PR #5, opened 2018-09-12
Resolve unmet peer dependencies

### LINKED ISSUE #4: React and React-native versions in peerDependencies.
@kmagiera Is there any specific version which should be kept in `package.json`?
With the latest upgrade of `react-navigation` some warnings appeared.

react-navigation/react-navigation#4982

dsznajder:
I do not see any specific features (maybe I missed something), but it seems reasonable to go for every `react` and `react-native` version if `react-navigation` does.

resolves #4

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,8 +36,8 @@
   "homepage": "https://github.com/kmagiera/react-native-screens#readme",
   "dependencies": {},
   "peerDependencies": {
-    "react": "16.0.0-alpha.6",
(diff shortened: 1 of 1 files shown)


LinusU:
Would you mind releasing a version with this patch in it?

kmagiera:
Just did -> alpha.10 includes it

LinusU:
Thanks!

PR #12, opened 2018-09-17
fixes setActive method when the param is not int type in Android

fangasvsass:
setActive method will receive double type,Double cannot be cast to java.lang.Integer


fangasvsass:
it will crash when navigate to any page in Android, Would you mind releasing a version for this

kmagiera:
Just published alpha.12 that includes this fix @fangasvsass

PR #14, opened 2018-09-19
Add typescript definitions

ckknight:
Only type I'm worried about is `onComponentRef`. It's a callback that takes an instance of a `React.Component`, but that instance is typed as `any` within here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "precommit": "lint-staged"
   },
   "main": "src/screens.js",
+  "types": "src/screens.d.ts",
   "files": [
     "src/",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -11,6 +11,7 @@
     "precommit": "lint-staged"
   },
   "main": "src/screens.js",
+  "types": "src/screens.d.ts",
```
kmagiera: We should add `src/screens.d.ts` to the "files" section below too in order for it to be included in the package sent to npm

kmagiera: Ah, wait, we actually have whole "src/" there so this should not be necessary

PR #22, opened 2018-10-02
 Fix for unsupported platforms

### LINKED ISSUE #13: Add screens.web.js with stubs for Screen and ScreenContainer
As per https://github.com/react-navigation/react-navigation/issues/5004 this would resolve react-navigation on web issues.

brunolemos:
Fixes #13 for real
Fixes https://github.com/react-navigation/react-navigation/issues/5004

Use different `index.js` files for supported and unsupported platforms.

Tested on iOS and on an existing web project that uses react-navigation.

---

'react-native' does not contain an export named 'requireNativeComponent'.

Necolas explained here that this import should not be on any file that runs on web: https://github.com/necolas/react-native-web/issues/507#issuecomment-307953785


kmagiera:
I actually went ahead and made these changes (now we only have `screens.web.js` and `screens.native.js` in `src/`)

I published it on NPM under 1.0.0-alpha.14, do you mind testing this with your react native web project when you get a minute @brunolemos ??

brunolemos:
> The reason is that packager picks screens.js

This is weird, the default packages prioritize `.xxx.js` over `.js`.

But ok, just tested your changes and it worked on web so all fine!

The way you did is cleaner, but I avoided doing this way because `.native.js` will also run on some other platforms like Windows (I think), and because now the fix is only targeting web specifically, not a fallback for "all unsupported platforms".

kmagiera:
Thanks for checking!

I wouldn't be concerned about other native platforms as long as they have `requireNativeComponent` method (I think only web is missing that). When the native component isn't available we are handling that gracefully as this package is not meant just yet to be obligatory. So there are people who don't have it installed and the native views are not available in which case we fallback to using Views.

PR #41, opened 2018-11-20
 Support project-wide properties, replace deprecated configuration

guix77:


joshuakelly:
+1 Having to update this myself right now.

PR #66, opened 2019-02-08
Add peek and pop

osdnk:
## Motivation
Peek and pop appears to be useful interaction and since it requires using UIControllers for registering previewing and controllers for managing peek-and-pop related actions. 

## Changes
Peek and pop interaction is handled by a special component `NativePeekableView` which renders nothing (actually renders `RNSPeekableWrapper`, which renders nothing – `view` is overrode natively) and does not impact layout changes. However, inside this component, we attach (in JS) a preview component which is natively connected for preview controller which we create for each instance of `NativePeekableView`. 

Actions of previewing are handled by `RNPreviewViewController` which delegated actions back to `RNSPeekableWrapper` and calls back JS.

In JS we manage preview component by state. On displaying native preview we render preview component and on disappear we hide preview component. 

Because it was needed to have a ref for wrapping screen, I've added `ScreenContext` which holds a reference for the outer screen component. 

Also, added example and upgraded RN – it was needed because of the static context, which was not available in the previous version of RN.


src/screens.native.js
```diff
@@ -86,3 +99,106 @@ export class ScreenContainer extends React.Component {
...
+          onPeek={this.onPeek}
+          onPop={this.props.onPop}
+          ref={this.preview}
+          previewActions={this.state.traversedActions}
+          onAction={this.onActionsEvent}>
+          <View style={StyleSheet.absoluteFill}>
+            {this.state.visible ? this.props.renderPreview() : null}
+          </View>
+        </NativePeekableView>
+      </React.Fragment>
```
adamczyk777: Why not `< />`? It's shorter, more readable and equivalent.

adamczyk777:
Forgot to mention, I get an error when running on android 😞 
```
FAILURE: Build failed with an exception.

* Where:
Script '/Users/jadamczyk/Projects/swmansion/react-native-screens/Example/node_modules/react-native/react.gradle' line: 95

* What went wrong:
A problem occurred configuring project ':app'.
> Could not get unknown property 'mergeResourcesProvider' for object of type com.android.build.gradle.internal.api.ApplicationVariantImpl.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 0s
```

That may be my fault but what I've done I just cloned the repo and checked out peek and pop and installed node modules

src/screens.native.js
osdnk: It leads to error in eslint ann popular editors e.g. webstorm

adamczyk777:
Also on iOS I got an error, let me know if it is my fault or what though. this is on iOS
image

osdnk:
Thanks, @adamczyk777!
Could you take a look now?

src/screens.native.js
```diff
@@ -86,3 +99,106 @@ export class ScreenContainer extends React.Component {
...
+        </View>
+        <NativePeekableView
+          onDisappear={this.onDisappear}
+          onPeek={this.onPeek}
+          onPop={this.props.onPop}
+          ref={this.preview}
+          previewActions={this.state.traversedActions}
+          onAction={this.onActionsEvent}>
+          <View style={StyleSheet.absoluteFill}>
+            {this.state.visible ? this.props.renderPreview() : null}
```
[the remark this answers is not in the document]
osdnk: Hm, I'd rather stay with it.

ios/RNSPeekableView.m
```diff
@@ -0,0 +1,166 @@
...
+  if (self = [super init]) {
+    _wrapper = wrapper;
+  }
+  return self;
+}
+
+- (UIViewController *)previewingContext:(id<UIViewControllerPreviewing>)previewingContext viewControllerForLocation:(CGPoint)location
+{
+  if (_wrapper.onPeek) {
+    _wrapper.onPeek(@{});
```
kmagiera: can't we just do `_wrapper.onPeek(nil)` instead of this?

ios/RNSPeekableView.m
```diff
@@ -0,0 +1,166 @@
...
+  }
+  return _wrapper.previewController;
+}
+
+- (void)previewingContext:(id<UIViewControllerPreviewing>)  previewingContext
```
kmagiera: formatting is broken here, too many whitespaces

ios/RNSPeekableView.m
```diff
@@ -0,0 +1,166 @@
...
+     commitViewController:(UIViewController *)viewControllerToCommit
+{
+  if (_wrapper.onPop) {
+    _wrapper.onPop(@{});
```
kmagiera: same comment

ios/RNSPeekableView.m
```diff
@@ -0,0 +1,166 @@
...
+  UIView *_child;
+}
+
+
+- (instancetype)initWithUIManager:(RCTUIManager *)uiManager
+{
+  if (self = [super init]) {
+    _uiManager = uiManager;
+  }
+  _previewController = [[RNPreviewViewController alloc] initWithWrapper:self];
```
kmagiera: this should be moved inside `if(self)` block

ios/RNSPeekableView.m
```diff
@@ -0,0 +1,166 @@
...
+  
+  return self;
+}
+
+- (void)setOnPop:(RCTDirectEventBlock)onPop
+{
+  _onPop = onPop;
```
kmagiera: do we really need all those methods here? seems like sth that should be synthetized automatically

src/peekableView.ios.js
```diff
@@ -0,0 +1,112 @@
...
+      mappedActions,
+    };
+  }
+
+  preview = React.createRef();
+  previewView = React.createRef();
+  childView = React.createRef();
+  componentDidMount() {
+    const childRef = React.Children.only(this.props.children)._nativeTag;
+    setImmediate(() => {
```
kmagiera: why setImmediate? Add comment if needed, preferably get rid of it

src/peekableView.ios.js
```diff
@@ -0,0 +1,112 @@
...
+      mappedActions
+    );
+    return {
+      traversedActions,
+      mappedActions,
+    };
+  }
+
+  preview = React.createRef();
+  previewView = React.createRef();
```
kmagiera: this is unused. I am confused now because it seem like children are what you seem to treat like preview. So maybe it'd make more sense not to call it `child` but `preview` which would also carry some additional semantics as opposed to "children" which does not say much about what is it used for in this case

ios/RNSPeekableView.m
osdnk: nah, it's needed

FrankFundel:
When will it be merged?

kmagiera:
Unfortunately we won't have capacity to work on this now. As it stayed open for way too long and many things have changed under the hood I will close the PR for now and hope we can revisit this topic in the future.

PR #68, opened 2019-02-12
Upgrade from android support v4 to androidX

bpeyrouBDHIS:
This PR aims to upgrade from android support v4 to androidX.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -6,7 +6,7 @@ buildscript {
     }
 
     dependencies {
-       classpath 'com.android.tools.build:gradle:3.1.4'
+       classpath 'com.android.tools.build:gradle:3.3.1'
     }
 }
 
@@ -18,8 +18,8 @@ def safeExtGet(prop, fallback) {
 }
 
 android {
-    compileSdkVersion safeExtGet('compileSdkVersion', 26)
-    buildToolsVersion safeExtGet('buildToolsVersion', '26.0.3')
+    compileSdkVersion safeExtGet('compileSdkVersion', 28)
+    buildToolsVersion safeExtGet('buildToolsVersion', '28.0.3')
 
     defaultConfig {
         minSdkVersion safeExtGet('minSdkVersion', 16)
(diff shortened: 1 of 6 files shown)


alanlima:
Hi @kmagiera 

I am just wondering if you have any date to merge this PR?

kmagiera:
Hey @bpeyrouBDHIS and @alanlima 

I'd like to consider this PR but not sure what kind of consequences it brings for end users. I see you update SDK version so I'm wondering how that'd work for people who are on SDK versions below 28 with their apps? Are we going to require them to upgrade their projects? What if they need to support older Android versions? What is the impact on the application footprint?

brunobar79:
@kmagiera AFAIK react-native v0.60 is gonna make the breaking change already so I suggest you to release it with a major version bump and a note in the readme ( RN < 0.6 use v1)

se-bastiaan:
Also see this discussion: https://github.com/react-native-community/discussions-and-proposals/issues/129

This change is required to support RN 0.60

radeno:
Guys, 1.0.0-aplha23 is backward incompatible with RN < 0.60 because of this PR.

You need to notify about it in README file.

radeno:
And what is much worse, react-navigation/native is unlocked to use any 1.0.0
https://github.com/react-navigation/native/blob/d54c3590825a3567f63fdea940a784699c653742/package.json#L95
So after `yarn upgrade`, all users without RN 0.60 will be hit, and not able compile on Android.

JustinKasad:
I am having the same issue!

I had to install using 
`yarn add https://github.com/kmagiera/react-native-screens#0627e742c8fd992148a17266dd39f8760142a16c` to get everything to work on RN 58.

I also had to do another hacky solution to fix this: 
https://github.com/kmagiera/react-native-screens/issues/124

@kmagiera Take a look at https://semver.org/
Proper versioning would alleviate these kinds of issues.

se-bastiaan:
This is why [jetifier](https://github.com/mikehardy/jetifier) exists. The migration to AndroidX is necessary to get to full support with RN0.60. However, if you want to stay on the support libraries the jetifier tool will make this possible via reverse-jetify.

cylim:
> This is why [jetifier](https://github.com/mikehardy/jetifier) exists. The migration to AndroidX is necessary to get to full support with RN0.60. However, if you want to stay on the support libraries the jetifier tool will make this possible via reverse-jetify. It is the tool recommended in [the release notes of RN0.60](https://facebook.github.io/react-native/blog/2019/07/03/version-60#androidx-support) as well.

This is for an AndroidX app to use not androidx libraries. It shouldn't work in the cases that our app is not androidx, but the library is androidx.

se-bastiaan:
I mentioned the [reverse option](https://github.com/mikehardy/jetifier#to-reverse-jetify--convert-node_modules-dependencies-to-support-libraries) which works fine for an app using support libraries that uses packages that are based on AndroidX.

PR #75, opened 2019-02-25
Update docs

scottmas:
The docs do not state that react-native-screens is also used by default for the tab and drawer navigators. I updated the docs to include that. I didn't see support in any of the other navigators that I looked at, but if there are any others, we should update the documentation for them as well.


brentvatne:
as per header on https://reactnavigation.org/docs/en/tab-navigator.html that navigator is deprecated and you should use https://reactnavigation.org/docs/en/bottom-tab-navigator.html / https://reactnavigation.org/docs/en/material-top-tab-navigator.html instead. screens are supported on those

scottmas:
Okay, that makes sense. Those navigators do not appear to have support for react-native-screens. Correct?

scottmas:
It makes more sense to me to just specify that all the navigators in react-navigation are supported. Is that the case for **all** the navigators though?

brentvatne:
yeah every navigator that is provided by the react-navigation org supports react-native-screens

scottmas:
Cool, so the readme text in the latest push should be accurate then.

PR #76, opened 2019-02-25
Rename screens.native.js -> screens.js

osdnk:
Eslint's `import/no-unresolved` was appearing on importing `screens`. The reason was that platform-specific extensions are not well supported by eslint.

Renaming screens.native.js to screens.js does not comes with unwanted effects because on browsers it will be still fallbacking to `screens.web.js`.


osdnk:
Fixes https://github.com/kmagiera/react-native-screens/issues/62

jarvisluong:
An alternative solution that keeps the web support is to make two files screens.ios and android, then exports all from the .native.js file 🤔

gsklee:
@kmagiera would you like to take a look? Should be a quick one.

skleeschulte:
[Looking at the react-native docs](https://facebook.github.io/react-native/docs/platform-specific-code.html#native-specific-extensions-ie-sharing-code-with-nodejs-and-web), I think using the .native.js extension is the correct way. [This issue can be fixed by updating the eslint-config](https://github.com/kmagiera/react-native-screens/issues/62#issuecomment-495158875).

satya164:
I usually prefer `index.native.js` and `index.js` because webpack doesn't support `.web.js` by default anymore while RN supports `.native.js` by default, so it's less configuration.

kacperkapusciak:
@osdnk could we close this PR for being no longer relevant 2 years later?

PR #77, opened 2019-02-26
Add bluring inputs on disactivating screen

osdnk:
## Motivation
This a solution for https://github.com/kmagiera/react-native-screens/issues/67

Keyboard doesn't dismiss after navigating to another screen and back. Even on `Keyboard.dismiss()` call. It seems like without Screens keyboard is dismissed on every navigation.

Repo: [snack.expo.io/@andrey/keyboard-bug](https://snack.expo.io/@andrey/keyboard-bug)
Focus TextInput -> Next Screen -> Back -> Keyboard is not dismissing

## Changes
Before setting screen to inactive, I added walking through all subviews and checking whether it's TextInput. If yes, it's being blurred then.

### DIFF
--- a/ios/RNSScreen.m
+++ b/ios/RNSScreen.m
@@ -1,5 +1,6 @@
 #import "RNSScreen.h"
 #import "RNSScreenContainer.h"
+#import <RCTText/RCTBaseTextInputView.h>
 
 @interface RNSScreen : UIViewController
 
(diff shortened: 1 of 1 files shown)


buoyad:
I can confirm this fixes the problem in iOS for us. ping @kmagiera do you think this can be merged?

Hirbod:
@WoLewicki can this maybe land into master?

WoLewicki:
cc @osdnk

tboba:
Hi @osdnk!
This PR has come a long way before it was finally tested 😅 I've just checked the changes and I see that they work fine. Alas, I can't merge this, since we've got `hideKeyboardOnSwipe` prop already which does the same thing. Hence, I need to close this PR.

Thanks!

PR #82, opened 2019-03-18
Fix non ascii character in comment

mayt:
I was getting the following compilation errors when I pulled in this repo.

```
error: unmappable character for encoding ASCII
```

The root cause seems to be the character below was not from the ASCII set.


osdnk:
Are you sure? 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔

mayt:
Does android build build for you?

Running through a unicode converter
```
 // ignore \u{2013} layer type is controlled by `transitioning` prop
```

osdnk:
Will try

compojoom:
@osdnk - I get the same with android studio. Was driving me nuts...

mayt:
Thanks!

compojoom:
can we get alpha 23?

PR #85, opened 2019-03-27
fix Screen.java:80: error: unmappable character for encoding ASCII   …

compojoom:
…  // ignore ??? layer type is controlled by `transitioning` prop

When building our app with Android studio we were always getting the following error:
> Task :react-native-screens:androidJavadoc FAILED
node_modules/react-native-screens/android/src/main/java/com/swmansion/rnscreens/Screen.java:80: error: unmappable character for encoding ASCII
    // ignore ??? layer type is controlled by `transitioning` prop
              ^
The file is encoded in UTF-8 - I don't know why the error talks about ASCII, but it also seems that the minus on that line is not a minus from the default charset?


compojoom:
closing as I didn't see this: https://github.com/kmagiera/react-native-screens/pull/82

PR #88, opened 2019-04-11
[web] Added active prop

EvanBacon:
# Why

[`react-navigation-stack`](https://github.com/react-navigation/react-navigation-stack/pull/105) relies on Screens to be removed when the active prop is falsey.

# How

Set `display: none` when the `active` prop is `false` or an `Animated.Value` with a value of `0`.

Related: https://github.com/react-navigation/react-navigation-stack/pull/105

### DIFF
--- a/src/screens.web.js
+++ b/src/screens.web.js
@@ -1,6 +1,6 @@
 import debounce from 'debounce';
 import React from 'react';
-import { Animated, StyleSheet, View } from 'react-native';
+import { Animated, View } from 'react-native';
 
 let _shouldUseScreens = true;
 
@@ -24,27 +24,20 @@ function isAnimatedValue(value) {
 function isPropTruthy(prop) {
   let activeValue = prop;
   if (isAnimatedValue(prop)) {
-    activeValue = !!prop.__getValue();
+    activeValue = prop.__getValue();
   }
 
-  return activeValue !== undefined && activeValue;
+  return !!activeValue;
 }
 
 export class Screen extends React.Component {
   listenerId = null;
 
-  state = { isActive: null };
-
   constructor(props) {
     super(props);
 
     this._onAnimatedValueUpdated = debounce(this._onAnimatedValueUpdated, 10);
-
     this._addListener(props.active);
-
-    this.state = {
-      isActive: isPropTruthy(props.active),
-    };
   }
 
   componentWillUnmount() {
@@ -52,6 +45,10 @@ export class Screen extends React.Component {
   }
 
   _addListener = possibleListener => {
+    if (this.listenerId)
+      throw new Error(
+        'Screen: Attempting to observe an animated value while another value is already observed.'
+      );
     if (isAnimatedValue(possibleListener)) {
       this.listenerId = possibleListener.addListener(
         this._onAnimatedValueUpdated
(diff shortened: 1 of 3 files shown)


src/screens.web.js
```diff
@@ -1,16 +1,38 @@
...
 }
 
-export const Screen = Animated.View;
+export class Screen extends React.Component {
+  render() {
+    const { style, active, ...rest } = this.props;
+    let viewStyle = [style];
+    let activeValue = active;
+    // Convert animated value to boolean.
+    if (active && active.__getValue) {
```
kmagiera: Looks like this relies on rerender happening when animated value changes. Is that what you wanted here? Maybe we should set animated listener and update `display` style in it instead?

src/screens.web.js
```diff
@@ -1,19 +1,110 @@
...
+    }
+  };
+
+  render() {
+    const { style, ...rest } = this.props;
+    const { isActive } = this.state;
+    let viewStyle = style;
+
+    if (!isActive) {
+      viewStyle = [style, styles.none];
```
kmagiera: is there no way we can control `display` prop other than re-rendering? Would it be possible to have `display = new Animated.Value('block')` and then set it with `display.setValue('none')`?

I'm pretty certain that with native driver only numeric values are supported with animated, but on web we don't use native driver so maybe we could do this? It would simplify this component a bit as we wouldn't need to use state at all.

What do you think?

EvanBacon: So I would hijack incoming animation values and update the new display animation value with their value? Wouldn't it also need to observe if the incoming value was just truthy?

ide: > So I would hijack incoming animation values and update the new display animation value with their value?

Sort of, but I don't think hijack is the right word since the parent of this component expects it to do whatever it wants with the `isActive`prop. This component's `isActive` prop can either be a regular value or an AnimatedValue, and in the latter case, this component subscribes to changes in the AnimatedValue. In either case, this component would trigger a separate animation that "animates" `block` to `none` immediately.

I put "animates" in quotes because this second, internal AnimatedValue is not used for animation. It is used because AnimatedValues bypass React's rendering and reconciliation. TBH something like `setNativeProps` would also work except for the fact that I'm not sure it handles StyleSheet values that could be in `this.props.style`.

src/screens.web.js
```diff
@@ -1,19 +1,110 @@
...
+import React from 'react';
+import { Animated, StyleSheet, View } from 'react-native';
+
+let _shouldUseScreens = true;
 
 export function useScreens(shouldUseScreens = true) {
   if (shouldUseScreens) {
-    console.warn('react-native-screens is not support on this platform.');
+    console.warn(
+      'react-native-screens is not fully support on this platform yet.'
```
ide:
```suggestion
      'react-native-screens is not fully supported on this platform yet.'
```

src/screens.web.js
kmagiera: do you want to try this other approach @EvanBacon ? otherwise this looks fine, just wanted to see if this implementation could be less complicated

EvanBacon: AFAICT the `Animated.Value` in react-native-web only animates numeric values.

PR #93, opened 2019-05-03
fix: memory leak on view <-> controller

### LINKED ISSUE #92: Memory leak in iOS
### bug
Used by in iOS will cause  memory leak.

image

image

### env
react-native: 0.59.4
react: 16.8.3
react-navigation: 3.5.1
react-native-screens: 1.0.0-alpha.22

### demo
```js
import React from "react";
import { View, Text } from "react-native";
import { createStackNavigator, createAppContainer } from "react-navigation";
import { useScreens } from 'react-native-screens';
useScreens();

class HomeScreen extends React.Component {
  render() {
    return (
      <View style={{ flex: 1, alignItems: "center", justifyContent: "center" }}>
        <Text>Home Screen</Text>
      </View>
    );
  }
}

const AppNavigator = createStackNavigator({
  Home: {
    screen: HomeScreen
  }
});

const AppContainer = createAppContainer(AppNavigator)

AppRegistry.registerComponent('AwesomeProject', () => AppContainer)
```

zhigang1992:
fix #92 and also fix memory leak on dev reload

### DIFF
--- a/ios/RNSScreenContainer.m
+++ b/ios/RNSScreenContainer.m
@@ -166,6 +166,7 @@
 
 @implementation RNSScreenContainerManager {
   NSMutableArray<RNSScreenContainerView *> *_markedContainers;
+  __weak RNSScreenContainerView *_containerView;
 }
 
 RCT_EXPORT_MODULE()
@@ -175,7 +176,9 @@ RCT_EXPORT_MODULE()
   if (!_markedContainers) {
     _markedContainers = [NSMutableArray new];
   }
-  return [[RNSScreenContainerView alloc] initWithManager:self];
+  RNSScreenContainerView *containerView = [[RNSScreenContainerView alloc] initWithManager:self];
+  _containerView = containerView;
+  return containerView;
 }
 
 - (void)markUpdated:(RNSScreenContainerView *)screen
@@ -193,5 +196,10 @@ RCT_EXPORT_MODULE()
     });
   }
 }
+
+- (void)invalidate {
+    [_containerView.controller removeFromParentViewController];
+}
+
 
 @end
--- a/ios/RNSScreen.h
+++ b/ios/RNSScreen.h
@@ -7,7 +7,7 @@
 @interface RNSScreenManager : RCTViewManager
 @end
 
-@interface RNSScreenView : RCTView <RCTInvalidating>
+@interface RNSScreenView : RCTView
 
 @property (weak, nonatomic) UIView<RNSScreenContainerDelegate> *reactSuperview;
 @property (nonatomic, retain) UIViewController *controller;
--- a/ios/RNSScreen.m
+++ b/ios/RNSScreen.m
@@ -42,12 +42,6 @@
   return _reactSuperview;
 }
 
-- (void)invalidate
-{
-  _controller.view = nil;
-  _controller = nil;
-}
-
 - (void)notifyFinishTransitioning
 {
   [_controller notifyFinishTransitioning];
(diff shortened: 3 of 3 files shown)


ios/RNSScreenContainer.m
```diff
@@ -194,4 +197,9 @@ - (void)markUpdated:(RNSScreenContainerView *)screen
   }
 }
 
+- (void)invalidate {
+    [_containerView.controller removeFromParentViewController];
```
zhigang1992: Need to remove root controller on reload.

brentvatne:
cc @kmagiera

cmmartin:
Friendly nudge @kmagiera 🙂Can we get this merged? Thanks

kmagiera:
Thanks @zhigang1992 for the PR and thanks @brentvatne @cmmartin for remained.

I responded on the original ticket and I believe this change isn't necessary. It seem like it's only been reported that there is a retain cycle between RNScreen and RNScreenView but this is by design. When RNScreenView is detached we break the cycle and allow for both RNScreenView and RNScreen to be freed. I verified that by adding `NSLog` statements to `dealloc` in both. Can you please respond here or on the ticket #92 if you believe that there is in fact a memory leak issue and not just retain cycle (which isn't in fact a problem on its own)

cmmartin:
I am playing a video with `react-native-video`. When I leave the screen and unmount that video component, the audio continues to play. If I shake and reload in development, the audio will still continue meaning I will eventually have 10+ versions of the audio playing on top of each other despite no video component being mounted. 

I originally thought this was an issue with `react-native-video` and even reported it, but, as @zhigang1992 [pointed out](https://github.com/react-native-community/react-native-video/issues/107#issuecomment-488998356), if I disable `react-native-screens`, the issue no longer occurs.

I will fork and apply this patch and confirm it resolves my issue and report back

kmagiera:
Any results @cmmartin ?

cmmartin:
Yes, sorry for the delay. Applying this PR did resolve my issue.

FYI my issue only occurs in development. 

The screens seem to persist reloads. To reproduce...
1) Load a screen with a `Video` component from react-native-video
2) Shake -> Reload

The audio from the video will continue to play and the audio from the new video (after reload) will play on top of it. After a few reloads, the device will become sluggish due to lack of memory, I assume.

All resolved with the PR. I am using my fork for now, but +1 for merging this

r0b0t3d:
I got same issue, this fix works for me with `react-native-video`
@zhigang1992 Could you resolve conflicting files?

kmagiera:
Any chance you can check if the problem still persist? We'd recently landed some changes that may affect this behavior: 102880c18be86dba4f9cc57e28d2be2b64719051

Happy to look more into this again if the above change does not resolve the probelm.

palmtrae:
Possibly the same issue here with `react-native-video`, this involves modals specifically. 

The effect is as @cmmartin describes, if the user starts a video in the modal, and decides to close that modal, the video will continue playing even though the component should've been unmounted at this point. However, pushing and poping a new screen within the same stack does dealloc the video object as expected.

We got a modal heavy app in development, and each new modal that we create and then later dismiss will remain in memory indefinitely until the application crashes because of memory pressure.

This is what it looks like after opening and closing a completely empty modal 20+ times (two new RNSScreen objects is created each time, one for NavigationController, one for the actual screen.)

Screenshot 2020-02-20 at 09 59 38

kmagiera:
Happy to hear it's fixed. Spent some time investigaing this issue and managed to come up with a fix in #378. Thank you all again for your input and ideas.

PR #97, opened 2019-05-13
Fix type of active prop in typescript definitions

satya164:
The TS definitions specify the prop as a boolean, but passing a boolean causes a crash.

### DIFF
--- a/src/screens.d.ts
+++ b/src/screens.d.ts
@@ -3,13 +3,13 @@
 
 declare module 'react-native-screens' {
   import { ComponentClass } from 'react';
-  import { ViewProps } from 'react-native';
+  import { ViewProps, Animated } from 'react-native';
 
   export function useScreens(shouldUseScreens?: boolean): void;
   export function screensEnabled(): boolean;
 
   export interface ScreenProps extends ViewProps {
-    active?: boolean;
+    active?: 0 | 1 | Animated.AnimatedInterpolation;
     onComponentRef?: (view: any) => void;
   }
   export const Screen: ComponentClass<ScreenProps>;


src/screens.d.ts
```diff
@@ -3,13 +3,13 @@
...
   import { ComponentClass } from 'react';
-  import { ViewProps } from 'react-native';
+  import { ViewProps, Animated } from 'react-native';
 
   export function useScreens(shouldUseScreens?: boolean): void;
   export function screensEnabled(): boolean;
 
   export interface ScreenProps extends ViewProps {
-    active?: boolean;
+    active?: 0 | 1 | Animated.AnimatedInterpolation;
```
osdnk: why? Shouldn't it be just `0 | 1`?

satya164: We're using an animated value in React Navigation. It's also an `Animated.View` so it should support animated nodes: https://github.com/kmagiera/react-native-screens/blob/0d082ad98566925366e35b6f6835c6dfa3d33cd5/src/screens.native.js#L61

osdnk: only style properties are supported by reanimated.

satya164: @osdnk The type is for vanilla Animated :)

osdnk: only style properties are supported by Animated as well

satya164: @osdnk It can support any prop. Check this out https://snack.expo.io/SJLybJP2N

satya164: This line https://github.com/facebook/react-native/blob/ddce74388a5662b5307582ca2e05710224c74ba0/Libraries/Animated/src/createAnimatedComponent.js#L149

osdnk: But it still won't be supported by native drivers, am I right?

satya164: Yea, it won't be supported by native driver

PR #99, opened 2019-05-13
Enhancement: change to useScreens to enableScreens

kelset:
This wants to be a small PR to improve the usability of the library, based on this conversation: https://twitter.com/grifotv/status/1127847192067215360.

Since the release of RNS there has been a new major player in the React game: hooks. And sadly `useScreens` recalls too closely Hooks, and this can lead to misunderstanding.

Changing it to `enableScreens` will make the difference clear, but at the same time it will be a BREAKING CHANGE for everyone using the lib.

So if we prefer to keep it around as useScreens I'm ok with it too, and we can close this.

Also, I did some tweaks to the README + fix some typos.

### DIFF
--- a/src/screens.native.js
+++ b/src/screens.native.js
@@ -8,7 +8,7 @@ import {
 } from 'react-native';
 import { version } from 'react-native/Libraries/Core/ReactNativeVersion';
 
-let USE_SCREENS = false;
+let ENABLE_SCREENS = false;
 
 // UIManager[`${moduleName}`] is deprecated in RN 0.58 and `getViewManagerConfig` is added.
 // We can remove this when we drop support for RN < 0.58.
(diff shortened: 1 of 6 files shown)


src/screens.native.js
```diff
@@ -17,17 +17,17 @@ const getViewManagerConfigCompat = name =>
     ? UIManager.getViewManagerConfig(name)
     : UIManager[name];
 
-export function useScreens(shouldUseScreens = true) {
-  USE_SCREENS = shouldUseScreens;
-  if (USE_SCREENS && !getViewManagerConfigCompat('RNSScreen')) {
+export function enableScreens(shouldEnableScreens = true) {
```
brentvatne: we should continue to export `useScreens` since there are a lot of places that will refer to it like this and it doesn't hurt to export it as well

kelset: So, just for me to understand, you'd want to have both "names" exported? Meaning, you could then either use `useScreens` or `enableScreens`?

brentvatne: yep

kelset: ok - lemme tackle that now 👍

kelset:
@brentvatne I've re-added the `useScreens` just as a "proxy" for enable - I think we should keep it around but like, "without leaving a footprint" in the typing so that we will discourage people from using it.

But if you prefer I can re-add that too. LMK

brentvatne:
I think `enableScreens` / `useScreens` will soon be removed, perhaps when 1.0 is released. At that point we'll probably invert it so it's enabled by default and you need to call `disableScreens` to turn it off. I think we might as well just export types for both right now, and just annotate it to say that you probably want to use `enableScreens` if you use hooks.

kelset:
Ok, I've re-added the TS typing for useScreens - LMK if there is any other change needed.

jsamr:
Hey, can you bump a new release from this PR? Just being annoyed by eslint react hook complaining

>  React Hook "useScreens" cannot be called at the top level. React Hooks must be called in a React function component or a custom React Hook function

kelset:
oh lol I totally forgot about this PR XD

We should also have in the release notes that this change is out there probably.

PR #115, opened 2019-06-18
Add lazy requiring screens components

osdnk:
## Motivation
If screens are not used, linking shouldn't be required.

### DIFF
--- a/src/screens.native.js
+++ b/src/screens.native.js
@@ -17,7 +17,7 @@ const getViewManagerConfigCompat = name =>
     ? UIManager.getViewManagerConfig(name)
     : UIManager[name];
 
-export function useScreens(shouldUseScreens = true) {
+function useScreens(shouldUseScreens = true) {
   USE_SCREENS = shouldUseScreens;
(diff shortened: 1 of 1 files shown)


osdnk:
@satya164 @adamczyk777

src/screens.native.js
```diff
@@ -26,20 +26,33 @@ export function useScreens(shouldUseScreens = true) {
...
+  get NativeScreenContainer() {
+    NativeScreenContainerValue =
+      NativeScreenContainerValue ||
+      requireNativeComponent('RNSScreenContainer', null);
+    return NativeScreenContainerValue;
+  },
+};
+
+const AnimatedNativeScreen = Animated.createAnimatedComponent(
+  ScreensNativeModules.NativeScreen
```
satya164: Oops I missed this. So this Animated component also needs to be created lazily since you're not lazily initialising on render, but initialising on property access. Since `ScreensNativeModules.NativeScreen` is called in the module scope, the variable will be initialised as the module is loaded and will throw if the component wasn't linked.

satya164: I have addressed this in #117

PR #116, opened 2019-06-18
Lazily require the native components so an error is not thrown when not linked

satya164:


satya164:
Oops. Didn't see #115

PR #117, opened 2019-07-01
Add lazy requiring screens components

satya164:
This is a follow up to #115 which fixes the animated screen to be initialised lazily.

cc @adamczyk777 will be awesome if you can merge this and publish a new version

### DIFF
--- a/src/screens.native.js
+++ b/src/screens.native.js
@@ -30,8 +30,11 @@ function screensEnabled() {
   return USE_SCREENS;
 }
 
+// We initialize these lazily so that importing the module doesn't throw error when not linked
+// This is necessary coz libraries such as React Navigation import the library where it may not be enabled
 let NativeScreenValue;
 let NativeScreenContainerValue;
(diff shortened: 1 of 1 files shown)


src/screens.native.js
```diff
@@ -57,32 +69,52 @@ export class Screen extends React.Component {
...
-        <AnimatedNativeScreen
-          {...rest}
-          ref={this.setRef}
-          style={StyleSheet.absoluteFill}>
-          <Animated.View style={style}>{children}</Animated.View>
-        </AnimatedNativeScreen>
-      );
+      AnimatedNativeScreen =
+        AnimatedNativeScreen ||
+        Animated.createAnimatedComponent(ScreensNativeModules.NativeScreen);
```
kmagiera: it is a bad idea to use `createAnimatedComponent` inside render. It's like you'd define a component class in render method and then returned it

satya164: It's only initialized once. All the subsequent renders will use the already initialized version.

PR #130, opened 2019-07-15
Stop disabling interaction of newly added screen

osdnk:
## Motivation

In react navigation we want to keep recognising gesture even if second screen beneath is active (in stack impl). Previously it was working bc there was only one gesture recogniser for whole stack, but we want to get rid of this pattern in order to extend customisation. 

### How it was? Screen below is rendered is gesture state is active so it's get rendered for a while (or no) and then gesture is cancelled immediately
https://streamable.com/4bkmt

### How it is after this fix
https://streamable.com/dho7v

### DIFF
--- a/ios/RNSScreenContainer.m
+++ b/ios/RNSScreenContainer.m
@@ -102,30 +102,11 @@
   }
 
   // detect if new screen is going to be activated
-  BOOL activeScreenAdded = NO;
+  RNSScreenView* activeScreenAdded = NULL;
   for (RNSScreenView *screen in _reactSubviews) {
     if (screen.active && ![_activeScreens containsObject:screen]) {
-      activeScreenAdded = YES;
-    }
-  }
-
-  // if we are adding new active screen, we perform remounting of all already marked as active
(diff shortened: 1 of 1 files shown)


ios/RNSScreenContainer.m
```diff
@@ -102,33 +102,18 @@ - (void)updateContainer
...
-      }
-    }
-
-    // add new screens in order they are placed in subviews array
-    for (RNSScreenView *screen in _reactSubviews) {
-      if (screen.active) {
-        [self attachScreen:screen];
-      }
-    }
+    activeScreenAdded.userInteractionEnabled = NO;
```
kmagiera: This code has been here to prevent view we are sliding away from responding to touches. If you delete this you will be able to do double taps on buttons etc. which isn't desirable. Also if you delete code that trigger `willMoveToWindow` this will likely interfere with the way keyboard handling work. What needs to be tested is the case when you have a keyboard open and you push new screen and also the opposite case when you pop from screen with keyboard opened.

PR #175, opened 2019-10-08
Call update bounds after transition on iOS

osdnk:
We observe that bounds are not updating at all. Method `updateBounds` is not called at all while transitioning.

As a result yoga reevaluation is not performed so content has bad size i.e. it's not possible to scroll to the very end in ScrollView because a component has a bad size in comparison to the parent (it should have the exact same height).   

It can be observed in navigation-ex and it works better after my fix: https://streamable.com/miun2

### DIFF
--- a/ios/RNSScreen.m
+++ b/ios/RNSScreen.m
@@ -240,6 +240,10 @@
   }
 }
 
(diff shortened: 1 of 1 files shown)


osdnk:
https://streamable.com/miun2
after

PR #182, opened 2019-10-18
Fix RN version check when using master

janicduplessis:
When using RN from master the version is 0.0.0 so we can just check if minor is 0 (this code breaks if major gets bump already anyway).


osdnk:
Uuuuuuu

PR #183, opened 2019-10-18
Apply custom title font attributes to largeTitle

davrosull:
Currently on iOS when you set `titleColor` it is not applied to the `largeTitle`.

This PR will apply the `titleColor` and `titleFontFamily` to the `largeTitle`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,6 +218,18 @@ Allows for customizing font family to be used for back button title on iOS.
 
 Allows for customizing font size to be used for back button title on iOS.
 
+#### `largeTitle` (iOS only)
+
+When set to `true` it makes the title display using the large title effect.
+
+#### `largeTitleFontFamily` (iOS only)
+
+Customize font family to be used for the large title.
+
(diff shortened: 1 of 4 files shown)


janicduplessis:
I think those should be new props to control the large title font size / family / color. I think it is common to want at least a different font weight (font family change) for the large title.

davrosull:
Yep - fair call! I tossing up whether to do it that way or the way I did. Thanks for the feedback and I'll update the PR with those changes.

davrosull:
Added `largeTitleFontFamily` and `largeTitleFontSize` to adjust the largeTitle font and size.

 Still using `titleColor` to adjust the color which I think makes sense - I wouldn't imagine a lot of users wanting to have different colors. 

I haven't looked at `fontWeight` of the title in this PR though.

osdnk:
Lovely

592da:
what about font weight? :) @davrosull @osdnk

PR #189, opened 2019-10-22
Fix useScreens -> enableScreens renaming

kmagiera:
There was a bug in PR that introduces enableScreens method to replace useScreens. The bug was that the method did not end up being exported (we use module.exports and not export syntax). On top of that I'm adding a deprecation warning to useScreens method as it interferes with some react hooks tooling.


shamilovtim:
@kmagiera deprecation warning never showed in my Expo app

PR #212, opened 2019-11-05
Fix compile error on xcode 10

sunnylqm:
Fixes https://github.com/kmagiera/react-native-screens/pull/195#issuecomment-549146288


osdnk:
Thanks!

PR #215, opened 2019-11-06
Fix screen container after rnsscreen layout change.

kmagiera:
For native stack we introduced a new way of layouting screen controllers. We no longer rely on flexbox to layout them so that we can use native stack to provide a correct dimensions and safe area paddings. Unfortunately this change broke the old screen container. With this commit we are adapting layout change to old screen container that will now layout its direct children on the native side by making them take up the whole space of the container.

### DIFF
--- a/ios/RNSScreen.m
+++ b/ios/RNSScreen.m
@@ -8,33 +8,6 @@
 #import <React/RCTShadowView.h>
 #import <React/RCTTouchHandler.h>
 
-@interface RNSScreenFrameData : NSObject
-@property (nonatomic, readonly) CGFloat rightInset;
-@property (nonatomic, readonly) CGFloat topInset;
-@property (nonatomic, readonly) CGFloat bottomInset;
-@property (nonatomic, readonly) CGFloat leftInset;
-@property (nonatomic, readonly) CGFloat navbarOffset;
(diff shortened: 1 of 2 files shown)


ajsmth:
is it possible to expose the old container so that we can take advantage of flex and manage the layout of the screen containers?

kmagiera:
screen container (a.k.a stack) can be layed out using flex, this is only for individual screens so that they match the dimension of the container.

Or maybe you'd like individual screens to not take the full screen? If so you can use a `transparentModal` or `containedTransparentModal` presentation options and use flex to adjust position and dimension of the content of the screen (Screen component would still take the whole container but it'd be transparent).

If that does not help your use case I'd be interested in learning about what a specific use case for managing size of individual screens would be.

ajsmth:
thanks for the response - you're right I was referring to individual screens, but not the Stack container just the old ScreenContainer. I've tried as far back as 2.0.0.-alpha.2 and it seems to respect the layouts, but more recent updates (i'm guessing this one) has removed it as you've described above.

I'm primarily interested in taking advantage of the `active` prop and controlling the position and transformations of Screen components myself, perhaps this is not the right use case for this lib!

PR #232, opened 2019-11-24
Implement largeTitle on Android

janicduplessis:
This adds support for large header on Android, similar to what exists on iOS. The main issue is that we need a way to communicate scroll to `CoordinatorLayout`. Normally this is done using `NestedScrollView` or a `RecyclerView` as a direct child of `CoordinatorLayout`. This is however not really possible with the current setup and the way react native works.

I was able to get it working by making `ReactScrollView` inherit `NestedScrollView` to test but we will need to find a better solution as I don't think this is upstreamable. There might also be some subtle behavior differences when the `NestedScrollView` is not a direct child of the `CoordinatorLayout`. This is to be verified with a fully native app.

I need to investigate more how `NestedScrollView` works to forward scroll events to the parent `CoordinatorLayout`. Some options I see:

-  Have a custom ScrollView component that would extend `NestedScrollView`. Probably the easiest to implement but requires app changes to work.
- Traverse subviews to find instances of `ReactScrollView` and find a way to hook into its scroll events. We could then have some proxy `ScrollView` that we insert as a child of `CoordinatorLayout` that sends the scroll events the way it expects it. The issue is we'd have to know and re-traverse whenever the view hierachy changes.

Summary of the changes:

- Use `CollapsingToolbarLayout` to implement large title behaviors
- Adds support for existing `largeTitle` and `largeTitleFontFamily` props
- Add `largeTitleFontSize` but currently unused, `CollapsingToolbarLayout` only allows passing a resource id for styling titles. See if we can hack something by traversing subviews.
- Add new `collapsable` prop to make the header hide during scroll, not sure about the name. It was simple to add support for this with `CollapsingToolbarLayout` so I figured why not.

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.java
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.java
@@ -6,21 +6,21 @@ import android.os.Bundle;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
+import android.widget.FrameLayout;
 import android.widget.LinearLayout;
-
-import androidx.annotation.Nullable;
-import androidx.appcompat.widget.Toolbar;
-import androidx.coordinatorlayout.widget.CoordinatorLayout;
 
 import com.facebook.react.uimanager.PixelUtil;
 import com.google.android.material.appbar.AppBarLayout;
+import com.google.android.material.appbar.CollapsingToolbarLayout;
+
+import androidx.annotation.Nullable;
+import androidx.coordinatorlayout.widget.CoordinatorLayout;
 
 public class ScreenStackFragment extends ScreenFragment {
 
   private static final float TOOLBAR_ELEVATION = PixelUtil.toPixelFromDIP(4);
 
   private AppBarLayout mAppBarLayout;
-  private Toolbar mToolbar;
   private boolean mShadowHidden;
 
   @SuppressLint("ValidFragment")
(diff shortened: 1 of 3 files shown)


bartzy:
@kmagiera Is this something that you think can be merged eventually?

janicduplessis:
It still requires a significant amount of work / testing. I can't give a timeframe but I'll try working on this again when I have some free time.

Hirbod:
would be awesome to see this on android, even though its not platform specific

tboba:
Hi @janicduplessis, thank you for submitting this PR!
As I've started to work on the refreshed version of something similar, which is an ability to just select header type from Material 3* (You can take a look on those changes in #1874) I'll close this Pull Request, but before that let me answer on some of the obstacles you've mentioned:

> The main issue is that we need a way to communicate scroll to CoordinatorLayout. Normally this is done using NestedScrollView or a RecyclerView as a direct child of CoordinatorLayout. This is however not really possible with the current setup and the way react native works.

I've managed to make the title collapsible on scroll without using NestedScrollView, so I can say I've omitted this obstacle by using `nestedScrollEnabled` prop on ScrollView - huge credits to @grahammendick for suggesting that!

> I need to investigate more how NestedScrollView works to forward scroll events to the parent CoordinatorLayout. Some options I see:
> [...]
> Traverse subviews to find instances of ReactScrollView and find a way to hook into its scroll events. We could then have some proxy ScrollView that we insert as a child of CoordinatorLayout that sends the scroll events the way it expects it. The issue is we'd have to know and re-traverse whenever the view hierachy changes.

I was looking if this could be possible, but I couldn't find a way how can I change the expanded height of the Collapsing Toolbar Layout on the run manually 😕 Using `nestedScrollEnabled` on the ScrollView is still easier though.

> Add largeTitleFontSize but currently unused, CollapsingToolbarLayout only allows passing a resource id for styling titles. See if we can hack something by traversing subviews.

Actually we can change the appearance of the expanded toolbar title - the font size of this title can be changed using `CollapsingToolbarLayout#setCollapsedTitleTextSize`, but we will add this property (along with property to change the `scrollFlags` later - just as you've created `collapsable` prop). 

If you've got any questions, regarding the new implementation please let me know about them in #1874 - for now I'll close this PR.

* - You can find more information about the Top App Bar in Material 3 [here](https://m3.material.io/components/top-app-bar/overview).

janicduplessis:
@tboba Very nice, looking forward to it!

PR #252, opened 2019-12-17
Fix not displaying view in navigation

osdnk:
fixes https://github.com/kmagiera/react-native-screens/issues/208

For stack overriding `reactSetFrame` for `active` does nothing because `active` is always `NO`, so there's no regression.

But it fixes in other cases (like stack navigation - not native one).

### DIFF
--- a/ios/RNSScreen.m
+++ b/ios/RNSScreen.m
@@ -31,12 +31,6 @@
   return self;
 }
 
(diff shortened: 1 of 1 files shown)


vijaykishan312:
not properly work on navigation screen 
on goBack and navigate screen to another screen screen is white 
react-native-screens@2.7.0

PR #254, opened 2019-12-26
Move gestureEnabled config to screen instead of heade…

janicduplessis:
When you have 2 screens in a stack with the bottom one with gestureEnabled=false using the back gesture causes the screen to become unresponsive. This seems to be caused by setting `interactivePopGestureRecognizer.enabled = NO` as soon as the gesture starts. This causes the gesture to cancel immediately and leaves the screen in an unresponsive state.

```js
<Stack><Screen gestureEnabled={false} /><Screen /></Stack>
```

To fix this instead of using `interactivePopGestureRecognizer.enabled` we can leverage the existing delegate that we have in `RNScreenStack`. In `gestureRecognizerShouldBegin` we can check if the top screen has gestures enabled.

To make this simpler I moved the `gestureEnabled` config to `Screen` instead of `HeaderConfig`. I think it also makes more sense conceptually since the gesture is tied to the screen and not the header. It also simplifies the android code a bit. 

This is a breaking change.

**Update**

This now only moves the config to screen since a separate fix was merged for the bug.

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.java
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.java
@@ -52,6 +52,7 @@ public class Screen extends ViewGroup implements ReactPointerEventsView {
   private boolean mTransitioning;
   private StackPresentation mStackPresentation = StackPresentation.PUSH;
   private StackAnimation mStackAnimation = StackAnimation.DEFAULT;
+  private boolean mGestureEnabled = true;
 
   public Screen(ReactContext context) {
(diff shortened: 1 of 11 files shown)


osdnk:
According to our discussion, I'd prefer to have it split into two PRs:
1. fix of the bug
2. API discussion
The first thing will be merged immediately and won't be blocked by 2.

janicduplessis:
Rebased, now it only moves the config since the fix was merged in another PR.

PR #258, opened 2019-12-30
[ios] release view from controller when popped from stack

r0b0t3d:
This issue is related to #92.

**Issue**
-------
I got issue when playing video with `react-native-video`. The video keep playing (still hear sound) when go back.
```
react-native: 0.61.5
react-native-screens: 2.0.0-alpha.22
```
**Solution**
-------
When controller popped, try to assign view to nil to break the cycle.

### DIFF
--- a/ios/RNSScreenStack.m
+++ b/ios/RNSScreenStack.m
@@ -22,6 +22,7 @@
   NSMutableArray<RNSScreenView *> *_reactSubviews;
   NSMutableSet<RNSScreenView *> *_dismissedScreens;
   NSMutableArray<UIViewController *> *_presentedModals;
+  __weak UIViewController* recentPopped;
   __weak RNSScreenStackManager *_manager;
 }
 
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
`react-native-video` should probably also implement `didMoveToWindow` and `willMoveToWindow`, so that the video always pauses/stops when transitioning away.

Example swift code:
```swift
@objc(didMoveToWindow)
  override func didMoveToWindow() {
    super.didMoveToWindow()

    let isVisible = superview != nil && window != nil
    let isDetachedView = superview != nil && window == nil

    if (isVisible) {
      
    } else if (isDetachedView && !invalidated) {
      self.pause()
    // If the bridge invalidates it should stop playing
    } else if (isDetachedView && invalidated) {
      self.stop()
    }
  }

  override func willMove(toWindow newWindow: UIWindow?) {
    super.willMove(toWindow: newWindow)

    let wasDetached = superview != nil && window == nil && newWindow != nil
    if (wasDetached) {
     
    }
  }
```

PR #272, opened 2020-01-07
[android] Fix re-attaching ScreenContainer to window

graszka22:
Code to reproduce and test: https://snack.expo.io/@angelikaserwa/humiliated-bagel.
Switch to the Settings tab, then go back to the Home tab and press the details button. Nothing happens. It was because after re-attaching we were using a destroyed FragmentManager that was cached inside the ScreenContainer class. 
Then, when we go back from Details to Home screen, using a hardware back button, an exception occured: `The specified child already has a parent. You must call removeView() on the child's parent first`.
I fixed this by calling `removeAllViews()` when detaching container from window and forcing an update on re-attach.


r0b0t3d:
@graszka22 I tried your code but got this bug.
In Home tab, it is fine to switch to other tabs. But when come to detail view and go back, Home tab is blank.

device-2020-01-18-113036

PR #285, opened 2020-01-17
Revert "[Android] Bottom tab not update"

kmagiera:
Reverts kmagiera/react-native-screens#275

#275 has been merged prematurely. The PR did not provide any description on why the proposed changes make sense and they seem to be causing weird effects. We were aware of the issue with destroyed fragment manager and I believe #272 should properly address it. I'm going to revert #275 in favor of merging #272


kmagiera:
cc @osdnk @r0b0t3d

r0b0t3d:
It is good to me. Will try and let you know

PR #307, opened 2020-02-05
Apply custom color to largeTitle

v-fernandez:
Currently you cannot set a different color to largeTitle then the one applied to the non large title. This PR adds `largeTitleColor` that allows setting the color for the largeTitle, following the existing `largeTitleFontFamily` and `largeTitleFontSize`.


kmagiera:
Thanks for sending this. Would be great if we could keep using titleColor for large title in case largeTitleColor isn't set. I think it is a reasonable default in most of the cases to keep the color the same. Now when largeTitleColor isn't provided we'd use platform default one. WDYT?

v-fernandez:
@kmagiera yes, that makes sense. I'll update the PR. Thanks for the feedback!

v-fernandez:
@kmagiera ok, changes made.

PR #318, opened 2020-02-10
feat: Add in additional modal types for iOS

matt-oakes:
Allows you to choose additional modal presentation styles for iOS. It adds the ability to force a full screen modal or choose a "form sheet" style. This only affects iOS.


wibb36:
@matt-oakes I've been very interested in this. Could you share an example in code on how to make that work?

matt-oakes:
@wibb36 Yes, there should have been an example here.

All you need to do is set the `stackPresentation` option of the screen in a native stack to `formSheet` to use the form sheet behaviour. This is all iOS only.

PR #323, opened 2020-02-11
fix(android): NullPointerException crash due to null top screen

matt-oakes:
The library would crash when the `mTopScreen` instance variable is null. This change checks if it null before calling the `getScreen()` method. It also marks the method as having a nullable return value. The only place this method is used is already setup to handle `null` responses.


kmagiera:
Thanks!

PR #325, opened 2020-02-12
Fix gestureEnabled for modal with no header on iOS

janicduplessis:
Since the config is only updated in the HeaderConfig view it will not get executed for stacks with no header. This fixes it by setting the prop directly in the Screen when the gestureEnabled prop is set.

### DIFF
--- a/ios/RNSScreen.m
+++ b/ios/RNSScreen.m
@@ -109,6 +109,17 @@
   }
 }
 
+- (void)setGestureEnabled:(BOOL)gestureEnabled
+{
+  #ifdef __IPHONE_13_0
+    if (@available(iOS 13.0, *)) {
+      _controller.modalInPresentation = !gestureEnabled;
+    }
+  #endif
+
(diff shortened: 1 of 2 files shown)


kmagiera:
Thanks for working on this! Based on the description it looks like you have problems with some config params which are applicable even when no header is set. The goal of `updateViewController` where you're making changes was to handle those type of settings before it returns. It is then surprising to me that it isn't working. Do you have any insights on why we can't have it in `updateViewController` (it should not be tied to header)? I'd prefer to keep that code in this method instead of having it distributed in a variety of places

janicduplessis:
@kmagiera The thing is `updateViewController` is only called when the screen has a ScreenStackHeaderConfig component. Gestures aren't really tied to the header (it should be possible to disable them on a screen with no header). I already moved the config to the Screen component in a previous PR (https://github.com/software-mansion/react-native-screens/pull/254) but this was left as part of the header config as I wasn't sure what it did.

PR #327, opened 2020-02-13
Fix screen already added issue

adamczyk777:
In order to fix errors with messages like 'Screen already added' taht happend when dismissing modal screens from JS Stack Navigators, I removed `mActiveScreenFragments`.  Instead, I used `mFragmentManager.getFreagments()`. FragmentManager manages them itself, there is no need for duplication.


adamczyk777:
Closing this pull request as @kmagiera already applied the changes we were discussing 😄

PR #330, opened 2020-02-14
Trigger transition finish event from native stack

kmagiera:
This change adds new event that gets dispatched from native stack when screen transitioning is finished.

The new event is used by react-navigation to update the library internal state that the triggered action has been finished. Previously we were relying solely on onAppear and onDisappear events, however those does not get triggered when new iOS 13 modals are used or when transparent modals are displayed. This is because with transparent modals the view below is still visible. We therefore needed another way of notifying the library that screen transition have finished despite the fact that disappear event couldn't be triggered.

As a part of this change I also refactored invalid ref cycle-break code on iOS which was ought to remove reference cycle between view and view controller. This code have been moved to viewWillDisappear callback.

Also on Android part small refactoring has been done and we removed the necessity to keep mActiveScreens array which was occasionally getting out of sync with the list of active fragments.

### DIFF
--- a/createNativeStackNavigator.js
+++ b/createNativeStackNavigator.js
@@ -27,16 +27,34 @@ function renderComponentOrThunk(componentOrThunk, props) {
 
 class StackView extends React.Component {
   _removeScene = route => {
+    console.warn('REMOVE SCENE');
     this.props.navigation.dispatch(StackActions.pop({ key: route.key }));
   };
 
-  _onSceneFocus = (route, descriptor) => {
+  _onAppear = (route, descriptor) => {
     descriptor.options &&
       descriptor.options.onAppear &&
       descriptor.options.onAppear();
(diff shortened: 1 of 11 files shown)


createNativeStackNavigator.js
```diff
@@ -27,18 +27,36 @@ function renderComponentOrThunk(componentOrThunk, props) {
...
-      StackActions.completeTransition({ toChildKey: route.key })
+      StackActions.completeTransition({
+        toChildKey: route.key,
+        key: this.props.navigation.state.key,
+      })
     );
   };
 
+  _onFinishTransitioning = () => {
+    console.warn('FINISH TRANSITIONING');
```
beetlebum: @kmagiera looks like this was left here accidentally, it's showing on my logs since I updated earlier today

PR #331, opened 2020-02-15
Fix several crashes in native stack related to fragment library

kmagiera:
This change fixes two crashes related to fragment library. The first issue was a crash caused by return transaction started while the previous transaction was ongoing (e.g., quickly adding new screen on top of the stack and immediately dismissing it). The main fix was applied in the fragment library and therefore as a part of this change we update the dependency to fragment:1.2.1 which is the current latest stable version. As a result of the fragment library change we started observing other issue. The second issue was caused by the fact that under certain circumstances the view associated with a fragment couldn't been added despite it still being attached to a parent. This was resulting in a crash. This change adds a cleanup code that properly detaches the view: we do it in onCreateView but also when the fragment destroys its view (in onViewDestroy callback). The latter is necessary because when fragments are restored the order of onCreateView calls is reversed which causes inner views to attach despite their fragment managers not being initialized.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -49,8 +49,9 @@ repositories {
 dependencies {
     implementation 'com.facebook.react:react-native:+'
     implementation 'androidx.appcompat:appcompat:1.1.0'
-    implementation 'androidx.coordinatorlayout:coordinatorlayout:1.0.0'
-    implementation 'com.google.android.material:material:1.0.0'
+    implementation 'androidx.fragment:fragment:1.2.1'
+    implementation 'androidx.coordinatorlayout:coordinatorlayout:1.1.0'
+    implementation 'com.google.android.material:material:1.1.0'
 }
 
 def configureReactNativePom(def pom) {
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
@@ -220,13 +220,6 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
   protected void onDetachedFromWindow() {
     super.onDetachedFromWindow();
     mIsAttached = false;
-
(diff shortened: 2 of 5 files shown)


Bardiamist:
@kmagiera Can it be related?

ApplicationNotResponding
Application Not Responding for at least 4000 ms.

```
com.facebook.react.bridge.ReadableNativeMap in importKeys
com.facebook.react.bridge.ReadableNativeMap in getLocalMap at line 47
com.facebook.react.bridge.ReadableNativeMap in access$000 at line 23
com.facebook.react.bridge.ReadableNativeMap$ReadableNativeMapKeySetIterator in <init> at line 243
com.facebook.react.bridge.ReadableNativeMap in keySetIterator at line 194
com.facebook.react.animated.PropsAnimatedNode in <init> at line 36
com.facebook.react.animated.NativeAnimatedNodesManager in createAnimatedNode at line 87
com.facebook.react.animated.NativeAnimatedModule$4 in execute at line 208
com.facebook.react.animated.NativeAnimatedModule$3 in execute at line 154
com.facebook.react.uimanager.UIViewOperationQueue$UIBlockOperation in execute at line 498
com.facebook.react.uimanager.UIViewOperationQueue$1 in run at line 779
com.facebook.react.uimanager.UIViewOperationQueue in flushPendingBatches at line 888
com.facebook.react.uimanager.UIViewOperationQueue in access$2200 at line 42
com.facebook.react.uimanager.UIViewOperationQueue$2 in runGuarded at line 846
com.facebook.react.bridge.GuardedRunnable in run at line 29
android.os.Handler in handleCallback at line 883
android.os.Handler in dispatchMessage at line 100
android.os.Looper in loop at line 237
android.app.ActivityThread in main at line 7811
java.lang.reflect.Method in invoke
com.android.internal.os.RuntimeInit$MethodAndArgsCaller in run at line 493
com.android.internal.os.ZygoteInit in main at line 1076
```

PR #345, opened 2020-02-18
Add native stack

satya164:


kmagiera:
Thanks for this. Any pointers on what we need to do when publishing? Or should we just publish typescript files without any build steps?

satya164:
I think for now publishing as is is ok. It's tricky to build with multiple entry points in project root.

Hirbod:
I hope this one lands on time into EXPO SDK 37, which is around the corner.

PR #350, opened 2020-02-20
Enable to control content insets in Android Toolbar from React Native

### LINKED ISSUE #332: [Android] Header does not have left and right padding on Android
Hi. 

I'm using `@react-navigation/native-stack` in my app, and I have an issue with Header on Android. When the header doesn't show back button, the title is stuck to the left side of the screen without any padding. 

The header right can be solved with CSS, but I couldn't find a way to do the same with title on Android.

Thank you

majkpal:
From now on you can set inset on the left and right side of toolbar when there are not additional actions and inset between action button and toolbar content, keep in mind that it still has space that is forced by ImageView containing icon. 

Also added fixes for few typos in docs 😄 
resolves #332

### DIFF
--- a/Example/nativeStack.js
+++ b/Example/nativeStack.js
@@ -64,7 +64,11 @@ export class Stack extends Component {
         stackAnimation="fade"
         active={1}
         onDismissed={() => this.removeByKey(key)}>
-        <ScreenStackHeaderConfig title={key}>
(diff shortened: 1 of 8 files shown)


kmagiera:
Closing this in favor of #396 – however would like to get those typo updates submitted in a separate PR if you will.

PR #391, opened 2020-02-26
Fix wrong import for react-navigation in v4 stack

satya164:
Navigators in v4 need to import from 'react-navigation' package instead of '@react-navigation/xxx'.

### DIFF
--- a/createNativeStackNavigator.js
+++ b/createNativeStackNavigator.js
@@ -5,7 +5,7 @@ import {
   SceneView,
   StackActions,
   createNavigator,
-} from '@react-navigation/core';
+} from 'react-navigation';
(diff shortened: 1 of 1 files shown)


kmagiera:
weirdly I'm using this in two projects and works fine. Care to explain why this change is needed?

satya164:
It works fine because `react-navigation` has a dependency on it, and yarn v1/npm flatten packages, so it can be resolved from `node_modules`. However, users don't add it to their `package.json`, so technically there's no guarantee that this package will be there in every scenario.

Also if we decide to vendor these stuff and not depend on `@react-navigation/core` anymore (which I'm considering), this will break since the package won't be there.

kmagiera:
Cool, just checked and appear to be working with this change as well.

PR #406, opened 2020-03-05
[docs] Remove unnecessary gradle dependencies

radko93:
According to #397, we don't need those dependencies anymore.


kmagiera:
Thanks!

PR #429, opened 2020-03-20
Add headerLeft function to render left side of header

zcmgyu:
I don't really understand why don't you implement the `headerLeft` function but I think it's really necessary when designing UI.

Screen Shot 2020-03-21 at 1 44 32 AM

### DIFF
--- a/native-stack/README.md
+++ b/native-stack/README.md
@@ -83,6 +83,10 @@ Title string used by the back button on iOS. Defaults to the previous scene's `h
 
 Whether the back button title should be visible or not. Defaults to `true`. Only supported on iOS.
 
+#### `headerLeft`
+
+Function which returns a React Element to display on the left side of the header.
+
 #### `headerRight`
 
 Function which returns a React Element to display on the right side of the header.
--- a/native-stack/types.tsx
+++ b/native-stack/types.tsx
@@ -110,6 +110,10 @@ export type NativeStackNavigationOptions = {
    */
   headerLargeTitle?: boolean;
   /**
+   * Function which returns a React Element to display on the left side of the header.
+   */
+  headerLeft?: () => React.ReactNode;
+  /**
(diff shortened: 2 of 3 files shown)


kmagiera:
I think this was due to some inconsistencies between iOS and Android. Not sure if those are still present but from what I remember headerLeft component would replace the back button on iOS (you need to use it on a second screen on the stack because root screen does not have back button) whereas on Android it renders next to the back button.

I'd be happy to merge this if this behavior is consistent. Can you please share screenshots of how it look on Android and iOS when you use it on a screen pushed on stack (not just the root screen)

zcmgyu:
Currently, I'm not working on Android. Let's me check and response ASAP

giaset:
Hey all, any progress on this? Would be awesome if this was merged soon, as having a custom component on the left side of the nav bar is a pretty common use case :)

quentez:
I've been trying to implement a menu button to the left of the title similar to this:

image

Would `headerLeft` be needed in that case or is that another way to achieve this?

Abhishek12345679:
do y'all have any plans to merge it ?

lucatk:
@kmagiera Please merge this. It is basic functionality which this native wrapper for iOS and Android screens needs to feature.
On current Android SDK version 28, this replaces the Android back button just like on iOS.

WoLewicki:
@zcmgyu thanks for your contribution! I didn't notice your PR and added `headerLeft` with some more info along with `headerCenter` and some other things in #457 while working on back button visibility. There are still some inconsistencies between Android and iOS. On Android, `headerLeft` makes the title also disappear in nested screens (It can be seen https://github.com/software-mansion/react-native-screens/pull/429#issuecomment-614605353) because of the weird positioning of that title otherwise, which doesn't happen on iOS. If someone would like to look into it, it would be great!

WoLewicki:
Closing this since the PR is outdated.

Issue #410, opened 2020-03-09
error in running project: Could not find swiperefreshlayout.jar (androidx.swiperefreshlayout:swiperefreshlayout:1.0.0).


MMiladSSanati:
I am trying to install react-native-screens in my project. I installed it without any problem but when i want to run the app, the following problem will occur:

`
Executing tasks: [:app:assembleDebug] in project D:\Programming\Projects\Sharmarket\mobile app\v1.0.0\Sharmarket\android

Gradle Daemon started in 3 s 898 ms
> Task :app:generatePackageList
> Task :app:preBuild
> Task :react-native-screens:preBuild UP-TO-DATE
> Task :react-native-screens:preDebugBuild UP-TO-DATE
> Task :react-native-screens:checkDebugManifest UP-TO-DATE
> Task :react-native-screens:processDebugManifest UP-TO-DATE
> Task :app:preDebugBuild
> Task :react-native-screens:compileDebugAidl NO-SOURCE
> Task :app:compileDebugAidl NO-SOURCE
> Task :react-native-screens:packageDebugRenderscript NO-SOURCE
> Task :app:checkDebugManifest
> Task :app:compileDebugRenderscript NO-SOURCE
> Task :app:generateDebugBuildConfig
> Task :app:bundleDebugJsAndAssets SKIPPED
> Task :app:prepareLintJar UP-TO-DATE
> Task :app:generateDebugSources
> Task :react-native-screens:generateDebugBuildConfig UP-TO-DATE
> Task :react-native-screens:compileDebugRenderscript NO-SOURCE
> Task :react-native-screens:generateDebugResValues UP-TO-DATE
> Task :react-native-screens:generateDebugResources UP-TO-DATE
> Task :react-native-screens:packageDebugResources UP-TO-DATE
> Task :react-native-screens:generateDebugRFile UP-TO-DATE
> Task :react-native-screens:prepareLintJar UP-TO-DATE
> Task :react-native-screens:generateDebugSources UP-TO-DATE
> Task :react-native-screens:javaPreCompileDebug FAILED

FAILURE: Build failed with an exception.

 What went wrong:
Execution failed for task ':react-native-screens:javaPreCompileDebug'.
> Could not resolve all files for configuration ':react-native-screens:debugCompileClasspath'.
   > Could not find swiperefreshlayout.jar (androidx.swiperefreshlayout:swiperefreshlayout:1.0.0).
     Searched in the following locations:
[...]

PR #439, opened 2020-03-25
Fix Could not find swiperefreshlayout.jar #410

loverdeveloper:
There is an error during the build about swiperefreshlayout.jar,
it's fixed by upgrade Swiperefreshlayout version
#410


WoLewicki:
Is this change needed? I cannot repro this problem. Isn't the problem with one's project rather than the library?

WoLewicki:
Closing due to no response. Feel free to comment in order to reopen.

PR #453, opened 2020-04-05
Don't clip custom header subviews on Android

janicduplessis:
This is mostly noticeable when using icons with ripple overflow.

Before

After


chirag04:
@WoLewicki not a pressing issue but was wondering if there were plans to cut a new release any time soon?

PR #454, opened 2020-04-05
Fix: add missing largeTitleBackgroundColor option

agarant:
Set the largeTitleBackgroundColor option.

PR #455, opened 2020-04-06
feat: support header title color

agarant:
Currently it was not possible to set a title color different than the tint color. This PR keeps the tintColor as the default header title color but also uses the headerStyle.color if defined.

### DIFF
--- a/createNativeStackNavigator.js
+++ b/createNativeStackNavigator.js
@@ -90,7 +90,7 @@ class StackView extends React.Component {
       translucent: translucent === undefined ? false : translucent,
       title,
       titleFontFamily: headerTitleStyle && headerTitleStyle.fontFamily,
-      titleColor: headerTintColor,
+      titleColor: (headerTitleStyle && headerTitleStyle.color) || headerTintColor,
       titleFontSize: headerTitleStyle && headerTitleStyle.fontSize,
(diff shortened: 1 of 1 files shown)


Collin3:
I think you should be able to achieve this by setting your screen options to something like the object below. This is how I was able to style the back button, but set the title text to a different color. I'm using the latest version (2.5.0).
```js
{
  headerTintColor: yourTintColorHere,
  headerTitleStyle: {
    color: yourTitleColorHere
  }
}
```

agarant:
@Collin3 Thanks for suggesting a workaround.

However, I still think that the fix is relevant as it is makes the api more intuitive and consistent. @WoLewicki do you mind taking a quick look at it. It is a 1 line PR. :)

WoLewicki:
Thanks for your contribution! Looks good, I am merging it.

Hirbod:
Maybe a hint in the native-stack readme?

WoLewicki:
Could someone make a PR with an update to the README and screens.d.ts that it would properly show the available options based on `createNativeStackNavigator.js`? I would love to merge a PR with it.

agarant:
@WoLewicki of course, I can make the update in #454, updating the required docs and types for the largetTitle property at the same time.

Issue #445, opened 2020-03-29
Include back button in ScreenStackHeaderLeftView


kyle-ssg:
Firstly, this lib is awesome.

When defining a custom ScreenStackHeaderLeftView, I feel that a lot of the times I'd want to still have the native back button where applicable. If I wish to be able to both go back and have a custom view next to it.

Also perhaps a side question on this, would I be able to detect if the stack can go back?

Edit: figured the last point out already - thanks

WoLewicki:
@kyle-ssg can you check if using `backButtonInCustomView` from #457 set to `true` gives you what you want?

WoLewicki:
Ok so I am closing this issue. Feel free to comment if there is something wrong.

PR #457, opened 2020-04-06
feat: add types for header views and back button visibility prop

WoLewicki:
Added type definitions for `headerLeft` and `headerCenter` and an option allowing the custom left view to be rendered to the right of the back button requested in #445.

### DIFF
--- a/native-stack/README.md
+++ b/native-stack/README.md
@@ -91,6 +91,14 @@ Whether the back button title should be visible or not. Defaults to `true`. Only
 
 Function which returns a React Element to display on the right side of the header.
 
+#### `headerLeft`
+
+Function which returns a React Element to display on the left side of the header. For now, on Android, using it will cause the title to also disappear.
+
+#### `headerCenter`
+
+Function which returns a React Element to display in the center of the header.
+
 #### `headerTranslucent`
 
 Boolean indicating whether the navigation bar is translucent. Only supported on iOS.
--- a/native-stack/README.md
+++ b/native-stack/README.md
@@ -2,9 +2,9 @@
 
 Provides a way for your app to transition between screens where each new screen is placed on top of a stack.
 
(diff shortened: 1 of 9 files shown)


native-stack/views/HeaderConfig.tsx
```diff
@@ -73,6 +79,14 @@ export default function HeaderConfig(props: Props) {
       {headerRight !== undefined ? (
         <ScreenStackHeaderRightView>{headerRight()}</ScreenStackHeaderRightView>
       ) : null}
+      {headerLeft !== undefined ? (
```
satya164, line 82: The reason these were not exposed is because of inconsistent behaviour between Android and iOS. Can you check if it still happens?

src/screens.d.ts
```diff
@@ -103,6 +103,10 @@ declare module 'react-native-screens' {
      * @description If set to true the back button will not be rendered as a part of navigation header.
      */
     hideBackButton?: boolean;
+    /**
+     * Whether to show the back button with custom left side of the header.
+     */
```
WoLewicki, line 108:
```suggestion
    /**
     * @host (iOS only)
     * @description Whether to show the back button with a custom left side of the header.
     */
```

native-stack/README.md
```diff
@@ -87,21 +91,29 @@ Whether the back button title should be visible or not. Defaults to `true`. Only
 
 Function which returns a React Element to display on the right side of the header.
 
+#### `headerLeft`
+
+Function which returns a React Element to display on the left side of the header. For now, on Android, using it will cause the title to also disappear.
```
chirag04, line 96: @WoLewicki any recommendation on how to implement a custom back button with a title on android? happy to send. PR but needs some guidance on how to proceed

WoLewicki: https://github.com/software-mansion/react-native-screens/pull/457/files#diff-d63d0fa38279f7aa0d11c045d518a176L242 removes the title. You can comment that line and see what happens and work on it to be handled properly.

PR #461, opened 2020-04-11
Simplify the implementation on web

satya164:
Use `Animated.createComponent` on web instead of adding listeners manually. This means:

- No extra listeners
- One less dependency (`debounce`)
- Smaller component hierarchy since `Animated.View` used the HOC anyway

This also fixes some issues on web:

- Fix `NativeScreen` ignoring the `active` prop
- Fix `enableScreens(false)` not respected
- Fix custom `display` value not being respected since it was set to `flex`
- Fix view not re-appearing sometimes when `active` state changes

I couldn't find the reasoning for the previous implementation (https://github.com/software-mansion/react-native-screens/pull/88), so I assumed that it was an oversight. Lemme know if I'm missing something.

### DIFF
--- a/src/screens.web.js
+++ b/src/screens.web.js
@@ -1,113 +1,35 @@
-import debounce from 'debounce';
 import React from 'react';
 import { Animated, View } from 'react-native';
 
-let _shouldEnableScreens = true;
+let ENABLE_SCREENS = true;
 
 export function enableScreens(shouldEnableScreens = true) {
-  if (shouldEnableScreens) {
-    console.warn(
-      'react-native-screens is not fully supported on this platform yet.'
-    );
-  }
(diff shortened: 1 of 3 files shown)


src/screens.web.js
```diff
@@ -1,113 +1,41 @@
-import debounce from 'debounce';
 import React from 'react';
 import { Animated, View } from 'react-native';
 
-let _shouldEnableScreens = true;
+let ENABLE_SCREENS = true;
 
 export function enableScreens(shouldEnableScreens = true) {
   if (shouldEnableScreens) {
     console.warn(
```
EvanBacon, line 8: Maybe we should get rid of this, or at least wrap it in `__DEV__`.

PR #477, opened 2020-04-24
Proper container children cleanup

kmagiera:
This change attempts to fix a problem when screen fragment views aren't properly detached from their parents despite `recycleView` being invoked on them.

The final outcome of the issue was that the app would crash with an error that view it tries to attach already has a parent. We observed this issue happening in production but with no clear repro steps the problem appeared to be a bit cryptic. Our investigation shown that `recycleView` does not reset the parent because the view is not listed in `mDisappearingChildren` in `ViewGroup`'s implementation.

We suspect that the above was due to us triggering `clearDisappearingChildren` directly in `Screen` and hence this PR updates that. 

The second change we make here is that we no longer attempt to reuse the main fragment view (crated by `onCreateView`). We used a custom container for stack fragments anyways and in this PR we adapt the same approach for `ScreenFragment` and make it so that the container view is not recycled (only the contents of the container that is managed by React Native is what we recycle).

Next, we update detach handler in containers. As we don't know the exact source of the parent reference cleanup we suspect that what is happening is that some children views remain attached to container past the child fragment lifecycle. This is due to the fact that fragments may perform transitions which may delay the process of child views being detached. As when the container is detached, its children are no longer visible we can safely drop all the children there. We restore fragment hierarchy when the container is attached anyways.
 
Lastly, we make small refactor in `ScreenContainer.java` which aims to extract fragment selection logic out of on attach handler. We may consider moving it elsewhere in the future or consider making it "asynchronous". This leads to some nullability check changes in that file as well as a new methods `setFragmentManager` being added.

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.java
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.java
@@ -107,12 +107,6 @@ public class Screen extends ViewGroup {
   }
 
   @Override
-  protected void onDetachedFromWindow() {
-    super.onDetachedFromWindow();
-    clearDisappearingChildren();
-  }
-
-  @Override
   protected void onAttachedToWindow() {
     super.onAttachedToWindow();
     // This method implements a workaround for RN's autoFocus functionality. Because of the way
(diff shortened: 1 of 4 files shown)


micabe:
nicely done! @kmagiera

PR #481, opened 2020-04-28
Fix user interface style in native stack view controllers

tsapeta:
# Why

In Expo client, which supports both light and dark styles, you can open an experience that has strict setting to light or dark mode. In that case experience's root view controller is set to the style specified by the user, however view controllers presented over it or added as a child have `overrideUserInterfaceStyle` set to `UIUserInterfaceStyleUnspecified` which causes the view controller to use the style that is set in system settings so we get some UIKit's views rendered with incorrect interface style 😞 This only affects view controllers presented modally, thus doing the same for pushed screens is not necessary.
@brentvatne prepared reproducible example: https://github.com/brentvatne/modal-example

# How

Native stack's view controllers will inherit `overrideUserInterfaceStyle` from presenting view controller.

# Test Plan

Example https://github.com/brentvatne/modal-example works as expected in Expo client with style set to Dark in system settings. _*This also requires one small change in the client itself, so it won't work in the current version in the app store._

### DIFF
--- a/ios/RNSScreenStack.m
+++ b/ios/RNSScreenStack.m
@@ -285,6 +285,12 @@
       for (NSUInteger i = changeRootIndex; i < controllers.count; i++) {
         UIViewController *next = controllers[i];
         BOOL lastModal = (i == controllers.count - 1);
+
+        if (@available(iOS 13.0, *)) {
+          // Inherit UI style from its parent - solves an issue with incorrect style being applied to some UIKit views like date picker or segmented control.
+          next.overrideUserInterfaceStyle = previous.overrideUserInterfaceStyle;
+        }
+
         [previous presentViewController:next
                                animated:lastModal
                              completion:^{
--- a/ios/RNSScreenStack.m
+++ b/ios/RNSScreenStack.m
@@ -286,10 +286,12 @@
         UIViewController *next = controllers[i];
         BOOL lastModal = (i == controllers.count - 1);
 
+#ifdef __IPHONE_13_0
         if (@available(iOS 13.0, *)) {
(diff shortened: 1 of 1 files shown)


ios/RNSScreenStack.m
```diff
@@ -285,6 +285,12 @@ - (void)setModalViewControllers:(NSArray<UIViewController *> *)controllers
       for (NSUInteger i = changeRootIndex; i < controllers.count; i++) {
         UIViewController *next = controllers[i];
         BOOL lastModal = (i == controllers.count - 1);
+
+        if (@available(iOS 13.0, *)) {
+          // Inherit UI style from its parent - solves an issue with incorrect style being applied to some UIKit views like date picker or segmented control.
+          next.overrideUserInterfaceStyle = previous.overrideUserInterfaceStyle;
```
kmagiera, line 291: I don't think this should make any difference in your case but I think it'd be more intuitive if presented controllers would inherit styles from their container and not from the sibling modal. If so I'd copy the property from `_controller.overrideUserInterfaceStyle` here and not from `previous`

ios/RNSScreenStack.m
```diff
@@ -285,6 +285,12 @@ - (void)setModalViewControllers:(NSArray<UIViewController *> *)controllers
       for (NSUInteger i = changeRootIndex; i < controllers.count; i++) {
         UIViewController *next = controllers[i];
         BOOL lastModal = (i == controllers.count - 1);
+
+        if (@available(iOS 13.0, *)) {
```
kmagiera, line 289: This also does not matter I think as soon appstore will stop accepting builds made with prev versions of xcode (if that isn't the case already) but for consistency can we add `#ifdef __IPHONE_13_0` to wrap this call much like we do in other cases where we do version checks

PR #486, opened 2020-04-29
feat: add replace animation

### LINKED ISSUE #394: Add animationTypeForReplace
I would like to have `animationTypeForReplace` option for `Stack.Screen`.

> **animationTypeForReplace**
>
>The type of animation to use when this screen replaces another screen. It takes the following values:
>
>`push` - The animation of a new screen being pushed will be used
>`pop` - The animation of a screen being popped will be used
>Defaults to `push`.
>
>When `pop` is used, the `pop` animation is applied to the screen being replaced

See: https://reactnavigation.org/docs/stack-navigator#animationtypeforreplace

WoLewicki:
Added animation options for `replace` action. At the update of the container, if the previous top screen is no longer on the stack and the new screen was not there before, it probably means that `replace` action was invoked, so we choose the animation of the new screen. Should close #394.

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.java
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.java
@@ -30,6 +30,11 @@ public class Screen extends ViewGroup {
     FADE
   }
 
+  public enum ReplaceAnimation {
+    PUSH,
+    POP
+  }
+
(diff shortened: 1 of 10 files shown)


enheit:
@WoLewicki @kmagiera any updates on this pull request?

WoLewicki:
@enheit you can always add it to your project via patch-package if you need it. It will be merged in the near future probably.

PR #488, opened 2020-04-30
fix: add better shadow handling for newer apis

### LINKED ISSUE #484: Header hide shadow on android do not work
Hello. I have set on Stack.Navigator property headerHideShadow: true and it works on IOS but on android on goBack it returns shadow or randomly adds it. Look into gif:

giphy

WoLewicki:
When `headerHideShadow` is set to true on a screen, pushing another screen on top of it and then going back makes the shadow appear again. It looks like a problem with the usage of deprecated method `setTargetElevation`. This change fixes the problem on the newer APIs by using another method of changing the elevation. Should resolve #484.

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.java
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.java
@@ -44,7 +44,6 @@ public class ScreenStackFragment extends ScreenFragment {
 
   private AppBarLayout mAppBarLayout;
   private Toolbar mToolbar;
-  private boolean mShadowHidden;
 
   @SuppressLint("ValidFragment")
   public ScreenStackFragment(Screen screenView) {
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.java
```diff
@@ -68,7 +70,11 @@ public void setToolbar(Toolbar toolbar) {
   }
 
   public void setToolbarShadowHidden(boolean hidden) {
-    if (mShadowHidden != hidden) {
+    if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.LOLLIPOP){
+      StateListAnimator stateListAnimator = new StateListAnimator();
+      stateListAnimator.addState(new int[0], ObjectAnimator.ofFloat(mAppBarLayout, "elevation",hidden ?  0 : TOOLBAR_ELEVATION));
+      mAppBarLayout.setStateListAnimator(stateListAnimator);
+    } else if (mShadowHidden != hidden) {
       mAppBarLayout.setTargetElevation(hidden ? 0 : TOOLBAR_ELEVATION);
```
chirag04, line 78: is it even worth keeping this around?

chirag04:
@WoLewicki would be great to ship this as well. thank you for the fix. 🙌

WoLewicki:
Closing due to #516.

PR #496, opened 2020-05-06
fix: disable gestures in native stack on Android

### LINKED ISSUE #402: Android: Navigation resets to first tab in nested navigations
I have a TabBar. 3 Tabs. Each Tab has its own stack. On iOS, being in Tab 3 and navigating inside that tab and switching tabs and returning back to tab 3: everything works as expected. I am still on the same screen.

When I do the same on android, I get back to the first screen of that Tab, while I should remain on the same inside stack screen. So the navigation state gets lost. Assume I have an open form and just clicked on a chat notification, returning back and everything is gone and I have to go over again.

React-native-screens 2.2.0.
react-native-screens/native-stack 
react-navigation 5

WoLewicki:
If `ScreenStack`s of native-stack navigators have the same `FragmentManager` (because each of them is not under a different `Screen` component), changing a screen in one of the stacks pops one screen in the other stack because of the  https://github.com/software-mansion/react-native-screens/blob/master/android/src/main/java/com/swmansion/rnscreens/ScreenStack.java#L34 being called in both `ScreenStack`s by the `FragmentManager`. The workaround is to set `gestureEnabled` to false in `createNativeStackNavigator` on Android since React-Navigation handles hw back button and setting it will make the https://github.com/software-mansion/react-native-screens/blob/master/android/src/main/java/com/swmansion/rnscreens/ScreenStack.java#L283 not to be called because the screen is not dismissable without `gestureEnabled` set to `true`. Should fix #402. PR also fixes some indentations since the file was not indented properly.

### DIFF
--- a/native-stack/views/NativeStackView.tsx
+++ b/native-stack/views/NativeStackView.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { View, StyleSheet } from 'react-native';
+import { View, StyleSheet, Platform } from 'react-native';
 import {
   ScreenStack,
   Screen as ScreenComponent,
@@ -46,7 +46,7 @@ export default function NativeStackView({
           <Screen
             key={route.key}
             style={StyleSheet.absoluteFill}
-            gestureEnabled={gestureEnabled}
+            gestureEnabled={Platform.OS === 'android' ? false : gestureEnabled}
             stackPresentation={stackPresentation}
             stackAnimation={stackAnimation}
             onAppear={() => {
--- a/createNativeStackNavigator.js
+++ b/createNativeStackNavigator.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { StyleSheet } from 'react-native';
+import { StyleSheet, Platform } from 'react-native';
 import {
   StackRouter,
   SceneView,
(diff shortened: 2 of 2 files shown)


native-stack/views/NativeStackView.tsx
```diff
@@ -46,7 +46,7 @@ export default function NativeStackView({
           <Screen
             key={route.key}
             style={StyleSheet.absoluteFill}
-            gestureEnabled={gestureEnabled}
+            gestureEnabled={Platform.OS === 'android' ? false : gestureEnabled}
```
chirag04, line 49: swipe gestures are common on android these days tho. there will be no way to have them on android with this change, yeah?

WoLewicki: `gestureEnabled` prop on Android handles dismissing screen by clicking hw back button and there are no swipe gestures supported by RNScreens on Android for now.

ku8ar:
@WoLewicki Can you tell me why property `gestureEnabled` exists on the native side if it is always false in JS? Maybe we should just remove this property in java code?

WoLewicki:
It is there e.g. for people who would like to make their own implementation of JS side of `react-native-screens`, with different handling of hw back button.

PR #498, opened 2020-05-06
Add onWillAppear, onWillDisappear and onDisappear props

janicduplessis:
WIP, just want to validate the api

These match exactly the timing of iOS `viewWillAppear`, `viewWillDisappear` and `viewDidDisappear`. The main motivation for adding those is support transition start / end events in the native-stack implementation like the JS version does.

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.java
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.java
@@ -55,6 +55,14 @@ public class Screen extends ViewGroup {
   private boolean mGestureEnabled = true;
 
   @Override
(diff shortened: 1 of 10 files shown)


janicduplessis:
TODO: Test / figure out how this works with regular Screen

hadnet:
@janicduplessis hi, man, thanks for your work. Any progress in order to use those start & end events?

janicduplessis:
I think this PR is good to merge, I haven't fully tested the timing of these events when using non-native stack, but they are not really useful in this case anyway. Been using this in prod for a few months.

janicduplessis:
Merging #499 instead

Issue #317, opened 2020-02-10
Consider exposing Animated.Value with transition progress (for shared element transitions)


Jarred-Sumner:
It'd be great if the native stack navigator supported shared element transitions. One way of implementing that, without `react-native-screens` needing to know anything about shared element transitions, is by exposing an `Animated.Value` with the progress of the current transition running.

I explored a little how one might implement this.

In here, if you switched from `[UIView animateWithDuration]` to [`UIViewPropertyAnimator`](https://developer.apple.com/documentation/uikit/uiviewpropertyanimator), you can get [`fractionComplete`](https://developer.apple.com/documentation/uikit/uiviewanimating/1649787-fractioncomplete):

https://github.com/kmagiera/react-native-screens/blob/ed997ef4ecd4cc99383386cb016bc8d79bf70279/ios/RNSScreenStack.m#L354-L375

This is an example of using UIViewPropertyAnimator with a custom view controller transition: https://stackoverflow.com/questions/27638806/transitioncoordinator-return-nil-for-custom-container-view-controller-in-ios8

From there, you would have a `CADisplayLink` which sets the value for the animated node with the `fractionComplete` from the `UIViewPropertyAnimator` instance.

nandorojo:
Would be awesome if this could make it before @expo releases sdk 37.

janicduplessis:
Could be possible to use this along with native animated. We could expose an animation progress event which could be connected to an animated value. Not sure if it could work with stock pop / push animations but with the custom fade one it seems possible (turns out good since it is probably the animation you want to use with shared elements).

Here's how it could look:

```ts
const progress = new Animated.Value(0);

<ScreenStack>
  <Screen onAnimationProgress={Animated.event([{nativeEvent: {progress}}], {useNativeDriver: true})} />
</ScreenStack>
```

Hirbod:
But Android does not have a fade animation. (at least currently in native-stack). We would need both platforms. 

I would be happy to do a shared animation with fade. It’s the most common pattern anyway.

mrousavy:
@Hirbod I'm not quite sure I understand, do you mean fade animation is the only animation for this use case? I've seen quite a few uses of shared element transitions and they're almost always more than just a simple fade.. The App Store's App of the day animation is a (very complex) example for this.
Don't get me wrong, native shared element fade animations are a good start!

Hirbod:
[...]

PR #499, opened 2020-05-06
[native-stack] Add transitionStart and transitionEnd events

janicduplessis:
Implement [these](https://github.com/react-navigation/react-navigation/blob/bea14aa26fd5cbfebc7973733c5cf1f44fd323aa/packages/stack/src/types.tsx#L25) events in native stack, similar to the events available in js stack.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -56,7 +56,7 @@ declare module 'react-native-screens' {
      */
     onWillAppear?: (e: NativeSyntheticEvent<NativeTouchEvent>) => void;
     /**
-     * @description A callback that gets called when the current screen will disappears. This is called as soon as the transition begins.
+     * @description A callback that gets called when the current screen will disappear. This is called as soon as the transition begins.
      */
     onWillDisappear?: (e: NativeSyntheticEvent<NativeTouchEvent>) => void;
     /**
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -52,7 +52,7 @@ declare module 'react-native-screens' {
     children?: React.ReactNode;
 
     /**
-     * @description A callback that gets called when the current screen will appears. This is called as soon as the transition begins.
+     * @description A callback that gets called when the current screen will appear. This is called as soon as the transition begins.
      */
     onWillAppear?: (e: NativeSyntheticEvent<NativeTouchEvent>) => void;
     /**
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -50,12 +50,25 @@ declare module 'react-native-screens' {
     active?: 0 | 1 | Animated.AnimatedInterpolation;
     onComponentRef?: (view: any) => void;
     children?: React.ReactNode;
+
     /**
-     *@description A callback that gets called when the current screen appears.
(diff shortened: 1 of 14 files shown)


Hirbod:
Nice one! I miss this events badly!

Hirbod:
@janicduplessis https://github.com/software-mansion/react-native-screens/issues/317
could this PR also take this into consideration? Without these I think react-native-shared-element won't be able to support native-stack

janicduplessis:
@Hirbod I think it is different, It could be possible to expose an animation progress event for the custom animations, so the fade one right now. Not sure if it is possible with the default push / pop animations.

Hirbod:
I think the fade one would be an extremely good start :)

src/index.d.ts
```diff
@@ -29,12 +29,25 @@ declare module 'react-native-screens' {
     active?: 0 | 1 | Animated.AnimatedInterpolation;
     onComponentRef?: (view: any) => void;
     children?: React.ReactNode;
+
+    /**
+     * @description A callback that gets called when the current screen will appears. This is called as soon as the transition begins.
```
WoLewicki, line 34:
```suggestion
     * @description A callback that gets called when the current screen will appear. This is called as soon as the transition begins.
```

src/index.d.ts
```diff
@@ -29,12 +29,25 @@ declare module 'react-native-screens' {
...
+     */
+    onWillAppear?: (e: NativeSyntheticEvent<NativeTouchEvent>) => void;
     /**
-     *@description A callback that gets called when the current screen appears.
+     * @description A callback that gets called when the current screen will disappears. This is called as soon as the transition begins.
```
WoLewicki, line 38:
```suggestion
     * @description A callback that gets called when the current screen will disappear. This is called as soon as the transition begins.
```

WoLewicki:
Could you also update https://github.com/software-mansion/react-native-screens/blob/master/native-stack/README.md with the new events? And I think it will be ready then.

janicduplessis:
Updated in 6a72acd1ea1d25d13aaf9ac607b2ee57ec5a7f35, Also removed finishedTransitioning doc since it was removed in a previous PR.

janicduplessis:
Edit: Needed a bit more work to get nested stacks working on Android. This adds some infra for nested stack to register themselves on their parent screen to allow dispatching events to child stacks.

We already looked up the parent screen to get the fragment manager but didn't keep references around. This keeps the reference to the parent Screen in ScreenStack and registers the child stacks in their parent Screen to allow communicating in both directions between nested stacks.

This is mostly useful to dispatch an event from a screen to its child stacks since traversing down the view hierarchy is not practical.

janicduplessis:
Here are logs on iOS and Android testing side by side with https://gist.github.com/janicduplessis/fd2d7b37b19126824a7c3dbbd9891444

image

https://gist.github.com/janicduplessis/33eb975e15ccfc99d6e29882818eec89

WoLewicki:
The PR looks like it is ready to be merged! Very good job @janicduplessis :tada:

PR #500, opened 2020-05-07
Add props to headerLeft and headerRight

AzzouDuGhetto:
This PR will allow users to custom headerRight and headerLeft components using user's theme color.
For example, if I want a custom settings icon in headerRight to navigate to my settings page, I will be able to retrieve color from my custom theme. It's already available in ReactNavigation.

`colors.primary` is used by the back button so it makes sense to use it for headerRight and headerLeft.

The same logic may be applied to headerCenter (Imagine a custom TextComponent), we may apply colors.text in this case.

Let me know if you find this useful (or not), or if I can make it better to be integrated.

### DIFF
--- a/native-stack/views/HeaderConfig.tsx
+++ b/native-stack/views/HeaderConfig.tsx
@@ -50,8 +50,8 @@ export default function HeaderConfig(props: Props) {
         headerTitle !== undefined
           ? headerTitle
           : title !== undefined
-          ? title
-          : route.name
+            ? title
+            : route.name
       }
       titleFontFamily={headerTitleStyle.fontFamily}
       titleFontSize={headerTitleStyle.fontSize}
@@ -59,8 +59,8 @@ export default function HeaderConfig(props: Props) {
         headerTitleStyle.color !== undefined
           ? headerTitleStyle.color
           : headerTintColor !== undefined
-          ? headerTintColor
-          : colors.text
+            ? headerTintColor
+            : colors.text
       }
       backTitle={headerBackTitleVisible ? headerBackTitle : ' '}
       backTitleFontFamily={headerBackTitleStyle.fontFamily}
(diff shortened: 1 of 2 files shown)


WoLewicki:
I think it would be best if you replicated the behavior of `color={headerTintColor !== undefined ? headerTintColor : colors.primary}`. Also could you add the same thing for the `headerCenter`?

AzzouDuGhetto:
@WoLewicki Done, I add it to `headerCenter` with `colors.text` too. I personally prefer to use `??` operators  (as defined here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) but if you want me to change back to the classic a !== undefined ? a : b, there's no problem.

WoLewicki:
Please change it to the classic form so we can keep it consistent. Maybe we will change all the occurrences to nullish coalescing soon. Also could you please use the `colors.primary` in the `headerCenter` too? It seems more fitting for every custom view to get the same default color.

WoLewicki:
Or we can leave it as it is, nevermind. Will change `headerCenter` only and merge it since we are releasing the library.

native-stack/views/HeaderConfig.tsx
```diff
@@ -77,14 +77,18 @@ export default function HeaderConfig(props: Props) {
...
       {headerLeft !== undefined ? (
-        <ScreenStackHeaderLeftView>{headerLeft()}</ScreenStackHeaderLeftView>
+        <ScreenStackHeaderLeftView>
+          {headerLeft({ tintColor: headerTintColor ?? colors.primary })}
+        </ScreenStackHeaderLeftView>
       ) : null}
       {headerCenter !== undefined ? (
         <ScreenStackHeaderCenterView>
-          {headerCenter()}
+          {headerCenter({ tintColor: headerTintColor ?? colors.text })}
```
WoLewicki, line 91:
```suggestion
          {headerCenter({ tintColor: headerTintColor ?? colors.primary })}
```

AzzouDuGhetto:
Wouldn’t it be more logical for `headerCenter` to mimic `headerTitle` behavior ? `headerCenter` will most likely be a custom TextComponent, so `colors.text` would make more sense ? Anyway, glad it’s getting merged 👍🏼

PR #502, opened 2020-05-08
Allow disabling modal animations

### LINKED ISSUE #494: Allow changing animation for modals
We're creating a navigator using `createNativeStackNavigator` and we have a screen with options `{ stackPresentation: 'transparentModal', stackAnimation: 'none' }`

We're doing our own animation with a backdrop, so we need to be able to disable the system animation

Currently `stackAnimation` only works with stack transitions

Tested on iOS

jacobp100:
Allows disabling modal animations via `stackAnimation: 'none'` on iOS. Android already handles this

Fixes #494

### DIFF
--- a/ios/RNSScreenStack.m
+++ b/ios/RNSScreenStack.m
@@ -293,8 +293,10 @@
         }
 #endif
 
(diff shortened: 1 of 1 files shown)


WoLewicki:
Looks good. Did you test it to check if the completion handler always fires despite no animation?

jacobp100:
Just tried it now - it does still fire!

WoLewicki:
Great! I am merging it then.

jacobp100:
Thanks!

PR #512, opened 2020-05-14
@wolewicki/add header blur effect

WoLewicki:
Added `appearance.backgroundEffect` options for blurring background in semi-transparent and transparent header.
PR also removes `configureWithDefaultBackground` option of appearance for translucent header since it applies `blurEffect` which causes problems (#520) and is not necessary since the prop for `blurEffect` will now be available.

### DIFF
--- a/ios/RNSScreenStackHeaderConfig.m
+++ b/ios/RNSScreenStackHeaderConfig.m
@@ -266,12 +266,7 @@
     // transparent background color
     [appearance configureWithTransparentBackground];
   } else {
-    // non-transparent background or default background
-    if (config.translucent) {
-      [appearance configureWithDefaultBackground];
-    } else {
-      [appearance configureWithOpaqueBackground];
-    }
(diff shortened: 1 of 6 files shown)


ios/RNSScreenStackHeaderConfig.m
```diff
@@ -467,6 +468,59 @@ + (void)updateViewController:(UIViewController *)vc withConfig:(RNSScreenStackHe
   }
 }
 
++ (UIBlurEffectStyle)blurEffectStyleFromEffectName:(NSString *)effectName
+{
+  if (@available(iOS 13.0, *)) {
+    if ([effectName  isEqual: @"extraLight"]) {
```
jacobp100, line 474: Could this be a `RCT_ENUM_CONVERTER`?

WoLewicki: Yeah, I will probably change it when I'll have some time.

Issue #485, opened 2020-04-29
Navbar window fitting android


tankers746:
I was wondering why the following was reverted: 
https://github.com/software-mansion/react-native-screens/commit/4f8efd2873516dd2d3c30661dffce99b5fa76f57

The current navbar on android doesn't take into account for using a translucent status bar:

`MainActivity.onCreate`

`getWindow().getDecorView().setSystemUiVisibility(View.SYSTEM_UI_FLAG_LAYOUT_STABLE | View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN);`

![image](https://user-images.githubusercontent.com/19326459/80616511-93935580-8a73-11ea-8e22-4372cfc845a2.png)

janicduplessis:
I'm using this patch as a workaround for now, we tried making it work properly before but ended up very difficult.

In ScreenStackHeaderConfig.java
```
@@ -153,6 +156,13 @@ public class ScreenStackHeaderConfig extends ViewGroup {
       getScreenFragment().setToolbar(mToolbar);
     }
 
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+      mToolbar.setPadding(0, getRootWindowInsets().getSystemWindowInsetTop(), 0, 0);
+    } else {
+      // Hacky fallback for old android.
+      mToolbar.setPadding(0, (int)(25 * getResources().getDisplayMetrics().density), 0, 0);
+    }
+
     activity.setSupportActionBar(mToolbar);
     ActionBar actionBar = activity.getSupportActionBar();
 ```

Maybe we could make it a prop so try to find some heuristic to figure out if we're drawing under the statusbar.

Hirbod:
Well.. we both got the same error on the same day ;)
Was about to release my app right now. So we need to set translucent to false again and go with black for now... 
So Expo -> translucent just dont work.

<img width="400" alt="Bildschirmfoto 2020-04-29 um 19 50 09" src="https://user-images.githubusercontent.com/504909/80629130-afd2ca80-8a52-11ea-90de-df14cd4ea2ca.png">

mrousavy:
@janicduplessis Your patch doesn't work for me, it's saying "package Build does not exist". Where did you import the `Build` class from?

mrousavy:
[...]

PR #518, opened 2020-05-20
feat: add translucent status bar handling

WoLewicki:
Added a listener on the insets of the window which mimics the behavior of RN's StatusBar: (https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/modules/statusbar/StatusBarModule.java#L145). The addition of padding is controlled by the exported `translucentStatusBar` prop. The prop is needed because of the incorrect behavior which occurred sometimes if the padding was only controlled by the listener, where the toolbar's top padding was added to the window insets top padding, causing the doubled top padding of the toolbar.
The situation handled here is the case where a user has his status bar set to translucent via RN StatusBar component with `translucent={true}` prop. In that case, the header's top padding is not applied properly which causes the content of the header to go under the Status Bar. See #485 for more info.

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
@@ -40,6 +40,7 @@ public class ScreenStackHeaderConfig extends ViewGroup {
   private boolean mIsShadowHidden;
   private boolean mDestroyed;
   private boolean mBackButtonInCustomView;
+  private boolean mTranslucentStatusBar;
   private int mTintColor;
   private final Toolbar mToolbar;
 
@@ -63,6 +64,20 @@ public class ScreenStackHeaderConfig extends ViewGroup {
           fragment.dismiss();
         }
       }
+    }
+  };
+
+  @RequiresApi(api = Build.VERSION_CODES.KITKAT_WATCH)
+  private OnApplyWindowInsetsListener mOnApplyWindowInsetsListener = new View.OnApplyWindowInsetsListener() {
+    @Override
+    public WindowInsets onApplyWindowInsets(View v, WindowInsets insets) {
+      WindowInsets defaultInsets = v.onApplyWindowInsets(insets);
+      mToolbar.setPadding(0, insets.getSystemWindowInsetTop(), 0, 0);
+      return defaultInsets.replaceSystemWindowInsets(
+              defaultInsets.getSystemWindowInsetLeft(),
+              0,
+              defaultInsets.getSystemWindowInsetRight(),
+              defaultInsets.getSystemWindowInsetBottom());
     }
   };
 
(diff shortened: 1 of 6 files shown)


android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
```diff
@@ -302,6 +312,23 @@ private TextView getTitleTextView() {
...
+      @Override
+      public WindowInsets onApplyWindowInsets(View v, WindowInsets insets) {
+        WindowInsets defaultInsets = v.onApplyWindowInsets(insets);
+        // this inset is set to 0 by translucent status bar
+        if (defaultInsets.getStableInsetTop() == 0) {
+          mToolbar.setPadding(0, insets.getSystemWindowInsetTop(), 0, 0);
+        } else {
+          mToolbar.setPadding(0, defaultInsets.getSystemWindowInsetTop(), 0, 0);
+        }
+        return insets.consumeSystemWindowInsets();
```
mczernek, line 327: You really consumed only top inset in here, haven't you? There should be a way of marking it.

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
```diff
@@ -173,6 +178,11 @@ public void onUpdate() {
     mToolbar.setNavigationOnClickListener(mBackClickListener);
 
 
+    // we apply padding if status bar is translucent
```
mczernek, line 181: You could use same constant as in `RequiresApi` annotation in `applyWindowInsets` method

janicduplessis:
Does this work if translucent was set using system ui visibility flags instead of using the RN statusbar module?

```
getWindow().getDecorView().setSystemUiVisibility(View.SYSTEM_UI_FLAG_LAYOUT_STABLE | View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN);
```

WoLewicki:
@janicduplessis I changed the behavior to mimic what the RN StatusBar does: https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/modules/statusbar/StatusBarModule.java#L119. It requires the user to apply a prop if he wants it to work correctly, but it eliminates the problems with additional padding added which sometimes occurred with the previous implementation. It should also work ok with system ui visibility flags, because you can now control the padding by yourself with the prop.

janicduplessis:
:+1: using a prop seems like a good approach to me.

Hirbod:
Maybe it should be called translucentStatusBarAndroid or something. That wording could be confusing

WoLewicki:
@Hirbod it has mentions of the platform in the types definition files. I think it should be enough since it is the convention in this library.

native-stack/types.tsx
```diff
@@ -210,6 +210,13 @@ export type NativeStackNavigationOptions = {
    * @platform ios
    */
   gestureEnabled?: boolean;
+  /**
+   * Set this prop to true whenever you have translucent status bar in order for it to work correctly. Defaults to `false`.
+   * Only supported on Android.
+   *
+   * @platform android
+   */
+  translucentStatusBar?: boolean;
```
brentvatne, line 219: `insetForTranslucentStatusBar` maybe? otherwise it feels like we are setting the status bar to translucent

WoLewicki:
Closing due to #545.

PR #545, opened 2020-06-18
Add headerTopInsetEnabled prop that defaults to true to account for translucent status bars on Android

brentvatne:
Follow up from https://github.com/software-mansion/react-native-screens/pull/518 using @janicduplessis' suggested solution. The inset is applied by default to encourage developers to use translucent status bars on Android, which is typically a better experience for users and consistent with the iOS behavior.

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
@@ -3,6 +3,7 @@ package com.swmansion.rnscreens;
 import android.content.Context;
 import android.graphics.PorterDuff;
 import android.graphics.drawable.Drawable;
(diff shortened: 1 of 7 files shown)


WoLewicki:
The situation handled here is the case where a user has his status bar set to translucent via RN `StatusBar` component with `translucent={true}` prop. In that case, the header's top padding is not applied properly which causes the content of the header to go under the Status Bar.

This PR introduces additional padding to the header, which is a **BREAKING CHANGE** and will cause the header to have too much top padding in case of not having a translucent status bar.

There are two ways to handle it:
1. Recommended - use translucent status bar via RN `StatusBar` component (if you already don't use it) - https://reactnative.dev/docs/statusbar#translucent.
2. Set `headerTopInsetEnabled` to `false` (it is `true` by default) if you want an opaque status bar.

Issue #573, opened 2020-07-21
[Android] Render content behind transparent navigation bar


kyle-ssg:
on iOS, when you set the background of the navbar to transparent, the content is able to be rendered behind the navbar.  

![image](https://user-images.githubusercontent.com/8608314/88090030-dfdac700-cb84-11ea-8e88-d59a83af4893.png)

Ideally, I'd like to have the ability to render content behind the navigation bar in android.

kylsYw:
I have the same problem

Issue #405, opened 2020-03-05
[native-stack] Custom image background on the whole screen


leshniak:
**Platform:** Android

**Problem:** I'd like to set a custom image background on the whole app screen, including the header bar. On react-navigation and non-native stack I can do it with https://reactnavigation.org/docs/stack-navigator/#headertransparent.

Is this possible on native-stack? Or will it be possible in the future?

WoLewicki:
So you basically need a transparent header?

leshniak:
Yes, I want to achieve an effect like this:
![IMG_1737](https://user-images.githubusercontent.com/752917/82296527-d5435c00-99a0-11ea-8931-ef0e218d08b2.PNG)

WoLewicki:
On iOS it is already an option with:
```js
 headerStyle: 
{
  backgroundColor: 'transparent'
}
```
On Android it is not available currently.

WoLewicki:
Please follow #575 for any updates on this.

PR #575, opened 2020-07-23
[Android] Add support for translucent header

andreibarabas:
Currently, the `headerTranslucent = true` is only supported on iOS. With this PR the setting will also work on Android

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,27 +109,27 @@ Allows for the customization of how the given screen should appear/disappear whe
 
 Defines how the method that should be used to present the given screen. It is a separate property from `stackAnimation` as the presentation mode can carry additional semantic. The allowed values are:
 
- - `push` – the new screen will be pushed onto a stack which on iOS means that the default animation will be slide from the side, the animation on Android may vary depending on the OS version and theme.
- - `modal` – Explained below.
- - `transparentModal` – Explained below.
- - `containedModal` – Explained below.
- - `containedTransparentModal` – Explained below.
- - `fullScreenModal` – Explained below.
- - `formSheet` – Explained below.
+- `push` – the new screen will be pushed onto a stack which on iOS means that the default animation will be slide from the side, the animation on Android may vary depending on the OS version and theme.
+- `modal` – Explained below.
+- `transparentModal` – Explained below.
+- `containedModal` – Explained below.
+- `containedTransparentModal` – Explained below.
+- `fullScreenModal` – Explained below.
+- `formSheet` – Explained below.
 
- For iOS:
- - `modal` will use [`UIModalPresentationAutomatic`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationautomatic?language=objc) on iOS 13 and later, and will use [`UIModalPresentationFullScreen`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationfullscreen?language=objc) on iOS 12 and earlier.
- - `fullScreenModal` will use [`UIModalPresentationFullScreen`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationfullscreen?language=objc)
- - `formSheet` will use [`UIModalPresentationFormSheet`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationformsheet?language=objc)
- - `transparentModal` will use [`UIModalPresentationOverFullScreen`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationoverfullscreen?language=objc)
- - `containedModal` will use [`UIModalPresentationCurrentContext`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationcurrentcontext?language=objc)
- - `containedTransparentModal` will use [`UIModalPresentationOverCurrentContext`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationovercurrentcontext?language=objc)
+For iOS:
+- `modal` will use [`UIModalPresentationAutomatic`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationautomatic?language=objc) on iOS 13 and later, and will use [`UIModalPresentationFullScreen`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationfullscreen?language=objc) on iOS 12 and earlier.
+- `fullScreenModal` will use [`UIModalPresentationFullScreen`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationfullscreen?language=objc)
+- `formSheet` will use [`UIModalPresentationFormSheet`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationformsheet?language=objc)
+- `transparentModal` will use [`UIModalPresentationOverFullScreen`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationoverfullscreen?language=objc)
+- `containedModal` will use [`UIModalPresentationCurrentContext`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationcurrentcontext?language=objc)
+- `containedTransparentModal` will use [`UIModalPresentationOverCurrentContext`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationovercurrentcontext?language=objc)
 
- For Android:
+For Android:
 
- `modal`, `containedModal`, `fullScreenModal`, `formSheet` will use `Screen.StackPresentation.MODAL`.
+`modal`, `containedModal`, `fullScreenModal`, `formSheet` will use `Screen.StackPresentation.MODAL`.
 
- `transparentModal`, `containedTransparentModal` will use `Screen.StackPresentation.TRANSPARENT_MODAL`.
+`transparentModal`, `containedTransparentModal` will use `Screen.StackPresentation.TRANSPARENT_MODAL`.
 
 ### `<ScreenStackHeaderConfig>`
 
@@ -191,9 +191,9 @@ String that applies `rtl` or `ltr` form to the stack.
 
 When set to `false` the back swipe gesture will be disabled when the parent `Screen` is on top of the stack. The default value is `true`.
 
-#### `translucent` (iOS only)
+#### `translucent`
 
-When set to `true`, it makes native navigation bar on iOS semi-transparent with blur effect. It is a common way of presenting a navigation bar introduced in iOS 11. The default value is `false`.
+When set to `true`, it makes native navigation bar on iOS semi-transparent with blur effect, while on Android is fully transparent. It is a common way of presenting a navigation bar introduced in iOS 11. The default value is `false`.
 
 #### `backTitle` (iOS only)
 
--- a/README.md
+++ b/README.md
@@ -193,7 +193,7 @@ When set to `false` the back swipe gesture will be disabled when the parent `Scr
 
 #### `translucent`
 
-When set to `true`, it makes native navigation bar on iOS semi-transparent with blur effect, while on Android is fully transparent. It is a common way of presenting a navigation bar introduced in iOS 11. The default value is `false`.
+When set to `true`, it allows the content to go under the navigation header, not bellow. If you want to create a transparent header, you should also set `backgroundColor` to `transparent`. The default value is `false`.
 
 #### `backTitle` (iOS only)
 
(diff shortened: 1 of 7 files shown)


kyle-ssg:
Ah awesome! I Guess that would resolve https://github.com/software-mansion/react-native-screens/issues/573.

kyle-ssg:
@andreibarabas This works exactly as I'd expect except when popping, it seems to force everything to render after the navbar.

andreibarabas:
Hey @kyle-ssg - i'll take a look

WoLewicki:
I would love to merge it! Please inform me when it is ready for review. It should also close #405.

andreibarabas:
Hey @WoLewicki , I need a couple more days to get to this. I'm currently caught up in a release, and once that is done can look more into this.

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
```diff
@@ -76,7 +77,6 @@ public ScreenStackHeaderConfig(Context context) {
     if (context.getTheme().resolveAttribute(android.R.attr.colorPrimary, tv, true)) {
       mToolbar.setBackgroundColor(tv.data);
     }
-    mToolbar.setClipChildren(false);
```
janicduplessis, line 79: Is there a reason why this was removed?

WoLewicki: Probably wrongly rebased to the current master.

WoLewicki:
@andreibarabas it is not a problem 🙂

WoLewicki:
How are things going on this PR @andreibarabas? Do you need any help or you just didn't have time to finish it?

andreibarabas:
Haven't yet had the time to dive deep into it, but will give it another shoot hopefully tomorrow / the day after that.

I also need this fixed for our production build, so top priority ;)

andreibarabas:
@WoLewicki - done ;) it works now, tested it with pop also.

WoLewicki:
Looks good! There are some other places where the information about the platform availability should be changed: https://github.com/software-mansion/react-native-screens/blob/master/README.md#L194, https://github.com/software-mansion/react-native-screens/blob/master/src/index.d.ts#L158, https://github.com/software-mansion/react-native-screens/blob/master/src/native-stack/types.tsx#L116. If you could change this, I would be grateful. Also, do you know of any options to have something like `contentInsetAdjustmentBehavior` for Android? Would be good to have consistent behavior for ScrollViews on both platforms.

kyle-ssg:
Hey @andreibarabas, from some brief testing I've noticed the title isn't rendering at least for me.

image

andreibarabas:
Hey @kyle-ssg . Title i have't checked. Our design system doesn't use them.

Let me check that and get back.

Any other issues?

kyle-ssg:
> Hey @kyle-ssg . Title i have't checked. Our design system doesn't use them.
> 
> Let me check that and get back.
> 
> Any other issues?

Other than that it looks awesome @andreibarabas!

andreibarabas:
@kyle-ssg - can you provide a snack maybe? I checked on our app and put a dummy title and it showed up. Here it is 
Screenshot_20200915_094510

@WoLewicki - updated the doc as per your suggestion. 

As for `contentInsetAdjustmentBehavior`, I'm not really an android developer, so would have to deeply research it. maybe in a different PR? Do you see that as blocking for this one?

WoLewicki:
@andreibarabas I don't think it is blocking since this transparent header is an optional thing anyway.

kyle-ssg:
Yeah, happy with that. I'll raise separately with repro

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
```diff
@@ -363,7 +367,7 @@ public void setTintColor(int color) {
 
   public void setTopInsetEnabled(boolean topInsetEnabled) { mIsTopInsetEnabled = topInsetEnabled; }
 
-  public void setBackgroundColor(int color) {
+  public void setBackgroundColor(Integer color) {
```
WoLewicki, line 370: Also, could you change it back to int? It is the way we pass all the color props and I think it is the preferred way of doing so.

andreibarabas: It was int, but int does not accept null. We need null to actually not put a color.

andreibarabas: For the ui to look as in the example, on iOS (and now android), you need to write


`headerTranslucent: true,
      headerStyle: {backgroundColor: 'transparent'},`

WoLewicki: If you make it an Integer, the value of 0 is passed for `backgroundColor: 'transparent',` and -1 if nothing is passed. There is no `null` option. Am I right?
If you make it an int, the same values are passed. 
So maybe it is better to leave the int and just check `if (mBackgroundColor != -1)` instead of 0?

andreibarabas: let me quickly do a test and see how it behaves

andreibarabas: @WoLewicki - you where right. I think I put null because it clearly says what you are doing (i.e: setting the color ONLY if the prop was set), but your solution works and i've reverted it like that to be consistent.

Here are some screenshots:

Screenshot_20200915_150323
Screenshot_20200915_150357
Screenshot_20200915_150423
Screenshot_20200915_150446

README.md
```diff
@@ -191,9 +191,9 @@ String that applies `rtl` or `ltr` form to the stack.
 
 When set to `false` the back swipe gesture will be disabled when the parent `Screen` is on top of the stack. The default value is `true`.
 
-#### `translucent` (iOS only)
+#### `translucent`
 
-When set to `true`, it makes native navigation bar on iOS semi-transparent with blur effect. It is a common way of presenting a navigation bar introduced in iOS 11. The default value is `false`.
+When set to `true`, it makes native navigation bar on iOS semi-transparent with blur effect, while on Android is fully transparent. It is a common way of presenting a navigation bar introduced in iOS 11. The default value is `false`.
```
WoLewicki, line 196: I also think this is not true for either of the platforms. All it does is make it possible for the content to be visible under the header due to not rendering content below the header but under it. It does not change the transparency by itself since it is controlled by `backgroundColor`. Or am I missing something?

andreibarabas: Yes, you are right. Updated.

andreibarabas:
hey @WoLewicki  - anything else you guys need to have this merged?

WoLewicki:
Didn't @kyle-ssg have problems with the title that needed to be resolved? Or are they gone?

andreibarabas:
I can't repro them. All screenshots I sent above include the title

brunocrpontes:
I tested this branch in a personal app and headers title just go away when use an headerLeft component in bar.

WoLewicki:
@brunocrpontes is it specific to translucent header? I think you are talking about this behavior: https://github.com/software-mansion/react-native-screens/tree/master/native-stack#headerleft and the same happens with normal header. Am I right?

brunocrpontes:
> @brunocrpontes is it specific to translucent header? I think you are talking about this behavior: https://github.com/software-mansion/react-native-screens/tree/master/native-stack#headerleft and the same happens with normal header. Am I right?

Yes, you're right.

WoLewicki:
@kyle-ssg so can you provide a repro showing the lack of title? I would like to test it and merge it if all is ok.

kyle-ssg:
Hey @WoLewicki sorry today I can't really busy release, it would make sense if it's the custom header left that's doing it though. I bet that is the usecase!

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
```diff
@@ -238,7 +242,7 @@ public void onUpdate() {
     }
 
     // background
-    if (mBackgroundColor != 0) {
+    if (mBackgroundColor != -1) {
```
kmagiera, line 245: Can you explain why this change is made here? If we expect "default" to be `-1` maybe we should also set the default in `@ReactProp` while defining bgColor:

```java
@ReactProp(name = "backgroundColor", customType = "Color", defaultInt = -1)
```

??

kmagiera (commented):
This looks good overal, I only have one question I left inline regarding one change. I'd like to get that one item clarified, and this is good to merge otherwise

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
WoLewicki: I think it is a good idea to add that for extra safety.

andreibarabas: Sure. will take care of it as soon as possible.

There is still an open issue from @kyle-ssg which reproduces only if there is a custom left component. I won't have time earlier than next week to investigate

WoLewicki: But I think this is not an issue since it is the expected behavior for the title to disappear when there is `headerLeft` component.

kyle-ssg: Regardless of if the headerLeft occurred before / after this work I think this is good to merge, I'll look into this separately and raise something if I think there's an issue. @WoLewicki @andreibarabas.

WoLewicki: On iOS, the situation is kind of similar but for `headerCenter`, where using it makes the title disappear.

andreibarabas: sure. it's fine with me. I added @kmagiera suggestion, so let me know if you need anything else from my side to get this merged.

WoLewicki:
Thanks for your contribution! The only problem I can spot is that passing a wrong color name is resolved to 0 so the header becomes transparent, but I think we can classify it as a problem of the programmer passing wrong options. I am merging it then!

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfigViewManager.java
```diff
@@ -89,9 +89,9 @@ public void setTitleColor(ScreenStackHeaderConfig config, int titleColor) {
     config.setTitleColor(titleColor);
   }
 
-  @ReactProp(name = "backgroundColor", customType = "Color")
-  public void setBackgroundColor(ScreenStackHeaderConfig config, int titleColor) {
-    config.setBackgroundColor(titleColor);
+  @ReactProp(name = "backgroundColor", customType = "Color", defaultInt = -1)
```
andrewbeckman, line 92: @andreibarabas I am noticing that when I try to set my header background color to "#fff" or "#ffffff" (i.e. white), I end up with a non-white header. If you take a look at the attached photo, you can see the header appears to be gray. Also note that the status bar at the top is white.

Any chance these bugs came in with the changes to this file?

Screen Shot 2020-10-04 at 12 12 00 PM

andrewbeckman: To clarify, this is when `headerTranslucent` is set to false.

andrewbeckman: After some further testing, @andreibarabas it looks like it was the change in ScreenStackHeaderConfig.java

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
andrewbeckman: I am testing these changes out and this line seems to have introduced a bug where I can't set the header background color to "#fff" or "#ffffff".

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfigViewManager.java
WoLewicki: Looks like the "#fff", "#ffffff" and `white` colors all have value of "-1" on the native side, so the newly added if makes it impossible to use these values. Can you check if https://github.com/software-mansion/react-native-screens/pull/656 fixes the issue?

andrewbeckman: @WoLewicki that fixed the issue! Thanks

andreibarabas: Guys, glad you solved it. Thanks! Just now had the time to groom the notifications.

PR #580, opened 2020-07-28
feat: add RTL support

### LINKED ISSUE #403: RTL SUPPORT
hi , i'm trying to use the native stack for iOS and android ,

the issue that when the app is rtl ... the animations and the headers still from left to right ,

this issue was in the react native navigation library and they added their support here 

Screen Shot 2020-03-02 at 10 18 12 AM

https://wix.github.io/react-native-navigation/#/docs/styling?id=options-object-format

kindly add this support so the library will be full and have all the same features like them

WoLewicki:
Added RTL support for both platforms. In order to use `rtl` on Android, you have to add `android:supportsRtl="true"` in the manifest of your app.
On Android, if you set the above flag in the manifest, the orientation changes without the need to do it programmatically. On iOS, the direction defaults to `ltr`, and only way to change the direction is to do it programmatically. Should resolve #403.

### DIFF
--- a/ios/RNSScreenStackHeaderConfig.m
+++ b/ios/RNSScreenStackHeaderConfig.m
@@ -355,20 +355,20 @@
   if (shouldHide) {
     return;
   }
-
-  if (config.direction) {
-    switch (config.direction) {
-      case RNSScreenDirectionLeftToRight:
-        navctr.view.semanticContentAttribute = UISemanticContentAttributeForceLeftToRight;
-        navctr.navigationBar.semanticContentAttribute = UISemanticContentAttributeForceLeftToRight;
-        break;
-      case RNSScreenDirectionRightToLeft:
-        navctr.view.semanticContentAttribute = UISemanticContentAttributeForceRightToLeft;
-        navctr.navigationBar.semanticContentAttribute = UISemanticContentAttributeForceRightToLeft;
-        break;
-      default:
-        break;
-    }
+  
+  // if we don't explicitly change the direction in config, we should get it from the phone settings
+  if (config.direction == UISemanticContentAttributeUnspecified) {
+    if ([NSLocale preferredLanguages].count != 0 &&
+         [NSLocale characterDirectionForLanguage:[[NSLocale preferredLanguages] objectAtIndex:0]] == NSLocaleLanguageDirectionRightToLeft) {
+       navctr.view.semanticContentAttribute = UISemanticContentAttributeForceRightToLeft;
+       navctr.navigationBar.semanticContentAttribute = UISemanticContentAttributeForceRightToLeft;
+     } else {
+       navctr.view.semanticContentAttribute = UISemanticContentAttributeForceLeftToRight;
+       navctr.navigationBar.semanticContentAttribute = UISemanticContentAttributeForceLeftToRight;
+     }
+  } else {
+    navctr.view.semanticContentAttribute = config.direction;
+    navctr.navigationBar.semanticContentAttribute = config.direction;
   }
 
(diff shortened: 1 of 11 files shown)


ios/RNSScreenStackHeaderConfig.m
```diff
@@ -355,6 +355,21 @@ + (void)updateViewController:(UIViewController *)vc withConfig:(RNSScreenStackHe
...
+    if ([NSLocale preferredLanguages].count != 0 &&
+         [NSLocale characterDirectionForLanguage:[[NSLocale preferredLanguages] objectAtIndex:0]] == NSLocaleLanguageDirectionRightToLeft) {
+       navctr.view.semanticContentAttribute = UISemanticContentAttributeForceRightToLeft;
+       navctr.navigationBar.semanticContentAttribute = UISemanticContentAttributeForceRightToLeft;
+     } else {
+       navctr.view.semanticContentAttribute = UISemanticContentAttributeForceLeftToRight;
+       navctr.navigationBar.semanticContentAttribute = UISemanticContentAttributeForceLeftToRight;
+     }
+  } else {
+    navctr.view.semanticContentAttribute = config.direction;
```
kmagiera, line 370: is it necessary to set it on the navigation view? Setting on the navbar isn't enough?

WoLewicki: Setting it only on the navbar does not change the side from which we can dismiss the screen with swipe gesture.

ios/RNSScreenStackHeaderConfig.m
```diff
@@ -355,6 +355,21 @@ + (void)updateViewController:(UIViewController *)vc withConfig:(RNSScreenStackHe
   if (shouldHide) {
     return;
   }
+  
+  // if we don't explicitly change the direction in config, we should get it from the phone settings
+  if (config.direction == UISemanticContentAttributeUnspecified) {
+    if ([NSLocale preferredLanguages].count != 0 &&
```
kmagiera, line 361: I think if this isn't the default behavior on the OS (I mean using the system settings by default), we shouldn't be following this approach either. If we want to do this anyways, it should work the same way on Android

WoLewicki: On Android, it works without the need of setting anything despite `android:supportsRtl="true"` in the manifest.

EhsanSarshar:
the issue still exist with new release of "react-native-screens": "^2.10.1",

WoLewicki:
@EhsanSarshar you mean with these changes applied? Because they are not released yet. And what is the exact issue? Can you provide a repo/snack that shows it?

EhsanSarshar:
@WoLewicki the issue is Rtl. the back arrow suppose to be in righ but it's in left

EhsanSarshar:
can I install from the master.

WoLewicki:
You can use [patch-package](https://www.npmjs.com/package/patch-package) and apply the changes from this PR by yourself till the next release.

WoLewicki:
Did you use `direction` prop set to `rtl`?

EhsanSarshar:
no. where should I set the props. the problem is with the Header native component

WoLewicki:
you can set this prop in the `screenOptions` of `native-stack` like this: 
```js
    <NavigationContainer>
      <Stack.Navigator
        screenOptions={{
          direction: 'rtl',
        }}>
        <Stack.Screen name="Settings" component={Settings} />
      </Stack.Navigator>
    </NavigationContainer>
```

EhsanSarshar:
@WoLewicki Thanks a lot it worked

PR #584, opened 2020-08-01
General Housekeeping

codecog:
This PR updates various dev dependencies as well as making some small changes to a couple of the various checks inside props and an update to one of the types to not use `{}`.

The eslint configuration has also had a bit of an update and i fixed the new errors that arose from that change.


.eslintrc.js
```diff
@@ -1,91 +1,55 @@
...
         'import/parsers': {
           '@typescript-eslint/parser': ['.ts', '.tsx'],
         },
-        'import/resolver': {
-          node: {
-            extensions: ['.js', '.ts', '.tsx'],
-          },
-        },
-      },
-      rules: {
```
WoLewicki, line 45: Could you keep the rules that are not included in the `recommended` option?

src/createNativeStackNavigator.js
```diff
@@ -49,7 +49,7 @@ class StackView extends React.Component {
 
   _onFinishTransitioning = () => {
     const { routes } = this.props.navigation.state;
-    let lastRoute = routes && routes.length && routes[routes.length - 1];
+    const lastRoute = routes && routes.length && routes[routes.length - 1];
```
WoLewicki, line 52: It can be shortened to `routes?.length && ...` right?

src/createNativeStackNavigator.js
```diff
@@ -1,20 +1,20 @@
...
 import {
-  StackRouter,
+  createNavigator,
   SceneView,
   StackActions,
-  createNavigator,
-} from 'react-navigation';
-import { HeaderBackButton } from 'react-navigation-stack';
+  StackRouter,
+} from '@react-navigation/native';
```
WoLewicki, line 6: It should be kept as `'react-navigation'` and `'react-navigation-stack'` because it is meant for older versions of react-navigation.

src/native-stack/navigators/createNativeStackNavigator.tsx
```diff
@@ -1,33 +1,33 @@
...
-  NativeStackNavigatorProps,
-  NativeStackNavigationOptions,
   NativeStackNavigationEventMap,
+  NativeStackNavigationOptions,
+  NativeStackNavigatorProps,
 } from '../types';
+import NativeStackView from '../views/NativeStackView';
 
-function NativeStackNavigator(props: NativeStackNavigatorProps) {
+const NativeStackNavigator: React.FC<NativeStackNavigatorProps> = ({
```
WoLewicki, line 19: Why change it to const?

src/native-stack/navigators/createNativeStackNavigator.tsx
```diff
@@ -67,13 +66,13 @@ function NativeStackNavigator(props: NativeStackNavigatorProps) {
 
   return (
     <NativeStackView
-      {...rest}
+      {...props}
```
WoLewicki, line 69: why change `rest` to `props`?

src/native-stack/views/HeaderConfig.tsx
```diff
@@ -1,107 +1,99 @@
...
 } from 'react-native-screens';
-import { Route, useTheme } from '@react-navigation/native';
 import { NativeStackNavigationOptions } from '../types';
 
 type Props = NativeStackNavigationOptions & {
   route: Route<string>;
 };
 
-export default function HeaderConfig(props: Props) {
+const HeaderConfig: React.FC<Props> = ({
```
WoLewicki, line 16: Same as above

src/native-stack/views/HeaderConfig.tsx
```diff
@@ -1,107 +1,99 @@
...
+      backgroundColor={
+        headerStyle.backgroundColor ? headerStyle.backgroundColor : colors.card
       }
-      topInsetEnabled={headerTopInsetEnabled}
       backTitle={headerBackTitleVisible ? headerBackTitle : ' '}
       backTitleFontFamily={headerBackTitleStyle.fontFamily}
       backTitleFontSize={headerBackTitleStyle.fontSize}
-      color={headerTintColor !== undefined ? headerTintColor : colors.primary}
+      blurEffect={headerStyle.blurEffect}
+      color={headerTintColor || colors.primary}
```
WoLewicki, line 54: I think you can apply `tintColor` here to be consistent.

src/native-stack/views/NativeStackView.tsx
```diff
@@ -1,39 +1,41 @@
...
+const isAndroid = Platform.OS === 'android';
 
 type Props = {
   state: StackNavigationState;
   navigation: NativeStackNavigationHelpers;
   descriptors: NativeStackDescriptorMap;
 };
 
-export default function NativeStackView({
+const NativeStackView: React.FC<Props> = ({
```
WoLewicki, line 28: Same as above.

src/native-stack/navigators/createNativeStackNavigator.tsx
codecog: May just be more of a personal thing but reading `...rest` in isolation can give some indirection where using `props` would explictly tell the reader it is the props passed into this component without having to go back to where the component is declared and then check that `...rest` is in fact spreading component props. Happy to revert this change to keep using rest though.

src/native-stack/navigators/createNativeStackNavigator.tsx
codecog: Simply to set the type to `React.FC<NativeStackNavigatorProps>` rather than setting the props type on the argument directly. Fine to change this back if there's seen to be no benefit.

src/native-stack/views/HeaderConfig.tsx
codecog: Same reasoning as why the other was changed. I also feel it's cleaner to destructure the props from inside the `()` rather than stating props and then stating props again along with destructuring them - it can be done in place of.

src/native-stack/views/NativeStackView.tsx
codecog: If it a case of preferring all of these to be reverted? Happy to do that and update the PR if so.

src/native-stack/navigators/createNativeStackNavigator.tsx
WoLewicki: We keep the implementation of this navigator kind of similar to how other react-navigation navigators are implemented, and `rest` is being used there, so I would rather keep it that way.

src/native-stack/navigators/createNativeStackNavigator.tsx
WoLewicki: Please keep it a function. The reasoning is the same, it's similar to how it is done in `react-navigation`.

src/native-stack/views/HeaderConfig.tsx
WoLewicki: I agree that we can destructure the props inside the (), but we should keep it as a function.

src/native-stack/views/HeaderConfig.tsx
WoLewicki: I mean you can do `color={tintColor}` from line 42.

WoLewicki:
Could you also rebase to the newest master? I see there is a conflict with one file.

src/native-stack/views/NativeStackView.tsx
```diff
@@ -42,18 +44,52 @@ export default function NativeStackView({
...
+
         return (
           <Screen
             key={route.key}
             style={StyleSheet.absoluteFill}
-            gestureEnabled={Platform.OS === 'android' ? false : gestureEnabled}
+            gestureEnabled={isAndroid ? false : gestureEnabled}
             stackPresentation={stackPresentation}
             stackAnimation={stackAnimation}
+            onWillAppear={() => {
```
codecog, line 62: @WoLewicki these extra methods were causing the merge conflict but i don't think they exist any longer, especially the emit type options. Only 'appear' and 'dismiss' are available. If i'm to remove then and only keep `onAppear` and `onDismissed` it then says it is a merge conflict.

src/createNativeStackNavigator.js
WoLewicki: You did not change it so why resolve conversation?

src/native-stack/views/NativeStackView.tsx
WoLewicki: Navigation events where recently added and you can see them on `master` https://github.com/software-mansion/react-native-screens/blob/master/src/native-stack/views/NativeStackView.tsx. I don't know why they appear as changes added by you. Did you pull the master branch before rebasing?

src/createNativeStackNavigator.js
codecog: This must have reverted when i resolved conflicts with that other file and it didn't unresolve this. Will push the change again.

codecog:
@WoLewicki Those conflicts should be sorted now, had to add this as a second remote and rebase from there instead of the fork (even though they had the same commit history, weird).

codecog:
Thanks @WoLewicki

PR #593, opened 2020-08-06
Fix ordering of appear / disappear events on Android to match iOS

janicduplessis:
Some potential issues with the different event ordering were pointed out and I think are worth fixing. We're kind of limited by the ordering of animation events on android so I don't really see another way to fix the ordering without adding a lot of complexity.

Android seems to dispatch both animation events right one after the other so I haven't seen any cases of race conditions with simply re-dispatching on the main thread for the appear events.

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenFragment.java
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenFragment.java
@@ -11,6 +11,7 @@ import android.view.inputmethod.InputMethodManager;
 import android.widget.FrameLayout;
 
 import com.facebook.react.bridge.ReactContext;
+import com.facebook.react.bridge.UiThreadUtil;
 import com.facebook.react.uimanager.UIManagerModule;
 
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


kmagiera:
I understand this change is applicable to willAppear and willDisappear only, and the order of these events is what we want to change, no?

Also is the didAppear/didDisappear order correct/opposite to "will" events?

Can you make sure this works correctly when animations are disabled (animation="none")?

WoLewicki:
On Android, the order of `willAppear` and `willDisappear` events is switched, the same for `didAppear/didDisappear`, which can be seen on the gist from #499: https://github.com/software-mansion/react-native-screens/pull/499#issuecomment-667345492. Testing it with disabled animations is a good point for sure.

janicduplessis:
As @WoLewicki mentioned it does apply to both, first change in the diff is for `willAppear` and the second for `didAppear`. 

The correct order is always to have the `xDisappear` one first. The case that was pointed out is if you have 2 screens A and B, A dismisses the keyboard in `willDisappear` and the B opens the keyboard in `willAppear`. When navigating from A to B you'd expect the keyboard to be dismissed and reopened immediately: `Keyboard.dismiss (from screen A) -> Keyboard.open (from screen B)`. However if the `xAppear` is called first it would cause the keyboard to stay closed since it would result in calls like `Keyboard.open (from screen B) -> Keyboard.dismiss (from screen A)`.

WoLewicki:
Closing in favor of #941.

PR #493, opened 2020-05-05
Fix: Add compiler flags for tvOS

hufkens:
At the moment the lib doesn't compile on tvOS. 
I added the necessary compiler flags.

The code was tested in our tvOS/iOS apps.

PR #598, opened 2020-08-10
Separate iOS specific code (Fix tvOS build)

Krisztiaan:
Closes #493 with changes applied and the original author @hufkens credited. @WoLewicki let me know if this would be ok to merge, as we would be using screens on tvOS as well, and currently need `patch-package` for this. Additional flags by @gaborkajtar.

### DIFF
--- a/ios/RNSScreen.m
+++ b/ios/RNSScreen.m
@@ -88,10 +88,12 @@
       break;
     case RNSScreenStackPresentationFullScreenModal:
       _controller.modalPresentationStyle = UIModalPresentationFullScreen;
+#if (TARGET_OS_IOS)
       break;
     case RNSScreenStackPresentationFormSheet:
       _controller.modalPresentationStyle = UIModalPresentationFormSheet;
       break;
+#endif
     case RNSScreenStackPresentationTransparentModal:
       _controller.modalPresentationStyle = UIModalPresentationOverFullScreen;
       break;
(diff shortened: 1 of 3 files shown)


WoLewicki:
So you checked and these are all places where this should be applied so it compiles on tvOS?

krisztiaan-uic:
This way it compiles for tvOS, however, we're still in active initial development of the apps, and are on the lookout if any other modifications are necessary. Currently using these exact changes via `patch-package`.

ios/RNSScreen.m
```diff
@@ -88,10 +88,12 @@ - (void)setStackPresentation:(RNSScreenStackPresentation)stackPresentation
       break;
     case RNSScreenStackPresentationFullScreenModal:
       _controller.modalPresentationStyle = UIModalPresentationFullScreen;
+#if (TARGET_OS_IOS)
       break;
```
WoLewicki, line 92:
```suggestion
      break;
#if (TARGET_OS_IOS)
```

ios/RNSScreenStackHeaderConfig.m
```diff
@@ -383,6 +388,8 @@ + (void)updateViewController:(UIViewController *)vc withConfig:(RNSScreenStackHe
     }
     navitem.largeTitleDisplayMode = config.largeTitle ? UINavigationItemLargeTitleDisplayModeAlways : UINavigationItemLargeTitleDisplayModeNever;
   }
+#endif
+
 #ifdef __IPHONE_13_0
```
WoLewicki, line 393: Why there is no `#if defined(__IPHONE_13_0) && TARGET_OS_IOS` but it is like that in line 537?

Krisztiaan: This is likely an oversight, changed all other occurences to this.

ios/RNSScreen.m
```diff
@@ -413,4 +417,3 @@ @implementation RCTConvert (RNSScreen)
 
 
 @end
-
```
WoLewicki, line 416: There should be blank line at the end of the file.

WoLewicki:
Thank you for your contribution! Looks like it is ready to be merged :tada:

WoLewicki:
Most probably next week.

PR #600, opened 2020-08-12
fix: change update logic

### LINKED ISSUE #432: headerRight randomly position the component incorrectly
I was testing the `createNativeStackNavigation` and ended up having this random issue where the component rendered by `headerRight` is being positioned incorrectly sometimes. There's nothing special in my code, I am just rendering a `<Button>` from `react-native`.

Here's the source code: https://github.com/lnmunhoz/react-native-experiments/blob/master/react-navigation-examples/examples/NativeNavigation.tsx.

Kapture 2020-03-21 at 20 56 50

## Update
The issue also happens when the `headerLargeTitle` is `false`.

image

WoLewicki:
Suggested by @kmagiera change that should resolve #432. 
We wait with adding to the parent controller until the stack is mounted and has its initial layout done. If we add it before layout, some of the items (specifically items from the navigation bar), won't be able to position properly. Also, the position and size of such items, even if it happens to change, won't be properly updated (this is perhaps some internal issue of UIKit). If we add it when the window is not attached, some of the view transitions will be blocked (i.e. modal transitions) and the internal view controller's state will get out of sync with what's on-screen without us knowing.

### DIFF
--- a/ios/RNSScreenStack.m
+++ b/ios/RNSScreenStack.m
@@ -25,11 +25,13 @@
   UINavigationController *_controller;
   NSMutableArray<RNSScreenView *> *_reactSubviews;
   __weak RNSScreenStackManager *_manager;
+  BOOL _hasLayout;
 }
 
 - (instancetype)initWithManager:(RNSScreenStackManager*)manager
 {
   if (self = [super init]) {
+    _hasLayout = NO;
     _manager = manager;
     _reactSubviews = [NSMutableArray new];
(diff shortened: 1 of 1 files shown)


FrankFundel:
When using this, the header is completely missing on first load. After I switch tabs the header appears.

WoLewicki:
@FrankFundel we are aware that this solution is not the best one and are working on finding a better one. Meanwhile, can you share a repo with the bug you mentioned? It would help us find all the existing issues.

PR #601, opened 2020-08-13
Add info about measuring header's height

### LINKED ISSUE #430: How to measure header height?
I implement padding to KeyboardAvoidingView.
In react-navigation I could get header height through useHeaderHeight hook.

```
import { useHeaderHeight } from '@react-navigation/stack';

// ...

const headerHeight = useHeaderHeight();
```

Could you mind share ideas to measure that header height?

WoLewicki:
Added info about measuring header's height on iOS. It can be helpful when using a translucent header. Should resolve #430.

### DIFF
--- a/native-stack/README.md
+++ b/native-stack/README.md
@@ -306,7 +306,18 @@ navigation.popToTop();
 
 ### Measuring header's height on iOS
 
-Using translucent header on iOS can result in the need of measuring your header's height. In order to do it, you can use `react-native-safe-area-context`. It should always be `safeArea.top` + 96 for large header and `safeArea.top` + 44 for small header.
+Using translucent header on iOS can result in the need of measuring your header's height. In order to do it, you can use `react-native-safe-area-context`. It can be measured like this:
+```js
+import {useSafeArea} from 'react-native-safe-area-context';
+
+...
+
+const statusBarInset = useSafeArea().top;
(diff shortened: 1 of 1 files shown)


native-stack/README.md
```diff
@@ -302,6 +302,24 @@ Pops all of the screens in the stack except the first one and navigates to it.
...
+import {useSafeArea} from 'react-native-safe-area-context';
+
+...
+
+const statusBarInset = useSafeArea().top; // inset of the status bar
+const smallHeaderInset = statusBarInset + 44; // inset to use for a small header since it's frame is equal to 44 + the frame of status bar
+const largeHeaderInset = statusBarInset + 96; // inset to use for a large header since it's frame is equal to 96 + the frame of status bar
+```
+
+You can also see an example of using these values with a `ScrollView` here: https://snack.expo.io/@wolewicki/ios-header-height.
```
jakub-gonet, line 320:
```suggestion
Using translucent header on iOS can result in the need of measuring your header's height. In order to do it, you can use `react-native-safe-area-context`. 
Inset of status bar can be measured by using `useSafeArea().top`, the small header has a height of 44 and the big header has a height of 96.

You can see an example of using these values with a `ScrollView` here: https://snack.expo.io/@wolewicki/ios-header-height.
```

PR #602, opened 2020-08-14
feat: enable dev menu on Android

### LINKED ISSUE #276: [Android]Can't access developer menu on physical device
Using `enableScreens` and `createNativeStackNavigator` in `App.js`:
```
import { enableScreens } from 'react-native-screens'
import createNativeStackNavigator from 'react-native-screens/createNativeStackNavigator'

enableScreens()
const AppNavigator = createNativeStackNavigator({
  Login
})
```
I can't access developer menu. But when I replace `createNativeStackNavigator` with `createStackNavigator`, I can access developer menu.

`react-native-screens` version: 2.0.0-alpha.22

WoLewicki:
Overridded method that is called on the key event on which the developer menu on Android should be opened. Change added only for debug mode. Should resolve #276.
Clicking `cmd+m` on macOS in debug mode should open the developer menu on Android, but Action Bar wants to handle that key event in order to trigger `showOverflowMenu` on the Toolbar. We then override this method and call React's `showDevOptionsDialog` in it.

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
@@ -56,10 +56,7 @@ public class ScreenStackHeaderConfig extends ViewGroup {
 
     @Override
     public boolean showOverflowMenu() {
-      Activity activity =  ((ThemedReactContext) getContext()).getCurrentActivity();
-      if (activity != null) {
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
```diff
@@ -45,6 +46,19 @@
   private int mDefaultStartInset;
   private int mDefaultStartInsetWithNavigation;
 
+  private class DebugMenuToolbar extends Toolbar {
```
kmagiera, line 49: static

Issue #105, opened 2019-05-30
Native C++ crashes on Android 9 on React Native 0.59


joshjhargreaves:
We recently upgraded our app from React Native `0.56`to `0.59` and started seeing crashes when navigating between screens on a test device Pixel 3 running Android 9 - we can also reproduce on a Pixel 3 emulator.

The cashes themselves are very consistent and happen often. Disabling `useScreens` on Android stops the crashes from happening. I will work on trying to create a repro project but for now all I have is the stacktrack which I apologize for in advance as I realize how unhelpful that is but thought there might be some value in opening an issue to give some visibility in case it is a general issue and not something tied to the specifics of our app setup.

The crash itself is happening in `libhwui.so`

```
2019-05-28 17:58:14.490 32395-32395/? A/DEBUG: *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
2019-05-28 17:58:14.490 32395-32395/? A/DEBUG: Build fingerprint: 'google/crosshatch/crosshatch:9/PD1A.180720.030/4972053:user/release-keys'
2019-05-28 17:58:14.490 32395-32395/? A/DEBUG: Revision: 'MP1.0'
2019-05-28 17:58:14.490 32395-32395/? A/DEBUG: ABI: 'arm64'
2019-05-28 17:58:14.490 32395-32395/? A/DEBUG: pid: 32138, tid: 32225, name: RenderThread  >>> com.bloomberg.android.plus <<<
2019-05-28 17:58:14.490 32395-32395/? A/DEBUG: signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x20
2019-05-28 17:58:14.491 32395-32395/? A/DEBUG: Cause: null pointer dereference
2019-05-28 17:58:14.491 32395-32395/? A/DEBUG:     x0  0000000000000000  x1  00000075caa43ae4  x2  0000000000000002  x3  000000751f6da518
2019-05-28 17:58:14.491 32395-32395/? A/DEBUG:     x4  0000000000000110  x5  000000751f6d8f14  x6  000000751f6d8f10  x7  000000751f6d8f0c
2019-05-28 17:58:14.491 32395-32395/? A/DEBUG:     x8  0000000000000000  x9  0000000000000000  x10 0000000000000000  x11 0000000000000000
[...]

PR #607, opened 2020-08-20
fix: webview in stack

WoLewicki:
Added check for if there is WebView on the screen we are transitioning to. If there is, we don't change the layer type in stack navigator to avoid crash seen in #105.

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.java
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.java
@@ -153,6 +153,9 @@ public class Screen extends ViewGroup {
     }
     mTransitioning = transitioning;
     boolean isWebViewInScreen = hasWebView(this);
+    if (isWebViewInScreen && getLayerType() != View.LAYER_TYPE_HARDWARE) {
+      return;
+    }
     super.setLayerType(transitioning && !isWebViewInScreen ? View.LAYER_TYPE_HARDWARE : View.LAYER_TYPE_NONE, null);
   }
 
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.java
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.java
@@ -152,17 +152,17 @@ public class Screen extends ViewGroup {
       return;
     }
     mTransitioning = transitioning;
-    boolean isWebViewInScreen = findWebView(this);
+    boolean isWebViewInScreen = hasWebView(this);
     super.setLayerType(transitioning && !isWebViewInScreen ? View.LAYER_TYPE_HARDWARE : View.LAYER_TYPE_NONE, null);
   }
 
-  private boolean findWebView(ViewGroup viewGroup) {
+  private boolean hasWebView(ViewGroup viewGroup) {
     for(int i = 0; i < viewGroup.getChildCount(); i++) {
       View child = viewGroup.getChildAt(i);
       if (child instanceof WebView) {
         return true;
       } else if (child instanceof ViewGroup) {
-         if (findWebView((ViewGroup) child)) {
+         if (hasWebView((ViewGroup) child)) {
            return true;
          }
       }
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/swmansion/rnscreens/Screen.java
```diff
@@ -151,7 +152,22 @@ public void setTransitioning(boolean transitioning) {
       return;
     }
     mTransitioning = transitioning;
-    super.setLayerType(transitioning ? View.LAYER_TYPE_HARDWARE : View.LAYER_TYPE_NONE, null);
+    boolean isWebViewInScreen = findWebView(this);
+    super.setLayerType(transitioning && !isWebViewInScreen ? View.LAYER_TYPE_HARDWARE : View.LAYER_TYPE_NONE, null);
+  }
+
+  private boolean findWebView(ViewGroup viewGroup) {
```
Krisztiaan, line 159: `find` should probably give back the `WebView` itself, not a `boolean` value

```suggestion
    boolean isWebViewInScreen = hasWebView(this);
    super.setLayerType(transitioning && !isWebViewInScreen ? View.LAYER_TYPE_HARDWARE : View.LAYER_TYPE_NONE, null);
  }

  private boolean hasWebView(ViewGroup viewGroup) {
```

android/src/main/java/com/swmansion/rnscreens/Screen.java
```diff
@@ -151,7 +152,22 @@ public void setTransitioning(boolean transitioning) {
...
+    for(int i = 0; i < viewGroup.getChildCount(); i++) {
+      View child = viewGroup.getChildAt(i);
+      if (child instanceof WebView) {
+        return true;
+      } else if (child instanceof ViewGroup) {
+         if (findWebView((ViewGroup) child)) {
```
Krisztiaan, line 165: `find` should probably give back the `WebView` itself, not a `boolean` value

```suggestion
    boolean isWebViewInScreen = hasWebView(this);
    super.setLayerType(transitioning && !isWebViewInScreen ? View.LAYER_TYPE_HARDWARE : View.LAYER_TYPE_NONE, null);
  }

  private boolean hasWebView(ViewGroup viewGroup) {
    for(int i = 0; i < viewGroup.getChildCount(); i++) {
      View child = viewGroup.getChildAt(i);
      if (child instanceof WebView) {
        return true;
      } else if (child instanceof ViewGroup) {
         if (hasWebView((ViewGroup) child)) {
```

Krisztiaan (approved):

Krisztiaan:
Hah, idk how it goes in this repo, but lgtm! :)

chirag04:
@WoLewicki can we merge this?

WoLewicki:
We will see after some discussion if there are no better options. For now, you can use [patch-package](https://www.npmjs.com/package/patch-package) and apply those changes in order to test if everything works right.

android/src/main/java/com/swmansion/rnscreens/Screen.java
```diff
@@ -151,7 +152,22 @@ public void setTransitioning(boolean transitioning) {
       return;
     }
     mTransitioning = transitioning;
-    super.setLayerType(transitioning ? View.LAYER_TYPE_HARDWARE : View.LAYER_TYPE_NONE, null);
+    boolean isWebViewInScreen = hasWebView(this);
+    super.setLayerType(transitioning && !isWebViewInScreen ? View.LAYER_TYPE_HARDWARE : View.LAYER_TYPE_NONE, null);
```
kmagiera, line 156: This line isn't consistent with the PR description. You say that we don't change the layer type in case there is a webview, but it appears that we change it to NONE. Should we avoid updating layer type altogether as stated in the description?

kmagiera (approved):
Rubber stamping this PR, please just address the inline comment I left prior to merging.

android/src/main/java/com/swmansion/rnscreens/Screen.java
WoLewicki: We could not change the layer type, but if it is already `HARDWARE` for some reason, it will probably lead to crash. I can add a check for if the layer is not `HARDWARE` and, if so, just do nothing when there is a `WebView`.

fangasvsass:
can we merge this?

PR #614, opened 2020-08-25
fix: add appContainer to modal screens

### LINKED ISSUE #268: Can't use rn element inspector on modal
When opening rn element inspector while a modal is open, the element inspector gets attached to the window behind. So it's not possible to inspect the modal screen.

No issues with the performance monitor though. 

Simulator Screen Shot - iPhone 11 - 2020-01-06 at 20 45 38

WoLewicki:
Following the implementation of `Modal` in `react-native`: https://github.com/facebook/react-native/blob/master/Libraries/Modal/Modal.js#L206, we attach each screen that is presented modally with an `AppContainer` that has a listener for `LogBox` and `ElementInspector`. This allows for these elements to pop up over the modal screen. There will be 2 `ElementInspectors` on the first screen due to 2 `AppContainer` components rendered in it. If we added check for if the `index === 0`, changing that index would unmount and remount screen content while calling e.g. `navigation.reset()`. Should resolve #268.

### DIFF
--- a/src/native-stack/views/NativeStackView.tsx
+++ b/src/native-stack/views/NativeStackView.tsx
@@ -4,7 +4,8 @@ import {
   useTheme,
 } from '@react-navigation/native';
 import * as React from 'react';
-import { Platform, StyleSheet, View } from 'react-native';
+import { Platform, StyleSheet, View, ViewProps } from 'react-native';
+// @ts-ignore Getting private component
 import AppContainer from 'react-native/Libraries/ReactNative/AppContainer';
 import {
   Screen as ScreenComponent,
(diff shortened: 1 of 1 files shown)


src/native-stack/views/NativeStackView.tsx
```diff
@@ -106,7 +107,16 @@ export default function NativeStackView({
               });
             }}>
             <HeaderConfig {...options} route={route} />
-            <View style={viewStyles}>{renderScene()}</View>
+            {__DEV__ &&
```
kmagiera, line 110: __DEV__ checks can be stripped in build time but not when used like this. You can define a debug container in top level as follows and the use a simple if (__DEV__) statement:

```


let Container = View;

if (__DEV__) {
  const DebugContainer = (props) => <AppContainer><View {...props}/></AppContainer>
  Container = DebugContainer;
}
```

kmagiera (approved):
Please respond to my inline comment. Accepting to unblock otherwise.

brentvatne:
it might be worth opening a pr against react-native to expose this as a public api, what do you think  @kmagiera?

kmagiera:
yes, I think it makes sense, react-native's modal uses that and I think there are some plans to move it out of the core repo. It'd be necessary to expose a way for modal to work outside of the core anyways.

PR #615, opened 2020-08-25
remove unused imports and add Component from React in Example Folder!.

luism3861:
this PR remove unused imports and add `Component` from React in Example Folder!.

### DIFF
--- a/Example/App.js
+++ b/Example/App.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Component } from 'react';
 import {
   Text,
   View,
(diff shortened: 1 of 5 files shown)


WoLewicki:
Thanks for your contribution! Maybe you would like to rewrite these views to functional API instead of classes? It would enable users to test e.g. hooks with new navigation events.

luism3861:
> Thanks for your contribution! Maybe you would like to rewrite these views to functional API instead of classes? It would enable users to test e.g. hooks with new navigation events.

of course i would like to work with rewrite all files with functional API.!

WoLewicki:
I think we can close this PR in favor of the one with functional API, am I right?

luism3861:
> I think we can close this PR in favor of the one with functional API, am I right?

yep you're right

luism3861:
@WoLewicki  yup you're right

PR #188, opened 2019-10-22
Disable interactions on previous screen

sayres:

Issue #120, opened 2019-07-07
[RFC] Allow enabling/disabling screens per subtree instead of globally


satya164:
The current way to enable screens is to call `useScreens(true)` at the root level globally in the app.

However, it'll be better to enable/disable screens for part of the app so we could disable it for problematic areas while still keeping it enabled for rest of the code. We can use the context API to achieve it.

One use case is to disable screens for modals in React Navigation where it seems to be problematic.

The API may look like this:

Configuration for a sub-tree:

```js
<ScreensProvider enabled={true}>
  // Content
</ScreensProider>
```

Configuration per-container:

```js
<ScreensContainer enabled={false}>
  // Content
<ScreensContainer>
```

By default `ScreensContainer` will use the parent configuration for enabling/disabling if this prop is not specified. Another approach will be to wrap this container in `ScreensProvider` and pass `enabled` instead of adding a prop on `ScreenContainer` directly.

Then we can do something like `enabled={mode === 'modal' ? false : true}`

cc @kmagiera @osdnk @brentvatne

WoLewicki:
I believe this option was added in #636, so I am closing this issue. Feel free to comment if I am wrong.

Issue #111, opened 2019-06-09
flickering and blinking view on navigation.goBack()


ammoradi:
Thanks for the awesome library which reduced **react navigation**'s memory usage in my case.
but when **navigation.goBack()** called, the view flickers.

```
react native version: 0.57.8
react navigation version: 2.18.0
react native screens version: 1.0.0-alpha.22
platform: Android
mode: Production
Device: Samsung Galaxy S6
Android Version: 7.0
```

fangasvsass:
react native version: 0.59.9
react navigation version: 3.11.0
react native screens version: 1.0.0-alpha.22 
platform: Android
also hava the problem

ammoradi:
@fangasvsass with your dependencies this bug did not appear.
also, I changed the usage position of `useScreens()` from top of my navigator config to top off my `App.js` file.
the suspicious thing is that I'm not sure this package working on the new position or not!!
#113

kotomono:
@ammoradi did u find out the real reason of the flickering? i have 0.59.7, react navigation version 3.11.0 (even tried the newest one) and alpha.22 of screens but i'm facing that problem. the usescreens() position is at the beginning of the app.js so that should not be the problem

ammoradi:
@kotomono can I see your App.js?
I think this issue highly depends on the position of `usescreens()`

NOTE: beware of using this lib. based on my Firebase Crashlytics report, this lib crashes sometimes. Although because of good memory handling of this lib, I'm using it up to now.

kotomono:
i simply put it on top of the page, even before any other imports

thanks,
Roberto

Il Mer 31 Lug 2019, 10:26 امــیر مــحـمــد <notifications@github.com> ha
scritto:

ammoradi:
@kotomono I used this in my index.js file like this:
```
useScreens();

AppRegistry.registerComponent(appName, () => App);
```

kotomono:
yep did that too.. sadly no changes.. it may be related to something else
in the views?


// @flow
import { useScreens } from 'react-native-screens';

import { AppRegistry } from 'react-native';
import App from 'App/App';
import { name as appName } from './app.json';
useScreens();
AppRegistry.registerComponent(appName, () => App);

‪Il giorno mer 31 lug 2019 alle ore 10:39 ‫امــیر مــحـمــد‬‎ <
notifications@github.com> ha scritto:‬

ammoradi:
@kotomono maybe it is related to `StackNavigators`.
[...]

PR #624, opened 2020-09-04
feat: add optional navigators

### LINKED ISSUE #134: Interaction not disabled on previous screen until slightly after transition
The [updateContainer](https://github.com/kmagiera/react-native-screens/blob/4c2aded42665708851f991fbc13c0b5148705d2f/ios/RNSScreenContainer.m#L86) method has logic to disable interaction during screen transitions, but it doesn't seem to work. I can still tap on buttons from the *previous* screen both during the transition and for a brief moment afterwards. Grainy GIF below, test app [here](https://github.com/jonthanon/react-native-screens-gesture-bug).

react-native-screens-bug

Here's the pseudocode of `updateContainer` as I understand it:

- Detach all screens that still exist but aren’t active anymore
- Detach all screens that were active but don’t exist anymore
- If any screens are newly-active, then…
    - Detach any ***already-active*** screens and disable their user interaction
    - Attach all active screens
- If anything changed to/from active and there’s only one active screen, then…
    - Enable user interaction on the one active screen

I noticed that only screens that were ***already active*** (i.e. the screen is currently active *and* was already in `_activeScreens`) are detached and have their user interaction disabled. Here's what I observe stepping through that method when a new screen is pushed. There are 4 calls to `updateContainer`, and the animated transition occurs between calls 3 and 4:

1. First call
	- Only the first screen is present and is already inactive
    - It gets removed from the active list and detached
2. Second call
	- Both the first and second screens are now present
    - First screen is inactive, second screen is active
    - Neither screen has its user interaction disabled, because this only happens for *already-active* screens
    - Second screen gets attached
    - Since things changed and there's only one active screen, the second screen's user interaction is enabled
3. Third call
	- Both screens still present
	- Both screens are now active
	- Only the second screen gets detached and has its user interaction disabled, because this only happens for *already-active* screens
    - Both screens attached
4. Fourth call (transition has just happened)
	- Both screens present
	- First screen is now inactive again
    - First screen gets detached
    - Since things changed and there's only one active screen, the second screen's user interaction is enabled

So, during the transition, the second screen's user interaction is enabled, disabled, and enabled again. Notice that the user interaction for the first screen is never changed. That said, if I observe the actual value of `userInteractionEnabled` for the two screens' corresponding UIViews at the start and end of `updateContainer`, I see wildly different things. The first screen always starts enabled, and is disabled by the end of the method. The second screen is usually enabled throughout, except for the fourth/final call, when it starts out disabled and ends up enabled. So clearly there are some side effects going on somewhere that I'm missing.

In any case, **the result is that I can still tap on buttons in the *previous* screen both during the transition and for a brief moment afterward**. At *some* point, possibly during or after the fourth call to `updateContainer`, something gestures to stop being triggered on the first screen, and enables them on the second screen, but it takes longer than it should (probably related to the delay people are seeing [here](https://github.com/kmagiera/react-native-screens/issues/95)).

WoLewicki:
Changed the behavior of `active` prop to contain 3 options (values):
- 0: Screen is inactive
- 1: Screen is active, but is not visible and is not the first responder, or is transitioning
- 2: Screen is active and on top  

Added new prop `shouldUseActivityState` to differentiate between different implementations for other libraries to choose their implementation of JS side.

Should resolve #134, #120, #61. Should be also possible to use this as a workaround for #111. Should change the behavior described in #95 and probably resolve it. Should also close #130, #188 due to change of logic.

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
@@ -239,10 +239,6 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
     return screenFragment.getScreen().getActivityState();
   }
 
-  protected boolean isScreenActive(ScreenFragment screenFragment) {
-    return getActivityState(screenFragment) != Screen.ActivityState.INACTIVE;
-  }
-
   protected boolean hasScreen(ScreenFragment screenFragment) {
     return mScreenFragments.contains(screenFragment);
   }
@@ -339,8 +335,7 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
     Set<Fragment> orphaned = new HashSet<>(mFragmentManager.getFragments());
     for (int i = 0, size = mScreenFragments.size(); i < size; i++) {
       ScreenFragment screenFragment = mScreenFragments.get(i);
-      boolean isActive = isScreenActive(screenFragment);
-      if (!isActive && screenFragment.isAdded()) {
+      if (getActivityState(screenFragment) == Screen.ActivityState.INACTIVE && screenFragment.isAdded()) {
         detachScreen(screenFragment);
       }
       orphaned.remove(screenFragment);
@@ -360,9 +355,8 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
 
     for (int i = 0, size = mScreenFragments.size(); i < size; i++) {
       ScreenFragment screenFragment = mScreenFragments.get(i);
-      Screen.ActivityState active = getActivityState(screenFragment);
-      if (active == Screen.ActivityState.ON_TOP) {
-        // top screen gets the active value of 2 after the end of transition
+      if (getActivityState(screenFragment) == Screen.ActivityState.ON_TOP) {
+        // if there is an "onTop" screen it means the transition has ended
         transitioning = false;
       }
     }
@@ -371,11 +365,11 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
     boolean addedBefore = false;
     for (int i = 0, size = mScreenFragments.size(); i < size; i++) {
       ScreenFragment screenFragment = mScreenFragments.get(i);
-      boolean isActive = isScreenActive(screenFragment);
-      if (isActive && !screenFragment.isAdded()) {
+      Screen.ActivityState activityState = getActivityState(screenFragment);
+      if (activityState != Screen.ActivityState.INACTIVE && !screenFragment.isAdded()) {
         addedBefore = true;
         attachScreen(screenFragment);
-      } else if (isActive && addedBefore) {
+      } else if (activityState != Screen.ActivityState.INACTIVE && addedBefore) {
         moveToFront(screenFragment);
       }
       screenFragment.getScreen().setTransitioning(transitioning);
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
@@ -208,12 +208,12 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
       mProcessingTransaction.runOnCommit(new Runnable() {
         @Override
         public void run() {
-         if (mProcessingTransaction == transaction) {
-           // we need to take into account that commit is initiated with some other transaction while
-           // the previous one is still processing. In this case mProcessingTransaction gets overwritten
-           // and we don't want to set it to null until the second transaction is finished.
-           mProcessingTransaction = null;
-         }
+          if (mProcessingTransaction == transaction) {
+            // we need to take into account that commit is initiated with some other transaction while
+            // the previous one is still processing. In this case mProcessingTransaction gets overwritten
+            // and we don't want to set it to null until the second transaction is finished.
+            mProcessingTransaction = null;
+          }
         }
       });
       mCurrentTransaction.commitAllowingStateLoss();
@@ -235,8 +235,12 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
     getOrCreateTransaction().remove(screenFragment);
   }
 
+  protected Screen.ActivityState getActivityState(ScreenFragment screenFragment) {
+    return screenFragment.getScreen().getActivityState();
+  }
+
   protected boolean isScreenActive(ScreenFragment screenFragment) {
-    return screenFragment.getScreen().isActive();
+    return getActivityState(screenFragment) != Screen.ActivityState.INACTIVE;
   }
 
   protected boolean hasScreen(ScreenFragment screenFragment) {
@@ -352,14 +356,16 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
       }
     }
 
-    // detect if we are "transitioning" based on the number of active screens
-    int activeScreens = 0;
+    boolean transitioning = true;
+
     for (int i = 0, size = mScreenFragments.size(); i < size; i++) {
-      if (isScreenActive(mScreenFragments.get(i))) {
-        activeScreens += 1;
+      ScreenFragment screenFragment = mScreenFragments.get(i);
+      Screen.ActivityState active = getActivityState(screenFragment);
+      if (active == Screen.ActivityState.ON_TOP) {
+        // top screen gets the active value of 2 after the end of transition
+        transitioning = false;
       }
     }
-    boolean transitioning = activeScreens > 1;
 
     // attach newly activated screens
     boolean addedBefore = false;
@@ -374,6 +380,7 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
       }
       screenFragment.getScreen().setTransitioning(transitioning);
     }
+
     tryCommitTransaction();
   }
 }
--- a/ios/RNSScreenContainer.m
+++ b/ios/RNSScreenContainer.m
@@ -145,30 +145,30 @@
 - (void)updateContainer
 {
   _needUpdate = NO;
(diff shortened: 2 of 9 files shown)


kmagiera (changes requested):
I don't quite understand the need of `isTop` property. I thought this can be resolved by adding a single prop (transitioning) and not two new props in total. Can you explain why we only want to change layer for the top screen? Why can't we use `active` prop for that? It appears to me like the role of `isTop` is very similar to `active` so maybe we should adjust the way `active` works?

WoLewicki:
We change the layer of all screens in a container, but the screen that does it, is the one that is transitioning. I don't think we can use `active` prop since there can be multiple active screens at once and do we have a way to decide which one of them is the top screen and should be a responder on iOS?

kmagiera:
What is the meaning of `transitioning` property then? If we go back on stack which screen is transitioning? I think when we push new one it is more or less clear.

WoLewicki:
`transitioning` prop follows the transition. `active` can not turn to 0 after the transition, so it cannot be used to decide if the transition ended. When going back, the screen that is moving out is transitioning, it has the highest index till the end of the transition, and after that, it is removed and the screen below gets the highest index. Going forward, the newly attached screen gets the highest index and it is the one transitioning too.

android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
```diff
@@ -235,8 +235,12 @@ private void detachScreen(ScreenFragment screenFragment) {
     getOrCreateTransaction().remove(screenFragment);
   }
 
+  protected int getScreenActive(ScreenFragment screenFragment) {
+    return screenFragment.getScreen().getActive();
+  }
+
   protected boolean isScreenActive(ScreenFragment screenFragment) {
-    return screenFragment.getScreen().isActive();
+    return getScreenActive(screenFragment) != 0;
```
kmagiera, line 243: it looks like after this change `active` changes its semantic. We should avoid calling it the same way when thinking of it as meaning different things. In particular you can get `active` as an int and also get active as boolean. This suggest that active has different meanins and we should use different naming for these two things

android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
```diff
@@ -352,14 +356,16 @@ protected void performUpdate() {
...
-    // detect if we are "transitioning" based on the number of active screens
-    int activeScreens = 0;
+    boolean transitioning = true;
+
     for (int i = 0, size = mScreenFragments.size(); i < size; i++) {
-      if (isScreenActive(mScreenFragments.get(i))) {
-        activeScreens += 1;
+      ScreenFragment screenFragment = mScreenFragments.get(i);
+      int active = getScreenActive(screenFragment);
+      if (active == 2) {
```
kmagiera, line 364: can we define a constant for this and name it appropriately such that this code is easier to follow

src/index.d.ts
```diff
@@ -48,7 +48,7 @@ declare module 'react-native-screens' {
   export type ScreenReplaceTypes = 'push' | 'pop';
 
   export interface ScreenProps extends ViewProps {
-    active?: 0 | 1 | Animated.AnimatedInterpolation;
+    active?: 0 | 1 | 2 | Animated.AnimatedInterpolation;
```
kmagiera, line 51: I'd perhaps rename this to something like "activityState" and keep old "active" prop for compatibility while just transforming its value.

Would also rename "shouldUseNewImplementation" to something that gives more clue (will "new" still be new in five months?). So given we rename prop to "activityState" then the other one can be named "shouldUseActivityState"

kmagiera (commented):
This is going in the right direction 💯 – left a few comments inline

ios/RNSScreenContainer.m
```diff
@@ -119,7 +119,7 @@ - (void)updateContainer
   // remove screens that are no longer active
   NSMutableSet *orphaned = [NSMutableSet setWithSet:_activeScreens];
   for (RNSScreenView *screen in _reactSubviews) {
-    if (!screen.active && [_activeScreens containsObject:screen]) {
+    if (screen.activityState == 0 && [_activeScreens containsObject:screen]) {
```
kmagiera, line 122: can we have an enum or named constant in objc code too?

src/index.native.js
```diff
@@ -110,8 +113,20 @@ class Screen extends React.Component {
 
       // When using RN from master version is 0.0.0
       if (version.minor >= 57 || version.minor === 0) {
-        const { enabled, ...rest } = this.props;
-        return <AnimatedNativeScreen {...rest} ref={this.setRef} />;
+        let { enabled, active, activityState, ...rest } = this.props;
+        if (active !== undefined && activityState === undefined) {
+          console.warn(
+            'You are using old React Navigation version. Update your React Navigation navigators to use the best working solution.'
```
kmagiera, line 119: I'd structure this message along these lines:
Appears like you are using old version of react-navigation library. Please update react-navigation to version X.Y.Z or above to take full advantage of new functionality added to react-native-screens

ios/RNSScreenContainer.m
```diff
@@ -133,38 +133,41 @@ - (void)updateContainer
   // detect if new screen is going to be activated
   BOOL activeScreenAdded = NO;
   for (RNSScreenView *screen in _reactSubviews) {
-    if (screen.active && ![_activeScreens containsObject:screen]) {
+    if (screen.activityState != 0 && ![_activeScreens containsObject:screen]) {
       activeScreenAdded = YES;
     }
   }
 
+  if (activeScreenRemoved || activeScreenAdded) {
```
kmagiera, line 141: Do we really need this fragment here? We already have some code for handling "interactionEnabled" prop below that checks the same condition, maybe we can put these fragments next to each other?

kmagiera (approved):
Accepting as I only hav some minor comments. Note that I also added a couple of comments in the previous review batch but mistakenly clicked "comment". Please address my previous comments too prior to merging

android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
```diff
@@ -235,8 +235,12 @@ private void detachScreen(ScreenFragment screenFragment) {
     getOrCreateTransaction().remove(screenFragment);
   }
 
+  protected Screen.ActivityState getActivityState(ScreenFragment screenFragment) {
+    return screenFragment.getScreen().getActivityState();
+  }
+
   protected boolean isScreenActive(ScreenFragment screenFragment) {
-    return screenFragment.getScreen().isActive();
+    return getActivityState(screenFragment) != Screen.ActivityState.INACTIVE;
```
kmagiera, line 243: I think it'd be better to abandon this method altogether. We only use it in `performUpdate` where one out of three calls have been replaced with direct comparison of activity state. We should do the same with two remaining calls. This method is not used outside + I think that using term "active" is going to cause more confusion

android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
```diff
@@ -352,14 +356,16 @@ protected void performUpdate() {
...
-    // detect if we are "transitioning" based on the number of active screens
-    int activeScreens = 0;
+    boolean transitioning = true;
+
     for (int i = 0, size = mScreenFragments.size(); i < size; i++) {
-      if (isScreenActive(mScreenFragments.get(i))) {
-        activeScreens += 1;
+      ScreenFragment screenFragment = mScreenFragments.get(i);
+      Screen.ActivityState active = getActivityState(screenFragment);
+      if (active == Screen.ActivityState.ON_TOP) {
```
kmagiera, line 364: Why is the activity state variable called "active" – considering the change made here I'd prefer to avoid further confusion and use one term over the other. Also I think we don't need this variable here at all

```suggestion
      if (getActivityState(screenFragment) == Screen.ActivityState.ON_TOP) {
```

ios/RNSScreenContainer.m
```diff
@@ -164,38 +163,39 @@ - (void)updateContainer
...
   }
 
-  // if we are down to one active screen it means the transitioning is over and we want to notify
-  // the transition has finished
-  if ((activeScreenRemoved || activeScreenAdded) && _activeScreens.count == 1) {
-    RNSScreenView *singleActiveScreen = [_activeScreens anyObject];
-    // restore interactions
-    singleActiveScreen.userInteractionEnabled = YES;
-    [singleActiveScreen notifyFinishTransitioning];
+  if (activeScreenRemoved || activeScreenAdded) {
```
kmagiera, line 188: rename these two as well, for example: `screenRemoved || screenAdded`

ios/RNSScreenContainer.m
```diff
@@ -164,38 +163,39 @@ - (void)updateContainer
...
+    // we disable interaction for during the transition until one of the screens has active == 2
```
kmagiera, line 189:
```suggestion
    // we disable interaction for the duration of the transition until one of the screens changes its state to "onTop"
```

ios/RNSScreenContainer.m
```diff
@@ -164,38 +163,39 @@ - (void)updateContainer
...
+    self.userInteractionEnabled = NO;
+    
+    for (RNSScreenView *screen in _reactSubviews) {
+      if (screen.activityState == RNSActivityStateOnTop) {
+        // if there is a screen with active == 2, it means the transition has ended, so we restore interactions
```
kmagiera, line 194:
```suggestion
        // if there is an "onTop" screen it means the transition has ended so we restore interactions
```

src/index.native.js
```diff
@@ -214,4 +229,6 @@ module.exports = {
   enableScreens,
   useScreens,
   screensEnabled,
+
+  shouldUseActivityState,
```
kmagiera, line 233:
```suggestion
  shouldUseActivityState,
```

kmagiera (approved):

PR #642, opened 2020-09-17
feat: add status bar managment

WoLewicki:
Added status bar management on iOS based on the VCs. The user must apply change described in README to `AppDelegate.m` and to the project's `Info.plist` in order for the new props to work.

### DIFF
--- a/ios/RNSScreen.h
+++ b/ios/RNSScreen.h
@@ -27,13 +27,6 @@ typedef NS_ENUM(NSInteger, RNSScreenReplaceAnimation) {
   RNSScreenReplaceAnimationPush,
 };
 
-typedef NS_ENUM(NSInteger, RNSStatusBarStyle) {
-  RNSStatusBarStyleAuto,
-  RNSStatusBarStyleInverted,
-  RNSStatusBarStyleLight,
-  RNSStatusBarStyleDark,
-};
-
 @interface RCTConvert (RNSScreen)
 
 + (RNSScreenStackPresentation)RNSScreenStackPresentation:(id)json;
--- a/ios/RNSScreen.h
+++ b/ios/RNSScreen.h
@@ -38,7 +38,6 @@ typedef NS_ENUM(NSInteger, RNSStatusBarStyle) {
 
 + (RNSScreenStackPresentation)RNSScreenStackPresentation:(id)json;
 + (RNSScreenStackAnimation)RNSScreenStackAnimation:(id)json;
-+ (RNSStatusBarStyle)RNSStatusBarStyle:(id)json;
 
 @end
 
@@ -68,9 +67,6 @@ typedef NS_ENUM(NSInteger, RNSStatusBarStyle) {
 @property (nonatomic) RNSScreenStackAnimation stackAnimation;
 @property (nonatomic) RNSScreenStackPresentation stackPresentation;
 @property (nonatomic) RNSScreenReplaceAnimation replaceAnimation;
-@property (nonatomic) RNSStatusBarStyle statusBarStyle;
-@property (nonatomic) UIStatusBarAnimation statusBarAnimation;
-@property (nonatomic) BOOL statusBarHidden;
 
 - (void)notifyFinishTransitioning;
 
--- a/ios/RNSScreen.h
+++ b/ios/RNSScreen.h
@@ -61,6 +61,7 @@ typedef NS_ENUM(NSInteger, RNSScreenReplaceAnimation) {
 @property (nonatomic) RNSScreenReplaceAnimation replaceAnimation;
 @property (nonatomic) UIStatusBarStyle statusBarStyle;
 @property (nonatomic) UIStatusBarAnimation statusBarAnimation;
+@property (nonatomic) BOOL statusBarHidden;
 
 - (void)notifyFinishTransitioning;
 
--- a/ios/RNSScreen.h
+++ b/ios/RNSScreen.h
@@ -27,10 +27,18 @@ typedef NS_ENUM(NSInteger, RNSScreenReplaceAnimation) {
   RNSScreenReplaceAnimationPush,
 };
 
+typedef NS_ENUM(NSInteger, RNSStatusBarStyle) {
+  RNSStatusBarStyleAuto,
+  RNSStatusBarStyleInverted,
+  RNSStatusBarStyleLight,
+  RNSStatusBarStyleDark,
+};
+
 @interface RCTConvert (RNSScreen)
 
 + (RNSScreenStackPresentation)RNSScreenStackPresentation:(id)json;
 + (RNSScreenStackAnimation)RNSScreenStackAnimation:(id)json;
++ (RNSStatusBarStyle)RNSStatusBarStyle:(id)json;
 
 @end
 
@@ -59,7 +67,7 @@ typedef NS_ENUM(NSInteger, RNSScreenReplaceAnimation) {
 @property (nonatomic) RNSScreenStackAnimation stackAnimation;
 @property (nonatomic) RNSScreenStackPresentation stackPresentation;
 @property (nonatomic) RNSScreenReplaceAnimation replaceAnimation;
-@property (nonatomic) UIStatusBarStyle statusBarStyle;
+@property (nonatomic) RNSStatusBarStyle statusBarStyle;
 @property (nonatomic) UIStatusBarAnimation statusBarAnimation;
 @property (nonatomic) BOOL statusBarHidden;
 
--- a/ios/RNSScreen.h
+++ b/ios/RNSScreen.h
@@ -34,7 +34,7 @@ typedef NS_ENUM(NSInteger, RNSScreenReplaceAnimation) {
 
 @end
 
-@interface RNSScreen : UIViewController <RNScreensViewControllerDelegate>
+@interface RNSScreen : UIViewController
 
 - (instancetype)initWithView:(UIView *)view;
 - (void)notifyFinishTransitioning;
--- a/ios/RNSScreen.h
+++ b/ios/RNSScreen.h
@@ -1,6 +1,7 @@
 #import <React/RCTViewManager.h>
 #import <React/RCTView.h>
 #import <React/RCTComponent.h>
+
 #import "RNSScreenContainer.h"
 
 @class RNSScreenContainerView;
@@ -34,7 +35,7 @@ typedef NS_ENUM(NSInteger, RNSScreenReplaceAnimation) {
 
 @end
 
-@interface RNSScreen : UIViewController
+@interface RNSScreen : UIViewController <RNScreensViewControllerDelegate>
 
 - (instancetype)initWithView:(UIView *)view;
 - (void)notifyFinishTransitioning;
--- a/ios/RNSScreen.h
+++ b/ios/RNSScreen.h
@@ -34,7 +34,7 @@ typedef NS_ENUM(NSInteger, RNSScreenReplaceAnimation) {
 
 @end
 
-@interface RNSScreen : UIViewController
+@interface RNSScreen : UIViewController <RNScreensViewControllerDelegate>
 
 - (instancetype)initWithView:(UIView *)view;
 - (void)notifyFinishTransitioning;
@@ -43,6 +43,7 @@ typedef NS_ENUM(NSInteger, RNSScreenReplaceAnimation) {
 @end
 
 @interface RNSScreenManager : RCTViewManager
+
 @end
 
 @interface RNSScreenView : RCTView
@@ -52,7 +53,7 @@ typedef NS_ENUM(NSInteger, RNSScreenReplaceAnimation) {
 @property (nonatomic, copy) RCTDirectEventBlock onDismissed;
 @property (nonatomic, copy) RCTDirectEventBlock onWillAppear;
 @property (nonatomic, copy) RCTDirectEventBlock onWillDisappear;
-@property (weak, nonatomic) UIView<RNSScreenContainerDelegate> *reactSuperview;
+@property (weak, nonatomic) UIView <RNSScreenContainerDelegate> *reactSuperview;
 @property (nonatomic, retain) UIViewController *controller;
 @property (nonatomic, readonly) BOOL dismissed;
 @property (nonatomic) BOOL active;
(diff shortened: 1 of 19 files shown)


kmagiera:
@brentvatne suggested that we don't stick with the core RN's statubar naming convention as it isn't intuitive when accounting for light/dark mode. Instead we could use the following options for the status bar color: 'auto', 'inverted', 'light', or 'dark' (see the new Expo's status bar API: https://docs.expo.io/versions/latest/sdk/status-bar/#statusbarstyle)

native-stack/README.md
```diff
@@ -179,6 +179,31 @@ How the given screen should appear/disappear when pushed or popped at the top of
 
 Defaults to `default`.
 
+### Status bar managment
+
+With `native-stack`, the status bar can be managed by `UIViewController` on iOS. It requires:
+
+1. Enabling (or deleting) `View controller-based status bar appearance` in your Info.plist file (and so disables the option to use React Native's `StatusBar` component).
```
WoLewicki, line 186:
```suggestion
1. Enabling (or deleting) `View controller-based status bar appearance` in your Info.plist file (it disables the option to use React Native's `StatusBar` component). 
```

ios/RNSScreen.m
```diff
@@ -142,6 +148,54 @@ - (void)setStackAnimation:(RNSScreenStackAnimation)stackAnimation
...
+        [self->_controller setNeedsStatusBarAppearanceUpdate];
+      }];
+    }
+  }
+}
+
+- (void)setStatusBarHidden:(BOOL)statusBarHidden
+{
+  if (statusBarHidden != _statusBarHidden) {
+    if (![[[NSBundle mainBundle] objectForInfoDictionaryKey:@"UIViewControllerBasedStatusBarAppearance"] boolValue]) {
```
kmagiera, line 186: Maybe extract this to a separate method? We don't need special warning per parameter. Alternatively it'd be the best to evaluate objectForInfoDictionaryKey call in global scope and store it in a const, mainBundle cannot be changed in runtime so it should be safe to only evaluate it this way

ios/RNSScreen.m
```diff
@@ -333,7 +458,9 @@ - (void)willMoveToParentViewController:(UIViewController *)parent
 - (void)viewWillAppear:(BOOL)animated
 {
   [super viewWillAppear:animated];
-
+  [UIView animateWithDuration:0.5 animations:^{
```
kmagiera, line 461: why 0.5? If this is needed then please add some comment to explain this

ios/RNSScreenContainer.m
```diff
@@ -148,8 +165,6 @@ - (void)updateContainer
           // for screens that were already active we want to mimick the effect UINavigationController
           // has when willMoveToWindow:nil is triggered before the animation starts
           [self prepareDetach:screen];
-          // disable interactions for the duration of transition
```
kmagiera, line 151: these changes seem irrelevant? bad merge?

kmagiera (commented):
left a few comments inline

ios/RNSScreen.m
WoLewicki: It is an arbitrary value taken from some SO thread. `Programming iOS 13` book uses a 0.4-second long transition but also doesn't specify why such value exactly, so I have no answer for this.

ios/RNSScreen.m
```diff
@@ -142,6 +149,50 @@ - (void)setStackAnimation:(RNSScreenStackAnimation)stackAnimation
...
+    _controller.modalPresentationCapturesStatusBarAppearance = YES;
+    [UIView animateWithDuration:0.5 animations:^{
+      [self->_controller setNeedsStatusBarAppearanceUpdate];
+    }];
+  }
+}
+
++ (void)checkUIVCBasedStatusBarAppearance
+{
+  if (viewControllerStatusBarDisabled) {
```
kmagiera, line 190: there is no need to define variables outside or mark them as extern. I'd perhaps rename this method to `assertViewControllerBasedStatusBarAppearenceSet` then the implementation should look as follows:


```objc
+ (void)assertViewControllerBasedStatusBarAppearenceSet
{
    static dispatch_once_t once;
    static bool viewControllerBasedAppearence;
    dispatch_once(&once, ^{
viewControllerBasedAppearence = [[[NSBundle mainBundle] objectForInfoDictionaryKey:@"UIViewControllerBasedStatusBarAppearance"] boolValue];
    });
    if (!viewControllerBasedAppearence) {
      RCTLogError(@"........");
    }
}
```

ios/RNSScreen.m
kmagiera: we should have that note included in the code as a comment then (preferably including the link to SO answer)

ios/RNSScreenStackHeaderConfig.h
```diff
@@ -50,7 +60,8 @@ typedef NS_ENUM(NSInteger, RNSScreenStackHeaderSubviewType) {
 
 + (RNSScreenStackHeaderSubviewType)RNSScreenStackHeaderSubviewType:(id)json;
 + (UIBlurEffectStyle)UIBlurEffectStyle:(id)json;
-+ (UISemanticContentAttribute)RNSScreenDirection:(id)json;
++ (UISemanticContentAttribute)UISemanticContentAttribute:(id)json;
```
WoLewicki, line 63: I changed it beacuse the method name was a relict of some of the previous implementations and could be misleading.

ios/RNSScreenStack.m
```diff
@@ -17,6 +18,28 @@ @interface RNSScreenStackView () <UINavigationControllerDelegate, UIAdaptivePres
...
+
+- (UIStatusBarStyle)preferredStatusBarStyle
+{
+  UIViewController *childVC =  [[self childViewControllers] lastObject];
+  return [childVC isKindOfClass:[RNSScreen class]] ? childVC.preferredStatusBarStyle : UIStatusBarStyleDefault;
+}
+
+- (UIStatusBarAnimation)preferredStatusBarUpdateAnimation
+{
+  return [[self childViewControllers] lastObject].preferredStatusBarUpdateAnimation;
```
kmagiera, line 31: maybe we should use `topViewController` instead of `childViewControllers lastObject` here and in the previous method?

ios/UIViewController+RNScreens.m
```diff
@@ -0,0 +1,52 @@
...
+  return [self reactNativeScreensChildViewControllerForStatusBarHidden];
+}
+
+- (UIStatusBarAnimation)reactNativeScreensPreferredStatusBarUpdateAnimation
+{
+  UIViewController* lastViewController = [[self childViewControllers] lastObject];
+  if ([lastViewController isKindOfClass:[RNScreensNavigationController class]] || [lastViewController isKindOfClass:[RNScreensViewController class]] || [lastViewController isKindOfClass:[RNSScreen class]]) {
+    return lastViewController.preferredStatusBarUpdateAnimation;
+  }
+  return [self reactNativeScreensPreferredStatusBarUpdateAnimation];
```
kmagiera, line 35: isn't this an infinite loop?

ios/UIViewController+RNScreens.m
```diff
@@ -0,0 +1,52 @@
...
+
+#import <objc/runtime.h>
+
+
+@implementation UIViewController (RNScreens)
+
+- (UIViewController *)reactNativeScreensChildViewControllerForStatusBarStyle
+{
+  UIViewController* lastViewController = [[self childViewControllers] lastObject];
+  if ([lastViewController isKindOfClass:[RNScreensNavigationController class]] || [lastViewController isKindOfClass:[RNScreensViewController class]] || [lastViewController isKindOfClass:[RNSScreen class]]) {
```
kmagiera, line 14: maybe we can define a protocol that all RNScreen navigators conform to in order to avoid all these checks.

ios/UIViewController+RNScreens.m
```diff
@@ -0,0 +1,52 @@
...
+    return lastViewController;
```
kmagiera, line 15: we have all these swizzled method defined for RNScreensViewController and for RNScreenNavigationController. It seems to me that it'd be better to redirect the call to those implementations instead of having it implemented twice. I understand that what we want to do here is to forward calls to last child controller, but we only want to do it so long `self` is not one of the view controllers defined by screens library (RNScreensVC or RNScreenNavConroller)

kmagiera (approved):
Left a couple more comments. Accepting to unblock

ios/UIViewController+RNScreens.m
WoLewicki: It will call the base method, that is how the swizzling works (based on 
 https://nshipster.com/method-swizzling/#invoking-_cmd)

ios/RNSScreen.m
```diff
@@ -295,6 +297,63 @@ - (instancetype)initWithView:(UIView *)view
...
+  }
+  RNSScreenStackHeaderConfig *config = [self findConfigForScreen];
+  return config && config.statusBarHidden ? config.statusBarHidden : NO;
+}
+
+- (RNSScreenStackHeaderConfig *)findConfigForScreen
+{
+  // if there is no child navigator and the parent is `RNSScreenContainer`, we should fallback to the parent's (that is not `RNSScreenContainer`) option
+  UIViewController *parent = [self parentViewController];
+  while ([parent isKindOfClass:[RNScreensViewController class]] || [parent isKindOfClass:[RNSScreen class]]) {
```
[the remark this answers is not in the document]
WoLewicki: Would it enter the loop if the `parent` was nil at the beginning? I think this while condition is not that bad since we iterate as long as the parent is one of those two kinds of VC. If we spotted something else, or we got to the top of the hierarchy, loop would stop.

ios/RNSScreen.m
```diff
@@ -295,6 +297,63 @@ - (instancetype)initWithView:(UIView *)view
...
+    if ([parent parentViewController] == nil && [parent isKindOfClass:[RNSScreen class]]) {
+      // we are at the top of hierarchy and the controller is RNSScreen so we are in modal screen
+      break;
+    } else {
+      parent = [parent parentViewController];
+    }
+  }
+  RNSScreenView *screenView = [parent isKindOfClass:[RNScreensNavigationController class]]
```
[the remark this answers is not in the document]
WoLewicki: If it is not, we will go to the `else` section of the second ternary operator, where we just return the `self.view`, which in the case of `RNSScreen` is always `RNSScreenView`. It probably is good to check if the child of `RNScreensNavigationController` is `RNSScreen` for sure and add comments about the other options.

ios/UIViewController+RNScreens.m
WoLewicki: `RNScreenNavigationController` doesn't use the swizzled methods, probably due to `UINavigationController` having own implementations of them, that don't call `super`, so I am not sure if it would help with anything.

kmagiera: How could then it be possible that `lastViewController` is of type `RNSScreen`? We should olways expect it to be of a container type (either RNSSVC or RNSNC)

WoLewicki: These methods can be used in `RNScreensViewController` too iirc.

ios/RNSScreen.m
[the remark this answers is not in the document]
kmagiera: What if instead of doing this parent traversal we eliminate the need of accessing the config in childViewControllerForStatusBarStyle and others? It looks like it'd be much easier to do the up-to-bottom traversal when we're asked to provide child component as opposed to always returning some child VC and in that child VC doing a reverse traversal to find a parent where config is set

ios/RNSScreen.h
```diff
@@ -51,7 +53,7 @@ typedef NS_ENUM(NSInteger, RNSScreenReplaceAnimation) {
 @property (nonatomic, copy) RCTDirectEventBlock onDismissed;
 @property (nonatomic, copy) RCTDirectEventBlock onWillAppear;
 @property (nonatomic, copy) RCTDirectEventBlock onWillDisappear;
-@property (weak, nonatomic) UIView<RNSScreenContainerDelegate> *reactSuperview;
+@property (weak, nonatomic) UIView <RNSScreenContainerDelegate> *reactSuperview;
```
kmagiera, line 56:
```suggestion
@property (weak, nonatomic) UIView<RNSScreenContainerDelegate> *reactSuperview;
```

kmagiera (approved):
Thanks for addressing all the previous comments. IMO it is much cleaner than all the previous iterations. Good job 👍

jacobp100:
Thanks for working on this! I'm not on a project that needs it at the moment, but it will be really helpful in the future! 😀

a-eid:
how does this feature work with expo based apps ?

brentvatne:
@a-eid - it's not enabled in Expo Go because we'd have to enable view controller based status bar appearance, but [in the nearish future it should be possible](https://blog.expo.io/expo-managed-workflow-in-2021-d1c9b68aa10)

a-eid:
@brentvatne thank you, I've read about this a few days ago on twitter. 
pretty existing. 

you mentioned that this is not enabled in `Expo Go`, does it work if you build the app ? 
- I've been trying `Statusbar` from RN and `expo-statusbar` and both of them don't work work `native-stack`.

brentvatne:
those will work in native-stack just fine. let me know what issues you encounter with it and include a minimal reproducible example in https://github.com/expo/expo/issues

a-eid:
@brentvatne yes they do, was a mistake on my part. 

thank you

brentvatne:
@a-eid - `RCTStatusBarManager` is a module from upstream react-native. if you don't use React Native's StatusBar API then you won't get the warning: https://github.com/facebook/react-native/blob/1465c8f3874cdee8c325ab4a4916fda0b3e43bdb/React/CoreModules/RCTStatusBarManager.mm#L148-L151

Hirbod:
@brentvatne @a-eid I have the same issue:
I am using an expo-dev-client (0.8.0) and my app is staying black after setting UIViewControllerBasedStatusBarAppearance to true. I searched my whole project and made sure no StatusBar imports are left, also removed expo-status-bar from my project. Sure this isn't an expo issue? Maybe a third party package inside node_modules is causing this...

Hirbod:
@brentvatne I think the issue is coming from here:
https://github.com/expo/expo/blob/master/packages/expo-dev-launcher/bundle/views/LauncherApp.tsx#L42

Thats the only existing StatusBar Import left

Hirbod:
@brentvatne I created an issue for it:
https://github.com/expo/expo/issues/15801

PR #644, opened 2020-09-18
fix: update on rotation

WoLewicki:
Schedule update of buttons on the screen rotation so they align correctly.


WoLewicki:
Closing in favor of #668.

PR #648, opened 2020-09-22
feat: add slide animations on Android

jiong-shen-cb:


android/src/main/java/com/swmansion/rnscreens/ScreenStack.java
```diff
@@ -209,21 +184,63 @@ public void run() {
...
+          getOrCreateTransaction().setCustomAnimations(R.anim.close_slide_in, R.anim.close_slide_out);
+          break;
+      }
+
+      if (!customAnimation) {
+        getOrCreateTransaction().setTransition(transition);
+      }
+    }
+
+    // remove all screens previously on stack
```
jiong-shen-cb, line 219: This portion is moved from the top because the animations need to be set first before any screens are added/removed

Hirbod:
Awesome.

Jalson1982:
Hope this will be in master soon. Great job @jiong-shen-cb

Hirbod:
What do you think about this approach @WoLewicki 
It is the XML way but looks nice and clean.

WoLewicki:
If only it works, I think using XMLs will be a good option :tada:.
I think we should prefix them somehow, like `rns_close_slide_in` to avoid any name clashes. Will test this PR in free time and please do so too, especially with nested navigators and `popToTop` options etc. so we can spot the problems quickly if there are any

jiong-shen-cb:
I will update the xml names, btw since I only added support for android (not an ios developer), if necessary I could use some help extend it for ios

itsjgf:
I have opened a PR on top of this one that adds support for `stackAnimation: 'slide'` on iOS & tweaks the modal animation on Android -> #662 662.

WoLewicki:
I think it would also be nice to have the option for `rtl` transition, from what I tested, only thing needed for the new transition is to swap `close` with `open` transition specs. Maybe it would be better to change the naming of these specs to something like `rns_slide_in_from_right`, `rns_slide_out_from_right`, `rns_slide_in_from_left`, `rns_slide_out_from_left`. It would require adding another animation option, probably `SLIDE_RTL` would be the best name. Can you do it @jiong-shen-cb?

jiong-shen-cb:
sure I can add it this week,

jiong-shen-cb:
@WoLewicki I added slide_rtl, please check
@itsjgf please merge my changes into your branch and add slide_rtl in iOS if you could.  Thanks!

WoLewicki:
@jiong-shen-cb could you also add the information about the new animation in:
- https://github.com/software-mansion/react-native-screens/blob/master/README.md#stackanimation
- https://github.com/software-mansion/react-native-screens/blob/master/native-stack/README.md#stackanimation
- https://github.com/software-mansion/react-native-screens/blob/master/createNativeStackNavigator/README.md#stackanimation
- https://github.com/software-mansion/react-native-screens/blob/master/src/index.d.ts#L26
- https://github.com/software-mansion/react-native-screens/blob/master/src/index.d.ts#L95 (you can write that it resolves to iOS's default transition)
- https://github.com/software-mansion/react-native-screens/blob/master/src/native-stack/types.tsx#L247

I think you will need to rebase to the newest master in order to do it, because I recently made changes to the ordering of the props.

jiong-shen-cb:
@WoLewicki Added and rebased to master, please check if they are good enough.

WoLewicki:
Looks perfect! The last (hopefully) change that is required is to add new animation options to iOS in order for it to be able to work correctly with those changes. Please change the method in https://github.com/software-mansion/react-native-screens/blob/master/ios/RNSScreen.m#L502 to:
 
```
RCT_ENUM_CONVERTER(RNSScreenStackAnimation, (@{
                                                  @"default": @(RNSScreenStackAnimationDefault),
                                                  @"none": @(RNSScreenStackAnimationNone),
                                                  @"fade": @(RNSScreenStackAnimationFade),
                                                  @"flip": @(RNSScreenStackAnimationFlip),
                                                  @"slide": @(RNSScreenStackAnimationDefault),
                                                  @"slide_rtl": @(RNSScreenStackAnimationDefault),
                                                  }), RNSScreenStackAnimationDefault, integerValue)
```

It will just add resolving the new options to the default animation.

jiong-shen-cb:
Thanks, updated

WoLewicki:
After consultations with @kmagiera, we decided that it would be very good to have some clips/gifs in this thread that show the transitions, especially going forward and back with using:
1. header back button
2. normal `goBack()` with e.g. a button
3. hw back button (the one at the bottom of the phone's screen)

It looks like now the 3. option is giving different results in animation due to using Android Fragment's back stack.

Another thing is the name of the transitions. `slide_rtl` can be thought of as something that changes the direction of the content, while in this case it is not directly connected to it and can be used with `ltr` and `rtl`. So maybe it would be good to change the names to e.g. `slide_from_right` and `slide_from_left`, or something even more adequate. What do you think?

jiong-shen-cb:
Yes I see the issue and will look into that this week.

jiong-shen-cb:
@WoLewicki Sorry I got delayed from other stuff last week. I think name changes are good (since rtl was actually slide from left). After comparing the build from master and this branch, I see same behavior on the hw back button from the Example app.  Could you provide some clips to show the difference?

WoLewicki:
I tested it again and I cannot reproduce it now, and I am not sure why. Could you update the PR according to the instructions? I also added a `TestsExample` project: https://github.com/software-mansion/react-native-screens/tree/master/TestsExample, (after rebasing to newest master you should be able to use it) could you add a `Test648` file in there which shows the application of the PR? Some basic and maybe nested stack that shows that it works how it should and compiles on both platforms. It will be easier for everyone to test it then.

jiong-shen-cb:
Great, will look at those tomorrow.

jiong-shen-cb:
Updated, please check.

WoLewicki:
Thanks for your contribution! It looks like it is ready to be merged :tada:

jiong-shen-cb:
Thank you! Is there a potential release date for the changes?

WoLewicki:
It depends on some factors, but next week most probably.

hadnet:
I've been waiting for this for a long time. Just to be sure case I'm not up to date with this guys but this will add the possibility to use slide animations (like iOS) on Android, right? When this release will be available?

WoLewicki:
Yes, this PR adds slide animations on transitions on Android. This should be released alongside with #662, and @itsjgf is still working on it, but you can always apply these changes and test them by using [patch-package](https://www.npmjs.com/package/patch-package).

hadnet:
Ok, guys, I tried everything and this slide animation is not working on Android. I installed version 2.16.0/1 2.17.X and no slide animation. These are the packages I'm using along with Expo SDK40. I tested using Android 11 and tested on an older Android 6.x but does not work either. The other options like flip, fade working good. Can anyone help me understand what is wrong here?

```json
    "@react-navigation/bottom-tabs": "^5.11.7",
    "@react-navigation/native": "^5.9.2",
    "@react-navigation/stack": "^5.14.2",
    "react-native-gesture-handler": "^1.9.0",
    "react-native-reanimated": "~1.13.0",
    "react-native-screens": "2.16.1",
```

WoLewicki:
If you use Expo managed workflow, the native code is included in the client and cannot be changed. `react-native-screens` package version in SDK 40 is `2.15.0`, so this commit is not available there yet, which means you cannot use it unfortunately.

PR #650, opened 2020-09-25
feat: orientation managment on iOS

WoLewicki:
Added orientation management on top of #642. Helper methods taken from [expo-screen-orientation](https://docs.expo.io/versions/latest/sdk/screen-orientation/).

### DIFF
--- a/ios/RNSScreen.m
+++ b/ios/RNSScreen.m
@@ -379,9 +379,16 @@
   // if there is no child navigator and the parent is `RNSScreenContainer`, we should fallback to the parent's (that is not `RNSScreenContainer`) option
   UIViewController *parent = [self parentViewController];
   while ([parent isKindOfClass:[RNScreensViewController class]] || [parent isKindOfClass:[RNSScreen class]]) {
-    parent = [parent parentViewController];
+    if ([parent parentViewController] == nil && [parent isKindOfClass:[RNSScreen class]]) {
+      // we are at the top of hierarchy and the controller is RNSScreen so we are in modal screen
+      break;
+    } else {
+      parent = [parent parentViewController];
+    }
   }
-  RNSScreenView *screenView = [parent isKindOfClass:[RNScreensNavigationController class]] ? ((RNSScreenView *)[[[parent childViewControllers] lastObject] view]) : ((RNSScreenView *)self.view);
+  RNSScreenView *screenView = [parent isKindOfClass:[RNScreensNavigationController class]]
+    ? ((RNSScreenView *)[[[parent childViewControllers] lastObject] view])
+    : [parent isKindOfClass:[RNSScreen class]] ? ((RNSScreenView *)parent.view) : ((RNSScreenView *)self.view);
   return [screenView statusBarStyleForRNSStatusBarStyle];
 }
 
@@ -394,9 +401,17 @@
   // if there is no child navigator and the parent is `RNSScreenContainer`, we should fallback to the parent's (that is not `RNSScreenContainer`) option
   UIViewController *parent = [self parentViewController];
   while ([parent isKindOfClass:[RNScreensViewController class]] || [parent isKindOfClass:[RNSScreen class]]) {
-    parent = [parent parentViewController];
+    if ([parent parentViewController] == nil && [parent isKindOfClass:[RNSScreen class]]) {
+      // we are at the top of hierarchy and the controller is RNSScreen so we are in modal screen
+      break;
+    } else {
+      parent = [parent parentViewController];
+    }
   }
-  return [parent isKindOfClass:[RNScreensNavigationController class]] ? ((RNSScreenView *)[[[parent childViewControllers] lastObject] view]).statusBarAnimation : ((RNSScreenView *)self.view).statusBarAnimation;
+  RNSScreenView *screenView = [parent isKindOfClass:[RNScreensNavigationController class]]
+    ? ((RNSScreenView *)[[[parent childViewControllers] lastObject] view])
+    : [parent isKindOfClass:[RNSScreen class]] ? ((RNSScreenView *)parent.view) : ((RNSScreenView *)self.view);
+  return screenView.statusBarAnimation;
 }
 
 - (BOOL)prefersStatusBarHidden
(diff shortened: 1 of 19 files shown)


src/index.d.ts
```diff
@@ -114,6 +123,18 @@ declare module 'react-native-screens' {
...
+     * @type "default" - resolves to "all" without "portrait_down"
+     * @type "all" – all orientations are permitted
+     * @type "portrait" – portrait orientations are permitted
+     * @type "portrait_up" – right-side portrait orientation is permitted
+     * @type "portrait_down" – upside-down portrait orientation is permitted
+     * @type "landscape" – landscape orientations are permitted
+     * @type "landscape_left" – landscape-left orientation is permitted
+     * @type "landscape_right" – landscape-right orientation is permitted
+     */
+    stackOrientationMask?: StackOrientationMaskTypes;
```
kmagiera, line 137: What was the intention behind this naming? For me it sound inaccurate for two reasons:
1) why does it mention stack? I think it'd be more intuitive if it was a screen property that is relevant to a single screen and not to the whole stack.
2) Why does it say "mask"? Masks are typically used when you are allowed to select multiple options. Instead here we have a list of allowed options and only one can be selected (yes, some options allow for both right and left portrait but it does not matter in my opinion). In order to match the mask behavior we should only have base level options defined and allow them to be mixed (e.g. we have portrait_left and portrait_right and allow both to be used as an array or expressed like a bit mask)

ios/UIViewController+RNScreens.m
```diff
@@ -34,6 +34,15 @@ - (UIStatusBarAnimation)RNSPreferredStatusBarUpdateAnimation
   return [self RNSPreferredStatusBarUpdateAnimation];
 }
 
+- (UIInterfaceOrientationMask)RNSSupportedInterfaceOrientations
```
kmagiera, line 37: typically method names start with lowercase in Objc. Also, the capital leter prefix is typically used for global names (class names, enums, exported functions) that, in Objc cannot be scoped/namespaced (this is the reason why the prefixes are used). When you define a class instance method it is already scoped to that class, so no need to add the prefix

ios/RNSScreen.m
```diff
@@ -402,6 +409,23 @@ - (BOOL)prefersStatusBarHidden
...
 
+- (UIInterfaceOrientationMask)supportedInterfaceOrientations
+{
+  UIViewController *presentedVC = [self presentedViewController];
+  if ([presentedVC isKindOfClass:[RNSScreen class]]) {
+    return presentedVC.supportedInterfaceOrientations;
+  }
+  
+  UIViewController *child = [[self childViewControllers] lastObject];
+  if ([child isKindOfClass:[RNScreensNavigationController class]] || [child isKindOfClass:[RNScreensViewController class]]) {
```
kmagiera, line 420: What is `RNScreensViewController` and `RNScreensNavigationController` ?

https://github.com/software-mansion/react-native-screens/search?q=RNScreensNavigationController

ios/RNSScreen.m
```diff
@@ -402,6 +409,23 @@ - (BOOL)prefersStatusBarHidden
...
+    return child.supportedInterfaceOrientations;
+  }
+  RNSScreenView *screenView = [self findScreenViewForScreenProps];
+  
+  UIInterfaceOrientationMask orientationMask = [RNSScreen interfaceOrientationMaskForStackOrientationMask:screenView.stackOrientationMask];
```
kmagiera, line 425: Wouldn't it make more sense if we kept orientation mask in a format that is compatible with UIKit? We introduced a new type and keep it along the instance only so that we can convert it at some point

ios/RNSScreen.m
```diff
@@ -196,6 +197,12 @@ - (void)setStatusBarHidden:(BOOL)statusBarHidden
   }
 }
 
+- (void)setStackOrientationMask:(RNSScreenOrientationMask)stackOrientationMask
+{
+  _stackOrientationMask = stackOrientationMask;
```
kmagiera, line 202: Commented about it below. What is the point of having our own type for orientation if the only thing we use it for is to convert to UIKit compatible type. This seem to add a lot of boilerplate

ios/RNSScreen.m
```diff
@@ -505,6 +529,81 @@ - (void)notifyFinishTransitioning
   _previousFirstResponder = nil;
 }
 
++ (UIInterfaceOrientationMask)interfaceOrientationMaskForStackOrientationMask:(RNSScreenOrientationMask)stackOrientationMask
```
kmagiera, line 532: One better place where I think this could be placed is RCTConvert interface. You could add RCTCOnvert extension for UIInterfaceOrientationMask, this way you no longer need to have to keep you own type and handle conversion from JS param to your own enum and then again from your enum to UIInterfaceOrientationMask

You can check here for an example usage: https://github.com/facebook/react-native/blob/master/React/Views/RCTConvert%2BCoreLocation.m

In that example there is a conversion defined for CLLocationCoordinate2D type. This way, in view managers and in native modules it is allowed to use CLLocationCoordinate2D type as an argument, framework will handle conversion automatically when that type is used as an argument.

kmagiera (commented):
Left a couple of inline comments

src/index.d.ts
WoLewicki: 1. I tried to follow the props `stackAnimation` and `stackPresentation`, but I am aware that this name was maybe not the best in that case.
2. It is the leftover from using the "mask" options on the native side.

`screenOrientation` would be the best name probably?

ios/UIViewController+RNScreens.m
WoLewicki: I followed the naming method for swizzling from `expo`: https://github.com/expo/expo/blob/3356f86046b0107a228cd9dd9cebe81e5fb39321/packages/@unimodules/react-native-adapter/ios/UMReactNativeAdapter/Services/UMReactFontManager.m#L14-L35 (cc @sjchmiela)

ios/RNSScreen.m
WoLewicki: They are subclasses for `ScreenContainer`'s `UIViewController` and `ScreenStack`'s `UINaviagationController` added in this PR to override methods connected to orientation and status bar managment.

ios/RNSScreen.m
WoLewicki: It is definitely a good idea.

WoLewicki:
Closing in favor of #675.

PR #656, opened 2020-10-05
fix: remove check for background color

WoLewicki:
"#fff", "#ffffff" and `white` colors all have a value of "-1" on the native side, and `transparent` color has a value of "0", so using Integer and checking if it is `null` is the best option now.

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
@@ -242,9 +242,7 @@ public class ScreenStackHeaderConfig extends ViewGroup {
     }
 
     // background
-    if (mBackgroundColor != -1) {
-      mToolbar.setBackgroundColor(mBackgroundColor);
-    }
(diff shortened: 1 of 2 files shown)


andrewbeckman:
> Looks like the "#fff", "#ffffff" and `white` colors all have a value of "-1" on the native side, so the newly added if makes it impossible to use these values for the header's background color.

I tested this and can confirm it fixes the issue.

andrewbeckman:
@WoLewicki I tested this and can confirm it fixes the issue. Thanks!

andrewbeckman (approved):

WoLewicki:
@andreibarabas it looks like the option with Integer for `backgroundColor` is the best one, since `-1` is reserved for `white` color. I am sorry for disagreeing with you on this option. @andrewbeckman can you check if this updated PR does not bring any regression?

kmagiera (approved):

andreibarabas:
@WoLewicki  - no worries. I couldn't remember why I put the exactly that option there, so didn't have a proper argument to support it. 

Thanks for stepping in and solving the final touches

PR #658, opened 2020-10-07
fix: support multiple transparent modal layers on android

henninghall:
When stacking multiple transparent modals on top of each other, the views below the top one are being removed from the screen. This leads to 2 problems:
1. Mulitple transparent modals are not visible. 
2. When navigating back from the modals, the javascript navigation state gets unsynced with the actual visible screens and the app ends up in a weird state.  

This PR fixes those issues, this already works as intended on iOS.


android/src/main/java/com/swmansion/rnscreens/ScreenStack.java
```diff
@@ -171,8 +171,20 @@ protected void performUpdate() {
...
 
-    for (ScreenStackFragment screen : mScreenFragments) {
-      // detach all screens that should not be visible
+    // detach all screens that should not be visible
+    for (int i = 0; i < mScreenFragments.size(); i++) {
+      ScreenStackFragment screen = mScreenFragments.get(i);
+
+      // Stop detaching screens if next screen is a transparent modal. This is needed to be able to
+      // see all screens above and including the last non transparent screen
+      boolean hasNextScreen = i < mScreenFragments.size() - 1;
```
WoLewicki, line 180:
```suggestion
      boolean hasNextScreen = i + 1 < mScreenFragments.size();
```

WoLewicki: I think this way it is more clear that you are checking the next screen.

WoLewicki (commented):
Could you provide a repo on which the changes could be visible and tested that they work OK?

henninghall:
@WoLewicki 
Here is a fork with the example project running with the new code. Tap option "Native stack bindings for RNN" and then "Transparent Modal"
https://github.com/henninghall/react-native-screens/tree/transparent-modal-bug-repro

android/src/main/java/com/swmansion/rnscreens/ScreenStack.java
```diff
@@ -171,8 +171,20 @@ protected void performUpdate() {
...
+    // detach all screens that should not be visible
+    for (int i = 0; i < mScreenFragments.size(); i++) {
+      ScreenStackFragment screen = mScreenFragments.get(i);
+
+      // Stop detaching screens if next screen is a transparent modal. This is needed to be able to
+      // see all screens above and including the last non transparent screen
+      boolean hasNextScreen = i + 1 < mScreenFragments.size();
+      if(hasNextScreen){
+        ScreenStackFragment nextScreen = mScreenFragments.get(i + 1);
+        if(nextScreen.getScreen().getStackPresentation() == Screen.StackPresentation.TRANSPARENT_MODAL){
```
kmagiera, line 183: this seem not to handle situation when you go back from non-transparent to transparent modal. To support the case when you can see through we used the `belowTop` variable. Apparently it only supports the case when you have a single level of transparent layers. In order to support multiple layers (what I understand this PR is addressing) it should modify all the logic that deals with `belowTop` including fragments where we attach `belowTop` (when we should attach possible multiple views) and bring them to front in the right order to maintain the proper z-order when attached.

henninghall:
@kmagiera I updated the PR according to your requested changes. It now handles the situation when navigating back from a non-transparent modal to a transparent one. Would be great if you could have another look 🙂

I also added an option to push transparent modal routes in the example project which will look like this: 

Screenshot 2020-12-29 at 19 43 12

WoLewicki:
Thanks for updating it! Can you provide an example in [TestExample](https://github.com/software-mansion/react-native-screens/tree/master/TestsExample) project instead? It uses the current code from the library and makes testing the changes much easier.

henninghall:
@WoLewicki Yeah, I've already done that!

WoLewicki:
@henninghall you added an example in the `Example` project, not the `TestsExample`. Sorry for the effort, but we try to follow a convention where you make a file `Testx` in the `TestsExample` project, where `x` is the number of issue/PR, so it is easy to find and check the reproduction. Could you paste your example into that one by making a simple `native-stack` with those screens and remove the code changing the `Example` project?

henninghall:
Aah now I see! Of course, I will fix it shortly.

henninghall:
@WoLewicki I updated the TestsExample now and removed the changes in Example project. 

Before this bugfix: 
Screenshot 2021-01-15 at 21 48 44

After this bugfix: 
Screenshot 2021-01-15 at 21 50 16

WoLewicki:
Ok I tested it and it seems to work just fine :tada: One thing that would be great in the test example is to have an option to go back from screens and also to push a non-transparent screen in order to be able to test how these two types behave when both being pushed in the same stack. If you could add it, I think this PR would be complete and ready to be merged.

henninghall:
@WoLewicki Here we go, added options and formatted the file. 

Screenshot 2021-01-22 at 21 38 56

henninghall:
@WoLewicki Any plans to merge this soon? 🙏

marcusstenbeck:
@kmagiera @WoLewicki Waiting on this! :)

android/src/main/java/com/swmansion/rnscreens/ScreenStack.java
```diff
@@ -144,25 +144,14 @@ protected boolean hasScreen(ScreenFragment screenFragment) {
   @Override
   protected void performUpdate() {
 
+    ArrayList<ScreenStackFragment> nonDismissedScreens = getNonDismissedScreens();
```
kmagiera, line 147: looks like an overkill to allocate and compute a whole array just to get the last element, maybe we can move it to a separate method but keep the previous loop here. Esp that we also make yet another call that allocates yet another array a few lines below

android/src/main/java/com/swmansion/rnscreens/ScreenStack.java
```diff
@@ -231,22 +220,27 @@ protected void performUpdate() {
...
-    if (belowTop != null && !belowTop.isAdded()) {
-      final ScreenStackFragment top = newTop;
-      getOrCreateTransaction().add(getId(), belowTop).runOnCommit(new Runnable() {
-        @Override
-        public void run() {
-          top.getScreen().bringToFront();
-        }
-      });
+    // attach screens visible below top
+    for (int i = visibleScreensBelowTop.size() - 1; i >= 0; i--){
```
kmagiera, line 232: Can we keep the previous simple loops in place instead of allocating copies of screens array so many times here? I'm fine with extracing it to a method with a more descriptive name. Note that it is a bit unintuitive why all the arrays are in the stack order while visibleScreensBelowTop array is reverse

android/src/main/java/com/swmansion/rnscreens/ScreenStack.java
```diff
@@ -313,4 +307,47 @@ private void setupBackHandlerIfNeeded(ScreenStackFragment topScreen) {
       mFragmentManager.addOnBackStackChangedListener(mBackStackListener);
     }
   }
+
+  private boolean isTransparent(ScreenStackFragment fragment){
```
kmagiera, line 311: this can be a static method, no?

kmagiera (approved):
This looks good right now. I'd like to get rid of added code that makes unnecessary array allocations. Note that the update method can run pretty frequently and these allocations are rather trivial to avoid so I can't justify keeping them for the sake of simplicity. I'm aproving this as the code looks good and don't want to block it any longer but please address the requested changes before we merge this.

henninghall:
@WoLewicki @kmagiera thanks for your comments, I've removed the unnecessary array allocations and restored the simple loops. Entirely out of curiosity, do you think that the additional array allocation actually could have notable performance implications?

android/src/main/java/com/swmansion/rnscreens/ScreenStack.java
henninghall: Sure!

Issue #577, opened 2020-07-27
Modal bugs out and reveals view behind when dragging down with gesture


mrousavy:
### Description

I'm sorry for the title, but I couldn't find better words to describe this.

When using the native stack navigator to present a screen modally, the screen slides up like a `pageSheet` or `formSheet` on iOS. A users reaction might be sliding it back down, which is native behaviour for action sheets in iOS. The problem is, that the screen moves up and shows what's beneath it. 

The weird this is, I cannot reproduce this issue on my iPhone 11 simulator, but it happens **every** time on my real iPhone 11 Pro device.

### Demo

Here's a demo:

![t](https://user-images.githubusercontent.com/15199031/88532559-1beaad80-d005-11ea-9d23-3b22fca3d041.gif)

### Additional Info

For my `NativeStackNavigationOptions` in the Parent Stack navigator I am using:

```js
stackPresentation: 'modal',
gestureEnabled: false,
```

, and the child stack navigator (the screens inside the modal) have the following `NativeStackNavigationOptions`:

```js
headerShown: true,
headerStyle: {
	backgroundColor: Colors.current.identity,
},
headerTintColor: Colors.current.white,
headerTitleStyle: {
	color: Colors.current.white,
	fontSize: 20,
},
```

### Versions

`package.json`
```json
...
"react": "16.13.1",
"react-native": "0.63.2",
"react-native-screens": "^2.9.0",
...
```

WoLewicki:
Can you provide a repo/snack with minimal configuration needed to reproduce the issue?

mrousavy:
@WoLewicki https://github.com/mrousavy/rnscreens577

mrousavy:
It goes pretty far up. Kinda funny

<img src="https://user-images.githubusercontent.com/15199031/88546500-befaf180-d01c-11ea-8142-d947735082ec.png" width="50%" />

mrousavy:
Also, here's another demo, seems related:

![RPReplay_Final1596031780 mov](https://user-images.githubusercontent.com/15199031/88811092-6eb79700-d1b6-11ea-806c-5b1a0c3fc1e9.gif)

I don't know why it flickers here and doesn't on the other demo, maybe because I'm using `SafeAreaView` here?

WoLewicki:
Yeah, it may have something to do with `SafeAreaView`. Could you try if it can be reproduced without it?

happyfloat:
[...]

PR #660, opened 2020-10-08
Fix: iOS modal page sheet frame should not be that of UINavigationController

Cal-um:
I was having [this issue](https://github.com/software-mansion/react-native-screens/issues/577#issuecomment-665689734) when pulling up an iOS 13 page sheet modal on phone devices. 

Cause of issue: The RNScreen frame was being set to equal the parent UINavigationController. This height is larger than the iOS modal since page sheet modals are not full screen. It corrects itself when `updateBounds` is called. However if you pull up the modal reactSetFrame is called multiple time to match the UINavigationControllers frame and a loop happens where frame is set, then bounds are updated causing a violent jitter to anything pinned to the bottom of the screen.

This fixed it by only calling `[super reactSetFrame:frame]` if the parent is not a NavigationController && the stack presentation is not RNSScreenStackPresentationModal. 

There may be a better way of casing this as it will affect FullScreenModals for < iOS 13 devices. I have not noticed any adverse affects with this.

### DIFF
--- a/ios/RNSScreen.m
+++ b/ios/RNSScreen.m
@@ -37,7 +37,7 @@
 - (void)reactSetFrame:(CGRect)frame
 {
   if (![self.reactViewController.parentViewController
-        isKindOfClass:[UINavigationController class]]) {
+        isKindOfClass:[UINavigationController class]] && self.stackPresentation != RNSScreenStackPresentationModal) {
     [super reactSetFrame:frame];
   }
   // when screen is mounted under UINavigationController it's size is controller


Cal-um:
Looking into the code more I think I may have diagnosed why this is happening wrongly but the fact is that calling  `[super reactSetFrame:frame]` in a iOS 13 modal causes this jittering behaviour.

Cal-um:
Here is a video of the wobble and the example [repo](https://github.com/Cal-um/iOS13-screens-bug)
------------
wobble
------------
This video is with the fix. Notice there is no wobble and the frame stays the same with no drop.
noWobble
------------
@WoLewicki Do you think this is a suitable fix for the [issue](https://github.com/software-mansion/react-native-screens/issues/577)

Issue #528, opened 2020-05-29
Header right elements disappear on rotate


JackMoseley2001:
Hello,

I've noticed that any elements placed inside header right (both using navigation.setOptions and view the Stack.Screen options) show when the stack is initially shown, but then if the device is rotated they disappear until the view is next reloaded.

This is only happening when the stack is created using createNativeStackNavigator and not when using React Navigation's createStackNavigator.

Anyone else encountering this? Any help would be appreciated.

leoz:
Same here.

WoLewicki:
The bug is most probably the same as #432, but we can keep this one open too. And it only happens on iOS right?

JackMoseley2001:
> The bug is most probably the same as #432, but we can keep this one open too. And it only happens on iOS right?

Thank you. I have only been able to test it on iOS so far, will have the chance to test it on an Android simulator today and will update once I have done so.

I saw on the post linked there was mention to expo - if linked, it’s worth noting that this issue I’m encountering is occurring in a bare react native project.

WoLewicki:
You can try the solution mentioned in #322 and see if it fixes the issue in your case.

JackMoseley2001:
Thanks, I will attempt this workaround also

JackMoseley2001:
> You can try the solution mentioned in #322 and see if it fixes the issue in your case.

I've attempted this solution, it works some of the time, but to get somewhat reliable results I needed to increase the delay time, resulting in the buttons appearing after some time. I did continue to find that the button would disappear though - going from portrait to landscape worked, but then it failed to show when going from landscape to portrait again

WoLewicki:
I will close it to keep the discussion about the issue in one place. Please see #432 for more information.

WoLewicki:
Can you check if #644 fixes your issue and do not introduce any new bugs?

JackMoseley2001:
Thank you for taking time to have a look at this problem.

I have installed and tested the pull request regarding this, but am still experiencing that the icon will disappear on rotate. I have tried it by setting the right icon within the screen itself using props, and within the stack navigator screen.

I have since reporting this issue moved away from using React Native for the meantime and therefore hope that anyone within issue #432 may be able to help more.

zibs:
Copy pasting [my comment](https://github.com/software-mansion/react-native-screens/issues/432#issuecomment-708490367) from #432 by request.

[...]

PR #668, opened 2020-10-18
Ensure header elements are visible on iOS

zibs:
Adds ensureHeaderButtonsVisible method in RNScreens.m
References issues #432 and #528.
References and incorporates PR #644 as well

Thanks @WoLewicki!

### DIFF
--- a/ios/RNSScreen.m
+++ b/ios/RNSScreen.m
@@ -420,7 +420,7 @@
 
 - (void)ensureHeaderButtonsVisible
 {
-  double delayInSeconds = 0.1;
+  double delayInSeconds = 0.5;
   dispatch_time_t popTime = dispatch_time(DISPATCH_TIME_NOW, (int64_t)(delayInSeconds * NSEC_PER_SEC));
   dispatch_after(popTime, dispatch_get_main_queue(), ^(void){
     [self.view.reactSubviews enumerateObjectsUsingBlock:^(UIView * _Nonnull view, NSUInteger idx, BOOL * _Nonnull stop) {
--- a/ios/RNSScreen.m
+++ b/ios/RNSScreen.m
@@ -330,10 +330,28 @@
   }
 }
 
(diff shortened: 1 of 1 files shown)


WoLewicki:
Thanks for submitting the PR! The main problem with this method is that it assumes that 0.1-second time is enough for the buttons to align correctly after it. It may be not true for screens with more content in them or slower phones. You can see https://github.com/software-mansion/react-native-screens/issues/528#issuecomment-694829190. Is it available for you to check if this 0.1-second time is always enough for this to work correctly?

zibs:
I can get back to you on this this week. It seems like after testing in the wild that 0.1 wasn't enough time when running CPU heavy transitions.

zibs:
So I see here that I've left it set at `0.5`, which might even be too much. In production we're running it at `0.2` and don't think we've had any issues since. Want me to increase it to `0.2` here? Or just leave it at `0.5`?

Other than that it should be good to go.

WoLewicki:
I am not sure if we will want to merge it for now. It is very much a workaround, and with some randomness included, so I would be very hesitant about doing so. I will ping here when it is clear if we want to do it. And for now, I think you can leave it as it is :tada:

casperstr:
What is the status on this? Can I do anything to help move this forward?

WoLewicki:
@casperstr see my comment above. I don't think there is much more to be done about it. If someone wants to use it, he should patch-package it for now. I think we would need a deterministic solution in order to merge such a change.

casperstr:
> @casperstr see my comment above. I don't think there is much more to be done about it. If someone wants to use it, he should patch-package it for now. I think we would need a deterministic solution in order to merge such a change.

The native stack is currently unusable if header buttons are required. Do we have any path forward to patch this with a deterministic solution?

WoLewicki:
I haven't found any other solution unfortunately, but I am open to contributions regarding this.

abram:
Hi @zibs, thanks for your work here. I ran into this issue today and I was grateful to find this PR and discussion. @WoLewicki, I've submitted PR #902 containing a solution I came up with. Hope it's helpful!

Issue #645, opened 2020-09-19
Issue with largeTitle and BottomTabNavigator without a header


juergengunz:
The first screen in the RootNavigator is a BottomTabNavigator without a header and I want to navigate to a Screen in the RootStackNavigator with largeTitles. If I navigate to the second Screen with headerLargeTitle:true - Initially the title is collapsed. 

Is there a way to force the large title to be not collapsed?

```
<RootStack.Screen options={{headerShown: false}} name="Home" component={TabScreen} />
<RootStack.Screen options={({ route }) => ({headerShown: true,title: route.params.name,headerLargeTitle: true})} name="Second" component={SecondScreen} />
```

WoLewicki:
Can you provide a repo/snack with minimal configuration needed to reproduce the issue?

giaset:
Having the same issue.

`<Tab.Navigator>
    <Tab.Screen name='LobbyStackNavigator' component={LobbyStackNavigator} />
</Tab.Navigator>`

and

```
<Stack.Navigator screenOptions={{ headerLargeTitle: true }}/>
    <Stack.Screen name='Lobby' component={LobbyScreen} />
</Stack.Navigator>
```

ArekChr:
Have the same issue, every large title header is collapsed inside bottom tab on app mounted

WoLewicki:
Does this behavior appear only if using `ScrollView` in the screen with large title?

ArekChr:
@WoLewicki  I have manipulatet props a bit, If I set remove paddingBottom contentContainerStyle from scrollview (which is needed for bottom tab translucent) then this doesn't occur I hope that helps

ArekChr:
@WoLewicki  This error doesn't occur if scrollview is empty and without props, hope that helps

ArekChr:
@WoLewicki this error also occur with contentContainerStyle={{ paddingBottom: 10 }} **but with empty scrollView**

WoLewicki:
It looks like the native iOS implementation of `RCTScrollView` is messing with the large header. It would require more debugging on the native side. Thanks for the information @ArekChr 🎉 .

ArekChr:
Im not pretty sure that it is RCTScrollView, With wix react-native-navigation this issue doesn't occur with the same config

WoLewicki:
That is interesting. I will try to look into it more. Please comment if you find some more clues about it!

alexanderstrom:
[...]

PR #670, opened 2020-10-19
fix: large header sizeToFit

### LINKED ISSUE #649: LargeHeader stays small after pop/goBack/swipe gesture on iOS 14+
Hi,

I just upgraded to Expo SDK 39 and before of that, I had never an issue navigating to a child screen (from LargeHeader to small header). When I go back now, it stays smalls instead of growing back to large.

What has changed and is this something known?

WoLewicki:
Applied https://developer.apple.com/forums/thread/660745 solution suggested by @julian-baumann to make the large header appear on the Screen load. Should resolve #649 and #645.

### DIFF
--- a/ios/RNSScreen.m
+++ b/ios/RNSScreen.m
@@ -334,6 +334,10 @@
 {
   [super viewWillAppear:animated];
 
+  UIViewController *parentVC = [self parentViewController];
+  if ([parentVC isKindOfClass:[UINavigationController class]]) {
+    [((UINavigationController *)parentVC).navigationBar sizeToFit];
+  }
   [((RNSScreenView *)self.view) notifyWillAppear];
 }
 


nickyleach:
Tried this patch but it seems like it's resizing the header back button instead of the title?

giphy-2

WoLewicki:
@nickyleach can you try and use the solution proposed here: https://github.com/software-mansion/react-native-screens/issues/645#issuecomment-712199841?

dylancom:
> @nickyleach can you try and use the solution proposed here: [#645 (comment)](https://github.com/software-mansion/react-native-screens/issues/645#issuecomment-712199841)?

@WoLewicki the proposed solution is working for me.

savelichalex:
@WoLewicki how can we help to merge it? We have to patch it in out code...

WoLewicki:
@savelichalex I am afraid the change proposed here was rather a quick workaround, not a solution to the problem, and so it might not be ever merged in order not to introduce any other bugs. I think it is better to apply https://github.com/software-mansion/react-native-screens/issues/645#issuecomment-712199841 as it seems to consistently solve the problem.

savelichalex:
@WoLewicki the thing is we apply both that workarounds to get proper behaviour. If you look at gif above, you may notice that when it go back to `settings` screens the header is folded, though it shouldn't (you could see proper behaviour in default iOS settings app), and the change from current PR helps us to fix such problem.

Should I open a new issue for that? Or there are other ways how we could solve it? We are ready to contribute.

WoLewicki:
I still don't get why you need this PR. https://github.com/software-mansion/react-native-screens/issues/645#issuecomment-712199841 answers the issue of a folded header and this PR tries to do the same thing exactly.

Hirbod:
@WoLewicki but what about Expo users? We need a native fix shipped with react-native-screens (if possible) to be included in SDK 41. Patching isn't an option for folks like me, because I love the managed experience. (Even though downsides are that I am sometimes handcuffed)

WoLewicki:
This PR is not a solution to the problem, rather a workaround, which can result in other bugs, so we most probably don't want to merge it unfortunately. I believe the real solution is the change of `react-native` implementation. If you need this with Expo managed workflow, can't you use the workarounds using JS code pointed in the connected issue?

Hirbod:
None of them worked for me @WoLewicki

WoLewicki:
I have no easy solution for this then unfortunately. Probably when EAS comes live for everyone, it will be simpler to patch native code for managed workflow.

pouyarezvani:
Hey, @WoLewicki I am having the same issue...
I am using `headerLargeTitle` in all of my bottom navigation tabs, and when I enter any of these pages, the`headerLargeTitle` is always collapsed. Was there a solution for this proposed since the time of the creation of this issue?

WoLewicki:
I am afraid not, only the one from this PR and the ones from the issues.

tboba:
@WoLewicki I think this problem is already solved in screens (I've also double-checked Test1072 and I see that large header is expanded. This is possibly working because of this line - https://github.com/software-mansion/react-native-screens/blob/9fbab8df17bf0ad3c9a05ee074896c44cd545095/ios/RNSScreenStack.mm#L160), PR #202. Do you think we can close this PR? Or does it fix some corner case I can't see right now?

PR #679, opened 2020-10-22
feat: android orientation management

WoLewicki:
## Description

`SC` will stand for `ScreenContainer`. Added orientation management on Android. On every update of `native-stack` or navigators with using `SC` we check if there is orientation change needed. We do it by checking the top `Screen` and, if there is no child `HeaderConfig` (which has the priority) and there is a `HeaderConfig` above the `Screen`, changing the orientation according to it. If there are no child `SC`s in the `Screen` in `onContainerUpdate` even though the `Screen` has a child `SC`, it means that the child has not fired `onAttachedToWindow`, where the registration of child `SC` happens. It is not a problem, because then the correct update of orientation will be fired in `onAttachedToWindow` of the child `Screen` `HeaderConfig`, if it exists.

## Changes

- Added `getHeaderConfig` for unified checking for if `Screen` has a header
- Abstracted `updateChildFragments` method for updating `SC`/`ScreenStack` children on update
- Added `hasChildScreenWithConfig` that parses child `SCs` and checks if any of their `Screens` have `HeaderConfig`

`Test42.js` in `TestsExample` project.

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/src/index.d.ts
- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
@@ -151,8 +151,7 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
     return mScreenFragments.get(index).getScreen();
   }
 
-  @Nullable
-  public Screen getTopScreen() {
+  public @Nullable Screen getTopScreen() {
     for (ScreenFragment screenFragment: mScreenFragments) {
       if (getActivityState(screenFragment) == Screen.ActivityState.ON_TOP) {
         return screenFragment.getScreen();
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
@@ -153,7 +153,6 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
 
   private void setFragmentManager(FragmentManager fm) {
     mFragmentManager = fm;
-    updateIfNeeded();
   }
 
   private void setupFragmentManager() {
@@ -169,6 +168,7 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
       setFragmentManager(screenFragment.getChildFragmentManager());
       mParentScreenFragment = screenFragment;
       mParentScreenFragment.registerChildScreenContainer(this);
+      updateIfNeeded();
       return;
     }
 
@@ -191,6 +191,7 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
               "In order to use RNScreens components your app's activity need to extend ReactFragmentActivity or ReactCompatActivity");
     }
     setFragmentManager(((FragmentActivity) context).getSupportFragmentManager());
+    updateIfNeeded();
   }
 
   protected FragmentTransaction getOrCreateTransaction() {
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
@@ -328,6 +328,7 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
     mFragmentManager.executePendingTransactions();
 
     performUpdate();
+    updateChildFragments();
   }
 
   protected void performUpdate() {
@@ -376,5 +377,12 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
     }
 
     tryCommitTransaction();
+  }
+
+  protected void updateChildFragments() {
+    for (int i = 0, size = mScreenFragments.size(); i < size; i++) {
+      ScreenFragment screenFragment = mScreenFragments.get(i);
+      screenFragment.onContainerUpdate();
+    }
   }
 }
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
@@ -366,6 +366,7 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
 
     Screen topScreen = getTopScreen();
     if (topScreen != null) {
+      // if there is an "onTop" screen it means the transition has ended
       transitioning = false;
     }
 
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
@@ -153,6 +153,7 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
 
   private void setFragmentManager(FragmentManager fm) {
     mFragmentManager = fm;
+    updateIfNeeded();
   }
 
   private void setupFragmentManager() {
@@ -168,7 +169,6 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
       setFragmentManager(screenFragment.getChildFragmentManager());
       mParentScreenFragment = screenFragment;
       mParentScreenFragment.registerChildScreenContainer(this);
-      updateIfNeeded();
       return;
     }
 
@@ -191,7 +191,6 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
               "In order to use RNScreens components your app's activity need to extend ReactFragmentActivity or ReactCompatActivity");
     }
     setFragmentManager(((FragmentActivity) context).getSupportFragmentManager());
-    updateIfNeeded();
   }
 
   protected FragmentTransaction getOrCreateTransaction() {
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
@@ -382,7 +382,9 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
   protected void updateChildFragments() {
     for (int i = 0, size = mScreenFragments.size(); i < size; i++) {
       ScreenFragment screenFragment = mScreenFragments.get(i);
-      screenFragment.onContainerUpdate();
+      if (getActivityState(screenFragment) == Screen.ActivityState.ON_TOP) {
+        screenFragment.onContainerUpdate();
+      }
     }
   }
 }
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
@@ -136,8 +136,8 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
   }
 
   protected void removeAllScreens() {
-    for (int i = 0, size = mScreenFragments.size(); i < size; i++) {
-      mScreenFragments.get(i).getScreen().setContainer(null);
+    for (ScreenFragment screenFragment: mScreenFragments) {
+      screenFragment.getScreen().setContainer(null);
     }
     mScreenFragments.clear();
     markUpdated();
@@ -149,6 +149,16 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
 
   protected Screen getScreenAt(int index) {
     return mScreenFragments.get(index).getScreen();
+  }
+
+  @Nullable
+  public Screen getTopScreen() {
+    for (ScreenFragment screenFragment: mScreenFragments) {
+      if (getActivityState(screenFragment) == Screen.ActivityState.ON_TOP) {
+        return screenFragment.getScreen();
+      }
+    }
+    return null;
   }
 
   private void setFragmentManager(FragmentManager fm) {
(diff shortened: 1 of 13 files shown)


android/src/main/java/com/swmansion/rnscreens/ScreenFragment.java
```diff
@@ -66,6 +66,47 @@ public Screen getScreen() {
...
+            return (ScreenStackHeaderConfig) child;
+          }
+        }
+      }
+      parent = parent.getParent();
+    }
+    return null;
+  }
+
+  protected static boolean checkIfChildWithConfig(View view) {
```
kmagiera, line 94: rename this to make it more descriptive. Do we really want to traverse all view hierarchy? Maybe it should be enought to only traverse fragments hierarchy?

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
```diff
@@ -169,6 +171,12 @@ public void onUpdate() {
       }
     }
 
+    // orientation
+    if (!ScreenFragment.checkIfChildWithConfig(getScreen() != null ? getScreen().getChildAt(1) : null)) {
```
kmagiera, line 175: This look weird. What's the point of calling checkIfChildWithConfig(null)? Does it have side effect? I also commented that the name of this method isn't really very informative. Also its role isn't very clear by looking at that name neither by looking at the way we use it in different places.

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
```diff
@@ -345,6 +353,10 @@ private TextView getTitleTextView() {
     return null;
   }
 
+  public int getScreenOrientation(){
```
kmagiera, line 356:
```suggestion
  public int getScreenOrientation() {
```

kmagiera (approved):
Accepting to unblock but I'm a bit hesitant of the code that does view hierarchy traversal. I don't understand why is it necessary – change events are fired at all levels when new screen appear, no?

Lets maybe discuss that last issue over call. Looks good otherwise.

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.java
```diff
@@ -109,7 +109,8 @@ public void setToolbarTranslucent(boolean translucent) {
     }
   }
 
-  public void onStackUpdate() {
+  @Override
+  public void onContainerUpdate() {
     View child = mScreenView.getChildAt(0);
```
kmagiera, line 114: call super?

android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
```diff
@@ -377,4 +378,13 @@ protected void performUpdate() {
 
     tryCommitTransaction();
   }
+
+  protected void updateChildFragments() {
+    for (int i = 0, size = mScreenFragments.size(); i < size; i++) {
+      ScreenFragment screenFragment = mScreenFragments.get(i);
```
jakub-gonet, line 384: This likely can be replaced with for-each loop.

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.java
```diff
@@ -66,6 +64,63 @@ public Screen getScreen() {
...
+        }
+        if (hasChildScreenWithConfig(topScreen)) {
+          return true;
+        }
+      } else {
+        // in ScreenContainer we check only the screen that is on top
+        for (int i = 0; i < sc.getScreenCount(); i++) {
+          Screen childScreen = sc.getScreenAt(i);
+          if (childScreen.getActivityState() == Screen.ActivityState.ON_TOP) {
+            if (hasChildScreenWithConfig(childScreen)) {
```
jakub-gonet, line 110: Those ifs can be merged

WoLewicki: It may result in worse code readability, but I have no strong opinion on that.

jakub-gonet: I think writing this as a conditions' conjunction will be more readable than introducing another indent.

android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
```diff
@@ -377,4 +378,13 @@ protected void performUpdate() {
 
     tryCommitTransaction();
   }
+
+  protected void updateChildFragments() {
```
kmagiera, line 382: There is nothing that we update in this method. maybe rename to `notifyOnUpdate` or `notifyContainerUpdate` or sth else along these lines

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
```diff
@@ -169,6 +171,12 @@ public void onUpdate() {
       }
     }
 
+    // orientation
+    if (getScreenFragment() == null || !getScreenFragment().hasChildScreenWithConfig(getScreen())) {
+      // we check if there is no child that provides config, since then we shouldn't change orientation here (child with index 0 is this header config)
```
kmagiera, line 176: is this comment outdated? why are we referring to index 0 if no code around here references children nor indices

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.java
```diff
@@ -66,6 +64,63 @@ public Screen getScreen() {
     return mScreenView;
   }
 
+  public void onContainerUpdate() {
+    ScreenStackHeaderConfig config = findHeaderConfig();
+    boolean configInChild = hasChildScreenWithConfig(getScreen());
+    if (!configInChild && config != null && config.getScreenFragment().getActivity() != null) {
+      // if there is no child with config, we look for a parent with config to set the orientation
```
kmagiera, line 71: I'm still a bit confused by this. Why we check if children have config while `findHeaderConfig` searches for the config in current container or in its parents? If it isn't trivial to explain maybe we can discuss over call and figure out best way to describe it here.

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.java
```diff
@@ -66,6 +64,63 @@ public Screen getScreen() {
...
+  }
+
+  protected boolean hasChildScreenWithConfig(Screen screen) {
+    if (screen == null) {
+      return false;
+    }
+    for (ScreenContainer sc : screen.getFragment().getChildScreenContainers()) {
+      if (sc instanceof ScreenStack) {
+        // we check only the top screen of stack for header config
+        Screen topScreen = ((ScreenStack) sc).getTopScreen();
```
kmagiera, line 97: We may already be doing this in othre places but it struct me that we refer to subclasses (stack) from places that operate on a lower level of abstraction (screen) . Maybe we can add some interface method to screen container such that we can avoid testing against the type here.

kmagiera (commented):
Left a few more comments inline. sorry for delaying this but I think we are really close.

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
WoLewicki: Yes, it is outdated, there was getting child by index before.

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.java
WoLewicki: Let's have a structure of `ScreenStack` -> `ScreenContainer` -> `ScreenStack`. We do this check in case the update of the container was triggered after setting the orientation in the child stack, which would result in setting the orientation of the parent `ScreenStack` even though the child `ScreenStack` should have priority. At the same time, we want to have an option to change the orientation on every update of container, because the middle `ScreenContainer` can have children that are not `ScreenStack`s, so after navigating from a child that is a `ScreenStack` to a child that is not, we want the orientation to be taken and applied from the parent `ScreenStack`.

Is this explanation clear enough?

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.java
WoLewicki: I think we don't do it in other places. I can add `getTopScreen` method to `ScreenContainer` that will return the `Screen` with `ActivityState.ON_TOP`.

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.java
WoLewicki: Also, it is probably good to change the code to only look for the parent `Screen` with `config` if there is no child with `config` to avoid traversing the hierarchy when not needed.

android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
```diff
@@ -151,6 +151,16 @@ protected Screen getScreenAt(int index) {
     return mScreenFragments.get(index).getScreen();
   }
 
+  @Nullable
+  public Screen getTopScreen() {
```
kmagiera, line 155: We usually put `@Nullable` next to the returned type, i.e. `public @Nullable String` as "nullable" is a return type modifier rather than a method modifier

android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
```diff
@@ -328,13 +338,13 @@ private final void onUpdate() {
     mFragmentManager.executePendingTransactions();
 
     performUpdate();
+    notifyContainerUpdate();
```
kmagiera, line 341: shouldn't this be called directly from `perfomUpdate`? Are there any other ways `performUpdate` method is called when we don't want `notify` to be triggered? If not why?

android/src/main/java/com/swmansion/rnscreens/ScreenStack.java
```diff
@@ -54,6 +55,8 @@ public void dismiss(ScreenStackFragment screenFragment) {
     markUpdated();
   }
 
+  @Nullable
```
kmagiera, line 58: same comment about `@Nullable`

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.java
```diff
@@ -66,6 +64,53 @@ public Screen getScreen() {
...
+        if (headerConfig != null) {
+          return headerConfig;
+        }
+      }
+      parent = parent.getParent();
+    }
+    return null;
+  }
+
+  protected boolean hasChildScreenWithConfig(Screen screen) {
```
kmagiera, line 92: Maybe it'd be worth keeping track of whether there are children with config upon container updates. It seem like this parameter can change when we add new children or when the "top screen changes". Maybe we could add some bottom-up notification mechanism that'd be triggered when such an even occur that'd allow to keep the up-to-date information whether there are children with config instead of doing this traversal which potentially runs on every level in screen container hierarchy and runs the same code for the same container instances multiple times.

kmagiera (approved):
Accepting this to ublock. Code looks mostly fine, I left a few more comments.

android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
WoLewicki: I think not. I added this here since we call it both in `ScreenContainer` and `ScreenStack` this way and if it is in `perfomUpdate`, then we need to use this method in both classes. I don't have a strong opinion on this.

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.java
```diff
@@ -66,6 +65,68 @@ public Screen getScreen() {
...
+      }
+    }
+    return false;
+  }
+
+  protected void markParentFragments(boolean hasChildWithConfig) {
+    ViewParent parent = getScreen().getContainer();
+    while (parent != null) {
+      if (parent instanceof Screen) {
+        ((Screen) parent).getFragment().mHasChildWithConfig = hasChildWithConfig;
```
kmagiera, line 118: If this does not result in `mHasChildWithConfig` change there is no point in updating other parents up the view hierarchy (there will be no updates as well) – we can break the loop if this occurs

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.java
```diff
@@ -66,6 +65,68 @@ public Screen getScreen() {
...
+        return true;
+      }
+      if (topScreen.getFragment().mHasChildWithConfig) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  protected void markParentFragments(boolean hasChildWithConfig) {
```
kmagiera, line 114: I don't think this is correct. If there is one childrent with config and one without, then when the second one updates, it will send information to the parent that there is no config (but the first childrent still has config).

I think that instead this method should actually check flags with all its children (not recursively) and then if the new value for the flag is different than the previous one also notify its parent

Issue #678, opened 2020-10-22
BottomTabNavigator hide/show bug in Native Stack


VanDu97:
I had a problem with BottomTabNavigation have some Stack in there.
 "@react-navigation/stack": "^5.9.3"
"react-native-screens": "^2.11.0",

WoLewicki:
Can you provide a repo/snack with minimal configuration needed to reproduce the issue?

VanDu97:
The same issues was https://github.com/software-mansion/react-native-screens/issues/364

WoLewicki:
So if it is a duplicate of #364, can I close it then? Unfortunately, I cannot help you more than what is mentioned there already.

VanDu97:
but not solved. Please fix this!

WoLewicki:
Can you check if #680 fixes your issue? It seems to fix the problem for the code given in #364.

VanDu97:
> Can you check if #680 fixes your issue? It seems to fix the problem for the code given in #364.

Thank you for your help. It's fixed my problem.

WoLewicki:
#680 was closed in favor of #712. It fixes the issue of #364 from what I tested and seems not to introduce new, wrong behavior.

PR #680, opened 2020-10-22
feat: change condition

WoLewicki:
Change of condition suggested by @kmagiera.

### DIFF
--- a/ios/RNSScreen.m
+++ b/ios/RNSScreen.m
@@ -382,7 +382,7 @@
 {
   [super viewDidLayoutSubviews];
 
-  if (!CGRectEqualToRect(_lastViewFrame, self.view.frame)) {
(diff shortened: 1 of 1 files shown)


Cal-um:
@WoLewicki Here is the debugger with the frame extending below the view
Screenshot 2020-10-27 at 09 44 54

This is the same screen presented with react-native modal which does not extend below the screen.

Screenshot 2020-10-27 at 09 44 23

As per the height measured the react native modal has a smaller height. Any idea why we are getting a larger height in the react native screens presentation?

WoLewicki:
Probably because this modal is a `RNSScreen` and we decide the height of its frame. I cannot say much more at this moment.

artyorsh:
Coming from https://github.com/software-mansion/react-native-screens/issues/678, can confirm that works properly in scope of that issue

WoLewicki:
Closing in favor of #712.

PR #681, opened 2020-10-24
fix: fix types for native stack to match latest react navigation

satya164:
The PR updates the type annotations in native stack to match the following changes in React Navigation:

https://github.com/react-navigation/react-navigation/commit/7dc2f5832e371473f3263c01ab39824eb9e2057d
https://github.com/react-navigation/react-navigation/commit/f51086edea42f2382dac8c6914aac8574132114b

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,8 +52,8 @@
   "devDependencies": {
     "@babel/core": "^7.11.0",
     "@react-native-community/bob": "^0.15.1",
-    "@react-navigation/native": "^5.7.2",
-    "@react-navigation/stack": "^5.8.0",
+    "@react-navigation/native": "^5.8.0",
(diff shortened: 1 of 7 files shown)


WoLewicki:
Shouldn't `<ParamListBase>` be added here also: https://github.com/software-mansion/react-native-screens/blob/master/src/native-stack/views/NativeStackView.tsx#L43 ?

package.json
```diff
@@ -52,8 +52,8 @@
   "devDependencies": {
     "@babel/core": "^7.11.0",
     "@react-native-community/bob": "^0.15.1",
-    "@react-navigation/native": "^5.7.2",
```
WoLewicki, line 55: Could you update the packages in Example app too so they match each other?

PR #685, opened 2020-10-28
feat: add first look on GH actions

WoLewicki:
Added first look on GH actions.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,53 @@
+## Description
+
+<!--
+Description and motivation for this PR.
+
+Include Fixes #<number> if this is fixing some issue.
+
+Fixes # .
+-->
+
+## Changes
+
+<!--
+Please describe things you've changed here, make a **high level** overview, if change is simple you can omit this section.
+
+For example:
+
+- Updated `about.md` docs
+
+-->
+
+<!--
+
+## Screenshots / GIFs
+
+Here you can add screenshots / GIFs documenting your change.
+
+You can add before / after section if you're changing some behavior.
+
+### Before
+
+### After
+
+-->
+
+## Test code and steps to reproduce
+
+<!--
+Please include code that can be used to test this change and short description how this example should work.
+This snippet should be as minimal as possible and ready to be pasted into editor (don't exclude exports or remove "not important" parts of reproduction example)
+-->
+
+## Checklist
+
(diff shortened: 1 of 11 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,18 @@
+name: Build Android and iOS
+
+on:
+  push:
+    branches:
+      - master
```
jakub-gonet, line 6: You probably want to lint it on PRs too.

package.json
```diff
@@ -6,10 +6,10 @@
     "check-types": "tsc --noEmit",
     "start": "node node_modules/react-native/local-cli/cli.js start",
     "test": "npm run format && npm run lint && npm run test:unit",
-    "test:unit": "jest",
+    "test:unit": "jest --passWithNoTests",
     "format": "prettier --write --list-different './src/**/*.{js,ts,tsx}'",
     "lint": "eslint --ext '.js,.ts,.tsx' --fix src && npm run check-types",
-    "precommit": "lint-staged",
+    "precommit": "npm run test",
```
jakub-gonet, line 12: I think this should be replaced by `yarn run test`, you should also change other commands to use yarn (except `npm login`)

src/createNativeStackNavigator.js
```diff
@@ -14,7 +14,9 @@ import {
   SceneView,
   StackActions,
   StackRouter,
+  // eslint-disable-next-line import/no-unresolved
```
jakub-gonet, line 17: Why did you ignore this line? If you use react-navigation why you don't depend on it in package.json?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,53 @@
...
+## Checklist
+
+- [ ] Included code example that can be used to test this change
+- [ ] Updated TS types
+- Updated documentation: <!-- For adding new props to native-stack -->
+- [ ] https://github.com/software-mansion/react-native-screens/blob/master/README.md
+- [ ] https://github.com/software-mansion/react-native-screens/blob/master/native-stack/README.md
+- [ ] https://github.com/software-mansion/react-native-screens/blob/master/createNativeStackNavigator/README.md
+- [ ] https://github.com/software-mansion/react-native-screens/blob/master/src/index.d.ts
+- [ ] https://github.com/software-mansion/react-native-screens/blob/master/src/native-stack/types.tsx
```
jakub-gonet, line 52: I think you can indent it to group those links

PR #689, opened 2020-10-30
Fix system font weight for large title

glennreyes:
The font weight of the large title didn't match exactly to other common large titles in ios preinstalled apps.

#### Before

image

#### After

image

#### Health app (for reference)

image


WoLewicki:
Thanks for your contribution!

PR #709, opened 2020-11-13
feat: export new prop for web

### LINKED ISSUE #701: src/index.js is missing a shouldUseActivityState export
## Description

As of 2.14 the `src/index.native.js` file exports a `shouldUseActivityState` value, while the `src/index.js` file does not. This triggers a warning when using `react-navigation` in conjunction with `react-native-web`. 

The fallback `index.js` should export some sort of value for this variable to align the APIs.

### Expected behavior

Expect no `react-native-screens` related warnings when running using above environment.

### Actual behavior

Receive build-time warning.

## Package versions

- React Native Screens: 2.14

WoLewicki:
## Description

Export `shouldUseActivityState` in web. Fixes #701 .

## Changes

Added `shouldUseActivityState` to the exported constants in `index.js` with `true` value, which can be useful when the support for `active` prop is dropped in the `react-navigation`.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -15,10 +15,7 @@ export class NativeScreen extends React.Component {
   render() {
     let { active, activityState, style, enabled = true, ...rest } = this.props;
     if (active !== undefined && activityState === undefined) {
-      console.warn(
-        'It appears that you are using old version of react-navigation library. Please update @react-navigation/bottom-tabs, @react-navigation/stack and @react-navigation/drawer to version 5.10.0 or above to take full advantage of new functionality added to react-native-screens'
-      );
-      activityState = active !== 0 ? 2 : 0; // in the new version, we need one of the screens to have value of 2 after the transition
+      activityState = active !== 0 ? 2 : 0; // change taken from index.native.js
     }
     return (
       <View
--- a/src/index.js
+++ b/src/index.js
@@ -13,13 +13,20 @@ export function screensEnabled() {
 
 export class NativeScreen extends React.Component {
   render() {
-    const { active, style, enabled = true, ...rest } = this.props;
-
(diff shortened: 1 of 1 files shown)


satya164 (approved):

src/index.js
```diff
@@ -13,13 +13,20 @@ export function screensEnabled() {
 
 export class NativeScreen extends React.Component {
   render() {
-    const { active, style, enabled = true, ...rest } = this.props;
-
+    let { active, activityState, style, enabled = true, ...rest } = this.props;
+    if (active !== undefined && activityState === undefined) {
+      console.warn(
+        'It appears that you are using old version of react-navigation library. Please update @react-navigation/bottom-tabs, @react-navigation/stack and @react-navigation/drawer to version 5.10.0 or above to take full advantage of new functionality added to react-native-screens'
```
satya164, line 19: I'm not sure about this warning. This doesn't necessarily mean that it's used by React Navigation. This warning will also appear in 4.x versions of React Navigation libs which won't be easy to upgrade.

WoLewicki: I added the same warning in `index.native.js`. I think we would like to inform users that upgrading `react-native-screens` without the update of `react-navigation` can result in some bugs. How would you see it?

satya164: @WoLewicki I thought use of old API should stay the same, no? btw we'll drop use of old API in v6 entirely.

WoLewicki: On the native side, the old API was removed, the new one can be not entirely compatible.

satya164: Hmm, maybe consider bumping the version to major if the change isn't backward compatible.

PR #714, opened 2020-11-17
fix: check if activityState is nil

### LINKED ISSUE #702: Push animation between screens not working
This is the version that was working for me:

```
    "@react-navigation/bottom-tabs": "^5.10.2",
    "@react-navigation/native": "^5.8.2",
    "@react-navigation/stack": "^5.11.1",
    "react": "17.0.1",
    "react-native": "0.63.3",
    "react-native-screens": "^2.12.0",
```

I recently updated to this version and on iOS, navigation no longer has any animation, it happens instantly:

```
    "@react-navigation/bottom-tabs": "^5.11.1",
    "@react-navigation/native": "^5.8.9",
    "@react-navigation/stack": "^5.12.6",
    "react": "17.0.1",
    "react-native": "0.63.3",
    "react-native-screens": "^2.14.0",
```

I have deleted Podfile.lock and `npx pod-install ios`, still not work.

When I changed `createNativeStackNavigator()` to `createStackNavigator()`, it worked fine again with animation.

Please let me know if it's a bug or what config I need to use to have the slide (push) animation?

WoLewicki:
## Description

Changed the logic of resolving `activityState` to take into account the initial `nil` value of it by passing the `NSNumber` instead of `int` (which was resolved to `0`) on iOS. The logic for restoring user interaction in `ScreenContainer` had to be changed too since it did not take into account going from value `1` to `2` for one of `Screen`s. It worked before because, at the end of the transition, the `activityState` of all `Screen`s was changed to `0` due to the initial `nil` pass, and then switched back to the proper value in the next evaluation, which resulted in `screenAdded` being set to `YES`. A similar situation, but without the need for change in `ScreenContainer` applies to Android. Fixes #702.

## Changes

Added check for `nil` value of `activityState`. Added new prop `activityStateOrNil`. Changed logic in `ScreenContainer`'s `updateContainer`.

`Test702.js` file with reproduction in `TestsExample`.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.java
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.java
@@ -27,7 +27,14 @@ public class ScreenViewManager extends ViewGroupManager<Screen> {
   }
 
   @ReactProp(name = "activityState")
-  public void setActivityState(Screen view, int activityState) {
+  public void setActivityState(Screen view, Integer activityState) {
+    if (activityState == null) {
+      // Null will be provided when activityState is set as an animated value and we change
+      // it from JS to be a plain value (non animated).
+      // In this scenario we first trigger the code which resets it to the default, which is null.
+      // In that case we want to ignore such update because soon after we get the actual, valid state passed from JS.
+      return;
+    }
     if (activityState == 0) {
       view.setActivityState(Screen.ActivityState.INACTIVE);
     } else if (activityState == 1) {
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.java
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.java
@@ -31,8 +31,8 @@ public class ScreenViewManager extends ViewGroupManager<Screen> {
     if (activityState == null) {
       // Null will be provided when activityState is set as an animated value and we change
       // it from JS to be a plain value (non animated).
-      // In this scenario we first trigger the code which resets it to the default, which is null.
-      // In that case we want to ignore such update because soon after we get the actual, valid state passed from JS.
+      // In case when null is received, we want to ignore such value and not make
+      // any updates as the actual non-null value will follow immediately.
       return;
     }
     if (activityState == 0) {
--- a/ios/RNSScreen.m
+++ b/ios/RNSScreen.m
@@ -58,9 +58,10 @@
   [_bridge.uiManager setSize:self.bounds.size forView:self];
 }
 
-- (void)setActivityState:(int)activityState
+- (void)setActivityStateOrNil:(NSNumber *)activityStateOrNil
 {
-  if (activityState != _activityState) {
+  int activityState = [activityStateOrNil intValue];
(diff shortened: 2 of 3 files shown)


dylancom:
@WoLewicki it brings back the animation, but it seems that my modal screens lose interactivity. Once I open a modal, I can't press any button to go out of it anymore.

WoLewicki:
@dylancom can you provide a reproduction that shows this?

dylancom:
@WoLewicki https://github.com/dylancom/react-native-screens-animation-bug
The same branch as I provided already, contains the modal setup. So if you open the modal with the "Open Modal" button.

WoLewicki:
I do not have such an issue. Did you apply changes from both files? I test in in https://github.com/software-mansion/react-native-screens/blob/master/TestsExample/src/Test702.js with changes applied.

dylancom:
@WoLewicki Oops sorry, thought the 2nd file only had spacing changes. Seems to work now!

ios/RNSScreen.m
```diff
@@ -58,9 +58,10 @@ - (void)updateBounds
   [_bridge.uiManager setSize:self.bounds.size forView:self];
 }
 
-- (void)setActivityState:(int)activityState
+- (void)setActivityStateOrNil:(NSNumber *)activityStateOrNil
```
kmagiera, line 61: Would be great to document it somewhere in the code the purpose of special handling nil value and also explain when nil can be provided. That is we should mention that `nil` will be provided when `activityState` it is set as an animated value and we change it from JS to be a plain value (non animated). I understand that in this scenario we first trigger the code which resets it to the default, which is nil, and which maps to 0 meaning Inactive. In that case we want to ignore such update as soon after we get the actual state passed from JS.

kmagiera (approved):
Code looks good. However, I'd like to place some additional comments explaining why we have this unusual handling added. Accepting to unblock but please explain the meaning of nil and circumstances when it may occur in the code that handles it prior to merging.

kmagiera:
Also the PR title does not look too descriptive. Can you update the title to better reflect what we are changing here?

android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.java
```diff
@@ -27,7 +27,14 @@ protected Screen createViewInstance(ThemedReactContext reactContext) {
   }
 
   @ReactProp(name = "activityState")
-  public void setActivityState(Screen view, int activityState) {
+  public void setActivityState(Screen view, Integer activityState) {
+    if (activityState == null) {
+      // Null will be provided when activityState is set as an animated value and we change
+      // it from JS to be a plain value (non animated).
+      // In this scenario we first trigger the code which resets it to the default, which is null.
+      // In that case we want to ignore such update because soon after we get the actual, valid state passed from JS.
```
kmagiera, line 35:
```suggestion
      // In case when null is received we want to ignore such value and not make 
      // any updates as the actual non-null value will follow immediately.
```

kmagiera (approved):

PR #716, opened 2020-11-18
Multiple log box showing

Cal-um:
## Description

This fixes an incorrectly used prop to the `DebugContainer`. This was causing an appContainer to be rendered on every stackPresentation type causing a logbox to be rendered on every screen within a stack. So 1 on that main stack from RN and one additional one on every screen.

## Changes

Change stackAnimation to stackPresentation on NativeStackView.ts props.

No evidence given as it is clearly visible via the code that this was being passed incorrectly. Logbox shown on modals.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/native-stack/views/NativeStackView.tsx
+++ b/src/native-stack/views/NativeStackView.tsx
@@ -12,6 +12,7 @@ import {
   Screen as ScreenComponent,
   ScreenProps,
   ScreenStack,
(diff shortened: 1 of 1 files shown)


WoLewicki (approved):
Thanks for your contribution! It was clearly a mistake, thank you for spotting it.

PR #722, opened 2020-11-24
fix: make back button visible

### LINKED ISSUE #654: react-native-screens RTL  back arrow icon is not visible in ios
the problem only exist with the release build in real device

WoLewicki:
## Description

Added traversing through the subviews of `navigationBar` and changing their direction. Also, added the second call for updating the header. Those changes combined resolve in the back button being visible in `rtl` direction. Fixes #654.

## Changes

- Updated `Screen.m` to update the header at the beginning of the transition
- Added method for traversing through subviews of `navigationBar` and applying the direction to them. It is needed to change the `direction` of some private views that seem to be not updated along with `navigationBar` update

`Test654.js` in the `TestsExample` project.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.m
+++ b/ios/RNSScreen.m
@@ -425,9 +425,28 @@
   }
 }
 
+- (void)viewWillTransitionToSize:(CGSize)size withTransitionCoordinator:(id<UIViewControllerTransitionCoordinator>)coordinator
+{
+  [self relayoutHeader];
+}
+
+- (void)relayoutHeader
+{
+  double delayInSeconds = 0.1;
+  dispatch_time_t popTime = dispatch_time(DISPATCH_TIME_NOW, (int64_t)(delayInSeconds * NSEC_PER_SEC));
+  dispatch_after(popTime, dispatch_get_main_queue(), ^(void){
+    [self.view.reactSubviews enumerateObjectsUsingBlock:^(UIView * _Nonnull view, NSUInteger idx, BOOL * _Nonnull stop) {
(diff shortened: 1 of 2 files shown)


Hirbod:
Do you think this could land soon, maybe before SDK 40 is out? If there could be a last minute release and maybe a little trigger into expo, they might add it into SDK 40 before the beta is over

WoLewicki:
If the commit is not available in Expo's SDK 40 beta, it most probably won't be available in SDK 40 unfortunately, unless it is very crucial bug fix. This PR is rather a workaround and I am not sure about if it should be merged.

ajpaulingalls:
I have verified that this works, would love to get it into the next release so I don't need to keep my patch-package...

ajpaulingalls:
Looks like this missed the 2.17 release.  Any chance it can make it into the next one so I don't have to update my patch package again?  RTL is a critical requirement for me...

WoLewicki:
As I said in the previous comment, this PR is a workaround and unfortunately has a non-deterministic behavior (`0.1` second here: https://github.com/software-mansion/react-native-screens/pull/722/files#diff-e988f8ab84e7063d6cc9ad4e634806389882ce6727b5c95f3cfbc9f9d3921a67R435) depending on the speed of the phone etc. I don't think it should be merged as long as we don't find a deterministic solution to the problem.

WoLewicki:
Closing since #654 seems to be fixed.

Issue #686, opened 2020-10-29
tvOS RNScreens build errors


dktrvamp:
![Screen Shot 2020-10-28 at 6 04 25 pm](https://user-images.githubusercontent.com/8250990/97605289-13bf5700-19d4-11eb-832d-615fb41eeec1.png)
React Native For TVOS:
"react-native": "npm:react-native-tvos@0.63.1-1",

Breaks tvOS
"react-native-screens": "^2.12.0",

downgraded to and it works fine
"react-native-screens": "2.11.0",

WoLewicki:
Can you check if applying #675 resolves all the issues with tvOS? I tried to add the checks in all places there.

dktrvamp:
@WoLewicki I am a bit confused. #675 is part of the `2.11.0` version. This bug I reported was introduced in `2.12.0` 
Here is what I did.
I installed the following dependency:  `"react-native-screens": "software-mansion/react-native-screens#675/head", `
however my `yarn.lock` shows
```
react-native-screens@software-mansion/react-native-screens#675/head:
  version "2.11.0"
  resolved "https://codeload.github.com/software-mansion/react-native-screens/tar.gz/55f3ea54b1f7d988c7ad62512b15021e3e10c9ea"
```

Please Advise!

WoLewicki:
#675 is not merged so it is not available in 2.11.0. Did you copy all the changes from the file on the screenshot? I added checks there: https://github.com/software-mansion/react-native-screens/pull/675/files#diff-c58852b1c4aee1faddadefe30184a27d6598427b60dcf335c889df3e03d55424R8 which I cannot see on your screenshot.

WoLewicki:
The bugs were introduced in `2.12.0` because the other PR (#642), which is first available in `2.12.0`, introduced them.

WoLewicki:
@dktrvamp can you check if #724 fixes all the problems with the compilation?

csantos1113:
> @dktrvamp can you check if #724 fixes all the problems with the compilation?

Hi @WoLewicki - I tried this in my project:
```
"react-native-screens": "software-mansion/react-native-screens#724/head",
```

and now got the following errors:
<img width="1272" alt="Screen Shot 2020-12-15 at 6 37 05 PM" src="https://user-images.githubusercontent.com/9648559/102285642-b12f0580-3f04-11eb-8d1b-73fda47a0778.png">

PR #675, opened 2020-10-21
feat: add screen orientation on iOS

WoLewicki:
## Description

Added orientation management on iOS.

## Changes

Changed implementation details of `findChildVCForConfig` to always return the exact VC that should be used.
Added checks for `OS_TV`, where these APIs are unavailable.

## Test code and steps to reproduce

`Test42.js` in `TestsExample` project.

## Checklist

- [x] Included code example that can be used to test this change
- [x] Updated TS types
- [x] Updated documentation: 
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/README.md
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/native-stack/README.md
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/createNativeStackNavigator/README.md
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/src/index.d.ts
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/src/native-stack/types.tsx
- [x] Ensured that CI passes

PR #724, opened 2020-11-25
fix: add checks for tvOS and unify macros

WoLewicki:
## Description

Added checks for `tvOS` to make it compile. Changed all `#if (TARGET_OS_IOS)` to `#if !TARGET_OS_TV` since it is the option used in other repositories. Changed all `#ifdef __IPHONE_13_0` to `#if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && defined(__IPHONE_13_0) && \
    __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_0` for the same reason. Silenced some warnings by adding the missing case to switch, `self->` and `API_AVAILABLE(ios(13.0))`.

Compile the project on `tvOS`.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStackHeaderConfig.m
+++ b/ios/RNSScreenStackHeaderConfig.m
@@ -126,7 +126,8 @@
   [navbar setTintColor:[config.color colorWithAlphaComponent:CGColorGetAlpha(config.color.CGColor) - 0.01]];
   [navbar setTintColor:config.color];
 
-#if defined(__IPHONE_13_0) && TARGET_OS_IOS
+#if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && defined(__IPHONE_13_0) && \
+    __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_0
   if (@available(iOS 13.0, *)) {
     // font customized on the navigation item level, so nothing to do here
   } else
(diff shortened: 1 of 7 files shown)


ios/RNSScreenStackHeaderConfig.m
```diff
@@ -602,7 +617,8 @@ + (NSMutableDictionary *)blurEffectsForIOSVersion
       @"prominent": @(UIBlurEffectStyleProminent),
     }];
   }
-#if defined(__IPHONE_13_0) && TARGET_OS_IOS
+#if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && defined(__IPHONE_13_0) && \
```
csantos1113, line 620: After attempting to use this branch directly as suggested in this comment: https://github.com/software-mansion/react-native-screens/issues/686#issuecomment-733722458 I was still having issues trying to build to tvOS.

I made this change and it seemed to work.

```suggestion
#if !TARGET_OS_TV && defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && defined(__IPHONE_13_0) && \
```

WoLewicki: Thank you for checking. Are there any other places where there is an error?

WoLewicki: Can you check on branch from #675? I added some more checks there because of new features and updated the old ones. If it works OK there, I think we can close this PR in favor of that one.

csantos1113: Hey @WoLewicki - Tested #675 on my project and I did not get any errors. 👏

WoLewicki:
Closing in favor of #675.

PR #739, opened 2020-12-07
feat: remove testing on commit

WoLewicki:
## Description

Removed `yarn test:unit` on commit since there are no tests so it makes the `precommit` longer without a reason.


jakub-gonet (approved):

jakub-gonet (commented):
Remember to fix CI too (remove test task).

PR #772, opened 2021-01-12
feat: add setting frame in notifyWillAppear

### LINKED ISSUE #750: [2.16.1] - Lazy loaded tabs renders stack headers without safe area insets.
## Description
The latest version (2.16.1) and more related to this merge https://github.com/software-mansion/react-native-screens/pull/712, has issues where the header insets are not applied if not using lazy loaded tabs from react navigation `lazy={false}`.

## Steps To Reproduce

1. Navigate to the last tab and then navigate to the next screen within that tab and you will see the header rendered with no top insets applied.

## Snack or minimal code example

```
import React from 'react';
import { View, Button } from 'react-native';
import { enableScreens } from 'react-native-screens';
import { createBottomTabNavigator } from '@react-navigation/bottom-tabs';
import { createNativeStackNavigator } from 'react-native-screens/native-stack';

enableScreens();

const Tab = createBottomTabNavigator();
export default function TabNav() {
  return (
    <NavigationContainer>
      <Tab.Navigator lazy={false}>
        <Tab.Screen name="Tab1" component={Tab1} />
        <Tab.Screen name="Tab2" component={StackNav} />
      </Tab.Navigator>
    </NavigationContainer>
  );
}

function StackNav() {
  return (
    <Stack.Navigator>
      <Stack.Screen name="Screen1" component={Screen1} options={{ headerShown: false }} />
      <Stack.Screen name="Screen2" component={Screen2} />
    </Stack.Navigator>
  );
}

const Tab1 = () => <View style={{ flex: 1, backgroundColor: 'red' }} />;
const Screen1 = ({ navigation }) => (
  <View style={{ flex: 1, backgroundColor: 'green', justifyContent: 'center' }}>
    <Button title="navigate" onPress={() => navigation.navigate('Screen2')} />
  </View>
);
const Screen2 = () => <View style={{ flex: 1 }} />;
```

## Package versions

- React: 16.13.1
- React Native: 0.63.4

WoLewicki:
## Description

Suggested by @kmagiera change adding setting frame in the `viewWillAppear`. It ensures that the `parentViewController` is set when trying to set the frame, and if the parent is not `UINavigationController`, then the frame is properly set. It fixes the issue that appeared after #712 was merged, where there is a check for `parentViewController` not being `nil` introduced. The issue can be spotted and should fix #750.

## Changes

Added setting frame in `notifyWillAppear`.

### Before

image

### After

image

-->

`Test750.js` in `TestsExample` project.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.m
+++ b/ios/RNSScreen.m
@@ -204,7 +204,7 @@
     self.onWillAppear(nil);
   }
   // we do it here too because at this moment the `parentViewController` is already not nil,
-  // so if it is not UINavCtr, the frame will be updated to the correct one.
+  // so if the parent is not UINavCtr, the frame will be updated to the correct one.
   [self reactSetFrame:_reactFrame];
 }
 
--- a/ios/RNSScreen.m
+++ b/ios/RNSScreen.m
@@ -15,6 +15,7 @@
   __weak RCTBridge *_bridge;
   RNSScreen *_controller;
   RCTTouchHandler *_touchHandler;
+  CGRect _reactFrame;
 }
 
 @synthesize controller = _controller;
@@ -36,6 +37,7 @@
 
 - (void)reactSetFrame:(CGRect)frame
 {
+  _reactFrame = frame;
   UIViewController *parentVC = self.reactViewController.parentViewController;
   if (parentVC != nil && ![parentVC isKindOfClass:[UINavigationController class]]) {
     [super reactSetFrame:frame];
@@ -201,6 +203,9 @@
   if (self.onWillAppear) {
     self.onWillAppear(nil);
   }
+  // we do it here too because at this moment the `parentViewController` is already not nil,
+  // so if it is not UINavCtr, the frame will be updated to the correct one.
+  [self reactSetFrame:_reactFrame];
 }
 
 - (void)notifyWillDisappear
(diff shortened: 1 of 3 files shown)


ios/RNSScreen.m
```diff
@@ -201,6 +203,9 @@ - (void)notifyWillAppear
   if (self.onWillAppear) {
     self.onWillAppear(nil);
   }
+  // we do it here too because at this moment the `parentViewController` is already not nil,
+  // so if it is not UINavCtr, the frame will be updated to the correct one.
```
kmagiera, line 207:
```suggestion
  // so if it is not under a UINavCtr, the frame will be updated to the correct one.
```

kmagiera (approved):
In this code we veryfi if a *parent* is an UINavController not the controller itself. I added a note on that in the comment you added in RNScreen.m but please also update the PR description to match that

ios/RNSScreen.m
WoLewicki, line 207:
```suggestion
  // so if the parent is not UINavCtr, the frame will be updated to the correct one.
```

mahi8813:
Hi, Can we have a release with the fix pls

PR #784, opened 2021-01-22
feat: add cancel touchHandler on swipe in nested stack in modal

### LINKED ISSUE #780: Touchables still respond (and fire) when swiping left on iOS inside a stack that is inside a modal
### Screenshots

Wrong behaviour:

wrong

Expected behavior:

good

## Steps To Reproduce

1. Add a native stack (Modal)
2. Inside the modal add another stack (no matter if native or not)
3. When swiping left on iOS, if while you swipe there is a touchable, touchable activates. If you release the swipe, touchable fires

### Expected behavior

When swiping left, dismiss touchable activation (same as if is not a modal).

### Actual behavior

When swiping left inside a modal, any touchable in the way gets activated.

## Snack or minimal code example

Can be reproduced here: https://github.com/ferrannp/native-stack-playground

## Package versions

- React: `16.13.1`
- React Native: `0.63.4`
- React Native Screens: `2.16.1`

WoLewicki:
## Description

Added replicating the behavior of canceling `touchHandler` on swipe gesture in nested stacks inside a modal.
If the nested stack is not `native-stack`, the swipe still triggers button click event. This bug should be resolved with https://github.com/software-mansion/react-native-gesture-handler/pull/1323.

Should fix #780.

## Changes

Updated `RNSScreen.m` to be able to get `_touchHandler` and added getting top `Screen` of current view hierarchy in a `ScreenStack.m` in order to call `cancel` on its `touchHandler`.

`Test780.js`.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.h
+++ b/ios/RNSScreen.h
@@ -1,6 +1,7 @@
 #import <React/RCTViewManager.h>
 #import <React/RCTView.h>
 #import <React/RCTComponent.h>
+#import <React/RCTTouchHandler.h>
 
 #import "RNSScreenContainer.h"
 
@@ -69,6 +70,7 @@ typedef NS_ENUM(NSInteger, RNSActivityState) {
 @property (nonatomic) RNSScreenReplaceAnimation replaceAnimation;
 
 - (void)notifyFinishTransitioning;
+- (RCTTouchHandler *)reactTouchHandler;
 
 @end
 
--- a/ios/RNSScreen.h
+++ b/ios/RNSScreen.h
@@ -1,7 +1,6 @@
 #import <React/RCTViewManager.h>
 #import <React/RCTView.h>
 #import <React/RCTComponent.h>
-#import <React/RCTTouchHandler.h>
 
 #import "RNSScreenContainer.h"
 
@@ -70,7 +69,6 @@ typedef NS_ENUM(NSInteger, RNSActivityState) {
 @property (nonatomic) RNSScreenReplaceAnimation replaceAnimation;
 
 - (void)notifyFinishTransitioning;
-- (RCTTouchHandler *)reactTouchHandler;
 
 @end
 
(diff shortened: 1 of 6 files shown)


ios/RNSScreen.h
```diff
@@ -69,6 +70,7 @@ typedef NS_ENUM(NSInteger, RNSActivityState) {
 @property (nonatomic) RNSScreenReplaceAnimation replaceAnimation;
 
 - (void)notifyFinishTransitioning;
+- (RCTTouchHandler *)reactTouchHandler;
```
kmagiera, line 73: this should be `touchHandler` and not `reactTouchHandler` I believe

ios/RNSScreenStack.m
```diff
@@ -144,10 +144,13 @@ - (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
   // Without the below code the Touchable will remain active (highlighted) for the duration of back
   // gesture and onPress may fire when we release the finger.
   UIView *parent = _controller.view;
-  while (parent != nil && ![parent isKindOfClass:[RCTRootContentView class]]) parent = parent.superview;
-  RCTRootContentView *rootView = (RCTRootContentView *)parent;
-  [rootView.touchHandler cancel];
-
+  while (parent != nil && ![parent respondsToSelector:@selector(touchHandler)]) parent = parent.superview;
+  if (parent != nil) {
+    RCTTouchHandler *touchHandler = [parent performSelector:@selector(touchHandler)];
```
kmagiera, line 149: You should be able to just do `[parent touchHandler]` here no?

ios/RNSScreen.h
WoLewicki: Do we need it in `RNSScreen.h`?

kmagiera (approved):
Left two comments inline, looks good otherwise

ios/RNSScreenStack.m
WoLewicki: `parent` is a `UIView` so it does not have `touchHandler`.

PR #787, opened 2021-01-25
 chore: fix babel and metro config

satya164:


WoLewicki:
Your changes were already applied with other ones in one of my commits. Thanks for the help!

PR #792, opened 2021-01-27
Update Android Tools gralde version

someonewithpc:
## Description
This was necessary to fix https://github.com/software-mansion/react-native-gesture-handler/issues/494

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -10,7 +10,7 @@ buildscript {
         }
 
         dependencies {
-            classpath("com.android.tools.build:gradle:3.6.2")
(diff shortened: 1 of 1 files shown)


WoLewicki:
I made a PR removing this `buildscript` since it would only be used when `if (project == rootProject)`, but `react-native-screens` is not designed to be used as a standalone project. Can I close this PR then?

someonewithpc:
If that means it uses the same as the root project, then sounds good :)

WoLewicki:
You can check #799 and see if it introduces any unwanted behavior.

PR #798, opened 2021-02-03
feat: change modal logic

### LINKED ISSUE #767: Wrong icon color when push modal on iOS
## Description

When form sheet or page sheet modal is pushed, info icon on status bar should change color to white as shown

### Screenshots

| RNC      | Right |
| ----------- | ----------- |
|ezgif-6-a1f17d5dd8c6|ezgif-6-dfd7927aed0c |

## Package versions

- React: 16.13.1
- React Native: 0.63.3

WoLewicki:
## Description

PR introducing 3 changes. 
- disabled the ability to modify the status bar appearance in non-fullScreen modals since it destroys the animation (see gifs in #767) and introduces the necessity to control the color of status bar in every modal, which does not work well e.g. with phones with iOS <13 since there are only fullScreen modals there. The system seems to pretty well deal with all the cases where this change could be problematic, so I think it is worth to do this.
- extended an `if` clause in `RNSScreenStackHeaderConfig.m` to fix the bug when trying to change options via `navigation.setOptions` while being in `fullScreenModal`. Till now, when the code reached the changed line, if did not go into `if` due to `nav` being `nil` for `fullScreenModal`, which resolved to `nextVC` also being `nil`. We now check if `nav` is `nil`, which should mean that we are in `fullScreenModal`, and if `vc` is not `nil` at the same time, we want to update the screen.
- Removed running `tsc --noEmit` in `TestsExample` project since it made the CI fail for some reason with `.tsx` files.

Fixes #767.

## Changes

Changed `RNSScreen.m` to not include option for changing status bar appearance in non-fullScreen modals.
Changed `RNSScreenStackHeaderConfig.m` to fix the bug with fullScreen modals.
Added `tsconfig.json` in `TestsExample` to be able to test TS code in it.
Removed checking files from folders other than `src` in the main `tsconfig.json`.

`Test642.tsx` in `TestsExample` project.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/TestsExample/ios/Podfile.lock
+++ b/TestsExample/ios/Podfile.lock
@@ -308,7 +308,7 @@ PODS:
     - React
   - RNReanimated (1.13.1):
     - React
-  - RNScreens (2.17.0):
+  - RNScreens (2.17.1):
     - React-Core
   - RNSearchBar (3.5.1):
     - React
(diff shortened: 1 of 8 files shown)


kmagiera (approved):

hirbod:
@WoLewicki for me, this breaks the ability to get correct colors. We have formSheets and `modal` and iOS (16.4) inverts the colors for me. So its turning the time to white on white background and white on a black background. I want to control this but tried like 1 hour until I found this. Is there no way for us to control this now?

WoLewicki:
It was so long ago that I have no idea tbh 😅 Did this problem appear on 16.4? Or was it always like this?

Issue #800, opened 2021-02-04
Allow headers when stack presentation = "modal"


sbalay:
## Description

I KNOW, there are bunch other closed issues with the same topic (#365 #541 #559), please bear with me 😄 

As discussed in #260, at the moment it seems that the accepted solution to show headers inside a modal, is to create a separate navigator inside the modal to get a new header.

Given that `react-navigation` strongly recommends to avoid nesting as much as possible, I find this behaviour quite problematic when trying to keep a flat navigation structure.

### Expected behavior

Show headers in screens where stackPresentation is set to modal, according to the option `headerShown`

### Actual behavior

Headers are never shown

## Package versions

- React: 16.13.1
- React Native: 0.63.2
- React Native Screens: 2.16.1

WoLewicki:
Can you check if #821 is what you need? Please treat it rather as a PoC for now since it must be thoroughly tested.

sbalay:
@WoLewicki I'll try to test this branch locally during this week and get back to you after that.

sbalay:
@WoLewicki I've done some simple tests and the functionality looks correct.

I noticed that when a modal is pushed on top of another modal we do not show the back button in the left back button, even though the screens are in the same stack. I can still swipe down to dismiss the last modal and go back to the initial one though.
I don't have a lot of knowledge around iOS standards, so I'm not sure if this behaviour is ok or if the left back button should be shown.

WoLewicki:
The back button is not shown because under the hood we create a new `native-stack` in every modal screen just for the header, so it is always the first screen of the navigator. I don't think there is a solution for this since it is not a header of the initial stack and so there is no native option for going back with a back button.

PR #821, opened 2021-02-18
feat: add headers to modal

WoLewicki:
## Description

Added option for achieving headers in modal on iOS. The change is based on adding additional `ScreenStack` to every `modal` `Screen` component and adding the `HeaderConfig` component to upper or lower `Screen` based on the `stackPresentation` to ensure proper header transitions. Modal headers will not have the back button on iOS, since the header comes from the nested stack there and has no previous screen. Requested in #800.

## Changes

Changed `NativeStackView.tsx` to contain nested `ScreenStack` with proper `HeaderConfig` component manipulation. Same with v4 `native-stack`.

`Test800.tsx` in `TestsExample`. It will require to test many scenarios in order to make sure it doesn't break anything.

## Checklist

- [x] Ensured that CI passes


TestsExample/src/Test800.tsx
```diff
@@ -0,0 +1,59 @@
...
+}
+
+function Second({
+  navigation,
+}: {
+  navigation: NativeStackNavigationProp<ParamListBase>;
+}) {
+  return (
+    <Button
+      title="Tap me for second screen"
```
kacperkapusciak, line 47:
```suggestion
      title="Tap me for the first screen"
```
Small typo in example

kacperkapusciak (commented):
I've found out that the header background color behaves unpredictably. It seems like it sometimes makes itself transparent:

https://user-images.githubusercontent.com/39658211/108481379-6d57d400-7298-11eb-9290-8a014f030643.mov

https://user-images.githubusercontent.com/39658211/108481346-6335d580-7298-11eb-8296-a1371ba34719.mov


image

src/native-stack/views/NativeStackView.tsx
```diff
@@ -129,13 +183,18 @@ export default function NativeStackView({
...
-            <Container
-              style={viewStyles}
-              // @ts-ignore Wrong props passed to View
-              stackPresentation={stackPresentation}>
-              {renderScene()}
-            </Container>
+            <HeaderConfig
+              {...options}
+              route={route}
+              headerShown={isHeaderInPush}
```
kacperkapusciak, line 189: It looks like `isHeaderInPush` variable is redundant. I think when it's `InModal` it can't be `InPush` at the same time.
```suggestion
              headerShown={!isHeaderInModal}
```

src/createNativeStackNavigator.js
```diff
@@ -234,6 +266,13 @@ class StackView extends React.Component {
       stackAnimation = 'none';
     }
 
+    const hasHeader =
+      this.props.navigationConfig?.headerMode !== 'none' &&
+      options.header !== null;
+
+    const isHeaderInModal = stackPresentation !== 'push' && hasHeader;
+    const isHeaderInPush = stackPresentation === 'push' && hasHeader;
```
kacperkapusciak, line 274: Like in `NativeStackView.tsx` I think `isHeaderInPush` variable is reduntant because when it's `InModal` it can't be `InPush` so just use `!isHeaderInModal` in line `302`.
```suggestion
    const isHeaderInModal = stackPresentation !== 'push' && hasHeader;
    
```

kacperkapusciak (commented):
Left some comments but looks good overall 🙌  Good job! 🎉

src/native-stack/views/NativeStackView.tsx
```diff
@@ -77,6 +126,11 @@ export default function NativeStackView({
           contentStyle,
         ];
 
+        const isHeaderInModal =
+          stackPresentation !== 'push' && headerShown !== false;
+        const isHeaderInPush =
+          stackPresentation === 'push' && headerShown !== false;
```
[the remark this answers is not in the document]
kacperkapusciak: But on the other hand, it does make sense to be shown by default.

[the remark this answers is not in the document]
WoLewicki: We can make it non-default, but now at least everyone will see that there is such an option in modals. They can always turn this off.

src/native-stack/views/NativeStackView.tsx
WoLewicki: But it can also be none of them with `headerShown === false`.

src/createNativeStackNavigator.js
WoLewicki: Same as above.

kacperkapusciak (approved):

src/createNativeStackNavigator.tsx
```diff
@@ -330,6 +331,41 @@ class StackView extends React.Component<Props> {
...
+    route: NavigationRoute<NavigationParams>,
+    navigation: NavigationScreenProp<
+      NavigationRoute<NavigationParams>,
+      NavigationParams
+    >,
+    SceneComponent: React.ComponentType<Record<string, unknown>>,
+    index: number,
+    descriptor: NativeStackDescriptor
+  ) => {
+    if (isHeaderInModal) {
```
satya164, line 346: This may be problematic if this option changes dynamically as it'll unmount the previous component and render new one, and all state will be destroyed.

niklasgrewe:
any news on this pr? I am looking also for this feature 👍

WoLewicki:
Hopefully this will be merged soon. You can always [patch-package](https://www.npmjs.com/package/patch-package) these changes into your project to see how it works and give us feedback.

AzzouDuGhetto:
> I've found out that the header background color behaves unpredictably. It seems like it sometimes makes itself transparent:
> 
>  Modal.header.color.cyan.mov 
>  Modal.header.color.mov 
> image

Same here, otherwise it's very nice to finally see that implemented 🥳

KingAmo:
any update?

src/native-stack/views/NativeStackView.tsx
```diff
@@ -78,6 +130,24 @@ export default function NativeStackView({
...
+        if (
+          !didWarn &&
+          stackPresentation !== 'push' &&
+          headerShown !== undefined
+        ) {
+          didWarn = true;
+          console.warn(
+            'Be aware that changing the visibility of header in modal on iOS will result in resetting the state of the screen.'
+          );
+        }
```
alexandersandberg, line 142: @WoLewicki Would it not make more sense to warn about this whenever a visibility change actually happens?

Now I get a warning every time I run the app and open a modal despite having a static `headerShown: false` that will never cause this issue.

WoLewicki: We would need to keep track of the previous value of this prop in order to do that, and I am not sure if it is a good idea. Also, if you do not want the header to be shown, you can just not change the prop in modal screens at all. You can easily remove the warning by using `LogBox.ignoreLogs`.

alexandersandberg: Yeah, good point.

I see it's triggered due to `screenOptions={{ headerShown: false }}` on the navigator. I know you suggest setting `headerShown` on each screen instead, but why is that suggested exactly?

>You can easily remove the warning by using LogBox.ignoreLogs.

`headerShown: undefined` on the modal screen works as well, as it seems to override the navigator's `headerShown`.

WoLewicki: Yeah, setting props on individual screens override the options set in navigator's `screenOptions`.

xmflsct: I understand, then basically I can ignore the warning if I did not intentionally change `headerShown` prop. Just to make sure I am not doing anything wrong by paying attention to all warnings. :)

PR #830, opened 2021-02-24
fix: iOS 14 back button menu

WoLewicki:
## Description

Fixed broken navigation using a back button menu available on iOS >= 14.

Also, added a `disableBackButtonMenu` prop for disabling the back button menu based on https://github.com/wix/react-native-navigation/pull/6874.

### Before

https://user-images.githubusercontent.com/39658211/119121637-f0baca00-ba2d-11eb-9973-f2049adc685e.mov

### After

https://user-images.githubusercontent.com/39658211/119121664-f9130500-ba2d-11eb-85a1-c8946cf8649b.mov

## Changes

Added logic for dismissing more than one screen on iOS based on `dismissCount` value and handled it on the JS side accordingly.

Added `RNSUIBarButtonItem` to add the behaviour of not setting the `UIMenu` for the back button and added proper prop for handling it.

`Test830.tsx` in `TestsExample` project.

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/src/index.d.ts
- [x] Ensured that CI passes


kacperkapusciak (approved):

ferrannp:
@WoLewicki any plans to release this? 🙏 .

src/createNativeStackNavigator.tsx
```diff
@@ -537,7 +546,7 @@ function createStackNavigator(
 
       if (backRouteIndex > 0) {
         const newRoutes = [...state.routes];
-        newRoutes.splice(backRouteIndex, 1);
+        newRoutes.splice(backRouteIndex, dismissCount > 0 ? dismissCount : 1);
```
kacperkapusciak, line 549: and that's why here even if you pass the correct number of screens to be dismissed it starts at the wrong index (always previous screen)

kacperkapusciak (changes requested):
Great job on making the back button menu working on v5! 🎉  There's a small issue on v4

src/createNativeStackNavigator.tsx
kacperkapusciak: Fixed in 5e9ebd3

kacperkapusciak (approved):
LGTM!

nandorojo:
@WoLewicki Thanks for this PR!

Question – is it possible to use the iOS back button menu with a custom header that I have on my screen?

I'd like to wrap my custom back button here with the menu. Is there an exported component I could do that with? For instance:

```tsx
<BackButtonMenu>
  <MyComponent />
</BackButtonMenu>
```

https://user-images.githubusercontent.com/13172299/121788155-10f72800-cb99-11eb-9681-9972a6ab8a67.mov

WoLewicki:
Native back button item is a part of `UINavigationItem`, which is connected to `UINavigationBar`, so you cannot use that button without having the native `UINavigationBar` as far as I know. One solution I can think of is to make your own native button with `UIMenu` added to it, which elements would trigger the native navigation action when clicked.

bryanltobing:
> Native back button item is a part of UINavigationItem, which is connected to UINavigationBar, so you cannot use that button without having the native UINavigationBar as far as I know. One solution I can think of is to make your own native button with UIMenu added to it, which elements would trigger the native navigation action when clicked.

wonder if this is still not possible?

PR #839, opened 2021-03-01
fix: schedule only one update after transition

### LINKED ISSUE #791: RNScreensNavigationController is pushing the same view controller instance more than once
## Description

We are seeing the following crash quite frequently on iOS since we started to use `react-native-screens/createNativeStackNavigator`. Any help would be much appreciated.

```
Exception Type: EXC_CRASH (SIGABRT)
Crashed Thread: 0

Application Specific Information:
<RNScreensNavigationController: 0x1038e6800> is pushing the same view controller instance (<RNSScreen: 0x12f123e20>) more than once which is not supported and is most likely an error in the application

Thread 0 Crashed:
0   CoreFoundation                  0x3259749d8         __exceptionPreprocess
1   libobjc.A.dylib                 0x34e15ab50         objc_exception_throw
2   UIKitCore                       0x3298bc6c0         -[UINavigationController pushViewController:transition:forceImmediate:]
3   UIKitCore                       0x32a4dbfb4         -[_UIAfterCACommitBlock run]
4   UIKitCore                       0x32a041a64         _runAfterCACommitDeferredBlocks
5   UIKitCore                       0x32a030f94         _cleanUpAfterCAFlushAndRunDeferredBlocks
6   UIKitCore                       0x32a062564         _afterCACommitHandler
7   CoreFoundation                  0x3258f3878         __CFRUNLOOP_IS_CALLING_OUT_TO_AN_OBSERVER_CALLBACK_FUNCTION__
8   CoreFoundation                  0x3258edf4c         __CFRunLoopDoObservers
9   CoreFoundation                  0x3258edbf0         CFRunLoopRunSpecific
10  GraphicsServices                0x3533f9594         GSEventRunModal
11  UIKitCore                       0x32a0323d4         -[UIApplication _run]
12  UIKitCore                       0x32a037954         UIApplicationMain
13  ########                        0x200d3d5d4         main (main.m:7)
```

## Steps To Reproduce
We see this crash in our Sentry issues but don't know how to reproduce it yet.

## Package versions
- React: 16.13.1
- React Native: 0.63.3
- React Navigation: 4.4.3

WoLewicki:
## Description

Added a flag that makes it possible to schedule only one transition for after the current transition is completed. It solves the problem of adding multiple `completion` blocks to after the transition, which all fire at once after the transition, resulting in pushing the same VCs multiple times. Should fix #791.

## Changes

Added check for modals in `RNSScreenStack.m`.
Update `eslintrc.js` in `TestsExample`.

`Test791.tsx` in `TestsExample` project.

## Checklist

- [x] Ensured that CI passes


.eslintrc.js
```diff
@@ -87,6 +87,7 @@ module.exports = {
         'no-unused-vars': 'off',
         'no-array-constructor': 'off',
         'no-use-before-define': 'off',
+        '@typescript-eslint/explicit-module-boundary-types': 'off',
```
kacperkapusciak, line 90: I have mixed feelings about disabling this eslint rule

TestsExample/src/Test791.tsx
```diff
@@ -0,0 +1,60 @@
+import { NavigationContainer, ParamListBase } from '@react-navigation/native';
+import React, {useEffect} from 'react';
+import {StyleSheet, Button, View, Text} from 'react-native';
+import {createNativeStackNavigator, NativeStackNavigationProp} from 'react-native-screens/native-stack';
+
+type Props = {
+  navigation: NativeStackNavigationProp<ParamListBase>;
+}
+
+const MainScreen = ({navigation}: Props) => {
```
kacperkapusciak, line 10:
```suggestion
const MainScreen = ({navigation}: Props): JSX.Element => {
```

TestsExample/src/Test791.tsx
```diff
@@ -0,0 +1,60 @@
...
+  }, [])
+
+  return (
+    <View style={styles.screen}>
+      <Text>Issue 791</Text>
+    </View>
+  )
+}
+
+const PushScreen = ({navigation}: Props) => (
```
kacperkapusciak, line 22:
```suggestion
const PushScreen = ({navigation}: Props): JSX.Element => (
```

TestsExample/src/Test791.tsx
```diff
@@ -0,0 +1,60 @@
...
+  <View style={styles.screen}>
+    <Button onPress={() => navigation.push('Push')} title="Mash this button to push" />
+    <Button onPress={() => navigation.pop()} title="Mash this button to pop" />
+  </View>
+);
+
+const Stack = createNativeStackNavigator();
+
+const App = () => (
```
kacperkapusciak, line 31:
```suggestion
const App = (): JSX.Element => (
```

kacperkapusciak (commented):
I'd wait for @kmagiera to express his opinion but it does solve the issue.

PR #860, opened 2021-03-18
feat: move status bar and orientation management to Screen

### LINKED ISSUE #836: react-native-screens/native-stack BUG ORIENTATION
i use  createNativeStackNavigator  from 'react-native-screens/native-stack';

i have one problem in two cases:
first is when i fixed orientation: dont work.
secont when i use wonday/react-native-orientation-locker has the same problem orientation dont fixed.
if i donte use react native screens the orientation fixed was ok.
i force fix orientation with calling react native locker  every page with timeout 100ms and ok but, i would like no need use this pratice.

WoLewicki:
## Description

Moved the props used for status bar and orientation management from the `HeaderConfig` to `Screen` component in order to make it available for them to be used by all navigators, not only `native-stack`. It should also fix #836 due to check if the orientation has been set on Android.

## Changes

Moved the logic from `HeaderConfig` into `Screen` both on iOS and Android.
On iOS, there has been added new class with static methods used when managing those props.
On Android, for `native-stack` (where the `HeaderConfig` lives), the logic of setting the proper `screenOrientation` was left in `HeaderConfig` because it is required to have the `Fragment` and `Activity` ready to set it, and it is not available when setting the prop.

`Test42.tsx` for orientation and `Test642.tsx` for status bar management in `TestsExample` project.

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [x] Ensured that CI passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,21 @@ Allows for the customization of the type of animation to use when this screen re
 - `push` – performs push animation
 - `pop` – performs pop animation (default)
 
+#### `screenOrientation`
+
+Sets the current screen's available orientations and forces rotation if current orientation is not included. Possible values:
+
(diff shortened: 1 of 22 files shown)


WoLewicki:
Closed in favor of #863.

PR #869, opened 2021-03-25
feat: enable screens by default

WoLewicki:
## Description

Enabled native components of `react-native-screens` by default. It results in no need of calling `enableScreens()` at the top of your app to use them 🎉 

## Changes

Changed `index.native.tsx`.

Do not call `enableScreens()` at the top of your app and spot that native components are used.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/index.native.tsx
+++ b/src/index.native.tsx
@@ -23,7 +23,7 @@ import {
   ScreenStackHeaderConfigProps,
 } from './types';
 
-let ENABLE_SCREENS = false;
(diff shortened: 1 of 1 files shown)


yanush:
Is it still possible to disable it, even though it's being enabled by default?

kacperkapusciak:
@yanush add these lines to entry file of your app:
```js
import { enableScreens } from 'react-native-screens';

enableScreens(false);
```

PR #872, opened 2021-03-31
docs: remove from READMEs info about enabledScreens being required

kacperkapusciak:
## Description

Calling `enableScreens()` at the top of your application is no longer required since `v3.0.0` and this should be reflected in the documentation.

## Changes

- Updated README.md
- Updated native-stack/README.md
- Updated createNativeStackNavigator/README.md

## Checklist

- [x] Updated documentation: 

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,9 +43,7 @@ Read usage guide depending on if you are [using Expo](#usage-in-expo-with-react-
 
 Screens support is built into [react-navigation](https://github.com/react-navigation/react-navigation) starting from version [2.14.0](https://github.com/react-navigation/react-navigation/releases/tag/2.14.0) for all the different navigator types (stack, tab, drawer, etc). We plan on adding it to other navigators shortly.
 
-To configure react-navigation to use screens instead of plain RN Views for rendering screen views, follow the steps below:
-
-1.  Add this library as a dependency to your project:
+To configure react-navigation to use screens instead of plain RN Views for rendering screen views, simply add this library as a dependency to your project:
 
 ```bash
 # bare React Native project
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -55,19 +53,9 @@ yarn add react-native-screens
...
-enableScreens();
-```
-
-Note that the above code needs to execute before the first render of a navigation screen. You can check the Example's app [App.js](https://github.com/kmagiera/react-native-screens/blob/master/Example/App.js#L16) file as a reference.
-
-3.  Make sure that the version of [react-navigation](https://github.com/react-navigation/react-navigation) you are using is 2.14.0 or higher
+Just make sure that the version of [react-navigation](https://github.com/react-navigation/react-navigation) you are using is 2.14.0 or higher.
 
-4.  You are all set 🎉 – when screens are enabled in your application code react-navigation will automatically use them instead of relying on plain React Native Views.
+You are all set 🎉 – when screens are enabled in your application code react-navigation will automatically use them instead of relying on plain React Native Views.
```
WoLewicki, line 58: Please add here a note with the same note as in other files, with the information how to disable them.

createNativeStackNavigator/README.md
```diff
@@ -310,3 +302,13 @@ const largeHeaderInset = statusBarInset + 96; // inset to use for a large header
 ```
 
 You can also see an example of using these values with a `ScrollView` here: https://snack.expo.io/@wolewicki/ios-header-height.
+
+### Disabling `react-native-screens`
```
WoLewicki, line 306: Please move it to the place where there was information about enabling them. Same for below file.

WoLewicki (approved):
Please address my comments. Looks good otherwise!

README.md
kacperkapusciak: Fixed in e6e16cc492349bb0433e3ef952d8afe16fcd6f59

createNativeStackNavigator/README.md
kacperkapusciak: Fixed in e6e16cc492349bb0433e3ef952d8afe16fcd6f59

PR #848, opened 2021-03-09
feat: add slide_from_bottom & fade_from_bottom animations

kacperkapusciak:
## Description
PR originally authored by @itsjgf in https://github.com/software-mansion/react-native-screens/pull/662
Add support for `slide_from_bottom` and `fade_from_bottom` animations on both platforms.
`fade_from_bottom` animation mimics default Android Nougat animation.

Fixes #851, #651

## Changes

- Added slide_from_bottom animation
- Added fade_from_bottom animation
- Added example
- Updated types
- Updated documentation

## Screenshots / GIFs

`slide_from_bottom` on iOS

https://user-images.githubusercontent.com/39658211/123950477-04017380-d9a4-11eb-80ba-82c3804fffac.mp4

`slide_from_bottom` on Android

https://user-images.githubusercontent.com/39658211/123950560-1c718e00-d9a4-11eb-9176-2bdaab235f0f.mp4

`fade_from_bottom` on iOS

https://user-images.githubusercontent.com/39658211/123950617-2bf0d700-d9a4-11eb-820e-2c324d597ff8.mp4

`fade_from_bottom` on Android

https://user-images.githubusercontent.com/39658211/123950690-3ca14d00-d9a4-11eb-8195-d270e0eac5ea.mp4

## Checklist

- [x] Included code example that can be used to test this change
- [x] Updated TS types
- [x] Updated documentation: 
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/README.md
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/native-stack/README.md
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/createNativeStackNavigator/README.md
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/src/types.tsx
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/src/native-stack/types.tsx
- [x] Ensured that CI passes

PR #890, opened 2021-04-15
feat: expose transition progress

WoLewicki:
## Description

PR adding event with transition progress sent while animating between screens of `native-stack`. The progress is sent to both transitioning screens.

## Changes

- Added `onTransitionProgress` event
- Added logic for calculating, and sending to JS, transition progress for all kinds (hopefully) of transitions for iOS and Android

`Test887.tsx`

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [x] Ensured that CI passes


Hirbod:
CC @IjzerenHein
Is this something that works for you to implement react-native-shared-element with native-stack?

mrousavy:
What would be needed in order to pipe the `RCTDirectEventBlock` (`onTransitionProgress` event) through reanimated v2 to handle it completely synchronous on the UI thread (in a REA worklet)?

I think this would be beneficial since pushing a screen could potentially be a timeframe where the JS thread is really busy (rendering the newly pushed screen, effects, etc), so running the animations directly on the UI thread would yield much greater performance.

ios/RNSScreenStack.m
```diff
@@ -83,6 +86,24 @@ - (UIViewController *)reactViewController
 
 - (void)navigationController:(UINavigationController *)navigationController willShowViewController:(UIViewController *)viewController animated:(BOOL)animated
 {
+//  if (animated && [viewController.view isKindOfClass:[RNSScreenView class]] && _controller.transitionCoordinator != nil) {
```
kacperkapusciak, line 89: Are these left on purpose?

ios/RNSScreenStack.m
```diff
@@ -94,6 +115,14 @@ - (void)navigationController:(UINavigationController *)navigationController will
   [RNSScreenStackHeaderConfig willShowViewController:viewController animated:animated withConfig:config];
 }
 
+- (void)handleAnimation
+{
+//  if (_transitioningView != nil) {
```
kacperkapusciak, line 120: Same here.

ios/RNSScreenStack.m
WoLewicki: These are remainings of the previous implementation which I didn't want to lose. I'll remove them.

src/native-stack/types.tsx
```diff
@@ -222,6 +223,10 @@ export type NativeStackNavigationOptions = {
    * Boolean indicating whether the navigation bar is translucent.
    */
   headerTranslucent?: boolean;
+  /**
+   * A callback called every frame during the transition of screens of `native-stack`.
+   */
+  onTransitionProgress?: (e: TransitionProgressEvent) => void;
```
satya164, line 229: IMO there shouldn't be such an option. We don't expose events in options, we have a separate event system for it. But it also probably shouldn't be a regular event coz then we would need to do animations in JS only. Instead we should expose an `Animated.Value` like JS stack does. JS stack has a `useCardAnimation()` hook that we can use in screens in stack. It has many properties in JS stack, but for native stack, this should be enough:

```js
{
  /**
   * Values for the current screen.
   */
  current: {
    /**
     * Animated node representing the progress value of the current screen.
     */
    progress: Animated.AnimatedInterpolation;
  };
  /**
   * Values for the screen after this one in the stack.
   * This can be `undefined` in case the screen animating is the last one.
   */
  next?: {
    /**
     * Animated node representing the progress value of the next screen.
     */
    progress: Animated.AnimatedInterpolation;
  };
};
```

We'd basically need to create animated values for each screen that we can update with `Animated.event` and `useNativeDriver: true`.

ios/RNSScreen.m
```diff
@@ -233,6 +233,16 @@ - (void)notifyDisappear
   }
 }
 
+- (void)notifyTransitionProgress:(double)progress closing:(BOOL)closing
```
satya164, line 236: Let's ensure that this works with `Animated.event` with`useNativeDriver: true`.

android/src/main/java/com/swmansion/rnscreens/ScreenTransitionProgressEvent.java
```diff
@@ -0,0 +1,41 @@
+package com.swmansion.rnscreens;
+
+import com.facebook.react.bridge.Arguments;
+import com.facebook.react.bridge.WritableMap;
+import com.facebook.react.uimanager.events.Event;
+import com.facebook.react.uimanager.events.RCTEventEmitter;
+
+public class ScreenTransitionProgressEvent extends Event<ScreenAppearEvent> {
+
+  public static final String EVENT_NAME = "topTransitionProgress";
```
satya164, line 10: Is there a reason to call it `topTransitionProgress` instead of `onTransitionProgress`?

satya164 (changes requested):

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.java
```diff
@@ -184,14 +195,70 @@ protected void dispatchOnDisappear() {
...
+
+    dispatchTransitionProgress(1.0f, true, true);
+  }
+
+  public void incrementCoalescingKey() {
+    mCoalescingKey++;
+  }
+
+  protected void dispatchTransitionProgress(float alpha, boolean closing, boolean shouldIncrementCoalescingKey) {
+    if (shouldIncrementCoalescingKey) {
```
kmagiera, line 208: What's the purpose of incrementing the key?

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.java
```diff
@@ -238,6 +305,14 @@ public void run() {
     }
   }
 
+  public boolean isSendingProgress() {
+    return mIsSendingProgress;
+  }
+
+  public void setIsSendingProgress(boolean isSendingProgress) {
```
kmagiera, line 312: typically when getter starts with "is" the setter does not need "is" in java (that's the getter/setter naming convention), so this should just be "setSendingProgress"

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.java
```diff
@@ -184,14 +195,70 @@ protected void dispatchOnDisappear() {
...
+    mCoalescingKey++;
+  }
+
+  protected void dispatchTransitionProgress(float alpha, boolean closing, boolean shouldIncrementCoalescingKey) {
+    if (shouldIncrementCoalescingKey) {
+      incrementCoalescingKey();
+    }
+    sendTransitionProgressEvent(alpha, closing);
+
+    if (mAboveScreen != null && mAboveScreen.getFragment() != null && !mAboveScreen.getFragment().isSendingProgress()) {
```
kmagiera, line 213: Can you add a comment explaining this here. Why would more than one view send progress at a given time? Is there a problem when they do? I understand that progress updates will be delivered to separate views

android/src/main/java/com/swmansion/rnscreens/ScreenTransitionProgressEvent.java
```diff
@@ -0,0 +1,41 @@
...
+
+  @Override
+  public String getEventName() {
+    return EVENT_NAME;
+  }
+
+  @Override
+  public short getCoalescingKey() {
+    // All events for a given view can be coalesced.
+    return mCoalescingKey;
```
kmagiera, line 31: if you want all events for single view to be coalesced you should just return the same key here. I think that what would be bettere here actually would be to make the key depend on `mClosing` such that we don't coalesce closing and opening events

createNativeStackNavigator/README.md
```diff
@@ -185,6 +185,10 @@ A Boolean to that lets you opt out of insetting the header. You may want to * se
 
 Boolean indicating whether the navigation bar is translucent.
 
+#### `onTransitionProgress`
+
+A callback called every frame during the transition of screens.
```
kmagiera, line 190: Maybe also explain that the event provides the information of the transition progress (0-1)

ios/RNSScreen.m
```diff
@@ -612,6 +636,53 @@ - (void)notifyFinishTransitioning
...
+  }
+}
+
+- (void)notifyTransitionProgress:(double)progress closing:(BOOL)closing
+{
+  [((RNSScreenView *)self.view) notifyTransitionProgress:progress closing:closing];
+  // if we are in a modal, we want to send transition to the above screen too, which might not trigger appear/disappear events
+  // e.g. when we are in iOS >= 13 default modal
+  // we also check if we are not already sending the progress of transition and if it does not present another modal,
+  // because otherwise we would send progress to the screen 2 levels higher then the current one
```
kmagiera, line 680: I don't understand this. Why would this method send events to the view that is not transitioning at the moment?

ios/RNSScreen.m
```diff
@@ -612,6 +636,53 @@ - (void)notifyFinishTransitioning
...
+    _fakeView = fakeView;
+    [self.transitionCoordinator animateAlongsideTransition:^(id<UIViewControllerTransitionCoordinatorContext>  _Nonnull context) {
+      [[context containerView] addSubview:fakeView];
+      fakeView.alpha = 1.0;
+      self->_animationTimer = [CADisplayLink displayLinkWithTarget:self selector:@selector(handleAnimation)];
+      [self->_animationTimer addToRunLoop:[NSRunLoop currentRunLoop] forMode:NSDefaultRunLoopMode];
+    } completion:^(id<UIViewControllerTransitionCoordinatorContext>  _Nonnull context) {
+      [self->_animationTimer setPaused:YES];
+      [self->_animationTimer invalidate];
+      self->_fakeView = nil;
```
kmagiera, line 656: do we need fakeView to be retained under `this`?

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.java
```diff
@@ -39,6 +46,10 @@ protected static View recycleView(View view) {
   }
 
   protected Screen mScreenView;
+  private boolean mIsSendingProgress;
+  private Screen mAboveScreen;
```
kmagiera, line 50: I'd be hesitant of storing reference to other screens here especially if we want to get reference to a screen that's directly before this one in the container. Note that this order can bemodified and that other screen can be removed (e.g. as a result of JS update) from the container without any notice. In this case we'd keep a reference to an orphaned view here

kmagiera (changes requested):

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.java
WoLewicki: I think we want the progress of `0.0` and `1.0` to be always sent to the JS, and such values are dispatched in lifecycle events, so we increment the key in those methods.

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.java
WoLewicki: We send the transition progress to both of the transitioning screens. We do it by subscribing to the ongoing transition of fragments: https://github.com/software-mansion/react-native-screens/pull/890/files#diff-2c98f4cb51ad28895af60cd79d14b67a66256480a554d81a9dfa6f51c84061c2R41. But if we are pushing a `transparentModal`, the fragment below does not participate in the transition, so it does not call the `onAnimationStart` method. In order to achieve the state when always 2 views send progress during the transition, we check if above screen is sending progress, and, if not, we do it from the screen being pushed. Same applies to modals on iOS.

android/src/main/java/com/swmansion/rnscreens/ScreenTransitionProgressEvent.java
WoLewicki: This comment should not be there, it slipped in when I copied the code regarding this. We already increment the key on closing and opening events, as mentioned in the https://github.com/software-mansion/react-native-screens/pull/890#discussion_r646477369.

ios/RNSScreen.m
WoLewicki: The explanation is the same as in https://github.com/software-mansion/react-native-screens/pull/890#discussion_r646561799.

ios/RNSScreen.m
WoLewicki: We use it in every tick of `_animationTimer` and I am not sure how else can it be used in the `handleAnimation` of it (see https://github.com/software-mansion/react-native-screens/pull/890/files#diff-e988f8ab84e7063d6cc9ad4e634806389882ce6727b5c95f3cfbc9f9d3921a67R665)

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.java
WoLewicki: We could make `mAboveScreen` null at the end of the transition so it is only alive for the lifetime of it.

WoLewicki: ^ done

src/native-stack/views/NativeStackView.tsx
```diff
@@ -107,6 +107,122 @@ const MaybeNestedStack = ({
...
 
+type RouteViewProps = {
+  route: Route<string>;
+  index: number;
+  navigation: NativeStackNavigationHelpers;
+  descriptors: NativeStackDescriptorMap;
+  navigatorKey: string;
+};
+
+function RouteView({
```
kmagiera, line 118: what now is the purpose of this component? I remember previously when we discussed some implementation in the past there were some animated values being generated on a per route basis, but now it seems to no longer be the case?

src/reanimated/ReanimatedScreen.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  setRef = (ref: React.ElementRef<typeof View> | null): void => {
+    this.tag = findNodeHandle(ref);
+    this.ref = ref;
+    this.props.onComponentRef?.(ref);
+    if (this.eventHandler == null && this.tag != null) {
+      const progressRef = this.progress;
+      const closingRef = this.closing;
+      const goingForwardRef = this.goingForward;
+      this.eventHandler = new WorkletEventHandler(
```
kmagiera, line 41: now that this component is extracted maybe we can use useEvent hook from reanimated instead of doing it this way?

kmagiera (commented):
It appears like quite a bit of complexity is added in order to support transparent screen case. Can you explain what the issue is in that case? I wonder if we can find a cleaner way to support this usecase avoiding quite a complex logic for handling it?

My second worry is about interupting transition (e.g. you go back before the animation is over) – have you tested this scenarios? Any thoughts on that?

Also please see my comments I made inline.

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.java
```diff
@@ -182,14 +188,80 @@ protected void dispatchOnDisappear() {
...
+    sendTransitionProgressEvent(alpha, closing, goingForward);
+
+    if (mAboveScreen != null
+        && mAboveScreen.getFragment() != null
+        && !mAboveScreen.getFragment().isSendingProgress()) {
+      if (shouldIncrementCoalescingKey) {
+        mAboveScreen.getFragment().incrementCoalescingKey();
+      }
+      // if we are in transparentModal presentation, only one screen is sending progress
+      mAboveScreen.getFragment().sendTransitionProgressEvent(alpha, !closing, goingForward);
```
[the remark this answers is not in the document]
WoLewicki: I am not sure what more than the comment in `ScreenStack.java` (https://github.com/software-mansion/react-native-screens/pull/890/files#diff-b1439496d79d456c74fa09061a5e3fef9e2e7c35e7c3ba434343ada55695abf6R259). As for removing `mAboveScreen`, we could just not send progress to non-animating screens, or maybe deal with it from JS. I am open to other options since this one is certainly not perfect.

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.java
```diff
@@ -22,7 +24,61 @@
...
+          public void onAnimationRepeat(Animation animation) {}
+        };
+
+    @Override
+    public void startAnimation(Animation animation) {
+      // For some reason View##onAnimationEnd doesn't get called for
+      // exit transitions so we use this hack.
+      AnimationSet set = new AnimationSet(true);
+      set.addAnimation(animation);
+      set.setAnimationListener(mAnimationListener);
```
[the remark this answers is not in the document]
WoLewicki: I will definitely have to look if it works correctly after merging #848 or find another way to attach a listener for during the animation. I haven't looked too much for other options, so maybe there are better ways of doing it, maybe returning the current exact transformation of a screen on every frame, something similar to what is done on iOS.

ios/RNSScreen.m
```diff
@@ -524,19 +541,38 @@ - (void)viewWillAppear:(BOOL)animated
...
+      _goingForward = NO;
     }
   } else {
     // this event is also triggered if we cancelled the swipe.
     // The _isSwiping is still true, but we don't want to notify then
     _shouldNotify = NO;
   }
 
+  if (_goingBackWithJS) {
+    // we are going back with JS, so previous assumptions may have been wrong
```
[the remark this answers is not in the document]
WoLewicki: Maybe it should be called `goingBackFromJS`, it means that the navigation state changed on the JS side and current top screen is no longer in the stack. Then the assumption `_goingForward = parentChildrenCount - 1 == selfIndex;` is wrong as it only works when navigating natively by swipe or native header back button.

ios/RNSScreenStack.m
```diff
@@ -350,6 +350,15 @@ - (void)setModalViewControllers:(NSArray<UIViewController *> *)controllers
         BOOL shouldAnimate = lastModal && [next isKindOfClass:[RNSScreen class]] &&
             ((RNSScreenView *)next.view).stackAnimation != RNSScreenStackAnimationNone;
 
+        if ([next isKindOfClass:[RNSScreen class]]) {
+          if ([previous isKindOfClass:[RNSScreen class]]) {
+            ((RNSScreen *)next).presentingScreen = (RNSScreen *)previous;
```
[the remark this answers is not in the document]
WoLewicki: What do you mean by existing methods? I could not find another way to make sure we always send the progress to the correct screen, e.g. when dismissing 2 screens at once from JS. As for weak references, is it not enough to set the `presentingScreen` property as weak (https://github.com/software-mansion/react-native-screens/pull/890/files?file-filters%5B%5D=.h&file-filters%5B%5D=.java&file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.lock&file-filters%5B%5D=.md&file-filters%5B%5D=.tsx#diff-2ee26dd0f63d0d5fffc71b24dbee03d15940367a83bcb2a5c95e3b9e9470e131R67)?

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.java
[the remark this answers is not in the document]
WoLewicki: I merged the code from #848 and it still works correctly. Looks like the same instance of animation is used when going back (we don't attach a listener then), so the listener is still there.

src/native-stack/views/NativeStackView.tsx
WoLewicki: Yes, I'll revert it.

src/index.native.tsx
```diff
@@ -149,7 +155,31 @@ class Screen extends React.Component<ScreenProps> {
           statusBarColor={processedColor}
           activityState={activityState}
           ref={this.setRef}
-        />
+          onTransitionProgress={Animated.event(
```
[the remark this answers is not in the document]
WoLewicki: I think we came to a conclusion that it is always present in `stack` navigator and we want to have the same behavior in `native-stack`.

ios/RNSScreen.m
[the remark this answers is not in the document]
WoLewicki: Update: changed it to methods provided by `UIViewController` as suggested in its documentation.

kmagiera (approved):
This looks much cleaner after recent changes. Don't have any more comments and I think it is fine to merge it. 👏👏👏👏👏

PR #941, opened 2021-05-13
fix: make events dispatch in correct order

WoLewicki:
## Description

PR made on top of #593 fixing the correct order (always willAppear/willDisappear before appear/disappear) on Android to match iOS where it is already correct.

## Changes

- Added dispatching `onAppear` and `onDisappear` on the ui thread suggested in #593 both for normal and `none` transitions.
- Added code in `ScreenStack` making sure that the `willAppear` and `appear` are always dispatched when the new stack appears.
- Changed adding parent fragment order to be present when we setup fragment manager since we sometimes want to know if it is nested in `onUpdate` method.

`Test593.tsx` in `TestsExample` project.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/TestsExample/App.js
+++ b/TestsExample/App.js
@@ -11,6 +11,7 @@ import Test550 from './src/Test550';
 import Test556 from './src/Test556';
 import Test564 from './src/Test564';
 import Test577 from './src/Test577';
+import Test593 from './src/Test593';
 import Test619 from './src/Test619';
 import Test624 from './src/Test624';
 import Test640 from './src/Test640';
--- a/TestsExample/ios/Podfile.lock
+++ b/TestsExample/ios/Podfile.lock
@@ -355,7 +355,7 @@ PODS:
     - React-RCTVibration
     - ReactCommon/turbomodule/core
     - Yoga
-  - RNScreens (3.1.1):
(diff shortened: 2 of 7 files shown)


WoLewicki:
The test case was going from the `Status` screen into `Deeper` screen with nested stack having `Privacy` screen as a first one. Then going from `Privacy` to `Another`, which is a second screen in the nested stack. Then going back straight to `Status` screen, so popping the nested stack and `Deeper` screen at once.

With `stackAnimation: 'slide_from_left', 'default', 'fade'`:
image

With `stackAnimation: 'none'`:
image

We can spot the difference in `stackAnimation: 'none'` between the platforms since on Android we dispatch both willAppear-appear and willDisappear-disappear immediately one after another, so first the screen we are transitioning from dispatches willDisappear and disappear and then the screen we are transitioning to dispatches willAppear and appear. On iOS, the ordering is somehow switched too, but the must-have, so first xDisappear then xAppear still applies for both platforms.

WoLewicki:
@janicduplessis could you check if it is exactly what we were trying to achieve with #593?

janicduplessis:
@WoLewicki Cool! Yes looks like I was trying to achieve. LGTM

PR #965, opened 2021-06-07
fix: dont warn about header if prop not changed

WoLewicki:
## Description

Added changes from https://github.com/react-navigation/react-navigation/blob/main/packages/native-stack/src/views/NativeStackView.tsx#L28 to display warning about header in modal only after the prop has been changed, so it will not display the warning if the user specified `headerShown: true` in the options for the screen.

`Test800.tsx` in `TestsExample` project.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,9 @@
     "url": "https://github.com/software-mansion/react-native-screens/issues"
   },
   "homepage": "https://github.com/software-mansion/react-native-screens#readme",
+  "dependencies": {
+    "warn-once": "^0.1.0"
+  },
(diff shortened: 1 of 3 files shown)


kacperkapusciak (approved):
LGTM

robinvw1:
Thanks @WoLewicki! I got this old warning `Be aware that changing the visibility of header in modal on iOS will result in resetting the state of the screen.` every time, without changing the `headerShown` prop. Hopefully this will land soon in the new release. :)

PR #970, opened 2021-06-08
fix: keyboard dismissing on swipe

WoLewicki:
## Description

Added code resolving in dismissing keyboard on swipe to go back. Other methods don't seem to work, even calling `Keyboard.dismiss()` in the start of swipe, which resolves in calling `[self resignFirstResponder];` on the `UITextView` (see https://github.com/facebook/react-native/blob/0afd71a18d19edef3d6603cabd2672ffde5ba30e/React/Views/UIView%2BReact.m#L276), but it seems not enough, like if there was some other responder still not resigning. The same thing works in normal `stack` navigator.

## Changes

Added `[self endEditing:YES];` in `notifyWillDisappear` method of `RNSScreenView`.

`Test958.tsx` in `TestsExample` project.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/TestsExample/App.js
+++ b/TestsExample/App.js
@@ -45,6 +45,7 @@ import Test861 from './src/Test861';
 import Test865 from './src/Test865';
 import Test881 from './src/Test881';
 import Test898 from './src/Test898';
+import Test958 from './src/Test958';
 
 export default function App() {
   return <Test42 />;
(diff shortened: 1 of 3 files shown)


kacperkapusciak (changes requested):
Xcode complains about your solution with this warning message:
```
[Snapshotting] Snapshotting a view (0x7fc0c1054880, UIInputSetContainerView) that has not been rendered at least once requires afterScreenUpdates:YES.
```
To reproduce: write something in the input, try using the back gesture and release without going back.

WoLewicki:
I found such a comment: https://developer.apple.com/forums/thread/63306?answerId=231289022#231289022 and https://stackoverflow.com/a/24625591/13006595. I think that when the transition starts, the snapshot of the disappearing view is taken, or something similar to this. I am not sure what should we do about this.

WoLewicki:
Closing since it is not expected behavior.

PR #974, opened 2021-06-10
feat: remove enableScreens from example app

### LINKED ISSUE #973: No need to run enableScreens()
As of version 3.0.0 `You no longer need to call enableScreens() at the top of your application` ([click](https://github.com/software-mansion/react-native-screens/releases/tag/3.0.0))

https://github.com/software-mansion/react-native-screens/blob/b5171609cf8b0455aca0ad9710aff3f997cea1a6/Example/App.tsx#L30

kacperkapusciak:
## Description

Example app uses the newest version of `react-native-screens` thus it no longer requires to call `enableScreens()`.

Fixes #973

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/Example/App.tsx
+++ b/Example/App.tsx
@@ -27,8 +27,6 @@ import Modals from './src/screens/Modals';
 import Orientation from './src/screens/Orientation';
 import SearchBar from './src/screens/SearchBar';
 
-enableScreens();
-
 if (Platform.OS === 'android') {
   StatusBar.setTranslucent(true);
 }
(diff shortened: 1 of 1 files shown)


Example/App.tsx
```diff
@@ -27,8 +27,6 @@ import Modals from './src/screens/Modals';
 import Orientation from './src/screens/Orientation';
 import SearchBar from './src/screens/SearchBar';
 
-enableScreens();
```
WoLewicki, line 30: I think you should also remove the import of the method since it is not used

kacperkapusciak: Fixed in af9aec6

PR #761, opened 2020-12-28
feat: add simple push/pop transition on ios

WoLewicki:
## Description

Added a simple transition on iOS, similar to the default one, but without the animation of header and shadow. The code has been mainly taken from https://stackoverflow.com/a/45827454/13006595. Requested in #628.

## Changes

Added new `stackAnimation`: `simple_push`. Added proper handling on iOS and Android.

## Test code and steps to reproduce

`Test761.tsx` in `TestsExample` project.

## Checklist

- [x] v4 implementation
- [x] `rtl` on swipe detection
- [x] Included code example that can be used to test this change
- [x] Updated TS types
- [x] Updated documentation: 
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/README.md
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/native-stack/README.md
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/createNativeStackNavigator/README.md
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/src/types.tsx
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/src/native-stack/types.tsx
- [x] Ensured that CI passes

PR #984, opened 2021-06-15
strip tvOS incompatible code in tvOS builds

### LINKED ISSUE #950: TvOS compiling issues
## Description

The package is not compiling on TvOS version 3.2.0 and 3.3.0
It successfully compiles on version 3.1.1

## Package versions

- React: 16.13.1
- React Native: npm:react-native-tvos@0.63.1-4
- React Native Screens: 3.2.0 or 3.3.0

Kazetsukai:
## Description

Fixing compile errors on tvOS
Fixes #950

## Changes

- wraps code that won't compile on tvOS with compile-time conditionals

Attempt to use the library as-is in a tvOS React Native project, and you will have compile errors relating to some calls not being available on tvOS.

## Checklist

- [ ] Ensured that CI passes


kacperkapusciak (changes requested):
The `lib/` folder isn't meant to be committed. These files are automatically generated and they shouldn't be in the repo. It's added to the [.gitignore here](https://github.com/software-mansion/react-native-screens/blob/master/.gitignore#L51) and I'm confused how and why you managed to commit it anyway.

ios/RNSScreenStack.m
```diff
@@ -507,6 +511,7 @@ - (void)dismissOnReload
   return nil;
 }
 
+#if !TARGET_OS_TV
```
WoLewicki, line 514: Should we eliminate whole `gestureRecognizerShouldBegin` method on TVOS? Is it never used there?

Kazetsukai: I am probably misunderstanding something here, but I thought this was only used for the edge swipe recognisers defined in `setupGestureHandlers` which don't exist on tvOS. My Obj-C skills aren't the best, does removing this interact with something outside of `RNSScreenStack`?

Kazetsukai:
Woops, was playing around with something locally, didn't mean to add lib to this, reverted.

ios/RNSScreenStack.m
WoLewicki: You can see here the implementation before `setupGestureHandlers` was introduced: https://github.com/software-mansion/react-native-screens/pull/761/files#diff-1c3e13a1b0dcbaa5f7cff5f3ce12cc915fb509725faf693fb65cd23ceb49c4a9L141, so I think we should do something like
```objc
#if TARGET_OS_TV
	return YES;
#else
	// rest of the code
```

and it should start here: https://github.com/software-mansion/react-native-screens/pull/761/files#diff-1c3e13a1b0dcbaa5f7cff5f3ce12cc915fb509725faf693fb65cd23ceb49c4a9R526, but I am not sure if this is 100% correct and if this method is ever fired on `tvOS`.

WoLewicki: If we do it like that, maybe we should also resolve using `simple_push` to default transition so we don't end up with some weird behaviors on `tvOS`. We could do that by returning `nil` for `tvOS` and `RNSScreenStackAnimationSimplePush` option here: `https://github.com/software-mansion/react-native-screens/blob/a6f91b6b7ef800f87c1ccc1dea95e1ebcb5604e7/ios/RNSScreenStack.m#L503`.

WoLewicki:
I will close this since #988 has been merged. Thanks @Kazetsukai for your contribution and I added you as a co-author of #988 :tada:

PR #987, opened 2021-06-16
chore: format native code

WoLewicki:
## Description

Added scripts for linting java code from https://github.com/facebook/react-native/blob/b176617d1e0cb7834698f147db4522290098ea00/scripts/lint-java.js and for iOS code from https://stackoverflow.com/a/48459965/13006595.

## Checklist

- [x] Ensured that CI passes


package.json
```diff
@@ -6,7 +6,9 @@
     "check-types": "tsc --noEmit",
     "start": "react-native start",
     "test:unit": "jest --passWithNoTests",
-    "format": "prettier --write --list-different './src/**/*.{js,ts,tsx}'",
+    "format": "prettier --write --list-different './src/**/*.{js,ts,tsx}' && yarn clang-format && yarn lint-java",
+    "lint-java": "node ./scripts/lint-java.js",
+    "clang-format": "find ios/ -iname *.h -o -iname *.m -o -iname *.cpp | xargs clang-format -i",
```
kacperkapusciak, line 11: For this to work, you have to have `clang-format` installed on your machine. We should definitely add a [clang-format](https://github.com/angular/clang-format#readme) package as a dev-dependency to the repo. That way we avoid a situation where a (to be) contributor can't even commit the code without an additional setup.

PR #1006, opened 2021-07-08
async -> sync

Szymon20000:
## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
@@ -23,17 +23,7 @@ public class ScreenContainer<T extends ScreenFragment> extends ViewGroup {
   protected final ArrayList<T> mScreenFragments = new ArrayList<>();
 
   protected @Nullable FragmentManager mFragmentManager;
-  private @Nullable FragmentTransaction mCurrentTransaction;
-  private @Nullable FragmentTransaction mProcessingTransaction;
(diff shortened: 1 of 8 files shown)


android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java
```diff
@@ -201,47 +185,36 @@ private void setupFragmentManager() {
   }
 
   protected FragmentTransaction getOrCreateTransaction() {
-    if (mCurrentTransaction == null) {
-      mCurrentTransaction = mFragmentManager.beginTransaction();
-      mCurrentTransaction.setReorderingAllowed(true);
+    if (mCurrentTransaction != null) {
```
kmagiera, line 188: this seems to no longer be necessary if we create new transaction and then commit it with each single operation. As a consequence we wouldn't need the `mCurrentTransaction` field nor need to call getOrCreaeTransaction nor tryCommitTransaction

WoLewicki: I think it was left for usage with `ScreenStack`, where we have animations in `Fragments`, so we want to concatenate all the operations into one and run animations on them. Btw, I would change the name `commitNowAllowingStateLoss` to something like `commitTransaction`.

WoLewicki:
Closed in favor of #1066.

Issue #619, opened 2020-08-30
Header Height jumps with nested navigators


dotconnor:
When you have a nested navigators like this: `NativeStack -> BottomTabs -> NativeStack`, it creates issues when rendering the header from the innermost NativeStack. Only appears to happen when there some other navigator anywhere between two NativeStack ones. For example `Stack -> BottomTabs -> NativeStack` works correctly and `NativeStack -> NativeStack` work correctly.

This seems to be the same bug as #540.

https://github.com/dotconnor/ScreensHeader

## Expected

![Kapture 2020-08-30 at 05 47 11](https://user-images.githubusercontent.com/14323370/91656166-8d929d00-ea84-11ea-8264-bf9776ee003c.gif)

## Actual

![Kapture 2020-08-30 at 05 49 15](https://user-images.githubusercontent.com/14323370/91656176-a438f400-ea84-11ea-8967-c360452414cb.gif)

WoLewicki:
Looks like it is the same issue as https://github.com/software-mansion/react-native-screens/issues/564. Am I right?

dotconnor:
Just tested this, and the issue still happens regardless of the `View` inside the screen, secondly it also happens regardless if the translucent header option is set or not.

WoLewicki:
There seem to be some layout issues when having a `UIViewController` (from `ScreenContainer` of `bottom-tabs`) between 2 `UINavigationController`s (from `ScreenStack` of `native-stack`). As a workaround, you can make the top navigator a normal `stack` (since you don't use any of the `native-stack` specific options, right?).

ArekChr:
> There seem to be some layout issues when having a `UIViewController` (from `ScreenContainer` of `bottom-tabs`) between 2 `UINavigationController`s (from `ScreenStack` of `native-stack`). As a workaround, you can make the top navigator a normal `stack` (since you don't use any of the `native-stack` specific options, right?).

I'm using large and blurred headers from native-stack

WoLewicki:
@ArekChr It is not a solution to the bug, but a workaround, and I know it is wanted for many cases.

burakgormek:
@WoLewicki Any update? I have same issue. I just noticed that this is only happens if stack has animation.
stackAnimation 'none' works without an issue.

simplified test: https://snack.expo.io/KrD6WIJ2l (looks like iOS only)

WoLewicki:
[...]

Issue #564, opened 2020-07-13
Jumpdown ScrollView with auto insets on BottomTab inside NativeStack


SDSLeon:
related issue in react-navigation was closed so I open it here

**Current Behavior**

ScrollView with contentInsetAdjustmentBehavior="automatic" initialy render under translucent header and after stack animation jumps down; 

GIF: [url](https://storage.de.cloud.ovh.net/v1/AUTH_b2cffe8f45324c2bba39e8db1aedb58f/cloudconvert-files/87d0b94d-9da2-417d-837a-5ad54414fe58/Запись%20экрана%202020-07-13%20в%2015.35.56.gif?temp_url_sig=0998bd7d02e92abf5590769b1d8ef8edcac63104&temp_url_expires=1594731587)

**Expected Behavior**

ScrollView with contentInsetAdjustmentBehavior="automatic" initialy render on bottom of header; 

**How to reproduce**
Example: https://github.com/SDSLeon/nativeStackPlusTabTranslucentScroll

after some time of investigation:
1. Problem is when bottomTabsNavigation uses react-native-screens to render tab content
![image](https://user-images.githubusercontent.com/6071034/87339550-72f57a80-c54f-11ea-9abf-ed44ae56660e.png)
 

**Your Environment**

| software                      | version |
| ----------------------------- | ------- |
| iOS                | all versions
| @react-navigation/native      | 5.7.0
| @react-navigation/bottom-tabs | 5.7.0
| react-native-screens          | 2.9.0
| react-native                  | 0.63.0
| expo                          |
| node                          | 13.11
| npm or yarn                   | yarn 1.22.4

WoLewicki:
Can you check if #1029 resolves your issues and does not introduce new ones? It requires applying PR from `react-navigation` for it to work: https://github.com/react-navigation/react-navigation/pull/9772.

PR #1029, opened 2021-07-28
feat: expose new container for 2-state navigators

WoLewicki:
## Description

PR adding `RNSScreenNavigationContainer` component on iOS. This component should be used in `bottom-tabs` and `drawer` navigators of `react-navigation` in order to resolve layout issues connected to integration between `UIViewController` and `UINavigationController` containers (see #619 and #564). `stack` navigator unfortunately cannot use `UINavigationController` since it needs the previous screens to be visible e.g. during the transition. Connected PR in `react-navigation`: https://github.com/react-navigation/react-navigation/pull/9772

## Changes

Added new component and `hasTwoStates` prop for `ScreenContainer` which controls which native component should be used.

`Test619.js` and `Test564.js` to see the layout issues gone. It needs setting `hasTwoStates` props on `bottom-tabs` navigator in order for this to work.

## Checklist

- [x] Updated TS types
- [x] Ensured that CI passes


src/types.tsx
```diff
@@ -179,6 +179,10 @@ export interface ScreenContainerProps extends ViewProps {
    * A prop that gives users an option to switch between using Screens for the navigator (container). All children screens should have the same value of their "enabled" prop as their container.
    */
   enabled?: boolean;
+  /**
+   * A prop to be used in drawer and bottom-tabs navigators for better implementation of ScreenContainer on iOS.
+   */
+  tabsOrDrawer?: boolean;
```
satya164, line 185: I think a prop name like `hasTwoStates` (didn't think a lot about the name) or something like that would be more generic and more appropriate

src/index.native.tsx
```diff
@@ -47,6 +47,8 @@ function enableScreens(shouldEnableScreens = true): void {
 
 // const that tells if the library should use new implementation, will be undefined for older versions
 const shouldUseActivityState = true;
+const isNavigationContainerAvailable =
```
satya164, line 50: Is there a case when this won't be available when someone is using this version of react-native-screens?

WoLewicki: What if someone updates to this version of library in Expo managed workflow? I know it should not be done, but 🤷‍♂️ .

satya164: @WoLewicki updating to an incompatible version of the library could also cause other issues as well, and surface bugs which are already fixed in latest version, so probably not a use case we should support.

kacperkapusciak (approved):

Issue #251, opened 2019-12-14
[Feature Request] Implement gestureResponseDistance


harrisrobin:
Hello,

I was wondering if gestureResponseDistance is a feature that's planned to be implemented in this library. Currently, if you use native stack you lose `gestureResponseDistance` which is [implemented by stack navigator](https://github.com/react-navigation/stack/issues/309).

Cheers 🍻

Jalson1982:
Is there any chance to implement this please?

osdnk:
Hi, 
We'd like to ephazite that priority of React Native Screens is to deliver a native-like experience with a minimal cost of modifying what's actually delivered by the Platform. 
Therefore we don't want to go for supporting features that are not ready out of the box or easy to implement by native components' kit.
Feel free to make this change in your fork or patch but I need to want you that messing with UIKit is what you most likely don't want to do 😃

WoLewicki:
I am closing this issue. Feel free to comment if you have any questions.

tgensol:
Do you think that feature can be implemented ? 

If you see an iOS user, they will all struggle to move back as every others native applications put a gestureDistance bigger (instagram, facebook, snapchat,...).... 

I stopped using your lib and went back to **@react-navigation/stack** because the main issue we had from users was that difficulty for going back in our app (and we asked thousands of people). 

Or do you have any idea how to change it in your lib ? 

Thanks !

WoLewicki:
@tgensol I am afraid there is no such option, as @osdnk mentioned. Although if you manage to find even a hacky solution to this, we are always open for PRs.

JackShannon:
@WoLewicki 

https://github.com/fastred/SloppySwiper

Would it be possible to use this library?

PR #1030, opened 2021-07-29
feat: add full width back gesture on ios

JackShannon:
## Description

https://github.com/software-mansion/react-native-screens/issues/251

This is a hacky solution. I think in order for this to be part of the library there would need to be an option to turn it on/off.

## Changes

Changed the gestureRecognizer to a UIPanGestureRecognizer.

## Screenshots / GIFs

### Before
You have to swipe from the left edge

### After
You can swipe from anywhere:

ezgif-6-ab8f70027cea

-->

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStack.m
+++ b/ios/RNSScreenStack.m
@@ -45,6 +45,30 @@
 {
   return [self topViewController].supportedInterfaceOrientations;
 }
+
+- (void)viewDidLoad
+{
+  [super viewDidLoad];
+  [self setupFullWidthBackGesture];
+}
+
+- (void)setupFullWidthBackGesture
+{
+  NSArray *targets = [self.interactivePopGestureRecognizer valueForKey:@"_targets"];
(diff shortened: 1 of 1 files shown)


JackShannon:
An alternative solution could be to add our own UIScreenEdgePanGestureRecognizer instead of UIPanGestureRecognizer. https://developer.apple.com/documentation/uikit/uiscreenedgepangesturerecognizer?language=objc

That way the api could be consistent with react navigations stack navigator and could use gestureResponseDistance to set the edges.

WoLewicki:
I tested this PR and unfortunately it resolves in some nondeterministic behaviors like freeze of screen when trying to swipe the first screen on stack. I added another version of this PR in #1072. Can you check if it works as you would expect it to and if so, can we close this PR?

PR #1032, opened 2021-08-02
fix: touch event canceled with gesture disabled

### LINKED ISSUE #1022: RN Touch events are canceled even when {gestureEnabled: false} on iOS when using native stack navigation
## Description

For example, if you have a component at the edge like a slider, it's touches are canceled if you start sliding from the left edge. That might be alright if the screen has gesture enabled, but I don't think it's correct to do that when gestureEnabled: false for a screen (totally possible I am missing a case).

I haven't worked with Objective C before but I think the issue is because of this.

https://github.com/software-mansion/react-native-screens/blob/c3998ad5f95d2bbd881d19bce4de52b151fafde4/ios/RNSScreenStack.m#L530-L553

Specifically, the part where touchHandler is cancelled and reset without checking if gesture's enabled for a screen.

I fixed it by applying this patch locally that tests if topScreen.gestureEnabled is truthy before cancelling/resetting touch handlers :
```
@@
diff --git a/node_modules/react-native-screens/ios/RNSScreenStack.m b/node_modules/react-native-screens/ios/RNSScreenStack.m
index 7b76e12..f44f11a 100644
--- a/node_modules/react-native-screens/ios/RNSScreenStack.m
+++ b/node_modules/react-native-screens/ios/RNSScreenStack.m
@@ -509,20 +509,22 @@

 - (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
 {
-  // cancel touches in parent, this is needed to cancel RN touch events. For example when Touchable
-  // item is close to an edge and we start pulling from edge we want the Touchable to be cancelled.
-  // Without the below code the Touchable will remain active (highlighted) for the duration of back
-  // gesture and onPress may fire when we release the finger.
-  UIView *parent = _controller.view;
-  while (parent != nil && ![parent respondsToSelector:@selector(touchHandler)]) parent = parent.superview;
-  if (parent != nil) {
-    RCTTouchHandler *touchHandler = [parent performSelector:@selector(touchHandler)];
-    [touchHandler cancel];
-    [touchHandler reset];
-  }
-  
   RNSScreenView *topScreen = (RNSScreenView *)_controller.viewControllers.lastObject.view;

+  if(topScreen.gestureEnabled) {
+    // cancel touches in parent, this is needed to cancel RN touch events. For example when Touchable
+    // item is close to an edge and we start pulling from edge we want the Touchable to be cancelled.
+    // Without the below code the Touchable will remain active (highlighted) for the duration of back
+    // gesture and onPress may fire when we release the finger.
+    UIView *parent = _controller.view;
+    while (parent != nil && ![parent respondsToSelector:@selector(touchHandler)]) parent = parent.superview;
+    if (parent != nil) {
+      RCTTouchHandler *touchHandler = [parent performSelector:@selector(touchHandler)];
+      [touchHandler cancel];
+      [touchHandler reset];
+    }
+  }
+  
   if (!topScreen.gestureEnabled || _controller.viewControllers.count < 2) {
     return NO;
   }
   ```
## Steps To Reproduce
1. Use native stack with gestureEnabled: false on iOS
2. Use a slider. For ex [this](https://www.npmjs.com/package/@sharcoux/slider), just import and use `<Slider maximumValue={100} />`
3. Start sliding from the left edge, you'll see that the slider is interrupted

### Expected behavior
Touch handler shouldn't be interrupted when gestureEnabled:false

### Actual behavior
Touch handler is interrupted
## Package versions

- React Native: 0.63.4
- React Native Screens: 3.4.0

### PR DESCRIPTION
**Pull request by Ubax**
## Description
Fixes #1022

## Changes
Adds event cancelation only when gesture should be recognized.

Test1032.tsx

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/TestsExample/src/Test1032.tsx
+++ b/TestsExample/src/Test1032.tsx
@@ -1,0 +1,47 @@
+import * as React from 'react';
+import { Alert, Button, Switch, Text, View } from 'react-native';
+import { NavigationContainer, ParamListBase } from '@react-navigation/native';
+import { createNativeStackNavigator, NativeStackNavigationProp } from 'react-native-screens/native-stack';
+
+const Stack = createNativeStackNavigator();
+
+export default function App() {
+    const [gestureEnabled, setGestureEnable] = React.useState(false);
+    return (
+        <View style={{ flex: 1, paddingBottom: 200 }}>
+            <NavigationContainer>
+                <Stack.Navigator screenOptions={{ gestureEnabled }}>
+                    <Stack.Screen name="Top" component={First} options={{ replaceAnimation: 'push' }} />
+                    <Stack.Screen name="Top1" component={Second} options={{ replaceAnimation: 'push' }} />
+                </Stack.Navigator>
+            </NavigationContainer >
+            <View>
+                <Text>Gesture enabled</Text>
+                <Switch value={gestureEnabled} onValueChange={setGestureEnable} />
+            </View>
+        </View>
+    );
+}
+
+function First({
+    navigation,
+}: {
+    navigation: NativeStackNavigationProp<ParamListBase>;
+}) {
+    return (
+        <View style={{ backgroundColor: '#FFF' }}>
+            <Button title="Tap me for second screen" onPress={() => navigation.push('Top1')} />
+        </View>
+
+    );
+}
+
+function Second() {
+    return (
+        <View style={{ backgroundColor: '#FFF' }}>
+            <Button title="Tap me to see test" onPress={() => Alert.alert('Test')} />
+        </View>
+    );
+}
+
+
--- a/TestsExample/src/Test1032.tsx
+++ b/TestsExample/src/Test1032.tsx
@@ -1,47 +1,50 @@
 import * as React from 'react';
-import { Alert, Button, Switch, Text, View } from 'react-native';
-import { NavigationContainer, ParamListBase } from '@react-navigation/native';
-import { createNativeStackNavigator, NativeStackNavigationProp } from 'react-native-screens/native-stack';
+import {Alert, Button, Switch, Text, View} from 'react-native';
+import {NavigationContainer, ParamListBase} from '@react-navigation/native';
+import {
+  createNativeStackNavigator,
+  NativeStackNavigationProp,
+} from 'react-native-screens/native-stack';
 
 const Stack = createNativeStackNavigator();
 
 export default function App() {
-    const [gestureEnabled, setGestureEnable] = React.useState(false);
-    return (
-        <View style={{ flex: 1, paddingBottom: 200 }}>
-            <NavigationContainer>
-                <Stack.Navigator screenOptions={{ gestureEnabled }}>
-                    <Stack.Screen name="Top" component={First} options={{ replaceAnimation: 'push' }} />
-                    <Stack.Screen name="Top1" component={Second} options={{ replaceAnimation: 'push' }} />
-                </Stack.Navigator>
-            </NavigationContainer >
-            <View>
-                <Text>Gesture enabled</Text>
-                <Switch value={gestureEnabled} onValueChange={setGestureEnable} />
-            </View>
-        </View>
-    );
+  const [gestureEnabled, setGestureEnable] = React.useState(false);
+  return (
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: TestsExample/src/Test1032.tsx [hunk 1]
```diff
@@ -0,0 +1,47 @@
...
+
+const Stack = createNativeStackNavigator();
+
+export default function App() {
+    const [gestureEnabled, setGestureEnable] = React.useState(false);
+    return (
+        <View style={{ flex: 1, paddingBottom: 200 }}>
+            <NavigationContainer>
+                <Stack.Navigator screenOptions={{ gestureEnabled }}>
+                    <Stack.Screen name="Top" component={First} options={{ replaceAnimation: 'push' }} />
```
WoLewicki, line 14: `options={{ replaceAnimation: 'push' }}` is this option needed for something?

TestsExample/src/Test1032.tsx
```diff
@@ -0,0 +1,47 @@
...
+            <Button title="Tap me for second screen" onPress={() => navigation.push('Top1')} />
+        </View>
+
+    );
+}
+
+function Second() {
+    return (
+        <View style={{ backgroundColor: '#FFF' }}>
+            <Button title="Tap me to see test" onPress={() => Alert.alert('Test')} />
```
WoLewicki, line 42: Maybe better something like `Swipe back on me to see if button triggers`

TestsExample/src/Test1032.tsx
```diff
@@ -0,0 +1,47 @@
...
+        <View style={{ flex: 1, paddingBottom: 200 }}>
+            <NavigationContainer>
+                <Stack.Navigator screenOptions={{ gestureEnabled }}>
+                    <Stack.Screen name="Top" component={First} options={{ replaceAnimation: 'push' }} />
+                    <Stack.Screen name="Top1" component={Second} options={{ replaceAnimation: 'push' }} />
+                </Stack.Navigator>
+            </NavigationContainer >
+            <View>
+                <Text>Gesture enabled</Text>
+                <Switch value={gestureEnabled} onValueChange={setGestureEnable} />
```
WoLewicki, line 20: You can switch `gestureEnabled` directly in `Second` by adding such switch there and using `navigation.setOptions`.

ios/RNSScreenStack.m
```diff
@@ -526,7 +526,7 @@ - (void)dismissOnReload
   return nil;
 }
 
-- (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
+- (void)cancelTouchHandlerIfExists
```
WoLewicki, line 529: I think `cancelTouchesInParent` fits better according to the comment.

ios/RNSScreenStack.m
```diff
@@ -556,13 +560,16 @@ - (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
                           ((RNSGestureRecognizer *)gestureRecognizer).edges == UIRectEdgeRight) ||
         (_controller.view.semanticContentAttribute != UISemanticContentAttributeForceRightToLeft &&
          ((RNSGestureRecognizer *)gestureRecognizer).edges == UIRectEdgeLeft);
-    if (isCorrectEdge) {
-      return topScreen.stackAnimation == RNSScreenStackAnimationSimplePush;
+    if (isCorrectEdge && topScreen.stackAnimation == RNSScreenStackAnimationSimplePush) {
+      [self cancelTouchHandlerIfExists];
+      return YES;
     }
     return NO;
```
WoLewicki, line 567: this `return NO;` seems unnecessary here.

WoLewicki (approved):
Looks good! I left some comments, please answer them and I think we can merge it then.

TestsExample/src/Test1032.tsx
Ubax: No. Actually it is a leftover

TestsExample/src/Test1032.tsx
Ubax: I think that being able to change it on both screens is useful, so I would keep it outside

kacperkapusciak (approved):
LGTM 👏

PR #1035, opened 2021-08-03
feat: configuring native-stack transitions from JS

WoLewicki:
## Description

Added an option to provide custom transitions between screens from JS. It requires setting `stackAnimation: 'custom'` and providing `animationSpec` on screen.

## Changes

Added `stackAnimation: 'custom'` and `animationSpec` prop to `Screen` component.

`Test1035.tsx`

## Checklist

- [x] Updated TS types
- [ ] Updated documentation: 
- [x] Ensured that CI passes


Hirbod:
This would be a great feature to land!

sbearben:
Agree this would be awesome - curious, was this PR mostly an experiment to gauge feasibility?

WoLewicki:
I think we will not go this way with custom transitions. After we release SET(https://github.com/software-mansion/react-native-screens/pull/1591), I think we can extend the API to be able to animate whole screens with it so we'll get the benefit of all preconfigured animations etc.

PR #1041, opened 2021-08-09
fix: animation when replacing ios modal

WoLewicki:
## Description

PR changing the order of updating the VCs of `native-stack` in order to first dismiss the modal and then push/pop normal screens.

### Before

https://user-images.githubusercontent.com/32481228/128707566-317ccc1a-b324-4bd0-9d1b-09b5c09e0fed.mp4

### After

https://user-images.githubusercontent.com/32481228/128707575-dd6f1e67-89bb-4077-965c-00d654d8a3d0.mp4

`Test1041.tsx`

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/TestsExample/App.js
+++ b/TestsExample/App.js
@@ -49,6 +49,7 @@ import Test898 from './src/Test898';
 import Test913 from './src/Test913';
 import Test999 from './src/Test999';
 import Test1032 from './src/Test1032';
+import Test1041 from './src/Test1041';
 
 export default function App() {
   return <Test42 />;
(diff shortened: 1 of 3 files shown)


tboba:
@WoLewicki why do we want to have such behaviour with modals? In my opinion, views behind the modals should navigate immediately to the next screen, without waiting for the end of the modal animation.
I can't find quite good reproducer for the native app with the modal that navigate to another screen, to check how native apps behave normally though. Do you know some of the iOS native apps with such interaction?
What do you think?

WoLewicki:
I don't quite remember why we needed this so we can close it. But on the other hand it seems legit that first the modal closes and then we navigate so the user sees all of the actions that are taking place.

PR #1055, opened 2021-08-17
chore: Adds build checks 

Ubax:
## Description

Adds build checks for iOS, tvOS and android.

Checks should pass in this PR.
They should fail in #1056. 

## Checklist

- [ ] Ensured that CI passes


.github/workflows/android-build-test.yml
```diff
@@ -0,0 +1,21 @@
...
+      - name: Use Node.js 14
+        uses: actions/setup-node@v2
+        with:
+          node-version: 14
+      - name: Install dependencies
+        run: yarn && cd TestsExample && yarn
+      - name: run tests
+        uses: reactivecircus/android-emulator-runner@v2
+        with:
+          api-level: 29
```
WoLewicki, line 20: I am not sure about this one. We support Android versions back to `16` API level, so maybe we should test against the lowest available version? wdyt @kacperkapusciak ?

Ubax: I just copy pasted this value. I'll change it to 16

Ubax: I also changed `minSdkVersion` of TestsExample to 16 (previously 21)

Ubax: I changed it to 21

.github/workflows/android-build-test.yml
```diff
@@ -0,0 +1,19 @@
...
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: Use Node.js 14
+        uses: actions/setup-node@v2
+        with:
+          node-version: 14
+          cache: 'yarn'
+      - name: Install dependencies
+        run: yarn && cd TestsExample && yarn
```
kacperkapusciak, line 17: Why is `yarn` run only here in the main dir of `react-native-screens` and omitted in other builds? Are we certain that this job will run first and packages will be cached for others? It so I'd still split it to separate steps.

.github/workflows/tv-os-build-test.yml
```diff
@@ -0,0 +1,22 @@
...
+        with:
+          node-version: 14
+          cache: 'yarn'
+          cache-dependency-path: 'TVOSExample/yarn.lock'
+      - name: Install node dependencies
+        run: cd TVOSExample && yarn
+      - name: Install pods
+        run: cd TVOSExample/ios && pod install
+      - name: Build app
+        run: cd TVOSExample && yarn tv-os
```
kacperkapusciak, line 22: Use `working-directory` rather than changing directory on every step:
https://stackoverflow.com/a/58142276/9999202
Same for other builds

.github/workflows/android-build-test.yml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: Use Node.js 14
+        uses: actions/setup-node@v2
+        with:
+          node-version: 14
+          cache: 'yarn'
+      - name: Install dependencies
```
kacperkapusciak, line 16: In iOS and TvOS builds `Install node dependencies` name was used. Both of them are fine but stick to one

kacperkapusciak (changes requested):
Great job! 🎉  Left some small nit-picks to fix 😃

.github/workflows/android-build-test.yml
Ubax: It was there because before I was also running emulator and the app crashed without root dependencies. 
I decided that it is not necessary to run it (without emulator job takes 3 minutes instead of 5 and can be ran on ubuntu).

Ubax: I removed the unnecessary `yarn`

kacperkapusciak (approved):

henninghall:
Nice one! 🥳

PR #1062, opened 2021-08-19
fix: iOS screen orientation crash

henninghall:
## Description

[This change](https://github.com/software-mansion/react-native-screens/pull/863/files#diff-e988f8ab84e7063d6cc9ad4e634806389882ce6727b5c95f3cfbc9f9d3921a67L365-R399) in PR https://github.com/software-mansion/react-native-screens/pull/863 introduces a crash in some situtations. 

In our case, we are not providing an orientation to the "react-native-screen"-screen and instead we are changing the orientation manually. There are probably several situations where you want to be able to do that.

## Changes

Restored the functionality by returning `UIInterfaceOrientationMaskAllButUpsideDown` in cases where orientation is not provided to the screen.

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.m
+++ b/ios/RNSScreen.m
@@ -405,7 +405,8 @@
   UIViewController *vc = [self findChildVCForConfigAndTrait:RNSWindowTraitOrientation includingModals:YES];
 
   if ([vc isKindOfClass:[RNSScreen class]]) {
-    return ((RNSScreenView *)vc.view).screenOrientation;
+    UIInterfaceOrientationMask orientation = ((RNSScreenView *)vc.view).screenOrientation;
+    return orientation ? orientation : UIInterfaceOrientationMaskAllButUpsideDown;
   }
   return UIInterfaceOrientationMaskAllButUpsideDown;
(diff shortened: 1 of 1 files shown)


WoLewicki:
Can you provide a repo with reproduction of the crash? Isn't the default value of the orientation `UIInterfaceOrientationMaskAllButUpsideDown`? Also, if you do not provide `screenOrientation` on any of the screens, I think it shouldn't ever enter that `if` clause, so the reproduction would be even more valuable.

tboba:
Hi @henninghall, apologies for quite late response. I believe this change is quite outdated, since in `RNSScreen.mm` there is such change, but in other form 😅 

https://github.com/software-mansion/react-native-screens/blob/main/ios/RNSScreen.mm#L1327

Therefore, I'll close that issue.
Thanks!

PR #1072, opened 2021-08-24
feat: add full width swipe option

WoLewicki:
## Description

PR adding an option on iOS to begin the swipe gesture in any place on the screen. Since it comes with a custom gesture recognizer, the transition animation must be provided, therefore `simple_push` animation has been chosen as the closest one to the default. Requested in #251.

**Be aware that using this option with `gestureEnabled` set to `true` will cancel JS recognizers on the screen when using vertical or horizontal swipe gesture.**

## Changes

Added `fullWidthGestureEnabled` prop with the described behavior.

`Test1072.tsx`

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
- [ ] Ensured that CI passes


ios/RNSScreenStack.m
```diff
@@ -553,23 +562,37 @@ - (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
   [self cancelTouchesInParent];
   return YES;
 #else
-  // custom animation should be always served by custom recognizers
+  if (topScreen.fullWidthGestureEnabled) {
+    // we want only `RNSPanGestureRecognizer` to be able to recognize when
+    // `fullWidthGestureEnabled` is set and only then
```
kmagiera, line 567: rephrase

ios/RNSScreenStackAnimator.m
```diff
@@ -50,20 +50,23 @@ - (void)animateTransition:(id<UIViewControllerContextTransitioning>)transitionCo
...
+      // we are swiping with full width gesture
+      if (screen.customAnimationOnSwipe) {
+        [self animateTransitionWithStackAnimation:screen.stackAnimation
+                                transitionContext:transitionContext
+                                             toVC:toViewController
+                                           fromVC:fromViewController];
+      } else {
+        [self animateSimplePushWithTransitionContext:transitionContext toVC:toViewController fromVC:fromViewController];
+      }
+      return;
```
kmagiera, line 63: I personally try to avoid early returns if there exists a more logical structure of the conditions. It appears like it'd suffice to put the remainig code in `else` block here and that'd result in the same behavior w/o a need of using `return` here

kmagiera (approved):
Looks good, left a couple minor comments inline.

ios/RNSScreenStack.m
WoLewicki: `and only then` is not necessary since there is a comment about it already: https://github.com/software-mansion/react-native-screens/pull/1072/files#diff-1c3e13a1b0dcbaa5f7cff5f3ce12cc915fb509725faf693fb65cd23ceb49c4a9R595, so I will remove it.

kacperkapusciak (approved):

timurridjanovic:
Is this supposed to work in @react-navigation/native-stack?
Can't seem to make it work

WoLewicki:
I think that since https://github.com/react-navigation/react-navigation/blob/main/packages/native-stack/src/views/NativeStackView.native.tsx#L176 does not spread all of the props, all of the new ones need to be added there explicitly, so we need new version of `@react-navigation/native-stack` in order for new props from `react-native-screens` to work. We would love to accept a PR with such change :tada: You can ping me and/or @kacperkapusciak  in it for faster review.

ettnn:
Any updates on this @WoLewicki ?

WoLewicki:
Are you asking about v6 version of this? If so, I am afraid there is still no PR with this option there.

kyang6:
I'd love to help out, is the PR simply adding fullWidthGestureEnabled as a prop to NativeStackView.native.tsx and then passing it down into Screen?

WoLewicki:
@kyang6 I believe there should not be much that needs to be done to make it work.

kyang6:
Gotcha! Quick question, has fullWidthGestureEnabled been added to ScreenProps?

PR #1096, opened 2021-09-01
feat: move `react-native-window-view` to screens package

WoLewicki:
## Description

Moved the `WindowView` from https://github.com/WoLewicki/react-native-window-view to this repo and renamed it to `RNSFullWindowOverlay`.

## Changes

Added `RNSFullWindowOverlay` files to JS and iOS files. Removed the necessity of adding changes to `AppDelegate.m` by providing an extension of `UIWindow` in the package.

`Test1096.tsx`

## Checklist

- [x] Updated TS types
- [ ] Updated documentation: 
- [x] Ensured that CI passes


nandorojo:
@WoLewicki thank you for the great work on this. I think this feature could be a huge addition to React Native iOS in general.

I have a few questions:
1. Can you use multiple `WindowView` components? If so, will they work like the `<Modal />`, where the ones rendered last will go on top?
2. Does it matter _where_ the `WindowView` is rendered? For example, rather than putting it at the root of my app, can I treat it like a `Modal` and put it anywhere?

A big use-case I'm thinking about here is creating custom components like `Popover`s, `Menu`s, `Alert`s, etc. On iOS, this has always been a challenge: the only solution with a good DX is to use the native `Modal`, but on iOS, you can't open more than one `Modal` at a time. So I'm wondering if this component could finally be the solution.

Thanks again!

WoLewicki:
Ad 1. as you can see here: https://github.com/software-mansion/react-native-screens/pull/1096/files#diff-174f03b4ee9b6d6719c80e57d200b2057b704709f1b3f83218e31b54819d22d4R6 the `RNSOverlayViewContainers` are brung to front in the order in which they were added, so yeah, it should work like that.

Ad 2. as long as the view is not destroyed by React, it should stay in the native view hierarchy, so I think you can think of it as kind of similar thing to `Modal`.

Probably the best thing you can do is to test it by yourself :tada: and see if there are any problems.

nandorojo:
This PR might be an underrated solution to many of my iOS issues...thank you so much @WoLewicki! I will definitely test it.

A bit of a naïve question here, what is the best way to test this PR in an app?

nandorojo:
One other question, related to the previous conversation about `pointerEvents="box-none"`. Is this behavior possible?

```tsx
<>
  <App />
  <OverlayView interceptTouches={false} shown>
    <Notifications /> 👈 this should still intercept touches when it renders
  </OverlayView>
</>
```

Ideally, its children can intercept touches, even if the overlay itself doesn't.

WoLewicki:
One way of testing is adding this package to your project in `package.json` straight from this commit, something like this: "react-native-screens": "github:software-mansion/react-native-screens#ede07fb". Another way is to clone the repo, checkout to this branch and apply changes inside e.g. `TestsExample` project to test your use-cases.

As for the behavior you mentioned, I believe it is not possible now due to simple logic enabling/disabling the whole view hierarchy of `OverlayView`, but it could be changed with some effort for sure and I will try to do it when I get some free time.

WoLewicki:
@nandorojo can you check the newest commit? It should make the whole view not intercept touches, but its children should intercept them now. I added the behavior in the example, where the `OverlayView` takes the whole screen and you can click on views below it if there is none of its child there. If it works and does not destroy any of `react-native's` behaviors, I think we can merge it. It would then probably make it to next Expo SDK.

nandorojo:
Looks good! I'll test it out. Would be great for it to make SDK 43.

arberK:
@WoLewicki  I tested this. Works very good for me. A much-needed feature

src/index.native.tsx
```diff
@@ -329,6 +337,14 @@ module.exports = {
 
     return ScreensNativeModules.NativeSearchBar;
   },
+  get OverlayView() {
```
kmagiera, line 340: I'd perhaps change the name to sth like "FullWindowOverlay" to make it cleaner that it does not just overlay the container where its put or the screen under which it is rendered but instead it displays in front of all the content in the window.

TestsExample/src/Test1096.tsx
```diff
@@ -0,0 +1,105 @@
...
+            title="dismiss rn modal"
+            onPress={() => setIsShowModal(false)}
+          />
+          <Button
+            title="Show/hide window view"
+            onPress={() => setShown(!shown)}
+          />
+        </View>
+      </Modal>
+      <OverlayView style={{position: 'absolute', width: '100%', height: '100%', justifyContent: 'center' }} shown={shown} draggable={draggable} interceptTouches={interceptTouches}>
```
kmagiera, line 53: What happens if there are more than one of such components rendered in the tree?

kmagiera (commented):
Left two comments inline. Also, I think it'd be good to cleanup draggable/shown parameters as we discussed.

TestsExample/src/Test1096.tsx
WoLewicki: They will be rendered on top of each other in the same order as they were added to `React` view hierarchy at initialization since this code will be called: https://github.com/software-mansion/react-native-screens/pull/1096/files#diff-4882eafb07e471e6bd8bf6558dca65eacab42001de16e1dc1ca7b23c54ab765fR108 and then the `RNWindow` will move all containers to front in order they were added: https://github.com/software-mansion/react-native-screens/pull/1096/files#diff-174f03b4ee9b6d6719c80e57d200b2057b704709f1b3f83218e31b54819d22d4R11

kmagiera (approved):
Looks fine for the first iteration. Looking forward to getting full platform compatibility for this one!

nandorojo:
> Looking forward to getting full platform compatibility for this one!

Me too, this one is big time!

src/index.native.tsx
nandorojo: `WindowOverlay` could work too.

jakob-p:
When i try to use this with expo im getting 
image

kacperkapusciak:
@jakob-p what Expo SDK version are you using?

Expo SDK 42 comes with `react-native-screens` `v3.4.0`. 

FullWindowsOverlay is available from [version >=3.7.0 ](https://github.com/software-mansion/react-native-screens/releases/tag/3.7.0) of `react-native-screens` which will be available in [Expo SDK 43](https://github.com/expo/expo/blob/master/CHANGELOG.md#-3rd-party-library-updates-1).



Cheers

jakob-p:
Im using Expo SDK 42, but tried to upgrade react-native-screens manually to 3.8.0. 

I'll try it with Expo SDK 43 when its officially released.

mfazekas:
FWIW I've added an android POC for FullWindowOverlay - see #2421

PR #1103, opened 2021-09-03
chore: add windows CI

kacperkapusciak:
## Description

This PR adds build check for Windows platform for `react-native-screens`'s Example app. PR based on CI used in [`react-native-webview` repo](https://github.com/react-native-webview/react-native-webview).

## Changes

- Bootstrapped `react-native-windows` in Example project
- Added `.github\workflows\windows-build-test.yml`

## Checklist

- [x] Ensured that CI passes


dannyvv:
Hi @kacperkapusciak, I took a peek at the failure at the suggestion of @asklar 

I believe the following is happening:
* This repo has two package.json files:
  * `D:\a\react-native-screens\react-native-screens\package.json`
     * This declares react-native@0.63.2
  * `D:\a\react-native-screens\react-native-screens\Example\package.json`
     *  This declares a dependency on react-native@0.65.2
* The pipeline is restoring both package.json files
* React-Native-Windows msbuild logic is written to assume it is running in a 'yarn workspace' where the outer package.json as a `workspace` field and aggregates all packages up to the highest folder. 
* react-native-windows uses the msbuild feature:
   `<ReactNativePackageDir Condition="'$(ReactNativePackageDir)' == ''">$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), 'node_modules\react-native\package.json'))\node_modules\react-native\</ReactNativePackageDir>`
   to resolve where the react-native Npm Package is... 
* This seems to be finding the package at `D:\a\react-native-screens\react-native-screens\node_modules\react-native` instead of `D:\a\react-native-screens\react-native-screens\Example\node_modules\react-native` 
* So react-native-windows is trying to load TurboCxxModule.cpp from the 0.63.2 version of react-native but the code layout or react-native has changed since, so the expected layout of 0.65 does not match

Possible solutions:
* Do not run `yarn install` at the root of the repo
* Bump react-native at the root of the repo to 0.65

If these don't work we'll need some more diagnostics:
* Add a step before the build step: `dir /s/b` to print the directory structure
* Create a pipeline artifact for `Example\*.binlog` so we can get better diagnostics

kacperkapusciak:
Hi @dannyvv,
Thank you so so much for the explanation and for caring!
I've figured it out that the script tries to use files that aren't present in 0.63.2 but I couldn't understand why it happens but it makes sense now.

I'll try to fix the CI with your advice and I'll update you on how it went.

Thanks! 🙌

asklar:
thanks for taking a look @dannyvv : )
the new error is in trying to find the RNW root dir, this is the error:
`This project references targets in your node_modules\react-native-windows folder. The missing file is \node_modules\react-native-windows\\PropertySheets\External\Microsoft.ReactNative.Uwp.CppLib.props`

note that it resolved the root dir to the empty string so it's tryin to find stuff under \node_modules (i.e. the node_modules directory at the root of the drive).

This comes from this line of msbuild:
https://github.com/software-mansion/react-native-screens/blob/master/windows/RNScreens/RNScreens.vcxproj#L21

`<ReactNativeWindowsDir Condition="'$(ReactNativeWindowsDir)' == ''">$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), 'node_modules\react-native-windows\package.json'))\node_modules\react-native-windows\</ReactNativeWindowsDir>`

which tries to find a directory with a file named `node_modules\react-native-windows\package.json` starting from the vcxproj's directory, and retrying upwards

kacperkapusciak:
Hi @dannyvv,
because the proposed solutions didn't help I've added the printing of the directory structure step and pipeline artifact to gather logs as you've asked.

Hi @asklar,
I can't wrap my head around how to overcome this without breaking the package for other developers. Any propositions?

asklar:
> Hi @dannyvv Danny van Velzen FTE, because the proposed solutions didn't help I've added the printing of the directory structure step and pipeline artifact to gather logs as you've asked.
> 
> Hi @asklar Alexander Sklar FTE, I can't wrap my head around how to overcome this without breaking the package for other developers. Any propositions?

tagging @chiaramooney to take a look. I _think_ the repo layout would benefit from being set up as a yarn workspace

acoates-ms:
You could add a `Directory.Build.props` file at the root of the repo.  Just make sure it doesn't get included in the published npm package.

```xml
<?xml version="1.0" encoding="utf-8"?>
<Project>

  <!-- 
    Due to the mono-repo setup in this repo, we should ensure that when building the example project the projects
    are pointing to the same place for react-native and react-native-windows.
    -->

  <PropertyGroup>
    <ReactNativeWindowsDir>$(MSBuildThisFileDirectory)\Example\node_modules\react-native-windows\</ReactNativeWindowsDir>
    <ReactNativePackageDir>$(MSBuildThisFileDirectory)\Example\node_modules\react-native\</ReactNativePackageDir>
  </PropertyGroup>

</Project>
```

PR #1116, opened 2021-09-09
fix: delay ScreenContainer updates in same manner as ScreenStack

WoLewicki:
## Description

Added delayed update logic to `ScreenContainer` in order to resolve issues with too early detached nested navigators when going back from a screen in `native-stack` with nested different navigators. This change was part of synchronous fragment updates introduced in #1066. Going back like that triggers `removeAllScreens` code of the nested navigator and it resolved in detaching all of the child screens before the update of `native-stack` was dispatched, leading to blank screen from the start of navigation.

`Test860.tsx` to see that without this change the nested navigators are detached at the start of transition.

## Checklist

- [x] Ensured that CI passes


kacperkapusciak (approved):

kmagiera (approved):
Looks good 👍

PR #1133, opened 2021-09-16
fix: commit remove and add separately in moveToFront

WoLewicki:
## Description

PR fixing `stack` navigator's `detachPreviousScreen` behavior with new versions of `appCompat`. Seems like one of the new implementations of `Fragment` added some kind of optimization of redundant operations. Therefore, `moveToFront` method does not work correctly there, since we remove and then reattach the same fragment in order to make it appear in front. Committing the `remove` and `add` action separately brings back the correct behavior, probably due to no option of optimization because of synchronous behavior of these actions.

## Changes

Separated `remove` and `add` actions in `moveToFront` method in `ScreenContainer.kt`.

`Test624.tsx` modal presentation screens with `appCompat` version `1.3.1`.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt
@@ -181,7 +181,8 @@ open class ScreenContainer<T : ScreenFragment>(context: Context?) : ViewGroup(co
     }
 
     private fun moveToFront(screenFragment: ScreenFragment) {
-        createTransaction().remove(screenFragment).add(id, screenFragment).commitNowAllowingStateLoss()
+        createTransaction().remove(screenFragment).commitNowAllowingStateLoss()
+        createTransaction().add(id, screenFragment).commitNowAllowingStateLoss()
     }
 
     private fun detachScreen(screenFragment: ScreenFragment) {


cristianoccazinsp:
@WoLewicki this does seem to fix the screen overlap issue, but it introduces a new (or two?) one.

When popping screens, there's a tiny flash of the screen now, did you see the same behaviour in your testing?


https://user-images.githubusercontent.com/48869228/133615782-4caf8cd0-9e27-492b-86ca-a4294377641d.mp4


Secondly, it brings back the issue I was observing before with appcompat 1.1 and 1.2, where surface/texture views reset right before popping, causing the camera review of react native camera to stop and start during the transition (this is likely to occur to any other libraries that use surface views, like react native video, map views, etc.)

https://user-images.githubusercontent.com/48869228/133615996-3acdecca-3910-4ef9-8e58-4089a1f33f0e.mp4

Interesting enough, if we pop the screen during this slow test animation, the surface issue does not happen.

WoLewicki:
Yeah, I spotted the flickering after some testing, which is a result of none of the screens being attached for one frame. This seems weird since both `remove` and then `add` is done on `UI` thread. It may be connected to how the `Fragments` work under the hood 😕  It is sad that fixing one behavior leads to destroying other ones.
Also, it seems that when having `appCompat` `1.3.1`, there is no animation of going back when not using `detachPreviousScreen: false` since `moveToFront` does not work there. Can you confirm that?

cristianoccazinsp:
Not sure what you mean by no animation when going back. The two videos above are always animated when popping, aren't they? The screen behind animates left to right.

Could it be just an issue somewhere with the ordering of the fragments in the stack? It really seems like it is just a matter of the non-detached screen staying behind the other fragment as opposed to on top.

Also, to remove the flickering, you can just use `createTransaction().hide` and  `createTransaction().show` instead of removing and adding the fragments, but that still leaves you with the overlapped screen.

WoLewicki:
> Not sure what you mean by no animation when going back. The two videos above are always animated when popping, aren't they? The screen behind animates left to right.

I tried some simple example, e.g. this: https://github.com/software-mansion/react-native-screens/blob/master/TestsExample/src/Test111.js, and the second screen disappears when being removed because it is rendered behind the first one. When providing `detachPreviousScreen: false`, it works correctly.


> Could it be just an issue somewhere with the ordering of the fragments in the stack? It really seems like it is just a matter of the non-detached screen staying behind the other fragment as opposed to on top.
> 
> Also, to remove the flickering, you can just use `createTransaction().hide` and `createTransaction().show` instead of removing and adding the fragments, but that still leaves you with the overlapped screen.

Yeah, I tried `hide` and `show`, but it does not change the ordering of rendering of fragments unfortunately. This ordering needs to be changed and is done by removing and adding the fragment since it pushes the fragment to the front.

cristianoccazinsp:
I see what you mean now with the example.

cristianoccazinsp:
@WoLewicki I've been trying to debug this (while also trying to understand how all these fragments work). 

Basically, we have screen A, B (detach=true), and C configured.  Default = A, push B, push C, then pop C. Native code does something as:
- attachScreen (B) (first push)
- detachScreen (A) (second push)
- attachScreen (C) (second push)
- detachScreen (B) (second push)
- -- we now pop C, and ultimately end up with A in front of B, instead of the other way around --
- attachScreen (B)
- moveToFront (C)
- detachScreen (C)
- attachScreen (A)
- moveToFront (B)


I think the order of the pop operations are a bit odd. Perhaps it's just a re ordering issue of the fragments that now became more obvious with the improved transactions. I will try to mess around with these operations to see if there's a way to make them behave correctly. Do you see anything else that may be causing this problem?

cristianoccazinsp:
Tested a bit more and forced `mFragmentManager` to respect the fragment order so the fragment behind doesn't end in the top of the stack, and the issue still happens, so something else has to be setting the zindex/ordering of the fragment behind the top one.

cristianoccazinsp:
@WoLewicki I think I solved the issue, would you please test something like the following? Very rough, but it fixed all of my issues:

- Odd animation behaviour when popping (like the issue from your example)
- Issues with `detachPreviousScreen`
- Camera view / surface no longer re-creates when popping


In short, fix the "move to front" logic, and ensure everything runs in the same transaction
```
var addedBefore = false
var trans = createTransaction()
var pendingFront: ArrayList<T> = ArrayList<T>()

for (screenFragment in mScreenFragments) {
    val activityState = getActivityState(screenFragment)
    if (activityState !== ActivityState.INACTIVE && !screenFragment.isAdded) {
        addedBefore = true
        //attachScreen(screenFragment)
        trans.add(id, screenFragment)
    } else if (activityState !== ActivityState.INACTIVE && addedBefore) {
        trans.remove(screenFragment)
        pendingFront.add(screenFragment)
    }
    screenFragment.screen.setTransitioning(transitioning)
}

for (screenFragment in pendingFront) {
    trans.add(id, screenFragment)
}

trans.commitNowAllowingStateLoss()
```

Please do some testing and let me know if this can be improved.

WoLewicki:
I changed the code as you suggested so all of the transactions are committed at once and I haven't spotted any new issues. This needs to be tested thoroughly and we could probably merge it then :tada: Thanks for deep investigation!

cristianoccazinsp:
@WoLewicki thanks for improving the code!

Any thoughts on the second addition that I see you've also included, basically to fix the bring to front issue? Do you think that removing and adding the fragment that way may cause issues? It's nearly the same behaviour as before, but it ensures that the fragment order is preserved. I borrowed the idea from the iOS counterpart that does something similar.

Did it also fix the issue for the test file you've spotted earlier (Test111.js)?  I will do some more testing with the test app and my own app that was having issues. I do see there's a slight difference in the code where the orphaned fragments are now also removed in the same transaction, which should be fine as well.

Lastly, do you think these changes may be related to https://github.com/software-mansion/react-native-screens/issues/1138 and/or https://github.com/software-mansion/react-native-screens/issues/1136 ?

cristianoccazinsp:
@WoLewicki how's the testing for this? Any insights whether or not it will be merged and released soon?

WoLewicki:
I cannot give an exact date, but after much testing, if everything stays OK, it will probably be merged. I would suggest you to use [patch-package](https://www.npmjs.com/package/patch-package) with these changes till that time since it is a fix for non-default options and therefore does not have big priority.

cristianoccazinsp:
@chungweileong94 any chance RN 0.66 upgraded some of the androidx dependencies? I believe this fixes the issue when not using the fragment version this library assumes is being used.

What do you get if you run `./gradlew app:dependencies |grep androidx.fragment:fragment` ? 1.2.x or 1.3.x ?

chungweileong94:
@WoLewicki I updated one of my [RN projects](https://github.com/chungweileong94/ReactNavigationLagDemo) to 0.66.1, and it happens as well.

So the project is using `react-navigation`, and so far the bug only exists if I use `TransitionPresets.SlideFromRightIOS`. Because the back navigation is too fast, so I enable the gesture for better demonstration in the video below

https://user-images.githubusercontent.com/15154097/138022581-004c76a3-6791-4ce6-bbe2-596914c5d52f.mov

chungweileong94:
@cristianoccazinsp It seems like it's using 1.3x, correct me if I'm wrong

```
> ./gradlew app:dependencies |grep androidx.fragment:fragment
|    |    +--- androidx.fragment:fragment:1.0.0
|    |    +--- androidx.fragment:fragment:1.3.4
+--- androidx.fragment:fragment:{strictly 1.3.4} -> 1.3.4 (c)
|    |    +--- androidx.fragment:fragment:1.3.4
|    |    +--- androidx.fragment:fragment:1.0.0 -> 1.3.4 (*)
     +--- androidx.fragment:fragment:1.2.1 -> 1.3.4 (*)
          +--- androidx.fragment:fragment:1.0.0 -> 1.3.4 (*)
               +--- androidx.fragment:fragment:1.1.0 -> 1.3.4 (*)
|    |    +--- androidx.fragment:fragment:1.3.4
+--- androidx.fragment:fragment:{strictly 1.3.4} -> 1.3.4 (c)
|    |    +--- androidx.fragment:fragment:1.3.4
|    |    +--- androidx.fragment:fragment:1.0.0 -> 1.3.4 (*)
     +--- androidx.fragment:fragment:1.2.1 -> 1.3.4 (*)
          +--- androidx.fragment:fragment:1.0.0 -> 1.3.4 (*)
               +--- androidx.fragment:fragment:1.1.0 -> 1.3.4 (*)
|    |    +--- androidx.fragment:fragment:1.1.0 -> 1.2.1
+--- androidx.fragment:fragment:{strictly 1.2.1} -> 1.2.1 (c)
|    |    +--- androidx.fragment:fragment:1.1.0 -> 1.2.1
|    |    +--- androidx.fragment:fragment:1.0.0 -> 1.2.1 (*)
     +--- androidx.fragment:fragment:1.2.1 (*)
          +--- androidx.fragment:fragment:1.0.0 -> 1.2.1 (*)
               +--- androidx.fragment:fragment:1.1.0 -> 1.2.1 (*)
|    |    +--- androidx.fragment:fragment:1.1.0 -> 1.2.1
+--- androidx.fragment:fragment:{strictly 1.2.1} -> 1.2.1 (c)
|    |    +--- androidx.fragment:fragment:1.1.0 -> 1.2.1
|    |    +--- androidx.fragment:fragment:1.0.0 -> 1.2.1 (*)
     +--- androidx.fragment:fragment:1.2.1 (*)
          +--- androidx.fragment:fragment:1.0.0 -> 1.2.1 (*)
               +--- androidx.fragment:fragment:1.1.0 -> 1.2.1 (*)
```

jumpinjammin:
Can someone explain to me why this fix even works? @WoLewicki 

I'm confused when we iterate over fragments but have a single boolean `addedBefore` that represents adding a previous fragment, but later we use that property to determine if we should detach any fragment here:

https://github.com/software-mansion/react-native-screens/pull/1133/files#diff-f1499e0f24ea7e4e2b14ebda8d78911fefedf76bd6112eeb6513ac2ae1c19bb7R331

Further, where are the separate commits of the transaction for removals and additions? The only one I see is at the end.

https://github.com/software-mansion/react-native-screens/pull/1133/files#diff-f1499e0f24ea7e4e2b14ebda8d78911fefedf76bd6112eeb6513ac2ae1c19bb7R341

WoLewicki:
@jumpinjammin the description of PR is not valid and the newest implementation came with this comment: https://github.com/software-mansion/react-native-screens/pull/1133/files#issuecomment-923943317. As for `addedBedore`, I believe it was done like this since we iterate from the deepest screens on the stack, so if we encounter a screen which should be added, all the screens above it should be added as well, but I may have forgotten something. Is there an issue regarding this PR?

PR #1139, opened 2021-09-20
chore: Adds screenshot artifact to iOS e2e

Ubax:
## Description

When iOS e2e test fail save screenshots as artifacts

## Checklist
- [ ] Ensured that CI passes

### DIFF
--- a/.github/workflows/ios-e2e-test.yml
+++ b/.github/workflows/ios-e2e-test.yml
@@ -45,5 +45,5 @@ jobs:
       - uses: actions/upload-artifact@v2
         if: ${{ failure() }}
         with:
-          name: android-fail-screen-shoots
+          name: ios-fail-screen-shoots
           path: ${{ env.WORKING_DIRECTORY }}/artifacts
--- a/.github/workflows/ios-e2e-test.yml
+++ b/.github/workflows/ios-e2e-test.yml
@@ -2,7 +2,7 @@ name: Test iOS e2e
 on:
   pull_request:
     paths:
-        - 'Example/**'
+      - 'Example/**'
   push:
     branches:
       - master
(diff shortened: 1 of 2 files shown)


.github/workflows/ios-e2e-test.yml
```diff
@@ -42,3 +42,8 @@ jobs:
       - name: Test app
         working-directory: ${{ env.WORKING_DIRECTORY }}
         run: yarn test-e2e-ios
+      - uses: actions/upload-artifact@v2
+        if: ${{ failure() }}
+        with:
+          name: android-fail-screen-shoots
```
WoLewicki, line 48: Why is there `android` if the PR mentions `iOS`?

Ubax: A copy paste typo 🤦

.github/workflows/ios-e2e-test.yml
```diff
@@ -2,7 +2,7 @@ name: Test iOS e2e
 on:
   pull_request:
     paths:
-        - 'Example/**'
+      - 'Example/**'
```
kacperkapusciak, line 5: A question that should've been to previous PRs but I couldn't do a review in time before merging - does this mean e2e tests are only run on PRs when the `Example/` project changes? Shouldn't it also be run when `ios/` and `android/` are being changed? cc @WoLewicki

WoLewicki: Yeah, for sure these tests should be run for such changes.

Ubax: It takes about 40 minutes for tests to run. I am not sure if we want them ran on every change in most PRs. They will always be run on the master after merging. Also one can ran them locally if they are not sure if something works

WoLewicki: Only master option seems enough for me.

kacperkapusciak: Okay, running e2e on every commit is overkill but if it's possible we could maybe run these tests 1 time a day (in the night) on a hanging PR? If it's not possible it's okay (not too sure if GitHub Actions supports these kind of tricks). Our master is not directly production so it should be fine as you've said.

Ubax: Unfortunately we can always do cron actions on master branch (or other specified in the workflow).

PR #1151, opened 2021-09-24
fix: add hasTwoStates iOS check

burakgormek:
## Description

PR adds iOS check for NativeScreenNavigationContainer.  Android trying to call RNSScreenNavigationContainer which is only had on iOS.
image

## Changes
Added iOS check before the NativeScreenNavigationContainerValue init. Actually line 235 calls NativeScreenNavigationContainer if navigation hasTwoStates but I combined both check in one place.

1- Add navigation patch https://github.com/react-navigation/react-navigation/pull/9772
2- Run Test619 on Android

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/src/index.native.tsx
+++ b/src/index.native.tsx
@@ -85,7 +85,9 @@ const ScreensNativeModules = {
   get NativeScreenNavigationContainer() {
     NativeScreenNavigationContainerValue =
       NativeScreenNavigationContainerValue ||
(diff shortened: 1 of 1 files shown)


WoLewicki (approved):
Thanks for the fix!

Issue #996, opened 2021-06-27
[native-stack] Focused search bar causes new screens to have incorrect header


nandorojo:
## Description

I'm using the iOS search bar to search artists. Works as expected.

When you click on an artist, it opens the `Artist` screen.

The `Artist` screen has `headerShown: false`, since I have a custom header on that screen.

```jsx
// this is the screen that search navigates to
<Screen
  name="Artist"
  component={ArtistScreen}
  options={{
    headerShown: false,
  }}
  getId={({ params }) => {
    return params?.slug
  }}
/>
```

However, if I open the artist screen **_while the search bar is focused_**, it adds a header to the `Artist` screen, and creates a glitchy experience when I go back.

### Screenshots

**Works**

When the search bar is _not focused_, and I click an artist, it works normally. It navigates to the `Artist` screen, and there's no header.

Video: https://www.loom.com/share/174403203eb941f9a5731ba62253ab11

**Doesn't work**

However, when the search bar _is focused_, it incorrectly adds a header to the `Artist` screen. Then, going back causes a weird jump.

Video: https://www.loom.com/share/8f8037145df649f880b891f52b099520

## Steps To Reproduce

1. Create a native stack screen with a search bar
2. Navigate to a screen that has `headerShown: false` while the search bar is focused.
3. Try going back, too

### Expected behavior

Having a `searchBar` should not add a header when opening a screen that has `headerShown: false`.

### Actual behavior

A focused `searchBar` causes screens opened to have a header. It also causes a layout glitch when you go back. I have `contentInsetAdjustmentBehavior="automatic"` on my `ScrollView` like I saw in the `TestExample` file.

## Snack or minimal code example

Expo Snack doesn't support 3.4.0. Let me know if you need me to make a full reproduction.

## Package versions

- React: 16.8.3
- React Native: 0.63.4
- React Native Screens: 3.4.0

nandorojo:
Setting `searchBar.hideNavigationBar: false` makes the header bug go away. However, going back still causes the layout shift if the search bar was focused before opening the artist screen.

Also, I'd prefer to have `hideNavigationBar: true`. But I'll leave it false for now as a workaround.

viet97:
https://user-images.githubusercontent.com/24263655/124250855-7793b400-db4f-11eb-98d1-c68d3964373c.mp4

i'm having same problem. Is it caused by createNativeStackNavigator() ?

Ubax:
[...]

PR #1153, opened 2021-09-27
fix: Adds dispatch for hiding header

Ubax:
## Description

Fixes issue #996. 

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.m
+++ b/ios/RNSScreen.m
@@ -564,6 +564,26 @@
     _shouldNotify = NO;
   }
 
+#if !TARGET_OS_TV
+  NSUInteger currentIndex = [self.navigationController.viewControllers indexOfObject:self];
+
+  if (@available(iOS 11.0, *) && currentIndex > 0) {
+    UINavigationItem *prevNavigationItem =
+        [self.navigationController.viewControllers objectAtIndex:currentIndex - 1].navigationItem;
+
+    BOOL wasSearchBarActiveActive = prevNavigationItem.searchController.active;
+    BOOL shouldHideHeader = ((RNSScreenStackHeaderConfig *)self.view.reactSubviews[0]).hide;
+    if (wasSearchBarActiveActive && shouldHideHeader) {
+      double delayInSeconds = 0;
+      dispatch_time_t popTime = dispatch_time(DISPATCH_TIME_NOW, (int64_t)(delayInSeconds * NSEC_PER_SEC));
+      dispatch_after(popTime, dispatch_get_main_queue(), ^(void) {
+        // We need to hide header beacuse if searchbar was active, then iOS would show deafult header for some reason
+        [self.navigationController setNavigationBarHidden:YES animated:NO];
+      });
+    }
+  }
+#endif
+
   // as per documentation of these methods
   _goingForward = [self isBeingPresented] || [self isMovingToParentViewController];
 
--- a/TestsExample/src/Issue996.tsx
+++ b/TestsExample/src/Issue996.tsx
@@ -1,0 +1,87 @@
+import * as React from 'react';
+import {Button, View} from 'react-native';
+import {NavigationContainer, ParamListBase} from '@react-navigation/native';
+import {
+  createNativeStackNavigator,
+  NativeStackNavigationProp,
+} from 'react-native-screens/native-stack';
+import {SafeAreaView} from 'react-native-safe-area-context';
+import { ScrollView } from 'react-native-gesture-handler';
+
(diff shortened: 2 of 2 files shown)


ios/RNSScreen.m
```diff
@@ -564,6 +564,26 @@ - (void)viewWillAppear:(BOOL)animated
     _shouldNotify = NO;
   }
 
+#if !TARGET_OS_TV
+  NSUInteger currentIndex = [self.navigationController.viewControllers indexOfObject:self];
+
+  if (@available(iOS 11.0, *) && currentIndex > 0) {
```
kacperkapusciak, line 570:
```suggestion
  if (@available(iOS 11.0, *)) {
    if (currentIndex > 0) {
```
I'd separate these statements because the Obj-c compiler gets confused when you combine `available` with logic operators (because it shouldn't be checked conditionally - it's >=iOS 11 or it's not) and warns with

 ```
 @available does not guard availability here; use if (@available) instead
 ```

ios/RNSScreen.m
```diff
@@ -564,6 +564,26 @@ - (void)viewWillAppear:(BOOL)animated
...
+    UINavigationItem *prevNavigationItem =
+        [self.navigationController.viewControllers objectAtIndex:currentIndex - 1].navigationItem;
+
+    BOOL wasSearchBarActiveActive = prevNavigationItem.searchController.active;
+    BOOL shouldHideHeader = ((RNSScreenStackHeaderConfig *)self.view.reactSubviews[0]).hide;
+    if (wasSearchBarActiveActive && shouldHideHeader) {
+      double delayInSeconds = 0;
+      dispatch_time_t popTime = dispatch_time(DISPATCH_TIME_NOW, (int64_t)(delayInSeconds * NSEC_PER_SEC));
```
kacperkapusciak, line 578:
```suggestion
      dispatch_time_t popTime = dispatch_time(DISPATCH_TIME_NOW, 0);
```
Reduced complexity

ios/RNSScreen.m
```diff
@@ -564,6 +564,29 @@ - (void)viewWillAppear:(BOOL)animated
     _shouldNotify = NO;
   }
 
+#if !TARGET_OS_TV
+  NSUInteger currentIndex = [self.navigationController.viewControllers indexOfObject:self];
```
WoLewicki, line 568: I think it should be inside the below `if` since before `iOS` 11 it will be just unused variable.

WoLewicki (commented):
I think it would be best to extract the new native code into a method and call it there. There is already much code in the `viewWillAppear` method. This PR should also be tested thoroughly in all kinds of navigation, like swiping, clicking back button, navigating from JS, since these options can return different values.

Ubax:
@WoLewicki I added videos before and after

ios/RNSScreen.m
```diff
@@ -575,6 +577,32 @@ - (void)viewWillAppear:(BOOL)animated
...
+#if !TARGET_OS_TV
+  if (@available(iOS 13.0, *)) {
+    NSUInteger currentIndex = [self.navigationController.viewControllers indexOfObject:self];
+
+    if (currentIndex > 0 && [self.view.reactSubviews[0] isKindOfClass:[RNSScreenStackHeaderConfig class]]) {
+      UINavigationItem *prevNavigationItem =
+          [self.navigationController.viewControllers objectAtIndex:currentIndex - 1].navigationItem;
+      RNSScreenStackHeaderConfig *config = ((RNSScreenStackHeaderConfig *)self.view.reactSubviews[0]);
+
+      BOOL wasSearchBarActiveActive = prevNavigationItem.searchController.active;
```
WoLewicki, line 591:
```suggestion
      BOOL wasSearchBarActive = prevNavigationItem.searchController.active;
```

remember to change it below in `if`

ios/RNSScreen.m
```diff
@@ -575,6 +577,32 @@ - (void)viewWillAppear:(BOOL)animated
...
+      BOOL shouldHideHeader = config.hide;
+
+      if (wasSearchBarActiveActive && shouldHideHeader) {
+        dispatch_time_t popTime = dispatch_time(DISPATCH_TIME_NOW, 0);
+        dispatch_after(popTime, dispatch_get_main_queue(), ^(void) {
+          // We need to hide header beacuse if searchbar was active, then iOS would show default header for some reason
```
WoLewicki, line 597: Could you add a short comment why is it done like that and not immediately?

WoLewicki (approved):

kacperkapusciak (approved):

PR #1155, opened 2021-09-28
fix: make method for checking if trait is set

WoLewicki:
## Description

PR exposing a way to check if there was a window trait (status bar or orientation) set by user in any of the navigators. In such case, it should take precedence over other methods, like `expo-screen-orientation`. Calling 
```
+ (BOOL)shouldAskScreensForTrait:(RNSWindowTrait)trait
                 includingModals:(BOOL)includingModals
                inViewController:(UIViewController *)vc;
```

should return `NO` if there was no screen in view hierarchy that requests a specific trait and `YES` otherwise.

PR to be used e.g. in `Expo`.

***These methods are used by providing their exact signature as a string and therefore their name should not be changed.***

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.h
+++ b/ios/RNSScreen.h
@@ -65,6 +65,7 @@ typedef NS_ENUM(NSInteger, RNSWindowTrait) {
 
 - (instancetype)initWithView:(UIView *)view;
 - (void)notifyFinishTransitioning;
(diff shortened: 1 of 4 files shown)


kacperkapusciak (approved):

zwh1666258377:
hi~ When is the next release coming @kacperkapusciak

PR #1190, opened 2021-10-26
fix: make stack animations work same on both platforms

WoLewicki:
## Description

PR properly resolving if there should be an animation of transition between screens on both platforms. The logic is that if we perform `push` transition, we ask the screen we are transitioning to, and if we do `pop` transition, we ask the disappearing screen. It works the same for `replace` animation, so if the screen we are transitioning to specifies `replaceAnimation: 'push'`, we push it and ask it for animation, and if it specifies `replaceAnimation: 'pop'`, we ask the screen which is being replaced.

## Checklist

- [ ] Ensured that CI passes


kacperkapusciak (approved):
LGTM

PR #1191, opened 2021-10-28
fix: overlay in fullscreen modal

WoLewicki:
## Description

When the `fullScreenModal` is presented, the `FullWindowOverlay` is removed from native view hierarchy since the view receives `didMoveToWindow` with `nil` value. This behavior can be removed, but then the only way to remove it will be to invalidate it, which means removing the component on the react side. Such change would result e.g. in view not disappearing when moving between screens of navigators since these screens are not removed from the React view hierarchy.

`Test1096.tsx` in `TestsExample` app.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/TestsExample/src/Test1096.tsx
+++ b/TestsExample/src/Test1096.tsx
@@ -22,6 +22,10 @@ function Home({
         title="Go to RNScreens modal"
         onPress={() => navigation.navigate('Modal')}
       />
+      <Button
+        title="Go to RNScreens fullScreenModal"
+        onPress={() => navigation.navigate('FullScreenModal')}
+      />
       <Modal
         animationType="slide"
         visible={isShowModal}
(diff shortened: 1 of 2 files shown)


gabrieldonadel:
Any updates on this?

WoLewicki:
This PR introduces some changes that are breaking the old behavior, so I am not sure if it will be merged. You can always use [patch-package](https://www.npmjs.com/package/patch-package) and apply those changes in your app if you need them.

Shaninnik:
This is a very important change, in my case `FullWindowOverlay` completely disappeared and never reappeared after using `react-native-image-crop-picker` that calls native `presentViewController` method. Took me a while to figure out what was going on. I am sure many other libs are using `presentViewController`... I've applied the patch but maybe this PR can be re considered? Or added as an optional prop on `FullWindowOverlay` to let developers choose which behavior they need. At the very least maybe some warning added to readme?

kkafar:
Highly likely this won't ever be a subject of further development. Can reopen if needed.

PR #1208, opened 2021-11-10
feat: don't set display: 'none' on frozen screens

### LINKED ISSUE #1207: Pressing back fast on iOS may show frozen scene
Thanks @software-mansion and @kmagiera for your stellar work on `react-freeze`. 

I’ve been knee-deep in `Suspense` this past week and a half [adding freeze support to the Navigation router](https://github.com/grahammendick/navigation/pull/547). I solved a thorny issue along the way that I think your implementation might suffer from so I’m getting in touch to try and help out.

## Showing a Frozen Scene
If you navigate from A to B to C then you end up with a stack A → B → C where scene A is frozen (I’m actually freezing both A and B). On iOS, if you then press back twice very fast you can end up showing A in a frozen state. The navigation happens on the native side immediately and A stays frozen until the JavaScript navigation catches up and unfreezes A. This might be delayed, for example, if the JavaScript thread is busy or A takes some time to render. 

The problem is that when A is frozen, React Native sets its display to 'none'. So you can’t afford to show a frozen scene to the user even for a short amount of time. I fixed this by removing the display property from the list of `validAttributes` for the native Scene component. The list of valid attributes is attached to a component’s ref so you can modify it inside a ref callback. That way, React Native won’t send the display property to native when Suspense tries to set it to 'none'.

```jsx
<NativeScreen
  ref={(ref) => {
    if (ref?.viewConfig?.validAttributes?.style) {
      ref.viewConfig.validAttributes.style = {
        ...ref.viewConfig.validAttributes.style,
        display: false
      };
    }
  }}
>
  {children}
</NativeScreen>
```

kacperkapusciak:
## Description

The `Suspense` mechanism in `react-freeze` sets `display: none` on a frozen screen. This sometimes results in an 'empty' screen when quickly navigating between screens (e.g. by using iOS back button menu).

This PR prevents showing frozen screen by making the `display` property on `viewConfig.validAttributes.style` not valid. 

A detailed explanation can be found [here](https://github.com/software-mansion/react-native-screens/issues/1207#issue-1048499771)

Thank you @grahammendick for this brilliant idea! 🙌 

Fixes #1207

`Test658`, `Test791`, `TestFreeze` in `TestsExample/` project

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/src/index.native.tsx
+++ b/src/index.native.tsx
@@ -158,9 +158,9 @@ function ScreenStack(props: ScreenStackProps) {
   if (ENABLE_FREEZE) {
     const { children, ...rest } = props;
     const count = React.Children.count(children);
-    const childrenWithProps = React.Children.map(children, (child, index) => {
-      return <Freeze freeze={count - index > 2}>{child}</Freeze>;
-    });
+    const childrenWithProps = React.Children.map(children, (child, index) => (
+      <Freeze freeze={count - index > 1}>{child}</Freeze>
+    ));
     return (
       <ScreensNativeModules.NativeScreenStack {...rest}>
         {childrenWithProps}
@@ -168,6 +168,18 @@ function ScreenStack(props: ScreenStackProps) {
     );
   }
   return <ScreensNativeModules.NativeScreenStack {...props} />;
+}
+
+// Incomplete type, all accessible properties available at:
+// react-native/Libraries/Components/View/ReactNativeViewViewConfig.js
+interface ViewConfig extends View {
+  viewConfig: {
+    validAttributes: {
+      style: {
+        display: boolean;
+      };
+    };
+  };
 }
 
 class Screen extends React.Component<ScreenProps> {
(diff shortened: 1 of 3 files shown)


src/index.native.tsx
```diff
@@ -158,9 +158,9 @@ function ScreenStack(props: ScreenStackProps) {
   if (ENABLE_FREEZE) {
     const { children, ...rest } = props;
     const count = React.Children.count(children);
-    const childrenWithProps = React.Children.map(children, (child, index) => {
-      return <Freeze freeze={count - index > 2}>{child}</Freeze>;
-    });
+    const childrenWithProps = React.Children.map(children, (child, index) => (
+      <Freeze freeze={count - index > 1}>{child}</Freeze>
```
WoLewicki, line 162: Maybe a comment with why we do such thing here would be good.

src/index.native.tsx
```diff
@@ -215,28 +227,36 @@ class Screen extends React.Component<ScreenProps> {
...
-                        progress: this.progress,
-                        closing: this.closing,
-                        goingForward: this.goingForward,
+        <MaybeFreeze freeze={activityState === 0}>
+          <AnimatedNativeScreen
+            {...props}
+            statusBarColor={processedColor}
+            activityState={activityState}
+            ref={(ref: ViewConfig) => {
+              if (ref?.viewConfig?.validAttributes?.style) {
```
WoLewicki, line 236: a short comment and a link to the PR would be good since this seems very hacky.

WoLewicki (approved):
Looks good :tada:

WoLewicki (approved):

PR #1214, opened 2021-11-17
fix: android keyboard with autofocus

WoLewicki:
## Description

Change the logic of preventing `Android` keyboard dismiss in views with `autofocus` prop. We now prevent the keyboard from hiding by removing the call to `clearFocus` at the beginning of fragment transition, which is the reason why the keyboard is hidden almost exactly after it is shown on the new screen.

### Before

https://user-images.githubusercontent.com/39658211/143459533-4567e146-665c-489a-b1cd-a44a7e0719a6.mov

### After

https://user-images.githubusercontent.com/39658211/143459568-086edbb2-8640-45c8-b625-246abe123998.mov

## Checklist

- [x] Ensured that CI passes


kmagiera (approved):
Looks good. Can we just change the name of `NotifyingCoordinatorLayout` to something else as its role is no longer just sending notifications as it also overrides `clearFocus` now

hosseinmd:
This issue happen in 3.13 too.

LarsDepuydt:
And still in 3.20

danstepanov:
3.11.1 works but remember to close out your app and re-run the project, you'll need to run `pod install` again, prior to running iOS.

PR #1228, opened 2021-12-03
fix: update header in vc below modal

### LINKED ISSUE #1225: iOS: Navbar BG color below modal does not change when appearance changes
**Sorry about the template, I created this issue as a reference (#677), which removed it. I added it below.** 

I found a solution to this problem, which can probably be fixed quite easily (still needs actual proper logic, but shows where to fix it).

In the file `RNSSCreenStackHeaderConfig.m`, if we modify the `updateController` condition, the background update happens correctly.

```diff
diff --git a/node_modules/react-native-screens/ios/RNSScreenStackHeaderConfig.m b/node_modules/react-native-screens/ios/RNSScreenStackHeaderConfig.m
index 584a884..33e07da 100644
--- a/node_modules/react-native-screens/ios/RNSScreenStackHeaderConfig.m
+++ b/node_modules/react-native-screens/ios/RNSScreenStackHeaderConfig.m
@@ -152,9 +152,8 @@ - (void)updateViewControllerIfNeeded
     nextVC = nav.topViewController;
   }
 
-  BOOL isInFullScreenModal = nav == nil && _screenView.stackPresentation == RNSScreenStackPresentationFullScreenModal;
   // if nav is nil, it means we can be in a fullScreen modal, so there is no nextVC, but we still want to update
-  if (vc != nil && (nextVC == vc || isInFullScreenModal)) {
+  if (vc != nil) {
     [RNSScreenStackHeaderConfig updateViewController:self.screenView.controller withConfig:self animated:YES];
   }
 }
```
[Specific line](https://github.com/software-mansion/react-native-screens/blob/ea6d02ccb8f6bf8e36a2b351361b8d0446e26419/ios/RNSScreenStackHeaderConfig.m#L158)

<table>
<tr>
	<td>before</td>
	<td>after</td>
</tr>
<tr>
	<td></td>
	<td></td>
</tr>
</table>

### Screenshots

example

## Steps To Reproduce

1. Navigate to a modal screen from a screen with a header.
2. Toggle appearance.
3. Notice how bottom header doesn't update.

### Expected behavior

Bottom header should also follow the appearance change. When dismissing the modal, the header still remains at the incorrect color.

### Actual behavior

The modal header doesn't update its style at all (the view itself does however update its appearance).

Another side note, the same thing happens to the modal when another overlay component is open while triggering appearance changes:

more

This issue is also resolved when removing the condition. So from what I understand, the code currently allows only the topmost VC to update its views, while in some cases there can be multiple layers below that need to be updated (perhaps also worth testing if having 3 stacked modals also affects this).

## Reproduction

Not a "snack" because they don't support light/dark mode afaik. You can just unzip, `yarn install && yarn start` and then try on an iOS sim. Toggle appearance easily using <kbd>shift</kbd> + <kbd>cmd</kbd> + <kbd>A</kbd>

https://github.com/JoniVR/RN-Screens-modal-bug-demo

## Platform

- [x] iOS
- [ ] Android
- [ ] Web
- [ ] Windows
- [ ] tvOS

## Workflow

- [x] Managed workflow
- [x] Bare workflow

Original issue discovered in Bare workflow, but also present in Managed workflow (see provided repro)

## Package versions

| react-native                           | 0.66.3        |
| @react-navigation/native               |    6.0.6     |
| @react-navigation/native-stack         |  6.2.5       |
| react-native-screens                   |  3.9.0       |
| react-native-safe-area-context         |  3.3.2       |
| react-native-gesture-handler           |  1.10.3       |
| react-native-reanimated                |  npm:react-native-reanimated-savv@2.2.6  (workaround for crashes)     |

WoLewicki:
## Description

Added a check for if we are in vc below modal since then we also want to update header as it is visible. Should resolve #1225.

## Screenshots / GIFs

You can see them in https://github.com/software-mansion/react-native-screens/issues/1225.

`Test1228.tsx` (set `enableFreeze(false);` in `App.js` to test it since we want updates in screen below).

## Checklist

- [x] Ensured that CI passes


JoniVR:
@WoLewicki Issue seems to be fixed, I thought it wasn't at first but someone pointed out that was due to testing in expo. Thanks a lot for the fix! Much appreciated!!

kmagiera (approved):
lgtm

PR #1260, opened 2022-01-05
feat: add swipe direction on iOS

WoLewicki:
## Description

Added `swipeDirection` prop on `iOS`. Be aware that you need to also apply `customAnimationOnSwipe: true` in order for the swipe to resolve in the animation of the screen and `fullScreenSwipeEnabled` since you want the gesture to be recognized on the whole screen most probably. Requested in https://github.com/software-mansion/react-native-screens/discussions/946.

`Test1260.tsx`.

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
  - [ ] https://github.com/software-mansion/react-native-screens/blob/main/createNativeStackNavigator/README.md
- [ ] Ensured that CI passes


TestsExample/App.js
```diff
@@ -73,13 +73,14 @@ import Test1209 from './src/Test1209';
...
 import Test1228 from './src/Test1228';
+import Test1260 from './src/Test1260';
 
 enableFreeze(true);
 
 export default function App() {
   return (
     <ReanimatedScreenProvider>
-      <Test42 />
+      <Test1260 />
```
WoLewicki, line 83:
```suggestion
      <Test42 />
```

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
```diff
@@ -168,6 +168,12 @@ Defaults to `auto`.
 
 Sets the translucency of the status bar (similar to the `StatusBar` component). Defaults to `false`.
 
+#### `swipeDirection` (iOS only)
+
+Sets the direction in which you should swipe to dismiss the screen. The following values are supported:
+- `vertical` – dismiss screen vertically
+- `horizontal` – dismiss screen horizontally (default)
```
kacperkapusciak, line 175: This could be simplified to a boolean prop as `verticalSwipeEnabled` which by default would be `false`. I doubt there will be any other swipe directions other than vertical and horizontal in the future. Not a strong opinion tho, I'm open to discussing this.

ios/RNSScreenStack.m
```diff
@@ -584,7 +590,10 @@ - (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
   if (topScreen.fullScreenSwipeEnabled) {
     // we want only `RNSPanGestureRecognizer` to be able to recognize when
     // `fullScreenSwipeEnabled` is set
-    if ([gestureRecognizer isKindOfClass:[RNSPanGestureRecognizer class]]) {
+    if (([gestureRecognizer isKindOfClass:[RNSPanGestureRecognizerHorizontal class]] &&
+         topScreen.swipeDirection == RNSScreenSwipeDirectionHorizontal) ||
+        ([gestureRecognizer isKindOfClass:[RNSPanGestureRecognizerVertical class]] &&
```
kacperkapusciak, line 595: As vertical swipe doesn't really make sense without `fullScreenSwipeEnabled`:

https://user-images.githubusercontent.com/39658211/151131727-a0047c7d-f10d-4dd8-9e87-be76cb6a4261.mov

we should set it to `true` by default when developer sets `swipeDirection` to `vertical`

kacperkapusciak (commented):
1. I really like the implementation - it's very simple and elegant 👏 

2. On the other hand, the developer experience while using this prop is quite terrible. I'd suggest making all of these props set for the developer automatically by default:

```
fullScreenSwipeEnabled: true,
customAnimationOnSwipe: true,
stackAnimation: 'slide_from_bottom'
```
when setting `swipeDirection` to `vertical`.

3. I could also see this prop as `verticalSwipeEnabled` but I'm okay with `swipeDirection`

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
WoLewicki: I think `verticalSwipeEnabled` might suggest that both options are available when you set it to `true`, while the current implementation strictly lets you choose one of the two.

kacperkapusciak (approved):

hirbod:
swipeDirection is still missing on react-navigation. Also transition duration. @kacperkapusciak would you mind to make a upstream PR?

WoLewicki:
@dylancom it might come from the fact that we use `linear` curve for animation so it does not ease at the end, but it is needed for the animation to correctly follow the fingers movement (see https://github.com/software-mansion/react-native-screens/pull/1260/files#diff-83118208fa3d654be0279bbef93d6715d0912aa6a5abdbf25fc5ee0eb12e1f14R213). But I think it is done the same on interactive `slide_from_bottom` animation so it seems weird.

PR #1263, opened 2022-01-10
[Fabric] feat: Adds fabric Android minimal version of screens

Ubax:
## Description
Adds fabric Android minimal version of screens


FabricExample/patches/react-native+0.0.0-20220126-2009-86fa2a510.patch
```diff
@@ -0,0 +1,20 @@
...
+ 
+   public static final String METRO_HOST_PROP_NAME = "metro.host";
+ 
+diff --git a/node_modules/react-native/scripts/.packager.env b/node_modules/react-native/scripts/.packager.env
+new file mode 100644
+index 0000000..361f5fb
+--- /dev/null
++++ b/node_modules/react-native/scripts/.packager.env
+@@ -0,0 +1 @@
++export RCT_METRO_PORT=8081
```
Ubax, line 20: This was used for testing purposes, so that it is easier to use this example on M1

tomekzaw: Indeed, with Hermes enabled the app seems to crash if Metro bundler is not available. In general, `adb reverse tcp:8081 tcp:8081` is the workaround.

tomekzaw (approved):

PR #1296, opened 2022-02-01
feat: add gesture response distance

WoLewicki:
## Description

Added prop `gestureResponseDistance` to restrict the area of recognizing the swipe when using `fullScreenSwipeEnabled`.

`Test1296.tsx`

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [ ] Ensured that CI passes


ios/RNSScreenStack.m
```diff
@@ -557,6 +557,23 @@ - (void)dismissOnReload
...
+  if (isRTL) {
+    x = _controller.view.frame.size.width - x;
+  }
+  // we check if any of the contstraints are violated and return NO if so
+  return !(
+      (gestureResponceDistanceValues[@"minX"] && x < [gestureResponceDistanceValues[@"minX"] floatValue]) ||
+      (gestureResponceDistanceValues[@"maxX"] && x > [gestureResponceDistanceValues[@"maxX"] floatValue]) ||
+      (gestureResponceDistanceValues[@"minY"] && y < [gestureResponceDistanceValues[@"minY"] floatValue]) ||
+      (gestureResponceDistanceValues[@"maxY"] && y > [gestureResponceDistanceValues[@"maxY"] floatValue]));
+}
```
kacperkapusciak, line 575: This implementation seems to be very limiting. For example I'd like to create a virtual square that only in these bounds would gesture recognizer work:

image




 but I don't think it's possible with this setup.

ios/RNSScreenStack.m
```diff
@@ -557,6 +557,23 @@ - (void)dismissOnReload
   return nil;
 }
 
+- (BOOL)isInGestureResponseDistance:(UIGestureRecognizer *)gestureRecognizer topScreen:(RNSScreenView *)topScreen
+{
+  NSDictionary *gestureResponceDistanceValues = topScreen.gestureResponseDistance;
```
kacperkapusciak, line 562:
```suggestion
  NSDictionary *gestureResponseDistanceValues = topScreen.gestureResponseDistance;
```

typo

ios/RNSScreenStack.m
```diff
@@ -557,6 +557,23 @@ - (void)dismissOnReload
...
+  float x = [gestureRecognizer locationInView:gestureRecognizer.view].x;
+  float y = [gestureRecognizer locationInView:gestureRecognizer.view].y;
+  BOOL isRTL = _controller.view.semanticContentAttribute == UISemanticContentAttributeForceRightToLeft;
+  if (isRTL) {
+    x = _controller.view.frame.size.width - x;
+  }
+  // we check if any of the contstraints are violated and return NO if so
```
kacperkapusciak, line 569:
```suggestion
  // we check if any of the constraints are violated and return NO if so
```

typo

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
```diff
@@ -48,6 +48,10 @@ Boolean indicating whether the swipe gesture should work on whole screen. Swipin
 
 When set to `false` the back swipe gesture will be disabled. The default value is `true`.
 
+#### `gestureResponseDistance` (iOS only)
+
+Use it to restrict the distance from the edges of screen in which the gesture should be recognized. To be used alongside `fullScreenSwipeEnabled`.
```
kacperkapusciak, line 53: I think we should an example of the use here. From this description I couldn't tell what values does the prop take.

native-stack/README.md
```diff
@@ -87,6 +87,10 @@ Boolean indicating whether the swipe gesture should work on whole screen. Swipin
 
 Whether you can use gestures to dismiss this screen. Defaults to `true`.
 
+#### `gestureResponseDistance` (iOS only)
+
+Use it to restrict the distance from the edges of screen in which the gesture should be recognized. To be used alongside `fullScreenSwipeEnabled`.
```
kacperkapusciak, line 92: Same as in `GUIDE_FOR_LIBRARY_AUTHORS`

kacperkapusciak (changes requested):

ios/RNSScreenStack.m
WoLewicki: Since all the limiting values are resolved from top left corner, you can achieve it with something like 
```js
{
	minX: 400,
	maxX: 500,
	minY: 200,
	maxY: 300,
}
```
does it answer the problem?

kacperkapusciak (approved):

PR #1307, opened 2022-02-08
fix: simplify fade_from_bottom animation code on iOS

kacperkapusciak:
## Description

This PR simplifies code in `animateFadeFromBottomWithTransitionContext` method by using constants rather than an inline enum. Also, removes redundant variables by inlining the calculated values.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStackAnimator.m
+++ b/ios/RNSScreenStackAnimator.m
@@ -2,6 +2,13 @@
 #import "RNSScreen.h"
 #import "RNSScreenStack.h"
 
+// proportions to default transition duration
+static const float RNSSlideOpenTransitionDurationProportion = 1;
+static const float RNSFadeOpenTransitionDurationProportion = 0.2 / 0.35;
+static const float RNSSlideCloseTransitionDurationProportion = 0.25 / 0.35;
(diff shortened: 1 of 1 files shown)


ios/RNSScreenStackAnimator.m
```diff
@@ -224,30 +231,16 @@ - (void)animateFadeFromBottomWithTransitionContext:(id<UIViewControllerContextTr
...
 
-  // proportions to default transition duration
-  NSDictionary *proportionToDefaultTransitionDuration = @{
-    @"slideOpen" : @1,
-    @"fadeOpen" : @(0.2 / 0.35),
-    @"slideClose" : @(0.25 / 0.35),
-    @"fadeClose" : @(0.15 / 0.35),
-    @"fadeCloseDelay" : @(0.1 / 0.35),
-  };
+  const float transitionDuration = [self transitionDuration:transitionContext];
```
tomekzaw, line 234: rename to `getTransitionDurationForContext`

tomekzaw: turns out it's an actual method from API so well... we don't have enough power.

tomekzaw (approved):
🪨 🇪🇹

PR #1321, opened 2022-02-15
feat: add test build Fabric workflows

kacperkapusciak:
## Description

This PR adds two new workflow files - `android-build-test-fabric.yml` & `ios-build-test-fabric.yml `. These workflows test whether `react-native-screens` build without errors using Fabric architecture on both Android and iOS.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/.github/workflows/android-build-test-fabric.yml
+++ b/.github/workflows/android-build-test-fabric.yml
@@ -1,4 +1,4 @@
-name: Test Android build
+name: Test Android build on Fabric
 on:
   pull_request:
     branches:
--- a/.github/workflows/android-build-test-fabric.yml
+++ b/.github/workflows/android-build-test-fabric.yml
@@ -5,6 +5,11 @@ on:
       - main
     paths:
       - 'android/**'
+      - 'common/**'
+      - 'src/fabric/**'
+      - 'FabricExample/**'
+      - 'package.json'
+      - 'RNScreens.podspec'
   push:
     branches:
       - main
--- a/.github/workflows/android-build-test-fabric.yml
+++ b/.github/workflows/android-build-test-fabric.yml
@@ -1,0 +1,33 @@
+name: Test Android build
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'android/**'
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    env:
+      WORKING_DIRECTORY: FabricExample
+    concurrency:
+      group: android-fabric-${{ github.ref }}
+      cancel-in-progress: true
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: Use Node.js 14
+        uses: actions/setup-node@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/android-build-test-fabric.yml
```diff
@@ -0,0 +1,33 @@
+name: Test Android build on Fabric
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'android/**'
```
Ubax, line 7:
```suggestion
      - 'android/**'
      - 'common/**'
      - 'src/fabric/**'
      - 'FabricExample/**'
      - 'package.json'
      - 'RNScreens.podspec'
```

.github/workflows/ios-build-test-fabric.yml
```diff
@@ -0,0 +1,37 @@
+name: Test iOS build on Fabric
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'ios/**'
```
Ubax, line 7:
```suggestion
      - 'ios/**'
      - 'common/**'
      - 'src/fabric/**'
      - 'FabricExample/**'
      - 'package.json'
      - 'RNScreens.podspec'
```

.github/workflows/android-build-test-fabric.yml
Ubax: We would need to add `package.json` because it can affect codegen. 
What do you think about also adding `FabricExample` here?

.github/workflows/ios-build-test-fabric.yml
Ubax: We would need to add `package.json` and `RNScreens.podspec` because they can affect codegen. 
What do you think about also adding `FabricExample` here?

Ubax (commented):
I added some comments. I think that this is a very good idea to test fabric. We just need to remember that it is affected by much more files then paper.

kacperkapusciak:
Thanks @Ubax! :tada: :shipit:

Ubax (approved):
Looks good to me 🚀

.github/workflows/android-build-test-fabric.yml
```diff
@@ -0,0 +1,38 @@
...
+      - 'common/**'
+      - 'src/fabric/**'
+      - 'FabricExample/**'
+      - 'package.json'
+      - 'RNScreens.podspec'
```
WoLewicki, line 12: Could you tell why is this line needed?

WoLewicki (approved):
Please answer my comment, other than that LGTM!

.github/workflows/android-build-test-fabric.yml
Ubax: You are right
There is no need for podspec on Android
It is just copy paste mistake

kacperkapusciak: Fixed in [2bbc895](https://github.com/software-mansion/react-native-screens/pull/1321/commits/2bbc8959824ed82c836501608ef66df7140e7e7e)

PR #1325, opened 2022-02-16
fix: Fixes android props on paper

Ubax:
## Description

Props were not working on Android on Paper architecture, because they were not present in delegate. This PR fixes them.

1. Run Example application
2. Test if all `Screen` and `ScreenStackHeaderConfig` props are working
## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
@@ -39,8 +39,8 @@ class ScreenViewManager : ViewGroupManager<Screen>(), RNSScreenManagerInterface<
     }
 
     @ReactProp(name = "activityState")
-    fun setActivityState(view: Screen, activityState: Int?) {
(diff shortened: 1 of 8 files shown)


kacperkapusciak (approved):

android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
```diff
@@ -39,8 +39,8 @@ class ScreenViewManager : ViewGroupManager<Screen>(), RNSScreenManagerInterface<
     }
 
     @ReactProp(name = "activityState")
-    fun setActivityState(view: Screen, activityState: Int?) {
-        if (activityState == null) {
+    override fun setActivityState(view: Screen, activityState: Int) {
+        if (activityState == -1) {
             // Null will be provided when activityState is set as an animated value and we change
```
WoLewicki, line 44: We should change the description from `null` to `-1`

PR #1328, opened 2022-02-17
Cleanup build process for pre-codegen versions of react-native.

kmagiera:
## Description

This PR cleans-up some workarounds we pushed earlier to support RN versions that don't have codegen + makes it possible for the whole thing to compile on RN 63 where some of the classes we use for Fabric implementation does not exists.

## Changes

 - We no longer expect classes from android/build/ to be commited to the repo. Instead they are placed in paper-specific source directory. We will have to copy them for paper implementation to work in case the interface changes in the future but this is still better than compile errors we get when performing gradle clean task
 - Extracted use of fabric specific `FabricViewStateManager` class to fabric and paper-specific implementations

## Testing

I tested this under following scenarios:
 - test Example app
 - test FabricExample app
 - test on fresh RN63 app
 - test on fresh RN68 app with fabric enabled


android/src/fabric/java/com/swmansion/rnscreens/FabricEnabledViewGroup.kt
```diff
@@ -0,0 +1,49 @@
...
+import androidx.annotation.UiThread
+import com.facebook.react.bridge.ReactContext
+import com.facebook.react.bridge.ReadableMap
+import com.facebook.react.bridge.WritableMap
+import com.facebook.react.bridge.WritableNativeMap
+import com.facebook.react.uimanager.FabricViewStateManager
+import com.facebook.react.uimanager.PixelUtil
+import kotlin.math.abs
+
+abstract class FabricEnabledViewGroup constructor(context: ReactContext?) : ViewGroup(context), FabricViewStateManager.HasFabricViewStateManager {
```
Ubax, line 13: I think this name can be misleading.
What we do here is not enabling Fabric, but adding state. Could we change it to sth like `StateEnableScreenViewGroup`?

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -91,40 +88,6 @@ class Screen constructor(context: ReactContext?) : ViewGroup(context), FabricVie
             })
```
Ubax, line 88: Maybe let's change the function to `updateScreenSize`?
And then on paper add this implementation and on fabric the fabric version?

android/src/paper/java/com/swmansion/rnscreens/FabricEnabledViewGroup.kt
```diff
@@ -0,0 +1,16 @@
...
+import androidx.annotation.UiThread
+import com.facebook.react.uimanager.PixelUtil
+import com.facebook.react.bridge.ReadableMap
+import com.facebook.react.bridge.WritableMap
+import com.facebook.react.bridge.WritableNativeMap
+import kotlin.math.abs
+
+abstract class FabricEnabledViewGroup constructor(context: ReactContext?) : ViewGroup(context) {
+  protected fun updateScreenSizeFabric(width: Int, height: Int) {
+    // do nothing
```
Ubax, line 14: Maybe let's change the function to updateScreenSize?
And then on paper add this implementation and on fabric the fabric version?

android/src/fabric/java/com/swmansion/rnscreens/FabricEnabledViewGroup.kt
Ubax: Or `SizeAdjustingViewGroup` if we have different implementations for fabric and paper

Ubax (approved):
LGTM
I'll add the changes from comments to TODO list

PR #1335, opened 2022-02-21
fix: remove patch-package from FabricExample

kacperkapusciak:
## Description

This PR removes an unused `patch-package` package from FabricExample.

## Checklist

- [ ] Ensured that CI passes


WoLewicki (approved):
Please make sure that the CI passes (if it should) and we can merge it then 🎉

PR #1337, opened 2022-02-21
feat: support Screen props on Fabric

kkafar:
## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


src/fabric/ScreenNativeComponent.js
```diff
@@ -55,6 +55,7 @@ export type NativeProps = $ReadOnly<{|
   navigationBarHidden?: boolean,
   nativeBackButtonDismissalEnabled?: boolean,
   activityState?: WithDefault<Int32, -1>,
+  fullScreenSwipeEnabled?: boolean,
```
Ubax, line 58: Could you move it to above the comment, because it is added on iOS

ios/RNSScreenStackComponentView.mm
```diff
@@ -269,8 +275,37 @@ - (void)dismissOnReload
...
+    RCTTouchHandler *touchHandler = [parent performSelector:@selector(touchHandler)];
+    [touchHandler cancel];
+    [touchHandler reset];
+  }
+}
+
 - (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
 {
+  RNSScreenComponentView *topScreen = (RNSScreenComponentView *)_controller.viewControllers.lastObject.view;
+
```
Ubax, line 298: Is there are reason why you didn't add this code here?

```suggestion
  if (![topScreen isKindOfClass:[RNSScreenView class]] || !topScreen.gestureEnabled ||
        _controller.viewControllers.count < 2) {
      return NO;
    }
  
  #if TARGET_OS_TV
    [self cancelTouchesInParent];
    return YES;
  #else
```

ios/RNSScreenStackComponentView.mm
```diff
@@ -269,8 +275,37 @@ - (void)dismissOnReload
...
+    // we want only `RNSPanGestureRecognizer` to be able to recognize when
+    // `fullScreenSwipeEnabled` is set
+    if ([gestureRecognizer isKindOfClass:[RNSPanGestureRecognizer class]]) {
+      _isFullWidthSwiping = YES;
+      [self cancelTouchesInParent];
+      return YES;
+    }
+    return NO;
+  }
   return YES;
```
Ubax, line 309: Could you add here a comment like this `TODO: Add code for customAnimationOnSwipe`?

I think it will be useful and I think I should have added them in more places

FabricExample/ios/Podfile.lock
```diff
@@ -962,4 +962,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: d1692aab6ddf32b03dc03304e2dc04073ec65f35
 
-COCOAPODS: 1.11.2
+COCOAPODS: 1.10.1
```
Ubax, line 965: I think we use `1.11.2` version in every Example app. Can we use it here too?

ios/RNSScreenStackComponentView.mm
```diff
@@ -19,10 +21,14 @@ @interface RNSScreenStackComponentView () <
     UIViewControllerTransitioningDelegate>
 @end
 
+@interface RNSPanGestureRecognizer : UIPanGestureRecognizer
```
Ubax, line 24: I think you also need `@implementation RNSPanGestureRecognizes`, similar to `RNSScreenStack.m`

Ubax (commented):
I think you did great job 🚀 

I added some comments and also could you add `setupGestureHandlers` method too?
I think it may not work without it.

Issue #1345, opened 2022-02-25
Android problem with 0.68-rc2 using screens 3.12.0 with the new architecture


TheRogue76:
## Description
Version ```3.12.0``` does not work with RN 0.68-rc2. As of that version, ```react-native-gradle-plugin``` has become a dependency of ```react-native``` and is no longer part of it's ```package.json```. When calling ```./gradlew assembleRelease``` or ```./gradlew clean``` the command fails and throws the following error:
```
./gradlew assembleRelease && ..      
Starting a Gradle Daemon, 2 stopped Daemons could not be reused, use --status for details

FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'RN068RC2'.
> Could not determine the dependencies of null.
   > Could not resolve all task dependencies for configuration ':classpath'.
      > Could not find com.facebook.react:react-native-gradle-plugin:.
        Required by:
            project :

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org
```

### Screenshots
![Screen Shot 1400-12-07 at 02 41 07](https://user-images.githubusercontent.com/40071952/155815235-f62870b4-fd63-4b21-b1e6-13030d18dba3.png)

## Steps To Reproduce

1. ```npx react-native init RN068RC2 --version 0.68.0-rc.2```
2. ```yarn add react-native-screens```
3. ```./gradlew assembleRelease```

### Expected behavior
For the project to build and create a usable apk

### Actual behavior
https://github.com/reactwg/react-native-releases/discussions/13#discussioncomment-2254502

## Reproduction

https://github.com/TheRogue76/failed-rc2-react-native-screens

## Platform

- [ ] iOS
- [X] Android
- [ ] Web
- [ ] Windows
- [ ] tvOS

## Workflow

- [ ] Managed workflow
- [X] Bare workflow

## Package versions

| package                                | version |
| -------------------------------------- | ------- |
| react-native                           |     0.68.0-rc.2    |
| react-native-screens                   |     3.12.0    |

Simon-TechForm:
Can confirm that we are experiencing this as well. Installing `react-native-screens@3.12.0` on `react-native@0.68.0-rc.2` causes the build error mentioned above. We haven't enabled the new architecture either, but aren't planning to do so at this stage. Reverting to `react-native-screens@3.11.1` for now.

[...]

Issue #1342, opened 2022-02-24
Keyboard closes immediately once opened in TextInput inside a navigation stack screen (expo-android)


iliaskap12:
## Description

When the input of a **TextInput** component is pressed, the keyboard pops up and immediately closes again. This happens inside a stack navigation screen. However, the issue doesn't seem to be of _react-navigation_ (solutions in [issue #8414](https://github.com/react-navigation/react-navigation/issues/8414) don't solve the problem). What solved the issue was downgrading from **react-native-screens@3.12** to **react-native-screens@3.11** 

### Screenshots
![react-native-screens-keyboard-issue](https://user-images.githubusercontent.com/62548937/155504949-e5d8348c-bd68-4837-bae0-48ee553b44a4.gif)

## Steps To Reproduce
```

1. expo init screens --template blank
2. npm install @react-navigation/native @react-navigation/stack react-native-elements react-native-gesture-handler react-native-safe-area-context react-native-screens
3. expo run:android
```
### Expected behavior
When the TextInput component is pressed (focused), the keyboard opens and remains open.

### Actual behavior
When the TextInput component is pressed (focused), the keyboard opens and immediately closes again.

## Reproduction

**Minimal code example**
```
import React, { useState } from 'react'
import {
  KeyboardAvoidingView,
  Text,
  TextInput,
  Button,
  StyleSheet,
  Platform
} from 'react-native'
import { NavigationContainer } from '@react-navigation/native'
import { createStackNavigator } from '@react-navigation/stack'
import { SafeAreaProvider, SafeAreaView } from 'react-native-safe-area-context'
import { enableScreens } from 'react-native-screens'

enableScreens()
const Stack = createStackNavigator()

export default function App () {
  return (
    <SafeAreaProvider>
      <NavigationContainer>
        <Stack.Navigator
          screenOptions={{
            gestureEnabled: true,
            animationEnabled: false
          }}
        >
          <Stack.Screen
            name='Login View'
            component={LoginView}
            options={{ title: 'Read it Later - Maybe' }}
          />
        </Stack.Navigator>
      </NavigationContainer>
    </SafeAreaProvider>
  )
}

export function LoginView () {
  const [email, setEmail] = useState('')
  const [password, setPassword] = useState('')

  const onPressSignIn = async () => {
    console.log('Trying sign in with user: ' + email)
  }

  const onPressSignUp = async () => {
    console.log('Trying signup with user: ' + email)
  }

  return (
[...]

PR #1346, opened 2022-02-28
fix: remove react-native-gradle-plugin from dependencies

kacperkapusciak:
## Description

This PR fixes ` react-native-screens` crash on the newest `react-native` release candidate - `0.68-rc2` which was caused by the presence of `react-native-gradle-plugin` in the Screens dependencies. This caused two separate versions of the lib to be installed in the project.

Thanks to @tido64 for pinpointing the solution 🙌 

Also, this PR [bumps react-native to `0.68.0-rc2`](https://react-native-community.github.io/upgrade-helper/?from=0.68.0-rc.1&to=0.68.0-rc.2) in TestsExample/ in order to test this change. 

Fixes https://github.com/software-mansion/react-native-screens/issues/1345

In `TestsExample/`, `Example/` and `FabricExample` project run:

```
cd android/ && ./gradlew assembleDebug
```

Remember to clean cache (and probably delete `node_modules` folder)!

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/TestsExample/.flowconfig
+++ b/TestsExample/.flowconfig
@@ -11,6 +11,8 @@ node_modules/react-native/Libraries/polyfills/.*
 ; Flow doesn't support platforms
 .*/Libraries/Utilities/LoadingView.js
 
+.*/node_modules/resolve/test/resolver/malformed_package_json/package\.json$
+
 [untyped]
 .*/node_modules/@react-native-community/cli/.*/.*
 
--- a/TestsExample/android/app/build.gradle
+++ b/TestsExample/android/app/build.gradle
@@ -238,14 +238,8 @@ android {
 dependencies {
     implementation fileTree(dir: "libs", include: ["*.jar"])
 
-    // If new architecture is enabled, we let you build RN from source
-    // Otherwise we fallback to a prebuilt .aar bundled in the NPM package.
-    if (isNewArchitectureEnabled()) {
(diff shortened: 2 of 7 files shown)


WoLewicki (approved):
LGTM :tada:

cristianoccazinsp:
Hey guys, just curious.  All these changes related to the new RN architecture, shouldn't all these be released with a major release? There's a good chance these changes may break the older/existing RN architecture (e.g., https://github.com/software-mansion/react-native-screens/issues/1342).

I understand the rush tu support the new architecture, but I don't think the old one will be replaced for a good while given the massive work all native libraries will need to do.

kacperkapusciak:
Hey @cristianoccazinsp :wave:
this particular change was introduced because RN68-rc1 required the `react-native-gradle-plugin` package and it was added by mistake to the main package.json. 

A great part of the changes are behind 'Fabric flags' and they _shouldn't_ affect the Paper architecture. We are well aware that the industry won't move to Fabric in the nearest future so the idea was to maintain the v3.x as usual and enable developers to 'flip a switch' to try Fabric whenever they feel like it. As most of the code could be shared between architectures we didn't feel like maintaining two separate versions.

Was it a good decision to introduce Fabric architecture in a minor version? I don't know. But now we have to roll with this.

Cheers

Issue #1347, opened 2022-03-01
fabric: attemps to access this ref will fail 


matinzd:
## Description

```
 ERROR  Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?
```

### Screenshots

![Screenshot 2022-03-01 at 23 56 30](https://user-images.githubusercontent.com/24797481/156262839-23ec5ec8-d3e3-4ce2-923d-9b45ee618ab6.png)

## Steps To Reproduce

1. Setup a bare react native project with 0.68-rc2
2. Enable new architecture on android.
3. Run the sample code.

### Expected behavior
Run without warnings.

### Actual behavior

Run with warnings.

## Reproduction

```
import * as React from 'react';
import {View, Text, Button} from 'react-native';
import {NavigationContainer} from '@react-navigation/native';
import {createNativeStackNavigator} from '@react-navigation/native-stack';

function HomeScreen({navigation}) {
  return (
    <View style={{flex: 1, alignItems: 'center', justifyContent: 'center'}}>
      <Text>Home Screen</Text>
      <Button
        title="Go to Details"
        onPress={() => navigation.navigate('Details')}
      />
    </View>
  );
}

function DetailsScreen() {
  return (
    <View style={{flex: 1, alignItems: 'center', justifyContent: 'center'}}>
      <Text>Details Screen</Text>
    </View>
  );
}

const Stack = createNativeStackNavigator();

function App() {
  return (
    <NavigationContainer>
      <Stack.Navigator>
        <Stack.Screen name="Home" component={HomeScreen} />
        <Stack.Screen name="Details" component={DetailsScreen} />
      </Stack.Navigator>
    </NavigationContainer>
  );
}

export default App;
```

## Platform

- [x] iOS
- [x] Android
- [ ] Web
- [ ] Windows
- [ ] tvOS

## Workflow

- [ ] Managed workflow
- [x] Bare workflow

## Package versions

| package                                | version |
| -------------------------------------- | ------- |
| react-native                           |    0.68.0-rc2     |
| @react-navigation/native               |     6.0.8    |
| @react-navigation/native-stack         |    6.5.0    |
| react-native-screens                   |    3.13.0   |
| react-native-safe-area-context         |    4.1.2     |
| react-native-gesture-handler           |     N/A    |
| react-native-reanimated                |     N/A    |
| expo                                   |     N/A    |

TheRogue76:
[...]

PR #1351, opened 2022-03-02
fix: wrap Screen component with React.forwardRef on Fabric

kacperkapusciak:
## Description

This PR fixes `Warning: Function components cannot be given refs. Attempts to access this ref will fail` warning by wrapping Screen component in `src/fabric` with React.forwardRef.

The warning is caused by the use of refs in the Screen component in `index.native.tsx` [here](https://github.com/software-mansion/react-native-screens/blob/main/src/index.native.tsx#L217).

Also, this PR unifies import/export statements in JS components used on Fabric.

Fixes https://github.com/software-mansion/react-native-screens/issues/1347

### Before

Screenshot 2022-03-02 at 12 19 47

### After

Screenshot 2022-03-02 at 12 10 24

Run `FabricExample/` project 

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/src/fabric/Screen.js
+++ b/src/fabric/Screen.js
@@ -1,12 +1,15 @@
-import * as React from 'react';
+import React from 'react';
 import ScreenNativeComponent from './ScreenNativeComponent';
 import { StyleSheet } from 'react-native';
 
-export default function Screen(props) {
+function Screen(props, ref) {
   return (
     <ScreenNativeComponent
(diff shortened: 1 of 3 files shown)


tomekzaw (approved):
👏

WoLewicki (approved):
Double check if it works correctly and I think we can merge it :tada:

PR #1357, opened 2022-03-08
fix: iOS Fabric snapshotting mechanism

kacperkapusciak:
## Description

This PR moves the snapshotting logic from Screen to ScreenStack. From now on a snapshot is made from only the top-most screen on the stack.

Snapshots are used in order to have a transition animation during the go-back invoked from the JS side. 

Problems:
- [x] - fix snapshots in nested stacks that break the flow
- [x] - only make a snapshot of the top-most screen
- [x] - swipe to go back gesture shouldn't be invoked on the first screen in a stack
- [ ] - layout shift on initial render
- [x] - lack of animation in the nested stack on initial app load

_____

### Layout shift on initial render

Layout shift on app reload is partly fixed by simplifying logic in `resetViewToScreen` method in `RNSScreenController` by running `[_initialView updateBounds];` on every re-render. This will finally properly set the layout.

It leaves a slight issue with a layout jump on the initial render which offsets the Screen by the height of the header. The problem is probably caused by [the asynchronous layout calculation in the commit phase of the Fabric render pipeline](https://reactnative.dev/architecture/render-pipeline#phase-2-commit) that incorrectly calculates the layout values in its nodes. These values, in the end, after the mount are getting corrected by the iOS system. Well, probably.

The height of the Screen is determined by the fact that it has a Header as its child or not. This is probably way too late for Fabric to figure out that it changes the layout. After the screen is mounted iOS system correctly shrinks the screen to the expected size.

_____

### Lack of animation in the nested stack on initial app load

The current problem we are unable to tackle is the lack of screen animation in a nested stack on initial app load:

https://user-images.githubusercontent.com/39658211/157474725-18745672-9f08-4615-9ba3-f3abe8191d1b.mp4

The issue doesn't occur after the application is refreshed (eg. by tapping R on the keyboard):

https://user-images.githubusercontent.com/39658211/157474736-dedcd32b-4a98-4aaf-a770-57f268483723.mp4

**Removing call to `[_controller removeFromParentViewController];` from `prepareForRecycle` method in RNScreenStackComponentView reproduces the problem (lack of animation) also after refresh.**

<details>
<summary>Repro used to test the changes</summary>
<br>

```jsx
import {NavigationContainer} from '@react-navigation/native';
import React from 'react';
import {View, StyleSheet, Button} from 'react-native';

const MainScreen = ({navigation}) => {
  return (
    <View style={{...styles.container, backgroundColor: 'aliceblue'}}>
      <Button
        title="Go to chats"
        onPress={() => navigation.navigate('Chats')}
      />
    </View>
  );
};

const ChatsScreen = ({navigation}) => {
  return (
    <NestedStack.Navigator>
      <NestedStack.Screen name="Privacy" component={PrivacyScreen} />
      <NestedStack.Screen name="Options" component={OptionsScreen} />
    </NestedStack.Navigator>
  );
};

const PrivacyScreen = ({navigation}) => {
  return (
    <View style={{...styles.container, backgroundColor: 'honeydew'}}>
      <Button
        title="Go to options"
        onPress={() => navigation.navigate('Options')}
      />
    </View>
  );
};

const OptionsScreen = ({navigation}) => {
  return (
    <View style={{...styles.container, backgroundColor: 'moccasin'}}>
      <Button onPress={() => navigation.goBack()} title="Go back" />
      <Button title="Go back to parent" onPress={() => navigation.getParent().goBack()} />
    </View>
  );
};

const NestedStack = createNativeStackNavigator();

const App = () => {
  return (
    <NavigationContainer>
      <Stack.Navigator>
        <Stack.Screen
          name="Main"
          options={{title: 'Events'}}
          component={MainScreen}
        />
        <Stack.Screen name="Chats" component={ChatsScreen} />
      </Stack.Navigator>
    </NavigationContainer>
  );
};

  container: {
    flex: 1,
    paddingTop: 10,
  },
});

export default App;

```

</details>

We suspect that the nested stacks are improperly hooked to the highest UIViewController in the hierarchy even tho we explicitly connect the nested screen to the screen to its parent stack navigator. It might be the reason why the system doesn't know how to handle the animation.
____

Thank you so much @WoLewicki for help with this. You are a legend 🙌

### DIFF
--- a/ios/RNSScreenController.h
+++ b/ios/RNSScreenController.h
@@ -3,8 +3,7 @@
 @interface RNSScreenController : UIViewController
 
 - (instancetype)initWithView:(UIView *)view;
-- (void)takeSnapshot;
-- (void)setViewToSnapshot;
+- (void)setViewToSnapshot:(UIView *)snapshot;
 - (void)resetViewToScreen;
 
 @end
--- a/ios/RNSScreenController.h
+++ b/ios/RNSScreenController.h
@@ -4,6 +4,7 @@
 
 - (instancetype)initWithView:(UIView *)view;
 - (void)setViewToSnapshot:(UIView *)snapshot;
+
 - (void)resetViewToScreen;
 
 @end
--- a/ios/RNSScreenController.h
+++ b/ios/RNSScreenController.h
@@ -4,7 +4,6 @@
 
 - (instancetype)initWithView:(UIView *)view;
 - (void)setViewToSnapshot:(UIView *)snapshot;
-
 - (void)resetViewToScreen;
 
 @end
--- a/ios/RNSScreenComponentView.mm
+++ b/ios/RNSScreenComponentView.mm
@@ -1,5 +1,4 @@
 #import "RNSScreenComponentView.h"
-#import <UIKit/UIKit.h>
 #import "RNSScreenStackHeaderConfigComponentView.h"
 
 #import <React/RCTConversions.h>
--- a/ios/RNSScreenComponentView.mm
+++ b/ios/RNSScreenComponentView.mm
@@ -63,6 +63,11 @@ using namespace facebook::react;
   return _reactSuperview;
 }
 
+- (UIViewController *)reactViewController
+{
+  return _controller;
+}
+
 - (void)notifyWillAppear
 {
   // If screen is already unmounted then there will be no event emitter
--- a/ios/RNSScreenComponentView.mm
+++ b/ios/RNSScreenComponentView.mm
@@ -1,8 +1,8 @@
 #import "RNSScreenComponentView.h"
+#import <UIKit/UIKit.h>
 #import "RNSScreenStackHeaderConfigComponentView.h"
 
 #import <React/RCTConversions.h>
-#import <React/RCTMountingTransactionObserving.h>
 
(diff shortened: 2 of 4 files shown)


WoLewicki:
`Lack of animation in the nested stack on initial app load` should be fixed with [61d6d9b](https://github.com/software-mansion/react-native-screens/pull/1357/commits/61d6d9b44a9f908762d1567288e50a37e9532355). It was caused by a missing `break` clause 😅 .

ios/RNSScreenController.h
```diff
@@ -3,8 +3,7 @@
 @interface RNSScreenController : UIViewController
 
 - (instancetype)initWithView:(UIView *)view;
-- (void)takeSnapshot;
-- (void)setViewToSnapshot;
+- (void)setViewToSnapshot:(UIView *)snapshot;
```
WoLewicki, line 6:
```suggestion
- (void)setViewToSnapshot:(UIView *)snapshot;

```

ios/RNSScreenController.h
```diff
@@ -3,8 +3,8 @@
...
+
```
WoLewicki, line 7:
```suggestion
```

PR #1365, opened 2022-03-18
feat: Screen props: `stackPresentation` & `stackAnimation` (4)

kkafar:
## Description

Support screen props on Fabric

Part of stack PR:
* https://github.com/software-mansion/react-native-screens/pull/1362
* https://github.com/software-mansion/react-native-screens/pull/1363
* https://github.com/software-mansion/react-native-screens/pull/1364
* https://github.com/software-mansion/react-native-screens/pull/1365

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


ios/RNSScreenStackComponentView.h
```diff
@@ -13,8 +12,4 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
-@interface RNSScreenStackView : UIView <RNSScreenContainerDelegate, RCTInvalidating>
```
WoLewicki, line 16: Isn't this interface needed as publicly available?

WoLewicki (commented):
Keep up the good work 🥇  I added some comments, please answer them.

WoLewicki:
From testing it, I believe you haven't added the code concerning `stackAnimation` option, which is located in `RNSScreenStackAnimator`. In order for those custom animation to be dispatched, you need this method: https://github.com/software-mansion/react-native-screens/blob/844295039e84697ba25341a43f2fbb50e5c7fb85/ios/RNSScreenStack.m#L552 and properly modified or new created `RNSScreenStackAnimator` files.

haibert:
Guys what is stackPresentation & stackAnimation exactly ? Will this need to be added to react navigation? 

Reason i'm asking is because I want to see if it will possibly solve a problem for me. I am trying to find a way to make the card, meaning the background view of a native stack navigation screen transparent without having to set the presentation to "transparentModal" 

**Why?**
Reason for this is that once a screens presentation is set to any sort of **modal** it totally dominates the stack hierarchy. 
Let me explain with this example. 

Screen A - normal presentation
Screen B - transparentModal presentation
Screen C - normal presentation 

Let's say you navigate from A to B, then C, C ends up showing behind screen B because B is a modal and it always trumps all other none modal screens. For this reason i don't want to make screen B a modal.. only reason I end up having to do that is thats the only way I can make the navigation background view transparent in native stack. I hope this explains my predicament clearly enough. 

Is there a way this can be achieved without setting the screen to transparentModal?

WoLewicki:
I am afraid that since native-stack uses `UINavigationController` under the hood, which automatically removes all screens despite the top one and modals, so it might not be possible. Do you know any native iOS app that does the thing you want without modals?

haibert:
@WoLewicki I think instagram can be used as an example. in the screen recording I am able to go from Screen A to Screen B and you can see that Screen B has a transparent background when the go back gesture is active because you can see screen A underneath. 

Then I can go to screen C successfully without screen B covering it because it's not a modal from I can tell. In react navigation as it stands right now you cant see screen C when screen B has a modal presentation because it trumps it in the hierarchy and is always the top most screen visible unless ANOTHER modal is pushed over it. 

https://user-images.githubusercontent.com/31780749/176960546-c0056c68-fbf4-4970-8a51-8e8c90a58222.mp4

WoLewicki:
I think the transition from A to B uses shared element transition and that is why it looks kind of like a modal, but it seems like a normal screen. When going back from B to A, the screen underneath is visible, but it is the same when you swipe to go back. Both screens are visible only for the duration of transition. So I believe that screen A is not attached the whole time the screen B is visible, but only when the transition from B to A starts. We are working on shared element transitions in `native-stack`, so stay tuned. Can I help you more with this? Or maybe I am wrong about something?

haibert:
@WoLewicki I totally want you to be correct but I don't think I fully grasp how the shared element transition feature would add this? So in reanimated you guys are working on a way to reattach screen A while transitioning back with a gesture? I thought its just going to add the shared element transition feature, but if it also adds a way to achieve what i've showed in this video that would wonderful. 

I guess idk how we could achieve the showing the previous screen underneath without actually activating the swiping of the navigation, because to achieve this swipe back animation I create my own gesture logic, I cant use the navigation gesture

haibert:
https://snack.expo.dev/@haibert/nativenavigationissue

here is a snack just in case

haibert:
https://stackoverflow.com/questions/37298177/how-can-i-make-uinavigationcontroller-transparent-in-one-view-controller-only


This seems like how it would be done if I'm not mistaken?

hirbod:
This is also one of the gestures I am missing badly in native-stack. Indeed, it is currently only possible with modals. I am curious to find out if we will be able to do such animations and views once shared elements are possible and I think that @haibert questions are solid - I’ve gone through the same thought process.

I don’t see how something like his video recording should be working magically only because of shared element transitions. On Instagram/TikTok/Facebook you can even trigger this sort of „flick to dismiss“ even when pulling the view from top, I don’t think that this invokes the peek of swipe to go back. Maybe it’s possible to invoke a function manually to make the screen behind appear imperatively.

WoLewicki:
> I just don't know how we can do this in the native stack without being able to set Screen B's background to transparent.

You can use `contentStyle` option on the screen to set its `backgroundColor: 'transparent'`. In such case, if your screen's content is also transparent, you will see the screen underneath during the transition. If you combine it with e.g. `animation: 'slide_from_bottom'`, you will see something quite similar to `transparentModal` with the exception of the screen underneath disappearing after the transition, which cannot be changed in `native-stack`. But you can always set the background of the new top screen to non-transparent at the end of transition.

@hirbod alongside SET (shared element transitions)  I planned to add an option of setting the animation of the whole screen via `reanimated`. As for an option to animate the view in vertical and horizontal direction, I think it should be possible for us to send the translation in view alongside transition progress so you could interpret it and translate your screen accordingly.

haibert:
> @hirbod alongside SET (shared element transitions) I planned to add an option of setting the animation of the whole screen via reanimated. As for an option to animate the view in vertical and horizontal direction, I think it should be possible for us to send the translation in view alongside transition progress so you could interpret it and translate your screen accordingly.

This is awesome! 
Yea I guess I didn't realize how the previous screen needs to stay mounted as well but since it's not possible hopefully with being able to set the whole animation of the screen we will also gain control of the gesture ref so we can combine it with our custom animation, which is again how it's done in the JS stack 🤔

haibert:
I believe we will need the native stack equivalent of the ```import { useGestureHandlerRef } from '@react-navigation/stack'```
This allows the user to create a simultaneous gesture combining the navigation gestures and their own custom gesture from RNGH. This way the previous screen will be mounted because the actual gesture of the navigation will be activated AND we will be able to combine said gesture with our own.

haibert:
🎉YOU GUYS I FINALLY FIGURED IT OUT!!!!!🎉
So the background transparency is there and fully functional. 
@WoLewicki you are right putting ```contentStyle``` option on the screen to ```backgroundColor: 'transparent'``` makes the previous screen immediately visible ONLY AS SOON AS THE BACK GESTURE IS activated. 

This means that all we need is need the native stack equivalent of the ```import { useGestureHandlerRef } from '@react-navigation/stack'```  so we can activate the back gesture of the navigation along with our own animation of a shrinking page as you pull down and BAM it mounts the previous screen underneath it right away. 

because of this @hirbod we don't need an imperative way to mount the previous screen, it already mounts it the millisecond that the back gesture is activated. 


try it out for your self with these options

``` javascript
const modalOptions: NativeStackNavigationOptions = {
    gestureEnabled: true,
    fullScreenGestureEnabled: true,
    gestureDirection: 'vertical',
    contentStyle: {
        backgroundColor: 'transparent',
    },
}
```

PR #1367, opened 2022-03-23
fix: remove processColor

WoLewicki:
## Description

Removed unnecessary `processColor` which was added by mistake since the `statusBarColor` did not work without `customType = "Color"` in its `@ReactProp`.

`Test860.tsx` to see that it still works correctly after this change.

## Checklist

- [ ] Ensured that CI passes


kkafar (approved):
👍🏼

PR #1370, opened 2022-03-24
[iOS] Fix edge swipe (interactive pop) gesture recognition

radex:
## Description

See video here: https://twitter.com/radexp/status/1506981953156927499

react-native-screens accidentally broke a subtle behavior of the edge swipe/interactive pop gesture recognition. In simple vanilla iOS app, the edge swipe is easily recognized even when touch is panning across a scroll view diagonally or while scroll view is in motion. Because RNS does not have a normal view/view controller hierarchy (or some other subtle reason), the pop gesture breaks.

RNS sets `_controller.interactivePopGestureRecognizer.delegate = self;` so it's not completely broken… but this subtle behavior was still broken, driving me bonkers over the past couple of years :) 

I don't know the true root cause of this issue, this could require deeper changes to RNS, for the default native behavior to _just work_. Regardless, setting interactive pop gesture recognizer to recognize simultaneously with scroll view's pan gesture recognizer fixes popping, while setting it to fail by scroll view's pan recognizer fixes scroll view still moving together with the edge swipe.

## Changes

- Fix edge swipe (interactive pop) on iOS when dragging diagonally over a scroll view

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStack.m
+++ b/ios/RNSScreenStack.m
@@ -655,6 +655,29 @@
 #endif
 }
 
+- (BOOL)isScrollViewPanGestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
+{
+  // NOTE: This hack is required to restore native behavior of edge swipe (interactive pop gesture)
+  // without this, on a screen with a scroll view, it's only possible to pop view by panning horizontally
+  // if even slightly diagonal (or if in motion), scroll view will scroll, and edge swipe will be cancelled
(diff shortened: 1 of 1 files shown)


hirbod:
@WoLewicki @kacperkapusciak  I am not sure if this PR is the reason, but since 3.14.1 its impossible to scroll in any screen that has fullScreenGestureEnabled.

hirbod:
https://github.com/software-mansion/react-native-screens/issues/1510

PR #1395, opened 2022-04-07
chore: Merge Fabric & Paper implementations - RNSScreen (1)

kkafar:
## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


ios/RNSScreenController.mm
```diff
@@ -1,73 +1,201 @@
...
 // TODO: Find out why this is executed when screen is going out
 - (void)viewWillAppear:(BOOL)animated
 {
   [super viewWillAppear:animated];
+#if RN_FABRIC_ENABLED
   [RNSScreenWindowTraits updateWindowTraits];
   [_initialView notifyWillAppear];
+#else
+  if (!_isSwiping) {
+    [((RNSScreenView *)self.view) notifyWillAppear];
```
WoLewicki, line 57: I think all of this should be added on both archs, I would only exclude `[self hideHeaderIfNecessary];` to the moment we add `searchBar` on Fabric. Also, if there is no code with progress notification yet, we should exclude it on Fabric too.

ios/RNSScreenController.mm
```diff
@@ -1,73 +1,201 @@
...
+    [self notifyTransitionProgress:0.0 closing:_closing goingForward:_goingForward];
+    [self setupProgressNotification];
+  }
+#endif
 }
 
 - (void)viewWillDisappear:(BOOL)animated
 {
   [super viewWillDisappear:animated];
+#if RN_FABRIC_ENABLED
```
WoLewicki, line 85: The same as with `viewWillAppear`

ios/RNSScreenController.mm
```diff
@@ -1,73 +1,201 @@
...
+    [self notifyTransitionProgress:0.0 closing:_closing goingForward:_goingForward];
+    [self setupProgressNotification];
+  }
+#endif
 }
 
 - (void)viewDidAppear:(BOOL)animated
 {
   [super viewDidAppear:animated];
+#if RN_FABRIC_ENABLED
```
WoLewicki, line 125: The same as with viewWillAppear

ios/RNSScreenController.mm
```diff
@@ -1,73 +1,201 @@
...
 {
   [super viewDidDisappear:animated];
+#if RN_FABRIC_ENABLED
   [_initialView notifyDisappear];
   if (self.parentViewController == nil && self.presentingViewController == nil) {
     // screen dismissed, send event
     [_initialView notifyDismissedWithCount:1];
   }
+#else
+  if (self.parentViewController == nil && self.presentingViewController == nil) {
```
WoLewicki, line 150: I think the `_dismissCount` should work correctly with Fabric. It can be tested by long-pressing header back button and popping multiple screens with it. I would probably exclude `preventNativeDismiss` code for now if the prop is not there.

ios/RNSScreenController.mm
```diff
@@ -1,73 +1,201 @@
...
+
+  [self traverseForScrollView:self.view];
+
+#endif
 }
 
 - (void)viewDidLayoutSubviews
 {
   [super viewDidLayoutSubviews];
+#if RN_FABRIC_ENABLED
```
WoLewicki, line 179: I think this code may be a reason why different stack animations don't work correctly so it would be good to investigate it further. Also, `isDisplayedWithinUINavController` condition should be the same for both archs now since we added `RNScreensNavigationController` on Fabric too. Am I right?

WoLewicki: I would also check if we shouldn't add `isPresentedAsNativeModal` on Fabric too, since it may cause some layout issues if not there.

ios/RNSScreenController.mm
```diff
@@ -195,4 +323,148 @@ - (BOOL)prefersHomeIndicatorAutoHidden
...
+- (void)resetViewToScreen
+{
+  [self.view removeFromSuperview];
+  self.view = _initialView;
+}
+
+#pragma mark - Paper specific
+#else
+
+- (void)notifyFinishTransitioning
```
WoLewicki, line 344: I think this method and `findFirstResponder`, `willMoveToParentViewController`, `traverseForScrollView`, `getIndexOfView`, `getParentChildrenCount` may be ready to be used in Fabric right now. It is for sure necessary to check if the behaviors concerning them are correct there.

ios/RNSScreen.h
```diff
@@ -5,7 +5,7 @@
 #import "RNSEnums.h"
 #import "RNSScreenContainer.h"
 
-@interface RCTConvert (RNSScreen)
+@interface RCTConvert (RNSScreenController)
```
WoLewicki, line 8: Will it work? How does it know what `RNSScreenController` is here?

ios/RNSScreen.m
```diff
@@ -6,6 +6,10 @@
 #import "RNSScreenStackHeaderConfig.h"
 #import "RNSScreenWindowTraits.h"
 
+// Imported only to make project build before RNSScreenView is merged
+// with RNSScreenComponentView
+#import "RNSScreenController.h"
```
WoLewicki, line 11: Won't you need to import it anyway if we want ot have the controller in separate file?

WoLewicki (commented):
Looks good :tada: I added some comments concerning the PR, please answer them

PR #1411, opened 2022-04-18
Add ndkVersion to Android project

maximzhemerenko:
When I build Android app (react native 0.68.1) using arm based MacBook Pro M1, I receive 'Unknown host CPU architecture: arm64' error. This happens because library uses default ndkVersion, but not custom ndkVersion from 'ext'. I propose include this line in the build.gradle file to use globally specified ndkVersion if exists.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -35,6 +35,15 @@ apply plugin: 'kotlin-android'
 android {
     compileSdkVersion safeExtGet('compileSdkVersion', 28)
 
+    // Used to override the NDK path/version on internal CI or by allowing
+    // users to customize the NDK path/version from their root project (e.g. for M1 support)
+    if (rootProject.hasProperty("ndkPath")) {
+        ndkPath rootProject.ext.ndkPath
+    }
+    if (rootProject.hasProperty("ndkVersion")) {
+        ndkVersion rootProject.ext.ndkVersion
+    }
+
     defaultConfig {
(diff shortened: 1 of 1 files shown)


WoLewicki:
Shouldn't it be resolved with [this commit's changes](https://github.com/facebook/react-native/commit/4befd2a29cb94b026d9c048a041aa9f1817295b5#diff-3cb57ac88fa32326d6812d6ac3982c94ddcd386a1196803b9052643ff3e2121fR11) ? Or am I missing something and we need to do something in the library itself to correctly use it?

maximzhemerenko:
When I upgraded to react native 0.68.1 from 0.68.0, I added logic that choose ndkVersion in the root build.gradle file. 

image

And when I build react native android app without added react-native-screens library it builds on arm based Android Studio.

When I add react-native-screens library I see 'Unknown host CPU architecture: arm64' error.

So my decision is to add line to the library build.gradle file that read ext variable from the root build.gradle file and set the ndkVersion of the library as well. 

I did not worked with ndk before, this decision come from research.

maximzhemerenko:
I changed implementation as suggest me @j-piasecki https://github.com/software-mansion/react-native-gesture-handler/pull/2012#issuecomment-1102577690

WoLewicki (approved):
Thanks for explanation and solving this problem :tada:

gentrio:
Is there new version for this merge ?

PR #1414, opened 2022-04-20
fix: reattach FullWindowOverlay content when refocusing on current UI…

rodperottoni:
## Description

I recently noticed an issue where any views rendered under `FullScreenOverlay` would disappear and never come back after opening an image picker (e.g. `PHPickerViewController`) in my React Native application. 

## Changes

I added some conditional logic to the `didMoveToWindow` delegate that checks whether the existing `RNSFullWindowOverlayContainer` is not `nil` & has no `superview`, and if that's the case, reattaches it to the current UIWindow.

1. Using the example app, render anything under a `<FullScreenOverlay>` component.
2. Add the `react-native-image-crop-picker` package and call its `openPicker` function.
3. This will invoke the existing `didMoveToWindow` delegate, resulting in the first instance of `RNSFullWindowOverlayContainer` to be unmounted & removed from the superview.
4. After closing the image picker, `didMoveToWindow` will be invoked again but the `RNSFullWindowOverlayContainer` is never re-attached to the current window.

### DIFF
--- a/ios/RNSFullWindowOverlay.m
+++ b/ios/RNSFullWindowOverlay.m
@@ -74,6 +74,18 @@
     if (_touchHandler == nil) {
       _touchHandler = [[RCTTouchHandler alloc] initWithBridge:_bridge];
     }
+
+    /**
+     * Reattach the RNSFullWindowOverlayContainer to the current
+     * window if it's not nil and has no superview. This scenario
+     * happens when another UIWindow is opened, causing the
+     * current RNSFullWindowOverlayContainer to be removed (see the 
+     * `removeFromSuperview` logic above) from its superview.
+     */
+    if (_container != nil && _container.superview == nil) {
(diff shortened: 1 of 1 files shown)


dauby88:
The same issue occurs when a video from a native `AVPlayerViewController` goes fullscreen. Just like the picker described in step 2, the window overlay container gets removed and never comes back.

dauby88:
This issue is fixed for me by this: https://github.com/software-mansion/react-native-screens/pull/1872
I no longer need this patch.

tboba:
@dauby88 hi, thanks for the info! @rodperottoni can you confirm the issue this PR fixes no longer exists for you on the newest version of screens? Or should we still leave this PR?
I would also like to apologize for the huge delay of reviewing this PR, since in `react-native-screens` there are things that need to be cared of 🙏

PR #1415, opened 2022-04-20
refact: merge Fabric & Paper impls: ScreenView & ScreenController (1)

kkafar:
## Description

* https://github.com/software-mansion/react-native-screens/pull/1415
* https://github.com/software-mansion/react-native-screens/pull/1416
* https://github.com/software-mansion/react-native-screens/pull/1417
* https://github.com/software-mansion/react-native-screens/pull/1418

PRs mentioned just above aim to merge new Fabric components implementation to Paper.

They should be merged in ascending order (each PR has a order number associated).

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


ios/RNSConvert.h
```diff
@@ -1,6 +1,7 @@
 #ifndef RNSConvert_h
```
WoLewicki, line 1: Are those 2 first lines needed?

ios/RNSScreen.mm
```diff
@@ -824,6 +1086,16 @@ - (UIView *)view
 
 @implementation RCTConvert (RNSScreen)
 
+RCT_ENUM_CONVERTER(
```
WoLewicki, line 1089: Did it just work previously and then it stopped? Or what is the reason of adding it here?

ios/RNSScreen.mm
```diff
@@ -391,72 +599,183 @@ @implementation RNSScreen {
...
+  }
+#else
+  // The below code makes the screen view adapt dimensions provided by the system. We take these
+  // into account only when the view is mounted under RNScreensNavigationController in which case system
+  // provides additional padding to account for possible header, and in the case when screen is
+  // shown as a native modal, as the final dimensions of the modal on iOS 12+ are shorter than the
+  // screen size
+  BOOL isDisplayedWithinUINavController =
+      [self.parentViewController isKindOfClass:[RNScreensNavigationController class]];
+  BOOL isPresentedAsNativeModal = self.parentViewController == nil && self.presentingViewController != nil;
```
WoLewicki, line 769: We should check if `isPresentedAsNativeModal` should not also be needed for both architectures to provide correct frames in modal. If so, I think we can extract those 2 BOOLs outside `ifdef` since they are the same (`UINavigationController` should be changed to `RNScreensNavigationController`)

ios/RNSScreen.mm
```diff
@@ -391,72 +599,183 @@ @implementation RNSScreen {
...
 }
 
+#pragma mark - Common
+
 - (instancetype)initWithView:(UIView *)view
 {
   if (self = [super init]) {
     self.view = view;
+#ifdef RN_FABRIC_ENABLED
+    if ([view isKindOfClass:[RNSScreenView class]]) {
```
WoLewicki, line 612: Is there a case when this `if` fails?

ios/RNSScreen.mm
```diff
@@ -2,57 +2,83 @@
...
-  // the controller. In that case we ignore react layout system from managing
-  // the screen dimensions and we wait for the screen VC to update and then we
-  // pass the dimensions to ui view manager to take into account when laying out
-  // subviews
+  _controller = [[RNSScreen alloc] initWithView:self];
+  _stackPresentation = RNSScreenStackPresentationPush;
+  _stackAnimation = RNSScreenStackAnimationDefault;
+  _gestureEnabled = YES;
+  _replaceAnimation = RNSScreenReplaceAnimationPop;
+  _dismissed = NO;
```
WoLewicki, line 76: All those props below can probably be not set since they should be `NO` by default. Could you check it and remove those assignments if so?

WoLewicki (commented):
Good work :tada: I added some comments for the future PRs.

ios/RNSConvert.h
kkafar: Indeed, they are not. Removed them already.

ios/RNSScreen.mm
kkafar: I used this convert method explicitly in `updateProps` method. Other solution I see is to change of incoming cpp type of `statusBarAnimation` prop from `std::string` to cpp enum - I can then add converting method in `RNSConvert`.

ios/RNSScreen.mm
kkafar: I'm not aware of such case in our code base, but `initWithView:` instance method is exposed publicly in `RNSScreen` interface so it is not wrong idea to make sure that the right type is passed at runtime. However my guess is that this check was added in `RNSScreenController` to make sure that no one passes there a `RNSScreenView` (it required `RNSScreenComponentView` before the merge happened`. We can either extend this `if statement` for both architecture or remove it completely. What do you think?

ios/RNSScreen.mm
kkafar: I see your point: these are most likely redundant lines of code and I agree. However I see no harm in making this initialisation explicit, it only improves the readability. There should be no impact on performance / robustness. What do you think?

ios/RNSScreen.mm
kkafar: As we discussed, I noted this down and will address it in following PR

ios/RNSScreen.mm
WoLewicki: As for readability, I have no strong opinion. We can keep them like this.

ios/RNSScreen.mm
WoLewicki: If we did not check it previously, I think there is no need to do it now as well so we can remove it.

ios/RNSScreen.mm
kkafar: 20a0a92

PR #1433, opened 2022-04-29
chore: add `hideKeyboardOnSwipe` prop for Fabric (3)

kkafar:
## Description

Part of stack PR. See: 

This PR adds `hideKeyboardOnSwipe` prop for Fabric.

## Changes

See [commits](https://github.com/software-mansion/react-native-screens/pull/1433/commits).

TODO

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
@@ -144,6 +144,8 @@ class ScreenViewManager : ViewGroupManager<Screen>(), RNSScreenManagerInterface<
 
     override fun setTransitionDuration(view: Screen?, value: Int) = Unit
 
+    override fun setHideKeyboardOnSwipe(view: Screen?, value: Boolean) = Unit
+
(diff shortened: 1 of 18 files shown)


android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
```diff
@@ -144,6 +144,8 @@ class ScreenViewManager : ViewGroupManager<Screen>(), RNSScreenManagerInterface<
 
     override fun setTransitionDuration(view: Screen?, value: Int) = Unit
 
+    override fun setHideKeyboardOnSwipe(view: Screen?, value: Boolean) = Unit
```
WoLewicki, line 147: What about adding new props to interface and delegate so it works on old arch?

kkafar: As I'm testing these things on [last PR](https://github.com/software-mansion/react-native-screens/pull/1446) I'll copy the interfaces there, so I don't have to do it for every prop.

PR #1443, opened 2022-05-04
chore: add `homeIndicatorHidden` & `activityState` props for Fabric (6)

kkafar:
## Description

Part of stack PR. See: 

This PR adds `homeIndicatorHidden` & `activityState` props for Fabric.

## Changes

See [commits](https://github.com/software-mansion/react-native-screens/pull/1443/commits).

TODO

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -457,8 +457,10 @@
   [self setHideKeyboardOnSwipe:newScreenProps.hideKeyboardOnSwipe];
 
   [self setCustomAnimationOnSwipe:newScreenProps.customAnimationOnSwipe];
-  
-  [self setGestureResponseDistance:[RNSConvert gestureResponseDistanceDictFromCppStruct:newScreenProps.gestureResponseDistance]];
+
+  [self
+      setGestureResponseDistance:[RNSConvert
+                                     gestureResponseDistanceDictFromCppStruct:newScreenProps.gestureResponseDistance]];
 
 #if !TARGET_OS_TV
   if (newScreenProps.statusBarHidden != oldScreenProps.statusBarHidden) {
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -199,7 +199,7 @@
 - (void)setActivityStateOrNil:(NSNumber *)activityStateOrNil
 {
   int activityState = [activityStateOrNil intValue];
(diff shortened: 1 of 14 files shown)


ios/RNSScreen.mm
```diff
@@ -192,6 +192,19 @@ - (void)setReplaceAnimation:(RNSScreenReplaceAnimation)replaceAnimation
   _replaceAnimation = replaceAnimation;
 }
 
+// Nil will be provided when activityState is set as an animated value and we change
+// it from JS to be a plain value (non animated).
+// In case when nil is received, we want to ignore such value and not make
+// any updates as the actual non-nil value will follow immediately.
+- (void)setActivityStateOrNil:(NSNumber *)activityStateOrNil
```
WoLewicki, line 199: I think we will have to use this method correctly in `updateProps`, maybe by passing `nil` if the `activityStateOrNil` value is `-1`, but it will need to be tested extensively for sure.

kkafar: see: [08bac0d](https://github.com/software-mansion/react-native-screens/pull/1443/commits/08bac0d09525f14bcab0b630751fd7c137a5a2ba)

I decided to go this way, but it might be also a good idea to [extend this definition of RNSActivityState` enum](https://github.com/software-mansion/react-native-screens/blob/08bac0d09525f14bcab0b630751fd7c137a5a2ba/ios/RNSEnums.h#L31-L35) to:

```objc
typedef NS_ENUM(NSInteger, RNSActivityState) {
  RNSActivityStateUndefined = -1,
  RNSActivityStateInactive = 0,
  RNSActivityStateTransitioningOrBelowTop = 1,
  RNSActivityStateOnTop = 2
};
```

and then change `activityState` type from int to actually `RNSActivityState`. I believe it would make thing much easier to handle (right now any `int` value can be assigned to `activityState` prop).

WoLewicki: It would be best to just check if the current implementation works correctly and if so, I would leave it as it is for now. The values coming from JS may contain many values between 0 and 2 (see https://github.com/react-navigation/react-navigation/blob/5b13f818f3c701ce33c2a30f5dd96763069556e4/packages/stack/src/views/Stack/CardStack.tsx#L596) so I think it might not even work for now if we applied `Int32` in codegen specification.

PR #1446, opened 2022-05-05
chore: unify business logic between archs (8)

kkafar:
## Description

Part of stack PR. See: 

This PR unifies some business logic between Paper & Fabric implementations.

## Changes

See [commits](https://github.com/software-mansion/react-native-screens/pull/1446/commits).

TODO

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -16,12 +16,12 @@
 #import "RNSConvert.h"
 #else
 #import <React/RCTTouchHandler.h>
-#import "RNSScreenStack.h"
 #endif
 
 #import <React/RCTShadowView.h>
 #import <React/RCTUIManager.h>
 #import "RNSScreenStackHeaderConfig.h"
+#import "RNSScreenStack.h"
 
 @interface RNSScreenView ()
 #ifdef RN_FABRIC_ENABLED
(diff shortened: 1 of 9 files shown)


ios/RNSScreen.mm
```diff
@@ -783,12 +778,16 @@ - (void)viewDidLayoutSubviews
   BOOL isDisplayedWithinUINavController =
       [self.parentViewController isKindOfClass:[RNScreensNavigationController class]];
   BOOL isPresentedAsNativeModal = self.parentViewController == nil && self.presentingViewController != nil;
+
   if ((isDisplayedWithinUINavController || isPresentedAsNativeModal) &&
       !CGRectEqualToRect(_lastViewFrame, self.view.frame)) {
```
WoLewicki, line 783: This condition: `!CGRectEqualToRect(_lastViewFrame, self.view.frame)` is not present in Fabric and it might destroy the behavior so it would be better to check it twice since there was a reason why it was not added there.

kkafar: See: [63e4938](https://github.com/software-mansion/react-native-screens/pull/1446/commits/63e49388c70440c1ed8d27fe0e0d19a0a3f2e1ae)

PR #1451, opened 2022-05-17
feat: use recommended WindowInsetsControllerCompat for StatusBar management

kirillzyusko:
## Description

Rewrote:
- setStyle;
- setHidden;
- setColor

methods to use `WindowInsetsControllerCompat` for StatusBar management (instead of old approach with changing window flags).

## Context And Motivation

I'm currently working on [react-native-keyboard-controller](https://github.com/kirillzyusko/react-native-keyboard-controller) library which heavily relies on `WindowInsets*` API. However if the code uses old way with window flags management - it breaks `WindowInsets*` API and I can not listen for keyboard events anymore using `ViewCompat.setWindowInsetsAnimationCallback`. So in this PR I rewrote some methods to new style and used `WindowInsetsControllerCompat` instead of managing window flags.

However I couldn't find a way how to rewrite `setTranslucent` method 🤔 (though in your example app it works without any issues, I mean old and new approach can live together without any conflicts). But I think it would be better to rewrite everything to one style and I would be highly appreciate you if you could help me to find a way how to rewrite it 😊 

## Changes

- Used `WindowInsetsControllerCompat` for `StatusBar` management

## Screenshots / GIFs

<table>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
  <tr>
    <td>image</td>
    <td>image</td>
  </tr>
</table>

Before we had a lot of places using deprecated API.

I have used your example with StatusBar settings - everything works as before, I didn't notice any differences 🤷‍♂️ 

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
@@ -8,9 +8,7 @@ import android.app.Activity
 import android.content.pm.ActivityInfo
 import android.graphics.Color
 import android.os.Build
-import android.view.View
 import android.view.ViewParent
-import android.view.WindowManager
 import androidx.core.view.ViewCompat
 import androidx.core.view.WindowCompat
 import androidx.core.view.WindowInsetsCompat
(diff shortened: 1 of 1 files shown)


hirbod:
Great work @kirillzyusko. @WoLewicki might be able to help migrating setTranslucent

WoLewicki:
Thanks for submitting a PR :tada:. I briefly checked it on `Test860.tsx` to see if there are any regressions and it seems that the only change is slightly different animation of hiding status bar. Could you double check if I am right and there are no other problems with it? Current implementation was taken from `react-native` codebase (https://github.com/facebook/react-native/blob/4f1aa4d686291d05d5fcab2bdcf691bb375d6fec/ReactAndroid/src/main/java/com/facebook/react/modules/statusbar/StatusBarModule.java) so it would nice to check why your approach was not used there. As for `setTranslucent`, I am afraid it is the most tricky implementation and is connected with `headerTopInsetEnabled` prop so I would not change it if not necessary.

hirbod:
@WoLewicki I think the reason why react-native uses the old approach is - well - it's old code. The WindowInsets API is pretty "new" and allows for cool keyboard animation and observation.

kirillzyusko:
Hi @WoLewicki 

> Could you double check if I am right and there are no other problems with it? 

I've checked this example many times and I couldn't find any regressions. Regarding a different animation when status bar hiding occurs - I think it can be a case, but there is no way to control it 🤷‍♂️ 

> so it would nice to check why your approach was not used there

I think @hirbod is right. Theirs implementation was written at time, when `WindowInsets` wasn't available yet. Actually I was going to submit a PR to RN repo, once this PR will be merged and I'll find a way how to manage translucent prop properly.

So, what do you think? Is there any blockers preventing this PR to be merged?

hirbod:
@WoLewicki anything holding back here?

PR #1458, opened 2022-05-26
feat: bring Fabric to ScreenContainer component

kkafar:
## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
```diff
@@ -35,8 +35,12 @@ class ScreenViewManager : ViewGroupManager<Screen>(), RNSScreenManagerInterface<
         return Screen(reactContext)
     }
 
+    override fun setActivityState(view: Screen, activityState: Float) {
+        setActivityState(view, activityState.toInt())
```
WoLewicki, line 39: It would be good to check if it works correctly in all cases since `toInt()` does `floor` on the value so there might be some issues with it. Leaving this comment here so the information does not get lost.

PR #1470, opened 2022-06-02
feat: bring Fabric to SearchBar

kkafar:
## Description

Adopted existing search bar implementation for Fabric. 

NOTE: For now, there is a bug when using search bar (header) & scroll view (scroll view has incorrect offset -> some content might be "hidden behind header"), but is seems that it is not search bar specific - it concerns header - scrollview interaction in general.

## Changes

* Created `SearchBarNativeComponent` for codegen
* Adopted existing search bar implementation for Fabric. 
* Small refactoring of event emitting methods so for each event, there is one method for both architerctures.
* Added `Test758` from `TestsExample`

## Screenshots / GIFs

https://user-images.githubusercontent.com/50801299/175048110-23d35fe8-767b-424e-9f29-af7717ec6e62.mov

`Test758.tsx` in `FabricTestExample`

## Checklist

- [x] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


src/fabric/SearchBarNativeComponent.js
```diff
@@ -0,0 +1,62 @@
...
+
+type ChangeTextEvent = $ReadOnly<{|
+  text?: string,
+|}>;
+
+type AutoCapitalizeType = 'none' | 'words' | 'sentences' | 'characters';
+
+type NativeProps = $ReadOnly<{|
+  ...ViewProps,
+  // implement these on iOS
```
WoLewicki, line 30: Is this comment still necessary?

ios/RNSSearchBar.mm
```diff
@@ -163,26 +252,75 @@ - (void)searchBarCancelButtonClicked:(UISearchBar *)searchBar
...
+}
+#endif
+
 @implementation RNSSearchBarManager
 
 RCT_EXPORT_MODULE()
 
 - (UIView *)view
 {
+#ifdef RN_FABRIC_ENABLED
```
WoLewicki, line 319: Do we need this code here? Here we just `ifdef` the whole method: https://github.com/software-mansion/react-native-screens/blob/ba6b06bfca0537cd2183cc49e75555e93da97c1b/ios/RNSScreenStackHeaderSubview.mm#L105

ios/RNSSearchBar.mm
```diff
@@ -163,26 +252,75 @@ - (void)searchBarCancelButtonClicked:(UISearchBar *)searchBar
...
+
+  if (oldScreenProps.placeholder != newScreenProps.placeholder) {
+    [self setPlaceholder:RCTNSStringFromStringNilIfEmpty(newScreenProps.placeholder)];
+  }
+
+  if (oldScreenProps.autoCapitalize != newScreenProps.autoCapitalize) {
+    [self setAutoCapitalize:[RNSConvert UITextAutocapitalizationTypeFromCppEquivalent:newScreenProps.autoCapitalize]];
   }
+
+  // TODO: implement props with Color types
```
WoLewicki, line 291: Is there a problem with this? I think you just need to use `RCTUIColorFromSharedColor` function iirc

WoLewicki (commented):
Looks good! Please answer the comments. I think there is one more part of code that needs to be updated regarding search bar: https://github.com/software-mansion/react-native-screens/blob/ba6b06bfca0537cd2183cc49e75555e93da97c1b/ios/RNSScreen.mm#L1013. It would be good to check the issue and repro regarding it to see if adding it resolves the issue mentioned there.

ios/RNSSearchBar.mm
kkafar: Yep, as I stated in comment of [5b8be70](https://github.com/software-mansion/react-native-screens/pull/1470/commits/5b8be70954bc25bc838d6d146b2d927c5dc0e30b), and here I changed it: [0247ea1](https://github.com/software-mansion/react-native-screens/pull/1470/commits/0247ea18fc35c098d04e71c1445837520c1e49a9)

ios/RNSSearchBar.mm
kkafar: I just did not know out of hand how colours work (they're passed as strings in JS, but I guess they're internally converted somewhere) and decided to postpone implementing these props (but I will add them in this PR). Thanks for telling me the right function.

src/fabric/SearchBarNativeComponent.js
kkafar: I updated the TODO messages, but did not delete them completely as they are of use for me. I'll make sure to keep them in sync with the work progress and that before this PR is merged they are deleted.

PR #1479, opened 2022-06-13
feat: Shared elements transitions

fdecampredon:
## Description

Basic implementation of Shared element transitions, the API is inspired by [wix/react-native-navigation](https://github.com/wix/react-native-navigation) but only supports basic transition of size and position.
The `align` part of the api is inspired by [IjzerenHein/react-native-shared-element](https://github.com/IjzerenHein/react-native-shared-element).

I will experiment more with my fork on a real project in the following days and might improve this solution or add missing features if something important is missing for our project.

## Changes

- Updated types to allow describing shared element transitions (available only for ScreenStack)
- Implemented Shared elements transitions on IOS native part
- Implemented Shared elements transitions on Android native part

On both platform shared elements transitions consist of : 
- making a "snapshot" of the initial element
- hiding the initial element and the target element
- rendering and animating the "snapshot"
- restoring the state of the initial element and the target element
- removing the snapshot

Surprisingly, the android implementation was a chalenge for me, even if the Android SDK supports shared elements out of the box, using provided android transition like [ChangeBounds](https://developer.android.com/reference/android/transition/ChangeBounds) leads to strange bugs and behaviors, especially with ReactImageView.

## Screenshots / GIFs
Simulator Screen Recording - iPhone 13 - 2022-06-13 at 17 44 27

I added a TestCase on both TestExample and FabricTestExample, the FabricTestExample project on Android was quite buggy on my side (even without my changes) and I am unable to test it properly at the moment.

## Checklist

- [x] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


fdecampredon:
Just to sum up our research (if some people are interested, what I doubt) we finally decided, for the moment, to give up on this work and to use a  custom solution for our application use case.

Shared Element Transitions are a lot more tricky than what I thought when I started to work on this, and there is a lot to take into account depending on your use case than just animating some views.
For example when to make the new screen interactive, avoiding some content flash when the shared element is removed, etc...

In the end I feel that it's really hard to design a generic solution, at least if you do more than using some image/text transitions, and that a well made custom animation (native or with react-native-reanimated), specifically designed for your use case, will always achieves a better user experience.

I'll stop working on this and I'm closing the draft.

PR #1491, opened 2022-06-27
fix: `hideHeaderIfNecessary` for Fabric

kkafar:
## Description

Added the fix introduced here: 

* https://github.com/software-mansion/react-native-screens/pull/1153

for Fabric.

## Changes

* Moved `hideHeaderIfNecessary` method to common section
* Added fix for Fabric: when returning from second screen application would crash due to order of view unmounting (child-first) on Fabric (calling `view.reactSubviews[0]` would cause application to crash)

* Moved `Test1153`. See: https://github.com/software-mansion/react-native-screens/pull/1153

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -1078,7 +1078,7 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
 
     // we need to check whether reactSubviews array is empty, because on Fabric child nodes are unmounted first ->
     // reactSubviews array may be empty
-    if (currentIndex > 0 && [self.screenView.reactSubviews count] &&
+    if (currentIndex > 0 && [self.screenView.reactSubviews count] > 0 &&
         [self.screenView.reactSubviews[0] isKindOfClass:[RNSScreenStackHeaderConfig class]]) {
       UINavigationItem *prevNavigationItem =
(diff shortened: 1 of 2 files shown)


ios/RNSScreen.mm
```diff
@@ -1107,7 +1076,10 @@ - (void)hideHeaderIfNecessary
   if (@available(iOS 13.0, *)) {
     NSUInteger currentIndex = [self.navigationController.viewControllers indexOfObject:self];
 
-    if (currentIndex > 0 && [self.screenView.reactSubviews[0] isKindOfClass:[RNSScreenStackHeaderConfig class]]) {
+    // we need to check whether reactSubviews array is empty, because on Fabric child nodes are unmounted first ->
+    // reactSubviews array may be empty
+    if (currentIndex > 0 && [self.screenView.reactSubviews count] &&
```
WoLewicki, line 1081:
```suggestion
    if (currentIndex > 0 && [self.screenView.reactSubviews count] > 0 &&
```

seems more straightforward

WoLewicki (approved):
I added one comment, other than that it looks good. Good job :tada:

PR #1496, opened 2022-06-28
fix: reset values after animations and make snapshot only when necessary

WoLewicki:
## Description

PR adding logic of resetting screens props after transition animations since they are recycled afterwards and often have wrong alpha/transform then. Also added application of snapshot only if `window` is present which should mean that screen has been removed via JS navigation actions.

Mainly `Test887.tsx` with going to different modal and non-modal screens to see if they all work correctly.

## Checklist

- [ ] Ensured that CI passes


kkafar (approved):
lgtm

Issue #1501, opened 2022-07-01
version@3.14.0 ios build error: RNSScreenViewEvent.h:2:9: 'React/RCTEventDispatcherProtocol.h' file not found


cengit:
## Description

version@3.14.0 ios build error: 

```sh
❌ /Users/runner/work/1/s/node_modules/react-native-screens/ios/RNSScreenViewEvent.h:2:9: 'React/RCTEventDispatcherProtocol.h' file not found

#import <React/RCTEventDispatcherProtocol.h>

^

▸ Compiling RNSScreenStackHeaderSubview.mm

Skipping duplicate build file in Copy Bundle Resources build phase: /Users/runner/work/1/s/node_modules/performance_frontend/src/font/Montserrat-SemiBold.ttf (in target 'eBriefcase' from project 'eBriefcase')

** ARCHIVE FAILED **

The following build commands failed:

CompileC /Users/runner/Library/Developer/Xcode/DerivedData/eBriefcase-fpyythrpiwdiswbagvblncwborcj/Build/Intermediates.noindex/ArchiveIntermediates/eBriefcase/IntermediateBuildFilesPath/Pods.build/Release-iphoneos/RNScreens.build/Objects-normal/arm64/RNSScreenViewEvent.o /Users/runner/work/1/s/node_modules/react-native-screens/ios/RNSScreenViewEvent.mm normal arm64 objective-c++ com.apple.compilers.llvm.clang.1_0.compiler (in target 'RNScreens' from project 'Pods')

(1 failure)
```

I guess this may be library version not matched, but don't know which library. can you please give some advice.

## Steps To Reproduce

1. use react-native-screens version@3.14.0

## Platform

- [x] iOS

## Architecture

- [x] Paper
- [ ] Fabric

## Workflow

- [x] Managed workflow
- [ ] Bare workflow

## Package versions

| package                                | version |
| -------------------------------------- | ------- |
| react-native                           |  0.63       |
| @react-navigation/native               |    3.8.4     |
| @react-navigation/native-stack         |  5.0.4       |
| react-native-screens                   |   3.14.0      |
| react-native-safe-area-context         |  3.4.1       |
| react-native-gesture-handler           |  1.10.3       |
| react-native-reanimated                |   1.13.4      |

kkafar:
Hi @cengit!
[...]

PR #1502, opened 2022-07-01
fix: bring back iOS Paper compatibility with `react-native@0.63`

kkafar:
## Description

Fixes: https://github.com/software-mansion/react-native-screens/issues/1501

#1472 introduced `RNSScreenViewEvent` which uses `React/RCTEventDispatcherProtocol.h` header. Before `react-native@0.64` this header was named `React/RCTEventDispatcher.h`, so now, when using `react-native@0.63.x` with `react-native-screens@3.14.0` the build fails as the header can not be found. 

## Changes

* As `RNSScreenViewEvent` is used only on Fabric (& on Fabric we support `react-native@0.69+`, I "ifdefed" it, so on Paper it is not defined.

1. `npx react-native init TestProject --version 0.63.2`
2. `yarn add react-native-screens@3.14.0`
3. Try to build iOS

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenViewEvent.h
+++ b/ios/RNSScreenViewEvent.h
@@ -1,3 +1,5 @@
+#ifdef RN_FABRIC_ENABLED
+
 #import <React/RCTBridge+Private.h>
 #import <React/RCTEventDispatcherProtocol.h>
 
@@ -10,3 +12,5 @@
                      goingForward:(int)goingForward;
 
 @end
+
+#endif // RN_FABRIC_ENABLED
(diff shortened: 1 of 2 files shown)


kkafar:
I'm not merging it yet, as more issues occurred after this fix (:D), however it is not clear yet whether they are related to `react-native-screens`

kkafar:
Closed as we dropped support for < 0.64 since 3.14

PR #1503, opened 2022-07-01
fix: CI

kkafar:
## Description

Improved worklows logic to tackle some issues with outdated Podfile.lock & gradle cache.

## Changes

* `ios-build-test-fabric`: introduced mechanism such that, when first attempt to run `pod install` fails, `rm -fr Pods build Podfile.lock` is executed and second attempt to install the pods follows on. *This is awful* but I don't see different solution for this moment.
* updated version of `react-native-gradle-plugin` - required `reactNativeDir` prop did not exist in 0.0.3 -> build crashed

See that CI here passes

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


ios/RNSScreenViewEvent.mm
```diff
@@ -21,6 +21,7 @@ - (instancetype)initWithEventName:(NSString *)eventName
   if ((self = [super init])) {
     _eventName = [eventName copy];
     _viewTag = reactTag;
+
```
WoLewicki, line 24: why blank line here?

kkafar: As I stated in this commit message: "[random change to trigger the CI](https://github.com/software-mansion/react-native-screens/pull/1503/commits/00c1e6e31b5e695eb7e83a3d6c792251e46017e0)" - so that I don't have to trigger workflows manually. It will be obviously deleted when the job is done (I will just remove this commit).

PR #1507, opened 2022-07-04
fix: default value for `gestureResponseDistance` 

kkafar:
## Description

Moved code that sets default value for `gestureResponseDistance` prop on `Screen` component "deeper", so that, when 3rd party library uses our primitives it does not have to set the default value by itself. Behaviour is unchanged.

## Changes

* The code was moved to `index.native.tsx`

`Test1296` -- see that `gestureResponseDistance` prop works just fine

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/src/index.native.tsx
+++ b/src/index.native.tsx
@@ -253,13 +253,6 @@ class Screen extends React.Component<ScreenProps> {
         ...props
       } = rest;
 
-      const defaultedGestureResponseDistance = {
-        start: gestureResponseDistance?.start ?? -1,
-        end: gestureResponseDistance?.end ?? -1,
-        top: gestureResponseDistance?.top ?? -1,
-        bottom: gestureResponseDistance?.bottom ?? -1,
-      };
-
       if (active !== undefined && activityState === undefined) {
         console.warn(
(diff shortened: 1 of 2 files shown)


src/index.native.tsx
```diff
@@ -249,9 +249,17 @@ class Screen extends React.Component<ScreenProps> {
...
+        gestureResponseDistance,
         ...props
       } = rest;
 
+      const defaultedGestureResponseDistance = {
+        start: gestureResponseDistance?.start ?? -1,
+        end: gestureResponseDistance?.end ?? -1,
+        top: gestureResponseDistance?.top ?? -1,
+        bottom: gestureResponseDistance?.bottom ?? -1,
+      };
```
kacperkapusciak, line 261: This could be inlined to avoid creating an unnecessary variable. Otherwise LGTM

kacperkapusciak (approved):

src/index.native.tsx
kkafar: See: [77883f](https://github.com/software-mansion/react-native-screens/pull/1507/commits/77883ff9eec7f0527657bff171492cc6b5ab5e69)

kacperkapusciak (approved):

PR #1509, opened 2022-07-04
fix: get properties from top-most screen on swipe

kacperkapusciak:
## Description

This PR fixes an issue on which `swipeDirection` was required to be set on both on current and previous screen because the value after one tick was incorrectly taken from the previous screen.

Thanks @WoLewicki for the solution 👏 

### Before

https://user-images.githubusercontent.com/39658211/177128499-ef973722-31af-4587-8d47-e225a45989db.mov

### After

https://user-images.githubusercontent.com/39658211/177128248-b827e713-f8e8-4675-a45b-114003cc3c3f.mov

TestsExample/Test1509.tsx

## Checklist


WoLewicki (approved):
Good job :tada:

hirbod:
@WoLewicki do you still remember the fullScreen swipe issue I've showed you at App.js conf?
When you swipe on the edge, it will trigger the default swipe gesture instead of the custom. 

Maybe thats a good point to fix this as well :D

WoLewicki:
@hirbod good point! I'll look into it. Do you have a quick repro for this and could paste it even here?

hirbod:
@WoLewicki this should work
https://snack.expo.dev/@hirbod/swipe-pop-issue

I could not reproduce this in Expo Go though.

kkafar:
Hi @hirbod 
I'm looking at [your snack](https://snack.expo.dev/@hirbod/swipe-pop-issue) right now & that's right :D Depending on where your gesture starts it is handled by different gesture recognisers. When you start the gesture near the edge native animation is being played, otherwise animation defaults to 'simple_push'. So if you want to have the same effect on whole screen you should try to set `animation: "simple_push"`. 
However I see that this is not the best behaviour, and we could maybe set the animation to 'simple_push' by default.

WoLewicki:
But still the native header transition is only possible with the default gesture recognizer 😢

hirbod:
Yeah the header is not that bad though, the shadow is also not a big deal breaker - just a visual thing

PR #1512, opened 2022-07-05
fix: handle multiple gesture recognizers

### LINKED ISSUE #1510: [bug][ios] 3.14+ breaks all ScrollViews/FlatList when inside a stack with fullScreenGestureEnabled
## Description

Hi @WoLewicki, I hold up to my promise to find every bug in RNS :D

After the update to 3.14.1 it is impossible to scroll in a view (ScrollView, FlatList etc) with `fullScreenGestureEnabled`.

The gesture recognizer takes total control and does not allow scrolling anymore. If you set `fullScreenGestureEnabled: false`, it works.

The problem does not occur with 3.13.1. I suspect the regression comes from https://github.com/software-mansion/react-native-screens/pull/1370

### Screenshots

| 3.13.1 (works)                                |  3.14.1 (broken) |
| <video src="https://user-images.githubusercontent.com/504909/177221531-0e25fcc2-d616-40fb-b0f8-2b2f81ab02a0.mp4" />        |   <video src="https://user-images.githubusercontent.com/504909/177221439-e20ac859-95d2-43c6-86bd-043f1f46fa45.mp4" />    |

## Steps To Reproduce

1. Create a Stack Screen with `fullScreenGestureEnabled`.
2. Add ScrollView inside of it and navigate to it
3. Try to scroll - impossible. The horizontal pop gesture kicks in

### Expected behavior

Should not break scrolling. Also messes around with react-native-gesture-handlers `activeOffsetY`. If set to -20, 20, it does also not work anymore, had to change it to -5, 5.

### Actual behavior

Scrolling impossible with `fullScreenGestureEnabled`
Breaks scrolling

## Reproduction

https://snack.expo.dev/@hirbod/scrollview

Just drop the ScrollView inside a Screen with fullScreenGestureEnabled set too true. Since Snack does not Support 3.14.1, it's not a real repro.

## Platform

- [x] iOS
- [ ] Android
- [ ] Web
- [ ] Windows
- [ ] tvOS

## Architecture

- [x] Paper
- [ ] Fabric

## Workflow

- [ ] Managed workflow
- [x] Bare workflow

## Package versions

| react-native                           |  0.68.2       |
| @react-navigation/native               |    6.0.10     |
| @react-navigation/native-stack         |      6.6.2   |
| react-native-screens                   |     3.14.1    |
| react-native-safe-area-context         |    4.3.1     |
| react-native-gesture-handler           |  2.4.2       |
| react-native-reanimated                |    2.8.0     |
| expo                                   |    SDK 45 (custom dev client)     |

kkafar:
## Description

Fixes #1510

Fixed & improved `RNSScreenStackView` logic regarding `UIGestureRecognizerDelegate` protocol.

## Changes

* Added `gestureRecognizer:shouldReceiveEvent:`. Events for particular gesture recognisers are now filtered before `gestureRecognizerShouldBegin` method is called.
* Any `UIScreenEdgePanGestureRecognizer` must fail for scroll view gesture recogniser to begin.
* When `fullScreenSwipeEnabled == true` scroll view gesture recogniser is prioritised before our `RNSPanGestureRecognizer`.
* Simplified logic in `gestureRecognzerShouldBegin`.

Now, when adding another gesture recogniser with screen stack as delegate one must make sure that events are properly filtered.

## Screenshots / GIFs

### Before

As on screen recordings attached in #1510

### After

```
fullScreenSwipeEnabled: true,
gestureEnabled: true,
stackAnimation: 'simple_push',
customAnimationOnSwipe: true,
```

https://user-images.githubusercontent.com/50801299/177521059-56d015c7-763f-4cb7-a6d3-030a48bb8e2b.mov

```
fullScreenSwipeEnabled: false,
gestureEnabled: true,
stackAnimation: 'simple_push',
customAnimationOnSwipe: false,
```

https://user-images.githubusercontent.com/50801299/177521435-4e4bcd48-de22-4ace-9355-32cf99f2ba3d.mov

`Test1072`

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStack.mm
+++ b/ios/RNSScreenStack.mm
@@ -600,11 +600,6 @@
 {
   RNSScreenView *topScreen = (RNSScreenView *)_controller.viewControllers.lastObject.view;
 
-  if (![topScreen isKindOfClass:[RNSScreenView class]] || !topScreen.gestureEnabled ||
-      _controller.viewControllers.count < 2) {
-    return NO;
-  }
-
 #if TARGET_OS_TV
   [self cancelTouchesInParent];
   return YES;
@@ -817,6 +812,19 @@
   }
   UIScrollView *scrollView = (UIScrollView *)gestureRecognizer.view;
   return scrollView.panGestureRecognizer == gestureRecognizer;
+}
+
+// RNSScreenStackView is a UIGestureRecognizerDelegate for two types of gesture recognizers:
+// RNSPanGestureRecognizer, RNSScreenEdgeGestureRecognizer
+- (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldReceiveEvent:(UIEvent *)event
+{
+  RNSScreenView *topScreen = (RNSScreenView *)_controller.viewControllers.lastObject.view;
+
+  if (![topScreen isKindOfClass:[RNSScreenView class]] || !topScreen.gestureEnabled ||
+      _controller.viewControllers.count < 2) {
+    return NO;
+  }
+  return YES;
 }
 
 - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
--- a/ios/RNSScreenStack.mm
+++ b/ios/RNSScreenStack.mm
@@ -604,11 +604,10 @@
   [self cancelTouchesInParent];
   return YES;
 #else
-  if (topScreen.fullScreenSwipeEnabled) {
-    // we want only `RNSPanGestureRecognizer` to be able to recognize when
-    // `fullScreenSwipeEnabled` is set, and we are in the bounds set by user
-    if ([gestureRecognizer isKindOfClass:[RNSPanGestureRecognizer class]] &&
-        [self isInGestureResponseDistance:gestureRecognizer topScreen:topScreen]) {
+  // RNSPanGestureRecognizer will receive events iff topScreen.fullScreenSwipeEnabled == YES;
+  // Events are filtered in gestureRecognizer:shouldReceiveEvent: method
+  if ([gestureRecognizer isKindOfClass:[RNSPanGestureRecognizer class]]) {
+    if ([self isInGestureResponseDistance:gestureRecognizer topScreen:topScreen]) {
       _isFullWidthSwiping = YES;
       [self cancelTouchesInParent];
       return YES;
@@ -616,6 +615,7 @@
     return NO;
   }
 
+  // Now we're dealing with RNSScreenEdgeGestureRecognizer (or _UIParallaxTransitionPanGestureRecognizer)
   if (topScreen.customAnimationOnSwipe && [RNSScreenStackAnimator isCustomAnimation:topScreen.stackAnimation]) {
     if ([gestureRecognizer isKindOfClass:[RNSScreenEdgeGestureRecognizer class]]) {
       // if we do not set any explicit `semanticContentAttribute`, it is `UISemanticContentAttributeUnspecified` instead
@@ -634,10 +634,8 @@
     if ([gestureRecognizer isKindOfClass:[RNSScreenEdgeGestureRecognizer class]]) {
       // it should only recognize with `customAnimationOnSwipe` set
       return NO;
-    } else if ([gestureRecognizer isKindOfClass:[RNSPanGestureRecognizer class]]) {
-      // it should only recognize with `fullScreenSwipeEnabled` set
-      return NO;
     }
+    // _UIParallaxTransitionPanGestureRecognizer (other...)
     [self cancelTouchesInParent];
     return YES;
   }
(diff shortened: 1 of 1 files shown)


hirbod:
@kkafar I would like to take the opportunity to talk about a general issue with multiple recognizers.
@intergalacticspacehighway and I use this patch regular (and adapting) once RNS releases a new version.

https://github.com/intergalacticspacehighway/pager-view-ios-swipe-back-fix/blob/last-page-scroll-fix/patches/react-native-screens+3.13.1.patch

This patch allows us to use https://github.com/callstack/react-native-pager-view and https://github.com/satya164/react-native-tab-view in a way that also Twitter does.

While having `fullScreenGestureEnabled` enabled, we are still able to swipe the tabs and when reached the first tab to also activate the pop gesture (edge + fullscreen pop). This PR also adds a `shouldRecognizeSimultaneouslyWithGestureRecognizer` which breaks our patch or at least made it more complicated.

Do you think there is any chance to have this somehow supported OOTB by RNS? I understand that RNS would most likely not add ReactNativePageView a dependency, but maybe you are aware of a more generic way to have this supported.

Here is a video demonstrating what this patch does. Without this patch, fullScreenGestureEnabled would just take over and break swiping. Using only edge swiping, it will even break and make it impossible to swipe back. The edge issue has been fixed lately but fullScreen need this patch.

https://user-images.githubusercontent.com/504909/177304435-2f46e402-ca5d-48b9-9b5e-40b851d01b47.mp4

hirbod:
I pulled in your changes and the bugs I reported are gone. Unfortunately, it did not "fix" the Tab Pager Fullscreen Swiping, so I will need to patch it

kkafar:
Hi @hirbod 
I believe we are not willing to depend on RNPageView, and looking at your patch implementation it seems to me that it is highly related to how page view is implemented.
There is something I've noted down and we will implement at some point in the future: activating the pop gesture when scrollview has `offset == 0`. It would also allow us to dismiss the screen with scroll view with vertical gesture (`swipeDirection == 'vertical'`). However I dunno whether it would satisfy you.

hirbod:
I just upgraded to SDK 46 (custom dev client) and it looks like this change messes around with plenty of gestures. Like I can't even close Gorhoms Bottom Sheet anymore because the fullscreenSwipe messes with it. I still have to investigate but it specially has issues with ScrollViews and FlatLists

hirbod:
False alarm. It was RNGH which caused the issue. We filed an issue there

timurridjanovic:
> There is something I've noted down and we will implement at some point in the future: activating the pop gesture when scrollview has `offset == 0`. It would also allow us to dismiss the screen with scroll view with vertical gesture (`swipeDirection == 'vertical'`).

@kkafar Is this something that is still planned or has this been solved? I'm trying to figure out how I can swipe away a vertical modal that has a scrollview in it.

timurridjanovic:
@kkafar Sorry to be annoying with this, but I would like to follow up on my question. I'd just like to know if the feature you mentioned (quoted in my previous comment) is planned in a recent future? Thx

kkafar:
HI @timurridjanovic, 
sorry for the delayed response. This is still something that I have in my TODO list. Best estimation I can give at this moment is: "it will be implemented in near future"... But will definitely notify you when it's done.

timurridjanovic:
@kkafar Thank you!

Madumo:
>There is something I've noted down and we will implement at some point in the future: activating the pop gesture when scrollview has offset == 0. It would also allow us to dismiss the screen with scroll view with vertical gesture (swipeDirection == 'vertical').

@kkafar do you know if there's any workaround possible to implement this behaviour currently? The "swipe down to dismiss" feature works with those options:

```
    presentation: 'card',
    animation: 'slide_from_bottom',
    fullScreenGestureEnabled: true,
    customAnimationOnGesture: true,
    gestureDirection: 'vertical',
```

But i can't make it work when there's a scrollview that swallows the gesture events... Is there any way to propagate those events manually when i know that the scrollOffset is 0? Or do you have any idea on how you would implement this? I might take a crack at it.

kkafar:
Hey @Madumo 
This would surely require native changes. I would start playing around with this method:

```
- (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
    shouldRecognizeSimultaneouslyWithGestureRecognizer:(UIGestureRecognizer *)otherGestureRecognizer
```

and others affected in this PR. I would check whether scollview is attached. If it is the case - check its offset. If it is < / <= than some threshold (0?) allow our gesture recognizer to receive the event (recognise simultaneously with the scroll view gesture recognizer). Then in gesture handling method (I do not recall whether it is currently overrided) trigger screen dismissal if the event comes with y offset > 0.
 
Any PR is welcome, however as you brought it to my attention I might take crack at it once I finish my current task.

RaphBlanchet:
Hey @kkafar, I work with @Madumo and we looked into fixing this issue. Following your recommendations, we ended up with this patch that seems to be fixing the problem for us: https://gist.github.com/RaphBlanchet/d260376370b39ee5c9ba7e4c45bcd939

We can now drag the screen down to dismiss it, even if the gesture is from the underlying ScrollView. We are even able to keep the bounces of the ScrollView by cancelling the ScrollView's gesture when we detect we are doing a dismiss.

If this patch looks good to you, I will do a PR to include this fix!

Also, regarding the flag you mentioned, I think that checking the already existing `fullScreenGestureEnabled` and `gestureDirection` would be more than enough, instead of adding a new one. This is a pretty common behaviour on iOS to dismiss a screen even if we scroll from inside a scroll view, so this patch would be more like a "bug fix" than a new feature. What do you think?

kkafar:
@RaphBlanchet, took a look at the gist and it looks like it could work. Definitely open up a PR!

RaphBlanchet:
@kkafar I just opened this PR: https://github.com/software-mansion/react-native-screens/pull/1902

Sorry for the delay, just found some time to do it, hope it helps! :)

PR #1526, opened 2022-07-13
fix: iOS multiple windows scenario

alexbumbu:
Using RNSFullWindowOverlay in multiple windows scenarios is causing weird presentation issues and it's causing a crash for applications which use UISceneDelegate. This is a fix I propose for these issues.
For fixing this I ended up using keyWindow, even though it's deprecated starting with iOS 13. After doing some research I found that filtering for the foreground active scene, and retrieve the keyWindow this way won't cover all the cases (more details in this post: https://teng.pub/technical/2021/11/9/uiapplication-key-window-replacement).

### DIFF
--- a/ios/RNSFullWindowOverlay.mm
+++ b/ios/RNSFullWindowOverlay.mm
@@ -83,7 +83,7 @@
 
 - (void)show
 {
-  UIWindow *window = RCTSharedApplication().delegate.window;
+  UIWindow *window = RCTSharedApplication().keyWindow;
   [window addSubview:_container];
 }
 


WoLewicki:
Seems that RN also uses this prop in their implementation: https://github.com/facebook/react-native/blob/8bd3edec88148d0ab1f225d2119435681fbbba33/React/CoreModules/RCTDevLoadingView.mm#L117 so I think it is ok to do it like that. Did you test if it does not destroy any of the current behaviors (https://github.com/software-mansion/react-native-screens/blob/cc5de7ba28e90ecce9bbb3464ef0a625eaae032f/TestsExample/src/Test1096.tsx is good for testing it)? If so, I think we can merge it.

tboba:
Hi @alexbumbu, thanks for creating this PR!

I think your solution is great, but better solution would be to choose the keyWindow from RCTKeyWindow(). I have created a separate PR for that (just to match the description with our template and not ingerent into your changes). If you still have a reproducer that shows the scenario with multiple windows, could you check if this PR works properly for you?

PR #1531, opened 2022-07-18
fix: invalid dismissCount value when closing modals

### LINKED ISSUE #1473: Pressing on Modal Background causing navigation to pop two routes.
## Description

https://user-images.githubusercontent.com/14292004/171832187-3ab0c483-1cf3-4b08-8b56-0e073aa999bb.mp4

## Steps To Reproduce

1. Navigate To Screen
2. Navigate To Modal in the same Stack
3. Press modal background.

### Expected behavior
Should pop to screen behind 

### Actual behavior
Pop x2 

## Reproduction
The issue is introduced in https://github.com/software-mansion/react-native-screens/pull/830
`dismissCount` seems to be incorrect when dealing with modals.

## Platform
- [ x] iOS
- [ ] Android
- [ ] Web
- [ ] Windows
- [ ] tvOS

## Workflow

- [x ] Managed workflow
- [ ] Bare workflow

## Package versions

| react-native                           | 0.68.1|
| @react-navigation/native               |6.0.10|
| @react-navigation/native-stack         |6.6.1|
| react-native-screens                   |3.11.1|
| react-native-safe-area-context         |         |
| react-native-gesture-handler           |         |
| react-native-reanimated                |         |
| expo                                   |         |

kkafar:
## Description

Fixes #1473 

When modals are pushed onto the stack, in view hierarchy they end up not being descendants of `RNSScreenNavigationController` -> [When a modal is dismissed & we ask `RNSScreen` with modal view for its navigation controller](https://github.com/software-mansion/react-native-screens/blob/cc5de7ba28e90ecce9bbb3464ef0a625eaae032f/ios/RNSScreen.mm#L753-L755) (`[self navigationController]`) we receive null. This caused `getIndexOfView:` method to return `-1` (as the view was not found) what resulted in wrong value of `_dismissCount`. 

## Changes

Added a check for case when `navigationController` of `RNSScreen` is null - now in such case we default to `_dismissCount == 1`

### Before

https://user-images.githubusercontent.com/50801299/179496086-0064e0bd-bf0e-4475-8a46-01432c75ec73.mov

### After

https://user-images.githubusercontent.com/50801299/179495969-403d0cd0-ee9f-4439-8d3e-febf6f4c6498.mov

`Test1473` in test example apps.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/FabricTestExample/App.js
+++ b/FabricTestExample/App.js
@@ -78,6 +78,7 @@ import Test1260 from './src/Test1260';
 import Test1296 from './src/Test1296';
 import Test1299 from './src/Test1299';
 import Test1419 from './src/Test1419';
+import Test1473 from './src/Test1473';
 import Test1476 from './src/Test1476';
 
 enableFreeze(true);
--- a/FabricTestExample/src/Test1473.tsx
+++ b/FabricTestExample/src/Test1473.tsx
@@ -1,0 +1,78 @@
+import React from 'react';
+import {StyleSheet, View, Text, Button} from 'react-native';
+import {NavigationContainer} from '@react-navigation/native';
+import {createNativeStackNavigator} from 'react-native-screens/native-stack';
+
+const Stack = createNativeStackNavigator();
+
+function First(props) {
+  return (
+    <View style={styles.container}>
+      <Text>This is the first screen</Text>
+      <Button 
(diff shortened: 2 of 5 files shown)


hirbod:
Thanks you very much for fixing this, even though I did not provide a repro!

PR #1585, opened 2022-09-10
chore(Android): add autolinking on Fabric

kkafar:
## Description

This PR integrates the library with Android autolinking mechanism 🥳 

Up to this moment, `react-native-screens` had custom linking mechanism which builded `rnscreens_modules` & `rnscreens_common` shared libraries and loaded them during runtime, leveraging JNI's `OnLoad` method to add our custom component descriptors providers to `ComponentDescriptorProviderRegistry` instance.  Since this PR forward, this custom loading mechanism is no longer needed as it shall be done by RNCLI autolinking. 

See my comments below this PR description to see comments on fragile changes.

**IMPORTANT**: *This PR requires `@react-native-community/cli` `9.1.0` or newer (support for `cmakeListsPath` in `react-native.config.js` & Android autolinking). This version is shipped with `react-native` since `0.70.0` => with merging of this PR we drop support for older `react-native` versions on Fabric.*

## Changes

* Removed custom extrenalNativeBuild configuration from library's gradle build.
* Removed custom loading mechanism that was used to load `rnscreen_modules` shared library.
* Removed code responsible for building `rnscreens_modules` SO.
* Added custom CMake build for `react_codegen_rnscreens` target (replacement of `rnscreen_modules`). See (1) below.
* Added RNCLI configuration. Note that now all library's component descriptors **must** be enumerated manually. See (2) below.

(1)

Autolinking expects `react_codegen_rnscreens` to be the name of the library. The name is autogenerated.

I believe it would be possible to create separate target containing only custom cpp code with adding a line like: 

```cmake
target_link_libraries(${CMAKE_PROJECT_NAME} rnscreens_modules)
```
in our custom `CMakeLists.txt` so `react_cdegen_rnscreens` contains only codegened symbols but I did not test that.

(2)

If component has `interfaceOnly` codegen config option set to `true` codegen does not generate part of the required code, including component descriptor (CD). This is the case for `ScreenNativeComponent`. This leads to its manually created CD not being picked up by autolinking ==> we need to somehow inform RNCLI that this CD exists. This is done by specifying `componentDescriptors` option in `react-native.config.js` file. AFAIK it does not allow to extend set of codegen'd CDs, it only allows to overwrite it, hence we need to specify all CDs manually.

Build any of FabricExample applications in the repository (or create fresh React Native app) and see that they work properly

## Checklist

- [x] Ensured that CI passes


common/cpp/react/renderer/components/rnscreens/RNSScreenState.h
```diff
@@ -32,6 +32,7 @@ class RNSScreenState final {
   const Size frameSize{};
 
 #ifdef ANDROID
+   __attribute__((visibility("default")))
   folly::dynamic getDynamic() const;
```
kkafar, line 36: For some reason this symbol was not visible during linkage of final app's SO as it was not exported from `libreact_codegen_rnscreens.so` library (when investigated with `nm` tool it was marked as a local symbol). Adding this attribute resolves the issue.

kkafar (commented):
Some implementation notes below

android/src/main/jni/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
+cmake_minimum_required(VERSION 3.13)
+set(CMAKE_VERBOSE_MAKEFILE on)
+
+set(RNS_ANDROID_DIR ${CMAKE_CURRENT_SOURCE_DIR}/../../..)
+set(RNS_COMMON_DIR ${RNS_ANDROID_DIR}/../common/cpp)
+set(RNS_GENERATED_DIR ${RNS_ANDROID_DIR}/build/generated)
+set(RNS_GENERATED_JNI_DIR ${RNS_GENERATED_DIR}/source/codegen/jni)
+set(RNS_GENERATED_REACT_DIR ${RNS_GENERATED_JNI_DIR}/react/renderer/components/rnscreens)
```
WoLewicki, line 8: Would it be possible to extract `rnscreens` to a variable so that library authors could change only that variable for their usecase?

android/src/main/jni/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
+cmake_minimum_required(VERSION 3.13)
+set(CMAKE_VERBOSE_MAKEFILE on)
+
+set(RNS_ANDROID_DIR ${CMAKE_CURRENT_SOURCE_DIR}/../../..)
+set(RNS_COMMON_DIR ${RNS_ANDROID_DIR}/../common/cpp)
```
tomekzaw, line 5: I'd like to suggest renaming `RNS_` to `RNSCREENS_`

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.h
```diff
@@ -4,13 +4,14 @@
...
+#include <jsi/jsi.h>
 
 namespace facebook {
 namespace react {
 
-extern const char RNSScreenComponentName[];
+JSI_EXPORT extern const char RNSScreenComponentName[];
 
-class RNSScreenShadowNode final : public ConcreteViewShadowNode<
+class JSI_EXPORT RNSScreenShadowNode final : public ConcreteViewShadowNode<
```
tomekzaw, line 14: I'd like to suggest defining
```cpp
#define RNSCREENS_EXPORT __attribute__((visibility("default")))
```
and using `RNSCREENS_EXPORT` instead.

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
```diff
@@ -6,7 +6,7 @@
 
 namespace facebook {
 namespace react {
-
+
```
tomekzaw, line 9: Unnecessary whitespace here.

android/src/main/jni/rnscreens.h
```diff
@@ -0,0 +1,15 @@
...
+
+#include <ReactCommon/JavaTurboModule.h>
+#include <ReactCommon/TurboModule.h>
+#include <jsi/jsi.h>
+#include <react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h>
+
+namespace facebook {
+namespace react {
+
+JSI_EXPORT
```
tomekzaw, line 11: Is `JSI_EXPORT` used here in some kind of official guide? If not, why not `RNSCREENS_EXPORT`?

android/src/main/jni/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
...
+  -fexceptions
+  -frtti
+  -std=c++17
+  -Wall
+  -Wpedantic
+  -Wno-gnu-zero-variadic-macro-arguments
+)
+
+file(GLOB rnscreens_SRCS CONFIGURE_DEPENDS *.cpp ${RNS_COMMON_DIR}/react/renderer/components/rnscreens/*.cpp)
+file(GLOB rnscreens_codegen_SRCS CONFIGURE_DEPENDS ${RNS_GENERATED_REACT_DIR}/*cpp)
```
tomekzaw, line 20:
```suggestion
file(GLOB rnscreens_codegen_SRCS CONFIGURE_DEPENDS ${RNS_GENERATED_REACT_DIR}/*.cpp)
```

android/src/main/jni/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
+cmake_minimum_required(VERSION 3.13)
+set(CMAKE_VERBOSE_MAKEFILE on)
```
tomekzaw, line 2:
```suggestion
set(CMAKE_VERBOSE_MAKEFILE ON)
```

android/src/main/jni/CMakeLists.txt
kkafar: I guess I could do it, but it comes down to simply using find & replace option in code editor. I will, however refine variable names and do some cleanup.

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.h
kkafar: See: [2cfe611](https://github.com/software-mansion/react-native-screens/pull/1585/commits/2cfe611a7fc27467286f2733ea1f3a87d1958dcf)

I went with `JSI_EXPORT` macro as it handles not only `GNU` compiler attributes, but also  `MSVC`'s ones.

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
kkafar: [8ce7490](https://github.com/software-mansion/react-native-screens/pull/1585/commits/8ce7490631c62a489487f48e72824dbe8ff83622)

android/src/main/jni/rnscreens.h
kkafar: I went with usage of `JSI_EXPORT` because it handles broader set of compilers (not only GNU). I could provide custom macro implementation, but:

1. JSI's one is ready to be used
2. If we wanted to achieve similar functionality to `JSI_EXPORT` it would require to detect what compiler is being used - I guess it's good idea to use already estabilished solutions.

android/src/main/jni/CMakeLists.txt
kkafar: [213e592](https://github.com/software-mansion/react-native-screens/pull/1585/commits/213e59226c5d93ea2f1de078f3836072e3ea6ae1)

android/src/main/jni/CMakeLists.txt
kkafar: So I went with your suggestion. All it takes now is to change `LIB_LITERAL` constant. However some further customisation may be required as you noticed in RNSVG (adding custom dependencies etc.)

android/src/main/jni/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
+cmake_minimum_required(VERSION 3.13)
+set(CMAKE_VERBOSE_MAKEFILE ON)
+
+set(LIB_LITERAL rnscreens)
```
tomekzaw, line 4:
```suggestion
set(LIB_NAME rnscreens)
```

tomekzaw (approved):
🚀

WoLewicki (approved):
I added some comments. Sorry for late review 😅

android/src/main/jni/rnscreens.cpp
```diff
@@ -0,0 +1,12 @@
+#include "rnscreens.h"
```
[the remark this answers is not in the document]
kkafar: https://github.com/software-mansion/react-native-screens/pull/1595

common/cpp/react/renderer/components/rnscreens/RNSScreenState.h
kkafar: Note that this explicit GNU compiler attribute was changed in [2cfe611](https://github.com/software-mansion/react-native-screens/pull/1585/commits/2cfe611a7fc27467286f2733ea1f3a87d1958dcf) to usage os `JSI_EXPORT` macro.
Also see this comment: https://github.com/software-mansion/react-native-screens/pull/1585#discussion_r972940173

android/src/main/jni/rnscreens.h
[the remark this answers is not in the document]
kkafar: See https://github.com/software-mansion/react-native-screens/pull/1585#discussion_r972235122

PR #1596, opened 2022-09-19
fix: make enabling device orientation notifications internal

kacperkapusciak:
## Description

UIKit documentation states that in order to successfully use orientation in iOS apps the orientation has to be enabled via [the `beginGeneratingDeviceOrientationNotifications ` method](https://developer.apple.com/documentation/uikit/uidevice/1620041-begingeneratingdeviceorientation?language=objc). This was addressed in https://github.com/software-mansion/react-native-screens/pull/1543.

This PR attempts to make this change internal and also removing an additional `react-native-screens` installation step from README.

PR made based on code from `expo-screen-orientation` package [from this line](https://github.com/expo/expo/blob/519ac3d86864da9dd920a9ff8d0c880f6ec81a58/ios/versioned/sdk46/EXScreenOrientation/EXScreenOrientation/ABI46_0_0EXScreenOrientationRegistry.m#L38). 

Fixes https://github.com/react-navigation/react-navigation.github.io/issues/1177 

Note by @kkafar:

For now I went with placing the code in ctor/dtor of `RNSScreen` view manager.
The view manager construction & lifecycle is managed by React Native's internals (and is executed during application startup). This behaviour is considered stable (changing this would be braking change for RN Paper architecture).

I considered also:

1. Creating a native module, so we can execute required code in its lifetime (ctor, dtor) methods -- but it seemed too much... Also we would have to load this module from JS (prevent lazy loading).
2. Using `dispatch_once` in `+ (voi)enforceDesiredDeviceOrientation` method - and calling only the `begin...` method (as there would be no place to call the responding `end...` method.
3. Calling `begin...`/`end...` methods every time `enforceDesiredDeviceOrientation` is being called.
4. Adding the code to ctor/dtor of `RNSScreenWindowTraits` and creating a static object. But I had objections. I feared  that compiler might optimise-out (remove) this static variable since it is unused. I've found some [mails](https://gcc.gnu.org/pipermail/gcc-help/2021-April/140125.html) form [gcc-help mailing list](https://gcc.gnu.org/mailman/listinfo/gcc-help) indicating that it might be the case. 

## Changes

- Moved `beginGeneratingDeviceOrientationNotifications` and `endGeneratingDeviceOrientationNotifications` methods from applications `AppDelegate`s to `RNSScreenManager` `init`/`dealloc` methods.

## Testing

I attached a breakpoint to see that added code ends actually being called. 

Orientation playground in `Example/` app.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -45,14 +45,6 @@
 #endif
 }
 
-#if !TARGET_OS_TV
-// See:
-// 1. https://github.com/software-mansion/react-native-screens/pull/1543
-// 2. https://github.com/software-mansion/react-native-screens/pull/1596
-// This line aims to ensure that device orientation notifications are enabled
-static RNSScreenWindowTraits *orientationNotificationLifecycle = [RNSScreenWindowTraits new];
-#endif
-
 #ifdef RN_FABRIC_ENABLED
 - (instancetype)initWithFrame:(CGRect)frame
 {
@@ -1205,6 +1197,29 @@ RCT_EXPORT_VIEW_PROPERTY(statusBarHidden, BOOL)
 RCT_EXPORT_VIEW_PROPERTY(statusBarStyle, RNSStatusBarStyle)
 RCT_EXPORT_VIEW_PROPERTY(homeIndicatorHidden, BOOL)
 #endif
+
+#if !TARGET_OS_TV
+// See:
+// 1. https://github.com/software-mansion/react-native-screens/pull/1543
+// 2. https://github.com/software-mansion/react-native-screens/pull/1596
+// This class is instatiated from React Native's internals during application startup
+- (instancetype)init
+{
+  if (self = [super init]) {
+    dispatch_async(dispatch_get_main_queue(), ^{
+      [[UIDevice currentDevice] beginGeneratingDeviceOrientationNotifications];
+    });
+  }
+  return self;
+}
+
+- (void)dealloc
+{
+  dispatch_sync(dispatch_get_main_queue(), ^{
+    [[UIDevice currentDevice] endGeneratingDeviceOrientationNotifications];
+  });
+}
+#endif // !TARGET_OS_TV
 
 - (UIView *)view
 {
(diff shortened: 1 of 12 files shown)


WoLewicki (approved):
LGTM! :tada:

ios/RNSScreenWindowTraits.mm
```diff
@@ -4,6 +4,25 @@
...
+}
+
+- (void)dealloc
+{
+#if !TARGET_OS_TV
+  dispatch_sync(dispatch_get_main_queue(), ^{
+    [[UIDevice currentDevice] endGeneratingDeviceOrientationNotifications];
+  });
+#endif
+}
```
kkafar, line 24: Haven't tested it yet, but as we talked - I believe we do not instantiate this class -> this code is dead. 

I'm leaving this note so we don't forget and merge it by accident.

kkafar (changes requested):
Great idea to remove this installation step. 

I'm leaving a note below - as we talked, so we don't bypass the objections and mistakenly merge it.

ios/RNSScreen.mm
```diff
@@ -1198,6 +1198,29 @@ @implementation RNSScreenManager
...
+}
+
+- (void)dealloc
+{
+  dispatch_sync(dispatch_get_main_queue(), ^{
+    [[UIDevice currentDevice] endGeneratingDeviceOrientationNotifications];
+  });
+}
+#endif // !TARGET_OS_TV
+
```
kkafar, line 1223: I decided to add this code to `RNSScreenManager`. 
I'm asking for review, to make sure that there aren't any quirks in view manager's lifecycle.

kkafar (approved):
I'm asking for review, to make sure that there aren't any quirks in view manager's lifecycle.

ios/RNSScreen.mm
WoLewicki: Tbh, I am not familiar with the lifecycle of view managers, did you test how does it work on Fabric?

kkafar: Yep, I added ad breakpoint at `init` method and it got called during initial app loading.

PR #1609, opened 2022-10-07
#1045 Fix keyboard navigation in headers

### LINKED ISSUE #1045: Android Header not targetable with physical keyboard
## Description

When using the header option for any portion of the header (left, center, right), the header component is not targetable by physical keyboards.  This is problematic for apps that need to meet accessibility standards.

My assumption is that the toolbar defined in the [ScreenStackHeaderConfig](https://github.com/software-mansion/react-native-screens/blob/master/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java#L73) is the culprit.  Since this is not extending from the react view, I think the accessibility attributes are being skipped over.

I naively tried setting `ViewCompat.setImportantForAccessibility(mToolbar, ViewCompat.IMPORTANT_FOR_ACCESSIBILITY_YES);` in the `ScreenStackHeaderConfig`, but this does not address the issue.

A11Y is critically important for our application, and frequently overlooked in the react native community.  Until this is addressed I need to remove the native stack navigator for the standard one provided via react navigation unfortunately.  I hope this can be looked into so we can use the excellent performance benefits of this library to its full potential.

## Steps To Reproduce

1.  Create screen with header component in stack
2. Connect Android phone to bluetooth keyboard, or through usb port
3.  Navigate around screen using keyboard arrows or tab / shift + tab
4.  Notice header does not target like all touchables do in screen

### Expected behavior

Touchable buttons in header are targetable

### Actual behavior

All keyboard navigation targets end below the header component.

## Package versions

- React: 17.0.1
- React Native: 0.64.1
- React Native Screens: 3.3.0

zarubond:
## Description
Fixes #1045 

Adds an option to reach header buttons with keyboard navigation

## Changes

Based on the documentation https://developer.android.com/about/versions/oreo/android-8.0#kbnc I have set the header AppBarLayout to be keyboard navigation cluster. This way the user can switch between content cluster and header cluster with meta+tab keyboard combination.

## Screenshots / GIFs

Tabbing through content buttons, switching to header cluster, tabbing through header buttons and then back.

out

1. Create an app with header button or just a back button in the header
2. Use tab key to go through components in the screen content cluster
3. Press keyboard combination meta+tab, to get to header cluster
4. Now you can use tab key to go through components in header cluster
5. Pressing meta+tab will get you back to screen content cluster

\* This has been tested on Xiaomi phone with Android 8.1 and chromebook with Android 11 runtime. On Samsung with Android 9 this keyboard meta+tab combination is reserved for app switching, probably bug on Samsung side.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
@@ -3,6 +3,7 @@ package com.swmansion.rnscreens
 import android.annotation.SuppressLint
 import android.content.Context
 import android.graphics.Color
+import android.os.Build
 import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.Menu
@@ -112,6 +113,10 @@ class ScreenStackFragment : ScreenFragment {
             layoutParams = AppBarLayout.LayoutParams(
                 AppBarLayout.LayoutParams.MATCH_PARENT, AppBarLayout.LayoutParams.WRAP_CONTENT
             )
+
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                setKeyboardNavigationCluster(true)
+            }
         }
 
         view?.addView(mAppBarLayout)


zarubond:
Hi @kkafar , I am not sure should I talk to about feature additions. But would it be possible to review if this feature which fixes Samsung issue would be admissible to react-native-screens?

PR #1610, opened 2022-10-11
feat: add imperative API for search bar

kkafar:
## Description

This PR is a follow up to #1523 adding code for new React Native architecture.

Status:

* [x] iOS Paper
* [x] iOS Fabric
* [x] Android Paper
* [x] Android Fabric

TODO:

* [x] Check when native commands (`dispatchCommand`) mechanism was introduced and back to what RN version is it compatible with Paper (seems to be working properly at least back to rn 0.68, need to check earlier versions)
* [x] Implement Android
* [x] Refine type declarations in `types.tsx`

**Note**: while working on android implementation I've found few bugs

<details>
<summary>Scrollview does not respect header (Paper)</summary>

image

</details>

## Changes

Added following methods:

* `focus()`
* `blur()`
* `clearText()`
* `toggleCancelButton(show: boolean)`

For usage please see `Test1097` in `TestsExample`.

Added `Test1097` in `FabricTestExample` && `TestsExample`

## Checklist

- [x] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


ios/RNSSearchBar.mm
```diff
@@ -308,8 +327,13 @@ - (void)updateProps:(facebook::react::Props::Shared const &)props
...
 }
 
+- (void)handleCommand:(const NSString *)commandName args:(const NSArray *)args
+{
+  RCTRNSSearchBarHandleCommand(self, commandName, args);
+}
+
 #else
-#endif
+#endif // RN_FABRIC_ENABLED
```
WoLewicki, line 336:
```suggestion
#endif // RCT_NEW_ARCH_ENABLED
```

ios/RNSSearchBar.mm
```diff
@@ -348,4 +372,36 @@ - (UIView *)view
 RCT_EXPORT_VIEW_PROPERTY(onFocus, RCTBubblingEventBlock)
 RCT_EXPORT_VIEW_PROPERTY(onBlur, RCTBubblingEventBlock)
 
+RCT_EXPORT_METHOD(focus : (NSNumber *_Nonnull)reactTag)
```
WoLewicki, line 375: It seems risky to treat a native view as a module at the same time. In `react-native-svg`, we had such case and when migrating to new arch, we splitted it to a module and a view. Also, there are ways to not use the bridge in order to support the future bridgeless mode, and I think it would be nice if `react-native-screens` applied to it too (see e.g. https://github.com/software-mansion/react-native-svg/pull/1847/files#diff-92804e3153cd488997303254fe5c7e4ea48521712823779a79bca1edde23f899)

src/index.native.tsx
```diff
@@ -379,6 +408,64 @@ const ScreenStackHeaderBackButtonImage = (props: ImageProps): JSX.Element => (
...
+class SearchBar extends React.Component<SearchBarProps> {
+  nativeSearchBarRef: React.RefObject<SearchBarCommands>;
+
+  constructor(props: SearchBarProps) {
+    super(props);
+    this.nativeSearchBarRef = React.createRef();
+  }
+
+  blur() {
+    if (this.nativeSearchBarRef.current) {
```
WoLewicki, line 420: Maybe it would be good to warn the user if the ref is not yet set so it is known why it did not work?

ios/RNSSearchBar.mm
WoLewicki: On the other hand, I see you use commands, so why are those methods needed at all?

WoLewicki (commented):
I added some comments, please answer them. Other than that it looks promising! Good work!

ios/RNSSearchBar.mm
kkafar: [d68ed3f](https://github.com/software-mansion/react-native-screens/pull/1610/commits/d68ed3f3a1760179ff0db840888dab15f90769dc)

src/index.native.tsx
kkafar: [752244b](https://github.com/software-mansion/react-native-screens/pull/1610/commits/752244b85f28cfe79a8a4542a89aeb7d3be506dd)

PR #1646, opened 2022-11-22
fix: fix back button menu for headerBackTitleVisible prop

kacperkapusciak:
## Description

In the native-stack, there's a `headerBackTitleVisible` prop which allows hiding the back button title while keeping the default `<` arrow. It achieves it by passing [an empty string to the `backTitle` header prop](https://github.com/react-navigation/react-navigation/blob/%40react-navigation/native-stack%406.9.2/packages/native-stack/src/views/HeaderConfig.tsx#L171).  

Although it was a proper way of hiding the back button title prior to iOS 14 it has the negative effect of hiding all elements in the back button menu in >=iOS 14. 

This can be fixed by a relatively new `backButtonDisplayMode` API by setting its value to `minimal`.

Fixes react-navigation/react-navigation/issues/11015

### Before

https://user-images.githubusercontent.com/39658211/203362522-29afafe9-033e-4214-bfd1-9897b10941c1.mov

### After

https://user-images.githubusercontent.com/39658211/203359936-020bf4db-dede-4e9c-8c57-c042f1880711.mov

### No effect on iOS 13 (still uses an empty string as there's no back button menu available)

https://user-images.githubusercontent.com/39658211/203523164-6508ec6b-662b-45ce-9df0-46e81d4c900d.mov

TestsExample/Test1646.tsx

<details>
<summary>Code example</summary>
<br>

```tsx
import {NavigationContainer, useRoute} from '@react-navigation/native';
import {Button, View} from 'react-native';

  return (
    <NavigationContainer>
      <Stack.Navigator
        screenOptions={{
          headerBackTitleVisible: false,
        }}>
        <Stack.Screen
          name="Screen"
          component={Screen}
          options={({route}) => ({title: route.params?.title ?? 'Hello'})}
        />
      </Stack.Navigator>
    </NavigationContainer>
  );
}

function Screen({navigation}) {
  const route = useRoute();
  const count = route.params?.count ?? 0;
  return (
    <View>
      <Button
        onPress={() =>
          navigation.push(route.name, {
            title: `Hello ${count + 1}`,
            count: count + 1,
          })
        }
        title="Push a route"
      />
    </View>
  );
}
```

</details>

## Checklist

- [ ] Ensured that CI passes


kkafar (commented):
The code looks fine. I'll test it & ship it (if everything is ok)

kkafar (approved):
Looks good

kpoelhekke:
It seems like this change caused an issue in our codebase because it now always shows the `headerBackTitle` even when `headerBackTitleVisible: false`. This issue exists on both iOS 13 and iOS 14. After removing this code everything is working as before again.

kkafar:
@kpoelhekke, 
thanks for reporting. I'll look into it.

kpoelhekke:
Thanks @kkafar. FYI, these are the `screenOptions` that we pass:

```
{
        headerShown: true,
        headerBackTitleVisible: false,
        headerTintColor: theme.activeTab,
        headerBackImageSource: backIconPath,
        headerShadowVisible: false,
        orientation: "portrait_up",
        headerTitleAlign: "center",
      }
```

Gregoirevda:
This has been tested with screens all containing titles, but causes an issue when `headerBackTitleVisible: false`. Header back title can only be hidden by setting the previous screen title to `''`

PR #1665, opened 2022-12-13
Add "hidden" prop to `Screen` to support iOS Picture-in-Picture

jaredly:
## Description

Thanks for making this wonderful library! I've really enjoyed using it with react-navigation. I'm transitioning a large app from custom native navigation code (written before we adopted react-native) to react-navigation, and the only thing I'm having trouble with is iOS Picture-In-Picture.

If you're unfamiilar, Picture-in-Picture (PIP) refers to popping a video out into an "overlay" that persists as you navigate around the app, until you either close it or "restore" it, which results in the originating screen being pushed back onto the stack, and the video sliding back into its original location.

Here's a little video of that in action, for context:

https://user-images.githubusercontent.com/112170/207445170-f9fc77cd-cebe-44c9-b95e-d8338ad587c5.mp4

There are two things needed to support PIP well:
1) the ability to dismiss a screen without letting the resources be garbage collected. (when the originating ViewController is garbage collected, the PIP window automatically closes). Additionally, the screen must not be unmounted on the react side of things, so that events will still be handled properly (critically, the `onRestoreUserInterfaceForPictureInPictureStop` callback for react-native-video).
2) the ability to bring back the screen that the PIP overlay originated from, in response to the user tapping on the relevant icon in the overlay.

## Changes

The solution I've arrived at requires fairly minimal changes to react-native-screens, just adding a single `hidden` prop to the `Screen` component, which instructs the ScreenStack to ignore that screen for the purposes of constructing the arrays of ViewControllers. Additionally, when the hidden prop is modified (for the case of "bringing back" a dismissed screen), I clear the `_dismissed` flag.

The above screen recording demonstrates this change, in concert with the react-navigation changes.

lmk what the best strategy is here; do you want me to create a minimal example repo?

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
@@ -76,13 +76,6 @@ class ScreenViewManager : ViewGroupManager<Screen>(), RNSScreenManagerInterface<
         }
     }
 
-    /*
-    @ReactProp(name = "hidden", defaultBoolean = true)
-    override fun setGestureEnabled(view: Screen, hidden: Boolean) {
-        view.isHidden = hidden
-    }
-    */
-
     @ReactProp(name = "gestureEnabled", defaultBoolean = true)
     override fun setGestureEnabled(view: Screen, gestureEnabled: Boolean) {
         view.isGestureEnabled = gestureEnabled
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
@@ -76,6 +76,13 @@ class ScreenViewManager : ViewGroupManager<Screen>(), RNSScreenManagerInterface<
         }
     }
 
+    /*
+    @ReactProp(name = "hidden", defaultBoolean = true)
+    override fun setGestureEnabled(view: Screen, hidden: Boolean) {
+        view.isHidden = hidden
+    }
+    */
+
     @ReactProp(name = "gestureEnabled", defaultBoolean = true)
     override fun setGestureEnabled(view: Screen, gestureEnabled: Boolean) {
         view.isGestureEnabled = gestureEnabled
--- a/createNativeStackNavigator/README.md
+++ b/createNativeStackNavigator/README.md
@@ -189,6 +189,14 @@ A Boolean to that lets you opt out of insetting the header. You may want to * se
 
 Boolean indicating whether the navigation bar is translucent.
 
+#### `hidden` (iOS only)
+
+If a screen is marked `hidden`, the resources will be retained, but it will not be rendered
+in the stack. This is to support iOS picture-in-picture, where you have a video overlay originating
+on a screen, and even when popping the screen off the stack the resources need to be retained
+(not garbage collected), so that the video can continue playing. Additionally, this allows the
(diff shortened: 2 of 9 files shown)


kacperkapusciak (commented):
Hi @jaredly 👋 

I think thanks to this PR we've made a full circle in `react-native-screens` development 😅  

Screens where developed to fight a problem directly opposite to what you are facing - freeing resources of a screen that the user can't see. 

In Stack, Drawer and Bottom Tabs navigators there's a prop to disable this optimisation (disabling screens) - [`detachInactiveScreens`](https://reactnavigation.org/docs/stack-navigator/#detachinactivescreens). I think using a JS stack with `screens` disabled should work well for your use-case.

Fair, there's no prop like this in native-stack because it always felt like a step back but maybe, as you've presented a valid use-case, we should consider adding it

jaredly:
Interesting! I think `detatchInactiveScreens` is a slightly different mechanism, although I may be misunderstanding; with the current react-native-screens, picture in picture works if you navigate forward (adding screens to the stack), and only breaks when you navigate *back* (popping the originating screen off the stack, and causing it to get garbage collected).
At any rate, going with a js-only screens impl is a no-go for us, as we care a lot about native navigation animations & gestures.
Thanks for taking a look!

jaredly:
btw I put together a minimal react-native project illustrating the issue

jaredly:
@kacperkapusciak for what it's worth, switching to the js-implemented stack didn't help with the "originating view needs to stick around for PIP to keep going"

https://user-images.githubusercontent.com/112170/207755295-e8e39dd8-b871-4c11-9825-00dc7e321f8a.mp4

jaredly:
@kkafar sorry to bump, but this feature is blocking our adoption of react-native-screens & react-navigation. Anything I can do to make this easier to review?

kkafar:
Sorry for the delay! I'm recently a little bit buried in todo stuff. I'll try to take care of this PR soon.

jaredly:
bump? 🙏

jaredly:
@kkafar it would be really nice to get this merged in! Is there anything I can do on my end? I'd love to stop depending on a fork.

hirbod:
Very good feature, seems like so many valid and PRs are stale in RNS.

kkafar:
Hey, @hirbod, @maksg is currently assigned to implementing API for [`preload`](https://reactnavigation.org/docs/7.x/navigation-actions/#preload) functionality & it should cover this use case. 

As to the "stale PRs" I think you're absolutely right & I do remember about the PR of yours. Moreover we appreciate every PR. Currently in `react-native-screens`, however, Fabric related issues have the highest priority before 0.75.0 release of RN. Once the PR is open & waiting for my attention best thing to do is to tag & bump me from time to time. If you are unhappy with current pace of development & require prioritised attention it is best to contact us through https://swmansion.com/contact#contact-form.

tboba:
According to changes from [this](https://github.com/software-mansion/react-native-screens/pull/2389) PR, I believe these changes have been superseded by checking if activityState from screen is inactive? 
@jaredly Could you get the newest alpha of Screens 4.x and check if `preload` prop gives you the same behavior as requested here? You can check from `TestPreload.tsx` how you can dispatch preload action, using `navigation.dispatch` method.

kkafar:
Can be achieved with `activityState={0}`

PR #1668, opened 2022-12-19
Ensure full typings generation from index.native.tsx

alespergl:
## Description

The `index.native.d.ts` file that is generated during build doesn't contain all exported symbols because the source file doesn't use proper `export` statements, but instead assigns to the CommonJS `module.exports` variable.

This is a problem when consuming the package with tools that support suffix-based module resolution like the `moduleSuffixes` setting of `tsc`. The tool resolves the `index` to `index.native` and tries to use the corresponding declaration file, which is incomplete and the compilation fails.

Note: this change doesn't address the problem that assigning to `module.exports` is incompatible with ES6 modules and therefore the "module" build target is unusable in that context.

## Changes

- export symbols that are exportable as-is
- export declarations for symbols that are exported as getters on
  `module.exports` object
- instead of assigning to `module.exports`, use `Object.defineProperties` which
  doesn't conflict with the regular `export` statements

### DIFF
--- a/src/index.native.tsx
+++ b/src/index.native.tsx
@@ -440,7 +440,7 @@ class Screen extends React.Component<ScreenProps> {
   }
 }
 
-module.exports = {
+export {
   // these are classes so they are not evaluated until used
   // so no need to use getters for them
   Screen,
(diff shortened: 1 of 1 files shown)


alespergl:
@kkafar I created this PR a while back but is seems to have gone unnoticed. Do you know how to get it reviewed?
Thanks.

eurekaffeine:
My project is also written in `CommonJS`. Please support `CommonJS`! @kkafar

tboba:
Hey @alespergl, thanks for creating this issue!
Because of the PR #1982 that removed mixed CJS/ESM, I believe this PR is obsolete now. Hence, I'll close it, but if you have any questions or you're willing to contribute further, please open it! Thanks!

PR #1705, opened 2023-01-31
chore: change fabric flag

WoLewicki:
Changed `RN_FABRIC_ENABLED` to `RCT_NEW_ARCH_ENABLED` since it is the current standard for this flag.

### DIFF
--- a/RNScreens.podspec
+++ b/RNScreens.podspec
@@ -26,7 +26,7 @@ Pod::Spec.new do |s|
       "CLANG_CXX_LANGUAGE_STANDARD" => "c++17",
     }
     s.platforms       = { ios: '11.0', tvos: '11.0' }
-    s.compiler_flags  = folly_compiler_flags + ' ' + '-DRN_FABRIC_ENABLED'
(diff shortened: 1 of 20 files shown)


tomekzaw (approved):

kkafar (approved):
Just some **overanalysis** below :D 

I like the idea os using similar compile option name across whole SWM Stack, but maybe we should consider customising it a bit;

1. `RCT` prefix comes from `react-native` source code and represents that project. Different libraries should avoid using the same prefixes (it is rather a bad practice). 
2. I don't really like the idea of using the same name as in `react-native` repo, as it may lead to some confusion about where it is defined. On the opposite: using library's prefix makes it clear.
3. It is rather good practice in ObjC to prefix classes & symbols with your own "signature".

Thus my suggestion would be to use `RNS_NEW_ARCH_ENABLED` in case of this library and respectively `RNGH_NEW_ARCH_ENABLED`, `RNSVG_NEW_ARCH_ENABLED`, `REA_NEW_ARCH_ENABLED` in the rest of SWM stack. 

Merging it in current state would also be acceptable.

PR #1707, opened 2023-02-02
Add onMenuButton event to RNSScreenView

goranche:
Implement recogniser to trigger mentioned event.

## Description

Implements a recogniser on `RNSScreenView` for the tvOS remote that watches for the menu button being pressed.
When this happens, a JS event named `onMenuButton` is triggered.

## **NOTE**

This branch / PR is based on the 3.13.1 tag, since that was the version that was used in the example application we used for investigation and development.

Before merging (automatic merge should not be possible) the changes should be based (and modified to match) from the version that'll be used.

It seems Fabric support was added in later versions and some substantial changes were introduced. This should not affect the functionality, but a manual merge will be needed first.


goranche:
wrong repo, wanted to do the PR on the fork first, but github got in the way 🙈 
sorry about that

PR #1708, opened 2023-02-02
fix(iOS): status bar does not respect app theme

### LINKED ISSUE #1683: Header theme is set incorrectly
### Description

When setting a theme for the NavigationContainer other than the OS, setting the theme for the Stack navigator works incorrectly (searchBar, header card shadow style).

IMG_4058

### Steps to reproduce
see snack example

https://snack.expo.dev/@appomart/react-native-screens-theme-bug-report

### Screens version

~3.18.0

### React Native version

0.70.5

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

Expo managed workflow

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

kkafar:
## Description

Resolves #1683 

At the moment, when user set theme on `NavigationContainer` the status bar does not adapt. It keeps following OS theme. You have to set theme for status bar separately in screens options.
Everything works fine when the app theme is following the OS theme (vast majority of cases I guess):

```
const scheme = useColorScheme(); // From RN API
...
<NavigationContainer theme={scheme === 'dark' ? DarkTheme : DefaultTheme}>
...
```

The issue appears when for example OS theme is `dark`, but the app wants to force display in `light` -> then the status bar keeps respecting OS theme and rest of the application does not (except few other components such as search bar, but they will be handled in separate PR's).

Proposed changes fix the problem & modify the default behaviour in following ways:

* when following system theme everything works as it used to + there is a possibility to override status bar theme by applying a screen option directly on the `Screen` component,
* when forcing custom theme the status bar style will respect it unless it is overridden directly on the `Screen` component,
* when forcing custom theme you can still get the old behaviour (if for some reason you wish) by setting `statusBarStyle: 'auto'` directly on the screen / in `screenOptions`

**NOTE**: This changes apply only to native stack v5 (`react-native-screens`), not to `react-navigation`. The fix will have to be implemented separately there. 

**NOTE**: TBH I'm not really sure what is the desired look on Android, will eventually fix it in separate PR.

## Changes

* Enforced `light` or `dark` theme for status bar basing on theme from navigation context, when the user does not provide a value.

See `Test1683` in `FabricTestExample` && `TestsExamples`

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/FabricTestExample/App.js
+++ b/FabricTestExample/App.js
@@ -82,6 +82,7 @@ import Test1473 from './src/Test1473';
 import Test1476 from './src/Test1476';
 import Test1646 from './src/Test1646';
 import Test1649 from './src/Test1649';
+import Test1683 from './src/Test1683';
 
 enableFreeze(true);
 
--- a/FabricTestExample/src/Test1683.tsx
+++ b/FabricTestExample/src/Test1683.tsx
@@ -1,0 +1,74 @@
+import {View, StyleSheet, Button, useColorScheme} from 'react-native';
+import {
+  NavigationContainer,
+  DarkTheme,
+  DefaultTheme,
+  LightTheme,
+  Theme,
+} from '@react-navigation/native';
+// import {createNativeStackNavigator} from '@react-navigation/native-stack';
+import {createNativeStackNavigator} from 'react-native-screens/native-stack';
+import React, {createContext, useContext, useState} from 'react';
+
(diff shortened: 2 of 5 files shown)


kkafar:
When the CI passed I'll merge this PR to make my life a little bit easier when working on a fix for search bar. Any review is still welcome.

PR #1710, opened 2023-02-03
feat(iOS): add prop for setting search bar theme

kkafar:
## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


kkafar:
Likely won't proceed here. Will reopen if we decide to reuse it.

Issue #1650, opened 2022-12-01
presentation="modal" not working on android


Kotpes:
### Description

Having set `presentation="modal"` for a screen has no effect on that screen on Android (still opens full screen view)
```js
<Stack.Screen name="ComponentName" component={Component} options={() => ({
        presentation: 'modal'
})} />
```
Works as expected on iOS.

### Steps to reproduce

1. Add `presentation="modal"` to a screen `options`
2. See that it has no effect on android (still opens the full screen view) - see the snack attached 

### Snack or a link to a repository

https://snack.expo.dev/gg56CATOS

### Screens version

6.6.1

### React Native version

0.69.0

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Acknowledgements

Yes

lokeshdangi:
was facing the same issue. 
seems like native-stack has some issue with "modal" but 'containedTransparentModal'  was working.
"modal" works fine if we switch to a normal stack navigator

jaredly:
Yeah it looks like it's just not implemented? 🤔 weird that the documentation says 
> `modal`, `containedModal`, `fullScreenModal`, `formSheet` will use `Screen.StackPresentation.MODAL`.

when `MODAL` is essentially ignored in the android code.
```
react-native-screens on main ➜ rg MODAL android
android/src/main/java/com/swmansion/rnscreens/Screen.kt
240:        PUSH, MODAL, TRANSPARENT_MODAL

android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
58:                Screen.StackPresentation.MODAL
60:                Screen.StackPresentation.TRANSPARENT_MODAL

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
90:        var visibleBottom: ScreenStackFragment? = null // this is only set if newTop has TRANSPARENT_MODAL presentation mode
328:            fragment.screen.stackPresentation === Screen.StackPresentation.TRANSPARENT_MODAL
```

Given that `Screen`'s are backed by [androidx `Fragment`s](https://github.com/software-mansion/react-native-screens/blob/5682daffff5baaf943cec79ffe1913215653893b/android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt#L28), I expect the fix will be to have `modal` screens inherit from [DialogFragment](https://developer.android.com/reference/androidx/fragment/app/DialogFragment).

[...]

PR #1715, opened 2023-02-08
Fix documentation to indicate that "modal" presentation is not yet implemented on Android

jaredly:
unless I'm misreading the code, `StackPresentation` is essentially ignored in the android code. There is code to handle "rendering under" a transparentModal screen, but nothing to make a screen actually render as a dialog, e.g. by having a StackFragment inheriting from [DialogFragment](https://developer.android.com/reference/androidx/fragment/app/DialogFragment).


jaredly:
related to https://github.com/software-mansion/react-native-screens/issues/1650

kkafar:
Hi, thanks for PR! I won't merge it right now, as 3.20 was already released with this version of docs & there is chance that there will be something done for Android modal views before 3.21. I'll merge it in case the feature won't be ready for 3.21.

PR #1717, opened 2023-02-11
chore: set library namespace in build script

Sprimage:
## Description

Warning during Android build:

> package="com.swmansion.rnscreens" found in source AndroidManifest.xml: /home/dee/mobile/node_modules/react-native-screens/android/src/main/AndroidManifest.xml.
Setting the namespace via a source AndroidManifest.xml's package attribute is deprecated.
Instead set the namespace (or testNamespace) in the module's build.gradle file, as described here: https://developer.android.com/studio/build/configure-app-module#set-namespace

## Changes

- Updated `android/build.gradle` 
- ~Updated `android/src/main/AndroidManifest.xml`~ (@kkafar: please see [this commit description](https://github.com/software-mansion/react-native-screens/pull/1717/commits/e62dc0e3a6b56020dfbb8fc6a39beb2678a100f8)) and review notes.

Build test examples and notice that the warning no longer appears.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -1,5 +1,4 @@
 
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-          package="com.swmansion.rnscreens">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
 </manifest>
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -1,4 +1,5 @@
 
-<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+          package="com.swmansion.rnscreens">
 
 </manifest>
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -40,6 +40,8 @@ def reactNativeArchitectures() {
 android {
     compileSdkVersion safeExtGet('compileSdkVersion', 28)
 
+    namespace "com.swmansion.rnscreens"
+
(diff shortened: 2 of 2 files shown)


kkafar:
@Sprimage, thank you for noticing this & contribution. Changes look completely fine, but for some reason out Android detox (e2e tests) build fail. I'll investigate it and merge the PR if it is not related.

android/src/main/AndroidManifest.xml
```diff
@@ -1,5 +1,4 @@
 
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-          package="com.swmansion.rnscreens">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
```
kkafar, line 2: Due to some issues with parsing in older RN CLI versions we need to restore package declaration in the manifest file.

```suggestion
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
          package="com.swmansion.rnscreens">
```

kkafar (changes requested):
@Sprimage I tested this change and it seems that it breaks Android build on React Native 0.69 (and possible all older versions) due to some bug in RN CLI version that it ships with.

I did a bit playing around and it looks like absence of `package="com.swmansion.rnscreens"` in `AndroidManifest.xml` triggers the parsing issue (package definition is expected there).

Thankfully it seems that we can [leave package definition in manifest file](https://issuetracker.google.com/issues/200682321?pli=1) as long as it exactly matches namespace defined in build script (at least in AGP 8.0).

kkafar:
I'll commit my suggestions to push this forward.

kkafar (approved):
:+1:

PR #1726, opened 2023-02-24
fix(iOS): back button not respecting style options

tyler-coleman:
## Description

@kkafar:
#1646 introduced a bug, where if iOS version >= 14 styles for back title would not get applied, due to [ill defined `if/else-if` logic](https://github.com/software-mansion/react-native-screens/blob/5e1f7ecdf30573f3e441e73c4f1f05c1f3ab9ccb/ios/RNSScreenStackHeaderConfig.mm#L477-L491).

## Changes

@kkafar:
Removed the fix introduced in #1646 and replaced it with setting `backBarButtonItem`'s title to `nil` & `backBarButtonTitle` to appropriate value when back button title is supposed to be invisible. This results in no title being displayed next to back button, but the correct value is present in back menu (if one is enabled).

To achieve the demanded effect I needed to distinguish between states: 
1. back title visible or not next to back button
2. back title is nil / empty / whitespace only

Currently when `headerBackTitleVisible: false` is set in JS, [whitespace only string is send to native side](https://github.com/software-mansion/react-native-screens/blob/5e1f7ecdf30573f3e441e73c4f1f05c1f3ab9ccb/src/native-stack/views/HeaderConfig.tsx#L105) (same thing [happens in `react-navigation`](https://github.com/react-navigation/react-navigation/blob/2ec98f3fae051c60e76aed3698d1bca6bbf730ef/packages/native-stack/src/views/HeaderConfig.tsx#L171)), but the same information can be passed to native side when user sets:

```js
headerBackTitleVisible: true,
headerBackTitle: ' '
```

so we can not really determine on native side whether the back button title should be visible or not (in case of empty string / nil passed from the user we default to title of previous screen). Therefore I decided to pass `headerBackTitleVisible` prop value directly to the native side via **new prop `backTitleVisible` on `ScreenStackHeaderConfig` component** and base new logic on its (props) value.

### Before

https://user-images.githubusercontent.com/32227697/221241945-58d3bb51-f25a-4d0f-b74f-a4e18792535b.mp4

### After

https://user-images.githubusercontent.com/32227697/221241962-b25108c9-beb5-4bbb-b712-3830dfc1c250.mp4

`Test1726` in `FabricTestExample` & `TestsExample`

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [x] Ensured that CI passes


yusufyildirim:
I've been debugging this issue and the problem seems to be introduced in this PR: https://github.com/software-mansion/react-native-screens/commit/e871092d3ddf0b34ccc89efa988fe78fcbb79a1a

JacobJaffe:
Thank you! This fixes things perfectly for me after applying this as a patch.

kkafar:
Hi, I've tested the PR and it seems that it fixes this particular issue, but it brings back the one solved in #1646. 
We need to find better solution. Will be back with more detail.

tyler-coleman:
@kkafar That's unfortunate 😔 Sorry for not catching the regression before opening the PR. Would you like me to close this for now?

kkafar:
@tyler-coleman I've just found out that I can push to branch from your fork. So unless you are planning to restrict me from pushing or deleting the fork I think we can continue development here :) I have a grasp of whats going on & will work on fix.

tyler-coleman:
@kkafar Great! Go ahead, push 😊 I'll leave the fork up as long as it takes

kkafar:
@tyler-coleman, would you mind checking out whether it works as expected now?

kkafar:
cc @kacperkapusciak, we will need the new `backTitleVisible` prop to `ScreenStackHeaderConfig` in `react-navigation`.

itajenglish:
@tyler-coleman Thanks for the fix! However, I don't think it is working in nested navigators.

https://user-images.githubusercontent.com/889564/233818883-e539a2f2-781a-452d-8059-e3d68ef81c1d.mp4

PR #1728, opened 2023-03-06
feat: remove title with search bar

Gregoirevda:
## Description
When using a search bar in react navigation, there's currently no option to have a search bar without title.
Specifying a blank title leaves a blank area

Fixes https://github.com/react-navigation/react-navigation/issues/11257

## Changes
Added `removeTitle` property, which will place the search bar in the title

### Before
Screenshot 2023-03-06 at 10 23 32

### After
Screenshot 2023-03-06 at 10 23 52

```jsx
  <Stack.Navigator>
     <Stack.Screen name="SearchScreen" component={SearchView} 
        options={{
          title: 'Search',
          headerSearchBarOptions: {
            placeholder: 'Search',
            removeTitle: true,
          },
        }}
      />
  </Stack.Navigator>
```

### DIFF
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -597,13 +597,13 @@ namespace rct = facebook::react;
 #if !TARGET_OS_TV
           if (@available(iOS 11.0, *)) {
             RNSSearchBar *searchBar = subview.subviews[0];
-            if(searchBar.removeTitle) {
+            navitem.hidesSearchBarWhenScrolling = searchBar.hideWhenScrolling;
+
+            if(searchBar.hideTitle) {
               [navitem.searchController.searchBar sizeToFit];
               navitem.titleView = searchBar.controller.searchBar;
-              searchBar.controller.hidesNavigationBarDuringPresentation = false;
             } else {
               navitem.searchController = searchBar.controller;
-              navitem.hidesSearchBarWhenScrolling = searchBar.hideWhenScrolling;
             }
           }
 #endif
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -597,8 +597,14 @@ namespace rct = facebook::react;
 #if !TARGET_OS_TV
           if (@available(iOS 11.0, *)) {
             RNSSearchBar *searchBar = subview.subviews[0];
-            navitem.searchController = searchBar.controller;
-            navitem.hidesSearchBarWhenScrolling = searchBar.hideWhenScrolling;
+            if(searchBar.removeTitle) {
+              [navitem.searchController.searchBar sizeToFit];
+              navitem.titleView = searchBar.controller.searchBar;
+              searchBar.controller.hidesNavigationBarDuringPresentation = false;
+            } else {
+              navitem.searchController = searchBar.controller;
+              navitem.hidesSearchBarWhenScrolling = searchBar.hideWhenScrolling;
+            }
           }
 #endif
         }
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -602,6 +602,7 @@ namespace rct = facebook::react;
             if (searchBar.hideTitle) {
               [navitem.searchController.searchBar sizeToFit];
               navitem.titleView = searchBar.controller.searchBar;
+              searchBar.controller.hidesNavigationBarDuringPresentation = false;
             } else {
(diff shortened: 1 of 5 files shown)


native-stack/README.md
```diff
@@ -516,6 +516,12 @@ Boolean indicating whether to hide the search bar when scrolling.
 
 Defaults to `true`.
 
+#### `removeTitle` (iOS only)
+
+Boolean indicating whether to hide the title.
+
+Defaults to `false`.
```
kkafar, line 523: I would go with `hideTitle` prop name, as we with most props use `hideXXX` / `showXXX` convention.

ios/RNSSearchBar.mm
```diff
@@ -143,6 +144,11 @@ - (void)setHideWhenScrolling:(BOOL)hideWhenScrolling
   _hideWhenScrolling = hideWhenScrolling;
 }
 
+- (void)setRemoveTitle:(BOOL)removeTitle
+{
+  _removeTitle = removeTitle;
+}
```
kkafar, line 150: As long as there is no further logic this setter function is redundant, as it will be auto-created by compiler. I believe we can safely delete it.

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -597,8 +597,14 @@ + (void)updateViewController:(UIViewController *)vc
...
             RNSSearchBar *searchBar = subview.subviews[0];
-            navitem.searchController = searchBar.controller;
-            navitem.hidesSearchBarWhenScrolling = searchBar.hideWhenScrolling;
+            if(searchBar.removeTitle) {
+              [navitem.searchController.searchBar sizeToFit];
+              navitem.titleView = searchBar.controller.searchBar;
+              searchBar.controller.hidesNavigationBarDuringPresentation = false;
+            } else {
+              navitem.searchController = searchBar.controller;
+              navitem.hidesSearchBarWhenScrolling = searchBar.hideWhenScrolling;
```
kkafar, line 606: Shouldn't these options be set regardless of `searchBar.removeTitle` value? Even if the title is hidden, we want the search bar to be properly configured (fully functional).

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -597,8 +597,14 @@ + (void)updateViewController:(UIViewController *)vc
 #if !TARGET_OS_TV
           if (@available(iOS 11.0, *)) {
             RNSSearchBar *searchBar = subview.subviews[0];
-            navitem.searchController = searchBar.controller;
-            navitem.hidesSearchBarWhenScrolling = searchBar.hideWhenScrolling;
+            if(searchBar.removeTitle) {
```
kkafar, line 600: Just a missing whitespace
```suggestion
            if (searchBar.removeTitle) {
```

kkafar (changes requested):
Hi, great job so far, thank you!

I support the idea but beside comments I left below I have few more thoughts:

1. We still need to handle these props on Android side (noops)
2. There is still Fabric support to be add
3. I'm not convinced whether placing search bar as a title view is the way to go here - I'm afraid this **might** result in some bugs, as we often rely (in library code) on where the particular views are placed in the view hierarchy. I'll do some research & let you know if there is better solution. 

I think I'll handle points 1 & 2 once we have polished paper implementation.

In meantime please address my review comments.

ios/RNSScreenStackHeaderConfig.mm
Gregoirevda: I've added hidWhenScrolling back , but searchController can't be assigned

Gregoirevda:
3. I've tried many alternatives, the only one who was getting close was
`[navitem setPreferredSearchBarPlacement:UINavigationItemSearchBarPlacementInline];` (iOS 16+ and next to title)

Gregoirevda:
I've re-applied `hidesNavigationBarDuringPresentation = false` back as clicking on the searchBar would then hide the searchBar, cannot see a use case for that :D

Gregoirevda:
@kkafar animations are working properly both on iPhone and iPad (mainly cancel button appearing)

hamdij0maa:
any idea on when this change will be released please @kkafar @Gregoirevda ?

hamdij0maa:
> @hamdij0maa, I've not fount time to review it yet, but I'll try in incoming days & I did not plan releasing 3.21.0 earlier than few weeks from now. However we'll see whether we can release it sooner.

thank you <3

Gregoirevda:
@kkafar Could we get this merged?

kkafar:
I won't proceed here in any reasonable timeline. Will reopen once I change my mind.

PR #1729, opened 2023-03-06
WIP chore: getDefaultHeaderHeight native impl

kkafar:
## Description

Add native implementation of `getDefaultHeaderHeight` function. Right now this is done, because `react-native-reanimated` desires such native API, however in future this idea can be pushed towards removing dependency on `react-native-safe-area-context`.

## Changes

Just straight up implemented previous logic from JS impl on the native side. 
I decided to implement this as static method of new `RNSUtils` class / object on both platforms, however it is up to change. 

**NOTE** JS implementation used React Natives `Platform`'s API. It is accurate to the running platform, but there are cases where it fails to resolve device type correctly (phone / pad). It uses `[UIDevice currentDevice] userInterfaceIdiom]` [underneath](https://github.com/facebook/react-native/blob/419025df226dfad6a2be57c8d5515f103b96917b/React/CoreModules/RCTPlatform.mm#L22-L36) which happens [not to report actual device type but the build target](https://stackoverflow.com/a/13513542/8635522) (see discussion) - so to get actually correct result we should query for device model. Being aware of that I decided to keep JS logic and be consistent with results returned by `Platform` API.

-

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.kt
@@ -395,10 +395,4 @@ class ScreenStackHeaderConfig(context: Context) : ViewGroup(context) {
         }
         toolbar.clipChildren = false
     }
-
-    companion object {
-        @JvmStatic
-        fun getDefaultHeaderHeight(width: Int, height: Int, topInset: Int, stackPresentation: Screen.StackPresentation): Int =
-            56 + topInset
-    }
 }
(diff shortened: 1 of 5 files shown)


tboba:
Hi @kkafar, as there's ongoing PR with similar change (see #1802), is this Pull Request still needed?

kkafar:
No it is not. Thank you.

kkafar:
Superseded by #1802

PR #1749, opened 2023-03-30
fix: do not apply namespace if it is not available in agp

WoLewicki:
## Description

`namespace` method is available since AGP 7.x.


kkafar (approved):
:+1: Seems fine

PR #1750, opened 2023-03-31
make RNSFullWindowOverlay a modal for accessibility

aweary:
## Description

`RNSFullWindowOverlay` is meant to cover the entire window, but at the moment VoiceOver doesn't know this and you can still navigate other views underneath this one. 

## Changes

This sets the `accessibilityViewIsModal` to `YES`, which tells VoiceOver that it should ignore all sibling views (which in this case would be basically everything else since this is a subview of the `UIWindow`) and means VoiceOver can no longer escape this view and navigate hidden content.

It also adds `UIAccessibilityPostNotification(UIAccessibilityLayoutChangedNotification, _container)` to tell VoiceOver that it should move focus into this view when its added to the window.

* Use any existing example where `FullWindowOverlay` is covering content on iOS
* Enable VoiceOver and attempt to navigate the screen with up/down swipe gesture
* Note that without this change you can access content outside the `FullWindowOverlay`; with this change you cannot

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSFullWindowOverlay.mm
+++ b/ios/RNSFullWindowOverlay.mm
@@ -15,6 +15,15 @@
 
 @implementation RNSFullWindowOverlayContainer
 
+- (instancetype)initWithFrame:(CGRect)frame
+{
+  self = [super initWithFrame:frame];
+  if (self) {
+    self.accessibilityViewIsModal = YES;
+  }
+  return self;
+}
+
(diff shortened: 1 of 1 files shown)


kkafar:
Thank you and sorry for very late response. I'll cherry-pick your changes & create new PR with minor adjustments.

PR #1765, opened 2023-04-26
fix:Compatible with version 0.72

NiuGuohui:
## Description

The Geometry.h file is deprecated and will be removed in the next version of React Native.
`graphics/conversions.h` is deprecated and will be removed in the future.

## Changes

- Updated `RNScreenState.h`

### DIFF
--- a/common/cpp/react/renderer/components/rnscreens/RNSScreenState.h
+++ b/common/cpp/react/renderer/components/rnscreens/RNSScreenState.h
@@ -1,8 +1,7 @@
 #pragma once
 
 #include <react/renderer/graphics/Float.h>
-#include <react/renderer/graphics/Geometry.h>
-#include <react/renderer/graphics/conversions.h>
+#include <react/renderer/core/graphicsConversions.h>
 
 #ifdef ANDROID
 #include <folly/dynamic.h>


jankosecki:
From what I can see this is a breaking change - RN prior to 0.72 does not contain `react/renderer/core/graphicsConversions.h` file.
However it does currently break the build for Android since CMake uses `-Werror -W#warnings` flags. 
I believe RN uses C++17 so you could use `__has_include()` preprocessing to include the new file or old file:
```
#if __has_include (<react/renderer/core/graphicsConversions.h>)
  #include <react/renderer/core/graphicsConversions.h>
#else
  #include <react/renderer/graphics/Geometry.h>
  #include <react/renderer/graphics/conversions.h>
#endif
```


To anyone interested I have managed to go further by adding:
```
externalNativeBuild {
          cmake {
              cppFlags "-Wno-unused-parameter -Wno-#warnings"
          }
}
```

WoLewicki (approved):
LGTM! Thanks for your contribution :rocket:

Since we want to keep the support only for the latest version for Fabric, I'll merge it and release the new version of the library with support only for 0.72.

PR #1786, opened 2023-06-09
fix: search bar positions are incorrect in RTL

moaead:
## Description

- Fixes search bar close icon position in **android** (RTL)
- Fixes search bar close icon and search bar field position in **IOS** (RTL)

## Screenshots / GIFs

### Before

**IOS**

image

**Android**

image

### After

**IOS**

image

**Android**

image

Force RTL in android or IOS and enable header search options by `headerSearchBarOptions` field then you will see in both android and IOS it's incorrectly positioned.


0xRenshaw:
Any movement on this?

PR #1790, opened 2023-06-13
feat: remove v4 from repo

WoLewicki:
## Description

PR removing `createNativeStackNavigator.tsx` file from the repo along with all the code regarding it. It may be considered **BREAKING CHANGE** for people still using `react-navigation` v4 in their app and should not update the version of `react-native-screens` if they still want to use that deprecated version.

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [ ] Ensured that CI passes


package.json
```diff
@@ -14,8 +14,7 @@
     "lint-android": "./android/gradlew -p android spotlessCheck -q",
     "lint": "yarn lint-js && yarn lint-android",
     "release": "npm login && release-it",
-    "prepare": "bob build && husky install",
-    "postinstall": "patch-package"
```
tomekzaw, line 18: ❤️

tomekzaw (approved):

Issue #1671, opened 2022-12-21
Header height error on devices with dynamic island


outaTiME:
### Description

Hello there,
After upgrading to Expo 47 I started experiencing a visual issue on devices with dynamic island.

In short, if I run my app in the Expo Go on the simulator everything looks fine (as it did with SDK 46), but if I run it from the Expo Go on the device (or if I build a build with EAS) it looks different.

Here below, I leave some screenshots to better understand the incident:

Expo Go (simulator):

![Simulator Screen Shot - iPhone 14 Pro Max - 2022-12-16 at 01 14 20](https://user-images.githubusercontent.com/140472/208022286-f2ba198c-912b-400e-bf09-bafd5d15a9dc.jpeg)

Expo Go (device):

![IMG_F54D287D8EA3-1-new](https://user-images.githubusercontent.com/140472/208022297-516c969b-860d-4775-9941-4885ce7fb759.jpeg)

On the other hand, if I compare the versions between the clients, there is a slight difference in their number:

Expo Go version (simulator): 2.26.4.101

Expo Go version (device): 2.26.4.1016582

I don't quite understand what could be the cause of the problem, but this affects the calculation of the header for the native stack of the `react-navigation` (`useHeaderHeight`) and ends up being wrong.

I assume that the version of Expo Go running on the device (2.26.4.1016582) introduces some regression or dependency incompatibility, which is what causes that problem (and this also affects the EAS build).

Any help is appreciated 🙏

Additionally, I leave you the report of this same incident in another repositories, in which I have not received an answer yet:

https://github.com/expo/expo/issues/20498
https://github.com/react-navigation/react-navigation/issues/10989

### Steps to reproduce

1. Download the repository, install dependencies and run `yarn start`
2. Run the project in `Expo Go` on simulator and see that there is no gap between the header and the content
3. Run the project in `Expo Go` on device and observe the gap between the header and the content

### Snack or a link to a repository

https://github.com/outaTiME/dynamic-island-error

### Screens version

3.18.2

### React Native version

0.70.5

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Architecture

None

### Build type

None

### Device

iOS simulator

### Device model

iPhone 14 Pro / iPhone 14 Pro Max

### Acknowledgements

Yes

tboba:
Hi, @outaTiME! Thanks for your feedback 👍 Sorry for late response 😢 

[...]

PR #1802, opened 2023-06-26
fix: calculate values of useHeaderHeight natively

tboba:
## Description

Currently, `useHeaderHeight()` hook returns default values, hard-coded in JS code. It uses `react-native-safe-area-context` under the hood to calculate top inset of the status bar and returns a header height `56` (for android) or `64` (for iOS). 
This PR introduces new `useAnimatedHeaderHeight` hook that calculates header height on each layout change.

## Changes

- Added support for calculating header height dynamically on the old / new architecture for Android and iOS.
- Added event that is being executed on header height change.
- Added new `useAnimatedHeaderHeight` hook for getting header height that is being calculated dynamically.

### Before

https://github.com/software-mansion/react-native-screens/assets/23281839/1e129ce8-cc62-4333-a6d8-f7283f018441

### After

https://github.com/software-mansion/react-native-screens/assets/23281839/5c11f371-c3f7-4eda-9969-4eb02f9de26a

See `Test1802` in `TestsExample` or `FabricTestExample`.
There's also `Test1802b` included. It's a reproduction for an issue #1671 which shows that issue should be fixed after this change.

## Checklist

- [X] Updated TS types
- [X] Updated documentation:
  - [X] https://github.com/software-mansion/react-native-screens/blob/main/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
  - [X] https://github.com/software-mansion/react-native-screens/blob/main/native-stack/README.md
- [x] Ensured that CI passes


hirbod:
Will this also mitigate the issue with the jumping header on Android when using tabs?

tboba:
Hi @hirbod, I think it should, but as this PR is still in an early state I cannot guarantee that. If this problem is strictly related to the value that is represented by the `useHeaderHeight()`, I'll try to find the solution for this bug and test it in this PR additionally 🎉

outaTiME:
That’s such good news @tboba, this dynamic calculation of the header size works also correctly on iOS modals?

tboba:
@outaTiME I looked onto that and it looks like header height is not being calculated on full screen modal - I already know why! I'll make a patch tomorrow for that 👍

tboba:
Hi @hirbod! Unfortunately, this PR still doesn't resolve that problem 😕 However, I'll try to find the solution for it after merging that one 👍

ios/utils/RNSScreenModal.h
```diff
@@ -0,0 +1,11 @@
+NS_ASSUME_NONNULL_BEGIN
+
+@interface RNSScreenModal : NSObject
+
++ (BOOL)isFullscreenModal:(UIModalPresentationStyle)presentationStyle;
++ (BOOL)isGrabbableModal:(UIModalPresentationStyle)presentationStyle;
++ (BOOL)isTransparentModal:(UIModalPresentationStyle)presentationStyle;
```
kkafar, line 7: I would put these directly on `RNSScreenView` or maybe create a `ModalUtil` category or sth. No need to create separate interface in my opinion.

What do you think?

src/native-stack/views/NativeStackView.tsx
```diff
@@ -73,7 +78,7 @@ const MaybeNestedStack = ({
 
   const isHeaderInModal = isAndroid
     ? false
-    : stackPresentation !== 'push' && headerShown === true;
+    : stackPresentation !== 'push' && headerShown;
```
kkafar, line 81:
```suggestion
    : stackPresentation !== 'push' && headerShown === true;
```

It's JavaScript, let's be explicit.

kkafar (changes requested):
Great job 🥳 

I have not tested the usage of it yet. Just looked at the code for now.

General notes:

1. We should reconsider approach of setting the `dynamic` flag globally in screens, as I do not like having a possibly invalid value in `HeaderHeightContext` (do I understand it right? When `dynamic = false` the `.height` will have invalid value)
2. What do you think of `recalculate` -> `resolve` / `update` / just `calculate` (in all cases)?

I've also left some code notes below -- please answer them.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -236,6 +242,34 @@ class Screen constructor(context: ReactContext?) : FabricEnabledViewGroup(contex
...
         }
 
+    fun recalculateHeaderHeight() {
+        val summarizedHeight = when (stackPresentation) {
+            StackPresentation.TRANSPARENT_MODAL -> 0
+            else -> getSummarizedHeaderHeight()
+        }
+
+        UIManagerHelper.getEventDispatcherForReactTag(context as ReactContext, id)
+            ?.dispatchEvent(HeaderHeightChangeEvent(id, summarizedHeight))
```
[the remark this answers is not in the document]
tboba: Sounds good!

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -255,4 +289,11 @@ class Screen constructor(context: ReactContext?) : FabricEnabledViewGroup(contex
     enum class WindowTraits {
         ORIENTATION, COLOR, STYLE, TRANSLUCENT, HIDDEN, ANIMATED, NAVIGATION_BAR_COLOR, NAVIGATION_BAR_HIDDEN
     }
+
+    private fun Int.convertToDp(context: Context): Int {
+        val resources = context.resources
+        val metrics = resources.displayMetrics
+        val dp = this / (metrics.densityDpi / 160f)
```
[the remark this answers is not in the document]
tboba: In material design, one dp is equal to one pixel on a screen of density 160 dpi, so to calculate correct value I'm dividing density by 160 😅

ios/utils/RNSScreenModal.h
tboba: I wouldn't go with category - I've tried that once and it was a bad idea. If I choose to move to the category, I would have to move the whole implementation there to omit circular imports. There I could use methods like `viewDidLayoutSubviews` to use newly-implemented functions there, but I found a strange behavior of modals while doing it - when I was trying to add `viewDidDisappear` to category, after closing the modal it was reappearing again 🤯 after closing it twice, then it disappeared completely.

I'll move them to RNSScreenView 😄

android/src/main/java/com/swmansion/rnscreens/Screen.kt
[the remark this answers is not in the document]
kkafar: Let's put this information as a comment into the code

ios/RNSScreen.mm
```diff
@@ -964,6 +989,73 @@ - (void)viewDidLayoutSubviews
   }
 }
 
+- (void)dismissViewControllerAnimated:(BOOL)flag completion:(void (^)())completion
+{
+  [super dismissViewControllerAnimated:flag completion:completion];
+  BOOL isPresentedAsNativeModal = self.screenView.isPresentedAsNativeModal;
+
+  [self recalculateHeaderHeightIsModal:isPresentedAsNativeModal];
```
WoLewicki, line 997: I think we could move the `self.screenView.isPresentedAsNativeModal` here directly since it is quite descriptive by itself

ios/RNSScreen.mm
```diff
@@ -964,6 +989,73 @@ - (void)viewDidLayoutSubviews
...
+}
+
+- (CGFloat)getCalculatedHeaderHeight:(BOOL)isModal
+{
+  CGFloat navbarHeight = self.navigationController.navigationBar.frame.size.height;
+
+  // In case where screen is a modal, we want to calculate it's childViewController's height
+  if (isModal && self.childViewControllers.count > 0 && self.childViewControllers[0] != nil) {
```
WoLewicki, line 1005: I think something like this would be better since you cast it later.

```suggestion
  if (isModal && self.childViewControllers.count > 0 && [self.childViewControllers[0] isKindOfClass:UINavigationController.class]) {
```

WoLewicki: Also, what if we had double nested stack? Will it work correctly in such case?

src/native-stack/views/NativeStackView.tsx
```diff
@@ -304,6 +307,15 @@ const RouteView = ({
...
         });
       }}
+      onHeaderHeightChange={(e) => {
+        navigation.emit({
+          type: 'headerHeightChange',
+          data: { newHeight: e.nativeEvent.newHeight },
+          target: route.key,
+        });
+
+        setHeaderHeight(e.nativeEvent.newHeight);
```
WoLewicki, line 317: Won't it cause many many rerenders of the whole `Screen` each time the height changes and even if the user did not subscribe to those changes?

WoLewicki (commented):
I left some comments. Good job :tada:

android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
```diff
@@ -166,24 +167,16 @@ class ScreenViewManager : ViewGroupManager<Screen>(), RNSScreenManagerInterface<
...
-            MapBuilder.of("registrationName", "onTransitionProgress")
+    override fun getExportedCustomDirectEventTypeConstants(): MutableMap<String, Any> = mutableMapOf(
+            ScreenDismissedEvent.EVENT_NAME to MapBuilder.of("registrationName", "onDismissed"),
+            ScreenWillAppearEvent.EVENT_NAME to MapBuilder.of("registrationName", "onWillAppear"),
+            ScreenAppearEvent.EVENT_NAME to MapBuilder.of("registrationName", "onAppear"),
+            ScreenWillDisappearEvent.EVENT_NAME to MapBuilder.of("registrationName", "onWillDisappear"),
+            ScreenDisappearEvent.EVENT_NAME to MapBuilder.of("registrationName", "onDisappear"),
+            HeaderHeightChangeEvent.EVENT_NAME to MapBuilder.of("registrationName", "onHeaderHeightChange"),
+            HeaderBackButtonClickedEvent.EVENT_NAME to MapBuilder.of("registrationName", "onHeaderBackButtonClicked"),
+            ScreenTransitionProgressEvent.EVENT_NAME to MapBuilder.of("registrationName", "onTransitionProgress")
```
kacperkapusciak, line 178: We have a Android linter setup in RNScreens repo. You can use it by running `yarn format-android`. It ensures consistency in the codebase and shouldn't be altered by hand.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -236,6 +242,34 @@ class Screen constructor(context: ReactContext?) : FabricEnabledViewGroup(contex
             mNativeBackButtonDismissalEnabled = enableNativeBackButtonDismissal
         }
 
+    fun recalculateHeaderHeight() {
+        val summarizedHeight = when (stackPresentation) {
+            StackPresentation.TRANSPARENT_MODAL -> 0
```
kacperkapusciak, line 247: I believe transparent modals both on Android and iOS have headers and yielding `0` isn't correct

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -236,6 +242,34 @@ class Screen constructor(context: ReactContext?) : FabricEnabledViewGroup(contex
...
+            else -> getSummarizedHeaderHeight()
+        }
+
+        UIManagerHelper.getEventDispatcherForReactTag(context as ReactContext, id)
+            ?.dispatchEvent(HeaderHeightChangeEvent(id, summarizedHeight))
+    }
+
+    private fun getSummarizedHeaderHeight(): Int {
```
kacperkapusciak, line 255: 👍 to what's @kkafar saying about naming and let's also try to use similar naming on both platforms

ios/RNSScreen.mm
```diff
@@ -964,6 +989,73 @@ - (void)viewDidLayoutSubviews
...
+#else
+  // On TVOS, status bar doesn't exist
+  return CGSizeMake(0, 0);
+#endif // !TARGET_OS_TV
+}
+
+- (void)recalculateHeaderHeightIsModal:(BOOL)isModal
+{
+  if ([RNSScreenModal isTransparentModal:self.modalPresentationStyle]) {
+    [self.screenView notifyHeaderHeightChange:0];
```
kacperkapusciak, line 1044: I believe transparent modals both on Android and iOS have headers and yielding `0` isn't correct

kacperkapusciak (changes requested):

src/native-stack/views/NativeStackView.tsx
tboba: Hmm, that's a good catch 🤔 I'll reconsider it!

ios/RNSScreen.mm
WoLewicki: Modals on `iOS` do not have headers, but we discussed it with @tboba already. On `Android`, modal is exactly the same fragment as the normal screen, but with the transparent background iirc. Or am I wrong @kkafar ?

android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
tboba: Sure, thanks for the tip!

ios/RNSScreen.mm
tboba: @WoLewicki I checked that and it seems to work already!

ios/RNSScreen.mm
```diff
@@ -964,6 +1005,83 @@ - (void)viewDidLayoutSubviews
...
+
+  CGSize fallbackStatusBarSize = [[UIApplication sharedApplication] statusBarFrame].size;
+
+  if (@available(iOS 13.0, *)) {
+    CGSize primaryStatusBarSize = self.view.window.windowScene.statusBarManager.statusBarFrame.size;
+    if (primaryStatusBarSize.height == 0 || primaryStatusBarSize.width == 0)
+      return fallbackStatusBarSize;
+
+    return primaryStatusBarSize;
+  } else {
```
kkafar, line 1062: This code should be put inside ifdefs, as for someone on really old xcode it would fail to build. See [here](https://github.com/software-mansion/react-native-screens/blob/d40dc028a6e69a9aefbe56f6098fa9f517fc8190/ios/RNSScreen.mm#L120) for example.

ios/RNSScreenStack.mm
```diff
@@ -278,6 +310,11 @@ - (void)maybeAddToParentAndUpdateContainer
   }
 }
 
+- (void)addSubview:(UIView *)view
+{
+  [super addSubview:view];
+}
```
kkafar, line 316: I believe it is redundant

```suggestion
```

ios/RNSScreen.h
```diff
@@ -114,6 +118,9 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)notifyTransitionProgress:(double)progress closing:(BOOL)closing goingForward:(BOOL)goingForward;
 - (void)notifyDismissCancelledWithDismissCount:(int)dismissCount;
 - (BOOL)isModal;
+- (BOOL)isPresentedAsNativeModal;
+- (BOOL)isFullscreenModal;
+- (BOOL)isTransparentModal;
```
kkafar, line 123: If these methods are used in `RNSScreen.mm` only and I believe they are -- we do not need to declare them in `@interface` -- these can be declared & defined entirely in `RNSScreen.mm`.

kkafar (changes requested):
Few more remarks, please answer them. Great job so far 🥳 
Also I have few points on whether we need to calculate header height on view controller dismissal (talking about iOS here)?

ios/RNSScreen.h
tboba: Good point! I'll remove `isFullscreenModal` and `isTransparentModal`, as `isPresentedAsNativeModal` is being used in RNSScreenStack 😁

ios/RNSScreen.mm
tboba: Sure! I was thinking why we're putting those ifdefs 🤔

kkafar: FYI: The ```if (@available(iOS ... ))``` check is evaluated at runtime. When someone uses older xcode version without support for sdk >= iOS 13 the code inside the block could fail to build because of missing symbols at compile time.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -255,4 +281,12 @@ class Screen constructor(context: ReactContext?) : FabricEnabledViewGroup(contex
     enum class WindowTraits {
         ORIENTATION, COLOR, STYLE, TRANSLUCENT, HIDDEN, ANIMATED, NAVIGATION_BAR_COLOR, NAVIGATION_BAR_HIDDEN
     }
+
+    // One dp is equal to one pixel on a screen of density 160 dpi.
+    private fun Int.convertToDp(context: Context): Int {
+        val resources = context.resources
+        val metrics = resources.displayMetrics
+        val dp = this / (metrics.densityDpi / 160f)
```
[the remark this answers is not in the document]
tboba: @kkafar Looks like it is! I've tested the `toDIPFromPixel` method from it and I'm getting the same values. Thanks for pointing this out.

TestsExample/src/Test1671.tsx
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
```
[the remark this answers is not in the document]
tboba: Introduced in `d1175c4`.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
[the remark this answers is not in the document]
tboba: Introduced in https://github.com/software-mansion/react-native-screens/pull/1802/commits/78a81522b4970c9110ebfeb914cdbcf4b8b23a17.

ios/RNSScreen.mm
```diff
@@ -964,6 +1005,83 @@ - (void)viewDidLayoutSubviews
...
+
+// Checks whether this screen has any child view controllers of type RNScreensNavigationController.
+// Useful for checking if this screen has nested stack or is displayed at the top.
+- (BOOL)hasAnyChildNavigators
+{
+  for (int i = 0; i < [self.childViewControllers count]; i++) {
+    if ([self.childViewControllers[i] isKindOfClass:[RNScreensNavigationController class]]) {
+      return true;
+    }
+  }
```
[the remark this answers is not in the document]
tboba: Introduced in https://github.com/software-mansion/react-native-screens/pull/1802/commits/1f5d586aff8373e79fa535af38756bd4c6da8e68

ios/RNSScreen.mm
```diff
@@ -964,6 +1005,83 @@ - (void)viewDidLayoutSubviews
...
+
+  return false;
```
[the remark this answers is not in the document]
tboba: Introduced in https://github.com/software-mansion/react-native-screens/pull/1802/commits/1f5d586aff8373e79fa535af38756bd4c6da8e68.

ios/RNSScreen.mm
tboba: Introduced in https://github.com/software-mansion/react-native-screens/pull/1802/commits/1f5d586aff8373e79fa535af38756bd4c6da8e68.

ios/RNSScreenStack.mm
tboba: Introduced in https://github.com/software-mansion/react-native-screens/pull/1802/commits/9b2e310797d78f7c83296052878f00b276a4dfcd.

ios/RNSScreen.h
tboba: Introduced in https://github.com/software-mansion/react-native-screens/pull/1802/commits/ae03a777dff550b45fe0a31055e154fef7ea49ca.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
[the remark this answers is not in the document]
tboba: @kkafar We've talked already about leaving it as it is for now, but as I was talking with Wojciech we made a decision to try to catch this value natively. RNSScreen should be sufficient for catching this value for now. I'll also try to reduce calls for recalculating header height value, as NativeStackView should already store the information about the value in previous screen. I'll try to investigate that 👍

kkafar (commented):
We comin' home

TVOSExample/ios/TVOSExample.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>IDEDidComputeMac32BitWarning</key>
+	<true/>
+</dict>
+</plist>
```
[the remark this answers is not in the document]
tboba: It's surely unwanted 🤔

native-stack/README.md
```diff
@@ -750,10 +750,18 @@ navigation.popToTop();
...
+- `useAnimatedHeaderHeight` calculates the header's height dynamically. The value provided by this hook changes with every view layout (such as when dragging a large header). 
+Please beware of using this hook in heavy components, as it may result in performance issues.
 
 ```tsx
+// for using useHeaderHeight
 import {useHeaderHeight} from 'react-native-screens/native-stack';
+
+// for using useAnimatedHeaderHeight
+import {useAnimatedHeaderHeight} from 'react-native-screens/native-stack';
 ```
```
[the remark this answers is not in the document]
tboba: I think it should! I've forgot to add a mention about it :eyes:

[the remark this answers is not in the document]
tboba: Resolved in https://github.com/software-mansion/react-native-screens/pull/1802/commits/a68658b97f33653c31d281f10fda70130173b1e5.

TVOSExample/ios/TVOSExample.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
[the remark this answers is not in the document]
tboba: Resolved in https://github.com/software-mansion/react-native-screens/pull/1802/commits/a68658b97f33653c31d281f10fda70130173b1e5.

.gitignore
```diff
@@ -21,6 +21,7 @@ DerivedData
 *.ipa
 *.xcuserstate
 project.xcworkspace
+*.plist
```
kkafar, line 24: Hey, this should not be in top level `.gitignore`. See [`Info.plist`](https://github.com/software-mansion/react-native-screens/blob/main/FabricTestExample/ios/FabricTestExample/Info.plist) file which contains application specific configuration we want to have checked into version control.

kkafar (changes requested):
Approving besides what's mentioned below

.gitignore
tboba: You're right, I can also see there's a couple of `IDEWorkspaceChecks.plist` folks already in the project. Removing this line 👍 https://github.com/software-mansion/react-native-screens/pull/1802/commits/6b23ba464e6cad2b4de6336fffc7f0eb35d6738b

kkafar (approved):
Great job :+1:

Green from me, but let's wait for others to go through the changes as we chatted today.

kacperkapusciak (approved):
Amazing work 🎉 

Please address the comments that I've left and also please check out whether this won't crash on a separate project without Reanimated installed.

Overall - good work! :shipit:

ios/event/RNSHeaderHeightChangeEvent.mm
```diff
@@ -0,0 +1,44 @@
+#import "RNSHeaderHeightChangeEvent.h"
```
WoLewicki, line 1: Why is the folder named `event` not `events` as on android?

WoLewicki (approved):
If it works correctly in all scenarios, I think we are good to merge it :tada:

src/types.tsx
```diff
@@ -72,6 +72,10 @@ export type HeaderSubviewTypes =
   | 'center'
   | 'searchBar';
 
+export type HeaderChangeEventType = {
+  newHeight: number;
```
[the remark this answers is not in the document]
tboba: Sure! Fixed in https://github.com/software-mansion/react-native-screens/pull/1802/commits/db7b50571e9c813e737b254294dd31459efb0e89.

native-stack/README.md
```diff
@@ -750,10 +750,25 @@ navigation.popToTop();
 
 ### Measuring header's height
 
-To measure header's height, you can use `useHeaderHeight` hook.
+To measure header's height, you can use the `useHeaderHeight`, `useAnimatedHeaderHeight` or `useReanimatedHeaderHeight` hook.
+- `useHeaderHeight` calculates the header's height statically. The value provided by this hook only changes when the view appears or when there is a view rotation.
```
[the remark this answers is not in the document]
tboba: Fixed in https://github.com/software-mansion/react-native-screens/pull/1802/commits/970dfcae90c7c8fdab9136a6e4f0bfc6aca7b738 🎉

native-stack/README.md
```diff
@@ -750,10 +750,25 @@ navigation.popToTop();
 
 ### Measuring header's height
 
-To measure header's height, you can use `useHeaderHeight` hook.
+To measure header's height, you can use the `useHeaderHeight`, `useAnimatedHeaderHeight` or `useReanimatedHeaderHeight` hook.
+- `useHeaderHeight` calculates the header's height statically. The value provided by this hook only changes when the view appears or when there is a view rotation. 
+Use this hook if you are certain that your header height will not change, or if your component is heavy, and you want to avoid recalculating the header height with every view layout.
```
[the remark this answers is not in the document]
tboba: Fixed in https://github.com/software-mansion/react-native-screens/pull/1802/commits/970dfcae90c7c8fdab9136a6e4f0bfc6aca7b738 🎉

ios/event/RNSHeaderHeightChangeEvent.mm
tboba: That's a good point! Changed in https://github.com/software-mansion/react-native-screens/pull/1802/commits/bb80aef75672c8f4aa9d13b956fe49e417be269d.

ios/RNSScreenStack.mm
```diff
@@ -57,6 +57,31 @@ - (UIViewController *)childViewControllerForStatusBarHidden
...
 }
 
+- (void)viewDidLayoutSubviews
+{
+  if ([self.topViewController isKindOfClass:[RNSScreen class]]) {
+    RNSScreen *screenController = (RNSScreen *)self.topViewController;
+
+    // If RNSScreen includes a navigation controller of type RNScreensNavigationController, it should not calculate
+    // header height, as it could have nested stack.
+    if (!(screenController.childViewControllers.count > 0 && screenController.hasAnyChildNavigators)) {
```
WoLewicki, line 67: Why do we need to check the `count` here? Shouldn't `hasAnyChildNavigators` just return `NO` then?

ios/RNSScreenStack.mm
```diff
@@ -57,6 +57,31 @@ - (UIViewController *)childViewControllerForStatusBarHidden
   return [self topViewController];
 }
 
+- (void)viewDidLayoutSubviews
+{
+  if ([self.topViewController isKindOfClass:[RNSScreen class]]) {
```
WoLewicki, line 62: Shouldn't we call `[super viewDidLayoutSubviews]` here or is it not necessary?

ios/RNSScreenStack.mm
```diff
@@ -57,6 +57,31 @@ - (UIViewController *)childViewControllerForStatusBarHidden
...
+    RNSScreen *screenController = (RNSScreen *)self.topViewController;
+
+    // If RNSScreen includes a navigation controller of type RNScreensNavigationController, it should not calculate
+    // header height, as it could have nested stack.
```
WoLewicki, line 66: What about both navigators having headers? I know it is a weird case but should we care about it?

ios/RNSScreenStack.mm
```diff
@@ -57,6 +57,31 @@ - (UIViewController *)childViewControllerForStatusBarHidden
...
+    }
+  }
+}
+
+- (void)dismissViewControllerAnimated:(BOOL)flag completion:(void (^)())completion
+{
+  [super dismissViewControllerAnimated:flag completion:completion];
+  BOOL wasModalPresented = ((RNSScreen *)self.presentedViewController).screenView.isModal;
+  // Check if presented modal was nested
+  BOOL isTopViewCtrNativeModal = ((RNSScreen *)self.topViewController).screenView.isPresentedAsNativeModal;
```
WoLewicki, line 78: Is it ever `false`? I thought `dismissViewControllerAnimated` is called for modals only.

WoLewicki (commented):
Some more comments to answer when you have time 😅

FabricTestExample/src/Test1802.tsx
```diff
@@ -0,0 +1,174 @@
...
+  TapGestureHandler,
+} from 'react-native-gesture-handler';
+import { ReanimatedScreenProvider } from 'react-native-screens/reanimated';
+import { FullWindowOverlay } from 'react-native-screens';
+
+const Stack = createNativeStackNavigator();
+
+function ExampleScreen() {
+  const headerHeight = useAnimatedHeaderHeight();
+  // const headerHeight = useReanimatedHeaderHeight();
```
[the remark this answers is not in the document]
tboba: I tested it already and it was working without `useAnimatedStyles` 🤔

ios/RNSScreen.mm
```diff
@@ -977,6 +1026,87 @@ - (void)viewDidLayoutSubviews
...
 
+- (void)dismissViewControllerAnimated:(BOOL)flag completion:(void (^)())completion
+{
+  [super dismissViewControllerAnimated:flag completion:completion];
+  [self calculateHeaderHeightIsModal:self.screenView.isPresentedAsNativeModal];
+}
+
+// Checks whether this screen has any child view controllers of type RNScreensNavigationController.
+// Useful for checking if this screen has nested stack or is displayed at the top.
+- (BOOL)hasAnyChildNavigators
```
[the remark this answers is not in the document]
tboba: I was also thinking about that name, but I'm wondering if there could be a situation where the screen would not be nested and it still could have RNSNavigationController 🤔

ios/RNSScreenStack.mm
tboba: If you mean that we could have a nested stack with more than one header (which is a typical situation for nested stacks) then only the last one should call `calculateHeaderHeightIsModal`, so I assume we're right now caring about it (If I understand you correctly 😄)

ios/RNSScreen.mm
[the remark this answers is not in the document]
WoLewicki: There is no option for `RNSNavigationController` since it is the concept from our library for the stack navigator. If there was a child `RNSNavigationController` and it would not be a nested stack then it seems like a bug.

ha3:
@tboba great work, thank you!

I just tested the new `useHeaderHeight` with a header with a search bar in iOS. Returning value doesn't include the height of the search bar. It would be great also to cover this case.

tboba:
Hi @ha3, thanks for letting us know! 🙏 

It looks like `use(Re)animatedHeaderHeight` calculates the header correctly (modifying the test a bit I can confirm this hook is returning correct value - transformY for Animated.Value puts FullWindowOverlay underneath the header).

image

However, as I tested `useHeaderHeight` it seems that search bar is indeed not taken into account with calculations. 

image

As we're now thinking about changing the implementation of `useHeaderHeight` completely, we will take care of it some day, but we won't fix this in this PR. Stay tuned 🙏

android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt
```diff
@@ -103,6 +103,13 @@ open class ScreenContainer<T : ScreenFragment>(context: Context?) : ViewGroup(co
     open fun removeScreenAt(index: Int) {
         mScreenFragments[index].screen.container = null
         mScreenFragments.removeAt(index)
+
+        // If screen at given index had another screen before in stack,
+        // calculate its header height.
```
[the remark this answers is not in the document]
tboba: We need this, because when there's a screen that is exiting to its parent it seems that it doesn't call `calculateHeaderHeight` (You can check it on video below, where I'm testing it in ScreensExample). However, it looks like this is not the case in TestsExample, so I assume this is necessary for the nested navigators, when we're traversing from child navigator to the parent one? 🤔 



Also, I think caring about triggering this method in other navigators is not necessary - shouldn't this method be called only for ScreenContainer that contains only screen with given index? Otherwise calls from lines 104-105 would throw an exception there because of mScreenFragments[index] 😄

ios/RNSScreenStack.mm
tboba: It's not necessary, everything works normally without the `super` (Xcode also doesn't complain about it) but I can add one 👍

ios/RNSScreenStack.mm
tboba: Seems like it should! I'll try tomorrow making an example with nested stack and check if `hasAnyChildNavigators` is sufficient there

ios/RNSScreenStack.mm
tboba: It is - that's why we're checking if topViewController is _presented as native modal_. If it is, we're getting its top view controller's child view controller and passing its height.

Also, I'm trying to recall why there was `wasModalPresented` introduced 🤔 It seems that it is indeed obsolete.

ios/RNSScreenStack.mm
tboba: Added one in https://github.com/software-mansion/react-native-screens/pull/1802/commits/9557ec64ce6703d03bf0566d0b96fa2741b6ea18

ios/RNSScreen.mm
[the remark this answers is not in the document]
tboba: Renamed in https://github.com/software-mansion/react-native-screens/pull/1802/commits/d2622c4f0b80ab3939c09a3c369c3403b8d36138.

ios/RNSScreenStack.mm
tboba: Removed in https://github.com/software-mansion/react-native-screens/pull/1802/commits/94e52902d49a36e1f22a3059196898efc2a714ac, but I will check that further.

android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt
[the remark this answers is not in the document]
WoLewicki: I am not sure why should we trigger this event when going back to a screen? Could you elaborate on this more? I thought that each screen has its own header height measurement since each of them can have a header. So when going back to a screen, why would its header change from the time we left it?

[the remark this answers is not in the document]
tboba: @WoLewicki I deeply tested that and it looks like currently NativeStackView loses the information about previous header height 😕 

https://github.com/software-mansion/react-native-screens/assets/23281839/40a388da-7cfa-475f-8ebf-d38506b10cb3

I assume that's because `Animated` does not hold the information between renders. I managed to solve that by wrapping `Animated` with useRef, but to be honest I'm not convinced if doing `React.useRef().current` and performing operations on it as usual is a good solution (and even practice?). What do you think? 🤔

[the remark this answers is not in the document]
tboba: As I considered it, I don't see any points on which circumstances it might not work (I don't see any pitfalls either), so I assume we might go with that 🤔

ios/RNSScreenStack.mm
tboba: Can confirm that this part of if statement is unnecessary, I've changed it in https://github.com/software-mansion/react-native-screens/pull/1802/commits/271a4374fa394bf13d7516b348325698d8e86145.

android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt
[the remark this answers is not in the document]
tboba: I've removed the android part of the code and changed the JS side here: https://github.com/software-mansion/react-native-screens/pull/1802/commits/2edab997027f41c73cc9422c085170bce290bcf2.

android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt
```diff
@@ -103,6 +103,7 @@ open class ScreenContainer<T : ScreenFragment>(context: Context?) : ViewGroup(co
     open fun removeScreenAt(index: Int) {
         mScreenFragments[index].screen.container = null
         mScreenFragments.removeAt(index)
+
```
WoLewicki, line 106: Looks like a leftover

ios/RNSScreen.mm
```diff
@@ -980,6 +1030,81 @@ - (void)viewDidLayoutSubviews
   }
 }
 
+// Checks whether this screen has any child view controllers of type RNScreensNavigationController.
```
WoLewicki, line 1033: Why the name here is `RNScreensNavigationController` and in code it is `RNSNavigationController`?

tboba: Looks like the refactor tool didn't changed that when I was refactoring this class 🤔 Thanks for pointing this out!

ios/RNSScreen.mm
```diff
@@ -980,6 +1030,81 @@ - (void)viewDidLayoutSubviews
...
+}
+
+- (CGSize)getCalculatedStatusBarHeightIsModal:(BOOL)isModal
+{
+#if !TARGET_OS_TV
+  BOOL isDraggableModal = isModal && ![self.screenView isFullscreenModal];
+  BOOL isDraggableModalWithChildViewCtr =
+      isDraggableModal && self.childViewControllers.count > 0 && self.childViewControllers[0] != nil;
+
+  // When modal is floating (we can grab its header), we don't want to calculate status bar in it.
```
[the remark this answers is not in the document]
tboba: It looks like they have `headerShown` option enabled by default, but you're right that we should not count status bar here, as they're like 'transparent' 👍

android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt
tboba: Resolved in https://github.com/software-mansion/react-native-screens/pull/1802/commits/246d3a555b18a3ab5416851ad86a92c1cd683deb.

ios/RNSScreen.mm
tboba: Resolved in https://github.com/software-mansion/react-native-screens/pull/1802/commits/fe776ca0bfc904f0376cc9958d7ef7661eb101cf.

ios/RNSScreenStack.mm
```diff
@@ -203,25 +218,25 @@ - (void)presentationControllerDidDismiss:(UIPresentationController *)presentatio
...
+  [self updateContainer];
+  if (self.onFinishTransitioning) {
+    // instead of directly triggering onFinishTransitioning this time we enqueue the event on the
+    // main queue. We do that because onDismiss event is also enqueued and we want for the transition
+    // finish event to arrive later than onDismiss (see RNSScreen#notifyDismiss)
+    dispatch_async(dispatch_get_main_queue(), ^{
+      [self emitOnFinishTransitioningEvent];
+    });
   }
+#endif
```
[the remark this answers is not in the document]
tboba: Resolved in https://github.com/software-mansion/react-native-screens/pull/1802/commits/fe776ca0bfc904f0376cc9958d7ef7661eb101cf.

ios/RNSScreen.mm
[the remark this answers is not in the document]
tboba: Resolved in https://github.com/software-mansion/react-native-screens/pull/1802/commits/a7d90a8da36c53fe6df20ecae7890a2598fb5363.

ios/RNSScreen.mm
```diff
@@ -980,6 +1031,81 @@ - (void)viewDidLayoutSubviews
...
+  CGSize statusBarSize = [self getCalculatedStatusBarHeightIsModal:isModal];
+
+  // Unfortunately, UIKit doesn't care about switching width and height options on screen rotation.
+  // We should check if user has rotated its screen, so we're choosing the minimum value between the
+  // width and height.
+  CGFloat statusBarHeight = MIN(statusBarSize.width, statusBarSize.height);
+  CGFloat totalHeight = navbarHeight + statusBarHeight;
+
+  [self.screenView notifyHeaderHeightChange:totalHeight];
+}
```
kkafar, line 1107: Do we want to dispatch event to JS on each call of this method? 
I would suggest splitting it into two separate methods:

1. actually resolving header height and returning it

```objective-c
- (CGFloat)calculateHeaderHeightIsModal:(BOOL)isModal
{ ... }
```
2. sending the event

Or in minimal version rename the method so that it is **clear** that it produces side-effects.

ios/RNSScreenStack.mm
```diff
@@ -61,6 +61,21 @@ - (UIViewController *)childViewControllerForStatusBarHidden
...
+- (void)viewDidLayoutSubviews
+{
+  [super viewDidLayoutSubviews];
+  if ([self.topViewController isKindOfClass:[RNSScreen class]]) {
+    RNSScreen *screenController = (RNSScreen *)self.topViewController;
+
+    // Calculate header height during simple transition from one screen to another.
+    // If RNSScreen includes a navigation controller of type RNSNavigationController, it should not calculate
+    // header height, as it could have nested stack.
+    if (!screenController.hasNestedStack) {
```
kkafar, line 73: `hasNestedStack` is a method, not property. Does this call actually work?

src/fabric/ScreenNativeComponent.ts
```diff
@@ -21,6 +21,10 @@ type TransitionProgressEvent = Readonly<{
   goingForward: Int32;
 }>;
 
+type HeaderChangeEvent = Readonly<{
```
kkafar, line 24: `HeaderHeightChange`?

src/types.tsx
```diff
@@ -72,6 +72,10 @@ export type HeaderSubviewTypes =
   | 'center'
   | 'searchBar';
 
+export type HeaderChangeEventType = {
```
kkafar, line 75: I would rename this type as well.

kkafar (commented):
Few remarks and questions from me.

ios/RNSScreen.mm
```diff
@@ -395,6 +396,29 @@ - (void)notifyDisappear
...
+    std::dynamic_pointer_cast<const facebook::react::RNSScreenEventEmitter>(_eventEmitter)
+        ->onHeaderHeightChange(
+            facebook::react::RNSScreenEventEmitter::OnHeaderHeightChange{.headerHeight = headerHeight});
+  }
+
+  RNSHeaderHeightChangeEvent *event =
+      [[RNSHeaderHeightChangeEvent alloc] initWithEventName:@"onHeaderHeightChange"
+                                                   reactTag:[NSNumber numberWithInt:self.tag]
+                                               headerHeight:headerHeight];
+  [[RCTBridge currentBridge].eventDispatcher sendEvent:event];
```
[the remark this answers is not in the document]
tboba: We've discussed that topic in the office 🤌

ios/RNSScreenStack.mm
tboba: As I investigated this method it looks like this method is being called right now (which is strange, indeed). It looks that even without this method it reports correct values, but this method is also reducing event calls to JS, so I would prefer to leave it here 🤞

src/fabric/ScreenNativeComponent.ts
tboba: Renamed in https://github.com/software-mansion/react-native-screens/pull/1802/commits/3c9c0fbe13cb07071e71254d249b2f90a964c3a0 👍

src/types.tsx
tboba: Renamed in https://github.com/software-mansion/react-native-screens/pull/1802/commits/3c9c0fbe13cb07071e71254d249b2f90a964c3a0 👍

ios/RNSScreen.mm
tboba: Changed in https://github.com/software-mansion/react-native-screens/pull/1802/commits/aabfc7edf48d5521de9d6e4b34a6fdceef8e0e85.

ios/RNSScreen.mm
```diff
@@ -544,6 +563,34 @@ - (RNSScreenStackHeaderConfig *_Nullable)findHeaderConfig
...
+
+- (BOOL)isFullscreenModal
+{
+  switch (self.controller.modalPresentationStyle) {
+    case UIModalPresentationFullScreen:
+    case UIModalPresentationCurrentContext:
+    case UIModalPresentationOverCurrentContext:
+      return true;
+    default:
+      return false;
```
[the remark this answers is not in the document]
tboba: Changed in https://github.com/software-mansion/react-native-screens/pull/1802/commits/aabfc7edf48d5521de9d6e4b34a6fdceef8e0e85.

kkafar (approved):
I'm leaving green for the code changes, however I haven't tested the final version yet. I'll do it soon.

kkafar (approved):
🟢

PR #1810, opened 2023-07-05
chore: add onGestureCancel event

piaskowyk:
## Description

This PR adds new iOS only event. Event is triggered after canceled swipe back (as on video)

https://github.com/software-mansion/react-native-screens/assets/36106620/e9f63c7c-e6b4-4444-895a-d6b41c725963


WoLewicki (approved):
LGTM 🚀

WoLewicki:
Could you change the name of the PR also? It is not only for internal usage at the end cc @piaskowyk 😅

PR #1825, opened 2023-07-11
fix(iOS): buggy search bar / large title behaviour on Fabric

kkafar:
## Description

`UIKit` requires `ScrollView` to be at index 0 in given view's subview array to enable it's interaction with navigation bar. On Fabric our `MaybeNestedStack` view got flattened, however it was not removed from hierarchy -- instead the view was attached as first child of the `RNSScreenView` disabling system interaction between navigation bar and a scroll view. 

## Changes

1. Added `collapsable={false}` to `MaybeNestedStack` view, preventing it from being flattened by React Native. 
2. Moved `HeaderConfig` component "down" -- so it is attached as second child of a `Screen`. It was necessary, because currently on Fabric we are adding `RNSScreenStackHeaderConfig` into subviews of the `Screen` and it also interrupted the interaction. 
3. Refactored code that relied on `RNSScreenStackHeaderConfig` being first child of a `Screen` (if present at all).

Initially I've played around with fixing (modyfing) the view hierarchy on the native side just by removing / adding some views when scroll view was being attached under the `RNSScreenView`, however such approach led to cascade of bugs (fixing first, caused another one and so on). We can assume that in general discrepancy between native & shadow tree leads to unpleasant bugs.

Possible alternative solution would be to not add `RNSScreenStackHeaderConfig` as a subview of `RNSScreenView` on Fabric. I did not research it, but it seems that it would require more changes. 

**Note**

We need to consult with `@react-navigation` whether the `@react-navigation/native-stack` has to be adjusted.

### Before

https://github.com/software-mansion/react-native-screens/assets/50801299/3b89dee8-a307-44ba-a9f4-02e4dfdd95ac

### After

https://github.com/software-mansion/react-native-screens/assets/50801299/205f24d4-590b-402e-8b0a-848fd4482d59

`Test1097` in `FabricTestExample` can be used to test these changes. Just use the scroll view and observe that the navigation bar gets stretched / dwindled while scrolling down / up respectively. 

## Checklist

- [x] Ensured that CI passes


ios/RNSScreen.h
```diff
@@ -109,12 +112,16 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)notifyDisappear;
 - (void)updateBounds;
 - (void)notifyDismissedWithCount:(int)dismissCount;
+- (instancetype)initWithFrame:(CGRect)frame;
```
tboba, line 115: Are we using this initializer somewhere?

src/native-stack/views/NativeStackView.tsx
```diff
@@ -100,6 +100,7 @@ const MaybeNestedStack = ({
       ]}
       // @ts-ignore Wrong props passed to View
       stackPresentation={stackPresentation}
+      collapsable={false}
```
tboba, line 103: Why we need to set collapsable to false here? 🤔

tboba (changes requested):
Great job with this one! I just left some comments to consider 🎉

ios/RNSScreen.h
```diff
@@ -109,12 +112,16 @@ NS_ASSUME_NONNULL_BEGIN
...
 #endif
 
 - (void)notifyTransitionProgress:(double)progress closing:(BOOL)closing goingForward:(BOOL)goingForward;
 - (void)notifyDismissCancelledWithDismissCount:(int)dismissCount;
 - (BOOL)isModal;
 
+/// Looks for header configuration in instance's `reactSubviews` and returns it. If not present returns `nil`.
```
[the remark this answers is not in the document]
kkafar: Three dashes are actually intentional as they activate doc-style comment. You can try it out

ios/RNSScreen.h
kkafar: No, but it should've been added there before as it is an initialiser and comes handy in various scenarios to be able to init the screen (especially when debugging). Also it is planned to separate `RNSScreenView` impl from `RNSScreen` & `RNSScreenViewManager` thus it will be necessary change. 

I could settle for introducing this change in separate PR, but when commiting I initially decided to leave this one-liner as no-harmer.

src/native-stack/views/NativeStackView.tsx
kkafar: Good idea

kkafar: https://github.com/software-mansion/react-native-screens/pull/1825/commits/037a34813660bfa4db7880dd66968cf350c74c3d

t1gu1:
On the web, HeaderConfig must be first child of a Screen.
Otherwise A11y on web is kinda broken.

PR #1835, opened 2023-07-18
fix(Android): shorten alpha animation to 83ms on default enter-out

kkafar:
## Description

The issue was reported [here](https://github.com/react-navigation/react-navigation/issues/11438) on `@react-navigation` repo.

*Note* Awaiting for response from people that reported the issue

## Changes

Adjusted alpha animation duration as suggested in one of the comments. 

Basing on people opinion that this improves the situation.

## Checklist

- [x] Ensured that CI passes


kkafar (commented):
Basing on not-so-vivid discussion and people feedback gathered in

* https://github.com/react-navigation/react-navigation/issues/11438

this change improves the animation look. 

Merging it as temporary measure.

PR #1843, opened 2023-07-20
feat(iOS): add support for search bar placement

kkafar:
## Description

Add support for [search bar placement](https://developer.apple.com/documentation/uikit/uinavigationitemsearchbarplacement?language=objc) prop on iOS.

## Changes

Added both Fabric & paper native impl and exposed the prop on JS side.

`Test1097` in `TestExample` and `FabricTestExample`. 

Play around with value of `placement` prop in search bar configuration.

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [x] Ensured that CI passes

### DIFF
--- a/TestsExample/src/Test1097.tsx
+++ b/TestsExample/src/Test1097.tsx
@@ -50,6 +50,7 @@ function First({ navigation }: NativeStackScreenProps<ParamListBase>) {
     hideNavigationBar: false,
     autoCapitalize: 'sentences',
     placeholder: 'Some text',
+    placement: 'stacked',
     onChangeText: (e: NativeSyntheticEvent<{ text: string }>) =>
       setSearch(e.nativeEvent.text),
     onCancelButtonPress: () => console.warn('Cancel button pressed'),
--- a/native-stack/README.md
+++ b/native-stack/README.md
@@ -561,7 +561,6 @@ React.useLayoutEffect(() => {
 
 A callback that gets called when search bar is closing
 
-
 #### `onFocus`
 
 A callback that gets called when search bar has received focus.
@@ -580,6 +579,18 @@ Text displayed when search field is empty.
 
 Defaults to an empty string.
 
+#### `placement` (iOS only)
+
+Position of the search bar
+   
+Supported values:
+ 
+* `automatic` - the search bar is placed according to current layout
+* `inline` - the search bar is placed on the trailing edge of navigation bar
+* `stacked` - the searcr bar is placed below the other content in navigation bar
+
+Defaults to `stacked`
+  
 #### `textColor`
 
 The search field text color.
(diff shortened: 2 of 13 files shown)


native-stack/README.md
```diff
@@ -580,6 +579,18 @@ Text displayed when search field is empty.
...
+
+Position of the search bar
+   
+Supported values:
+ 
+* `automatic` - the search bar is placed according to current layout
+* `inline` - the search bar is placed on the trailing edge of navigation bar
+* `stacked` - the searcr bar is placed below the other content in navigation bar
+
+Defaults to `stacked`
```
tboba, line 592:
```suggestion
* `stacked` - the search bar is placed below the other content in navigation bar

Defaults to `stacked`.
```

TestsExample/src/Test1097.tsx
```diff
@@ -50,6 +50,7 @@ function First({ navigation }: NativeStackScreenProps<ParamListBase>) {
     hideNavigationBar: false,
     autoCapitalize: 'sentences',
     placeholder: 'Some text',
+    placement: 'stacked',
```
tboba, line 53: I would not change this test, as it was related to the other topic (related to Windows support, I guess? See #1097).
Also, if 'stacked' value is default, do we require defining this property here?

src/types.tsx
```diff
@@ -633,6 +634,20 @@ export interface SearchBarProps {
...
+  /**
+   * Position of the search bar
+   *
+   * Supported values:
+   *
+   * * `automatic` - the search bar is placed according to current layout
+   * * `inline` - the search bar is placed on the trailing edge of navigation bar
+   * * `stacked` - the searcr bar is placed below the other content in navigation bar
+   *
+   * Defaults to `stacked`
```
tboba, line 646:
```suggestion
   * * `stacked` - the search bar is placed below the other content in navigation bar
   *
   * Defaults to `stacked`.
```

tboba (commented):
Great job so far with this 🎉 

I'm just adding small minor comments there 😄

tboba:
⚠️ Also, I've found there is a weird issue after I changed the `placement` to `inline` in Test1097 and I believe it shouldn't look like that 🤔 Should it?

image

TestsExample/src/Test1097.tsx
kkafar: `Test1097` become test case for all search bar related stuff a long time ago -- that's why there is so much scrollview / searchbar code. Also we should not rename it as many PRs refer to this test in their description.

> Also, if 'stacked' value is default, do we require defining this property here?

It's easier to see where the value of prop should be placed / changed when looking for usage example / testing.

native-stack/README.md
kkafar: https://github.com/software-mansion/react-native-screens/pull/1843/commits/f17c946fd161485f4178415c6c4544fdf918f748

kkafar:
@tboba TBH I've no idea how to trigger 'automatic' behaviour as the Apple developer docs do not provide us with any specifics. 

> Also, I've found there is a weird issue after I changed the placement to inline in TestsExample/Test1097 and I believe it shouldn't look like that 🤔 Should it?

This is the native behaviour.

PR #1844, opened 2023-07-21
fix: Calculate large headers in `useHeaderHeight` hook

tboba:
## Description

Currently, `useHeaderHeight()` hook returns wrong values for using large headers on iOS. This PR introduces a fix that adds a logic for calculating large headers.

This change is also a part of ongoing work on completely new `useAnimatedHeaderHeight` hook, introduced in #1802. 

## Changes

- Added an if statement in `useHeaderHeight` that sets correct value for `headerHeight`.

### Before

<details>
<summary>iPhone 14 Pro (iOS 16.4)</summary>
image
</details>

<details>
<summary>iPhone 8 (iOS 16.4)</summary>
image
</details>

### After

<details>
<summary>iPhone 14 Pro (iOS 16.4)</summary>
image
</details>

<details>
<summary>iPhone 8 (iOS 16.4)</summary>
image
</details>

You can test changes, included in PR by selecting `Test1844` from `TestsExample`/`FabricTestExample`.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/native-stack/utils/getDefaultHeaderHeight.tsx
+++ b/src/native-stack/utils/getDefaultHeaderHeight.tsx
@@ -29,7 +29,7 @@ export default function getDefaultHeaderHeight(
         headerHeight = 32;
       } else {
         if (isFormSheetModal) headerHeight = formSheetModalHeight;
-        headerHeight = isLargeHeader ? 96 : 44;
+        else headerHeight = isLargeHeader ? 96 : 44;
       }
     }
   }
--- a/src/native-stack/utils/getDefaultHeaderHeight.tsx
+++ b/src/native-stack/utils/getDefaultHeaderHeight.tsx
@@ -7,7 +7,8 @@ const formSheetModalHeight = 56;
 export default function getDefaultHeaderHeight(
   layout: Layout,
   statusBarHeight: number,
-  stackPresentation: StackPresentationTypes
+  stackPresentation: StackPresentationTypes,
+  isLargeHeader: boolean
 ): number {
   // default header heights
   let headerHeight = Platform.OS === 'android' ? 56 : 64;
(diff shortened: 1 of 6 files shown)


src/native-stack/utils/getDefaultHeaderHeight.tsx
```diff
@@ -7,27 +7,29 @@ const formSheetModalHeight = 56;
...
     if (Platform.isPad || Platform.isTV) {
-      headerHeight = isFromSheetModal ? formSheetModalHeight : 50;
+      headerHeight = isFormSheetModal ? formSheetModalHeight : 50;
     } else {
       if (isLandscape) {
         headerHeight = 32;
       } else {
-        headerHeight = isFromSheetModal ? formSheetModalHeight : 44;
+        if (isFormSheetModal) headerHeight = formSheetModalHeight;
+        headerHeight = isLargeHeader ? 96 : 44;
```
kkafar, line 32: The `if` clause has no effect here. Final value of `header` height depends solely on value of `isLargeHeader`. It it should be the case -- please consider removing the `if` clause or adjust the logic.

kkafar (commented):
Looks fine beside one thing I mention below. 🥳 Please answer my concenrs.

src/native-stack/utils/getDefaultHeaderHeight.tsx
tboba: You're right 😅 I forgot about the `else` keyword.

tboba: Fixed in https://github.com/software-mansion/react-native-screens/pull/1844/commits/ebcbc9574f7c57bf00919940542d48f5fb34869a.

kkafar (approved):
I would consider adding braces around if/else, other that that I believe we are good to go

PR #1848, opened 2023-07-28
fix(iOS): invalid initial orientation of screen underneath full screen modal

kkafar:
## Description

In case of fullScreenModal we want to ask for orientation specifically the screen underneath,
so that when the modal is dismissed the controller underneath is already in expected orientation, without the need to first wait for the dimissal to complete and only then to rotate.

This still is not perfect, as you see the visual glitches / jumping on videos below, however it improves the situation.

## Changes

Asked the screen that presents another screen in `fullScreenModal` stack presentation for interface orientation mask.

### Before

https://github.com/software-mansion/react-native-screens/assets/50801299/7f30ca37-9e67-4812-982e-b6144cce827e

### After

https://github.com/software-mansion/react-native-screens/assets/50801299/47f0f57b-46e6-4adc-9bd7-1f87613106d8

`Test1775` in TE

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStack.mm
+++ b/ios/RNSScreenStack.mm
@@ -893,7 +893,6 @@ namespace react = facebook::react;
 {
   [self emitOnFinishTransitioningEvent];
   [RNSScreenWindowTraits updateWindowTraits];
-  [self setNeedsLayout];
 }
 #endif
 


LucasHimelfarb:
this PR should be merged

felire (approved):

matthargett:
is this waiting on something in particular before it's merged and released?

Jpunt:
I'm also wondering: Is this going to be merged any time soon?

kkafar:
This PR is waiting for me having some free throughput and finializing it. As far as I remember the solution was not perfect and didn't solve the problem completely.
Have you tested this PR, does it work for you? 

(you can install `react-native-screens` directly from github by putting `"react-native-screens": "software-mansion/react-native-screens#@kkafar/another-rotation-issue"` in your package json)

Jpunt:
@kkafar the solution seems to be working great in our first testing! One thing we've spotted so far is that the home-indicator (for iPhones without home-button) transitions a bit wonky. 

@jim-lub can you remember anything else?

kkafar:
Hey, I'll look into it rn

kkafar:
Also: If the PR in its current shape works for you, you can install `react-native-screens` directly from the branch / commit hash, by putting following in your `package.json`:

`"react-native-screens": "software-mansion/react-native-screens#@kkafar/another-rotation-issue"` 

(writing from memory, lookout for typos)

vargajacint:
I tested it on a real device and found the exiting animation is wrong: It's going from left to right not top to bottom if you are in landscape mode.

kkafar:
>I tested it on a real device and found the exiting animation is wrong: It's going from left to right not top to bottom if you are in landscape mode.

Would you mind providing screen recordings visualising what do you mean?

vargajacint:
> > I tested it on a real device and found the exiting animation is wrong: It's going from left to right not top to bottom if you are in landscape mode.
> 
> Would you mind providing screen recordings visualising what do you mean?
> 
> As I'm testing on iPhone 12 mini right now & I do not see difference between simulator and real device.

As you can see, if you keep your phone in landscape mode the screen (navigation.goBack) is closing from left to right (aka top->bottom in portrait mode)

https://github.com/software-mansion/react-native-screens/assets/42407460/cdfb58ca-f758-4b51-b280-61296ac96b25

kkafar:
Yeah, for now it is expected (however I see why this might be a subject for change). IIRC the stack animation reflects the orientation of the screen underneath, so if it is in portrait mode, the dismissal will be top-down from the perspective of the screen underneath (looking like it's left to right from the perspective of screen in landscape mode)

aleqsio:
@Jpunt you can use this patch file to install the patch without needing to keep the branch updated. remember to rerun pod install :) 

https://gist.github.com/aleqsio/b1cb8bd24f34ed697e21b0f28a1efe68

you can use [patch-package](https://github.com/ds300/patch-package) for this.

kesha-antonov:
We need this PR since I'm trying to make same implementation as Yandex Music and can't without with PR

This is what I'm trying to achieve: to prevent prev screens to rotate while fullscreen player is opened

https://github.com/user-attachments/assets/e8d10630-0d86-4399-97da-0783c4b907b6

JasperH8g:
Just wanted to chip in again: The patch above still works on iOS, on 4.16.0. The closing animation goes to the bottom of the portrait orientation (where it transitions to), but that could be something unrelated. The most important thing to me is that the underlying screen remains in portrait while the landscape modal is covering it. 

Can this please be merged, so that we don't have to rely on a patch like this?

By the way: I'm experiencing very similar issues on Android. I'm not sure if this is possible at all (apart from RN or RNS) on Android, but if anyone could provide a similar patch I'd be very grateful 🙏

Issue #1722, opened 2023-02-22
Sheet / detent buggy jumping behavior on iOS


ferrannp:
### Description

Not sure if you seen my answer here https://github.com/software-mansion/react-native-screens/pull/1649#issuecomment-1404050925, but trying the new `formSheet` with detent has a buggy behavior. You can see the content from the bottom when you slide. I think this bug is in `modal` too but it is not so visible. See the video:

https://user-images.githubusercontent.com/774577/214713301-25cabef3-7818-4c6b-880d-cd4866e7a441.mov

----

By the way, are you planning to implement this also on Android? Would love testing it or even helping. I think a native sheet really makes a difference.

### Steps to reproduce

 In your `Example` app, go to `StackPresentation.tsx`, and change this example:

```js
    <Stack.Screen
      name="FormSheet"
      component={FormScreen}
      options={{ stackPresentation: 'formSheet', sheetAllowedDetents: 'all' }}
    />
```

Can be any combo of `sheetAllowedDetents`.

### Snack or a link to a repository

https://github.com/software-mansion/react-native-screens/blob/main/Example/src/screens/StackPresentation.tsx#L187

### Screens version

Latest master of your Example

### React Native version

Latest master of your Example

### Platforms

iOS

### JavaScript runtime

None

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Acknowledgements

Yes

kkafar:
Thank you for reporting, I'll try to handle it. 

I've spent some time on Android implementation, but it is not so straightforward with current `react-native-screens` architecture - the *very short* story: we wrap every view inside custom `Fragment` subclass & to display modal view (I'm talking material components here) the view needs to be presented inside special `ModalFragment` (or sth close) & we can't inherit from multiple classes. If Android only had used composition over inheritance there would be no issue here, but it is like it is :D However I'm not abandoning the idea, will surely give it a second try.

ferrannp:
Thanks for answering @kkafar ! Hopefully we can make it work in Android.

By the way, any ideas about the iOS bug reported in here?

kkafar:
I believe #2045 fixed this issue at a cost of impaired flexbox model support in sheet.

PR #1852, opened 2023-08-01
feat(iOS): support for custom detents with `formSheet` presentation

kkafar:
## Description

This PR adds support for custom detents with `formSheet` modal presentation on `iOS >= 16.0`.

This new feature can be used with `native-stack v5` (`react-native-screens/native-stack`) via `sheetAllowedDetents` prop by passing an array of values to it (see prop description in docs).

I plan on exposing it in v6 (`@react-navigation/native-stack`) in https://github.com/react-navigation/react-navigation/pull/11516 or in a follow up PR. 

**Note**: I'm looking for any suggestions on better prop naming :D 

See #1870 for view flickering issue description & solution

Resolves #1772

## Changes

* [x] Implemented native logic
* [x] Exposed the prop from native side
* [x] Exposed the prop and set default values in JS
* [x] Added Android stubs 
* [x] Updated docs

## Implementation notes

Please look below for my review.

## Known issues

View flickering on Paper and Fabric (the exact cause is not determined yet, but the code responsible for this is in `updateBounds` method which is called on every layout recalculation and it reports current view size to react (/ updates shadow node).

See: 

* https://github.com/software-mansion/react-native-screens/issues/1722

`Test1649` in both `TestsExample` and `FabricTestExample` apps.

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [x] Ensured that CI passes


ferrannp:
Do you want some help to test this @kkafar ? Also, if you make this native for Android... you will be a hero 😅.

kkafar:
Hey @ferrannp, 
thank you and yes -- I would really use a hand in testing this stuff, any feedback is welcome.

Currently I'm working on Fabric implementation which has some intricacies and it is not ready yet. 
Paper, however should work just fine (but I have not tested how this behaves on iPad yet).

Also I'm still thinking on prop naming and best way of exposing this functionality, any suggestions would also be welcome. At the moment I'm using dedicated prop temporarily named `sheetCustomDetents` which has effect only if:

1. `stackPresentation == 'formSheet'`
2. `sheetAllowedDetents == 'custom'`

But I also consider accepting an array with `sheetAllowedDetents` prop (so the declaration would be `sheetAllowedDetents: SheetDetentTypes | number[]` and dispatching this internally.

ferrannp:
I will test it then @kkafar 👍 . For me this one `sheetAllowedDetents: SheetDetentTypes | number[]` makes more sense 👍 . Is there a limit of amount of detents do you know? Cause if there is a limit maybe we can be more explicit in the type `number[]`.

kkafar:
I'm not aware of any limitation on number of detents. 
Only restriction I'm aware of right now is that the array has to be sorted in ascending order.

ios/RNSScreen.h
```diff
@@ -85,9 +85,11 @@ namespace react = facebook::react;
 // Props controlling UISheetPresentationController
 @property (nonatomic) RNSScreenDetentType sheetAllowedDetents;
 @property (nonatomic) RNSScreenDetentType sheetLargestUndimmedDetent;
+@property (nonatomic) NSNumber *sheetCustomLargestUndimmedDetent;
```
kkafar, line 88: `sheetUndimmedDetentIndex`?

ios/RNSScreen.h
```diff
@@ -85,9 +85,11 @@ namespace react = facebook::react;
...
 @property (nonatomic) BOOL sheetGrabberVisible;
 @property (nonatomic) CGFloat sheetCornerRadius;
 @property (nonatomic) BOOL sheetExpandsWhenScrolledToEdge;
+@property (nonatomic) NSArray<NSNumber *> *sheetCustomDetents;
```
kkafar, line 92: `sheetDetentValues`, `sheetMaxDetentValueFractions` or some variation of these?

ios/RNSScreen.mm
```diff
@@ -571,31 +612,33 @@ - (void)updatePresentationStyle
...
+  NSMutableArray<UISheetPresentationControllerDetent *> *customDetents =
+      [NSMutableArray arrayWithCapacity:fractions.count];
+  int detentIndex = 0;
+  for (NSNumber *frac in fractions) {
+    NSString *ident = [[NSNumber numberWithInt:detentIndex] stringValue];
+    [customDetents addObject:[UISheetPresentationControllerDetent
+                                 customDetentWithIdentifier:ident
+                                                   resolver:^CGFloat(
+                                                       id<UISheetPresentationControllerDetentResolutionContext> ctx) {
+                                                     return ctx.maximumDetentValue * frac.floatValue;
```
kkafar, line 634: I'm not sure yet whether I should use `ctx.maximumDetentValue` or try using `large` system detent height (by using `resolveValueInContext:`). Needs to be tested.

ios/RNSScreen.mm
```diff
@@ -571,31 +612,33 @@ - (void)updatePresentationStyle
...
+#if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && defined(__IPHONE_16_0) && \
+    __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_16_0
+- (NSArray<UISheetPresentationControllerDetent *> *)detentsFromMaxHeightFractions:(NSArray<NSNumber *> *)fractions
+    API_AVAILABLE(ios(16.0))
+{
+  NSMutableArray<UISheetPresentationControllerDetent *> *customDetents =
+      [NSMutableArray arrayWithCapacity:fractions.count];
+  int detentIndex = 0;
+  for (NSNumber *frac in fractions) {
+    NSString *ident = [[NSNumber numberWithInt:detentIndex] stringValue];
```
kkafar, line 629: System API is organised is such way, that if I want to refer to particular detent I need to do it by its identifier (unique string). Thus I'm using indexes here. I rejected doing it by detent value (height) because the value is context dependent.

Alternative would be to make user define detents as objects with `id` / `name` field, e.g.:

```js
[{
  id: "3/5",
  value: 0.6
}]
```

But I did not like this idea even more so. 


This is why `sheetCustomLargestUndimmedDetent` (the name must be changed) takes an index as a value.

ios/utils/NSArray+RNSUtil.h
```diff
@@ -0,0 +1,12 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+@interface NSArray (RNSUtil)
+
+- (BOOL)rns_isEmpty;
+- (BOOL)rns_isNotEmpty;
+
+@end
+
+NS_ASSUME_NONNULL_END
```
kkafar, line 12: We should have done it long time ago...

TODO: replace all conditions of shape `someArray.count > 0` with call to one of these methods.

I used naming convention presented in Objective-C [docs](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/CustomizingExistingClasses/CustomizingExistingClasses.html#//apple_ref/doc/uid/TP40011210-CH6-SW1) to avoid name clashing and potential undefined behaviour.

kkafar (commented):
General remarks:

1. Prop definitions differ for native stack v5 and native component `ScreenProps`. 

In native stack we have:

```js
sheetAllowedDetents: SheetDetentTypes | number[],
sheetLargestUndimmedDetent: SheetDetentTypes | number,
```

while native component goes with four distinct props:

```js
sheetAllowedDetents: SheetDetentTypes,
sheetLargestUndimmedDetent: SheetDetentTypes,
sheetCustomDetents: number[],
sheetCustomLargestUndimmedDetent: number,
```

It made native implementation much easier. Alternatively we could take an array of strings: 

```js
sheetAllowedDetents: (SheetDetentTypes | string)[],
sheetLargestUndimmedDetent: SheetDetentTypes | string,
```

where string is a percentage, e.g. `'50%'`, and parse it on native side? 

Looking for any suggestions

2. Either native detent levels ('medium', 'large', 'all') or custom ones can be used, but not both at the same time.

This is because the values I receive must be sorted & exact height of native detents is not known and it is context dependent, thus I reject mixing them.

ios/RNSScreen.mm
```diff
@@ -107,7 +110,7 @@ - (void)updateBounds
 #ifdef RCT_NEW_ARCH_ENABLED
   if (_state != nullptr) {
     auto newState = react::RNSScreenState{RCTSizeFromCGSize(self.bounds.size)};
-    _state->updateState(std::move(newState));
+    //    _state->updateState(std::move(newState));
```
kkafar, line 113: Dont bother, I'm still playing around with flickering on Fabric

ios/RNSScreen.mm
WoLewicki: What about if someone provided 2 same values?

ios/utils/NSArray+RNSUtil.h
WoLewicki: Why would we need this? count > 0 seems straightforward enough.

native-stack/README.md
```diff
@@ -267,13 +272,16 @@ Defaults to `false`.
...
+If this prop is set to:
+
+- `large` - the view underneath won't be dimmed at any detent level
+- `medium` - the view underneath will be dimmed only when detent level is `large`
+- `all` - the view underneath will be dimmed for any detent level
 
- - `large` - the view underneath won't be dimmed at any detent level
- - `medium` - the view underneath will be dimmed only when detent level is `large`
- - `all` - the view underneath will be dimmed for any detent level
+In case `sheetAllowedDetents` prop is set to an array of numbers A, this prop can be set to an index of array A,
```
WoLewicki, line 281:
```suggestion
In case `sheetAllowedDetents` prop is set to an array of numbers, this prop can be set to an index of that array,
```

WoLewicki (commented):
Just a quick review with some questions. Looks good overall 🎉

ios/RNSScreen.mm
kkafar: iOS will treat them as a single detent. The resolved value in such case is rather iOS implementation specific and not documented. It's better to treat such case as undefined behaviour. 

I see your point though.

I guess we will stick to using indices.

ios/utils/NSArray+RNSUtil.h
kkafar: Tbh I'm baffled that `Objective-C` does not have this like every other civilised language (I'm thinking of `Kotlin`, `Scala`, `Swift`, `Rust`, even `C++` & `Java`)

It reads better, expresses intention in more explicit manner, less error prone. Only con I see here is a little weird spelling due to the prefix. 

I mean I can see it not being in scope of this PR, so if you insist on not having it right now I'll revert. Let me know.

native-stack/README.md
```diff
@@ -232,11 +232,16 @@ Defaults to `pop`.
 Describes heights where a sheet can rest. 
 Works only when `stackPresentation` is set to `formSheet`.
 
-Available values:
+This prop can be set either to native value:
```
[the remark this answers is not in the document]
kkafar: Yep, I agree
https://github.com/software-mansion/react-native-screens/pull/1852/commits/81efd882904abcfb95e967d6da19c43d36c586d7

native-stack/README.md
kkafar: https://github.com/software-mansion/react-native-screens/pull/1852/commits/3ca5cdfd50fbfeb5c4c808ff769fdf170e7e9dba

src/native-stack/types.tsx
```diff
@@ -366,11 +372,14 @@ export type NativeStackNavigationOptions = {
    * - `medium` - the view underneath will be dimmed only when detent level is `large`
    * - `all` - the view underneath will be dimmed for any detent level
    *
+   * In case `sheetAllowedDetents` prop is set to an array of numbers A, this prop can be set to an index of array A,
```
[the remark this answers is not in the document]
kkafar: https://github.com/software-mansion/react-native-screens/pull/1852/commits/3ca5cdfd50fbfeb5c4c808ff769fdf170e7e9dba

ios/utils/NSArray+RNSUtil.h
tboba: I see the point of it, but I would choose better names for those methods - maybe something like `__isEmpty` if we mean to use it internally? `rns` doesn't look good here though.

tboba (changes requested):
Great job with this one 🥳 However, I've got some suggestions that should be detented (haha) 😂😂🤣

ios/utils/NSArray+RNSUtil.h
tboba: I see the point of it, but I would choose better names for those methods - maybe something like `__isEmpty` if we mean to use it internally? `rns` doesn't look good here though.

ios/RNSScreen.h
tboba: I would stay with `sheetCustomDetents`, as `sheetAllowedDetents` is already there and it would be confusing to use just`sheetDetentValues`. Also, `sheetMaxDetentValueFractions` seems a bit odd for me 🤔

android/src/paper/java/com/facebook/react/viewmanagers/RNSScreenManagerDelegate.java
```diff
@@ -29,6 +30,9 @@ public void setProperty(T view, String propName, @Nullable Object value) {
       case "sheetLargestUndimmedDetent":
         mViewManager.setSheetLargestUndimmedDetent(view, (String) value);
         break;
+      case "sheetCustomLargestUndimmedDetent":
+        mViewManager.setSheetCustomLargestUndimmedDetent(view, value == null ? -1 : ((Double) value).intValue());
```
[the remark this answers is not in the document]
kkafar: This code is generated by codegen. No point modifying it.

FabricTestExample/src/Test1649.tsx
```diff
@@ -1,23 +1,76 @@
...
+const cornerRadiusAtom = jotai.atom(-1);
+const expandsWhenScrolledToEdgeAtom = jotai.atom(false);
+
+const sheetOptionsAtom = jotai.atom(
+  get => ({
+    sheetAllowedDetents: get(allowedDetentsAtom),
+    sheetLargestUndimmedDetent: get(largestUndimmedDetentAtom),
+    sheetGrabberVisible: get(grabberVisibleAtom),
+    sheetCornerRadius: get(cornerRadiusAtom),
+    sheetExpandsWhenScrolledToEdge: get(expandsWhenScrolledToEdgeAtom),
```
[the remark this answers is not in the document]
kkafar: You need to restart the app, reloading the bundle leads to buggy behaviour. I've just tested it & it seems to work as it is described in source code:

> // If there is a larger detent to expand to than the selected detent, and a descendent scroll view is scrolled to top, this controls whether scrolling down will expand to a larger detent.
// Useful to set to NO for non-modal sheets, where scrolling in the sheet should not expand the sheet and obscure the content above.
// Default: YES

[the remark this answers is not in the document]
tboba: Tested it again and now it seems to work 👍

FabricTestExample/src/Test1649.tsx
```diff
@@ -1,23 +1,76 @@
...
+};
 
 const Stack = createNativeStackNavigator();
 
+/// Sheet options
+// const allowedDetentsAtom = jotai.atom<SheetDetent>('all');
+// const largestUndimmedDetentAtom = jotai.atom<SheetDetentTypes | number>('all');
+
+const allowedDetentsAtom = jotai.atom<SheetDetent>([
+  0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9,
```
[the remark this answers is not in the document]
tboba: Nevermind, it seems that after clicking `Change detent level` I wasn't able to change detent level to custom one 🤔

ferrannp:
I am testing this and TBH, being able to do things like:

```js
sheetCornerRadius: 20,
sheetAllowedDetents: [0.4, 0.9],
sheetLargestUndimmedDetent: 0,
```

is just amazing 🔥 .

However: 

https://github.com/software-mansion/react-native-screens/issues/1722 This bug is very bad. You do not notice it in a `modal` somehow but here is very noticeable see:

https://github.com/software-mansion/react-native-screens/assets/774577/71ea2f1c-387e-4635-a188-77defecb3203

There is an [opened PR](https://github.com/software-mansion/react-native-screens/pull/1767)... but IMO we need to fix this. It kinda breaks the whole experience.

kkafar:
I'm aware of this flickering (on Fabric it gets even worse).

#1767 can't be merged unfortunately, because while it helps with this particular problem, it brings back other already solved issues. 

I will see what can be done here.

FabricTestExample/src/Test1649.tsx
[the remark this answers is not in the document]
kkafar: I improved the example code a bit in https://github.com/software-mansion/react-native-screens/pull/1852/commits/2430d81e640dacc86b115feff4b51946d6196816

ios/RNSScreen.mm
```diff
@@ -550,19 +553,57 @@ - (RNSScreenStackHeaderConfig *_Nullable)findHeaderConfig
...
+
       sheet.prefersScrollingExpandsWhenScrolledToEdge = _sheetExpandsWhenScrolledToEdge;
       sheet.prefersGrabberVisible = _sheetGrabberVisible;
       sheet.preferredCornerRadius =
           _sheetCornerRadius < 0 ? UISheetPresentationControllerAutomaticDimension : _sheetCornerRadius;
 
-      if (_sheetLargestUndimmedDetent == RNSScreenDetentTypeMedium) {
+      int detentIndex = _sheetCustomLargestUndimmedDetent != nil ? _sheetCustomLargestUndimmedDetent.intValue : -1;
+      if (detentIndex != -1 && [_sheetCustomDetents rns_isNotEmpty]) {
+        if (detentIndex >= 0 && detentIndex < _sheetCustomDetents.count) {
```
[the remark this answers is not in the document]
kkafar: It is possible that we receive value `< 0` other than `-1` from user, in this case I've decided to put dimming view on all detent levels.

ios/utils/NSArray+RNSUtil.h
WoLewicki: Yeah, let's discuss if we need this later and then add it in another PR if needed.

jacobjonsson:
> I am testing this and TBH, being able to do things like:
> 
> ```js
> sheetCornerRadius: 20,
> sheetAllowedDetents: [0.4, 0.9],
> sheetLargestUndimmedDetent: 0,
> ```
> 
> is just amazing 🔥 .
> 
> However:
> 
> #1722 This bug is very bad. You do not notice it in a `modal` somehow but here is very noticeable see:
> 
>  jumping.mov 
> There is an [opened PR](https://github.com/software-mansion/react-native-screens/pull/1767)... but IMO we need to fix this. It kinda breaks the whole experience.

I've found setting the background color on the screen component exported from react-native-screens rather than the child view to fix the flickering.

ios/utils/NSArray+RNSUtil.h
kkafar: https://github.com/software-mansion/react-native-screens/pull/1852/commits/6e97b33e0814136c924727a8b99e37335dac3d69

ios/RNSScreen.mm
kkafar: https://github.com/software-mansion/react-native-screens/pull/1852/commits/d1fc618a8612b32dbd635bfb3d85d1dd4e93a0cb

ansh:
Will I be able to use a `FlatList` or `ScrollView` inside the modal with `formSheet` presentation?

ansh:
How can I help launch this? What's left?

kkafar:
Hey @ansh,

This PR is holded back by https://github.com/software-mansion/react-native-screens/pull/1974, as I want to have largest common API surface between platforms.

Also I'm still fighting with the "clipping" effect that was reported above. Haven't found a solution without tradeoffs yet (or maybe I just did, as I'm pushing right now).

kkafar:
Superseded by #2045

PR #1866, opened 2023-08-09
fix(iOS): wrong title in back button menu for screens w/ hidden header

### LINKED ISSUE #1864: iOS back button/menu title always is "back" when headerShown is false
### Description

If you have a stack of screens, the back button and the back button menu show the title of the previous screen, except when the header of the previous screen is not shown, in which case the shown title will revert to "back" (or localised equivalent).

For example:
```
<Stack.Navigator>
  <Stack.Screen name="screenA" component={ScreenA} options={{headerTitle: 'screen a title', headerShown: false}}/>
  <Stack.Screen name="screenB" component={ScreenB} options={{headerTitle: 'screen b title'}}/>
  <Stack.Screen name="screenC" component={ScreenC} options={{headerTitle: 'screen c title', headerShown: false}} />
  <Stack.Screen name="screenD" component={ScreenD} options={{headerTitle: 'screen d title'}} />
  <Stack.Screen name="screenE" component={ScreenE} options={{headerTitle: 'screen e title'}} />
</Stack.Navigator>
```
The names of `screenA` and `screenC` will be `back` where as the others will show `screen x title`. I expect this to be `screen x title` for all screens, regardless of whether or not the header has been shown.

image image

I have provided a snack. I am not quite familiar with Expo and Snacks and the versions differ a little from my app (for example I am using react native 0.71.7 and react-native-screens 3.24 locally, which are newer) but it has the same issue on the newer versions.

### Steps to reproduce

1. Create a navigator with a screen where the option `headerShown` is `false` and another screen to navigate to from there
2. Navigate to the second screen, note the back button/menu

https://snack.expo.dev/@hanneke/navigation-ios-back-name-not-properly-set-repro

### Screens version

3.20.0

### React Native version

0.71.3

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

Real device

### Device model

iPhone 11 16.5.1

### Acknowledgements

Yes

kkafar:
## Description

Ah, here we go again...

When header is hidden (`headerShown: false` in v6, `hidden: true` in v5) the method that updates header configuration returns early just after setting some layout related and LTR/RTL options, thus `title` property of current `UINavigationItem` is left unset leading to system default name `Back` being displayed in "back context menu". 

Fixes #1864

## Changes

When returning early (because header is hidden) we now set the `UINavigationItem` property to proper value. 

`Test1864` in `FabricTestExample` & `TestsExample`.

I've also tested it in context of

* https://github.com/software-mansion/react-native-screens/pull/1646

and mix of both. Seems to work fine.

## Checklist

- [x] Ensured that CI passes


FabricTestExample/src/Test1864.tsx
```diff
@@ -0,0 +1,76 @@
+import * as React from 'react';
+import { Button, View, Text } from 'react-native';
+import { NavigationContainer, ParamListBase } from '@react-navigation/native';
+import { createNativeStackNavigator } from '@react-navigation/native-stack';
+import { NativeStackNavigationProp } from '../../lib/typescript/native-stack';
```
tboba, line 5: I would change the import to `react-native-screens/native-stack`.

TestsExample/src/Test1864.tsx
```diff
@@ -0,0 +1,76 @@
+import * as React from 'react';
+import { Button, View, Text } from 'react-native';
+import { NavigationContainer, ParamListBase } from '@react-navigation/native';
+import { createNativeStackNavigator } from '@react-navigation/native-stack';
+import { NativeStackNavigationProp } from '../../lib/typescript/native-stack';
```
tboba, line 5: Same as above.

tboba (approved):
Looks great! I've got just one note about the import 🤠

FabricTestExample/src/Test1864.tsx
kkafar: Auto import 🤷🏻‍♂️ 

https://github.com/software-mansion/react-native-screens/pull/1866/commits/ab40baae5ec7d55fab398fc9acea2a42a32640c7

PR #1867, opened 2023-08-09
chore(Android): migrate from deprecated `RCTEventEmitter`

### LINKED ISSUE #1759: warning: 'RCTEventEmitter' is deprecated. Deprecated in Java import com.facebook.react.uimanager.events.RCTEventEmitter
### Description

 warning: 'RCTEventEmitter' is deprecated. Deprecated in Java
import com.facebook.react.uimanager.events.RCTEventEmitter
I added this code to improve the UI screen navigation
  if (Platform.OS === 'android') {
    if (UIManager.setLayoutAnimationEnabledExperimental) {
      UIManager.setLayoutAnimationEnabledExperimental(true);
    }
  }

### Steps to reproduce

1. I am currently using react-native 0.71, "react-native-screens": "^3.20.0",  "@react-navigation/native": "^6.1.5",
    "@react-navigation/native-stack": "^6.9.11",
 

https://github.com/Peak-Pay/peakpay-reactnative-app

### Screens version

3.20.0

### React Native version

0.71.2

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Release mode

### Device

Real device

### Device model

Redmi Note 11 pro

### Acknowledgements

Yes

kkafar:
## Description

Fixes #1759

`RCTEventEmitter` was deprecated over 3 years ago in [this commit](https://github.com/facebook/react-native/commit/2fbbdbb2ce897e8da3f471b08b93f167d566db1d)

I don't know when it will be finally deleted, but: 

1. There is no reason to wait for this moment with migration
2. According to the [commit message](https://github.com/facebook/react-native/commit/2fbbdbb2ce897e8da3f471b08b93f167d566db1d) there can be some minor performance improvements on Fabric
3. People complain on wall of warnings during build process

I inspected ReactNative's source code and it seems that parent class impl of `Event.dispatch` method we used up to this point calls `Event.getEventData` internally and if anything is returned it dispatches the event using appropriate event emitter ([see here](https://github.com/facebook/react-native/blob/e2ef6b98a061508bdd70a4c2747fe061c49b5c39/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/uimanager/events/Event.java#L154-L168)).

**Note**: This PR breaks support for versions of `react-native` lower than `0.65.x`.

## Changes

In all event classes:

1. Removed `Event.dispatch` override
2. Overrided `Event.getEventData`
3. Added `surfaceId` to event ctor

In all event construction places: 

1. Resolved appropriate (🤞🏻) `surfaceId`
2. Updated event construction

My empirical impression: it works as earlier.
My methodical attempt to test it: 

1. Add console logs to event handlers in `RouteView` in `src/native-stack/views/NativeStackView.tsx` use e.g. `Test1072` or different ones
5. Play around with text example
6. Notice that events are received in JS as expected. 

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt
@@ -296,9 +296,10 @@ open class ScreenFragment : Fragment {
             // we only send dismissed even when the screen has been removed from its container
             val screenContext = screen.context
             if (screenContext is ReactContext) {
+                val surfaceId = UIManagerHelper.getSurfaceId(screenContext)
                 UIManagerHelper
                     .getEventDispatcherForReactTag(screenContext, screen.id)
-                    ?.dispatchEvent(ScreenDismissedEvent(screen.id))
+                    ?.dispatchEvent(ScreenDismissedEvent(surfaceId, screen.id))
             }
         }
         mChildScreenContainers.clear()
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt
@@ -194,9 +194,9 @@ open class ScreenFragment : Fragment {
                 fragment.setLastEventDispatched(event)
                 val surfaceId = UIManagerHelper.getSurfaceId(it)
                 val lifecycleEvent: Event<*> = when (event) {
-                    ScreenLifecycleEvent.WillAppear -> ScreenWillAppearEvent(it.id)
+                    ScreenLifecycleEvent.WillAppear -> ScreenWillAppearEvent(surfaceId, it.id)
                     ScreenLifecycleEvent.Appear -> ScreenAppearEvent(surfaceId, it.id)
-                    ScreenLifecycleEvent.WillDisappear -> ScreenWillDisappearEvent(it.id)
+                    ScreenLifecycleEvent.WillDisappear -> ScreenWillDisappearEvent(surfaceId, it.id)
                     ScreenLifecycleEvent.Disappear -> ScreenDisappearEvent(surfaceId, it.id)
                 }
                 val screenContext = screen.context as ReactContext
@@ -239,6 +239,7 @@ open class ScreenFragment : Fragment {
                     .getEventDispatcherForReactTag(screenContext, screen.id)
                     ?.dispatchEvent(
                         ScreenTransitionProgressEvent(
+                            UIManagerHelper.getSurfaceId(screenContext),
                             screen.id, mProgress, closing, goingForward, coalescingKey
                         )
                     )
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt
@@ -192,9 +192,10 @@ open class ScreenFragment : Fragment {
         if (fragment is ScreenStackFragment && fragment.canDispatchEvent(event)) {
             fragment.screen.let {
                 fragment.setLastEventDispatched(event)
+                val surfaceId = UIManagerHelper.getSurfaceId(it)
                 val lifecycleEvent: Event<*> = when (event) {
                     ScreenLifecycleEvent.WillAppear -> ScreenWillAppearEvent(it.id)
-                    ScreenLifecycleEvent.Appear -> ScreenAppearEvent(it.id)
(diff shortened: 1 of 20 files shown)


android/src/main/java/com/swmansion/rnscreens/SearchBarView.kt
```diff
@@ -144,6 +144,8 @@ class SearchBarView(reactContext: ReactContext?) : ReactViewGroup(reactContext)
         text?.let { screenStackFragment?.searchView?.setText(it) }
     }
 
+    private val surfaceId = UIManagerHelper.getSurfaceId(this)
+
```
kkafar, line 148: I wonder whether I can cache this value or not, but this computation should not be expensive.

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt
```diff
@@ -192,11 +192,12 @@ open class ScreenFragment : Fragment {
         if (fragment is ScreenStackFragment && fragment.canDispatchEvent(event)) {
             fragment.screen.let {
                 fragment.setLastEventDispatched(event)
+                val surfaceId = UIManagerHelper.getSurfaceId(it)
```
kkafar, line 195: I believe on Paper `surfaceId` gets resolved always to -1 -- it is Paper compatible.

kkafar (commented):
Just few remarks for reviewer.

android/src/main/java/com/swmansion/rnscreens/SearchBarView.kt
WoLewicki: I think we should not cache it since in the future `surfaceId` should change based on the part of the application you are in. For now noone does it, but we won't have to change it later if we do it correctly right now 🤷‍♂️

android/src/main/java/com/swmansion/rnscreens/events/ScreenDismissedEvent.kt
```diff
@@ -1,20 +1,17 @@
...
     override fun getEventName() = EVENT_NAME
 
     // All events for a given view can be coalesced.
     override fun getCoalescingKey(): Short = 0
 
-    override fun dispatch(rctEventEmitter: RCTEventEmitter) {
-        val args = Arguments.createMap()
-        // on Android we always dismiss one screen at a time
-        args.putInt("dismissCount", 1)
-        rctEventEmitter.receiveEvent(viewTag, eventName, args)
```
WoLewicki, line 17: Are you sure we can just remove this code? Is it done like this in other libs/core ?

WoLewicki (commented):
Just a couple of questions 🚀

android/src/main/java/com/swmansion/rnscreens/events/ScreenDismissedEvent.kt
kkafar: Just take a look into [default implementation of this method](https://github.com/facebook/react-native/blob/e2ef6b98a061508bdd70a4c2747fe061c49b5c39/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/uimanager/events/Event.java#L154-L168).

kkafar: Also I've just found [an example](https://github.com/facebook/react-native/commit/72d0ddc16f2f631003c3486e0a59e50c145ec613#diff-afc6f25f459dcf1e24a012e16862c869aa04b0de5795171efe14b7a88be08e64)

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt
kkafar: `UIManagerHelper.getSurfaceId` exists [since 0.65](https://github.com/facebook/react-native/commit/62f0dee2353b14ce1524dc62de5e1d2f1883a089) -> that leaves 0.64 behind -> this PR might wait for a while before it is merged. But this situation nudges me towards thoughts that maybe we should reopen the discussion on what should be the minimal supported RN version, because it is 8 versions back right now.

tboba (approved):
That's a yes from me! But as @WoLewicki said, before merging it let's discuss when we should support RN <0.65 (<0.66?) 👍

kkafar:
As we've reached agreement on dropping RN versions and I've just released 3.25.0 with already merged changes I'm merging this one.

PR #1872, opened 2023-08-21
fix: full window overlay cannot receive tap when modal is full screen

intergalacticspacehighway:
## Description
- `FullWindowOverlay` cannot receive tap events when it is shown on top of `UIModalPresentationFullScreen` Modal which is the default presentation style of the react native Modal.

- UIKit changes the layout hierarchy when the modal is full screen. It is mentioned in the below docs. The view does not move to the window if a full-screen modal is being presented. However, the `_container` moves to the window as it is appended explicitly using `addSubview` in the key window. 
- https://developer.apple.com/library/archive/featuredarticles/ViewControllerPGforiPhoneOS/PresentingaViewController.html

Screenshot 2023-08-21 at 9 45 19 AM

- So the `RNSFullWindowOverlay` view gets attached to the superview but they are not moved to the window. This PR adds a simple fix to replace `didMoveToWindow` with `didMoveToSuperview`. I considered an approach to move `didMoveToWindow` in the `RNSFullWindowOverlayContainer` but I don't see any potential issues with the current approach. Let me know.

## Changes
Attach touch handlers to the container in `didMoveToSuperview` instead of `didMoveToWindow`

Run the below snack on iOS.

- https://snack.expo.dev/@nishanbende/fullwindowoverlay-tap-events-on-full-screen-modal

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSFullWindowOverlay.mm
+++ b/ios/RNSFullWindowOverlay.mm
@@ -127,9 +127,9 @@
   [window addSubview:_container];
 }
 
-- (void)didMoveToWindow
+- (void)didMoveToSuperview
 {
-  if (self.window == nil) {
+  if (self.superview == nil) {
     if (_container != nil) {
       [_container removeFromSuperview];
       [_touchHandler detachFromView:_container];


kkafar (approved):
Hey @intergalacticspacehighway, the changes look good:

1. In case `RNSFullWindowOverlay` is being attached directly under `UIWindow` the old behaviour is maintained (`didMoveToSuperview` is called directly before `didMoveToWindow` by UIKit (-- my experimental conclusions)).
2. In case you described only `didMoveToSuperview` is called. 

I think we can merge it once CI passes. 

Thanks for contributing 🙌🏻

intergalacticspacehighway:
Yeah. That's what i concluded. Thanks for the quick review! 😄

PR #1873, opened 2023-08-21
fix(Android): Fix reporting invalid status bar height on Android

tboba:
## Description

Currently, on Android side it looks like we're using `useSafeAreaInsets` hook from `react-native-safe-area-context` improperly. This hook is sending proper status bar height only if there's option `statusBarTranslucent` set to true. Thus, we should use y-axis position of the frame - `useSafeAreaFrame` is giving us this option with its `y` value.

This change is a part of ongoing work in Pull Request #1802 that implements `use(Re)animatedHeaderHeight` hook.

## Changes

- Added a statement that checks if `statusBarTranslucent` option is set to true for Android.
- Refactorized a logic for getting status bar height and moved it to new function.

### Before

Screenshot 2023-08-21 at 10 22 13

We can see the correct value (from `react-native-safe-area-context` perspective) is being reported wrongly by useSafeAreaInsets after reporting initial inset. This probably occurs because of the ViewGroup being layout after sending initial metrics.

### After

Screenshot 2023-08-21 at 11 23 28

iOS devices are not impacted because of this change.
You can use `Test1844` to test the changes.

## Checklist

- [x] Ensured that CI passes


WoLewicki (commented):
I have no knowledge of how `react-native-safe-area-context` works under the hood, so if it is correct then we can merge it 🚀

src/native-stack/views/NativeStackView.tsx
```diff
@@ -423,6 +425,25 @@ export default function NativeStackView(props: Props) {
   );
 }
 
+function getStatusBarHeight(dimensions: Rect, isStatusBarTranslucent: boolean) {
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const topInset = useSafeAreaInsets().top;
```
kkafar, line 430: We can pass it as a argument to the function to avoid breaking eslint / react rules.

kkafar (approved):
The changes look fine, I haven't testem them in separation, will do it with whole context of the main PR.

src/native-stack/views/NativeStackView.tsx
tboba: Resolved in https://github.com/software-mansion/react-native-screens/pull/1873/commits/a19dbd98d529f7af5602f28d628348ce01f1db94.

PR #1883, opened 2023-09-06
fix(Android): Change implementation of `headerConfig` prop on Android

tboba:
## Description

It looks like the change introduced in #1825 has also came with a bug where at least Search Bar was not working on Android.
The reason for that has turned out to be the wrong implementation of `headerConfig` property on Android, which was looking for the first child and assuming that it is ScreenStackHeaderConfig - which was wrong, because the change in #1825 has moved the ScreenStackHeaderConfig to the bottom of the hierarchy.

Fixes not working Search Bar on Android (and some other unknown things maybe?).

## Changes

From now `Screen#headerConfig` will traverse through all children of the View and will check if one of them is `ScreenStackHeaderConfig`.

You can check Test1166 if SearchBar is working right now; from now the Search Bar in Example section should work properly.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -8,6 +8,7 @@ import android.util.SparseArray
 import android.view.ViewGroup
 import android.view.WindowManager
 import android.webkit.WebView
+import androidx.core.view.children
 import com.facebook.react.bridge.GuardedRunnable
 import com.facebook.react.bridge.ReactContext
 import com.facebook.react.uimanager.UIManagerModule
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -84,7 +85,15 @@ class Screen constructor(context: ReactContext?) : FabricEnabledViewGroup(contex
...
-        get() = getChildAt(0) as? ScreenStackHeaderConfig
+        get() {
+            for (child in children) {
+                if (child is ScreenStackHeaderConfig) {
+                    return child
+                }
+            }
+
+            return null
+        }
```
kkafar, line 96: we can have it done in more sassy way
```suggestion
        get() = children.find { it is ScreenStackHeaderConfig } as? ScreenStackHeaderConfig
```

kkafar (changes requested):
Great catch!

android/src/main/java/com/swmansion/rnscreens/Screen.kt
tboba: Thanks for the hint! Resolved in https://github.com/software-mansion/react-native-screens/pull/1883/commits/74c4f962631dfc1b0b0e6c95a8cf025b6032245b.

kkafar (approved):
CI 🟢 ➡️ merge

PR #1887, opened 2023-09-13
refact(Android): allow for different fragment types inside ScreenContainer

kkafar:
## Description

This PR is laying out ground work under not-so-distant Android modal & bottom sheet implementation.

Currently `ScreenContainer<T: ScreenFragment>: ViewGroup` (and thus `ScreenStack: ScreenContainer<ScreenStackFragment>`) operates on `ScreenFragment: Fragment` objects, making it impossible to use any other `Fragment` subclass with the stack (only single inheritance is possible on JVM). 

This PR changes the situation by introducing `ScreenFragmentWrapper` interface which both:

1. holds the reference to underlaying `Fragment`
2. provides additional API previously directly attached to the `ScreenFragment`

Also introduced `ScreenStackFragmentWrapper` extends `ScreenFragmentWrapper` with API previously attached to `ScreenStackFragment`. 

Thereafter by implementing such interface by the fragment subclass itself we can have both of worlds:

1. Customized fragment behaviour
2. Implementation of all methods required for fragment<->container communication.

Disclaimer: *I have decided to split "Android modals PR" into few smaller ones to reduce its size and complexity. I won't use "stack PR" technique to spare myself whole bunch of rebasing & conflict resolving.*

## Changes

1. `ScreenContainer` is no longer parameterized with `T: ScreenFragment`. It uses `ScreenFragmentWrapper` as screen-primitve base type now.
2. Our fragment types `ScreenFragment` & `ScreenStackFragment` now implement `ScreenFragmentWrapper` & `ScreenStackFragmentWrapper` interfaces respectively. 
3. `ScreenContainer` has access to the fragments it operates on (fragment API) via `ScreenFragmentWrapper.fragment` property.
4. Our custom methods previously implemented directly on `Fragment` subclass are exposed via the new interfaces so `ScreenContainer` still has access to them. 

See that test examples build & CI passes.

I did some manual testing in our example apps & didn't notice any regression. 

## Checklist

- [x] Ensured that CI passes


android/src/main/java/com/swmansion/rnscreens/FragmentHolder.kt
```diff
@@ -0,0 +1,7 @@
+package com.swmansion.rnscreens
+
+import androidx.fragment.app.Fragment
+
+interface FragmentHolder {
+    val fragment: Fragment
+}
```
kkafar, line 7: Decided to split this property to separate interface as this is the key idea behind this PR.

Also in the future I plan to make `ScreenFragmentWrapper` interfaces much more granular (split those into number of fine-grained interfaces.

android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt
```diff
@@ -16,9 +16,9 @@ import com.facebook.react.modules.core.ChoreographerCompat
 import com.facebook.react.modules.core.ReactChoreographer
 import com.swmansion.rnscreens.Screen.ActivityState
 
-open class ScreenContainer<T : ScreenFragment>(context: Context?) : ViewGroup(context) {
+open class ScreenContainer(context: Context?) : ViewGroup(context) {
```
kkafar, line 19: `ScreenContainer` type is no longer parameterized with `ScreenFragment` type. Instead whole fragment type variance is handled by `ScreenFragmentWrapper` interface.

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt
```diff
@@ -25,15 +25,18 @@ import com.swmansion.rnscreens.events.ScreenWillDisappearEvent
...
 
-open class ScreenFragment : Fragment {
+open class ScreenFragment : Fragment, ScreenFragmentWrapper {
     enum class ScreenLifecycleEvent {
         Appear, WillAppear, Disappear, WillDisappear
     }
 
+    override val fragment: Fragment
+        get() = this
+
```
kkafar, line 35: Fragment subclass itself implements the new interface. 

Thanks to `fragment` property:

```kotlin
override val fragment: Fragment
    get() = this
```

we can refer to this specific `Fragment` instance inside `ScreenContainer`.

Please note also that we are not creating any strong-reference cycle here as this property is computed on-the-fly.

kkafar (commented):
Few remarks for the future & reviewers

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt
```diff
@@ -149,17 +152,19 @@ open class ScreenFragment : Fragment {
         return null
     }
 
-    val childScreenContainers: List<ScreenContainer<*>>
+    override val childScreenContainers: List<ScreenContainer>
         get() = mChildScreenContainers
 
-    private fun canDispatchEvent(event: ScreenLifecycleEvent): Boolean = when (event) {
+//    override fun canDispatchLifecycleEvent(event: ScreenLifecycleEvent): Boolean = canDispatchEvent(event)
```
WoLewicki, line 158: Probably leftover.

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt
```diff
@@ -169,29 +174,30 @@ open class ScreenFragment : Fragment {
     }
 
     private fun dispatchOnWillAppear() {
```
WoLewicki, line 176: I think all these event methods should be moved to `ScreenStackFragment` and I am not sure why are they even in this class at all. We only use them in `native-stack` since JS handles the lifecycle and animation progress events in JS stack.

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt
```diff
@@ -247,19 +253,19 @@ open class ScreenFragment : Fragment {
         }
     }
 
-    fun registerChildScreenContainer(screenContainer: ScreenContainer<*>) {
-        mChildScreenContainers.add(screenContainer)
+    override fun addChildContainer(container: ScreenContainer) {
```
WoLewicki, line 256: Why do you change the method name without changing `mChildScreenContainers` name? What was wrong with the previous name?

android/src/main/java/com/swmansion/rnscreens/ext/ViewExt.kt
```diff
@@ -0,0 +1,18 @@
+package com.swmansion.rnscreens.ext
+
+import android.view.View
+import android.view.ViewGroup
+
+fun View.recycle(): View {
```
WoLewicki, line 6: What is it and why is it in this PR?

WoLewicki (commented):
I added some comments, other than them, LGTM 🚀

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt
kkafar: There is some logic that dispatches events in `ScreenFragment` (e.g.  in `onViewAnimationEnd()`) and uses it. I'll consider moving it in follow up refactors.

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt
kkafar: https://github.com/software-mansion/react-native-screens/pull/1887/commits/de2dadc13ea080c6eedc52a5e899b44e34148a83

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt
kkafar: I just wanted to change prefixes: `register/unregister` -> `add/remove`, the `...Screen...` part should be there.

 https://github.com/software-mansion/react-native-screens/pull/1887/commits/e947d9ab9fde981ff955d632cee543ced8367e8c

android/src/main/java/com/swmansion/rnscreens/ext/ViewExt.kt
kkafar: Cherrypicked one commit too much. 

https://github.com/software-mansion/react-native-screens/pull/1887/commits/ee40686d50d566f80fca4f4947b6278b50710c08

WoLewicki (approved):

PR #1894, opened 2023-09-18
feat(Android): Add focused states on page transitions

### LINKED ISSUE #1706: [ANDROID TV] Focus gets lost between page transitions
### Description

Focus management is very important for TV applications and there's this big problem that has been hurting `react-native-tvos` community [for a long time](https://github.com/react-native-tvos/react-native-tvos/issues/237).

TV app users typically navigate around the UI items using their remote's d-pad. On every d-pad action, the currently focused element changes on the screen. It's crucial to restore the focus when user navigates to a different screen and then comes back to the previous one. Users should always continue their journey where they left off.

I'm assuming due to  working principles `react-native-screens`, Android loses the track of last focused element and hence can not recover the focus natively. As a result, a new focus search begins on every page transition and focus gets lost. Here's how it looks:

https://user-images.githubusercontent.com/22980987/215900717-12ae4e2e-5ab6-4d5b-95a3-ec5883209a23.mov

As can be seen on the screen recording, I navigate to a screen by triggering `onPress` action of the `(4) Text input example` -- I'll call it Button 4 -- button. Later, I come back to the previous screen using the back button. Since `Button 4` was the last element that had the focus on the previous screen, it should regain the focus when we navigate back.

Here's the expected behavior:

https://user-images.githubusercontent.com/22980987/215901324-91542862-c8d0-43e5-a56f-cb8a0c77c701.mov

### Steps to reproduce

1. Clone https://github.com/yusufyildirim/TVInputDemo/tree/focus-bug-repro repo
2. Run the app using `yarn android` command
3. Focus one of the elements on the screen and press enter to navigate to another screen
4. Come back to previous screen
5. See the focus is lost

https://github.com/yusufyildirim/TVInputDemo/tree/focus-bug-repro repo

### Screens version

3.18.2

### React Native version

0.69.6

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

None

### Architecture

Paper (Old Architecture)

### Build type

None

### Device

Android emulator

### Device model

_No response_

### Acknowledgements

Yes

tboba:
## Description

Currently, on Android TV devices when user is switching between views the last focused child is lost. This seems like an incorrect behavior - In #1706 people are reporting that on Android TV the last focused element should be stored.

This PR introduces the change to remember last focused element on previous screen. 
For now I've added a statement that the last element should be stored only for Android TV, but we should discuss if we also should store those elements for classic Android - I've noticed that it's possible to focus an element on a normal screen, but I've achieved it only by switching between elements with arrows on keyboard (so is it possible to focus an element just by touching a screen? Maybe TalkBack is also focusing the elements?). 

<details><summary>
Focusing elements on Android
</summary>

https://github.com/software-mansion/react-native-screens/assets/23281839/4f62ae66-f411-4c45-b678-1e47c3fa4ff2

</details>

Fixes #1706. 

## Changes

- Added `isTelevision` method that checks if user is currently using Android TV.
- Added `lastFocusedChild` variable with an implementation for finding last focused elements.

## Screenshots / GIFs

### Before

https://github.com/software-mansion/react-native-screens/assets/23281839/0e3e0ad4-a202-43c3-b688-6c66f76a43f8

### After

https://github.com/software-mansion/react-native-screens/assets/23281839/998dfa5c-6def-47f2-b129-425a2257319a

You can test those changes on `Example` application by switching between examples with arrows on the keyboard.

## Checklist

- [ ] Ensured that CI passes


android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
```diff
@@ -123,6 +130,11 @@ class ScreenStackFragment : ScreenFragment, ScreenStackFragmentWrapper {
         return view
     }
 
+    override fun onStop() {
+        mLastFocusedChild = findLastFocusedChild()
```
yusufyildirim, line 134:
```suggestion
        if (isTelevision()) mLastFocusedChild = findLastFocusedChild()
```

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
```diff
@@ -163,6 +175,19 @@ class ScreenStackFragment : ScreenFragment, ScreenStackFragmentWrapper {
         }
     }
 
+    private fun findLastFocusedChild(): View? {
+        // If current device is not television, we don't want to save last focused child.
+        if (!isTelevision()) return null
```
yusufyildirim, line 180:
```suggestion
```

yusufyildirim (commented):
Sort of nit: It might be better to preserve `findLastFocusedChild` function's responsibility and move the `if(isTelevision())` logic to the actual spot where you call the function. Awesome work!

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
tboba: Sure, thanks for the suggestion! Changed in https://github.com/software-mansion/react-native-screens/pull/1894/commits/4466f8d6df6cac984534a13781bbc6f94f0491ec.

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
tboba: Changed in https://github.com/software-mansion/react-native-screens/pull/1894/commits/4466f8d6df6cac984534a13781bbc6f94f0491ec.

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt
```diff
@@ -155,6 +158,10 @@ open class ScreenFragment : Fragment, ScreenFragmentWrapper {
     override val childScreenContainers: List<ScreenContainer>
         get() = mChildScreenContainers
 
+    fun isTelevision(): Boolean {
+        return context?.getSystemService<UiModeManager>()?.currentModeType == Configuration.UI_MODE_TYPE_TELEVISION
+    }
+
```
kkafar, line 164: Is there a particular reason that all changes to the logic are in `ScreenStackFragment` but this method is defined here? 

Just nitpicking here: I would put this method in some kind of util module(`Util.kt`) as such usage does not make much sense for me: `fragmentInstance.isTelevision` (what does it mean that fragment is a television? :D)

My vision:

`fun isPlatformAndroidTV(context: Context?): Boolean`


Also I've noticed that preferred way of resolving the platform [has changed](https://developer.android.com/training/tv/start/hardware.html#runtime-check) since [this  SO answer](https://stackoverflow.com/a/30059511/8635522) :D

kkafar (commented):
Left some comment below

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt
tboba: To be honest I was thinking if I should create a new util class just for one method - that's why I thought about moving it to `ScreenStackFragment`, but you're correct that it should be moved somewhere else 😄

Some things to mention about the technique of detecting a device - I went for the "old" method because for some reason my Android Emulator is still being detected as a non-leanback device 🤔 But that could be my mistake, actually - when I was testing that method I was detecting the device with variable:

```kt
var isTelevision = context?.packageManager?.hasSystemFeature(PackageManager.FEATURE_LEANBACK) == true
```
In this case, `PackageManager#hasSystemFeature` and `UiModeManager#getCurrentModeType` were returning false on Android and Android TV, whereas
```kt
fun isTelevision(): Boolean {
  return context?.getSystemService<UiModeManager>()?.currentModeType == Configuration.UI_MODE_TYPE_TELEVISION
}
```

was returning `true` on Android TV and `false` on Android - so it seems that there is a difference between `variable-statement` and single return in a function. I'll try to test if single return in a function for `PackageManager#hasSystemFeature` will work 👍 

But except this reason I'm also in favor of using UiModeManager because it checks not only the system feature, but the whole device - so we could check in the future if the device is a watch, a car, etc. - with system feature it's possible just check if device has a camera, NFC and other things, related for couple of devices, so for me it could be easier in the future to just add one `case`, instead of changing whole method. What do you think about it @kkafar?

kkafar:
```kotlin
var isTelevision = context?.packageManager?.hasSystemFeature(PackageManager.FEATURE_LEANBACK) == true
```

This is evaluated only a single time during variable initialisation. If you want to achieve effect of computed variable you should use:

```kotlin
val isTelevision
    get() = ...
```

If the recommended way of doing things does not work, surely we should go with workaround 😄 But let's make sure it is necessary.

yusufyildirim: @tboba Make sure your test project has `LEANBACK_LAUNCHER` intent filter in AndroidManifest. See https://developer.android.com/training/tv/start/start#tv-activity

yusufyildirim: And maybe also https://developer.android.com/training/tv/start/start#leanback-req

tboba: @yusufyildirim Thanks for the tips! I've managed to make `hasSystemFeature` work on Android TV without steps you've mentioned, but they may be a good point for investigating why our example apps don't darken the buttons when someone is focusing them 🤔

tboba: Moved `isTelevision` to DeviceUtils in https://github.com/software-mansion/react-native-screens/commit/f31aee27daa5f09e6a82acdcd8a37e4867b17acd (which is not visible in this PR for some reason at the moment :thinking:)

kkafar (approved):
Has the "recommended way" started working?

tboba:
@kkafar Yep, it works now - however, I'm not satisfied with it, as it checks the feature inside the OS - not the device. Just a small note for us - if there will be an opportunity to add support for watches, car infotainments or something similar maybe we should reconsider using the old way of checking the device 😛

PR #1895, opened 2023-09-20
fix(iOS): Fix positioning of large header and search bar

### LINKED ISSUE #1884: headerLargeTitle & headerSearchBarOptions Doesn't Support RTL
### Description

When RTL is enabled, the headerLargeTitle & headerSearchBarOptions don't honour the text direction. Instead they remain LTR as shown in the below screen cuttings. 
- Restarting the app between changing text direction makes no difference. 
- The fact the text itself isn't a RTL language makes no difference. 
- Having the device in a RTL language makes no difference.
- Doesn't matter if you're using Paper or Fabric or the new/old arch.
- Doesn't matter if you're running in debug or release.
- Doesn't matter if you're running on a real or sim device.

**headerLargeTitle**
image
**headerSearchBarOptions(focused)**
image

I would expect them to be positioned right as shown below (taken from the contacts app)

**Large Title**
image
**Search bar(focused)**
image

### Steps to reproduce

1. Set the options direction: 'rtl'
2. Not sure if required, but Set I18nManager.forceRTL(true); in the App.js too
3. Restart the app by closing it completely from background and re-launching

Expected
- For text direction to be RTL for headerLargeTitle & headerSearchBarOptions

Actual
- headerLargeTitle & headerSearchBarOptions are still LTR

https://github.com/0xRenshaw/react-native-screens/tree/main/FabricExample

### Screens version

3.25.0

### React Native version

0.72.4

### Platforms

iOS

### JavaScript runtime

None

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Release mode

### Device

Real device

### Device model

_No response_

### Acknowledgements

Yes

tboba:
## Description

In the current state, the RTL mode didn't work correctly for large headers and search bar - even if user sets RTL mode in their app, the content of the large header and search was still on the left side. Also, what I've found, the icon of the back button was still pointing to the left, even if the back button was on the right side of the screen.

Fixes #1884.

## Changes

- Added a workaround for pointing arrow to the left - unfortunately I couldn't find a better solution for changing the direction of the arrow.
- Changed the direction of the contents in the navigation bar when user is in RTL mode.

### Before

image <br/>

image

### After

image<br/>

Screenshot 2023-09-20 at 10 27 51

1. Launch example application from `Example` directory.
2. Switch `Right to left` option to `true`
3. Check `Header Options` and `Search bar` tabs to see the difference.
4. You can also change `headerHideBackButton` and `headerLargeTitle` options inside the stack's navigator and go to `Other Searchbar example` to check how large header behaves with the search bar.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -567,6 +567,8 @@ namespace react = facebook::react;
 #endif
   }
 #if !TARGET_OS_TV
+  // Workaround for the wrong rotation of back button arrow in RTL mode.
+  navitem.hidesBackButton = true;
   navitem.hidesBackButton = config.hideBackButton;
 #endif
   navitem.leftBarButtonItem = nil;
@@ -623,6 +625,13 @@ namespace react = facebook::react;
       }
     }
   }
+
+  dispatch_async(dispatch_get_main_queue(), ^{
+    // Position the contents in the navigation bar, regarding to the direction.
+    for (UIView *view in navctr.navigationBar.subviews) {
+      view.semanticContentAttribute = config.direction;
+    }
+  });
 
   // This assignment should be done after `navitem.titleView = ...` assignment (iOS 16.0 bug).
   // See: https://github.com/software-mansion/react-native-screens/issues/1570 (comments)
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -626,7 +626,7 @@ namespace react = facebook::react;
     }
   }
 
-  dispatch_async(dispatch_get_main_queue(), ^{
+  dispatch_after(dispatch_time(DISPATCH_TIME_NOW, 0), dispatch_get_main_queue(), ^{
     // Position the contents in the navigation bar, regarding to the direction.
     for (UIView *view in navctr.navigationBar.subviews) {
       view.semanticContentAttribute = config.direction;


ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -624,6 +626,13 @@ + (void)updateViewController:(UIViewController *)vc
     }
   }
 
+  dispatch_async(dispatch_get_main_queue(), ^{
+    // Position the contents in the navigation bar, regarding to the direction.
+    for (UIView *view in navctr.navigationBar.subviews) {
+      view.semanticContentAttribute = config.direction;
+    }
+  });
```
tboba, line 634: One comment regarding this line - Unfortunately I haven't managed to set `semanticContentAttribute` without the `dispatch_async` method. The problem is that:

 >iOS applies appearance changes when a view enters a window, it doesn’t change the appearance of a view >that’s already in a window. To change the appearance of a view that’s currently in a window, remove the view >from the view hierarchy and then put it back.

(via [SO](https://stackoverflow.com/questions/49277773/semanticcontentattribute-is-not-working) and Apple docs)

with is quite problematic for us, because we're setting attributes from config when the `ScreenStackHeaderConfig` is already mounted to the view, so dispatching it asynchronously is a good workaround for setting those content attributes.

kkafar: If so, can not we move the logic responsible for applying content attributes to some other place that would give us opportunity to set this earlier? 

> which is quite problematic for us, because we're setting attributes from config when the ScreenStackHeaderConfig is already mounted to the view, so dispatching it asynchronously is a good workaround for setting those content attributes.

IIRC `ScreenStackHeaderConfig` is not attached to view hierarchy at all on Paper. My guess is that the problem arises because the moment we receive header config (it is added by RN via `insertSubview:atIndex:` method (writing from memory here)) the owning `Screen` and it's view controller are already mounted into hierarchy. If it is really the case I see two options:

1. Hack it on JS side so that screen is aware of ltr/rtl direction (is it not already?) and set content attriubutes at the moment screen is inserted as the child of stack view by RN
2. Go with the workaround you suggested. But tbh I do not understand why it works

kkafar (commented):
Answered your comment

ios/RNSScreenStackHeaderConfig.mm
tboba: I think we've discussed most of the things during the weekly meeting :D but I was unable to move the implementation of changing the position of elements inside `addSubview / insertReactSubview` 😕 

Moreover, I've discovered that what the documentation of `UIAppearance` says does not fit to our case 🤔 For changing the mode of elements to RTL the view of the navigation bar should be **already** mounted, so we need to apply the changes quickly after the view will appear - that's why I'm still standing for leaving the implementation of setting `semanticContentAttribute` in `dispatch_after` (previously this was `dispatch_async`, but I've changed it, as dispatch_async is dispatching the contents of its block immediately, whereas dispatch_after is executing it just after the block where its stays finishes its own execution).

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -567,6 +567,8 @@ + (void)updateViewController:(UIViewController *)vc
 #endif
   }
 #if !TARGET_OS_TV
+  // Workaround for the wrong rotation of back button arrow in RTL mode.
+  navitem.hidesBackButton = true;
   navitem.hidesBackButton = config.hideBackButton;
```
kkafar, line 572: Hey, would you mind describing why do we first set `navitem.hidesBackButton = true` and then immediately we set it to the `config.hideBackButton`? Whats the mechanism here?

kkafar (commented):
Asked a question in order to clarify

ios/RNSScreenStackHeaderConfig.mm
tboba: That's because when you're setting your application to RTL mode the arrow in the back button is still pointing to the left - which is wrong, because in RTL mode you should have an arrow pointing to the right. It seems that hiding back button for a while fixes this issue, because any update related to the back button modification updates the it. To be fair I don't know why this is working, unfortunately changing the `semanticContentAttribute` in `dispatch_after` doesn't fix that either.

WoLewicki (approved):
As long as it does not introduce any regressions, I am fine with it :rocket:

STonkoshkur:
@tboba the back button direction is still incorrect in RTL for iOS. I'm running the example application from `Example` directory using Xcode 15.2

image

tboba:
@STonkoshkur interesting, do you have some native apps that have RTL mode enabled? I just wanted to see how the back button should look like. In my case, I have the same back button as you, but if I unhide the default navigation bar of the navigation controller, it looks completely the same 🤔 

image

STonkoshkur:
@tboba sorry, I've moved my comment to the parent issue

I'm pretty sure back button should look like the one provided in the after-fix demo above:

image

tboba:
@STonkoshkur Yeah, I moved my comment also there 😄 
Oh, yeaaaah you're right. I believe this behavior might have changed after the iOS 17.2 release. I need to take an eye onto this 🤔

PR #1898, opened 2023-09-21
fix(CI): fix Android E2E Events test

tboba:
## Description

Every now and then our CI was failing on the specific test - `events.e2e.ts`. The issue was that the toasts that were displayed during the test were dismissed too quickly (even after 40 seconds of waiting). 
This PR fixes the problem with checking if the application is running on the Detox server.

When I was making that change I was thinking about couple approaches how to detect if the machine that runs the tests is Detox:
❌ Using `react-native-is-detox` lib: I didn't want to have additional lib that seems to be not maintained actively + this approach is error-prone: there's for sure a better alternative than checking the environment natively
❌ Using .env files as environment libraries (with `react-native-config` lib): it's a great idea for using multiple values but it's a huge over-exaggerate when dealing only with single state
✅ I think using launch arguments is a best approach here, as it is lightweight while using only one value.

## Changes

- Fixed failing E2E test (`events.e2e.ts`)
- Decreased the amount of time to dismiss the toast

## Checklist
- [ ] Ensured that CI passes


tboba:
Closing, as we're planing to experiment with switching our current CI engine to Maestro + it looks like my change still doesn't fix the results :(

PR #1903, opened 2023-09-27
fix(Android): Change elements visibility on search bar open

### LINKED ISSUE #1450: Search bar on Android incorrectly adjusts other header elements when open
## Description

On Android, the search bar does not properly adjust header elements.
- Elements placed via `headerRight` are "pushed" out of the way and to the left of the search bar
- If there are any elements placed via `headerLeft`, both the screen title and said elements are not adjusted at all and appear under the search bar

### Screenshots
Screen Shot 2022-05-17 at 10 33 52 AM

Screen Shot 2022-05-17 at 10 34 05 AM

## Steps To Reproduce

1. Add a search bar to the header by specifying `headerSearchBarOptions`
2. Add any element(s) via the `headerLeft` navigation option
3. (Optional) Add any elements to the `headerRight` navigation option
4. Open the search bar on android

### Expected behavior
- The header buttons on both the left side and the right side should no longer be visible. 
- The search bar should appear as the only element in the header.
- The input should be unobstructed.

### Actual behavior
- Any elements placed via `headerRight` seem to get "pushed" out of the way to the left of the search bar
- The screen title as well as any elements placed by `headerLeft` are not adjusted and appear under the search bar

## Reproduction

https://snack.expo.dev/@nrwinner/android-search-bug---react-native-screens-repro

## Platform

- [ ] iOS
- [x] Android
- [ ] Web
- [ ] Windows
- [ ] tvOS

## Workflow

- [x] Managed workflow
- [ ] Bare workflow

## Package versions

| react-native                           |    0.68.1     |
| @react-navigation/native               |    6.0.2     |
| @react-navigation/native-stack         |    6.1.0     |
| react-native-screens                   |   3.13.1      |
| react-native-safe-area-context         |     4.2.4    |
| expo                                   |     45.0.0    |

tboba:
## Description

Currently, when user opens the search bar all of the subviews on the toolbar are being pushed to the left.
Instead, we want to hide them or even remove them so that they won't be visible during the layout.
This PR fixes that problem by secretly hiding the toolbar elements, changing their visibility to `GONE`.

Fixes #1450.

## Changes

- Added a fix that is being controlled by `handleOpen` and `handleClose` methods in SearchBarView.

### Before

https://github.com/software-mansion/react-native-screens/assets/23281839/8ca8e18e-2cf6-4c04-85c8-51b5d69f1a61

### After

https://github.com/software-mansion/react-native-screens/assets/23281839/41330ea7-a12b-4e14-9a1e-2207a54ebed0

In our example app open `SearchBar.tsx` and add `headerRight: () => <Button title="Test"/> to the `navigation.setOptions({` declaration. Then launch the app and open `Search bar` example.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/SearchBarView.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/SearchBarView.kt
@@ -32,14 +32,18 @@ class SearchBarView(reactContext: ReactContext?) : ReactViewGroup(reactContext) 
 
     private var mAreListenersSet: Boolean = false
 
-    private val screenStackFragment: ScreenStackFragment?
+    private val headerConfig: ScreenStackHeaderConfig?
         get() {
             val currentParent = parent
             if (currentParent is ScreenStackHeaderSubview) {
-                return currentParent.config?.screenFragment
+                return currentParent.config
             }
+
             return null
         }
+
+    private val screenStackFragment: ScreenStackFragment?
+        get() = headerConfig?.screenFragment
 
     fun onUpdate() {
         setSearchViewProps()
@@ -110,10 +114,12 @@ class SearchBarView(reactContext: ReactContext?) : ReactViewGroup(reactContext) 
 
     private fun handleClose() {
         sendEvent(SearchBarCloseEvent(surfaceId, id))
+        setToolbarElementsVisibility(VISIBLE)
     }
 
     private fun handleOpen() {
         sendEvent(SearchBarOpenEvent(surfaceId, id))
+        setToolbarElementsVisibility(GONE)
     }
 
     private fun handleTextSubmit(newText: String?) {
@@ -142,6 +148,14 @@ class SearchBarView(reactContext: ReactContext?) : ReactViewGroup(reactContext) 
 
     fun handleSetTextJsRequest(text: String?) {
         text?.let { screenStackFragment?.searchView?.setText(it) }
+    }
+
+    private fun setToolbarElementsVisibility(visibility: Int) {
+        for (i in 0..(headerConfig?.configSubviewsCount?.minus(1) ?: 0)) {
+            val subview = headerConfig?.getConfigSubview(i)
+            if (subview?.type != ScreenStackHeaderSubview.Type.SEARCH_BAR)
+                subview?.visibility = visibility
+        }
     }
 
     private val surfaceId = UIManagerHelper.getSurfaceId(this)


android/src/main/java/com/swmansion/rnscreens/SearchBarView.kt
```diff
@@ -144,6 +150,14 @@ class SearchBarView(reactContext: ReactContext?) : ReactViewGroup(reactContext)
         text?.let { screenStackFragment?.searchView?.setText(it) }
     }
 
+    private fun setToolbarElementsVisibility(visibility: Int) {
+        for (i in 0..(headerConfig?.configSubviewsCount?.minus(1) ?: 0)) {
```
tboba, line 154: Just a one note about this line - since we need to iterate over the subviews I need to have closed range on the end of the subviews array. That's why I'm decrementing the count by one, but we can make this line better with the newly-introduced `rangeUntil` method for the ranges:

```suggestion
        val subviewsCount = headerConfig?.configSubviewsCount? ?: 0
        for (i in 0..<subviewsCount) {
```

which won't count the last element of a range. Unfortunately this requires us to update Kotlin to 1.9.10 - that's why I've made a PR #1904 to update Kotlin version.

android/src/main/java/com/swmansion/rnscreens/SearchBarView.kt
```diff
@@ -110,10 +114,12 @@ class SearchBarView(reactContext: ReactContext?) : ReactViewGroup(reactContext)
 
     private fun handleClose() {
         sendEvent(SearchBarCloseEvent(surfaceId, id))
+        setToolbarElementsVisibility(VISIBLE)
```
WoLewicki, line 117: Is there a possibility that this won't be fired, for example is we navigated when the search bar was open?

WoLewicki (approved):
I left one comment, please answer it. Other than that, LGTM :rocket:

android/src/main/java/com/swmansion/rnscreens/SearchBarView.kt
tboba: @WoLewicki I thought about that already and I'm unable to reproduce the scenario where the user comes back from the search bar and in the same action comes to the previous screen. Is it even possible? 🤔 

https://github.com/software-mansion/react-native-screens/assets/23281839/30b1bade-8274-41fb-bc6e-9e972c3ca8e2

tboba: So, I figured out how to do that (I'm calling `navigation.navigate('Main')` on `onClose`) and it's doing fine!

https://github.com/software-mansion/react-native-screens/assets/23281839/d08323fe-8b26-4ae1-aa2a-d3b839432097

android/src/main/java/com/swmansion/rnscreens/SearchBarView.kt
kkafar: I believe that using any recent Kotlin syntax would force users to manually bump Kotlin versions in their applications build scripts. We should not do that.

kkafar (commented):
Hey, GJ overall. 
Left single comment.

android/src/main/java/com/swmansion/rnscreens/SearchBarView.kt
tboba: @kkafar Sure!

kkafar (approved):
🟢

android/src/main/java/com/swmansion/rnscreens/SearchBarView.kt
WoLewicki: I think a better example would be to navigate forward and see what happens when you go back.

tboba: @WoLewicki So I've checked the scenario that you've suggested and looks like everything still works - I can't manage to have search bar closed when I'm going to the next screen and coming to the previous one.

https://github.com/software-mansion/react-native-screens/assets/23281839/269faa3d-1e8d-433d-b675-83c38a9ec3fe

I'm still thinking if changing the visibility of the element could still break at some point. Do you think we should correct the change just to make sure that we will avoid side-effects?

WoLewicki: I think it would be some weird edge case and we can leave it as it is for now, and if the issues arise in the future, we will fix it then.

Issue #1849, opened 2023-07-29
TR SYSTEM LANGUAGE PROBLEM


mehmettasan:
### Description

After installing the React Navigaton package, I get an error when I try to install this package while the system language is Turkish. When I change the system language to English, it installs without any problems. I am using windows 11 and the operating system is up to date. 
**this screenshot is for react-native-safe-area-context but same problem exists in this package.**
![Ekran görüntüsü_20230728_183744](https://github.com/software-mansion/react-native-screens/assets/83862655/08f01533-ddf5-48ff-9d12-86dcb7c4eeee)

### Steps to reproduce

1- change system language turkish
2- npm install @react-navigation/native
3- npm install react-native-screens react-native-safe-area-context
4- npx react-native run-android

### Snack or a link to a repository

none

### Screens version

3.23.0

### React Native version

0.72.3

### Platforms

Android, Windows

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

mehmettasan:
> 

i trying with clean project. if you want see this problem you can change your windows language turkish and just run project.

mehmettasan:
i can share a repo but if you using any language you cant get a error.

kkafar:
Hey @mehmettasan,

looking at the screenshot you provided, my first impression is that the problem might be  related to character encoding used by your language (notice that every `I` letter is replaced with some weird symbol)? To my best knowledge Turkish uses some diacritic signs that might not be encoded with ASCII (maybe even UTF-8) and different encoding could be problematic for compilers to handle? That is just my first guess and I must disclaim that my expertise is very limited in that matter 😅 However I'm not aware of anything special in any of those libs you mentioned (especially `react-native-screens`) that could lead to such problems. 

Is this problem of yours specific only to these libs? What happens if you try to build application that do not use them? Does the project compile then?

mehmettasan:
> Hey @mehmettasan,
> 
> looking at the screenshot you provided, my first impression is that the problem might be related to character encoding used by your language (notice that every `I` letter is replaced with some weird symbol)? To my best knowledge Turkish uses some diacritic signs that might not be encoded with ASCII (maybe even UTF-8) and different encoding could be problematic for compilers to handle? That is just my first guess and I must disclaim that my expertise is very limited in that matter 😅 However I'm not aware of anything special in any of those libs you mentioned (especially `react-native-screens`) that could lead to such problems.
> 
[...]

PR #1904, opened 2023-09-27
chore: Update Kotlin to 1.9.10

tboba:
## Description

Since we're still using Kotlin 1.8 I think it's a good idea to update kotlin to the newest 1.9.10 version.
Specifically I want to use one of the benefits of this new version in #1903 where I could use new `rangeUntil` method in `for` statement.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/Example/android/build.gradle
+++ b/Example/android/build.gradle
@@ -7,7 +7,7 @@ buildscript {
         compileSdkVersion = 33
         targetSdkVersion = 33
         ndkVersion = "23.1.7779620"
-        kotlinVersion = "1.8.22"
+        kotlinVersion = "1.9.10"
     }
     repositories {
         google()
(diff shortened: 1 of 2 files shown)


kkafar (commented):
Hey, I've just recently [bumped](https://github.com/software-mansion/react-native-screens/pull/1891) default Kotlin version to `1.8.0` in order to align it with the version in RN application template and potentially fix some [localisation issues](https://github.com/software-mansion/react-native-screens/issues/1849). 

Arguments against bumping the version again & using any fresh syntax (such as `rangeUntil`):

1. `rnsDefaultKotlinVersion` is only a fallback value. It is being used only in case developer did not specify `kotlinVersion` in their application `build.gradle` script. And since RN project template specifies this, you should not expect that the library will be compiled with `rnsDefaultKotlinVersion` Kotlin version.
2. Using any recent addition to Kotlin would require the library to be compiled with recent compiler --> forcing users to manually bump Kotlin version in their build scripts. We do not really want to do that.

tboba:
@kkafar great points there, I didn't know that you've already updated Kotlin to 1.8.0 😅 
Let's not upgrade it for now then.

PR #1905, opened 2023-09-28
chore: Move E2E tests to Maestro

tboba:
## Description

WIP WIP

## Changes

WIP WIP

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


khagesh:
I see that you are also running into same issue of identifying back button. Is there a way we can pass testID or accessibilityIdentifier to back button on Stack header config props?

khagesh:
I added this patch to make maestro pick testID

```patch
diff --git a/node_modules/react-native-screens/ios/RNSScreenStackHeaderConfig.mm b/node_modules/react-native-screens/ios/RNSScreenStackHeaderConfig.mm
index a5eb013..0e4cd3c 100644
--- a/node_modules/react-native-screens/ios/RNSScreenStackHeaderConfig.mm
+++ b/node_modules/react-native-screens/ios/RNSScreenStackHeaderConfig.mm
@@ -495,8 +495,11 @@ + (void)updateViewController:(UIViewController *)vc
                                                                              target:nil
                                                                              action:nil];
   [backBarButtonItem setMenuHidden:config.disableBackButtonMenu];
+  // add accessibilityIdentifier to backBarButtonItem
+  [backBarButtonItem setAccessibilityLabel:@"go back"];
 
   if (config.isBackTitleVisible) {
+    [backBarButtonItem setAccessibilityLabel:@"go back"];
     if (config.backTitleFontFamily || config.backTitleFontSize) {
       NSMutableDictionary *attrs = [NSMutableDictionary new];
       NSNumber *size = config.backTitleFontSize ?: @17;
@@ -521,6 +524,7 @@ + (void)updateViewController:(UIViewController *)vc
     // of backButtonTitle
     [backBarButtonItem setTitle:nil];
     prevItem.backButtonTitle = resolvedBackTitle;
+    [backBarButtonItem setAccessibilityLabel:@"go back"];
   }
   prevItem.backBarButtonItem = backBarButtonItem;
 
@@ -579,6 +583,7 @@ + (void)updateViewController:(UIViewController *)vc
         navitem.leftItemsSupplementBackButton = config.backButtonInCustomView;
 #endif
         UIBarButtonItem *buttonItem = [[UIBarButtonItem alloc] initWithCustomView:subview];
+        [buttonItem setAccessibilityLabel:@"go back"];
         navitem.leftBarButtonItem = buttonItem;
         break;
       }

```

At first, I added `accessibilityIdentifier` instead of `accessibilityLabel`, but maestro does not pick `accessibilityIdentifier` if `accessibilityLabel` is set. So, I had to change it to `accessibilityLabel`.
I know this is a bad practice to override `accessibilityLabel`, but for our purpose we are fine screen readers read "go back". So, if someone wants they can use this patch for maestro tests to use same text for back button on every screen.

tboba:
Hi @khagesh, I was facing the issue of identifying back button, but I just needed that for pressing back button on iOS - I think that currently we don't need to set accessibility label for that 😄 If you'd like to press back button on iOS, I'd suggest to just tap on the `"Back"` or set the condition to do that. We're not planning to set the accessibility labels just to make our E2E tests work, but if this patch matches your requirements - keep it up! Patching it locally should be sufficient.

tboba:
Closing this PR, as currently Maestro doesn't meet our needs and is lacking support of running the tests on runners from Github Actions 🥲
I think that Maestro is just not ready for our needs for now, so let's come back to the case of moving our tests to Maestro in the next 2 years.

khagesh:
The problem here is that back button's accessibilityLabel is always different for every screen. And hence I wanted to add accessibilityIdentifier on back button. testID from react native transforms to accessibilityIdentifier on iOS. So, adding testID support is not just of rn-screen's own tests. It would help others as well. Also, the reason I patched with `accessibilityLabel` is because of maestro. It only picks accessibilityLabel if both (label and identifier) are set.

PR #1909, opened 2023-10-03
chore: use `require` syntax for resolution of all native components

kkafar:
## Description

Unification of way the native components are resolved in `index.native.tsx`.

## Changes

* Every native component is now resolved by `require(<component_spec_file>)` instead of directly querying view registry by using `requireNativeComponent`
* Ensured that every component spec is exported as default from its module

CI & launched native apps -- seems to be working fine. 

TODO: Check backward compatibility of this change

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/src/index.native.tsx
+++ b/src/index.native.tsx
@@ -5,7 +5,6 @@ import {
   Image,
   ImageProps,
   Platform,
-  requireNativeComponent,
   StyleProp,
   StyleSheet,
   UIManager,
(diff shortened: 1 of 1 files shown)


tboba (commented):
LGTM! Have you tested already if backward compatibility works fine?

tboba (approved):

WoLewicki (approved):

PR #1910, opened 2023-10-03
chore(CI): trigger e2e tests on JS-only changes

kkafar:
## Description

Changes in JS may impact components behaviour -> e2e tests should be triggered

## Changes

Added a trigger event to e2e CI spec

-

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/.github/workflows/android-e2e-test.yml
+++ b/.github/workflows/android-e2e-test.yml
@@ -8,6 +8,7 @@ on:
       - 'common/**'
       - 'Example/**'
       - 'TestsExample/**'
+      - 'src/**'
   push:
     branches:
       - main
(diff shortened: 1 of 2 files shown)


tboba (approved):
👍

tboba:
I've got one question, not strictly related to your changes - should we listen in CI to the changes from TestsExample, if it won't change the logic of the main code? If so, shouldn't we also listen to the changes from FabricTestExample?

kkafar:
> should we listen in CI to the changes from TestsExample

I peeked into blame & it was me who actually added `TestsExample` to the list of dependencies of these workflows. Scratching my head now, because I can't remember the reason & indeed it seems unnecessary as e2e tests use `Example` application solely. I'll cleanup this in this PR. 

Thanks for that catch 👍🏻

PR #1912, opened 2023-10-04
fix(iOS): select correct VC for nested modal presentation

### LINKED ISSUE #1829: When using nested, native-stack navigators, a modal screen in the inner navigator cannot push a modal screen in the outer navigator
### Description

https://github.com/software-mansion/react-native-screens/assets/5456182/f8b58a06-4cbd-458d-9e46-387a10b54b0b

I admit that this wording is a little confusing but let me try to explain what's going on. This seems to affect iOS only.

In the demo snack, I expect to be able to launch the "outer modal" -- the screen with name `modal` -- in the outer navigator from any other screen in this app. For example:

* I can navigate to `modal` from the starting screen `home`.
* I can navigate to `modal` from the inner navigator base screen, `base`.

However, I am unable to navigate to `modal` from `inner-modal`.

This is unexpected because if I were to remove the `presentation: modal` from `inner-modal`, then `modal` would pop-up normally. In addition, if `inner-modal` were in the same navigator as `modal`, then it would work, as evidenced by `modal` being able to navigate to `modal-2.

In addition, after this occurs, nothing is able to launch `modal` anymore, including the buttons which worked before!

### Steps to reproduce

1. Open the snack; iOS simulator
2. Open the Inner Navigator
3. Open the Inner Modal
4. Try to open the Outer Modal from the Inner Modal
5. Observe that it does not work.

https://snack.expo.dev/fnbHe-IiWz

### Screens version

3.18.0

### React Native version

0.70.0

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

None

### Architecture

Paper

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

kkafar:
## Description

Closes #1829

Currently it is not possible to navigate to a modal from another modal view that is mounted in JS under nested stack (see #1829 description for better context).
This is the case, because modal is presented by `RNSNavigationController` corresponding to the stack under which it is mounted in JS ==> UIKit reports that such controller is already presenting.

IMO the last presented modal should be the one, to present new one.

## Changes

Added code handling above case, by checking whether modals from other stacks are present & using top-most modal for presentation in such case.

See `Test1829` in test example apps

## Checklist

- [x] Ensured that CI passes (merged only comments after final CI checks)


tboba (approved):
LGTM! Great job 🚀

PR #1913, opened 2023-10-05
feat: add full screen swipe

WoLewicki:
## Summary

This pull request introduces a new feature which adds screen "go back" transition animations based on gestures. The feature is a collaboration between three libraries: `react-native-reanimated`, `react-native-screens`, and `react-native-gesture-handler`. Implementing this feature required changes in both Reanimated and RNScreens. You can find the related pull request in the `react-native-reanimated` repository: [PR 1913](https://github.com/software-mansion/react-native-reanimated/pull/5274).

Dependency requirements:
- `react-native-reanimated@3.6+`
- `react-native-screens@3.28+`
- `react-native-gesture-handler@2+`

A demo of this feature was presented at the RNCK meetup: [RNCK Meetup Demo](https://www.youtube.com/live/UYhOULSwjAg?si=YE55L6nw8z4FgZ9-&t=4523)

#### Basic usage

To use this feature, you need to import the `GestureDetectorProvider` from `react-native-screens/gesture-handler` and wrap your navigation container with it. Here's an example:

```js
import { GestureDetectorProvider } from 'react-native-screens/gesture-handler';

<GestureHandlerRootView>
  <NavigationContainer>
    <GestureDetectorProvider>
      <Stack.Navigator
        screenOptions={{
          stackAnimation: 'none',
        }}>
        <Stack.Screen
          name="ScreenA"
          component={ScreenA}
        />
        <Stack.Screen
          name="ScreenB"
          component={ScreenB}
          options={{
            goBackGesture: 'swipeRight', // gestures that trigger the screen transition
          }}
        />
      </Stack.Navigator>
    </GestureDetectorProvider>
  </NavigationContainer>
</GestureHandlerRootView>
```

Available gestures:
- swipeRight
- swipeLeft
- swipeUp
- swipeDown
- verticalSwipe
- horizontalSwipe
- twoDimensionalSwipe

#### Preset usage

You can also use preset animations for the transition. Import `ScreenTransition` from `react-native-reanimated` and specify the desired preset in the `transitionAnimation` option. Example:

```js
import { ScreenTransition } from 'react-native-reanimated';

<Stack.Screen
  name="ScreenB"
  component={ScreenB}
  options={{
    goBackGesture: 'swipeRight',
    transitionAnimation: ScreenTransition.SwipeRightFade, // transition preset
  }}
/>
```

Available presets:
- SwipeRight
- SwipeLeft
- SwipeDown
- SwipeUp
- Horizontal
- Vertical
- TwoDimensional
- SwipeRightFade

You don't need to worry about the preset because each gesture is associated with a default transition preset.

#### Custom Animation

If you want to create a custom animation, you can define your own `AnimatedScreenTransition` object. Here's an example:

```js
const customTransition: AnimatedScreenTransition = {
  topScreenFrame: (event, screenSize) => {
    'worklet';
    const progress = event.translationX / screenSize.width;
    return {
      transform: [
        { translateX: 1.3 * event.translationX },
        { rotate: 20 * progress + 'deg' }
      ]
    };
  },
  belowTopScreenFrame: (event, screenSize) => {
    'worklet';
    const progress = event.translationX / screenSize.width;
    return {
      transform: [
        { scale: 0.7 + 0.3 * progress }
      ]
    };
  },
}

<Stack.Screen
  name="ScreenB"
  component={ScreenB}
  options={{
    goBackGesture: 'swipeRight',
    transitionAnimation: customTransition
  }}
/>
```

#### `screenEdgeGesture`

To trigger a gesture that starts from the edge of the screen, you can use the `screenEdgeGesture` option. Example:

```js
<Stack.Screen
  name="ScreenB"
  component={ScreenB}
  options={{
    goBackGesture: 'swipeRight',
    screenEdgeGesture: true
  }}
/>
```

This means you can trigger a transition with a gesture that starts from the edge of the screen.

## Test plan

Test example from Example App


src/fabric/NativeScreensModule.ts
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable @typescript-eslint/ban-types */
+// its needed for codegen to work
+import type { TurboModule } from 'react-native';
+import { TurboModuleRegistry } from 'react-native';
+import type { Double, Int32 } from 'react-native/Libraries/Types/CodegenTypes';
+
+export interface Spec extends TurboModule {
```
tboba, line 7: Why is it TurboModule? I believe we don't have currently any TurboModules in screens. Cannot it be just a native module?

react-native-reanimated-3.6.0.tgz
tboba, line 1: I don't like the concept of putting this binary into the main project directory. Even if we'll probably remove this binary soon, can we move this `.tgz` file somewhere else (maybe into the new `bin` directory)?

src/native-stack/views/NativeStackView.tsx
```diff
@@ -403,19 +403,29 @@ function NativeStackViewInner({
 }: Props): JSX.Element {
   const { key, routes } = state;
 
+  const topScreenOptions = descriptors[state.routes[state.index].key].options;
```
tboba, line 406: IMO I would extract `state.routes[state.index].key` or `state.routes[state.index]` to the separate variable just for readability reasons

src/types.tsx
```diff
@@ -397,6 +397,8 @@ export interface ScreenStackProps extends ViewProps {
    * A callback that gets called when the current screen finishes its transition.
    */
   onFinishTransitioning?: (e: NativeSyntheticEvent<TargetedEvent>) => void;
+  stackRef?: React.MutableRefObject<React.Ref<View>>;
+  ref?: React.MutableRefObject<React.Ref<View>>;
```
tboba, line 401: Let's maybe put a comment above `stackRef` and `ref` what is the difference between them

src/fabric/NativeScreensModule.web.ts
```diff
@@ -0,0 +1,22 @@
+import type { Double, Int32 } from 'react-native/Libraries/Types/CodegenTypes';
+
+export default {
+  startTransition(_tag: Int32 | null) {
+    console.log(
+      '[ReactNativeScreens] startTransition is not supported on web]'
```
tboba, line 6: Let's stick to the one pattern:

```suggestion
      '[RNScreens] startTransition is not supported on web.'
```

src/fabric/NativeScreensModule.web.ts
```diff
@@ -0,0 +1,22 @@
...
+    );
+    return false;
+  },
+  updateTransition(_tag: Int32 | null, _progress: Double) {
+    console.log(
+      '[ReactNativeScreens] updateTransition is not supported on web]'
```
tboba, line 12:
```suggestion
      '[RNScreens] updateTransition is not supported on web.'
```

src/fabric/NativeScreensModule.web.ts
```diff
@@ -0,0 +1,22 @@
...
+    );
+    return false;
+  },
+  finishTransition(_tag: Int32 | null, _canceled: boolean) {
+    console.log(
+      '[ReactNativeScreens] finishTransition is not supported on web]'
```
tboba, line 18:
```suggestion
      '[RNScreens] finishTransition is not supported on web.'
```

src/native-stack/types.tsx
```diff
@@ -468,3 +473,52 @@ export type NativeStackDescriptor = Descriptor<
...
+    screenSize: MeasuredDimensions
+  ) => Record<string, unknown>;
+  belowTopScreenFrame: (
+    event: PanGestureHandlerEventPayload,
+    screenSize: MeasuredDimensions
+  ) => Record<string, unknown>;
+};
+
+export type GestureProviderProps = PropsWithChildren<{
+  children: React.ReactNode;
```
tboba, line 519: As in `GestureDetectorProvider` we don't provide any children, I would change this to the optional parameter (even just for the correct IDE suggestions 😄)

```suggestion
  children?: React.ReactNode;
```

tboba (commented):
Just a small comments from me. Overall, good job guys! I really appreciate your sacrifice and determination on completing this PR ❤️ You are rockin' it! 🎸

ios/RNSModule.mm
```diff
@@ -0,0 +1,106 @@
...
+@synthesize bridge = _bridge;
+
+- (dispatch_queue_t)methodQueue
+{
+  // It seems that due to how UIBlocks work with uiManager, we need to call the methods there
+  // for the blocks to be dispatched before the batch is completed
+  return dispatch_get_main_queue();
+}
+
+RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(startTransition : (nonnull NSNumber *)stackTag)
```
kkafar, line 25: We can make use of `NS_ASSUME_NONNULL_BEGIN` in this file too

package.json
```diff
@@ -98,7 +99,8 @@
     "react": "18.2.0",
     "react-dom": "^18.2.0",
     "react-native": "0.72.4",
-    "react-native-reanimated": "^2.2.0",
+    "react-native-gesture-handler": "^2.13.3",
+    "react-native-reanimated": "file:./react-native-reanimated-3.6.0.tgz",
```
kkafar, line 103: This must be changed

src/gesture-handler/GestureDetector.tsx
```diff
@@ -0,0 +1,302 @@
...
+      progress = Math.max(progressX, progressY);
+    }
+    sharedEvent.value = event;
+    const stackTag = screenTransitionConfig.value.stackTag;
+    global._manageScreenTransition(
+      ScreenTransitionCommand.Update,
+      stackTag,
+      progress
+    );
+  }
```
kkafar, line 227: oh my :D

kkafar (commented):
Lot of nitpicking, few sensible comments & questions.
I do not yet understand the iOS code, will come back to it some time soon

src/gesture-handler/GestureDetector.tsx
```diff
@@ -0,0 +1,302 @@
...
+  const screenTransitionConfig = makeMutable(
+    {
+      stackTag: -1,
+      belowTopScreenTag: -1,
+      topScreenTag: -1,
+      sharedEvent,
+      startingGesturePosition,
+      screenTransition: transitionAnimation,
+      isTransitionCanceled: false,
+      goBackGesture: goBackGesture ?? 'swipeRight',
```
WoLewicki, line 121: Isn't `goBackGesture` already checked before that it is valid?

src/gesture-handler/GestureDetector.tsx
```diff
@@ -0,0 +1,302 @@
...
+
+  function onStart(event: GestureUpdateEvent<PanGestureHandlerEventPayload>) {
+    'worklet';
+    sharedEvent.value = event;
+    const transitionConfig = screenTransitionConfig.value;
+    const animatedRef = refCurrent;
+    if (!animatedRef) {
+      throw new Error('[Reanimated] Unable to recognize stack ref.');
+    }
+    const stackTag = (animatedRef as () => number)();
```
WoLewicki, line 142: We need many more comments on what is going on in this code, or it will be impossible to maintain it in the future xd

src/gesture-handler/GestureDetector.tsx
```diff
@@ -0,0 +1,302 @@
...
+    startScreenTransition(transitionConfig);
+    canPerformUpdates.value = true;
+  }
+
+  function onUpdate(event: GestureUpdateEvent<PanGestureHandlerEventPayload>) {
+    'worklet';
+    if (!canPerformUpdates.value) {
+      return;
+    }
+    if (goBackGesture === 'swipeRight' && event.translationX < 0) {
```
WoLewicki, line 182: We need to change all those numbers etc into descriptive variables, or maybe create small functions like `resetNegativeTranslation` here?

src/gesture-handler/GestureDetector.tsx
```diff
@@ -0,0 +1,302 @@
...
+      event.translationX = 0;
+    } else if (goBackGesture === 'swipeLeft' && event.translationX > 0) {
+      event.translationX = 0;
+    } else if (goBackGesture === 'swipeDown' && event.translationY < 0) {
+      event.translationY = 0;
+    } else if (goBackGesture === 'swipeUp' && event.translationY > 0) {
+      event.translationY = 0;
+    }
+    let progress = 0;
+    if (goBackGesture === 'swipeRight') {
```
WoLewicki, line 192: `calculateProgress` ?

src/gesture-handler/GestureDetector.tsx
```diff
@@ -0,0 +1,302 @@
...
+    );
+  }
+
+  function onEnd(event: GestureUpdateEvent<PanGestureHandlerEventPayload>) {
+    'worklet';
+    if (!canPerformUpdates.value) {
+      return;
+    }
+    const screenSize = screenTransitionConfig.value.screenDimensions;
+    const distanceX = event.translationX + event.velocityX * 0.3;
```
WoLewicki, line 235: Link to where those values were taken from.

src/gesture-handler/GestureDetector.tsx
```diff
@@ -0,0 +1,302 @@
...
+    const distanceY = event.translationY + event.velocityY * 0.3;
+    const requiredXDistance = screenSize.width / 2;
+    const requiredYDistance = screenSize.height / 2;
+    let isTransitionCanceled = false;
```
WoLewicki, line 239: function `checkIfTransitionCancelled` or something similar.

src/gesture-handler/GestureDetector.tsx
```diff
@@ -0,0 +1,302 @@
...
+  let panGesture = Gesture.Pan()
+    .onStart(onStart)
+    .onUpdate(onUpdate)
+    .onEnd(onEnd);
+
+  if (screenEdgeGesture) {
+    const HIT_SLOP_SIZE = 50;
+    if (goBackGesture === 'swipeRight') {
+      panGesture = panGesture
+        .activeOffsetX(20)
```
WoLewicki, line 278: magic variable `20`.

WoLewicki (commented):
I added some comments, please answer them. Overall it looks pretty good :rocket:

ios/RNSModule.h
```diff
@@ -0,0 +1,9 @@
+#import <React/RCTBridgeModule.h>
+
+@interface RNSModule : NSObject <RCTBridgeModule>
+
+- (nonnull NSArray<NSNumber *> *)_startTransition:(nonnull NSNumber *)stackTag;
+- (bool)_updateTransition:(nonnull NSNumber *)stackTag progress:(double)progress;
+- (bool)_finishTransition:(nonnull NSNumber *)stackTag canceled:(bool)canceled;
+
+@end
```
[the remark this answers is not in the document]
kkafar: If they are not present you must specify nullability (`nullable` / `nonnull`) for each pointer separately. If you fail to do it the compiler will cry, throwing warnings or errors, depending on the options passed to the compiler.

src/index.tsx
```diff
@@ -67,6 +69,10 @@ export const InnerScreen = View;
 
 export const ScreenContext = React.createContext(Screen);
 
+export const GHContext = React.createContext(
+  (_props: GestureProviderProps): React.ReactElement => <></>
```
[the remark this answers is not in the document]
tboba: I don't think that in React.Fragment you could pass the props, sadly 🫤 still, as it's not a native index.tsx, I would leave it as it is right now (just for autosuggesting purposes).

android/build.gradle
```diff
@@ -29,6 +30,22 @@ def isNewArchitectureEnabled() {
     return project.hasProperty("newArchEnabled") && project.newArchEnabled == "true"
 }
 
+def shouldUseCommonInterfaceFromReanimated() {
+    if (project.hasProperty("useCommonInterfaceFromReanimated")) {
```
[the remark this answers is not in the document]
piaskowyk: This is just temporary workaround, we can remove it just after the release of Reanimated (with custom screen animation feature)
https://github.com/software-mansion/react-native-reanimated/blob/7513d1e876933bf52be7ac9d9bd7af382219b13b/Example/android/gradle.properties#L48

src/gesture-handler/GestureDetector.tsx
```diff
@@ -0,0 +1,302 @@
...
+      screenDimensions: DefaultScreenDimensions,
+      onFinishAnimation: () => {
+        'worklet';
+      },
+    },
+    true
+  );
+  const refCurrent = stackRef.current;
+  if (refCurrent === null) {
+    throw new Error('[RNScreens] You have to specify `stackRefWrapper`');
```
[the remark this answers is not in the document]
piaskowyk: You are right, this is unused old code. I need to remove it.

react-native-reanimated-3.6.0.tgz
piaskowyk: I can't see the purpose of moving it to a different location if we plan on removing it before merging. Let's leave it as it is for now, and we'll remove it completely in the near future.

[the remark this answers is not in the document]
piaskowyk: We will remove it just before the merge, due to this is required to development at this moment.

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
```diff
@@ -31,6 +31,9 @@ class ScreenStack(context: Context?) : ScreenContainer(context) {
     override val topScreen: Screen?
         get() = mTopScreen?.screen
 
+    val fragments: ArrayList<ScreenStackFragmentWrapper>
+        get() = mStack
+
```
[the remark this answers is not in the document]
piaskowyk: Here is the usage: https://github.com/software-mansion/react-native-screens/blob/fa10b21519d3fc695090a362d4cb57c1536d0437/android/src/main/java/com/swmansion/rnscreens/ScreensModule.kt#L51

android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt
```diff
@@ -200,6 +202,32 @@ open class ScreenContainer(context: Context?) : ViewGroup(context) {
...
+
+    fun detachTop() {
+        val top = topScreen as Screen
+        if (context is ReactContext) {
+            val surfaceId = UIManagerHelper.getSurfaceId(context)
+            UIManagerHelper
+                .getEventDispatcherForReactTag(context as ReactContext, top.id)
+                ?.dispatchEvent(ScreenDismissedEvent(surfaceId, top.id))
+        }
+    }
```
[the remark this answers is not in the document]
piaskowyk: cc @WoLewicki

android/build.gradle
```diff
@@ -29,6 +30,22 @@ def isNewArchitectureEnabled() {
...
+        return project.useCommonInterfaceFromReanimated == "true"
+    }
+    def reanimated = rootProject.subprojects.find { it.name == 'react-native-reanimated' }
+    if (reanimated != null) {
+        def inputFile = new File(reanimated.projectDir, '../package.json')
+        def json = new JsonSlurper().parseText(inputFile.text)
+        def reanimatedVersion = json.version as String
+        def (major, minor, patch) = reanimatedVersion.tokenize('.')
+        return Integer.parseInt(major) >= 3 && Integer.parseInt(minor) >= 6
```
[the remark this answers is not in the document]
piaskowyk: I removed it 🎉

android/src/main/java/com/swmansion/rnscreens/RNScreensPackage.kt
```diff
@@ -7,7 +7,9 @@ import com.facebook.react.uimanager.ViewManager
 
 class RNScreensPackage : ReactPackage {
     override fun createNativeModules(reactContext: ReactApplicationContext): List<NativeModule> =
-        emptyList()
+        listOf<NativeModule>(
```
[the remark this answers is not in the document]
piaskowyk: Done

src/gesture-handler/GestureDetector.tsx
piaskowyk: it is empirically determined :/

src/gesture-handler/GestureDetector.tsx
piaskowyk: I removed it 😅

ios/RNSModule.mm
```diff
@@ -0,0 +1,106 @@
...
+
+- (nonnull NSArray<NSNumber *> *)_startTransition:(nonnull NSNumber *)stackTag
+{
+  RCTAssertMainQueue();
+  RNSScreenStackView *stackView = [self getStackView:stackTag];
+  if (stackView == nil || isActiveTransition) {
+    return @[];
+  }
+  NSArray<NSNumber *> *screenTags = @[];
+  auto screens = stackView.reactViewController.childViewControllers;
```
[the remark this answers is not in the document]
piaskowyk: I don't see any :/

ios/RNSModule.mm
```diff
@@ -0,0 +1,106 @@
...
+  unsigned long screenCount = [screens count];
+  if (screenCount > 1) {
+    NSNumber *topScreen = screens[screenCount - 1].view.reactTag;
+    NSNumber *belowTopScreen = screens[screenCount - 2].view.reactTag;
+    screens[screenCount - 2].view.transform = CGAffineTransformMake(1, 0, 0, 1, 0, 0);
```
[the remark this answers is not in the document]
piaskowyk: I uses it to restore screen transform state. Without this restoration it will act like that:

https://github.com/software-mansion/react-native-screens/assets/36106620/072a1ea7-f5b9-48a8-8aa1-7a0c6ba8fd03

ios/RNSModule.mm
```diff
@@ -0,0 +1,106 @@
...
+    isActiveTransition = true;
+    screenTags = @[ topScreen, belowTopScreen ];
+  }
+  [stackView startScreenTransition];
```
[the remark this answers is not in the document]
piaskowyk: I moved it inside above condition-block

ios/RNSScreenStack.mm
```diff
@@ -960,6 +961,33 @@ - (void)didUpdateReactSubviews
...
+  if (_interactionController == nil) {
+    _customAnimation = YES;
+    _interactionController = [UIPercentDrivenInteractiveTransition new];
+    [_controller popViewControllerAnimated:YES];
+  }
+}
+
+- (void)updateScreenTransition:(double)progress
+{
+  [_interactionController updateInteractiveTransition:progress];
```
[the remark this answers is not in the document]
piaskowyk: Done 🫡

piaskowyk:
I don't know why CI with Android e2e is failing 🤔

src/gesture-handler/GestureDetector.tsx
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    transitionConfig.stackTag = stackTag.value;
+    startingGesturePosition.value = event;
+    const animatedRefMock = () => {
+      return screenTransitionConfig.value.topScreenId;
+    };
+    const screenSize = measure(animatedRefMock as any);
+    if (screenSize == null) {
+      throw new Error('[Reanimated] Failed to measure screen.');
```
WoLewicki, line 185:
```suggestion
      throw new Error('[RNScreens] Failed to measure screen.');
```

src/gesture-handler/GestureDetector.tsx
```diff
@@ -0,0 +1,346 @@
...
+    return isTransitionCanceled;
+  }
+
+  function onEnd(event: GestureUpdateEvent<PanGestureHandlerEventPayload>) {
+    'worklet';
+    if (!canPerformUpdates.value) {
+      return;
+    }
+    const screenSize = screenTransitionConfig.value.screenDimensions;
+    const distanceX = event.translationX + event.velocityX * 0.3;
```
WoLewicki, line 294: Kinda magic number `0.3`

WoLewicki (commented):
Left some comments :rocket:

WoLewicki (commented):
Left some comments :rocket:

tboba:
Some cases i've tested so far:

## 1) swipeUp / swipeDown gestures
When I was setting `goBackGesture` to either swipeUp or swipeDown I wasn't able to close the screen 😕 
Fortunately it's easier to dismiss it with verticalGesture 🎉 

https://github.com/software-mansion/react-native-screens/assets/23281839/3c609e9d-e228-4e1e-af49-544574df453a

## 2) (iOS) swipeLeft gesture
When I was using `swipeLeft` gesture I was still able to dismiss a screen from the left corner. Since I believe this is a native iOS setting to dismiss a view by swiping from the left border of a device (is it?), will it be possible to ignore it?

https://github.com/software-mansion/react-native-screens/assets/23281839/d3b20248-830b-41f6-b96a-b80e81a18222

## 3) (Android) Dismissing native swipe gesture
On Android on any gesture set in `goBackGesture` it's possible to horizontally go back to the other screen without any animation. Should we ignore this behavior (or perform classic pop animation during this gesture)? What do you think?

https://github.com/software-mansion/react-native-screens/assets/23281839/9adb29d6-f1fd-40ad-a99a-6ac74af0b1a7

## 3) You can't really override the existing animation with a previous gesture
Will it be possible to cancel the animation that is currently playing (by holding a finger onto the screen or fast swiping it)? I believe this might be frustrating for users to disallow cancelling the animation for them.

https://github.com/software-mansion/react-native-screens/assets/23281839/b3882c93-744b-426a-96f8-af36362f32b6

## 4) Error about RNSScreenStackView
I see that there's an error that occurs when the user (on debug mode) tries to change the `goBackGesture` value on a screen that is currently shown on the interface. 

On Android, doing the same thing crashes app completely 💥.

https://github.com/software-mansion/react-native-screens/assets/23281839/d807d8a3-9123-4943-b872-907d91e92fc9

FabricTestExample/src/TestScreenAnimation.tsx
```diff
@@ -0,0 +1,87 @@
...
+const Stack = createNativeStackNavigator<StackParamList>();
+
+const App = (): JSX.Element => (
+  <GestureHandlerRootView style={{ flex: 1 }}>
+      <NavigationContainer>
+    <GestureDetectorProvider>
+        <Stack.Navigator
+          screenOptions={{
+            headerHideBackButton: true,
+            direction: I18nManager.isRTL ? 'rtl' : 'ltr',
```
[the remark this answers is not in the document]
piaskowyk: No, I just copied it from another example 😅

android/src/main/cpp/jni-adapter.cpp
```diff
@@ -0,0 +1,99 @@
...
+                return {0, 0};
+            }
+            jclass javaClass = currentEnv->GetObjectClass(globalThis);
+            jmethodID methodID = currentEnv->GetMethodID(
+                javaClass,
+                "startTransition",
+                "(Ljava/lang/Integer;)[I"
+            );
+            jclass integerClass = currentEnv->FindClass("java/lang/Integer");
+            jmethodID integerConstructor = currentEnv->GetMethodID(integerClass, "<init>", "(I)V");
```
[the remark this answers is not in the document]
piaskowyk: but actually this is JNI 😅

android/src/main/java/com/swmansion/rnscreens/ScreensModule.kt
```diff
@@ -0,0 +1,96 @@
...
+
+    @DoNotStrip
+    private fun finishTransition(reactTag: Int?, canceled: Boolean) {
+        UiThreadUtil.assertOnUiThread()
+        if (!isActiveTransition.get() || reactTag == null) {
+            return
+        }
+        val uiManager = UIManagerHelper.getUIManagerForReactTag(mReactContext, reactTag)
+        val stack = uiManager?.resolveView(reactTag)
+        if (stack is ScreenStack) {
```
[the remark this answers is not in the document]
piaskowyk: Error log will be enough or you would rather prefer to throw an exception?

Example/package.json
```diff
@@ -23,7 +23,8 @@
     "nanoid": "^4.0.2",
     "react": "18.2.0",
     "react-native": "0.72.4",
-    "react-native-gesture-handler": "^2.12.1",
+    "react-native-gesture-handler": "^2.13.1",
+    "react-native-reanimated": "3.7.0-nightly-20231213-f3b16198a",
```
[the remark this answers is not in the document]
piaskowyk: I think this is not necessary. We don't want any dependencies when it comes to time of release both (Reanimated and Screens) libraries.

src/gesture-handler/GestureDetector.tsx
```diff
@@ -0,0 +1,346 @@
...
+  measure,
+  startScreenTransition,
+  finishScreenTransition,
+  ScreenTransition,
+  makeMutable,
+} from 'react-native-reanimated';
+import type { GestureProviderProps } from 'src/native-stack/types';
+import { getShadowNodeWrapperAndTagFromRef, isFabric } from './fabricUtils';
+
+declare global {
```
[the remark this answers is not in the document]
piaskowyk: We no longer need it, I'll remove it.

src/gesture-handler/GestureDetector.tsx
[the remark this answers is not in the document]
tboba: @kkafar can I ask you for your opinion about that change? I know that feeling of looking into the code may be subjective, so I need more voices about that 🙏

src/native-stack/types.tsx
```diff
@@ -468,3 +473,53 @@ export type NativeStackDescriptor = Descriptor<
...
+export type GoBackGesture =
+  | 'swipeRight'
+  | 'swipeLeft'
+  | 'swipeUp'
+  | 'swipeDown'
+  | 'verticalSwipe'
+  | 'horizontalSwipe'
+  | 'twoDimensionalSwipe';
+
+export type PanGestureHandlerEventPayload = {
```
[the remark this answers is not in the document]
piaskowyk: If you don't mind, I would like to stick with this name because it is the original name from GestureHandler. I suggest something like this:
image
Is it ok?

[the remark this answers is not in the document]
tboba: Yeah, sure!

src/gesture-handler/GestureDetector.tsx
```diff
@@ -0,0 +1,346 @@
...
+      goBackGesture: goBackGesture ?? 'swipeRight',
+      screenDimensions: DefaultScreenDimensions,
+      onFinishAnimation: () => {
+        'worklet';
+      },
+    },
+    true
+  );
+  const stackTag = makeMutable(-1, true);
+  const screenTagToNodeWrapperUI = makeMutable<any>({}, true);
```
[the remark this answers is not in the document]
piaskowyk: 1. `findNodeHandle` is a specific method in React Native. Its purpose is to find the native tag of a native component. This tag allows us to identify the native component in the native view registry.

2. `screenTagToNodeWrapperUI` is a container that is specific to Reanimated and Fabric. In this case, `screenTagToNodeWrapperUI` is created as a mutable value using `makeMutable`. This means it is a shared value that can be modified on the JS thread and read on the UI thread. In Fabric, I need access to the `shadowNodeWrapper` in order to modify the style of the component. However, this `shadowNodeWrapper` is only available on JS and after the first render. So in the `useEffect` hook, I retrieve the `shadowNodeWrapper` of every screen and send it to the UI Reanimated runtime via shared value. Reanimated requires this information to apply animation updates to the component.

src/gesture-handler/GestureDetector.tsx
```diff
@@ -0,0 +1,346 @@
...
+  GestureUpdateEvent,
+} from 'react-native-gesture-handler';
+import {
+  useSharedValue,
+  measure,
+  startScreenTransition,
+  finishScreenTransition,
+  ScreenTransition,
+  makeMutable,
+} from 'react-native-reanimated';
```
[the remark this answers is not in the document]
piaskowyk: I totally agree, but can we do it in another PR? This one is already big enough.

src/gesture-handler/GestureDetector.tsx
```diff
@@ -0,0 +1,346 @@
...
+    };
+    const screenSize = measure(animatedRefMock as any);
+    if (screenSize == null) {
+      throw new Error('[Reanimated] Failed to measure screen.');
+    }
+    if (screenSize == null) {
+      canPerformUpdates.value = false;
+      RNScreensTurboModule.finishTransition(stackTag.value, true);
+      return;
+    }
```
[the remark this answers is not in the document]
piaskowyk: The `RNScreensTurboModule#startTransition` method, contains logic that check if we can start transition, so we need to call it before :/

android/src/main/java/com/swmansion/rnscreens/ScreensModule.kt
```diff
@@ -0,0 +1,96 @@
...
+            if (canceled) {
+                stack.detachBelowTop()
+            } else {
+                stack.detachTop()
+            }
+            isActiveTransition.set(false)
+        }
+        topScreenId = -1
+    }
```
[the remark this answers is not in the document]
piaskowyk: Yes, you're right. I understand your point. Since this module is currently responsible for one specific logic domain, it makes sense to keep it as it is. If in the future we find a need to add more functionalities that are not strictly related to screen animation, we can consider splitting this module into sub-classes. I discussed it also with Wojtek. Are you ok with it?

ios/RNSModule.mm
```diff
@@ -0,0 +1,146 @@
...
+
+- (RNSScreenStackView *)getStackView:(NSNumber *)stackTag
+{
+  RNSScreenStackView *view = [self getScreenStackView:stackTag];
+  if (![view isKindOfClass:[RNSScreenStackView class]]) {
+    RCTLogError(@"Invalid view type, expecting RNSScreenStackView, got: %@", view);
+    return nil;
+  }
+  return view;
+}
```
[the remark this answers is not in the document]
piaskowyk: same thing - https://github.com/software-mansion/react-native-screens/pull/1913#discussion_r1435000166 😅

android/src/main/java/com/swmansion/rnscreens/RNScreensPackage.kt
```diff
@@ -18,4 +23,31 @@ class RNScreensPackage : ReactPackage {
...
+            val moduleInfos: MutableMap<String, ReactModuleInfo> = HashMap()
+            val isTurboModule = BuildConfig.IS_NEW_ARCHITECTURE_ENABLED
+            moduleInfos[ScreensModule.NAME] = ReactModuleInfo(
+                ScreensModule.NAME,
+                ScreensModule.NAME,
+                false, // canOverrideExistingModule
+                false, // needsEagerInit
+                true, // hasConstants
+                false, // isCxxModule
+                true // isTurboModule
```
kkafar, line 48: Shouldn't we use `isTurboModule` value here, instead of straight up passing `true`? It this is not the case `isTurboModule` is redundant.

src/gesture-handler/GestureDetector.tsx
```diff
@@ -0,0 +1,337 @@
...
+
+type RNScreensTurboModuleType = {
+  startTransition: (stackTag: number) => {
+    topScreenTag: number;
+    belowTopScreenTag: number;
+    canStartTransition: boolean;
+  };
+  updateTransition: (stackTag: number, progress: number) => void;
+  finishTransition: (stackTag: number, isCanceled: boolean) => void;
+};
```
kkafar, line 28: Why do you type it this way, instead of declaring these methods on the module spec itself in `src/fabric/NativeScreensModule.ts`?

^ treat this as a question for a meeting ;D: why do we not expose these module's methods in the spec, as you need to consume them in reanimated any way? 

My current understanding is that you do not expose anything in public API (by not declaring any methods in the TS module spec anyway) & you just access the module's methods that were injected into the runtime's global object during module initialisation.

Why is this preferred instead of just defining methods on module, implementing them & using them from JS through module itself, not global object?

src/gesture-handler/GestureDetector.tsx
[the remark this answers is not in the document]
kkafar: I would love some comments especially on types. I'll add comments on what is not clear  for me straight in the code.

src/gesture-handler/GestureDetector.tsx
```diff
@@ -0,0 +1,337 @@
...
+
+const DefaultScreenDimensions = {
+  width: 0,
+  height: 0,
+  x: 0,
+  y: 0,
+  pageX: 0,
+  pageY: 0,
+};
```
kkafar, line 76: what are `pageX, pageY`?

src/gesture-handler/fabricUtils.ts
```diff
@@ -0,0 +1,35 @@
+interface HostInstance {
+  _internalInstanceHandle: {
+    stateNode: {
+      node: Record<string, unknown>;
+    };
+  };
+  _nativeTag: number;
+}
+
+type localGlobal = typeof global & Record<string, unknown>;
```
kkafar, line 10:
```suggestion
type LocalGlobal = typeof global & Record<string, unknown>;
```

src/gesture-handler/fabricUtils.ts
```diff
@@ -0,0 +1,35 @@
...
+};
+if (isFabric()) {
+  try {
+    findHostInstance =
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      require('react-native/Libraries/Renderer/shims/ReactFabric').findHostInstance_DEPRECATED;
+  } catch (e) {
+    throw new Error('[RNScreens] Cannot import `findHostInstance_DEPRECATED`.');
+  }
+}
```
kkafar, line 27: Will this work once we migrate completely to ESM? How will this conditional `require` work then?

kkafar (commented):
Left some questions & invitation for screens meeting 😅

FabricTestExample/src/TestScreenAnimation.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { NavigationContainer } from '@react-navigation/native';
+import { View, StyleSheet, I18nManager, Button } from 'react-native';
+import { GestureHandlerRootView } from 'react-native-gesture-handler';
+import {
+  createNativeStackNavigator,
+  NativeStackNavigationProp,
+} from 'react-native-screens/native-stack';
+import { GestureDetectorProvider } from 'react-native-screens/gesture-handler';
+import RNScreensModule from '../../src/fabric/NativeScreensModule';
+console.log('RNScreensModule', RNScreensModule);
```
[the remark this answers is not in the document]
piaskowyk: The `RNScreensModule` shouldn't be a publicly accessible part of the API. I used it for testing purposes and forgot to remove it.

android/src/main/cpp/jni-adapter.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include <jsi/jsi.h>
+#include "RNScreensTurboModule.h"
+
+using namespace facebook;
+
+jobject globalThis;
+
+extern "C"
+JNIEXPORT void JNICALL
+Java_com_swmansion_rnscreens_ScreensModule_nativeInstall(JNIEnv *env, jobject thiz, jlong jsiPtr) {
```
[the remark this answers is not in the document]
piaskowyk: Sure thing, we'll meet on the call 🫡

package.json
```diff
@@ -137,7 +142,7 @@
   ],
   "codegenConfig": {
     "name": "rnscreens",
-    "type": "components",
+    "type": "all",
```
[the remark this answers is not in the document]
piaskowyk: Previously, the `rnscreens` library only contained native views. However, this PR adds a native module as well. In order to satisfy `codegen`, we now need to specify that the library contains both types of native dependencies. I think that @WoLewicki can provide more information on this matter.

src/gesture-handler/fabricUtils.ts
piaskowyk: Should work without any problems, we have same code in Reanimated (https://github.com/software-mansion/react-native-reanimated/blob/main/src/reanimated2/fabricUtils.ts) and it works.

src/gesture-handler/GestureDetector.tsx
piaskowyk: `pageX, pageY` - absolute X and Y position

This comes from Reanimated's measure - here is detailed description: https://docs.swmansion.com/react-native-reanimated/docs/advanced/measure/#returns

src/gesture-handler/fabricUtils.ts
kkafar: AFAIK this won't emit proper ESM. @tboba you did some research on this topic with #1982, please take a look

tboba: @kkafar Yup, as I checked previously how `lib` directory is being transpiled, this `require` declaration is being left as it is in ESM source code. As this could bring errors in some bundlers about mixed ESM/CJS code, I would consider if we should import specific packages like that.

kkafar: So, I understand that if we merge this in current form, #1982 is basically useless and ESM/CJS still will be mixed leading to errors with some bundlers, right?

tboba: Most likely, but let me check that. I'm planning to test PR #1982 today with Vite bundler to check if this issue still appears - I'll also test this PR with this `require`.

kkafar: Thanks!

tboba: @kkafar Unfortunately as I quickly checked how `lib` is being generated, all of the `require` statements are unfortunately not transpiled into `import` ones. I think we should find another way how we can import such optional dependencies (maybe with `await import`?) since some bundlers (like Vite) will still complain about mixed CJS/ESM code. 
However, I still can't manage to test that behavior - I've made a small example with @react-navigation/native-stack, Vite and react-native-web and I'm not getting any errors about the code 🤔 I think that I'm doing possibly something wrong, so I asked about the repro in the issue.

kkafar: I guess this file: `react-native/Libraries/Renderer/shims/ReactFabric` is always there, however I can see loading Fabric renderer on Paper leading to some issues. We need to check that & if there are no issues I'm fully in favour of importing it standard way.

tboba: @WoLewicki the biggest problem is that `ReactFabric` only exists... on Fabric 😄 
I think that inline import would be possible here if this file could also exist on Paper - `if(isFabric())` is not without a reason there. But, as import has to be resolved from the global scope (it can't be imported after if or in any class/function, unless we are talking about the `await import`) it's not that easy though 😕

src/gesture-handler/GestureDetector.tsx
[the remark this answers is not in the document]
tboba: Fair, thanks for the explanation!

src/gesture-handler/GestureDetector.tsx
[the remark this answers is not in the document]
tboba: Yeah sure 👍

anisurrahman072:
If anyone face **"Reanimated"** issue then it may solve your problem 👇

- https://github.com/software-mansion/react-native-screens/issues/2103#issuecomment-2052601353

PR #1915, opened 2023-10-07
fix: Allow Reanimated Screen to check large header

tboba:
## Description

It looks that on Reanimated Screen we didn't check if on iOS there's large header enabled, which led to the wrong initial values, when user was using *at least* ReanimatedScreenProvider.

This feature fixes that by adding an ability to get screen descriptor by Reanimated Screen, which contains an information if user has enabled `headerLargeTitle` prop.

## Changes

- Changed the `useReanimatedTransitionProgress` leftover inside the `useReanimatedHeaderHeight` file
- Created new hook `useScreenInfo` that provides an information about the route view which is being used by ReanimatedNativeStackScreen (this hook is not meant to be public [at least for now] - that's why I didn't exported it)
- Moved the implementation of getStatusBarHeight to the separate file

### Before

image

### After

image

Add to `Test42` ReanimatedScreenProvider and enable `headerLargeTitle` prop for the `Home` screen.

## Checklist

- [X] Updated TS types
- [ ] Ensured that CI passes


src/native-stack/views/NativeStackView.tsx
```diff
@@ -264,6 +264,7 @@ const RouteView = ({
       key={route.key}
       enabled
       isNativeStack
+      screenDescriptor={descriptors[route.key]}
```
tboba, line 267: Just a small comment from me about this line: I'm creating `screenDescriptor` prop here to access the options for `HeaderConfig` that is not available from the `ScreenProps` prop the `ReanimatedNativeStackScreen` gets, but since we're providing such information - maybe we could provide all of the information to the screen via some sort of hook, like `useScreenProperties`? 

We could provide such information like an info about the route, screen index and descriptors, just as the user is receiving the information about the navigation from `useNavigation` hook. I could also benefit from it by using that hook for receiving an info about the descriptors of the screen.
What do you think?

tboba: Resolved it in https://github.com/software-mansion/react-native-screens/pull/1915/commits/28a4d92f3bd536b93de1a62beb42e091aa00d49a - as we don't want to push such a big object to the native side, I've replaced it to the hook.

src/native-stack/utils/ScreenInfoContext.tsx
```diff
@@ -0,0 +1,12 @@
...
+import { NativeStackNavigationOptions } from '../types';
+
+type ScreenInfoType = {
+  options: NativeStackNavigationOptions;
+};
+
+const ScreenInfoContext = React.createContext<ScreenInfoType>({
+  options: {},
+});
+
```
kkafar, line 11: Nitpick: If we're gonna go with context let's work on naming. I don't have any suggestion for now, unfortunately, but will work on it 🤞🏻

src/native-stack/utils/useScreenInfo.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+
+import ScreenInfoContext from './ScreenInfoContext';
+
+export default function useScreenInfo() {
+  const screenInfo = React.useContext(ScreenInfoContext);
+
+  if (screenInfo === undefined) {
+    throw new Error(
+      "Couldn't find information about the screen. Are you inside a screen in a navigator?"
```
kkafar, line 10:
```suggestion
      "Couldn't find information about the screen. Make sure that the hook is being used inside screen contained in navigator"
```

src/native-stack/views/NativeStackView.tsx
```diff
@@ -260,128 +261,130 @@ const RouteView = ({
...
-        <HeaderHeightContext.Provider value={staticHeaderHeight}>
-          <MaybeNestedStack
-            options={options}
-            route={route}
-            stackPresentation={stackPresentation}>
-            {renderScene()}
-          </MaybeNestedStack>
-          {/* HeaderConfig must not be first child of a Screen.
+    <ScreenInfoContext.Provider value={{ options }}>
+      <Screen
```
kkafar, line 265: Wrapping the screen inside context here will work only with navigation v5. v6 (hosted in `react-navigation` repo) uses its own `NativeStackView` & won't be wrapped inside such context, thus I'm postulating following action points:

1. We need to make sure that these changes are ported for native stack v6
2. We need to make sure that there won't be any crashes in case the context is not present

src/reanimated/ReanimatedNativeStackScreen.tsx
```diff
@@ -30,24 +32,29 @@ const ReanimatedNativeStackScreen = React.forwardRef<
...
-    statusBarHeight = 54;
-  }
+  const isStatusBarTranslucent = rest.statusBarTranslucent ?? false;
+  const statusBarHeight = getStatusBarHeight(
+    topInset,
+    dimensions,
+    isStatusBarTranslucent
+  );
+
+  const isLargeHeader = options.headerLargeTitle ?? false;
```
kkafar, line 49: Why do we apply this fix only for `ReaniamtedNativeStackScreen`? Where does the difference come from?

src/reanimated/useReanimatedHeaderHeight.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import ReanimatedHeaderHeightContext from './ReanimatedHeaderHeightContext';
 
-export default function useReanimatedTransitionProgress() {
+export default function useReanimatedHeaderHeight() {
```
kkafar, line 4: I'm not sure right now, but if it was part of public API we should not rename it like that. If it was not, it is ok then.

tboba: It wasn't yet, as useReanimatedHeaderHeight wasn't in any release and it was my mistake that I didn't renamed the export and function in this hook 😅

kkafar: I see, the change is fine then

src/reanimated/ReanimatedNativeStackScreen.tsx
tboba: This is applied just for the compliance between the NativeStackView and ReanimatedNativeStackScreen - I forgot to apply it here earlier 😅 Take a look on the [NativeStackView#L225](https://github.com/software-mansion/react-native-screens/blob/24689052c009f383657a74521c5ce875044ee2ef/src/native-stack/views/NativeStackView.tsx#L225).

src/native-stack/utils/ScreenInfoContext.tsx
kkafar: Update & remainder for me: We talked and I've suggested to pass this information alongside `enabled`, `isNativeStack` as a prop instead of using context.

src/native-stack/views/NativeStackView.tsx
tboba: Sure! Resolved in https://github.com/software-mansion/react-native-screens/pull/1915/commits/b2d876ac20a3972688451fdf3f9b62c12086f6c2.

src/native-stack/utils/ScreenInfoContext.tsx
tboba: Done! https://github.com/software-mansion/react-native-screens/pull/1915/commits/b2d876ac20a3972688451fdf3f9b62c12086f6c2

src/native-stack/utils/useScreenInfo.tsx
tboba: Since this hook has been removed, I'm resolving it. Context: https://github.com/software-mansion/react-native-screens/pull/1915/commits/b2d876ac20a3972688451fdf3f9b62c12086f6c2.

src/native-stack/views/NativeStackView.tsx
```diff
@@ -130,7 +130,11 @@ const MaybeNestedStack = ({
   if (isHeaderInModal) {
     return (
       <ScreenStack style={styles.container}>
-        <Screen enabled isNativeStack style={StyleSheet.absoluteFill}>
+        <Screen
+          enabled
+          isNativeStack
+          isLargeHeader={isLargeHeader}
```
kkafar, line 136: Nitpicking but `isLargeHeader` does not fit here for me. As the `Screen` is not a header, but it may have one, thus `hasLargeHeader` maybe? 

```suggestion
          isLargeHeader={isLargeHeader}
```

kkafar (commented):
Besides one note below it looks good. GJ 🎉 

However I'm still doubtful that this is needed. Why only `ReanimatedNativeStackScreen` requires this additional information? What happens to e.g. `InnerScreen`?

src/native-stack/views/NativeStackView.tsx
tboba: Sure!

tboba: Done in https://github.com/software-mansion/react-native-screens/pull/1915/commits/cf7dcb1b684a51d8f0799a3cc49342a676b6e658.

PR #1917, opened 2023-10-11
fix(iOS): change implementation of calculating status bar, refactor methods used on header height change

tboba:
## Description

It looks that sometimes when you're most likely on the first screen the initial header height (taken from the `getDefaultHeaderHeight` method) stays and is not being updated from the `onHeaderHeightChange` event. Also, when user hides the status bar it wasn't counted to the final value of header height. This PR fixes those problems and also other minor issues, related to the calculating header height.

## Changes

- Added calls for calculating header height on setting animated config and changing `statusBarHidden` prop.
- Changed implementation of `getCalculatedStatusBarHeightIsModal` method.
- Refactorized naming of the methods, related to the header height.
- Added asserting modal hierarchy.

You can change `Modals.tsx` file by adding this snippet:

```js
    const headerHeight = useAnimatedHeaderHeight();
    headerHeight.addListener((height) => console.log(height.value))
```

to the components and listen to the changes in `Modals` example. Then try to hide header in modals - there should be `0` value as a header height.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -1040,12 +1040,9 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
 
 - (BOOL)assertModalHierarchy
 {
-  if (self.childViewControllers.count == 0) {
+  if (self.childViewControllers.count > 1) {
     RCTLogError(
-        @"Modal has been rendered without the navigation controller, which is not allowed. Ensure that modal has RNSNavigationController mounted as a child view controller.");
-  } else if (self.childViewControllers.count > 1) {
-    RCTLogError(
-        @"Modal has been rendered with more than one navigation controller, which is not allowed. Ensure that modal has only one RNSNavigationController mounted as a child view controller.");
+        @"Modal has been rendered with more than one navigation controller, which is not allowed. Ensure that modal has none or only one RNSNavigationController mounted as a child view controller.");
   }
 
   return self.childViewControllers.count == 1 &&
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -1074,20 +1074,8 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
     return CGSizeMake(0, 0);
   }
 
-  CGSize fallbackStatusBarSize = [[UIApplication sharedApplication] statusBarFrame].size;
-
-#if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && defined(__IPHONE_13_0) && \
-    __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_0
-  if (@available(iOS 13.0, *)) {
-    CGSize primaryStatusBarSize = self.view.window.windowScene.statusBarManager.statusBarFrame.size;
-    if (primaryStatusBarSize.height == 0 || primaryStatusBarSize.width == 0)
-      return fallbackStatusBarSize;
-
-    return primaryStatusBarSize;
-  } else {
-    return fallbackStatusBarSize;
-  }
-#endif /* Check for iOS 13.0 */
+  CGFloat statusBarSize = self.navigationController.navigationBar.frame.origin.y;
+  return CGSizeMake(statusBarSize, statusBarSize);
 
 #else
   // On TVOS, status bar doesn't exist
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -1058,40 +1058,42 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
   return NO;
 }
 
-- (CGFloat)getNavigationBarHeightIsModal:(BOOL)isModal
+- (UINavigationController *)getVisibleNavigationControllerIsModal:(BOOL)isModal
 {
   UINavigationController *navctr = self.navigationController;
 
-  // In case where screen is a modal, we want to calculate childViewController's
-  // navigation bar height instead of the navigation controller from RNSScreen.
-  if (isModal && self.assertModalHierarchy) {
-    navctr = self.childViewControllers[0];
+  if (isModal) {
+    // In case where screen is a modal, we want to calculate childViewController's
+    // navigation bar height instead of the navigation controller from RNSScreen.
+    if (self.assertModalHierarchy) {
+      navctr = self.childViewControllers[0];
+    } else {
+      // If the modal does not meet requirements (there's no RNSNavigationController which means that probably it
+      // doesn't have header or there are more than one RNSNavigationController which is invalid) we don't want to
+      // return anything.
+      return nil;
+    }
   }
 
-  return navctr.navigationBar.frame.size.height;
-}
-
-- (CGFloat)getNavigationBarInsetIsModal:(BOOL)isModal
-{
-#if TARGET_OS_TV
-  // On TVOS there's no inset of navigation bar.
-  return 0;
-#endif // TARGET_OS_TV
-  UINavigationController *navctr = self.navigationController;
-
-  // In case where screen is a modal, we want to calculate childViewController's
-  // navigation bar inset instead of the navigation controller from RNSScreen.
-  if (isModal && self.assertModalHierarchy) {
-    navctr = self.childViewControllers[0];
-  }
-
-  return navctr.navigationBar.frame.origin.y;
+  return navctr;
 }
 
 - (CGFloat)calculateHeaderHeightIsModal:(BOOL)isModal
 {
-  CGFloat navbarHeight = [self getNavigationBarHeightIsModal:isModal];
-  CGFloat navbarInset = [self getNavigationBarInsetIsModal:isModal];
+  UINavigationController *navctr = [self getVisibleNavigationControllerIsModal:isModal];
+
+  if (navctr == nil || navctr.navigationBarHidden) {
+    return 0;
+  }
+
+  CGFloat navbarHeight = navctr.navigationBar.frame.size.height;
+#if !TARGET_OS_TV
+  CGFloat navbarInset = navctr.navigationBar.frame.origin.y;
+#else
+  // On TVOS there's no inset of navigation bar.
+  CGFloat navbarInset = 0;
+#endif // !TARGET_OS_TV
+
   return navbarHeight + navbarInset;
 }
 
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -1038,14 +1038,9 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
   }
 }
 
-- (BOOL)assertModalHierarchy
+- (BOOL)isModalWithHeader
 {
-  if (self.childViewControllers.count > 1) {
-    RCTLogError(
-        @"Modal has been rendered with more than one navigation controller, which is not allowed. Ensure that modal has none or only one RNSNavigationController mounted as a child view controller.");
-  }
-
-  return self.childViewControllers.count == 1 &&
+  return self.screenView.isModal && self.childViewControllers.count == 1 &&
       [self.childViewControllers[0] isKindOfClass:UINavigationController.class];
 }
 
@@ -1093,7 +1088,7 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
   if (isModal) {
     // In case where screen is a modal, we want to calculate childViewController's
     // navigation bar height instead of the navigation controller from RNSScreen.
-    if (self.assertModalHierarchy) {
+    if (self.isModalWithHeader) {
       navctr = self.childViewControllers[0];
     } else {
       // If the modal does not meet requirements (there's no RNSNavigationController which means that probably it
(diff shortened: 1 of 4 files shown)


ios/RNSScreen.mm
```diff
@@ -251,6 +251,9 @@ - (void)setStatusBarHidden:(BOOL)statusBarHidden
   _statusBarHidden = statusBarHidden;
   [RNSScreenWindowTraits assertViewControllerBasedStatusBarAppearenceSet];
   [RNSScreenWindowTraits updateStatusBarAppearance];
+  // As the status bar could change its visibility, we need to calculate header height
+  // for the correct value in `onHeaderHeightChange` event.
+  [self.controller calculateAndNotifyHeaderHeightChangeIsModal:NO];
```
kkafar, line 256: Hey, I'm not convinced RN that hiding the status bar actually changes the height of a header. How does that work? Does it work like I think: hiding status bar has no effect on header height visually, but actual header height increases by height of the status bar?

kkafar (commented):
I have left a question below

ios/RNSScreen.mm
tboba: Youre right! 😄 this has been fixed in https://github.com/software-mansion/react-native-screens/pull/1917/commits/6e3cbb9b3a306a6bdb039282b8300ddf922efdbe and calling the method has been removed in https://github.com/software-mansion/react-native-screens/pull/1917/commits/46364e59f2d4b4f23186b7f519178d0c495a1955.

ios/RNSScreen.mm
```diff
@@ -1058,50 +1058,32 @@ - (CGFloat)getCalculatedHeaderHeightIsModal:(BOOL)isModal
...
-  // We should check if user has rotated its screen, so we're choosing the minimum value between the
-  // width and height.
-  CGFloat statusBarHeight = MIN(statusBarSize.width, statusBarSize.height);
-  return navbarHeight + statusBarHeight;
+  CGFloat navbarHeight = [self getNavigationBarHeightIsModal:isModal];
+  CGFloat navbarInset = [self getNavigationBarInsetIsModal:isModal];
+  return navbarHeight + navbarInset;
 }
 
 - (void)calculateAndNotifyHeaderHeightChangeIsModal:(BOOL)isModal
```
tboba, line 1089: Just one comment from me - leaving the name `calculateHeaderHeightIsModal` is made on purpose here. For me `header` is the navigationBar + an inset of it - that's why I'm leaving it here.

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -147,6 +147,11 @@ - (void)updateViewControllerIfNeeded
   // if nav is nil, it means we can be in a fullScreen modal, so there is no nextVC, but we still want to update
   if (vc != nil && (nextVC == vc || isInFullScreenModal || isPresentingVC)) {
     [RNSScreenStackHeaderConfig updateViewController:self.screenView.controller withConfig:self animated:YES];
+    // As the header might have change in `updateViewController` we need to ensure that header height
+    // returned by the `onHeaderHeightChange` event is correct.
+    if ([vc isKindOfClass:[RNSScreen class]]) {
```
kkafar, line 152: I believe this check is redundant as `screenView.controller` is always of type `RNSScreen`.

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -147,6 +147,11 @@ - (void)updateViewControllerIfNeeded
...
+      [(RNSScreen *)self.screenView.controller calculateAndNotifyHeaderHeightChangeIsModal:NO];
```
kkafar, line 153:
```suggestion
      [self.screenView.controller calculateAndNotifyHeaderHeightChangeIsModal:NO];
```

This cast should not be necessary as `controller` property on screen returns appropriate type.

ios/RNSScreen.mm
```diff
@@ -1058,50 +1058,32 @@ - (CGFloat)getCalculatedHeaderHeightIsModal:(BOOL)isModal
...
-  CGSize statusBarSize = [self getCalculatedStatusBarHeightIsModal:isModal];
-
-  // Unfortunately, UIKit doesn't care about switching width and height options on screen rotation.
-  // We should check if user has rotated its screen, so we're choosing the minimum value between the
-  // width and height.
-  CGFloat statusBarHeight = MIN(statusBarSize.width, statusBarSize.height);
-  return navbarHeight + statusBarHeight;
+  CGFloat navbarHeight = [self getNavigationBarHeightIsModal:isModal];
+  CGFloat navbarInset = [self getNavigationBarInsetIsModal:isModal];
+  return navbarHeight + navbarInset;
```
kkafar, line 1086: How does this new method work with nested elements / modals? Is it good?

kkafar (commented):
Few more remarks. 

Also please update PR title before merging (as events are no longer sent on status bar visibility change)

ios/RNSScreenStackHeaderConfig.mm
tboba: In this case we cannot look for screenView, as this method is static 😅 Do you mean that we don't need to ensure if vc is RNSScreen?

ios/RNSScreen.mm
tboba: I've checked previously and it looks gooood 👍 As nested views does not contain initial inset this hook returns only the height of navigation bar - 44 for small headers. If there's not any nested views the hook works as usual. On modals there's also correct value - 56. Also it looks that the `if` statement for fullscreen modal is also unnecessary right now.

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -353,6 +358,11 @@ + (void)willShowViewController:(UIViewController *)vc
                     withConfig:(RNSScreenStackHeaderConfig *)config
 {
   [self updateViewController:vc withConfig:config animated:animated];
+  // As the header might have change in `updateViewController` we need to ensure that header height
+  // returned by the `onHeaderHeightChange` event is correct.
+  if ([vc isKindOfClass:[RNSScreen class]]) {
+    [(RNSScreen *)vc calculateAndNotifyHeaderHeightChangeIsModal:NO];
+  }
```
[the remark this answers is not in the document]
tboba: Let's keep it for now - I'm afraid if there will be some cases that `vc` won't be actually RNSScreen (RNSScreenStack?) and that may cause some crashes on iOS.

ios/RNSScreen.mm
tboba: Changed that in https://github.com/software-mansion/react-native-screens/pull/1917/commits/8884bef22a406281739146dce0ecb4985c482348.

ios/RNSScreenStackHeaderConfig.mm
tboba: Changed that in https://github.com/software-mansion/react-native-screens/pull/1917/commits/8884bef22a406281739146dce0ecb4985c482348.

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -147,6 +147,9 @@ - (void)updateViewControllerIfNeeded
   // if nav is nil, it means we can be in a fullScreen modal, so there is no nextVC, but we still want to update
   if (vc != nil && (nextVC == vc || isInFullScreenModal || isPresentingVC)) {
     [RNSScreenStackHeaderConfig updateViewController:self.screenView.controller withConfig:self animated:YES];
+    // As the header might have change in `updateViewController` we need to ensure that header height
+    // returned by the `onHeaderHeightChange` event is correct.
+    [(RNSScreen *)self.screenView.controller calculateAndNotifyHeaderHeightChangeIsModal:NO];
```
kkafar, line 152: The cast is still here

```suggestion
    [self.screenView.controller calculateAndNotifyHeaderHeightChangeIsModal:NO];
```

I believe it is redundant

kkafar (commented):
Looks good beside one thing I mention below

ios/RNSScreenStackHeaderConfig.mm
tboba: Yeah, I've left it on mistake, as I was in a rush :sweat_smile: Fixed in https://github.com/software-mansion/react-native-screens/pull/1917/commits/40df2fdab3cfd753dc7263246515ca6d449e0321.

kkafar (approved):
LGTM :tada:

tboba:
Hey @kkafar, as I've made some changes (as I've discovered a bug while testing modals 😅) and also refactorized even more some places, can you take a final look on them? 🙏

kkafar (commented):
Hey, I've found the use case I told you about yesterday: if `headerShown: false` is set on modal it won't have `RNSNavigationController` in its hierarchy.

image

Does the code cover this case? 
I believe 0 should be returned.

kkafar:
Also after this PR is approved, please update PR description so that it reflects what actually this PR does.

ios/RNSScreen.mm
```diff
@@ -1031,6 +1038,20 @@ - (void)viewDidLayoutSubviews
...
+        @"Modal has been rendered without the navigation controller, which is not allowed. Ensure that modal has RNSNavigationController mounted as a child view controller.");
+  } else if (self.childViewControllers.count > 1) {
+    RCTLogError(
+        @"Modal has been rendered with more than one navigation controller, which is not allowed. Ensure that modal has only one RNSNavigationController mounted as a child view controller.");
+  }
+
+  return self.childViewControllers.count == 1 &&
+      [self.childViewControllers[0] isKindOfClass:UINavigationController.class];
+}
+
```
kkafar, line 1054: Hey, as I've shown in last review & as we discussed it is possible that "modal" view controller hierarchy is comprised of only single `RNSScreen` (in case it has no header).

Thus these log messages might be misleading & incorrect.

ios/RNSScreen.mm
```diff
@@ -1087,21 +1084,56 @@ - (CGSize)getCalculatedStatusBarHeightIsModal:(BOOL)isModal
...
+  // If navigation controller doesn't exists (or it is hidden) we want to handle two possible cases.
+  // If there's no navigation controller for the modal, we simply don't want to return header height, as modal possibly
+  // does not have header and we don't want to count status bar. If there's no navigation controller for the view we
+  // just want to return status bar height (if it's hidden, it will simply return 0).
+  if (navctr == nil || navctr.isNavigationBarHidden) {
+    if (isModal)
+      return 0;
+    else
+      return MIN(statusBarSize.width, statusBarSize.height);
+  }
```
kkafar, line 1126:
```suggestion
  UINavigationController *navctr = [self getVisibleNavigationControllerIsModal:isModal];

  // If navigation controller doesn't exists (or it is hidden) we want to handle two possible cases.
  // If there's no navigation controller for the modal, we simply don't want to return header height, as modal possibly
  // does not have header and we don't want to count status bar. If there's no navigation controller for the view we
  // just want to return status bar height (if it's hidden, it will simply return 0).
  if (navctr == nil || navctr.isNavigationBarHidden) {
    if (isModal)
      return 0;
    else {
      CGSize statusBarSize = [self getStatusBarHeightIsModal:isModal];
      return MIN(statusBarSize.width, statusBarSize.height);
    }
  }
```

ios/RNSScreen.mm
```diff
@@ -1031,6 +1038,20 @@ - (void)viewDidLayoutSubviews
   }
 }
 
+- (BOOL)assertModalHierarchy
```
kkafar, line 1041: Assertion should throw error / panic on failure, we just print the error.

```suggestion
- (BOOL)isModalWithHeader
```

I remember we talked about this method, sorry for misleading suggestion back then.

kkafar: My suggestion is to join two checks in this method:

```objective-c
return self.isModal && self.childViewControllers.count == 1 && [self.childViewControllers[0] isKindOfClass:UINavigationController.class]
```

ios/RNSScreen.mm
```diff
@@ -251,6 +251,13 @@ - (void)setStatusBarHidden:(BOOL)statusBarHidden
   _statusBarHidden = statusBarHidden;
   [RNSScreenWindowTraits assertViewControllerBasedStatusBarAppearenceSet];
   [RNSScreenWindowTraits updateStatusBarAppearance];
+
+  // As the status bar could change its visibility, we need to calculate header
+  // height for the correct value in `onHeaderHeightChange` event when navigation
+  // bar is not visible.
+  if (self.controller.navigationController.navigationBarHidden && !self.isModal) {
+    [self.controller calculateAndNotifyHeaderHeightChangeIsModal:NO];
+  }
```
kkafar, line 260: I do not really get why we need to recalculate header height on status bar visibility change? Is it because we use status bar height in case: non-modal view && navigation controller is `nil` or navigation bar is hidden? If so, what happens if the status bar is hidden? Are correct values reported?

kkafar (commented):
Few more remarks. Getting close

ios/RNSScreen.mm
tboba: Fixed in https://github.com/software-mansion/react-native-screens/pull/1917/commits/e779764a5423b326af1d71e2cbc3da4cfd5ca201.

ios/RNSScreen.mm
tboba: Resolved in https://github.com/software-mansion/react-native-screens/pull/1917/commits/add6ee3ea469736574a26cece43b2dbfcb714437.

ios/RNSScreen.mm
tboba: Resolved in https://github.com/software-mansion/react-native-screens/pull/1917/commits/add6ee3ea469736574a26cece43b2dbfcb714437.

ios/RNSScreen.mm
tboba: That's because `viewDidLayoutSubviews` is not being called when the status bar visibility is being changed 😃 
Also values seems to be correct there.

kkafar (approved):
🟢

Issue #1515, opened 2022-07-06
RNScreens 3.14.1: Unresolved reference: attr.


kinuxroot:
## Description

Build app and the error occurs in build react-native-screens

### Screenshots
![图片](https://user-images.githubusercontent.com/6010996/177526458-62c6e04d-3906-4421-933a-c77409580e77.png)

## Steps To Reproduce

1.Clone this project [code repository](https://github.com/kinuxroot/react-native-screens-with-bugs)
2.Install node modules using yarn
3.Enter ./android directory and use gradle to build app

### Expected behavior

compiling success

### Actual behavior

compiling failed

<details>

<summary>gradlew build --stacktrace</summary>

> Task :react-native-safe-area-context:compileDebugKotlin
w: D:\Projects\DataEE\taoxia\taoxia-app\taoxia-app\node_modules\react-native-safe-area-context\android\src\main\java\com\th3rdwave\safeareacontext\SafeAreaView.kt: (50, 23): 'getter for uiImplementation: UIImplementation!' is deprecated. Deprecated in Java

> Task :app:stripDebugDebugSymbols
[...]

PR #1919, opened 2023-10-14
fix: Unresolved reference: attr after upgrading to 0.73.0-rc.2

ivanignatiev:
## Description

Small PR which probably resolves [#1515](https://github.com/software-mansion/react-native-screens/issues/1515). 

For my case, I got this build issue after upgrading from 0.72.5 to 0.73.0-rc.2:

```
> Task :react-native-screens:compileDebugKotlin FAILED
e: file:///Users/ivanignatiev/GitHub/****/node_modules/react-native-screens/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.kt:391:46 Unresolved reference: attr
```

I do not know Java or Kotlin, so, please guide me if additional work is needed.

@kkafar: Since AGP 8.0 the default value for `android.nonTransitiveRClass` property [has changed from `false` to `true`](https://developer.android.com/build/releases/past-releases/agp-8-0-0-release-notes). This requires resource file names to be fully qualified. React Native 0.73 depends on AGP 8.0 thus we get the build issues.

PR #1942, opened 2023-10-23
fix(Android): Fix project doesn't compile when `nonTransitiveRClass` is set to true

lukmccall:
## Description

A follow-up to the https://github.com/software-mansion/react-native-screens/issues/1515.
Fixes:
```
e: .../ScreenStackHeaderConfig.kt:391:46 Unresolved reference: attr
```
when the `nonTransitiveRClass` option is set to true.

The `nonTransitiveRClass` is a handy tool that can significantly reduce the size of your app and speed up the build time. It is important to support it. There is one reference to a resource incompatible with this rule: ' R.attr.colorPrimary`. To make it work, you must import it from the `android` package to indicate the build system where that resource is located.

## Changes

Use the absolute path to import the attribute.

- create a new RN project 
- add react-native-screens
- add `android.nonTransitiveRClass=true` to `gradle.properties` 

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.kt
@@ -390,7 +390,7 @@ class ScreenStackHeaderConfig(context: Context) : ViewGroup(context) {
 
         // set primary color as background by default
         val tv = TypedValue()
-        if (context.theme.resolveAttribute(R.attr.colorPrimary, tv, true)) {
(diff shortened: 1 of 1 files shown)


tboba:
HI @lukmccall, thanks for creating this PR!
I believe this change is already included in #1919, but I appreciate the time you've spent on testing the change and collaborating to `react-native-screens` ❤️ 

I'll close this PR for now as it is a duplicate, but please ask any questions related to that change there if you've got any!

Cheers 🙏

PR #1943, opened 2023-10-23
fix: downgrade native deps to avoid conflicts with other libs

kkafar:
## Description

So I've recently bumped versions of few native library dependencies in https://github.com/software-mansion/react-native-screens/pull/1891, but it also bumped some transitive dependencies that conflicted with lower versions that are required by packages such as `gesture-handler`, `reanimated`, `safe-area-context` etc.
I'm downgrading the versions as much I as can while also keeping possible high `material` version.

I downgraded kotlin to `1.6.21` in TestsExample && tested the build.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -124,10 +124,10 @@ repositories {
 
 dependencies {
     implementation 'com.facebook.react:react-native:+'
-    implementation 'androidx.appcompat:appcompat:1.6.1'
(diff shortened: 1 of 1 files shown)


tboba (approved):
I've just checked the change on `Example` app and it works great! 👍

PR #1944, opened 2023-10-23
fix(Android): Set stateWrapper in ScreenViewManager in Fabric

joemun:
## Description

We are in the process of migrating to New Architecture. We noticed a suspicious error message in LogCat, and discovered that `FabricEnabledViewGroup`'s state updates in `updateScreenSizeFabric()` would fail with "setState called without a StateWrapper", due to: 

https://github.com/facebook/react-native/blob/79d620dff7561e6c9370184f9f19e79643486716/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/uimanager/FabricViewStateManager.java#L65

## Changes

Call `fabricViewStateManager.setStateWrapper()` in `ScreenViewManager.kt`.

Screen Shot 2023-10-23 at 11 20 09 AM

Screen Shot 2023-10-23 at 11 32 28 AM

Run `FabricExample` and `Example`. View LogCat logs.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
@@ -42,9 +42,9 @@ class ScreenViewManager : ViewGroupManager<Screen>(), RNSScreenManagerInterface<
         props: ReactStylesDiffMap?,
         stateWrapper: StateWrapper?
     ): Any? {
-        if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED && view.fabricViewStateManager != null) {
+        if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
             // fabricViewStateManager should never be null in Fabric. The null check is only for Paper's empty impl.
-            view.fabricViewStateManager!!.setStateWrapper(stateWrapper)
+            view.fabricViewStateManager?.setStateWrapper(stateWrapper)
         }
         return super.updateState(view, props, stateWrapper)
     }
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
@@ -4,6 +4,8 @@ import com.facebook.react.bridge.JSApplicationIllegalArgumentException
 import com.facebook.react.bridge.ReadableMap
 import com.facebook.react.common.MapBuilder
 import com.facebook.react.module.annotations.ReactModule
+import com.facebook.react.uimanager.ReactStylesDiffMap
+import com.facebook.react.uimanager.StateWrapper
 import com.facebook.react.uimanager.ThemedReactContext
 import com.facebook.react.uimanager.ViewGroupManager
 import com.facebook.react.uimanager.ViewManagerDelegate
@@ -33,6 +35,15 @@ class ScreenViewManager : ViewGroupManager<Screen>(), RNSScreenManagerInterface<
 
     override fun setActivityState(view: Screen, activityState: Float) {
         setActivityState(view, activityState.toInt())
+    }
+
+    override fun updateState(
+        view: Screen,
+        props: ReactStylesDiffMap?,
+        stateWrapper: StateWrapper?
+    ): Any? {
+        view.fabricViewStateManager.setStateWrapper(stateWrapper)
+        return null
     }
 
     @ReactProp(name = "activityState")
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
```diff
@@ -35,6 +37,18 @@ class ScreenViewManager : ViewGroupManager<Screen>(), RNSScreenManagerInterface<
...
     }
 
+    override fun updateState(
+        view: Screen,
+        props: ReactStylesDiffMap?,
+        stateWrapper: StateWrapper?
+    ): Any? {
+        if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
+            view.fabricViewStateManager.setStateWrapper(stateWrapper)
+            return null
```
tboba, line 47: Why we're returning `null` here instead of returning `super.updateState` like below? Is it breaking some logic of the screens there?

tboba (commented):
Hi @joemun, great job! Thanks for creating the PR! 🎉 
Just one comment from me

android/src/paper/java/com/swmansion/rnscreens/FabricEnabledViewGroup.kt
```diff
@@ -2,8 +2,12 @@ package com.swmansion.rnscreens
 
 import android.view.ViewGroup
 import com.facebook.react.bridge.ReactContext
+import com.facebook.react.uimanager.FabricViewStateManager
 
 abstract class FabricEnabledViewGroup constructor(context: ReactContext?) : ViewGroup(context) {
+
+    val fabricViewStateManager get() = null as FabricViewStateManager
```
tboba, line 9: Also one more thing actually 😅 as `fabricViewStateManager` could be nullable, I'll go for typing this val as `FabricViewStateManager?`. I know that we're protecting our backs with `BuildConfig.IS_NEW_ARCHITECTURE_ENABLED`, but I'm afraid this condition may fail at some point and will crash the app. Also, it's more logical for me to make this `optional`, as this val is still reachable from main code and can be used in paper implementation (which is not correct, but it won't crash the app if we use that unintentionally). What do you think?

android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
```diff
@@ -35,6 +37,18 @@ class ScreenViewManager : ViewGroupManager<Screen>(), RNSScreenManagerInterface<
         setActivityState(view, activityState.toInt())
     }
 
+    override fun updateState(
+        view: Screen,
+        props: ReactStylesDiffMap?,
+        stateWrapper: StateWrapper?
+    ): Any? {
+        if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED && view.fabricViewStateManager != null) {
+            // fabricViewStateManager should never be null in Fabric. the null check is only for Paper's empty impl.
```
tboba, line 46:
```suggestion
            // fabricViewStateManager should never be null in Fabric. The null check is only for Paper's empty impl.
```

tboba (commented):
Just one more thing 😄

android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
```diff
@@ -35,6 +37,18 @@ class ScreenViewManager : ViewGroupManager<Screen>(), RNSScreenManagerInterface<
...
 
+    override fun updateState(
+        view: Screen,
+        props: ReactStylesDiffMap?,
+        stateWrapper: StateWrapper?
+    ): Any? {
+        if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED && view.fabricViewStateManager != null) {
+            // fabricViewStateManager should never be null in Fabric. The null check is only for Paper's empty impl.
+            view.fabricViewStateManager!!.setStateWrapper(stateWrapper)
+        }
```
tboba, line 48: Actually, after some time of thinking I think it would be better to change this to:

```suggestion
        if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
            // fabricViewStateManager should never be null in Fabric. The null check is only for Paper's empty impl.
            view.fabricViewStateManager?.setStateWrapper(stateWrapper)
        }
```

sorry that I didn't mentioned it above 🤞

tboba (approved):
LGTM!

tboba:
Hi @joemun! I'm happy to say that we've released new version of react-native-screens (**3.29.0**) that contains changes from this PR 🥳 

PR #1945, opened 2023-10-23
feat(Android): add ios like slide animation

alexandrius:
## Description

This PR adds iOS like slide in animation to Android.
But why?
iOS like slide in animations are extremely popular in Android apps. Twitter, Telegram, Slack and many other apps stack navigation resembles or flat out copies iOS behaviour. 
`react-navigation` non native stack gives us ability to have reimplementation of iOS slide in behavior. However, `expo-router` doesn't support non native stack. Besides that if you want native performance and iOS like behavior you are kinda stuck 🤷. 

## Changes

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/res/base/anim/rns_slide_in_from_left_ios.xml
+++ b/android/src/main/res/base/anim/rns_slide_in_from_left_ios.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <translate xmlns:android="http://schemas.android.com/apk/res/android"
     android:duration="@android:integer/config_shortAnimTime"
-    android:interpolator="@android:interpolator/accelerate_decelerate"
     android:fromXDelta="-30%"
     android:toXDelta="0%" />
--- a/android/src/main/res/base/anim/rns_slide_in_from_left_ios.xml
+++ b/android/src/main/res/base/anim/rns_slide_in_from_left_ios.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<translate xmlns:android="http://schemas.android.com/apk/res/android"
+    android:duration="@android:integer/config_shortAnimTime"
+    android:interpolator="@android:interpolator/accelerate_decelerate"
+    android:fromXDelta="-30%"
+    android:toXDelta="0%" />
--- a/android/src/main/res/base/anim/rns_slide_in_from_right_ios.xml
+++ b/android/src/main/res/base/anim/rns_slide_in_from_right_ios.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<translate xmlns:android="http://schemas.android.com/apk/res/android"
+    android:duration="@android:integer/config_shortAnimTime"
+    android:interpolator="@android:interpolator/accelerate_decelerate"
+    android:fromXDelta="100%"
+    android:toXDelta="0%" />
--- a/android/src/main/res/base/anim/rns_slide_out_to_left_ios.xml
+++ b/android/src/main/res/base/anim/rns_slide_out_to_left_ios.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<translate xmlns:android="http://schemas.android.com/apk/res/android"
+    android:duration="@android:integer/config_shortAnimTime"
+    android:interpolator="@android:interpolator/accelerate_decelerate"
+    android:fromXDelta="0%"
+    android:toXDelta="-30%"/>
--- a/android/src/main/res/base/anim/rns_slide_out_to_left_ios.xml
+++ b/android/src/main/res/base/anim/rns_slide_out_to_left_ios.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <translate xmlns:android="http://schemas.android.com/apk/res/android"
     android:duration="@android:integer/config_shortAnimTime"
-    android:interpolator="@android:interpolator/accelerate_decelerate"
     android:fromXDelta="0%"
     android:toXDelta="-30%"/>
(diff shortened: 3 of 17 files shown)


native-stack/README.md
```diff
@@ -287,6 +287,7 @@ How the given screen should appear/disappear when pushed or popped at the top of
 - `slide_from_bottom` – performs a slide from bottom animation
 - `slide_from_right` - slide in the new screen from right to left (Android only, resolves to default transition on iOS)
 - `slide_from_left` - slide in the new screen from left to right (Android only, resolves to default transition on iOS)
+- `cupetino` - iOS like slide in animation (Android only, resolves to default transition on iOS)
```
fobos531, line 290:
```suggestion
- `cupertino` - iOS like slide in animation (Android only, resolves to default transition on iOS)
```

fobos531 (changes requested):

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
```diff
@@ -203,6 +203,7 @@ Allows for the customization of how the given screen should appear/disappear whe
 - `"slide_from_bottom"` - slide in the new screen from bottom to top
 - `"slide_from_right"` - slide in the new screen from right to left (Android only, resolves to default transition on iOS)
 - `"slide_from_left"` - slide in the new screen from left to right (Android only, resolves to default transition on iOS)
+- `"cupetino"` - iOS like slide in animation (Android only, resolves to default transition on iOS)
```
tboba, line 206: I don't know if `cupertino` sounds better as an animation name here (it's not descriptive from the name). Maybe `slide_out_from_side` / `jump_from_side` would be better here?

native-stack/README.md
tboba, line 290: Same as above

android/src/main/res/base/anim/rns_slide_in_from_left_ios.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<translate xmlns:android="http://schemas.android.com/apk/res/android"
+    android:duration="@android:integer/config_shortAnimTime"
+    android:fromXDelta="-30%"
```
tboba, line 4: Why we're setting `-30%` here and not just `-100%`? I've just tested how the animation looks like on `fromXDelta="-100%"` and to be honest I cannot spot the difference there 🤔 Am I missing something?

android/src/main/res/base/anim/rns_slide_out_to_left_ios.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<translate xmlns:android="http://schemas.android.com/apk/res/android"
+    android:duration="@android:integer/config_shortAnimTime"
+    android:fromXDelta="0%"
+    android:toXDelta="-30%"/>
```
tboba, line 5: same as above

tboba (changes requested):
Hey @alexandrius, great job! Thanks for creating the PR 🥳 
Just left some comments there

android/src/main/res/base/anim/rns_slide_in_from_right_ios.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<translate xmlns:android="http://schemas.android.com/apk/res/android"
+    android:duration="@android:integer/config_shortAnimTime"
+    android:interpolator="@android:interpolator/accelerate_decelerate"
```
tboba, line 4: One more thing - I've also found that interpolator is only provided here. Is it needed?

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
alexandrius: It's inspired by [Flutter](https://docs.flutter.dev/ui/widgets/cupertino). I think `cupertino` is somewhat original and also explains what it actually means.
If it's a firm no - I can change it something like `slide_from_right_ios`.

android/src/main/res/base/anim/rns_slide_in_from_left_ios.xml
alexandrius: The bottom screen doesn't go away fully on iOS:

https://github.com/software-mansion/react-native-screens/assets/5978212/a24e9357-6e0d-4b25-adda-861f6cfdb61a

I captured few frames and it goes off roughly by `30%`

android/src/main/res/base/anim/rns_slide_in_from_right_ios.xml
alexandrius: Yeah basically it tries to somewhat resemble iOS only for `newTop` fragment. Dismissing and bottom fragments do not have the animation (Or I couldn't notice)

native-stack/README.md
alexandrius: Just want to note I'm expert at making rypos :D

android/src/main/res/base/anim/rns_slide_out_to_left_ios.xml
alexandrius: Answered above

android/src/main/res/base/anim/rns_slide_in_from_right_ios.xml
tboba: I understand, but doesn't we also require it in `rns_slide_in_from_left_ios` then?

alexandrius: Added. It looks better now

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
tboba: I know the point, but I don't think this name is not that descriptive as it could be 😕 
Actually, how about just `ios` as the name of the transition?

alexandrius: renamed

tboba (approved):
LGTM! 👍
@WoLewicki @kkafar could you check if the `ios` name for the animation would fit there? I'm for it, but just want to know your opinion about it.

WoLewicki:
I don't have an idea for a better name so seems OK to me 😅

tboba:
Let's get this merged then 🎉

tboba:
Hi @alexandrius! I'm happy to say that we've released new version of react-native-screens (**3.29.0**) which has this change included! 🥳 

Issue #1686, opened 2023-01-19
formSheet presentation only opening half-way in 3.19.0


gorbypark:
### Description

formSheet presentation screens are only opening half-way in 3.19.0.  I reverted to 3.18.2 and everything works as normal.  Looking through the change logs I'm not seeing anything obvious to the cause of this.  Originally I had thought the new formSheet detent PR had merged/released, but it hasn't yet https://github.com/software-mansion/react-native-screens/pull/1649#issuecomment-1396638960.  I have only tried on the "old architecture".  The two screenshots below show 3.19.0 and 3.18.2, respectively, with no other changes made besides the version.

3.19.0
<img width="415" alt="Screen Shot 3.19.0" src="https://user-images.githubusercontent.com/19569469/213394219-b7f5d6df-c729-4913-8590-031444b1f5b8.png">

3.18.2
<img width="415" alt="Screen Shot 3.18.2" src="https://user-images.githubusercontent.com/19569469/213412411-38d70b97-3d6f-4385-8729-7bd4c9f13299.png">

### Steps to reproduce

1. Upgrade to react-native-screens 3.19.0
2. set `'presentation: 'formSheet'` on a screen

### Snack or a link to a repository

.

### Screens version

3.19.0

### React Native version

0.71.0

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Acknowledgements

Yes

gorbypark:
After further investigation, it appears that the medium detent PR was accidentally merged in 3.19.0?  https://github.com/software-mansion/react-native-screens/compare/3.18.2...3.19.0 the changelog is showing it as included in 3.19.0

kkafar:
Yeah, this might be a my mistake, as I've left some part of native implementation intact... But while testing I didn't notice the change of default behaviour. I'll check this & let you know

kkafar:
Yep, everything works fine on main, I've made mistake when removing public API and wrong default value is passed in `3.19.0` 🤦🏻‍♂️ 

It will be fixed with next minor version (won't give you details, but it should not be long from now).

SMJ93:
Hey @kkafar, when you you plan on releasing `3.19.1`?

hirbod:
[...]

PR #1964, opened 2023-11-09
fix(iOS): change RNSScreenDetentType enum to work correctly on Fabric

aljs:
## Description

When opening a formSheet (on iOS) from @react-navigation/native-stack on new arch I've experienced the same issue as https://github.com/software-mansion/react-native-screens/issues/1686 . After some debugging I've discovered that `sheetAllowedDetents={'large'}` transfers to **RNSScreenDetentTypeMedium** and `sheetAllowedDetents={'medium'}` to **RNSScreenDetentTypeLarge**. This PR fixes it.

## Changes

Updated _ios/RNSEnums.h_

-

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSEnums.h
+++ b/ios/RNSEnums.h
@@ -59,8 +59,8 @@ typedef NS_ENUM(NSInteger, RNSScreenStackHeaderSubviewType) {
 };
 
 typedef NS_ENUM(NSInteger, RNSScreenDetentType) {
-  RNSScreenDetentTypeMedium,
   RNSScreenDetentTypeLarge,
+  RNSScreenDetentTypeMedium,
   RNSScreenDetentTypeAll,
 };
 


kkafar:
Hey @aljs, thanks for taking time for investigation and helping us out! However, I doubt that this is the real source of the issue. Most likely this fails because the feature has not been released & tested with `react-navigation`. I mean, it is possible to use it, as the props passed via `react-navigation` components fall through to `react-native-screens` components however there is some JS logic missing on side of `react-navigation`. I'll put it into my backlog! Thanks again!

kkafar:
Yep, I've not merged iOS detents to `react-navigation` yet, there is JS logic missing there.
Thank you for contribution, I'll handle this in `@react-navigation` repo.

PR #1969, opened 2023-11-14
fix(iOS): Don't include AVPlayerView in `traverseForScrollView` method

### LINKED ISSUE #1967: `traverseForScrollView` crash with AVPlayer [iOS 17.2 BETA 1]
## Description
iOS 17.2 beta 1 seems to crashing when using any implementation of `AVPlayer` with `react-native-screens`. Issue occurs only when we are using native controls

## Some informations
In new iOS beta (17.2 beta 1) Apple add new UI element to AVPlayer hierarchy

Screenshot 2023-11-10 at 15 10 24

When we are navigating away from screen, `react-native-screens` calls `traverseForScrollView` and when we are reaching it, it's throwing error that crash app

### Error Message
```
*** -[__NSArray0 objectAtIndex:]: index 0 beyond bounds for empty array
```

## Workaround
We can temporarily catch
Screenshot 2023-11-10 at 15 21 40
with "try catch", but it is not a very good solution because in the future we may miss some "important" error

I think we should wait for next iOS release - If this don't get fixed the next step would be to detect if you are in `AVPlayer` and if so do not call the function 

### Steps to reproduce

To reproduce this issue we need to install `xcode 15.1 beta 2` and iOS 17.2 sumulator.

Steps:
1. Open provided example app
2. Wait for video to load
3. Navigate away from screen
4. Application will crash

https://github.com/abanobmikaeel/react-native-video-crash/

### Screens version

3.27.0

### React Native version

0.72.6

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo bare workflow

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

Any

### Acknowledgements

Yes

tboba:
## Description

Currently, in iOS 17.2 Beta 1-2 there's a bug that causes application to crash after going to the second screen with the video player that contains controls. This PR fixes this issue.

Closes #1967.

## Changes

- Added an if statement that returns if view is type of `AVPlayerView`, as we don't want to signal it about decelerating (sadly, this header file is not in UIKit's public API 😕.

### Before

https://github.com/software-mansion/react-native-screens/assets/23281839/6d040536-45fb-41ea-ad8a-1c86352bf199

### After

https://github.com/software-mansion/react-native-screens/assets/23281839/6da618e4-2bf9-47fb-be9f-e73ae560f081

You can test this change on the repro, attached in the original issue: https://github.com/abanobmikaeel/react-native-video-crash/

## Checklist

- [X] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -1423,6 +1423,15 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
     // we don't want to send `scrollViewDidEndDecelerating` event to JS before the JS thread is ready
     return;
   }
+
+  if ([NSStringFromClass([view class]) isEqualToString:@"AVPlayerView"]) {
+    // Traversing through AVPlayerView is an uncommon edge case that causes the disappearing screen
+    // to an excessive traversal through all video player elements
+    // (e.g., for react-native-video, this includes all controls and additional video views).
+    // Thus, we want to avoid unnecessary traversals through these views.
+    return;
+  }
+
   if ([view isKindOfClass:[UIScrollView class]] &&
       ([[(UIScrollView *)view delegate] respondsToSelector:@selector(scrollViewDidEndDecelerating:)])) {
     [[(UIScrollView *)view delegate] scrollViewDidEndDecelerating:(id)view];


kkafar (approved):
Looks good 🟢

udeyrishi:
Thanks for the patch folks! 🙏🏼 What is the timeline for getting this out in a release?

timharding:
Thanks for fixing this. Just wanted to contribute that we've noticed it starting to pop up in the wild with people running the iOS 17.2 beta.

PR #1970, opened 2023-11-15
fix(iOS): Fix updating bounds while changing interface orientation

### LINKED ISSUE #1738: Incorrect layout when navigating between screens of different orientation on iOS
### Description

This issue is hard to consistently reproduce, however, when navigating between screens with a different orientation e.g. Landscape -> Portrait, the new screen is incorrectly laid out with the previous screen's orientation - see screenshot below of the view hierarchy after pushing a new portrait screen.

Screenshot 2023-03-14 at 16 06 14

https://user-images.githubusercontent.com/31568400/225103581-1b54811b-0113-46e0-a839-14fab60b828f.mp4

As you can see the `RNSScreenView` (child of the `RNSScreen` UIViewController) that I've navigated to has had its frame set to landscape dimensions even though it is actually a portrait screen, resulting in a broken layout. 

I've added a log to the `setFrame` method of the `RNSScreenView` to see where this gets called from and with what dimensions (if you scroll to the right you can see the calling method):
```jsx
// When it breaks ❌:
2023-03-12 20:41:11.536827+0000 OrientationTest[49171:3980390] 👉RNSScreenView setting frame to 0.000000x0.000000 called from 2   UIKitCore                           0x0000000131f58325 UIViewCommonInitWithFrame + 1849
2023-03-12 20:41:11.796876+0000 OrientationTest[49171:3980390] 👉RNSScreenView setting frame to 844.000000x390.000000 called from 2   UIKitCore                           0x0000000130f725fa -[UINavigationController _startCustomTransition:] + 1348
2023-03-12 20:41:11.809961+0000 OrientationTest[49171:3980390] 👉RNSScreenView setting frame to 844.000000x390.000000 called from 2   OrientationTest                            0x0000000104cdeff8 -[RNSScreenStackAnimator animateTransition:] + 392
2023-03-12 20:41:11.823229+0000 OrientationTest[49171:3980390] 👉RNSScreenView setting frame to 844.000000x390.000000 called from 2   OrientationTest                            0x0000000104ce0aa2 -[RNSScreenStackAnimator animateFadeWithTransitionContext:toVC:fromVC:] + 338

// When it lays out correctly ✅:
2023-03-12 20:43:29.145651+0000 OrientationTest[49171:3980390] 👉RNSScreenView setting frame to 0.000000x0.000000 called from 2   UIKitCore                           0x0000000131f58325 UIViewCommonInitWithFrame + 1849
2023-03-12 20:43:29.401373+0000 OrientationTest[49171:3980390] 👉RNSScreenView setting frame to 844.000000x390.000000 called from 2   UIKitCore                           0x0000000130f725fa -[UINavigationController _startCustomTransition:] + 1348
2023-03-12 20:43:29.408939+0000 OrientationTest[49171:3980390] 👉RNSScreenView setting frame to 844.000000x390.000000 called from 2   OrientationTest                            0x0000000104cdeff8 -[RNSScreenStackAnimator animateTransition:] + 392
2023-03-12 20:43:29.417094+0000 OrientationTest[49171:3980390] 👉RNSScreenView setting frame to 844.000000x390.000000 called from 2   OrientationTest                            0x0000000104ce0aa2 -[RNSScreenStackAnimator animateFadeWithTransitionContext:toVC:fromVC:] + 338
2023-03-12 20:43:29.519909+0000 OrientationTest[49171:3980390] 👉RNSScreenView setting frame to 844.000000x844.000000 called from 2   UIKitCore                           0x0000000130f89f46 -[UINavigationController _layoutViewController:] + 1194
2023-03-12 20:43:29.820028+0000 OrientationTest[49171:3980390] 👉RNSScreenView setting frame to 390.000000x844.000000 called from 2   UIKitCore                           0x0000000130f89f46 -[UINavigationController _layoutViewController:] + 1194
```
There are a couple of interesting things to note in the logs above:
- Firstly, the `[UINavigationController _startCustomTransition:]` & related transition animation methods set the new screen to landscape dimensions (even though the screen is specified as portrait). It looks like this is due to the fact the UINavigationController's `supportedInterfaceOrientations` are specified by its top view controller which in this case would still be landscape (before the transition has started and before the new view controllers have been set) & then the UINavigationController internally sets the final frame for the view we are transitioning to based on this in its various animated transition methods.
- Secondly, in the case where the layout is correct, there is a final `[UINavigationController _layoutViewController:]` presumably from it some other update to the navigation controller triggering a re-layout with the correct, portrait dimensions for the screen.

## Expected Behaviour
That when navigating to a portrait screen it should correctly layout with portrait dimensions.

### Steps to reproduce

1. Open example on an iPhone simulator (we've been able to reproduce it on a physical device, but its most easily reproducible here for some reason)
2. Navigate back and forth between Landscape -> Portrait screens
3. Observe the calls to `setFrame` in the `RNSScreenView`
4. Eventually the layout should break as shown in the above screenshot (this might take a number of tries)
5. See that the Portrait screen is laid out with Landscape dimensions in the XCode view hierarchy debugger

https://github.com/thomas-coldwell/rnscreen-orientation-glitch

### Screens version

3.20.0

### React Native version

0.71.3

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

iPhone 14

### Acknowledgements

Yes

tboba:
## Description

Currently while trying to change the interface orientation there's a bug that doesn't layout the next screen, causing the old frame to remain (which means that the next screen will be rotated regarding to the previous orientation.
This PR fixes that problem by calling `[_controller.view setNeedsLayout]` on the `navigationController`, where the view is being rotated.

When I was testing this PR I also came to another solution of setting the frame in `viewWillTransitionToSize` but I didn't spot any differences between one solution and this one.

```objc
- (void)viewWillTransitionToSize:(CGSize)size
       withTransitionCoordinator:(id<UIViewControllerTransitionCoordinator>)coordinator
{
  [super viewWillTransitionToSize:size withTransitionCoordinator:coordinator];
#ifdef RCT_NEW_ARCH_ENABLED
  [coordinator
      animateAlongsideTransition:^(id<UIViewControllerTransitionCoordinatorContext> coordinator) {
        self.screenView.frame = CGRectMake(0, 0, size.width, size.height);
      }
      completion:^(id<UIViewControllerTransitionCoordinatorContext> coordinator) {
        [self.screenView setNeedsLayout];
      }];
#endif
}
```

I've also tried to make a workaround of a reaaaally old bug in React Native: https://github.com/facebook/react-native/issues/16060, but unfortunately I failed with that 😕 

Closes #1738.

## Changes

- Adds call to `setNeedsLayout` during the interface orientation

### Before

https://github.com/software-mansion/react-native-screens/assets/23281839/bcc366ab-9757-4a31-89e0-241c22cdd5c6

### After

https://github.com/software-mansion/react-native-screens/assets/23281839/6634379a-2697-46a3-aed6-3cb156616817

You can find in this PR `Test1970.tsx` that contains test test I've tested while making this change.

## Checklist

- [x] Ensured that CI passes


kkafar (commented):
As we discussed: the iOS change is ok, but we still need to work on the view clipping (or at least determine why the view is clipped) & I would suggest that we remove Test1970 & reuse some other test example with orientation change (if there is one).

ios/RNSScreenStack.mm
```diff
@@ -801,6 +801,7 @@ - (void)navigationController:(UINavigationController *)navigationController
 {
   [self emitOnFinishTransitioningEvent];
   [RNSScreenWindowTraits updateWindowTraits];
+  [_controller.view setNeedsLayout];
```
kkafar, line 804: isn't the `_controller.view == self`?

kkafar (commented):
One more note

ios/RNSScreenStack.mm
tboba: Not quite - as I'm looking onto `_controller.view` it's UILayoutContainerView, whereas self is just RNSScreenStackView

dani-mp:
Are you folks planning on getting this merged? There are a bunch of related issues reported in different repositories about this.

Will94o:
Hello ! thanks for the fix but when will it be merged ?

tboba:
Hi @dani-mp! Could you suggest such patch for changing the orientation? I'm concerned if we could avoid calling `setNeedsLayout` on iOS.

kkafar (approved):
I think we can process with this PR

ios/RNSScreenStack.mm
```diff
@@ -923,6 +923,7 @@ - (void)navigationController:(UINavigationController *)navigationController
...
```
kkafar, line 926: One single note, let's leave a comment above this line why do we trigger layout here (e.g. link to this PR).

kkafar (approved):
Yeah, thanks for that comment.
I think we are good to go.

sanghavip16:
@tboba can you please release new React Native npm package of **"react-native-screens"** with above changes?

kkafar:
We're in release process currently, however still waiting for few PRs to land on main. I won't promise anything, cause last date I promised was that it will be ready for week ago, however it should be soon

PR #1972, opened 2023-11-16
fix: build error with rn 0.73-rc.4

mfazekas:
## Description

Fixes building agains react-native 0.73.0-rc4 in fabric mode

## Changes

Updated common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h to build with fabric on 0.73-rc4.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
+++ b/common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
@@ -12,24 +12,24 @@ class RNSScreenComponentDescriptor final
  public:
   using ConcreteComponentDescriptor::ConcreteComponentDescriptor;
 
-  void adopt(ShadowNode::Unshared const &shadowNode) const override {
+  void adopt(ShadowNode& shadowNode) const override {
     react_native_assert(
-        std::dynamic_pointer_cast<RNSScreenShadowNode>(shadowNode));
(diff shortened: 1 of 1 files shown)


kkafar:
Hey, I've not tested it throughly yet, however this seems like not backward compatible change. 
In case of Fabric, we do not care for backward compat RN, but it looks like it could break Paper backward compatibility and we can not afford it for now. I'll work on adding some compile-time constants for C++ code with RN major, minor & patch versions so that we can compile the right code for the right version (is there different solution)?

mfazekas:
@kkafar Yes it's a breaking change,  this only affected fabric builds, at least in our CI. 

At least on 0.73 this file was not part of the build when building for paper, only when building fabric.

image

https://github.com/rnmapbox/maps/actions/runs/6886704342


I've also mentioned this at https://github.com/reactwg/react-native-releases/discussions/64#discussioncomment-7585049, if there is a C++ macro we can already use. We can probably use Gradle to figure out RN versions and push that to C++.

kkafar:
This is rather unexpected, this file should be compiled (included into compiled file) during Paper build. If it is not the case, we are good, as we do not need to care for Fabric backward compatiblity. 
I'll wait for our CI & test in later on some local branch.

PS. Thanks for contributing!

Issue #1198, opened 2021-11-02
enableFreeze cause useIsFocused stop working


r0b0t3d:
## Description
In bottom tabs, when enabling `enableFreeze`, `useIsFocused` is not triggered when screen blur.

## Steps To Reproduce

1. Create tabs with 2 screens `Home` and `Settings`
2. In `Settings`, add `const isFocused = useIsFocused()`
3. Navigate to `Settings` screen, `isFocused` updated to `true` (this is correct)
4. Navigate back to `Home`, `isFocused` still `true`

### Expected behavior
At step 4, `isFocused` in `Settings` should be `false`

### Actual behavior
At step 4, `isFocused` in `Settings` is `true`

## Platform

I just checked on iOS, not sure about other platforms

- [x] iOS
- [ ] Android
- [ ] Web
- [ ] Windows
- [ ] tvOS

## Workflow

- [ ] Managed workflow
- [x] Bare workflow

## Package versions

| package                                | version |
| -------------------------------------- | ------- |
| react-native                           |     0.66.1    |
| @react-navigation/native               |     6.0.6    |
| @react-navigation/native-stack         |    6.2.5     |
| @react-navigation/bottom-tabs         |    6.0.9     |
| react-native-screens                   |    3.9.0     |

marhaupe:
I can confirm this behaviour, but I think it's expected, isn't it? When screens freeze I assume them to not to render at all which actually makes calling `useIsFocused` redundant. This is just an educated guess though, confirmation from maintainers would be appreciated.

r0b0t3d:
Currently, I'd like to refresh the content when a tab is focused. So I need `useIsFocused` to trigger the request.

Maybe you're right @marhaupe . That mean `useIsFocused` should be updated before the screen get frozen

tomasmozeris:
Thanks for effort to make this feature. Same issue here, I use `useIsFocused` for periodical fetches when screen is in focus, I tried using `useFocusEffect`, but it doesn't catch unfocused state. I could add these fetches as Component and it would behave the same with freeze / unfreeze, but if there's plans to fix it I would better wait

nandorojo:
Isn't this the point of freezing the screen? You disable rendering when it's not focused.

nandorojo:
Can you clarify the issues? Does it not trigger when you refocus?

tomasmozeris:
Update the data then user refocus the screen:
`
const isFocused = useIsFocused();
useEffect(() => {
[...]

PR #1980, opened 2023-12-02
fix: move DelayedFreeze setImmediate into an effect

amadeus:
## Description
Executing side effects in render is usually considered bad manners in react land, and given concurrent rendering could have unexpected results if renders are thrown away.

## Changes
This change makes it so setImmediate fires in an effect, and also cleans up after itself if for whatever reason the callback isn't fired. (although I think it's probably not possible with how setImmediate is implemented)

## Screenshots / GIFs
N/A

We (at Discord) ran essentially this patch for probably over a year now without issue and figured it might be good to give back.

Here's an issue related to it: https://github.com/software-mansion/react-native-screens/issues/1198

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/src/index.native.tsx
+++ b/src/index.native.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
-import React, { PropsWithChildren, ReactNode } from 'react';
+import React, { useEffect, PropsWithChildren, ReactNode } from 'react';
 import {
   Animated,
   Image,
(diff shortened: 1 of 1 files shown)


tboba:
Hi @amadeus! I'm happy to say that we've released new version of react-native-screens (**3.29.0**) which has this change included! 🥳 

FelipeSSantos1:
It appears that a recent commit has introduced an issue in the Jest file. Any file that employs `jest.useFakeTimers()` encounters a failure, throwing an error: `Exceeded timeout of 5000 ms for a hook.` This error points to `"@testing-library/react-native"` imports. The test resumes functioning only when we instruct it not to fake setImmediate with `jest.useFakeTimers({ doNotFake: ["setImmediate"] })`.

PR #1982, opened 2023-12-05
refact: remove mixed CJS/ESM, refactorize index.native.tsx

### LINKED ISSUE #1908: Module format in lib/module is using cjs not esm
### Description

This is causing vite to blow up, and it's really hard to fix without patching the package. I can help contribute a fix just want to get confirmation this is ok to change.

### Steps to reproduce

1. Import react-native-screens from vite
2. It errors, even when using `transformMixedEsModules`
3. `git clone git@github.com:tamagui/tamagui.git`
4. `yarn`
5. `yarn tamastack`
6. Load iOS simulator and load app, it errors

https://github.com/tamagui/tamagui

### Screens version

3.22.1

### React Native version

0.72.0

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

tboba:
## Description

Users are reporting that on Vite they cannot build their project with react-native-screens because of the mixed CJS/ESM files that are being created while building a bundle (you can see [here](https://publint.dev/react-native-screens@3.27.0) that some package managers reports errors about mixed CJS/ESM files). To reduce that behavior I've decided to remove CJS completely while bundling the project, resulting in building only ESM files. Unfortunately because of that I had to remove optional requiring process inside the index.native.tsx file, but this shouldn't have a large impact while using rn-screens. 

I also decided to move some parts of the screens implementation to separate files - this should improve readability, better understanding of code for newcomers and should improve developer experience overall. Having only imports and exports in index files is also a good practice - this was my main reason why I've planned to do that.

Closes #1908 - I'll try to ensure that this will fix Vite for sure 👍 

## Changes

- Disable bundling CJS files from react-native-screens
- Refactorize index.native.tsx files to separate files

First, try to bundle the project - you can see that inside `lib` there shouldn't be `common` directory with the CJS files.
Then, try to run FabricTestExample with a couple of tests. Application should work properly as usual.

## Developer notes
There are some points that I stumbled upon and I need to mention here.
- I've managed to move all of the native components from class to function components, **except**:
  - **Screen:** Unfortunately we need to stay with class components there, as for now we would like to keep behavior of using `setNativeProps` for a screen (does anybody do that? Or is react-native calling this method for a screen wherever? There's a field for a discussion).
  - **SearchBar:** Because of managing ref's state and dropping it down to the methods responsible for commands I was also unable to convert this to functional component.

- I tried to also refactor index.tsx file, but I see no reason to do this. For now I'm keeping it as it is (with only a slight changes to this file):
  - Because of a conflict of naming between SearchBarCommands (from types.tsx) and SearchBarCommands as a native component -> it's not that easy to fix, so I suggest fixing this in a future (might be also a good first issue).
  - I also tried to move `index.native.tsx` to `index.tsx` and to move `index.tsx` to `index.web.tsx`, but because of a conflict I described above and because I don't see the point of rendering conditionally native components depending if `Platform.OS !== 'web'` (and rendering a `View` if Platform.OS is web) I'm keeping the current naming.
- Let me know what do you think about the refactor of index.native.tsx! This change is a **Proof of concept** and I codenamed it as a second stage of this PR, so we might give it a try, but I'm all ears about your opinion - IMO it is worth merging :shipit:.

## Checklist

- [X] Ensured that nothing changed while refactoring index.native.tsx
- [X] Ensured that CI passes


native-stack/package.json
```diff
@@ -1,6 +1,5 @@
 {
-  "main": "../lib/commonjs/native-stack/index",
   "module": "../lib/module/native-stack/index",
   "react-native": "../src/native-stack/index",
-  "types": "../lib/typescript/native-stack/index"
+  "types": "../lib/typescript/native-stack/index.d.ts"
```
tboba, line 4: I've added file formats because react-native-builder bob (while using the newest version) throws an error while the file format is not specified.

src/components/ScreenContainer.tsx
```diff
@@ -0,0 +1,31 @@
...
+import React from 'react';
+import { ScreenContainerProps } from 'react-native-screens';
+import { isNativePlatformSupported, screensEnabled } from '../core';
+
+// Native components
+import ScreenContainerNativeComponent from '../fabric/ScreenContainerNativeComponent';
+import ScreenNavigationContainerNativeComponent from '../fabric/ScreenNavigationContainerNativeComponent';
+
+export const NativeScreenContainer: React.ComponentType<ScreenContainerProps> =
+  ScreenContainerNativeComponent as any;
```
WoLewicki, line 11: Is there no way that there would be no `any` here?

src/core.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export const enableFreeze = (shouldEnableReactFreeze = true) => {
+  if (!isNativePlatformSupported) {
+    return;
+  }
+
+  const minor = parseInt(version.split('.')[1]); // eg. takes 66 from '0.66.0'
+
+  // react-freeze requires react-native >=0.64, react-native from main is 0.0.0
+  if (!(minor === 0 || minor >= 64) && shouldEnableReactFreeze) {
```
WoLewicki, line 39: I think we should remove this already since we support RN >= 0.64.

src/index.tsx
```diff
@@ -9,57 +9,50 @@ import {
   SearchBarProps,
 } from './types';
 
+import { screensEnabled } from './core';
```
WoLewicki, line 12: Could we even maybe remove `index.native.tsx` at all and just do proper platform checks so it all works correctly? Or is there something we cannot manage like this?

WoLewicki (commented):
Looks good :rocket: I left some comments.

kkafar (commented):
Took a quick look now:

1. I like that each component lives in its dedicated source file now <3
2. Is there any advantage of converting every function to arrow function?
3. We need to make sure that there are no more similar issues to the one we debugged last week (component exported under different name after the refactor)

native-stack/package.json
tboba: I've removed it, as I was getting weird errors about unknown `_$_REQUIRE` functions while trying to launch an app.
This was happening on the first times when I was moving the app to ESM and that error does not happen again, so I believe we could bring the `main` back 👍

tboba: About the `.d.ts` addition - somehow react-native-builder-bob requires right now to have that specific file format, as it can't find file by `index` shortcut name anymore.

src/components/FullWindowOverlay.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { PropsWithChildren, ReactNode } from 'react';
+import { Platform, StyleProp, View, ViewStyle } from 'react-native';
+
+// Native components
+import FullWindowOverlayNativeComponent from '../fabric/FullWindowOverlayNativeComponent';
+const NativeFullWindowOverlay: React.ComponentType<
```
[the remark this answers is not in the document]
tboba: I just wanted to stick into the one pattern and not mix between `const`s and `function`s. In my opinion arrow functions looks better as functional components, so I decided to leave consts here.

src/components/Screen.tsx
```diff
@@ -0,0 +1,194 @@
...
+        active,
+        activityState,
+        children,
+        isNativeStack,
+        gestureResponseDistance,
+        onGestureCancel,
+        ...props
+      } = rest;
+
+      if (active !== undefined && activityState === undefined) {
```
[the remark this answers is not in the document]
tboba: Great, thanks for mentioning that. We can do that in a separate PR 👍

src/components/ScreenContainer.tsx
tboba: I would love to not have any `any` here, but this would require us to type all of the native components (that are default exports with codegenNativeComponent - I don't even know if we could type those codegens 🤔)

tboba:
@kkafar 
1. I'm glad you like it! 
2. As I answered above, I just wanted to stick with one convention
3. Yeah, this app should be really deeply tested before merging this PR, but as I've checked already it works as usual on many cases!

native-stack/package.json
tboba: Reverting `main` field - resolved in https://github.com/software-mansion/react-native-screens/pull/1982/commits/30873a08b3c2686c2dbbc848b29aa8994d81a909.

src/core.ts
tboba: Resolved in https://github.com/software-mansion/react-native-screens/pull/1982/commits/e67702499dbe2fc82bdc440d6f6768dc46c5e26a.

src/components/Screen.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  render() {
+    const {
+      enabled = screensEnabled(),
+      freezeOnBlur = freezeEnabled(),
+      ...rest
+    } = this.props;
+
+    // To maintain default behaviour of formSheet stack presentation style & and to have resonable
+    // defaults for new medium-detent iOS API we need to set defaults here
```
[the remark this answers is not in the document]
tboba: Resolved in https://github.com/software-mansion/react-native-screens/pull/1982/commits/2cfce53f8479b2c9bd1d66f6c5f6b58eddb7721a.

src/components/FullWindowOverlay.tsx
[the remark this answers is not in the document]
WoLewicki: As long as it works 🤷

src/components/ScreenContainer.tsx
WoLewicki: Aren't they automatically typed with the props you add as their `NativeProps` type to `codegenNativeComponent`?

tboba: I believe they are not (or am I doing something wrong?). Earlier, when there were optional requires in the code typescript didn't care about the types (since it didn't know what we are importing). Right now it can't handle proper typing (and I see some reasons there - you've mentioned that this component should be automatically typed and maybe you're right, but some NativeProps differs between what types we wanna provide for the users (here - ScreenContainerProps) and what we want to provide for Codegen (e.g. BubblingEventHandlers). You can find such case in Search Bar, where for the search bar commands we provide SearchBarCommandsType for the user and we expect that this native component will match syntactically whereas for the codegen we want to expect that `focus` event will be BubblingEventHandler.

image

tboba: You're probably right, but we still don't provide exact same types as we provide in NativeProps. Earlier, when there were optional requires in the code, Typescript didn't care about the types (since it didn't know what we are importing). Right now it can't handle proper typing (and I see some reasons there - you've mentioned that this component should be automatically typed and maybe you're right, but some NativeProps differs between what types we wanna provide for the users (here - ScreenContainerProps) and what we want to provide for Codegen (e.g. BubblingEventHandlers). You can find such case in Search Bar, where for the search bar commands we provide SearchBarCommandsType for the user and we expect that this native component will match syntactically whereas for the codegen we want to expect that `focus` event will be BubblingEventHandler.

image

native-stack/package.json
tboba: Small update about adding `.d.ts` format here - I see that some package managers were not detecting `index` file properly, so probably in react-native-builder-bob there's a requirement to do such file format 👍 

https://publint.dev/react-native-screens@3.27.0

src/components/SearchBar.tsx
```diff
@@ -0,0 +1,86 @@
...
+  clearText() {
+    this._callMethodWithRef(ref => NativeSearchBarCommands.clearText(ref));
+  }
+
+  setText(text: string) {
+    this._callMethodWithRef(ref => NativeSearchBarCommands.setText(ref, text));
+  }
+
+  render() {
+    if (!isSearchBarAvailableForCurrentPlatform) {
```
WoLewicki, line 75: Did you maybe implement `SearchBar` for windows @marlenecota?

WoLewicki (approved):
Could we merge all `web` files to so we don't have 2 options by doing inline requires? Or is there a problem with that approach? Other than that, LGTM 🚀

tboba:
@WoLewicki I tried already to render conditionally web components if `Platform.OS === 'web'` but because of the imports that are importing Codegen component it's not possible sadly to merge them 😔

WoLewicki:
I meant inline requires, so no importing at the top of the file, but after the `if` for not being on web.

src/index.native.tsx
```diff
@@ -1,619 +0,0 @@
...
-    flag: boolean
-  ) => void;
-  setText: (
-    viewRef: React.ElementRef<typeof ScreensNativeModules.NativeSearchBar>,
-    text: string
-  ) => void;
-};
-
-// We initialize these lazily so that importing the module doesn't throw error when not linked
-// This is necessary coz libraries such as React Navigation import the library where it may not be enabled
```
kkafar, line 100: How is this handled now? Was the point this comment makes considered?

src/types.tsx
```diff
@@ -7,6 +7,7 @@ import {
   TextInputFocusEventData,
   ColorValue,
 } from 'react-native';
+import React from 'react';
```
kkafar, line 10: Can be cleaned up? On the diff it looks like this is the only line added to this source file, and this import is not used at all.

tboba:
@WoLewicki The first thought that comes to my mind is that this would require us to use `require().default` declarations, which is not an ESM format. So I believe this won't work, but let me confirm that 👍

kkafar (commented):
Hey, I love these changes, but as they profoundly reorganise our code I'm a bit nickpicky. Few points I want us to consider before merging this:

1. Do we have to resign from emitting commonjs? Why can't it be shipped alongside ESM? The complaints I saw in https://github.com/software-mansion/react-native-screens/issues/1908 were about CJS/ESM being mixed, not that CJS exists at all. IMO if we can avoid introducing breaking changes by switching one flag somewhere so that CJS is also emitted, we should go this way. If this requires much more setup then I'm fine with going for ESM only (is Metro fine with that?).

2. Actually I'm sceptical and a bit opinionated on converting top-level symbols (our components) to arrow-style functions because **there are** differences. The two I'm mostly concerned about are:

    1. Can we actually lose some information on JS stacktraces? My so-so JS knowledge comes to play here, but if you have:
    ```javascript
    const Home = (props) => {...}
    ```
    will the stacktrace say that exception was thrown in `Home` component or in some anonymous function (that just happens to be assigned to `Home` constant).
    
    2. The one I'm actually sure of is [function hoisting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function#hoisting)
    It is actually easier to work with top level symbols that are hoisted, as it won't trigger `usage-before-declaration` type of errors (especially that in JS you won't get this kind of error, you will just call some `undefined` method). My guess is that this is why e.g. [`react-navigation` does this](https://github.com/react-navigation/react-navigation/blob/main/packages/native-stack/src/views/HeaderConfig.tsx).
    
    
Also some points down below 👇🏻

src/index.native.tsx
kkafar: ☝🏻 This looks like it might be blocking issue. Are we safe here, also cc @WoLewicki so that you see my concerns, as I'm not fully aware on mechanisms that this comment describes.

tboba:
Hi @kkafar, thanks for the code review!
1. Even if we completely resign from mixing ESM and CJS, I'm afraid that some tools (like Vite :D) will still throw an errors while building a project with our package. I heard that Vite deals with simple mixing of ESM and CJS, but this needs to be considered and investigated.
2. Both points you've mentioned are great and worth considering, thanks! But about hoisting, when I was making this PR I haven't spotted any `usage-before-declaration` warnings 🤔

kkafar:
1. I'm talking about emitting ESM/CJS to separate directories. Now as all code is rewritten  to ESM, the babel (or whatever transpiler we use under the hood) won't emit code that "mixes" these two approaches, as my understanding of the problem was that we emitted 'mixed code' to `lib/module`, that used both CJS & ESM. We could emit pure ESM to `lib/module`, pure CJS to `lib/common` & leave everyone happy.

I might hallucinate here a bit, as I could misunderstand the issue.

2. [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) does not mention this among `arrow functions` features & quick googling indicates that indeed, arrow functions do not support hoisting, and most likely won't as it would require hoisting of variables/constants (this is my thought :D)

tboba:
1. I don't know if that's possible :D Of course splitting ESM and CJS to separate dirs is possible and that's how react-native-screens works right now, but I don't know if we could handle bundling conditionally case like:

for lib/commonjs:
```tsx
let thing;

if (Platform.OS !== 'web') {
    thing = require('./fabric/ThingNativeComponent').default;
}
// and use only `thing` here, instead of imported native component
```

for lib/module:
```tsx
import ThingNativeComponent from './fabric/ThingNativeComponent';
```

Also, I was thinking about a concept of managing both `.mjs` and `.cjs` files, but it's a big no-no for me 😄

kkafar:
I've started [reading a bit](https://esbuild.github.io/api/#main-fields) on `main`/`module` & my opinion that whole JS ecosystem is just a mess became much stronger 😅 

> I don't know if that's possible :D Of course splitting ESM and CJS to separate dirs is possible and that's how react-native-screens works right now, but I don't know if we could handle bundling conditionally case like:

Why would we do such thing? Let me be more clear:

* We have now single codebase written using ESM only
* We transpile our code to JS & ESM to e.g. `lib/module` directory
* We transpile our code to JS & CJS to e.g. `lib/cjs` directory (or however we want to call it)

& next in our `package.json`:

```json
"main": "./lib/cjs/index.js"
"module": "./lib/module/index.js"
```

I'm going after interpretation of these fields as [ESBuild suggests](https://esbuild.github.io/api/#main-fields) & [Vite itself](https://v4.vitejs.dev/guide/build.html#library-mode)

tboba:
@kkafar Sure, that makes sense 😄 
But to be honest I'm just wondering, what are the pros of generating CommonJS files? Do we need them or is there any use case where they are needed (Node / SSR)? For example, I see that Reanimated does not generate them 🤔 Except the concept of optional `require` I don't see the point of generating it

tboba:
@kkafar Sure, that convinces me 👍 I'll do that in the next commit

src/index.native.tsx
WoLewicki: It was used to lazily load native components from RNScreens, but since we cannot use inline requires right now, and almost all of those components are used from the beginning, I wouldn't care too much about it. Other libs did not have such mechanism iirc.

src/components/SearchBar.tsx
marlenecota: I'm not very familiar with this module, but I believe it was a partial implementation that did not include SearchBar.

src/types.tsx
tboba: As we're using `React` in types (for example for defining`React.ref`) this import cannot be removed 😕

tboba:
@kkafar all things you've mentioned are done! Could you check those changes?

kkafar (approved):
Haven't had opportunity to check these changes in action, but the code looks OK.

Please make sure that it works as expected (e.g. Vite is supported now), right code is generated (ESM / CJS) to the right directory, and test on our Apps that they work as usual.

src/components/FullWindowOverlay.tsx
```diff
@@ -0,0 +1,25 @@
...
+  PropsWithChildren<{
+    style: StyleProp<ViewStyle>;
+  }>
+> = FullWindowOverlayNativeComponent as any;
+
+function FullWindowOverlay(props: { children: ReactNode }) {
+  if (Platform.OS !== 'ios') {
+    console.warn('Importing FullWindowOverlay is only valid on iOS devices.');
```
WoLewicki, line 14: Maybe we should change it to `Using FullWindowOverlay ...`. But we can do it in another PR ofc

tboba: Sure thing, I'll do it here since this is a pretty small change 👍

tboba: Done in https://github.com/software-mansion/react-native-screens/pull/1982/commits/2a6eedd1c027a51f328bab3034d3e5a3988cc6a8

PR #1987, opened 2023-12-10
feat: add cancelSearch command on SearchBar

Jasonzj:
## Description

Add cancelSearch native command to SearchBar imperative API.
The command provides a way for developers to programmatically cancel an ongoing search operation and clear the search input.

## Changes

- Added JS impl
- Added Native impl (both Android & iOS)
  - RNSSearchBar.mm
  - CustomSearchView.kt
  - SearchBarManager.kt
  - SearchBarView.kt
- Updated JS types
- Updated docs

Test1097 in both FabricTestExample & TestsExample

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [ ] Ensured that CI passes


tboba:
Hi @Jasonzj, thanks for creating this PR!
I believe there's already an option to cancel search from the search bar - you can use `clearText()` and `blur()` from the search bar. Is `cancelSearch` different from using these two commands?

Jasonzj:
Hi @tboba,

Thank you for your response and for taking the time to review my pull request.

Regarding the search functionality, the `clearText()` and `blur()` commands do not fully cancel the search. Specifically, on iOS, after invoking `clearText()` and `blur()`, the `UISearchController`'s `active` property remains `YES`, indicating that the search controller is still active and has not been properly dismissed. The `cancelSearch` command is designed to thoroughly close the search bar, emulating the behavior of manually tapping the cancel button on the search bar. The situation is similar on Android, where we need to ensure the search operation is completely terminated and the search interface is closed.

I hope this clarifies the intent and necessity of the `cancelSearch` command. Please let me know if you need any further information or if there's anything else I can assist with.

kkafar:
I'm just updating you @Jasonzj, that we have some internal debate whether we should add another command or just modify the behaviour of the `blur` command and treat it like a fix? 

Would love to hear your opinion, as the user, do you think that the `blur` command as it is right know (only losing the focus from the text input) is useful and should be left as it is?

Jasonzj:
@kkafar Thank you for involving me in this discussion. 

In my opinion, both `blur` and `cancelSearch` commands have their unique purposes in common app patterns. The `blur` command is useful when the user needs to interact with elements on the search interface without interrupting the search. On the other hand, the `cancelSearch` command enables a complete closure of the search operation when transitioning to a new view after a search result is selected.

I believe retaining both commands would provide more flexibility to developers tailoring to their specific use-cases.

tboba (approved):
I'm also for it 🚀

kkafar:
Thanks @Jasonzj, we will merge it. As I need to ensure Fabric support I'll take it from here

kkafar (approved):
Note: merge it once CI passes

PR #1988, opened 2023-12-12
fix(Android): modify the decorFitsSystemWindow parameter in setNavigationBarHidden

### LINKED ISSUE #1719: navigationBarColor props breaking safeArea / textinput & more
### Description

Seems like the issue I posted here: https://github.com/react-navigation/react-navigation/issues/11139
is actually due to react native screens, so I will post the bug here instead.

On android, when using react navigation, adding navigationBarColor in the screenOptions of a @react-navigation/native-stack will completely mess up the display of the app.

TextInput will no longer be pulled up above the keyboard and the view doesn't adjust naturally

SafeAreaView turns out messed up and needs to be manually added on some screens, while removed on others and it seems random.

### Steps to reproduce

Have a basic native-stack, with a basic screen, use the screen option navigationBarColor on it, and see the whole display being messed up.

https://reactnavigation.org/docs/native-stack-navigator

### Screens version

3.18.2

### React Native version

0.69.7

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

jiyong1:
## Description

Currently when user tries to change the appearance of the navigation bar (using navigationBarHidden / navigationBarColor props) the content of the screen jumps to the place that does not respect safeAreaView. This PR fixes it by removing `decorFitsSystemWindow` declaration, as it breaks the interface by matching it to `fit system window`.

Resolves #1719.

## Changes

- Removed `decorFitsSystemWindow` call from `ScreenWindowTraits.kt`

## Checklist

- [X] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
@@ -197,7 +197,7 @@ object ScreenWindowTraits {
         val screenForNavBarHidden = findScreenForTrait(screen, WindowTraits.NAVIGATION_BAR_HIDDEN)
         val hidden = screenForNavBarHidden?.isNavigationBarHidden ?: false
 
-        WindowCompat.setDecorFitsSystemWindows(window, hidden)
+        WindowCompat.setDecorFitsSystemWindows(window, !hidden)
         if (hidden) {
             WindowInsetsControllerCompat(window, window.decorView).let { controller ->
                 controller.hide(WindowInsetsCompat.Type.navigationBars())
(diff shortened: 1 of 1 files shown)


tboba:
Hi @jiyong1! Thanks for creating this issue!
I have one question about your changes - as I was testing couple months ago a behaviour of `decorFitsSystemWindows` I believe that without this method changing the appearance of the navigation bar (changing color / hiding it) looks better, as the content doesn't jump while fast reloading the app on debug mode. Have you tested how the content of the screen without calling `decorFitsSystemWindows`? If not, could you test it? 🙏

jiyong1:
Hi @tboba 
I tested as you suggested and it seems better without the method. Can I push a commit that removes the method to this pull request?

jiyong1:
@tboba Yes, I have confirmed that the issue is still resolved even after removing it.

tboba (approved):
I've tested that right now and it works like a charm! 🚀

Thanks! ❤️

PR #1990, opened 2023-12-12
feat: make RNScreens compatible with bridgeless mode

WoLewicki:
## Description

PR enabling bridgeless mode in `FabricTestExample` so we can check if the library is compatible with it. The only problem I encountered was the usage of `currentBridge` which now returns `nil` on `iOS`. I made a dirty hack to get the needed module from some private fields. It would be nice to find a better solution for this, but we have to start with something :rocket:

## Changes

- Provide a bridgeless-compatible solution for getting the event dispatcher.

`Test1802.tsx` where we use the `headerHeight` event with `Animated`. For now you have to comment the code using RNGH and Reanimated to test it since they are not yet compatible with bridgeless mode.


ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -310,7 +310,20 @@ + (UIImage *)loadBackButtonImageInViewController:(UIViewController *)vc withConf
...
-        RCTImageLoader *imageLoader = [subview.bridge moduleForClass:[RCTImageLoader class]];
+        RCTImageLoader *imageLoader;
+        if (subview.bridge) {
+          imageLoader = [subview.bridge moduleForClass:[RCTImageLoader class]];
+        } else {
+#ifdef RCT_NEW_ARCH_ENABLED
+          // see more info why we do this in RNSScreen dispatchEventForAnimatedObserver
+          id appDelegate = [[UIApplication sharedApplication] delegate];
+          RCTModuleRegistry *moduleRegistry =
+              (RCTModuleRegistry *)[[appDelegate valueForKey:@"_reactHost"] valueForKey:@"_moduleRegistry"];
```
RSNara, line 321: @WoLewicki, this is very interesting! 

I don't understand the full code. Why do we have to rely on the bridge/module registry for this use-case? Do you think there's any way to re-structure this so that we don't have to rely on the bridge or the module registry? 

I'm asking to know if we need to provide a better api for this. (We definitely shouldn't be doing [[appDelegate valueForKey:@"_reactHost"] valueForKey:@"_moduleRegistry"]).

RSNara: So, from what I can gather: 
- RNSScreenStackHeaderConfig is a component. 
- And it needs access to a native module: RCTImageLoader. 

As far as I'm aware, fabric components do not have access to native modules. (I think this was also the case with legacy components). And... I think there's been some hesitation about giving components access to native modules.

Let me follow up internally!

WoLewicki: Ok, thanks for information and waiting for follow up 🚀

RSNara: @WoLewicki, the recommended approach is to follow the footsteps of the <Image/> component.

Create a hand-written ComponentDescriptor. ComponentDescriptors have access to the contextContainer. The contextContainer contains the RCTImageLoader. 

https://github.com/facebook/react-native/blob/f7f9250f6e90a41e798a8f62738925a2859a751a/packages/react-native/ReactCommon/react/renderer/components/image/ImageComponentDescriptor.h#L20-L39

You'll also need to write a custom ShadowNode. The component descriptor will attach some c++ object to the shadow node that calls into [RCTImageLoader imageForUrl...]. And this Fabric component will be given access to the shadow node. From that shadow node, it can grab that c++ object and call the method that eventually triggers [RCTImageLoader imageForUrl...].

ios/RNSScreen.mm
```diff
@@ -101,6 +102,25 @@ - (UIViewController *)reactViewController
...
+    RCTModuleRegistry *moduleRegistry =
+        (RCTModuleRegistry *)[[appDelegate valueForKey:@"_reactHost"] valueForKey:@"_moduleRegistry"];
+    if (moduleRegistry) {
+      id<RCTEventDispatcherProtocol> legacyEventDispatcher = [moduleRegistry moduleForName:"EventDispatcher"
+                                                                     lazilyLoadIfNecessary:YES];
+
+      [legacyEventDispatcher notifyObserversOfEvent:event];
+    }
+  }
+}
```
RSNara, line 123: @WoLewicki, you'll have to take the <Image/> component approach for this one as well.

In the ContextContainer, we also provide the RCTEventDispatcher. 

https://github.com/facebook/react-native/blob/f7f9250f6e90a41e798a8f62738925a2859a751a/packages/react-native/ReactCommon/react/runtime/platform/ios/ReactCommon/RCTInstance.mm#L270-L277

WoLewicki:
Closing this PR in favor of changes that are already on `main`.

PR #1999, opened 2024-01-02
refact(Android): unify member-field naming convention in Kotlin

kkafar:
## Description

Up to now we had mixed naming conventions -> I've applied [Android official ones](https://developer.android.com/kotlin/style-guide#naming_2)

## Changes

Refactored whole library code so that we stick to single naming convention

Run any example & see that it runs as it should

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt
@@ -191,7 +191,7 @@ open class ScreenContainer(context: Context?) : ViewGroup(context) {
     }
 
     protected fun createTransaction(): FragmentTransaction {
-        return requireNotNull(fragmentManager) { "mFragmentManager is null when creating transaction" }
+        return requireNotNull(fragmentManager) { "fragment manager is null when creating transaction" }
             .beginTransaction()
             .setReorderingAllowed(true)
     }
@@ -316,7 +316,7 @@ open class ScreenContainer(context: Context?) : ViewGroup(context) {
             // detach screens that are no longer active
             val orphaned: MutableSet<Fragment> = HashSet(
                 requireNotNull(fragmentManager) {
-                    "mFragmentManager is null when performing update in ScreenContainer"
+                    "fragment manager is null when performing update in ScreenContainer"
                 }.fragments
             )
             for (fragmentWrapper in screenWrappers) {
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt
@@ -18,15 +18,15 @@ import com.swmansion.rnscreens.Screen.ActivityState
 
 open class ScreenContainer(context: Context?) : ViewGroup(context) {
     @JvmField
-    protected val mScreenFragments = ArrayList<ScreenFragmentWrapper>()
+    protected val screenWrappers = ArrayList<ScreenFragmentWrapper>()
     @JvmField
-    protected var mFragmentManager: FragmentManager? = null
-    private var mIsAttached = false
-    private var mNeedUpdate = false
-    private var mLayoutEnqueued = false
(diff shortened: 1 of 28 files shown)


android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt
```diff
@@ -191,7 +191,7 @@ open class ScreenContainer(context: Context?) : ViewGroup(context) {
     }
 
     protected fun createTransaction(): FragmentTransaction {
-        return requireNotNull(mFragmentManager) { "mFragmentManager is null when creating transaction" }
+        return requireNotNull(fragmentManager) { "mFragmentManager is null when creating transaction" }
```
tboba, line 194:
```suggestion
        return requireNotNull(fragmentManager) { "fragmentManager is null when creating transaction" }
```

android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt
```diff
@@ -315,11 +315,11 @@ open class ScreenContainer(context: Context?) : ViewGroup(context) {
         createTransaction().let {
             // detach screens that are no longer active
             val orphaned: MutableSet<Fragment> = HashSet(
-                requireNotNull(mFragmentManager) {
+                requireNotNull(fragmentManager) {
                     "mFragmentManager is null when performing update in ScreenContainer"
```
tboba, line 319:
```suggestion
                    "fragmentManager is null when performing update in ScreenContainer"
```

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.kt
```diff
@@ -22,28 +22,28 @@ import com.swmansion.rnscreens.events.HeaderAttachedEvent
 import com.swmansion.rnscreens.events.HeaderDetachedEvent
 
 class ScreenStackHeaderConfig(context: Context) : ViewGroup(context) {
-    private val mConfigSubviews = ArrayList<ScreenStackHeaderSubview>(3)
+    private val configSubviews = ArrayList<ScreenStackHeaderSubview>(3)
     val toolbar: CustomToolbar
-    var mIsHidden = false
+    var isHeaderHidden = false  // named this way to avoid conflict with platform's isHidden
```
tboba, line 27: Do you know if there were any other conflicts in naming between internal variables and our ones? I also remember that there were a conflict of `isHidden` when I was doing a PR with Material 3 Top App Bar, but haven't spot any other conflicts - I'm just curious about that :D

tboba (changes requested):
Great job with this one 🚀 I've found that except the files you've changed there are also files `ScreensShadowNode` and `SearchBarView` that needs to be cared of, as they still have variables with `m` prefix - please check them :D
Also, please make sure if you haven't left any comments or errors with old variable names.

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.kt
kkafar: If there would be more, there would be compile-time errors I believe as it would be unambiguous definition.

android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt
kkafar: [d3e38e2](https://github.com/software-mansion/react-native-screens/pull/1999/commits/d3e38e2c115a27b06b85e7a3a2ac1fccdafae199)

android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt
kkafar: [d3e38e2](https://github.com/software-mansion/react-native-screens/pull/1999/commits/d3e38e2c115a27b06b85e7a3a2ac1fccdafae199)

tboba (approved):
🚀

Issue #1985, opened 2023-12-07
Android Build Error RN 0.73 - New Architecture


AlirezaHadjar:
### Description

There's a build issue on Android while building the newly initialized app.

```groovy
[10/12] Building CXX object CMakeFiles/appmodules.dir/Users/alirezahadjar/Desktop/Projects/ReactNative/rnscreensRN73BuildError/android/app/build/generated/rncli/src/main/jni/rncli.cpp.o
  FAILED: CMakeFiles/appmodules.dir/Users/alirezahadjar/Desktop/Projects/ReactNative/rnscreensRN73BuildError/android/app/build/generated/rncli/src/main/jni/rncli.cpp.o 
[...]

PR #2001, opened 2024-01-03
chore(docs): update compatibility table with supported RN versions with Fabric

kkafar:
## Description

So in [discussion](https://github.com/software-mansion/react-native-screens/issues/1985#issuecomment-1875344124) of #1985 I was just made aware that we did not
update supported react-native versions in our readme, leading to some confusion.

## Changes

Updated the readme & also release notes of 3.28.0

N/A

## Checklist

- [x] Ensured that CI passes


tboba (approved):
Looks good to me!

PR #2004, opened 2024-01-05
fix(TVOS): add searchResultsController for Search Bar on TVOS

### LINKED ISSUE #2003: TVOS Crash when using native stack header search bar
### Description

When I add a header search bar to a screen on tvOS the app crashes as passing nil to UISearchController is not supported on tvOS.

Documentation of Apple.
https://developer.apple.com/documentation/uikit/uisearchcontroller/1618647-init

Exception:
Exception	NSException *	"Passing nil to -[UISearchController initWithSearchResultsController:] is not supported on this platform."	0x0000600000d54330

### Steps to reproduce

1. Add headerSearchBar to screen on tvOS
2. Navigate to Screen with headerSearchBar

Basically adding this to a screen and opening screen on tvOS.
image

https://github.com/Duell10111/ReactTube/tree/react-native-screen-search-bar-issue-tvos

### Screens version

4.29.0

### React Native version

0.72

### Platforms

tvOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

None

### Device

iOS simulator

### Device model

Apple TV 4K

### Acknowledgements

Yes

tboba:
## Description

On TVOS applications are crashing when user tries to open a screen with Search Bar. This happens because TVOS requires UISearchController to contain `searchResultsController`. this PR resolves it by adding simple `UIViewController` as searchResultsController.

Fixes #2003.

## Changes

- Fixed crash on TVOS when user tries to open screen with search bar.

### Before

https://github.com/software-mansion/react-native-screens/assets/23281839/4f7b727a-139e-45a0-8141-f40b3e45d101

### After

https://github.com/software-mansion/react-native-screens/assets/23281839/7138f27e-294b-4d66-b5aa-ff2745d0d406

## Checklist

- [X] Ensured that CI passes

### DIFF
--- a/ios/RNSSearchBar.mm
+++ b/ios/RNSSearchBar.mm
@@ -50,7 +50,13 @@ namespace react = facebook::react;
 
 - (void)initCommonProps
 {
+#if !TARGET_OS_TV
   _controller = [[UISearchController alloc] initWithSearchResultsController:nil];
+#else
+  // on TVOS UISearchController must contain searchResultsController.
+  _controller = [[UISearchController alloc] initWithSearchResultsController:[UIViewController new]];
+#endif
+
   _controller.searchBar.delegate = self;
   _hideWhenScrolling = YES;
   _placement = RNSSearchBarPlacementStacked;


kkafar:
I see the problem & why this change will mitigate the crash. My question is, will the search functionality work with random controller being passed in?

kkafar (commented):
I see the problem & why this change will mitigate the crash. My question is, will the search functionality work with random controller being passed in?

tboba:
@kkafar It should, however I believe TVOS doesn't respect its view and its props. I was trying to create new `UIViewController` and set for its view `backgroundColor` and some transformations (via CATransform3D). However, after clicking `Search` I was getting the same view as on the `After` video. I believe that in the case of TVOS, `UIViewController` is just a handy controller for TVOS to have a place to render the view with the keyboard and a "screen" button. However on iOS we can't put a `UIViewController` for a searchResultsController - after typing something to the search bar, content underneath it is unscrollable.

kkafar (approved):
Yeah, thanks for explanation.

PR #2011, opened 2024-01-16
fix(iOS): invalid orientation of contained modals

### LINKED ISSUE #2006: Orientation Bugs with transparentModal and containedTransparentModal
### Description

I've found two classes of orientation bugs related to `transparentModal` and `containedTransparentModal`. I've also created a small demo react native project that shows off these two bugs in isolation. One interesting thing to note -- when I was trying to reproduce these in an Expo project, the `transparentModal` bug did not reproduce, only in a vanilla react-native project which I've attached a link for here: https://github.com/amadeus/react-navigation-bug

These bugs apply to iOS only, as far as I can tell.

### `transparentModal` Bug:

1. Spawn a screen with `transparentModal`
2. Dismiss that view
3. Notice how safe areas are now locked to the portrait values regardless of orientation change?

This bug does not get triggered by any other modal type. Here's a video showing the bug:

https://github.com/software-mansion/react-native-screens/assets/83376/a141648d-53c4-4325-b752-500de5628452

Notice after i've spawned the `transparentModal`, all subsequent rotations show portrait safe areas in landscape orientation?

It also gets a bit weird if you spawn the modals in landscape the entire time -- notice when finally dismissing the transparent modal, the safe areas only break then:

https://github.com/software-mansion/react-native-screens/assets/83376/348479b2-a0af-4a4b-a612-3d70478c07c2

### `containedTransparentModal` Bug:

In trying to work around this issue on Discord with `containedTransparentModal`, I found another bug with `containedTransparentModal` where it breaks if you rotate the screen with the `containedTransparentModal` showing:

1. Spawn a `containedTransparentModal`
2. Rotate screen
3. Notice how `containedTransparentModal` doesn't seem to update it's dimensions with the new rotated size?

https://github.com/software-mansion/react-native-screens/assets/83376/8d96129f-93b7-43ee-a943-f73923fe7e7b

https://github.com/amadeus/react-navigation-bug

### Screens version

3.25.0 (but it appears to even repro in the latest version)

### React Native version

0.73.0

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Release mode

### Device

Real device

### Device model

_No response_

### Acknowledgements

Yes

tboba:
## Description

On iOS there's a bug with contained (and contained transparent) modals, where the modal does not change its orientation when the whole interface (and/or device) is being rotated.
That's because contained modals will keep its orientation when the interface is being rotated, as their controllers don't update the bounds -> `viewDidLayoutSubviews` and `updateBounds` methods were not called at the moment of changing the orientation.

This PR fixes this bug by updating the bounds when `layoutSubviews` method from RNSScreenStack is being called, meaning that no matter if there were previously any modals that have been dismissed or not, screen stack will always update the bounds of the modal views.

Fixes #2006.

## Changes

- Fixed the problem with invalid orientation of contained modals by updating their bounds.

### Before

https://github.com/software-mansion/react-native-screens/assets/23281839/069a2df7-ebb6-4de5-b8be-d5e1d045421b

### After

https://github.com/software-mansion/react-native-screens/assets/23281839/c3cbd102-4c68-4749-b2a0-4eaba6b8703f

You can use Test2008 from TestsExample and FabricTestExample to test the change from this PR.

## Checklist

- [x] Ensured that CI passes


ios/RNSScreenStack.mm
```diff
@@ -646,6 +646,17 @@ - (void)layoutSubviews
 {
   [super layoutSubviews];
   _controller.view.frame = self.bounds;
+
+  // We need to update the bounds of the modal views here, since
+  // for contained modals they are not updated by modals themselves.
+  for (UIViewController *modal in _presentedModals) {
+    BOOL isModalBeingDismissed = [modal isKindOfClass:[RNSScreen class]] && ((RNSScreen *)modal).isBeingDismissed;
```
WoLewicki, line 653: Could you add a comment why we are not taking modals being dismissed into account and why taking the modal's view's superview frame works correctly? Other than that, LGTM 🚀

tboba: Sure!

tboba: Resolved in https://github.com/software-mansion/react-native-screens/pull/2011/commits/40f4f32a411fa7f44d97c4edcd5bdf57769ddde0.

WoLewicki (approved):

Gongreg:
Hello. Thank you for fixing the issue! We've bumped into it, and I wanted to ask when could we expect react-native-screens release with this fix?

tboba:
Hi @Gongreg, I think we should release a new version in about ~1-2 weeks!

PR #2013, opened 2024-01-17
fix: change default value of context of ScreenGestureDetector, add warning for goBackGesture

### LINKED ISSUE #2010: Error: Invalid prop `gestureDetectorBridge` supplied to `React.Fragment` while opening a screen on iOS
### Description

It seems that PR #1913 has introduced a bug while opening (or going to another) screen. After opening any test from TestsExample and FabricTestExample, I'm receiving an error:

```
Warning: Invalid prop `gestureDetectorBridge` supplied to `React.Fragment`. React.Fragment can only have `key` and `children` props.
```

https://github.com/software-mansion/react-native-screens/assets/23281839/629aea1e-f0a1-4597-b161-cf29e7e9f382

### Steps to reproduce

1. Open any test case from TestsExample / FabricTestExample
2. See that after opening an app / after going to another screen warning appears on the console and on the screen.

https://github.com/software-mansion/react-native-screens

### Screens version

3.29.0

### React Native version

0.73.0

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

iPhone 15 Pro (iOS 17.2)

### Acknowledgements

Yes

tboba:
## Description
Previously, after going from one screen to another (or after opening the app) there was an error regarding ScreenGestureDetector:

`Warning: Invalid prop 'gestureDetectorBridge' supplied to 'React.Fragment'. React.Fragment can only have 'key' and 'children' props.`

This error indicates that we were passing props from ScreenGestureDetector to React.Fragment. Indeed, after looking onto the default value of GHContext, there was a Fragment that was receiving invalid props.
This PR fixes this by changing the look of React.Fragment, so we won't pass the props further.

Also, I've added a warning that is being shown when user tries to use `goBackGesture` and the navigator is not wrapped in ScreenGestureDetector.

Fixes #2010.

## Changes

- Fix warning of React.Fragment when the ScreenGestureDetector isn't mounted
- Add warning when user tries to use `goBackGesture` and the navigator is not wrapped in ScreenGestureDetector
- Reformated code a bit

### Before

https://github.com/software-mansion/react-native-screens/assets/23281839/f3517b73-de80-45a5-847e-bd817d3d2682

### After

Video shows behavior on going from one screen to another and a warning when there's prop `goBackGesture` without ScreenGestureProvider.

https://github.com/software-mansion/react-native-screens/assets/23281839/7b477ea6-08cc-40ed-afd5-879675776007

You can test the changes in a several ways:
- By checking Screen animation example from ScreensExample
- By checking Test42 from TestsExample
- By checking TestScreenAnimation from FabricTestExample

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/index.native.tsx
+++ b/src/index.native.tsx
@@ -1,11 +1,5 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
-import React, {
-  Fragment,
-  PropsWithChildren,
-  ReactNode,
-  useEffect,
-  useRef,
-} from 'react';
+import React, { PropsWithChildren, ReactNode, useEffect, useRef } from 'react';
 import {
   Animated,
   Image,
@@ -581,7 +575,9 @@ export type {
 const ScreenContext = React.createContext(InnerScreen);
 
 // context to be used when the user wants full screen swipe (see `gesture-handler` folder in repo)
-const GHContext = React.createContext(Fragment);
+const GHContext = React.createContext(
+  (props: { children: React.ReactNode }) => <>{props.children}</>
+);
 
 class Screen extends React.Component<ScreenProps> {
   static contextType = ScreenContext;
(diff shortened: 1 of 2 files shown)


WoLewicki (approved):
LGTM!

PR #2014, opened 2024-01-17
fix(iOS): stuttering back arrow animation when using headerLargeTitle

dylancom:
## Description
When navigating to a screen that has a large header title and the back title hidden, the back arrow animation bugs (kind of bounces in / stuttery animation).
I noticed that when setting the title to a string, it animates normally.
I found that using a space instead of a complete empty string, the title of the previous screen animates nicely, instead of disappearing instantly.
(Tested on iPhone 15 Pro, iOS 17.0)

EDIT:
I also just noticed that on an iPhone 8 plus (iOS 14.0) the back button completely disappears after the stuttery animation.
It seems that setting it to nil caused multiple bugs...

## Demo (bug)

https://github.com/software-mansion/react-native-screens/assets/12894112/3b104aab-0c03-4940-8692-905a664a5452

## Demo (fix)

https://github.com/software-mansion/react-native-screens/assets/12894112/d43b6c1e-4f62-4ab5-8954-6ad6f0792aac

## Reproduction
```
<Stack.Screen
  name="Home"
  component={HomeScreen}
  options={{ headerLargeTitle: true }}
/>
<Stack.Screen
  name="Details"
  component={DetailsScreen}
  options={{
    headerLargeTitle: true,
    headerBackTitleVisible: false,
  }}
/>

From Home Screen: navigation.navigate('Details')
```

### DIFF
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -528,7 +528,7 @@ namespace react = facebook::react;
 
     // When backBarButtonItem's title is null, back menu will use value
     // of backButtonTitle
-    [backBarButtonItem setTitle:nil];
+    [backBarButtonItem setTitle:@" "];
     prevItem.backButtonTitle = resolvedBackTitle;
   }
   prevItem.backBarButtonItem = backBarButtonItem;


tboba:
Hi @dylancom, thanks for creating this PR! 🥳 
Hm, that's strange 🤔 do you know on which iPhone and which iOS does this back button stutter? I'm seeing this behavior the first time!

dylancom:
You're welcome!
I have tested this on an iPhone 15 Pro, iOS 17.0.

tboba:
@dylancom Yeah, something's definitely wrong there 🤔 
Do you know if previously in RNScreens this bug was also happening for you?

dylancom:
@tboba I have no idea, I haven't been using headerLargeTitle in this way before.

dylancom:
@tboba anything more needed to see this merged?

focux:
I'm experiencing this same issue. I'm using an iPhone 15 Pro with iOS 17.2 and react-native-screens `^3.27.0`

tboba:
@dylancom not at the moment, thanks! Currently I'm struggling with fixing main branch, as it's quite broken, but I'll take a look on that in upcoming days 🙌
@focux thanks for confirming!

tboba:
@dylancom I've just tested this PR and it looks like this change fixes the bug with the title, but unfortunately it breaks the animation of appearing title after pressing back button 😢

Without the change:

https://github.com/software-mansion/react-native-screens/assets/23281839/ecbe7224-c2d3-479a-bcc9-396a166a09d2

With the change (notice that right now the title is being scaled instead of appearing from the side):

https://github.com/software-mansion/react-native-screens/assets/23281839/2242cf4b-5941-493c-9234-52bdda0d721d

Do you know if there's any other solution for fixing that back button item?

dylancom:
If you remove the space and just use an empty string, it doesn't scale.
`[backBarButtonItem setTitle:@""];`

tboba:
@dylancom Yeah, actually I was trying to set the title of the button to an empty string and it breaks the animation - hence I see that you've added an empty space to the title. Sorry for misleading you!

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -528,7 +528,7 @@ + (void)updateViewController:(UIViewController *)vc
 
     // When backBarButtonItem's title is null, back menu will use value
     // of backButtonTitle
-    [backBarButtonItem setTitle:nil];
+    [backBarButtonItem setTitle:@" "];
```
tboba, line 531: Unfortunately, this change still breaks how the hint for the previous screens looks like. With the changed title, it is being set to empty field, which is unacceptable. Can we think about the better approach or "hack" the hints in other way?

https://github.com/software-mansion/react-native-screens/assets/23281839/5cd8f6c1-0039-4c45-b5de-7ac9c194de8c

tboba (changes requested):

ios/RNSScreenStackHeaderConfig.mm
tboba: @dylancom thanks! I've tried some methods mentioned in StackOverflow, but unfortunately setting the text color to `Clear` doesn't fix that bug. Also, I've tried to create a custom menu that will contain the title of the previous screen, but somehow iOS doesn't replace it after calling `setMenu` (I also don't want to steer the menu manually). Hence I'm afraid I won't merge this for now - I don't want to sacrifice one functionality for another, but of course you can patch screens with this change on your own.

tboba:
@dylancom as I said in the subcomment, because of the bug that comes with this change, I won't merge this PR for now (I also can't find any other solution for fixing this issue). If you have any suggestion how this can be fixed, don't hesitate to re-open this PR! Otherwise, can you fill a new issue with the bug described here? You can link this PR to know the additional context of our attempts. Thanks!

PR #2015, opened 2024-01-18
fix(Android): change name of headerBackButtonClicked event

WoLewicki:
## Description

When running app, there is a soft error thrown with such message:
image
and it is true that only this event does not stick to the convention, so we should change it.

## Changes

Removed `Event` suffix in `topHeaderBackButtonClickedEvent` since it should not be there.


tboba (approved):
LGTM! Just one more thing I've spotted which is related to this event - in `ScreenNativeComponent.ts`, `onHeaderBackButtonClicked` is separated from all events (all events are on the top of the file, whereas this event is on the bottom of NativeProps). Can you occasionally move it to other events? :D

https://github.com/software-mansion/react-native-screens/blob/main/src/fabric/ScreenNativeComponent.ts#L99

PR #2019, opened 2024-01-24
fix: android screen stack animation

janicduplessis:
## Description

There was some issues with android animation, the root cause is the drawing order, not the actual animations. This reverts the changes to the animations so it matches stock android ones and always reverse drawing order for android api 33+.

## Changes

Revert the animation changes in https://github.com/software-mansion/react-native-screens/commit/00543fb35201e34dfb5e73e22ff12d274a1ea5f5 and https://github.com/software-mansion/react-native-screens/commit/05f4dd7c77608010d3990ae2af3ec54d5f735760.

Return true from needsDrawReordering when api 33+.

## Screenshots / GIFs

### Before

Api 33

https://github.com/software-mansion/react-native-screens/assets/2677334/8777838c-a2bb-4e43-bcc2-a9a51926eeb9

The mitigations for wrong drawing order makes it kind of hard to see the issue, but going frame by frame we can see the animation is weird initially

image

### After

Api 31

https://github.com/software-mansion/react-native-screens/assets/2677334/2df2eaaa-92ef-4bc0-9ef5-e64e241a38d3

Api 33

https://github.com/software-mansion/react-native-screens/assets/2677334/2588241b-9536-4726-aacc-dca7facb32ad

We can now see that the new screen is actually drawn on top of the old one so it looks better during the initial fade.

image

Tested in example app

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
@@ -2,6 +2,7 @@ package com.swmansion.rnscreens
 
 import android.content.Context
 import android.graphics.Canvas
+import android.os.Build
 import android.view.View
 import com.facebook.react.bridge.ReactContext
 import com.facebook.react.uimanager.UIManagerHelper
@@ -335,8 +336,9 @@ class ScreenStack(context: Context?) : ScreenContainer(context) {
             fragmentWrapper.screen.stackPresentation === Screen.StackPresentation.TRANSPARENT_MODAL
 
         private fun needsDrawReordering(fragmentWrapper: ScreenFragmentWrapper): Boolean =
-            fragmentWrapper.screen.stackAnimation === StackAnimation.SLIDE_FROM_BOTTOM ||
+            Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU ||
+                fragmentWrapper.screen.stackAnimation === StackAnimation.SLIDE_FROM_BOTTOM ||
                 fragmentWrapper.screen.stackAnimation === StackAnimation.FADE_FROM_BOTTOM ||
-                  fragmentWrapper.screen.stackAnimation === StackAnimation.IOS
+                fragmentWrapper.screen.stackAnimation === StackAnimation.IOS
     }
 }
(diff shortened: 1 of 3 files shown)


janicduplessis:
Looks like CI failures are unrelated?

kkafar:
Yeah, they most likely are. Once I check out the behaviour we will proceed with merging.
Thanks for contribution.

tboba:
Hi @janicduplessis, @kkafar - the issue with CI is not related to this PR. Looks like changes made in #1913 have broke our CI (I've already discovered that on screens without `goBackGesture`, back button on Android always acts like pressing and holding while tapping it). So I wouldn't worry about the CI, I've already reported to @piaskowyk about this problem and waiting for his response 👀

tboba:
@janicduplessis That looks soooo damn way better! 🤩 
I think the biggest advantage of this change is that **white remaining white fields are no more** (pictured below).

https://github.com/software-mansion/react-native-screens/assets/23281839/8e546c38-3d65-4030-8317-56de60435c89

Thanks!

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
```diff
@@ -335,8 +336,9 @@ class ScreenStack(context: Context?) : ScreenContainer(context) {
             fragmentWrapper.screen.stackPresentation === Screen.StackPresentation.TRANSPARENT_MODAL
 
         private fun needsDrawReordering(fragmentWrapper: ScreenFragmentWrapper): Boolean =
-            fragmentWrapper.screen.stackAnimation === StackAnimation.SLIDE_FROM_BOTTOM ||
+            Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU ||
```
tboba, line 339: Can you add a comment here why we're comparing Android version (just for a context, that since API 33 the default animation has changed and because of the "sliding" we need to draw reordering 👍)

tboba (commented):
Just one comment from me. Also, could you merge changes from main, so that the CI build will be fixed?

janicduplessis:
@tboba Done!

tboba:
Love it, thanks @janicduplessis!
If we're not in a hurry, I'll confirm from the other commenters from [this PR](https://github.com/react-navigation/react-navigation/issues/11438) that the issues with the animation are not the case from them. Is that alright?

janicduplessis:
Sounds good.

efstathiosntonas:
@janicduplessis thank you for this, can you please run some tests in foldable devices like the one in the screenshot? 

I see some weird jumpiness on my end

Screenshot 2024-02-14 at 17 06 32

tboba:
Hi @efstathiosntonas, I've just tested this animation with provided emulator configuration and I can't see any jumpiness there 🤔 Maybe you've got some other emulator enabled that stutters your foldable device?
For now I'll merge this PR, but we need to observe if there's any issues, related to the animation there.

PR #2023, opened 2024-01-31
fix: change default gesture from Tap to Fling, fix failing CI

tboba:
## Description

Unfortunately, the `Tap` gesture as the default gesture in ScreenGestureDetector breaks the back button on Android and iOS. Somehow, gesture detector while "tapping" just taps and holds the back button, resulting in breaking CI and our testing apps.
This PR changes this gesture to `Fling`, resulting in correct behavior. It cannot be changed to `Manual` gesture, since it also breaks the iOS CI.

## Changes

- Changed gesture from Tap to Fling
- Updated detox from 20.11.4 to 20.17.0
- Fixed Android E2E tests by building the bundle with .so files on x86_64 architecture
- Changed scroll action that was opening background apps preview on Android 33 (and higher) levels
- Adjusted CI parameters, fixed the typo of artifact names

You can go through Example app and try to click back button in `Events`. Right now it should work properly.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/Example/e2e/examplesTests/events.e2e.ts
+++ b/Example/e2e/examplesTests/events.e2e.ts
@@ -8,7 +8,7 @@ describe('Events', () => {
     await waitFor(element(by.id('root-screen-playground-Events')))
       .toBeVisible()
       .whileElement(by.id('root-screen-examples-scrollview'))
-      .scroll(200, 'down');
+      .scroll(100, 'down', NaN, 0.85);
   });
 
   it('should Events playground exist', async () => {
--- a/.github/workflows/android-e2e-test.yml
+++ b/.github/workflows/android-e2e-test.yml
@@ -14,7 +14,7 @@ on:
   workflow_dispatch:
 jobs:
   test:
-    runs-on: macos-12
+    runs-on: ubuntu-latest
     timeout-minutes: 60
     env:
       WORKING_DIRECTORY: Example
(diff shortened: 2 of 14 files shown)


Example/e2e/examplesTests/events.e2e.ts
```diff
@@ -8,7 +8,7 @@ describe('Events', () => {
     await waitFor(element(by.id('root-screen-playground-Events')))
       .toBeVisible()
       .whileElement(by.id('root-screen-examples-scrollview'))
-      .scroll(200, 'down');
+      .scroll(100, 'down', NaN, 0.85);
```
WoLewicki, line 11: What is the reason for changing this value?

WoLewicki:
Also, I am not sure I understand what is going on exactly. Why does gesture-handler call the `tap`? Are we testing custom animations on CI?

tboba:
@WoLewicki we're not testing them, but this change does not only limit to the case of the CI. Because of the default `Tap` option as an empty gesture, Android and iOS simulate that motion as holding the element (I don't specifically know why), resulting in unability to click "go back" button, because it detects the click as the hold motion.

WoLewicki:
Ok keep me updated 🚀

tboba:
@WoLewicki @kkafar I think the changes are ready! 🚀 
Could you take a look on them?

kkafar (commented):
I don't quite get why our tests start to fail. Is it because in `Example` app we now wrap whole application view hierarchy in `GestureDetectorProvider` and gestures are hijacked? Wouldn't it be better to detect from environment whether we are in e2etest scenario and just don't wrap it?

WoLewicki (approved):
I don't like the idea that we are changing the library code only for the e2e tests to pass, but if we are 100% sure it works correctly then it is ok to me.

PR #2025, opened 2024-02-01
feat: add visionOS support

okwasniewski:
## Description

This PR adds visionOS support. Changes are mostly around the same places where tvOS had `ifdefs`. 

## Changes

- Added `visionos` support

## Screenshots / GIFs

https://github.com/software-mansion/react-native-screens/assets/52801365/7618d537-1e31-4cde-9610-da109295d800

1. I've initialized new visionOS project inside of the repo and linked it (same as TVOS example) 

_Note: I didn't add new visionOS example (as they are many of them already) but soon `react-native-test-app` will support visionOS so that might be a solution to test visionOS in the long run_

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenWindowTraits.mm
+++ b/ios/RNSScreenWindowTraits.mm
@@ -22,7 +22,7 @@
 
 + (void)updateStatusBarAppearance
 {
-#if !TARGET_OS_TV
+#if !TARGET_OS_TV && !TARGET_OS_VISION
   [UIView animateWithDuration:0.4
                    animations:^{ // duration based on "Programming iOS 13" p. 311 implementation
 #if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && defined(__IPHONE_13_0) && \
@@ -56,7 +56,9 @@
 #endif
   {
     if (@available(iOS 11.0, *)) {
+#if !TARGET_OS_VISION
       [UIApplication.sharedApplication.keyWindow.rootViewController setNeedsUpdateOfHomeIndicatorAutoHidden];
+#endif
     }
   }
 #endif
@@ -134,7 +136,7 @@
 
 + (void)enforceDesiredDeviceOrientation
 {
-#if !TARGET_OS_TV
+#if !TARGET_OS_TV && !TARGET_OS_VISION
   dispatch_async(dispatch_get_main_queue(), ^{
     UIInterfaceOrientationMask orientationMask = UIInterfaceOrientationMaskAllButUpsideDown;
 #if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && defined(__IPHONE_13_0) && \
@@ -211,7 +213,7 @@
   [RNSScreenWindowTraits updateHomeIndicatorAutoHidden];
 }
 
-#if !TARGET_OS_TV
+#if !TARGET_OS_TV && !TARGET_OS_VISION
 // based on
 // https://stackoverflow.com/questions/57965701/statusbarorientation-was-deprecated-in-ios-13-0-when-attempting-to-get-app-ori/61249908#61249908
 + (UIInterfaceOrientation)interfaceOrientation
--- a/ios/utils/RNSUIBarButtonItem.h
+++ b/ios/utils/RNSUIBarButtonItem.h
@@ -1,3 +1,5 @@
+#import <UIKit/UIKit.h>
+
 @interface RNSUIBarButtonItem : UIBarButtonItem
 
 @property (nonatomic) BOOL menuHidden;
--- a/RNScreens.podspec
+++ b/RNScreens.podspec
@@ -60,7 +60,7 @@ Pod::Spec.new do |s|
   s.homepage     = "https://github.com/software-mansion/react-native-screens"
   s.license      = "MIT"
   s.author       = { "author" => "author@domain.cn" }
-  s.platforms    = { :ios => platform, :tvos => "11.0" }
+  s.platforms    = { :ios => platform, :tvos => "11.0", :visionos => "1.0" }
   s.source       = { :git => "https://github.com/software-mansion/react-native-screens.git", :tag => "#{s.version}" }
   s.source_files = source_files
   s.requires_arc = true
--- a/ios/RNSConvert.h
+++ b/ios/RNSConvert.h
@@ -26,8 +26,10 @@ namespace react = facebook::react;
 + (NSDictionary *)gestureResponseDistanceDictFromCppStruct:
     (const react::RNSScreenGestureResponseDistanceStruct &)gestureResponseDistance;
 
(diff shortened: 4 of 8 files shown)


ios/utils/RNSUIBarButtonItem.h
```diff
@@ -1,3 +1,5 @@
+#import <UIKit/UIKit.h>
```
okwasniewski, line 1: This was erroring out when building for visionOS, not sure why but I think it's good to have the import here

ios/RNSScreenWindowTraits.mm
```diff
@@ -56,7 +56,9 @@ + (void)updateHomeIndicatorAutoHidden
 #endif
   {
     if (@available(iOS 11.0, *)) {
+#if !TARGET_OS_VISION
       [UIApplication.sharedApplication.keyWindow.rootViewController setNeedsUpdateOfHomeIndicatorAutoHidden];
```
tomekzaw, line 60: Should we use `RCTKeyWindow` instead of `UIApplication.sharedApplication.keyWindow` here?

okwasniewski: Yes, definitely - I will add it

ios/RNSScreenWindowTraits.mm
```diff
@@ -22,43 +22,18 @@ + (void)assertViewControllerBasedStatusBarAppearenceSet
...
-                       if (firstWindow != nil) {
-                         [[firstWindow rootViewController] setNeedsStatusBarAppearanceUpdate];
-                       }
-                     } else
-#endif
-                     {
-                       [UIApplication.sharedApplication.keyWindow.rootViewController setNeedsStatusBarAppearanceUpdate];
-                     }
-                   }];
+    [RCTKeyWindow().rootViewController setNeedsStatusBarAppearanceUpdate];
```
okwasniewski, line 28: Small note on why I changed it: 

`RCTKeyWindow()` internally loops over windows: 

```objc
for (UIWindow *window in RCTSharedApplication().windows) {
    if (window.keyWindow) {
      return window;
    }
  }
```

The `window` property is available from iOS 2.0 so I think that this check wasn't needed here and we are safe to just use `RTCKeyWindow()`

tboba:
Hi there, thank you for this PR! 🎉 
I've managed to create a repro with the Example app and unfortunately I've found some things:

1. On entering the screen I'm getting the error from `assertViewControllerBasedStatusBarAppearenceSet` method (RNSScreenWindowTraits). I believe that ifdefing that method should be enough. Can we fix that? 💪 

https://github.com/software-mansion/react-native-screens/assets/23281839/abd22327-1eb4-4681-8b65-c4c46847f263

2. I don't know if the transparent modal should look like on the video below. Usually, transparent modals are used to appear with the previous UIView kept in the hierarchy, behind the modal view - Unlike contained transparent modal, which are working correctly in this case, I can't see same thing with transparent modals 🤔 what do you think about it? Could this be a bug in the VisionOS itself? Just for the context, transparent modals are displayed as `UIModalPresentationOverFullScreen`.

https://github.com/software-mansion/react-native-screens/assets/23281839/bfed11ed-345c-4f39-877b-d0372d8fb7b5

3. After clicking on the title of the screen, its style changes (to the default? I think).

https://github.com/software-mansion/react-native-screens/assets/23281839/3ccfc1dd-5dcd-4100-8bfe-98c0b584b435

And I think that's mostly all! Besides those things, I am quite astonished that screen animations and search bar actually works! Great job!
If you need the repro with the screens example on VisionOS, I can publish it!

okwasniewski:
Hey @tboba, Thanks for taking the time to check this PR. It would be great if you could share the repo 🙏

tboba:
@okwasniewski cool! Here's the repo: https://github.com/tboba/screens-visionos-example

okwasniewski:
@tboba 

1) 

visionOS doesn't have status bar so I added ifdefs to this method 😅 The error might be silenced after adding this key to Info.plist but I don't think it makes sense to show this warning.

2) 

I saw the same behaviour in RN Core - Im not sure why it is showing this way but it looks more like a bug in visionOS than in screens. Let's leave it as is for now and adjust when/if Apple provides a better way to do it.

3) 

Hah this is interesting, I think Apple didn't intended to change for visionOS.. Maybe we can show a warning when user attempts to change it?

tboba:
@okwasniewski I'm concerned if we should allow changing the header's style by clicking (or tapping? I only heard that the visionOS's cursor is being controlled by the eye :D) on it at all 😅

okwasniewski:
@tboba I meant we should ignore changing header color for visionOS (it should be always set to white). I guess what's happening is that the view refreshes when the user taps on it. 

Let's show warning or just ignore changing color when someone (developer) attempts to change it.

tboba:
@okwasniewski, yeah, I would also assume that 😄 
Yeah, in case of changing title's color let's throw a warning and ignore setting the style for VisionOS 👍

okwasniewski:
@tboba I added ifdefs to ignore changing the header title color, I found that adding warnings was really annoying as they were triggered all the time - I think it will be worth adding to documentation that changing color is ignored for this platform

tboba:
@okwasniewski hmm, on the JS side we have the package with `warnOnce` method that will throw the warn only once in the whole app runtime when the given condition is met. You can find the examples of the usage of it in `NativeStackView.tsx` file. What do you think about it? Maybe it is worth trying? I believe you will need to use that method in `HeaderConfig.tsx` file

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -227,7 +227,7 @@ + (void)setAnimatedConfig:(UIViewController *)vc withConfig:(RNSScreenStackHeade
       [navbar setTitleTextAttributes:attrs];
     }
 
-#if !TARGET_OS_TV
+#if !TARGET_OS_T && !TARGET_OS_VISIONV
```
tboba, line 230: One more thing - a typo 😅

tboba (changes requested):

ios/RNSScreenStackHeaderConfig.mm
okwasniewski: fixed 😅

okwasniewski:
@tboba Not sure why Android e2e is failing but it looks like a flaky e2e test

tboba:
@okwasniewski Yup, unfortunately Github Runners are not that performant to assert that some of the components from the example are visible, resulting in test fails 😕 I'll run this test once or twice and it should be green!

tboba (approved):
Looks good to me, thanks!

tomekzaw (approved):
LGTM!

PR #2028, opened 2024-02-07
 feat: correct measure with native header

WoLewicki:
## Description

PR adding the `getContentOriginOffset` to Screen's `ShadowNode` in order to account for native header on both platforms. It seems like for the different screen orientations don't work correctly on `Android` so it needs to be investigated too. It also adds `ModalScreen` concept for now only to have different shadow node there. We need it for different traits since the non-modal Screen can be placed below some other views, so it should not be treated as the root of the view hierarchy by yoga.

## Changes

- Added new prop to `Screen`'s state which is used in `getContentOriginOffset` method which is in turn used by `LayoutableShadowNode` when adding offsets to elements: https://github.com/facebook/react-native/blob/8da1da2e7320e23341e024a855e87c4762b220cc/packages/react-native/ReactCommon/react/renderer/core/LayoutableShadowNode.cpp#L54.
- Added `ModalScreen`

## Screenshots / GIFs

### After

-->

It needs to be tested thoroughly with different scenarios of having headers in nested stacks etc.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.h
```diff
@@ -19,6 +19,7 @@ class JSI_EXPORT RNSScreenShadowNode final : public ConcreteViewShadowNode<
  public:
   using ConcreteViewShadowNode::ConcreteViewShadowNode;
 
+  Point getContentOriginOffset() const override;
   static ShadowNodeTraits BaseTraits() {
     auto traits = ConcreteViewShadowNode::BaseTraits();
```
j-piasecki, line 24: I'll leave it here so it's not lost to time. We believe this trait is necessary for modals, but it breaks touchables (measure to be more precise) in every case where screen is not positioned ad 0,0. We'd need to remove this treat here and make new shadow node specific for modals.

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.h
```diff
@@ -19,6 +19,7 @@ class JSI_EXPORT RNSScreenShadowNode final : public ConcreteViewShadowNode<
...
     traits.set(ShadowNodeTraits::Trait::RootNodeKind);
```
j-piasecki, line 25: I'll leave it here so it's not lost to time. We believe this trait is necessary for modals, but it breaks touchables (measure to be more precise) in every case where screen is not positioned ad 0,0. We'd need to remove this treat here and make new shadow node specific for modals.

j-piasecki:
There were some things left to do, like for example compiling it on the old arch, where Wojti forgot to update the method signature 😄. The thing that needs to be for sure checked is the behavior of nested stacks with different header settings.

I _think_ I will be able to look into it tomorrow, no promises though so if you want to do it, go ahead.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -65,9 +65,9 @@ class Screen(context: ReactContext?) : FabricEnabledViewGroup(context) {
             val width = r - l
             val height = b - t
 
-            calculateHeaderHeight()
+            val headerHeight = calculateHeaderHeight()
```
j-piasecki, line 68: This breaks on the js stack, it returns the height of the header that doesn't exist

tboba:
Hi @WoLewicki, could you merge the changes from main to this PR? I see that you have also bumped the version to RN 0.73.4, which is already done on main 😄

tboba:
Hi @cortinico, just wanted to ask - are there any estimates when RN 0.74 will be officially published?

cortinico:
> Hi @cortinico, just wanted to ask - are there any estimates when RN 0.74 will be officially published?

We're planning to cut the branch next week. Stable release will follow in the coming weeks so it's not immediate future but still quite high on the agenda. I just wanted to make sure this PR doesn't get forgotten

common/cpp/react/renderer/components/rnscreens/RNSModalScreenComponentDescriptor.h
```diff
@@ -0,0 +1,41 @@
...
+class RNSModalScreenComponentDescriptor final
+    : public ConcreteComponentDescriptor<RNSModalScreenShadowNode> {
+ public:
+  using ConcreteComponentDescriptor::ConcreteComponentDescriptor;
+
+  void adopt(ShadowNode& shadowNode) const override {
+    react_native_assert(
+        dynamic_cast<RNSModalScreenShadowNode*>(&shadowNode));
+    auto& screenShadowNode =
+        static_cast<RNSModalScreenShadowNode&>(shadowNode);
```
tboba, line 19: I see that here we're asserting that shadowNode should be of type `RNSModalScreenShadowNode` (with the use of dynamic_cast), but then we're setting screenShadowNode as a static_cast. While this might work as usual, do you know if there are some edge cases when this might break? Or it won't work, when we will dynamic cast this shadow node as a `RNSModalScreenShadowNode`?

ios/RNSModalScreen.h
```diff
@@ -0,0 +1,21 @@
...
+
+@interface RCTConvert (RNSModalScreen)
+
++ (RNSScreenStackPresentation)RNSScreenStackPresentation:(id)json;
++ (RNSScreenStackAnimation)RNSScreenStackAnimation:(id)json;
+
+#if !TARGET_OS_TV
++ (RNSStatusBarStyle)RNSStatusBarStyle:(id)json;
++ (UIStatusBarAnimation)UIStatusBarAnimation:(id)json;
++ (UIInterfaceOrientationMask)UIInterfaceOrientationMask:(id)json;
```
tboba, line 11: Why do we need to have statusbar and orientation oriented methods in RNSModalScreen?

src/components/Screen.tsx
```diff
@@ -61,11 +65,17 @@ export class InnerScreen extends React.Component<ScreenProps> {
...
     if (enabled && isNativePlatformSupported) {
-      AnimatedNativeScreen =
-        AnimatedNativeScreen || Animated.createAnimatedComponent(NativeScreen);
+      const AnimatedScreen =
+        Platform.OS === 'android' ||
+        stackPresentation === 'push' ||
+        stackPresentation === 'containedModal' ||
+        stackPresentation === 'containedTransparentModal'
+          ? AnimatedNativeScreen
+          : AnimatedNativeModalScreen;
```
tboba, line 78: I would add some comment here, that because how iOS modals are being mounted in view hierarchy, we need to use separate component. Also, in the future we need to check if for sure Android modals are being mounted in the same view group as usual screens (and not in the separate view, same as on iOS).

src/fabric/ModalScreenNativeComponent.ts
```diff
@@ -0,0 +1,104 @@
...
+  stackAnimation?: WithDefault<StackAnimation, 'default'>;
+  transitionDuration?: WithDefault<Int32, 350>;
+  replaceAnimation?: WithDefault<ReplaceAnimation, 'pop'>;
+  swipeDirection?: WithDefault<SwipeDirection, 'horizontal'>;
+  hideKeyboardOnSwipe?: boolean;
+  activityState?: WithDefault<Float, -1.0>;
+  navigationBarColor?: ColorValue;
+  navigationBarHidden?: boolean;
+  nativeBackButtonDismissalEnabled?: boolean;
+}
```
tboba, line 100: I don't like the concept, that Modal screen is basically a copy of the screen component 😕 we need to take care of merging properties of both components in separate PR.

tboba (changes requested):
Just left some comments and questions before testing the PR. Please answer them 🎉

tboba:
@WoLewicki also, I think I've found another issue that isn't covered with this fix. When there's a headerLargeTitle and the content is wrapped with the scrollview, touchable's target is still in the incorrect place 🤔 

https://github.com/software-mansion/react-native-screens/assets/23281839/189cf4ba-fcb3-491c-9902-81a66918545b

This time I see that the button is under the whole content. As I've printed calculated header's height, it gives the correct value (96, as it does not count status bar). Do you know about that case?
image

FabricTestExample/Gemfile.lock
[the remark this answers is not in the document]
WoLewicki: We have a Gemfile so running bundle install creates the `Gemfile.lock`. I always use `bundle install` and then `pod install`.

FabricTestExample/android/app/src/main/java/com/fabrictestexample/MainApplication.kt
```diff
@@ -16,11 +16,11 @@ class MainApplication : Application(), ReactApplication {
...
-        override fun getPackages(): List<ReactPackage> {
-            // Packages that cannot be autolinked yet can be added manually here, for example:
-            // packages.add(new MyReactNativePackage());
-            return PackageList(this).packages
-        }
+        override fun getPackages(): List<ReactPackage> =
+            PackageList(this).packages.apply {
+              // Packages that cannot be autolinked yet can be added manually here, for example:
+              // add(MyReactNativePackage())
+            }
```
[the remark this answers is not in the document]
WoLewicki: This comes from the template so I'd leave it as it is for easier version bumps.

FabricTestExample/react-native.config.js
```diff
@@ -1,7 +0,0 @@
-module.exports = {
-  project: {
-    ios: {
-      automaticPodsInstallation: true,
-    },
-  },
-};
```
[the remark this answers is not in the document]
WoLewicki: This comes from the template so I'd leave it as it is for easier version bumps.

FabricTestExample/src/Test2028.tsx
```diff
@@ -0,0 +1,60 @@
+/**
+ * Sample React Native App
+ * https://github.com/facebook/react-native
+ *
+ * @format
+ */
```
WoLewicki, line 6:
```suggestion
```

common/cpp/react/renderer/components/rnscreens/RNSModalScreenComponentDescriptor.h
WoLewicki: I have no knowledge of it, I just copied it from `RNSScreenShadowNode` so it should work fine.

ios/RNSModalScreen.h
WoLewicki: They come with `RCTConvert`. We could try and see if those are really needed in a follow up PR.

src/fabric/ModalScreenNativeComponent.ts
WoLewicki: You cannot do it right now because that's how codegen works.

src/components/Screen.tsx
WoLewicki: > Also, in the future we need to check if for sure Android modals are being mounted in the same view group as usual screens (and not in the separate view, same as on iOS).

You mean in the future as when the new modals will be merged or do you mean the current situation? Currently, modals on `Android` are just transparent fragments with the previous ones not being detached so they are in the same view hierarchy.

common/cpp/react/renderer/components/rnscreens/RNSModalScreenComponentDescriptor.h
j-piasecki: AFAIK `dynamic_cast` and `static_cast` do the same thing but `dynamic_cast` does runtime type checks, while `static_cast` assumes that you know what you're doing. `react_native_assert` does nothing in release builds, so we do both casts only in debug.
If something is wrong, assert and `dynamic_cast` combination will catch that, but doing `static_cast` in production builds gets rid of type-safety overhead.

src/components/Screen.tsx
tboba: I mean the modals, that will be introduced in Screens, since they may be pushed into the separate dialog hierarchy, but we'll check that, when the modal PR will be opened

ios/RNSScreen.mm
```diff
@@ -61,8 +61,7 @@ - (instancetype)initWithFrame:(CGRect)frame
   }
   return self;
 }
-#endif // RCT_NEW_ARCH_ENABLED
-
+#else
 - (instancetype)initWithBridge:(RCTBridge *)bridge
```
tboba, line 65: I think we should also check if is new architecture enabled in places where we're declaring initWithBridge (e.g. RNSFullWindowOverlay, RNSSearchBar). I also see some inconsistency here - in RNSScreenStackHeaderSubview we're declaring this initWithBridge into the `#pragma mark - Paper specific` declaration. Maybe we should move this initializer into the pragmas everywhere? What do you think?

tboba (changes requested):

ios/RNSScreen.mm
tboba: Let's leave that on a separate PR 👍

WoLewicki: Yeah, we can do it in some follow-up for sure. It does not change anything on new arch since this method is just never called.

PR #2031, opened 2024-02-14
fix(iOS): Change retrieving window in FullWindowOverlay

tboba:
## Description

This PR is a follow up for the PR #1526.

Currently, in React Native there's `RCTKeyWindow()` method which goes through each window of `RCTSharedApplication()` and selects the window that is a key window of a given window. This is a better solution, since we don't rely on a single `keyWindow` from RCTSharedApplication.
I've also added a support for Fabric.

I've checked how does FullWindowOverlay work with the change and it works as usual.
Closes #1526.

## Changes

- Changed the method of retrieving window from `delegate.window` to `RCTKeyWindow()`

You can run `Test1844` and check if FullWindowOverlay works correctly.

## Checklist

- [X] Ensured that CI passes

### DIFF
--- a/ios/RNSFullWindowOverlay.mm
+++ b/ios/RNSFullWindowOverlay.mm
@@ -123,7 +123,7 @@
 
 - (void)show
 {
-  UIWindow *window = RCTSharedApplication().delegate.window;
+  UIWindow *window = RCTKeyWindow();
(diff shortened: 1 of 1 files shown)


WoLewicki (approved):
LGTM!

PR #2032, opened 2024-02-14
fix: Update React Native to 0.73.4, change Cocoapods version

tboba:
## Description

Because of the bug with Cocoapods, our CIs are failing on the cached dependencies. This issue has been fixed in React Native 0.73.3 already, but this version requires us to update Cocoapods version. This PR updates our example projects to 0.73.4 version (which is the newest at the moment) and updates the version of Cocoapods.
You can read more about the problem [here](https://github.com/facebook/react-native/issues/42698).

## Changes

- Updated examples that are matching 0.73 version to 0.73.4
- Updated Cocoapods version pattern in Gemfiles

## Checklist

- [x] Ensured that CI passes


Example/Gemfile
```diff
@@ -1,4 +1,8 @@
 source 'https://rubygems.org'
 # You may use http://rbenv.org/ or https://rvm.io/ to install and use this version
 ruby '>= 2.6.10'
-gem 'cocoapods', '~> 1.12'
+
+# Cocoapods 1.15 introduced a bug which break the build. We will remove the upper
+# bound in the template on Cocoapods with next React Native release.
+gem 'cocoapods', '>= 1.13', '< 1.15'
+gem 'activesupport', '>= 6.1.7.5', '< 7.1.0'
```
tboba, line 8: I know that Example app is quite outdated, since it's the only app that hasn't been updated to 0.73, but I'm worried that the bug might also occur there (also, example app passes `pod install` properly, so I believe we can also use those versions).

.github/workflows/ios-build-test-fabric.yml
```diff
@@ -39,6 +39,9 @@ jobs:
         working-directory: ${{ env.WORKING_DIRECTORY }}
         run: yarn
 
+      - name: Install Ruby gems
+        run: gem install cocoapods -v 1.15.2
+
```
kkafar, line 44: Isn't it enough to bump `cocoapods` dependeny in `Gemfile` of each application?

.github/workflows/ios-build-test.yml
```diff
@@ -35,6 +35,8 @@ jobs:
       - name: Install node dependencies
         working-directory: ${{ env.WORKING_DIRECTORY }}
         run: yarn
+      - name: Install Ruby gems
+        run: gem install cocoapods -v 1.15.2
```
kkafar, line 39: Same here

.github/workflows/ios-e2e-test.yml
```diff
@@ -39,6 +39,8 @@ jobs:
       - name: Install node dependencies
         working-directory: ${{ env.WORKING_DIRECTORY }}
         run: yarn
+      - name: Install Ruby gems
+        run: gem install cocoapods -v 1.15.2
```
kkafar, line 43: and here

Example/Gemfile
kkafar, line 8: Just set it to 1.15.2 here?

Example/ios/Podfile.lock
```diff
@@ -749,4 +749,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 86e380a4262db238c7a45428750af2d88465585c
 
-COCOAPODS: 1.14.3
+COCOAPODS: 1.11.3
```
kkafar, line 752: Something went wrong here? 1.11.3 is quite outdated

FabricExample/ios/FabricExample.xcodeproj/project.pbxproj
```diff
@@ -584,11 +584,7 @@
 					"-DFOLLY_CFG_NO_COROUTINES=1",
 					"-DRN_FABRIC_ENABLED",
 				);
-				OTHER_LDFLAGS = (
-					"$(inherited)",
-					"-Wl",
-					"-ld_classic",
-				);
+				OTHER_LDFLAGS = "$(inherited)";
```
kkafar, line 587: Is ignoring linker warnings intentional here?

kkafar (commented):
Left mainly questions

.github/workflows/ios-build-test-fabric.yml
tboba: I've tried to just use the range, but unfortunately Github Actions were still failing in my case

.github/workflows/ios-build-test.yml
tboba: Responded above 👍

.github/workflows/ios-e2e-test.yml
tboba: Responded above 👍

Example/Gemfile
tboba: Yeah, we can try to set strict 1.15.2 version

Example/ios/Podfile.lock
tboba: To be honest I don't know why it's targetting that old version, but I see that in other examples cocoapods version is the same 🤔

kkafar: I think it comes from the version installed on your device

kkafar (approved):
If CI is happy, I'm happy too

PR #2033, opened 2024-02-15
fix(iOS): Remove calculating status bar height in useAnimatedHeaderHeight when header is not shown

tboba:
## Description

This PR fixes the behaviour of `useAnimatedHeaderHeight`, when header is not shown.
Currently, it includes the status bar height when `headerShown` prop is false. This is wrong, since we don't have a header in such scenario and we don't want to include status bar there. Therefore, when navigation controller does not exist or `isNavigationBarHidden` is set to true, we simply return 0.

## Changes

- Removed calculating status bar height when header is not shown.

You can use `Test1802.tsx` to test these changes.

## Checklist

- [x] Ensured that CI passes


tboba:
@WoLewicki could you check this one before merging #2034?

PR #2036, opened 2024-02-16
chore: drop React Native 0.64 - 0.67 since 3.30.0

tboba:
## Description

Since React Native Screens 3.30.0, we will support only 0.68+ versions of React Native. This PR drops support for the versions, lower than that revision.

## Changes

- Added row to the table of supported versions in README
- Added missing bullet point for supporting visionOS
- Removed unnecessary annotation
- Removed note about bug in RN 0.61 (sadly this still occurs in newest RN)

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ To learn about how to use `react-native-screens` with Fabric architecture, head 
 - [x] iOS
 - [x] Android
 - [x] tvOS
+- [x] visionOS
 - [x] Windows
 - [x] Web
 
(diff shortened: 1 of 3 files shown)


tboba:
I wonder what about support for Windows - in the README file of that package we're mainly mentioning RN 0.62 / 0.63. What do you think, should we also update those versions in README (and maybe remove the mention about manual linking) or not?

PR #2060, opened 2024-03-07
fix(iOS): Add view check for getting StackView in `goBackGesture`

piaskowyk:
## Description

This PR fixes problem with goBackGesture screen transition. If stackView can not be found, transition just shouldn't start instead of crashing app and throwing the `Invalid view type, expecting RNSScreenStackView` error.

## Checklist

- [x] Ensured that CI passes


tboba (approved):
LGTM! Thanks 🎉

PR #2061, opened 2024-03-07
fix: Add constraints for velocity in `goBackGesture` screen transition

piaskowyk:
## Description

Without this constraint, especially on the simulator, `event.velocity` sometimes has a huge value. This can lead to unintended behavior where even a slight movement can trigger a transition back to the previous screen instead of canceling it.

## Checklist

- [x] Ensured that CI passes


tboba (approved):
Yeah, this is especially important, when velocity can sometimes hit to `2000` value 😅 Thanks!

PR #2062, opened 2024-03-11
fix(CI): stabilize Android E2E tests

kirillzyusko:
## Description

Used test butler to make e2e tests on Android stable.

## Changes

- added `utilBinaryPaths`
- added `test-butler.apk`

Run several times on CI 👀 

## Checklist

- [x] Ensured that CI passes


Example/e2e/apks/test-butler-app-2.2.1.apk
tboba, line 1: Maybe `apps` would be better, as a directory name? What do you think?

tboba (commented):
Great stuff! Thanks ❤️

tboba:
Also, I suspect that this commit may have introduced the error - https://github.com/software-mansion/react-native-screens/commit/421e22c61243a8c0761c786ccb68c3b209aa1131. Maybe it's worth checking if reverting this change fixes not responding app notification?

.github/workflows/android-e2e-test.yml
```diff
@@ -48,7 +48,7 @@ jobs:
         with:
           working-directory: ${{ env.WORKING_DIRECTORY }}
           api-level: 31
-          target: default
+          target: aosp_atd
```
kirillzyusko, line 51: It seems like this change does the trick 🤔 

Actually you have "System UI isn't responding" popup instead of ANR (Application non responsive). So it's a different story, and from what observed in my experience "System UI isn't responding" happens if runner is weak. I've tried:
- to use macos-14 runner (but in this case the emulator doesn't start at all)
- to manually suppress this popup (doesn't work and popup is still present)
- to switch to ATD image (Automated Test Device) - and it seems like it really helped

However I still think that keeping test butler can help in the future to prevent undesired ANR and make e2e tests more reliable 👀

kirillzyusko:
> Also, I suspect that this commit may have introduced the error - https://github.com/software-mansion/react-native-screens/commit/421e22c61243a8c0761c786ccb68c3b209aa1131. Maybe it's worth checking if reverting this change fixes not responding app notification (if at any chance it still occurs)?

@tboba I don't think that this commit may have an impact on "System UI isn't responding" popup, because this popup most likely gets shown before any e2e test execution and even before app installation.

One of assumption that somewhere something with this commit happened in parallel (maybe GH started to host macos-12 runners that on less powerful hardware, or maybe one of action has been updated and started to consume more resources) - hard to say what exactly could happen, but I think this commit doesn't introduce this regression 🤔

kirillzyusko:
@tboba Third attempt has failed because "11. Main | transitionStart | opening" wasn't found, but on screenshot I can see "13. Main | transitionStart | opening" is available 🤔 

Is it because of a bug in RNS (race condition)?

image

tboba:
@kirillzyusko Yup, that's a known issue 😕 This test case is sadly nondeterministic, but in the near future I have plans to rewrite our example app and write better E2E tests, so I think that's not a problem for now.

tboba (approved):
I think we can merge this! Thanks 🚀

PR #2068, opened 2024-03-19
fix: change the search icon's color according to search bar's state

Cancercookie:
## Description

In our project we have the following configuration:
```
{
  barTintColor: theme.neutral100w,
  hideWhenScrolling: false,
  tintColor: "white",
  headerIconColor: "white",
  hideNavigationBar: true,
  placeholder: LABELS.SEARCH_PLACEHOLDER,
  hintTextColor: "gray",
  shouldShowHintSearchIcon: false,
}
```

that on Android gives us: 

Screenshot 2024-03-19 at 10 59 16

what happens is that because of `headerIconColor: "white"` we get this result on open: 

Screenshot 2024-03-19 at 10 59 24

What happens is that `headerIconColor` is applied to the clear icon too, making it blend with the background.

I've tried implementing a solution to set the same color as `hintTextColor` in our codebase using onOpen and onClose events but it was unstable.

## Changes

It makes sense to me that the color of the clear icon should follow `hintTextColor` when  search is opened.
What I propose is setting the color once when we land on the page and then setting it again according to search state each time `setToolbarElementsVisibility` gets called

## Screenshots / GIFs

The end result: 

Screenshot 2024-03-19 at 11 12 53

Screenshot 2024-03-19 at 11 12 48

- Use the previous configuration as values for `headerSearchBarOptions` in any Screen.
- Open/close search field

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/SearchBarView.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/SearchBarView.kt
@@ -60,7 +60,6 @@ class SearchBarView(reactContext: ReactContext?) : ReactViewGroup(reactContext) 
             searchView.inputType = inputType.toAndroidInputType(autoCapitalize)
             searchViewFormatter?.setTextColor(textColor)
             searchViewFormatter?.setTintColor(tintColor)
-            searchViewFormatter?.setHeaderIconColor(headerIconColor)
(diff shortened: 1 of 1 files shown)


kkafar:
@kkafar

PR #2071, opened 2024-03-19
fix: handle setting `display` for `_viewConfig` attribute

WoLewicki:
## Description

PR fixing the app freezes due to `react-freeze` 😄


tboba (approved):
LGTM!

PR #2072, opened 2024-03-21
fix: remove display setting on native side

WoLewicki:
## Description

Draft PR for fixing the display on native side.

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt
@@ -303,7 +303,7 @@ open class ScreenContainer(context: Context?) : ViewGroup(context) {
     }
 
     private fun onScreenChanged() {
-        // we perform update in `onBeforeLayout` of `ScreensShadowNode` by adding an UIBlock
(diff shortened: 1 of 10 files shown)


tboba:
@WoLewicki since there's open issue on RN repo, addressing why did we open this PR, should we close this PR?

WoLewicki:
I believe it can stick for now, and when we resolve the problem in the core, then we can close this one too. Also, this PR is not strictly connected to the new bugs coming from `react-freeze`, its just the matter of how we make it work in our case.

kkafar:
I do not remember what this PR was about & what issues in core you refer to in the conversation. Can we close it, or is it still relevant?

WoLewicki:
It was connected to our hacky solution for keeping the previous screen visible when applying freeze. Since `react-freeze` or rather `Suspense`(?) set the display to `none` under the hood for suspended components, we wanted it to not apply. This PR was trying to move this behavior to the native side instead of JS. I am not aware of the current state of those things, so it's up to you what we want to do with it next.

kkafar:
Okay, thanks for context. This is now set in JS. I'll close it since it won't be worked on within any reasonable timeline, and we can always reopen it when needed.

PR #2074, opened 2024-03-22
fix: Throw error when app doesn't contain appropriate reanimated version

tboba:
## Description

While using `<GestureDetectorProvider>`, currently when user does not have appropriate react-native-reanimated version, an error is thrown about the undefined value of `ScreenTransition`.

This PR fixes that by throwing an error, but with appropriate message what might went wrong.

Try to use Reanimated 3.8.1 and wrap your navigator with GestureDetectorProvider. For our `Example`, just changing the version of Reanimated from nightly will be enough!

## Checklist

- [ ] Ensured that CI passes


tboba:
After talking with @WoLewicki this seems like an overkill - why we should need to handle such case?

PR #2075, opened 2024-03-25
fix(Android): add notifying for header height change, fix header height values

tboba:
## Description

This PR reverts the change from another PR #2028 that removed notifying about header height change event. Also, here I've also included changes that are fixing incorrect values being returned from `calculateHeaderHeight()` function, which causes to move the target of the touchables 🚀 

## Changes

- Fixed incorrect values, returned from `calculateHeaderHeight` method
- Reverted notifying about header height change

You can check `Test556.tsx` and `Test1072.tsx` to check if touchables are working correctly.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -68,7 +68,7 @@ class Screen(context: ReactContext?) : FabricEnabledViewGroup(context) {
             val headerHeight = calculateHeaderHeight()
             val totalHeight = headerHeight.first + headerHeight.second // action bar height + status bar height
             if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
-                updateScreenSizeFabric(width, height, headerHeight.first + headerHeight.second)
+                updateScreenSizeFabric(width, height, totalHeight)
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -65,12 +65,15 @@ class Screen(context: ReactContext?) : FabricEnabledViewGroup(context) {
             val width = r - l
             val height = b - t
 
-            val headerHeight = calculateHeaderHeight().first
+            val headerHeight = calculateHeaderHeight()
+            val totalHeight = headerHeight.first + headerHeight.second // action bar height + status bar height
             if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
-                updateScreenSizeFabric(width, height, headerHeight)
+                updateScreenSizeFabric(width, height, headerHeight.first + headerHeight.second)
```
WoLewicki, line 71:
```suggestion
                updateScreenSizeFabric(width, height, totalHeight)
```

WoLewicki (approved):
Added one comment. Other than that, LGTM :tada:

PR #2081, opened 2024-03-25
fix: remove postinstall step from package.json

tboba:
## Description

This PR resolves issue with not working installation on react-native-screens on 3.30.0. It seems that `postinstall` step is breaking due to not finding `react-navigation` submodule.

## Changes

- Removed postinstall step from package.json

## Checklist

- [x] Ensured that CI passes


kkafar (approved):
Yep 👍🏻

kkafar:
Let's release patch version straight away

PR #2084, opened 2024-03-27
feat: Migrate example apps to use react-navigation

tboba:
## Description

This PR migrates our example apps to use react-navigation. Because we're right now using react-navigation as a submodule (and we intent to remove native-stack v5 from codebase), it's mandatory to migrate all of the example apps to @react-navigation/native-stack. To do this, I've moved all imports for creating navigator to use native-stack v6 and changed the naming of the props. Right now all of the examples should work, but it's important to build `lib` directories from submodule (using `yarn build`) beforehand.

## Changes

- Moved all navigators to use `@react-navigation/native-stack`
- Changed naming of the props to match native-stack v6
- Changed E2E tests a little

You can just run example apps to test this change.

## Checklist

- [x] Ensured that CI passes


kirillzyusko:
Hi @tboba 

Just seen that it looks like you struggle with CI/e2e tests (timeout 120000ms) - I think I experience the same problem, but I haven't found a solution yet 🙈 

I've tried many things:
- changing XCode version: 14.2, 14.3, 15.0, 15.2
- changing macOS runner: macos-12, macos-13, macos-14
- switching between detox versions

And all that didn't help 😔 If you find a fix - please let me know what exactly helped you 😅 

From my observations- it looks like GH runners became very slow. I've captured videos and when I see them - app works veeeryyy slow 🤷‍♂️ So now I tend to think that something happened to GH hardware (but maybe a new version of software has been released somewhere and we use it now and it causes a slowness).

Anyway just wanted to share my observations here - maybe you'll find them useful 😅

tboba:
@kirillzyusko Oh, yeaaaah It took me soo long today to observe what have I changed in my files, that our emulators are struggling with running 😅 Something's definitely broken on GH infrastructure 😕 Thanks for letting me know!

kirillzyusko:
@tboba yeah, here is an issue on GitHub if you are interested in what caused it: https://github.com/actions/runner-images/issues/9591 👀

WoLewicki (approved):
LGTM!

.github/workflows/android-e2e-test.yml
```diff
@@ -38,8 +38,14 @@ jobs:
...
         run: yarn install && yarn submodules
       - name: Install node dependencies
+        id: install_deps
         working-directory: ${{ env.WORKING_DIRECTORY }}
+        continue-on-error: true
         run: yarn install
+      - if: steps.install_deps.outcome == 'failure'
+        name: Reinstall node dependencies
+        working-directory: ${{ env.WORKING_DIRECTORY }}
+        run: yarn
```
kkafar, line 48: I guess you added this step because something was failing. Can you describe that issue here?

Example/metro.config.js
```diff
@@ -35,42 +42,56 @@ const config = {
...
             __dirname,
             'node_modules',
             moduleName,
             'src',
-            'index.tsx'
-          ),
-          type: 'sourceFile',
-        };
-      }
+            `index${ext}`
```
kkafar, line 80: Why do we want to consider different extensions & paths here? Shouldn't `index.tsx` always be present? What was the issue here?

Example/src/screens/Events.tsx
```diff
@@ -256,14 +287,10 @@ const App = (): JSX.Element => {
 
   return (
     <ToastProvider>
-      <Stack.Navigator
-        screenOptions={{
-          direction: I18nManager.isRTL ? 'rtl' : 'ltr',
-          nativeBackButtonDismissalEnabled: true,
-        }}>
+      <Stack.Navigator>
```
kkafar, line 290: Is direction not supported on v6?

FabricExample/App.js
```diff
@@ -1,3 +1,4 @@
+import 'react-native-gesture-handler';
```
kkafar, line 1: Why do we import GH here?

kkafar (commented):
Few questions from my side

Example/src/screens/Events.tsx
tboba: Yep - currently it is fully automated on react-navigation side!

.github/workflows/android-e2e-test.yml
tboba: Previously, when CI was building node dependencies, it was struggling with file linking of react-navigation, after we've been doing `yarn submodules` on a project (it was mostly throwing file not found error). What's strange is that after the second run, the error was disappearing since yarn probably already refreshed the filesystem tree? I don't know how it works under the hood, I also believe this error should be gone now, but I left it in case of any emergency (and also for keeping iOS and Android runners similar, since on iOS this step is already there!).

Example/metro.config.js
tboba: I've mainly coded this for to dynamically resolve any conflict and ensure we're only resolving deps from node_modules, but the thing is that different dependency has different way of storing index file. Some indexes may be included from `src`, and some of them may lie in the main dependency directory. Also, it's not a rule to keep `.tsx` extension for index, since index may have `.js` or `.ts` extension - with this method we're solving most of the cases. I'm aware some of the deps may lie not in `src` dir and not in the main directory aswell, but I don't know specific scenario ¯\_(ツ)_/¯¯\_(ツ)_/¯.

FabricExample/App.js
tboba: That's how react-native-gesture-handler should be used, according to the installation steps from the [docs](https://docs.swmansion.com/react-native-gesture-handler/docs/fundamentals/installation). Also, I had to add this dependency here, since react-navigation requires having react-native-gesture-handler for using navigator from `@react-navigation/stack`. I know we won't mostly use this stack, as we want to mainly test native-stack of course, but in case of trying to use `createStackNavigator` let's leave this import here and disallow to throw weird exception of not bundled gesture-handler 😄

.github/workflows/android-e2e-test.yml
kkafar: Oki doki

Example/metro.config.js
kkafar: Ok, I guess more robust approach would be to compute intersection of app dependencies & react-navigation dependencies & do it only for those. 

As long as it works I don't mind this approach.

kkafar (approved):
Resolved all conversations thanks!

PR #2087, opened 2024-04-02
chore(Android): fix emit type parameters

mym0404:
## Description

Currently, on Android every event has one class, putted in generic - `ScreenAppearEvent`. This seems to be wrong, as every event should have their own class inside `Event` generic. This PR changes this by replacing `ScreenAppearEvent` to the appropriate event class.

## Changes

- Replaced ScreenAppearEvent to appropriate event types inside generics.

## Checklist

- [x] Ensured that CI passes


tboba (approved):
LGTM! Thanks for that change 🎉

Issue #2082, opened 2024-03-25
Android Build Error - Execution failed for task ':react-native-screens:configureCMakeRelWithDebInfo'


FraserKemp:
### Description

When assembling a dev release build on android I get a build error. 

Error shown is :

* What went wrong:

Execution failed for task ':react-native-screens:configureCMakeRelWithDebInfo'.
> [CXX1405] error when building with cmake using /Users/vagrant/git/node_modules/react-native-screens/android/CMakeLists.txt: Build command failed.
  Error while executing java process with main class com.google.prefab.cli.AppKt with arguments {--build-system cmake --platform android --abi armeabi-v7a --os-version 24 --stl c++_shared --ndk-version 24 --output /Users/vagrant/git/node_modules/react-native-screens/android/.cxx/RelWithDebInfo/513j6x30/prefab/armeabi-v7a/prefab}
  
  Usage: prefab [OPTIONS] [PACKAGE_PATH]...
  
  Error: Missing argument "PACKAGE_PATH".

### Steps to reproduce

- Create a new project "npx react-native@0.68.7 init RNScreenTest --version 0.68.7"
- run yarn install
- Run project via release variant in android studio or via command line
- See error

ERROR : > Task :react-native-screens:configureCMakeRelWithDebInfo FAILED
C/C++: prefabUsage: prefab [OPTIONS] [PACKAGE_PATH]...
C/C++: prefabError: Missing argument "PACKAGE_PATH".

### Snack or a link to a repository

https://github.com/FraserKemp/test-react-native-screens

### Screens version

3.30.1

### React Native version

0.68.7

### Platforms

Android

### JavaScript runtime

JSC

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Release mode

### Device

None

### Acknowledgements

Yes

hau1412:
You too, i need fix version

hau1412:
I downgrade version [3.29.0], it work

FraserKemp:
I also downgraded for now 👍

tboba:
Hi @FraserKemp, thanks for submitting this issue! Let me investigate it 🔎

kumarGautamAramark:
I am also having same issue - 
React native - 0.70.6
react-native-screens - "^3.15.0"

tboba:
[...]

PR #2088, opened 2024-04-02
fix(Android): Add backward compatibility for prefabs

piaskowyk:
## Description

This PR fixes backward compatibility with React Native versions older than 0.71. The issue was caused by using prefabs in the screen transition animation feature. I added a check that makes this feature available since React Native 0.71

Tested against RN versions:
- [x] 0.68
- [x] 0.69
- [x] 0.70
- [x] 0.71
- [x] 0.73

Fixes https://github.com/software-mansion/react-native-screens/issues/2082

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -46,6 +46,38 @@ def reactNativeArchitectures() {
     return value ? value.split(",") : ["armeabi-v7a", "x86", "x86_64", "arm64-v8a"]
 }
 
+def safeAppExtGet(prop, fallback) {
+    def appProject = rootProject.allprojects.find { it.plugins.hasPlugin('com.android.application') }
+    appProject?.ext?.has(prop) ? appProject.ext.get(prop) : fallback
+}
+
+def resolveReactNativeDirectory() {
+    def reactNativeLocation = safeAppExtGet("REACT_NATIVE_NODE_MODULES_DIR", null)
+    if (reactNativeLocation != null) {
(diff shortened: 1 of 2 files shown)


android/build.gradle
```diff
@@ -146,11 +180,11 @@ repositories {
 
 dependencies {
     implementation 'com.facebook.react:react-native:+'
-    implementation 'androidx.appcompat:appcompat:1.5.0'
+    implementation 'androidx.appcompat:appcompat:1.4.2'
     implementation 'androidx.fragment:fragment:1.3.6'
     implementation 'androidx.coordinatorlayout:coordinatorlayout:1.2.0'
     implementation 'androidx.swiperefreshlayout:swiperefreshlayout:1.1.0'
-    implementation 'com.google.android.material:material:1.9.0'
+    implementation 'com.google.android.material:material:1.6.1'
```
tboba, line 187: This is needed, because of the error while building a project: `Can't determine type for tag '<macro name="m3_comp_assist_chip_container_shape">?attr/shapeAppearanceCornerSmall</macro>`. Since we're not using any features from Material 3, it's more important to patch this (at least for now 😛).

tboba (approved):
LGTM! 🎉

WoLewicki:
Do we still need this code if we only support RN >= 71? https://github.com/software-mansion/react-native-screens#supported-react-native-version. cc @piaskowyk maybe we can revert some of the code.

PR #2095, opened 2024-04-04
fix: add missing events

WoLewicki:
## Description

Add missing `ScreenStackHeaderConfig` events to fabric spec and remove the manual registration of events.

Also changed `Bubbling` events to `Direct` which fixes extraneous lifecycle events. It was to be done quite a while ago already: https://github.com/software-mansion/react-native-screens/pull/1308#discussion_r805773152


tboba (approved):
LGTM! 👍

PR #2105, opened 2024-04-14
fix(iOS): restore native behavior of auto shortening back button title

### LINKED ISSUE #1589: Header back button does not shorten to 'Back' on iOS if there's not enough space
### Description

According to the [documentation](https://reactnavigation.org/docs/native-stack-navigator#headerbacktitle), the back button title is supposed to change to "Back" if there's not enough space for the full title of the previous screen. That is not what I'm experiencing. In the GIF below you can see that instead of changing to "Back" it will truncate the header title, and if it's long enough, the back button will use the entire header width, hiding the title.

Title of back button is long enough to truncate the header title

Title of back button is long enough to remove the header title and back button icon

### Steps to reproduce

1. Create a screen with a long enough title
2. From the newly created screen, navigate to another screen with default screen options
3. Observe the long title of the back button

https://github.com/DrOverbuild/BackButtonRepro

### Screens version

3.17.0

### React Native version

0.70.0

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

iPhone 13 (iOS 15.5)

### Acknowledgements

Yes

zetavg:
## Description

Restore the iOS native behavior of automatically shorting the title of the header back button to "Back" if there is not enough space, which is documented [here](https://reactnavigation.org/docs/header-buttons#customizing-the-back-button)[^1] but does not behave as expected since v3.21.

Fixes #1589.

## Changes

* Assign to `backBarButtonItem` only if actual customizations of the back button are being made.

## Screenshots / GIFs

| Before | After |
|--------|------|
| Broken 1 | Working 1 |

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

[^1]: According to the document, 'On iOS this includes a label next to the button, which shows the title of the previous screen when the title fits in the available space, otherwise it says "Back".'

### DIFF
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -511,7 +511,7 @@ namespace react = facebook::react;
                                                                              action:nil];
   [backBarButtonItem setMenuHidden:config.disableBackButtonMenu];
 
-  auto isBackButtonCustomized = !isBackTitleBlank || config.disableBackButtonMenu || NO;
+  auto isBackButtonCustomized = !isBackTitleBlank || config.disableBackButtonMenu;
 
   if (config.isBackTitleVisible) {
     if ((config.backTitleFontFamily &&
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -511,8 +511,13 @@ namespace react = facebook::react;
                                                                              action:nil];
   [backBarButtonItem setMenuHidden:config.disableBackButtonMenu];
 
+  auto isBackButtonCustomized = !isBackTitleBlank || config.disableBackButtonMenu || NO;
+
   if (config.isBackTitleVisible) {
-    if (config.backTitleFontFamily || config.backTitleFontSize) {
+    if ((config.backTitleFontFamily &&
+         ![config.backTitleFontFamily isEqual:@"System"]) ||
+        config.backTitleFontSize) {
+      isBackButtonCustomized = YES;
       NSMutableDictionary *attrs = [NSMutableDictionary new];
       NSNumber *size = config.backTitleFontSize ?: @17;
       if (config.backTitleFontFamily) {
@@ -535,9 +540,17 @@ namespace react = facebook::react;
     // When backBarButtonItem's title is null, back menu will use value
     // of backButtonTitle
     [backBarButtonItem setTitle:nil];
+    isBackButtonCustomized = YES;
     prevItem.backButtonTitle = resolvedBackTitle;
   }
-  prevItem.backBarButtonItem = backBarButtonItem;
+
+  // Prevent unnecessary assignment of backBarButtonItem if it is not customized,
+  // as assigning one will override the native behavior of automatically shortening
+  // the title to "Back" or hide the back title if there's not enough space.
+  // See: https://github.com/software-mansion/react-native-screens/issues/1589
+  if (isBackButtonCustomized) {
+    prevItem.backBarButtonItem = backBarButtonItem;
+  }
 
   if (@available(iOS 11.0, *)) {
     if (config.largeTitle) {
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -515,6 +515,10 @@ namespace react = facebook::react;
 
   if (config.isBackTitleVisible) {
     if ((config.backTitleFontFamily &&
+         // While being used by react-navigation, the `backTitleFontFamily` will
+         // be set to "System" by default - which is the system default font.
+         // To avoid always considering the font as customized, we need to have an additional check.
+         // See: https://github.com/software-mansion/react-native-screens/pull/2105#discussion_r1565222738
          ![config.backTitleFontFamily isEqual:@"System"]) ||
         config.backTitleFontSize) {
       isBackButtonCustomized = YES;


ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -511,8 +511,13 @@ + (void)updateViewController:(UIViewController *)vc
                                                                              action:nil];
   [backBarButtonItem setMenuHidden:config.disableBackButtonMenu];
 
+  auto isBackButtonCustomized = !isBackTitleBlank || config.disableBackButtonMenu || NO;
+
   if (config.isBackTitleVisible) {
-    if (config.backTitleFontFamily || config.backTitleFontSize) {
+    if ((config.backTitleFontFamily &&
+         ![config.backTitleFontFamily isEqual:@"System"]) ||
```
tboba, line 518: Why do we need to check the font family here?

tboba (commented):
Hey @zetavg, thanks for this PR! ❤️ 
Just one small comment from me 😄

ios/RNSScreenStackHeaderConfig.mm
zetavg: It's because due to the current implementation of `@react-navigation/native-stack`, `backTitleFontFamily` will always be [assigned to the font family of `fonts.regular`](https://github.com/react-navigation/react-navigation/blob/ce6d885/packages/native-stack/src/views/HeaderConfig.tsx#L75-L80), which [will be "System" if the default theme is used](https://github.com/react-navigation/react-navigation/blob/ce6d88559e4a1afeafa84fc839892bb846349d67/packages/native/src/theming/fonts.tsx#L28-L31).

It seems that the "System" font family is already the default on iOS, as setting the font family to "System" does not change the appearance, but in this case, will make `config.backTitleFontFamily` always resolve to `YES`.

tboba: I see. Do you know, if we could rely on the native component that represents the title of back button, instead of comparing `config.backTitleFontFamily`?

kkafar: Thanks for the PR, this is great!

Can we have this comment in the code please?

tboba: @zetavg I'm wondering if it would be possible to get title component of back title and check if it has default font set, but I'm afraid this could lead to some flickers (setting the font and other properties -> checking the font of the native component -> setting system back 😄), so I guess we can stay with the current solution.

tboba (approved):
LGTM! Let's just wait for the comment 🚀

ios/RNSScreenStackHeaderConfig.mm
zetavg: Thanks for explaining, I think I got your point! I do think the current solution of comparing the font against `"System"` isn't perfect, one way is that there may be different ways to use the same system font (maybe `".AppleSystemUIFont"`? I'm not sure), and if `react-navigation` changes to also assign a default value to `config.backTitleFontSize` this will break again.

Your solution inspired me to think of a possible way to get this native behavior to work while still customizing the back button (such as changing the font or customizing the shortened "Back" text) - measure the width of the `headerTitle` and `headerRight` to see if there's enough space on layout, and replace the `backButtonTitle` with the shortened text or hide it. Well, flickering may be a problem.

zetavg:
Comment added! Please check if it's suitable 😃

tboba:
Yeah, I think for now it's a pass 🎉

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -511,8 +511,17 @@ + (void)updateViewController:(UIViewController *)vc
                                                                              action:nil];
   [backBarButtonItem setMenuHidden:config.disableBackButtonMenu];
 
+  auto isBackButtonCustomized = !isBackTitleBlank || config.disableBackButtonMenu || NO;
```
kkafar, line 514:
```suggestion
  auto isBackButtonCustomized = !isBackTitleBlank || config.disableBackButtonMenu;
```

This should give the same result

kkafar (changes requested):
Just one last thing

ios/RNSScreenStackHeaderConfig.mm
zetavg: Got it; thanks for pointing this out!

zetavg:
Hi @kkafar, just a ping to let you know I have applied your suggestion (thanks for that!) for the change you requested. If there's anything I can do to get this PR merged, please let me know.

kkafar (approved):
I think we're all good here.

I'll retrigger CI before merging.

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -535,9 +544,17 @@ + (void)updateViewController:(UIViewController *)vc
     // When backBarButtonItem's title is null, back menu will use value
     // of backButtonTitle
     [backBarButtonItem setTitle:nil];
+    isBackButtonCustomized = YES;
```
maciekstosio, line 547: Hi @zetavg, I know that this PR is already merged, but I wonder why do you set `isBackButtonCustomized` to `true` here.  My understanding is that this else block is for the case when we hide back button, but want to preserve it in back menu. So, by default when we have navigationItem hidden we don't display anything here - nothing is customized, when we want to custom hide it from back manu then we set `config.disableBackButtonMenu` which is in initial condition anyway.

zetavg: Quick reply - I think it's because this `isBackButtonCustomized` is used to determine whether we need to assign the `backBarButtonItem` to `prevItem.backBarButtonItem` later on line 556.

Since `backBarButtonItem` has been changed here (`[backBarButtonItem setTitle:nil];`), my understanding is that the changed `backBarButtonItem` has to be assigned to `prevItem.backBarButtonItem` for the changes to take effect. So it should be considered customized - or changed. Maybe it's better to rename `isBackButtonCustomized` to `isBackBarButtonItemChanged` so that it will be more clear (or use another way to track if we have changed anything on backBarButtonItem).

My guess is that if we do not set `isBackButtonCustomized` to true here, the back title won't be hidden as expected. I may need to verify it though.

zetavg: > Maybe it's better to rename `isBackButtonCustomized` to `isBackBarButtonItemChanged` ...

Oh, I see that you had already done this in #2123. 🎉

PR #2107, opened 2024-04-19
fix: don't use AnimatedNativeScreen when stackPresentation prop is not set

WoLewicki:
## Description

All navigators except `native-stack` does not set `stackPresentation` so it resolved in `AnimatedNativeModalScreen` making all navigators that are nested or under some views not work with pressability. Follow-up to https://github.com/software-mansion/react-native-screens/pull/2028/

## Changes

- Added `undefined` case for setting AnimatedNativeScreen

You can check `Test1214.tsx` or `Test2028.tsx` (after changing import to `@react-navigation/stack`) in order to test targets of touchables.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/components/Screen.tsx
+++ b/src/components/Screen.tsx
@@ -79,6 +79,7 @@ export class InnerScreen extends React.Component<ScreenProps> {
       // Due to how Yoga resolves layout, we need to have different components for modal nad non-modal screens
       const AnimatedScreen =
         Platform.OS === 'android' ||
+        stackPresentation === undefined ||
         stackPresentation === 'push' ||
         stackPresentation === 'containedModal' ||
         stackPresentation === 'containedTransparentModal'


tboba (approved):
Just tested it out on both architectures, LGTM! 🎉

roryabraham:
verified this fixed the problem in a release build of the Expensify app

src/components/Screen.tsx
```diff
@@ -79,6 +79,7 @@ export class InnerScreen extends React.Component<ScreenProps> {
       // Due to how Yoga resolves layout, we need to have different components for modal nad non-modal screens
       const AnimatedScreen =
         Platform.OS === 'android' ||
+        stackPresentation === undefined ||
```
satya164, line 82: Wouldn't it be better to do something like `stackPresentation === 'modal' ? AnimatedNativeModalScreen : AnimatedNativeScreen` - use modal when necessary and regular screen for rest, instead of specifying all the cases where it won't be a modal?

WoLewicki: This check will probably be updated when modals on `Android` and other types of modals on `iOS` by @kkafar will be merged. I'd leave it like that for now and come back to it when it's done but if you feel like it should be done here, then we can do it 🤷‍♂️

PR #2120, opened 2024-05-07
feat: remove currentBridge occurences

WoLewicki:
## Description

PR removing occurences of  `currentBridge` in the code since we should not use it (see https://github.com/software-mansion/react-native-reanimated/issues/5497#issuecomment-2083400038). Now in `RNSScreen.mm` we use the same hack as `ScrollView` (see the comment in the code). `RNSModule` now has the `setBridge` method so it gets the bridge on new arch in bridgeless too. As for `RNSScreenStackHeaderSubview`, I removed the call and changed it to `nil` since it did not work either way on new arch.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


WoLewicki:
We have to wait for https://github.com/facebook/react-native/pull/44474 to be released to properly handle all of this.

WoLewicki:
Nope, we should close it. I'll have to check what is the state of those things and if they are properly handled right now.

PR #2121, opened 2024-05-08
feat: proper snapshot when having multiple modals

WoLewicki:
## Description

WIP

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStack.mm
+++ b/ios/RNSScreenStack.mm
@@ -923,10 +923,6 @@ namespace react = facebook::react;
 {
   [self emitOnFinishTransitioningEvent];
   [RNSScreenWindowTraits updateWindowTraits];
-  // Because of the bug that caused view to have incorrect dimensions while changing the orientation,
-  // we need to signalize that it needs to be layouted.
-  // see https://github.com/software-mansion/react-native-screens/pull/1970
-  [_controller.view setNeedsLayout];
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


ios/RNSScreenStack.mm
```diff
@@ -923,10 +923,6 @@ - (void)navigationController:(UINavigationController *)navigationController
 {
   [self emitOnFinishTransitioningEvent];
   [RNSScreenWindowTraits updateWindowTraits];
-  // Because of the bug that caused view to have incorrect dimensions while changing the orientation,
-  // we need to signalize that it needs to be layouted.
-  // see https://github.com/software-mansion/react-native-screens/pull/1970
-  [_controller.view setNeedsLayout];
```
kkafar, line 929: Is this related & removed intentionally?

ios/RNSScreenStack.mm
```diff
@@ -1175,7 +1171,11 @@ - (void)unmountChildComponentView:(UIView<RCTComponentViewProtocol> *)childCompo
 
 - (void)takeSnapshot
 {
-  _snapshot = [_controller.visibleViewController.view snapshotViewAfterScreenUpdates:NO];
+  if (_presentedModals.count < 2) {
+    _snapshot = [_controller.visibleViewController.view snapshotViewAfterScreenUpdates:NO];
+  } else {
+    _snapshot = [[_presentedModals.lastObject view] snapshotViewAfterScreenUpdates:NO];
+  }
```
kkafar, line 1178: Yeah, I see. 
There, most likely, still be an issue when we dismiss a modal from different stack. I believe you can test this on `Test1829` (if not present in `FabricTestExample` it is in `TestsExample`).

kkafar (commented):
Notes

ios/RNSScreenStack.mm
WoLewicki: Nope, will revert it.

ios/RNSScreenStack.mm
WoLewicki: Yep, it does not work in such case, do we want to handle it right now?

kkafar (approved):
Answered your question

ios/RNSScreenStack.mm
kkafar: I think we won't be really able to come up with anything different that a heuristic solution before we do proper modal presenting logic refactor.

Let's proceed now, I'll create note on our task board so that this is not forgotten.

kkafar (approved):
🤦🏻‍♂️ Now, I've answered the question

PR #2126, opened 2024-05-10
refact(Android): convert example apps Groovy scripts to Kotlin

alduzy:
## Description

This PR intents to replace Groovy gradle scripts with Kotlin in the example apps.
Initially it was intended to convert root/android gradle scripts to Kotlin as well, but It didn't work out - changing library's build.gradle to Kotlin resulted in build errors and CI failures.

## Changes

- changed .gradle files extension to .gradle.kts
- converted scripts to Kotlin

- check if the example apps sync and build properly (clean project and/or invalidate caches if necessary)

## Checklist

- [x] Ensured that CI passes


tboba:
@alduzy hmmm, looks like Lint runner is failing, do you know what might be the cause while performing `yarn lint`?

alduzy:
There is currently a problem regarding `classpath("com.android.tools.build:gradle:4.2.2")` in android/build.gradle.kts as it syncs, builds and runs fine with Example, but requires min. 8.1.0 to work with FabricExample (which unfortunately does not work with Example).
Removing the dependency resolves the issue but causes lint stage of this PR to fail.
I attach the sync error below:
```
class com.android.build.gradle.internal.dsl.ExternalNativeBuildOptions_Decorated cannot be cast to class com.android.build.api.dsl.ExternalNativeBuildOptions (com.android.build.gradle.internal.dsl.ExternalNativeBuildOptions_Decorated is in unnamed module of loader org.gradle.internal.classloader.VisitableURLClassLoader$InstrumentingVisitableURLClassLoader @26772656; com.android.build.api.dsl.ExternalNativeBuildOptions is in unnamed module of loader org.gradle.internal.classloader.VisitableURLClassLoader$InstrumentingVisitableURLClassLoader @66e3ac0e)
```

Issue #2115, opened 2024-04-28
TypeError: undefined is not an object (evaluating '_$$_REQUIRE(_dependencyMap[11], "react-native-screens").ScreenStack')


kabundege:
### Error
Error: Requiring module "node_modules/react-native-screens/src/index.tsx", which threw an exception: Error: Exception in HostObject::get(propName:RNSModule): java.lang.UnsatisfiedLinkError: dalvik.system.PathClassLoader[DexPathList[[zip file "/data/app/com.app-MOOsrlTZxxpxhwDiZsIz2g==/base.apk"],nativeLibraryDirectories=[/data/app/com.app-MOOsrlTZxxpxhwDiZsIz2g==/lib/arm, /data/app/com.app-MOOsrlTZxxpxhwDiZsIz2g==/base.apk!/lib/armeabi-v7a, /system/lib, /product/lib]]] couldn't find "librnscreens.so"

### Another Error
TypeError: undefined is not an object (evaluating '_$$_REQUIRE(_dependencyMap[11], "react-native-screens").ScreenStack')
````
This error is located at:
    in NativeStackViewInner (created by NativeStackView)
    in RNCSafeAreaProvider (created by SafeAreaProvider)
    in SafeAreaProvider (created by SafeAreaProviderCompat)
    in SafeAreaProviderCompat (created by NativeStackView)
    in NativeStackView (created by NativeStackNavigator)
    in PreventRemoveProvider (created by NavigationContent)
    in NavigationContent
    in Unknown (created by NativeStackNavigator)
    in NativeStackNavigator (created by Navigation)
    in ErrorBoundary (created by ErrorBoundary)
    in ErrorBoundary (created by Navigation)
    in RNGestureHandlerRootView (created by GestureHandlerRootView)
    in GestureHandlerRootView (created by Navigation)
    in Navigation (created by OptimizedComponent)
    in OptimizedComponent
    in Unknown (created by OptimizedComponent)
    in OptimizedComponent
    in Unknown (created by App)
    in EnsureSingleNavigator
    in BaseNavigationContainer
    in ThemeProvider
    in NavigationContainerInner (created by App)
    in PersistGate (created by App)
    in Provider (created by App)
    in NetworkConnectivity (created by NetworkProvider)
    in NetworkProvider (created by App)
    in _QueryClientProvider (created by App)
    in _QueryClientProvider (created by _PersistQueryClientProvider)
    in _PersistQueryClientProvider (created by App)
[...]

PR #2128, opened 2024-05-10
Catch UnsatisfiedLinkError

piaskowyk:
## Description

Fixes https://github.com/software-mansion/react-native-screens/issues/2115

The `System.loadLibrary("rnscreens")` can throw an exception that inherits from the `Error` class instead of the `Exception` class.


kkafar (commented):
Makes sense.

My only worry is whether this is only type of exception that can be thrown here. Maybe going with more generic error type could be worth considering

piaskowyk:
According to [the Android documentation](https://developer.android.com/reference/java/lang/System#loadLibrary(java.lang.String)) in our case only [UnsatisfiedLinkError](https://developer.android.com/reference/java/lang/UnsatisfiedLinkError) exception can be thrown.

kkafar (approved):
This convinces me

PR #2129, opened 2024-05-12
fix(iOS): onNativeDismissCancelled called too early during modal dismissal

zetavg:
## Description

Currently, on iOS, `onNativeDismissCancelled` is being called while the user begins to drag down a modal (which calls the callback function of the `usePreventRemove` hook, if `preventRemove` is `true`).

This is not a common behavior of iOS apps and can sometimes be annoying: the callback will be triggered while the user is scrolling around a `ScrollView` in the modal and has hit the top, which is interrupting.

| Ideal | Before | After |
|-------|--------|-------|
| <a href="https://github.com/software-mansion/react-native-screens/assets/3784687/167828c8-eb3a-4b4b-99a7-b9eb7ec55f06">Ideal</a> | <a href="https://github.com/software-mansion/react-native-screens/assets/3784687/6c166757-8c73-4e90-9236-9faf84514a14">Before</a> | <a href="https://github.com/software-mansion/react-native-screens/assets/3784687/7993f36c-3455-4b23-8a28-f46956655f8a">After</a> |
| The event fires when the user **releses** the drag. If the user did not intend to dismiss the modal (the user did not drag it down far enough or dragged it back to its original position before release), the event will not fire. | The event fires as soon as the modal is dragged down, regardless of whether the user intends to dismiss the modal or not. (In this case, the user just wants to scroll to the top.) | Same as "Ideal". |

This PR adjusts this to the more ideal behavior.

## Changes

Move the call of `notifyDismissCancelled` from [`presentationControllerShouldDismiss`](https://developer.apple.com/documentation/uikit/uiadaptivepresentationcontrollerdelegate/3229890-presentationcontrollershoulddism) to [`presentationControllerDidAttemptToDismiss`](https://developer.apple.com/documentation/uikit/uiadaptivepresentationcontrollerdelegate/3229888-presentationcontrollerdidattempt) - which "Notifies the delegate that a user-initiated attempt to dismiss a view was prevented", and has the same platform avalibality as `presentationControllerShouldDismiss`.

Use the new "Prevent Remove" example:

### Full test recordings

**Modal**

https://github.com/software-mansion/react-native-screens/assets/3784687/16777817-3949-413c-83ad-f0f7136af158

**Normal Screen** (this should not be affected since I believe that preventing going back from a normal screen is not handled here)

https://github.com/software-mansion/react-native-screens/assets/3784687/41f25247-eb54-4ce8-9348-690d0a925e8f

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


zetavg:
Note: both #1959 and this PR can achieve #1950. But I think this should be the default behavior of `onNativeDismissCancelled` and `usePreventRemove` - firing the callback when the user releases the drag.

kkafar:
Hi @zetavg, thanks for preparing these changes :tada:. I took a glance & these changes look good, however I'll need to find time to test this! Will get back you once I do.

kkafar:
@kkafar Giving myself a notification to take care of this PR somewhere next week.

WoLewicki (approved):
LGTM!

kkafar (approved):
I finally got to test the changes, sorry for such delay. I've merged main & merged this implementation with recent changes from https://github.com/software-mansion/react-native-screens/pull/2184

> [!note]
We should deprecate emitting `gestureCancel` in `presentationControllerDidAttemptToDismiss` in favour of emitting only `dismissCancelled` to align fully with react-navigation v7 API. *Created ticket for this*.

Thank you for your contribution @zetavg!

kkafar:
Will merge once the CI passes.

PR #2130, opened 2024-05-13
fix: Prevent setting RTL mode in test examples

tboba:
## Description

This PR is a follow-up to #2084 that fixes setting incorrect RTL mode in our example apps.
The reason why it is being forced is that `I18nManager.forceRTL(true)` is being called from the global scope, which causes to force RTL, even if we're only importing the component **from file**.

## Changes

- Moved `I18nManager.forceRTL(true);` to inner scope of App components
- Added `I18nManager.forceRTL(false);` in App.js files to prevent propagating RTL mode to other tests.

Just run TestsExample & FabricTestExample ¯\_(ツ)_/¯

## Checklist

- [X] Ensured that CI passes

### DIFF
--- a/FabricTestExample/App.js
+++ b/FabricTestExample/App.js
@@ -1,5 +1,6 @@
 /* eslint-disable no-unused-vars */
 import React from 'react';
+import { I18nManager } from 'react-native';
 
 import { enableFreeze } from 'react-native-screens';
 
@@ -97,6 +98,10 @@ import Test2008 from './src/Test2008';
 import Test2028 from './src/Test2028';
 import Test2048 from './src/Test2048';
 import Test2069 from './src/Test2069';
+
+// Prevent RTL tests from propagating direction to other tests.
+// Remember to reload the app after switching from the test with RTL mode.
+I18nManager.forceRTL(false);
 
 enableFreeze(true);
 
--- a/FabricTestExample/src/Test654.js
+++ b/FabricTestExample/src/Test654.js
@@ -5,9 +5,9 @@ import { createNativeStackNavigator } from '@react-navigation/native-stack';
 
 const Stack = createNativeStackNavigator();
 
-I18nManager.forceRTL(true);
-
 export default function App() {
+  I18nManager.forceRTL(true);
+
   return (
     <NavigationContainer>
       <Stack.Navigator>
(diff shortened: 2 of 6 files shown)


FabricTestExample/App.js
```diff
@@ -98,6 +99,10 @@ import Test2028 from './src/Test2028';
 import Test2048 from './src/Test2048';
 import Test2069 from './src/Test2069';
 
+// Prevent RTL tests from propagating direction to other tests.
+// Remember to reload the app after switching from the test with RTL mode.
+I18nManager.forceRTL(false);
```
maciekstosio, line 104: I wonder if it wouldn't be better to revert back on component unmount instead of global, what do you think?

maciekstosio (approved):
One idea open for discussion, but in general looks good!

FabricTestExample/App.js
tboba: Yeah, great idea! Resolved in https://github.com/software-mansion/react-native-screens/pull/2130/commits/791b21be203fb8e5801c6d6f0013776c66b00179.

FabricTestExample/src/Test654.js
```diff
@@ -5,9 +5,15 @@ import { createNativeStackNavigator } from '@react-navigation/native-stack';
...
-I18nManager.forceRTL(true);
-
 export default function App() {
+  I18nManager.forceRTL(true);
+
+  React.useEffect(() => {
+    return () => {
+      I18nManager.forceRTL(false);
+    };
+  });
```
kkafar, line 15: Shouldn't it be put inside useEffect?

kkafar (commented):
Just single question?

FabricTestExample/src/Test654.js
tboba: Moved in https://github.com/software-mansion/react-native-screens/pull/2130/commits/0d3f60080b1e931ccc0d417442ea1f833d3e039e.

maciekstosio: Now you could convert it into a hook to simplify the code :D

tboba: Makes no sense just for two tests to be honest, but let's remember about it when we'll add more tests 👍

PR #2132, opened 2024-05-13
refactor: searchbar rewritten as functional component

alduzy:
## Description

This PR intents to change SearchBar component into a functional component.

## Changes

Rewritten SearchBar component as function leveraging useImperativeHandle to apply new methods

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/src/components/SearchBar.tsx
+++ b/src/components/SearchBar.tsx
@@ -18,18 +18,18 @@ export const NativeSearchBar: React.ComponentType<SearchBarProps> &
 export const NativeSearchBarCommands: SearchBarCommandsType =
   SearchBarNativeCommands as SearchBarCommandsType;
 
-type ViewRef = React.ElementRef<typeof NativeSearchBar>;
+type NativeSearchBarRef = React.ElementRef<typeof NativeSearchBar>;
 
 type SearchBarCommandsType = {
-  blur: (viewRef: ViewRef) => void;
-  focus: (viewRef: ViewRef) => void;
-  clearText: (viewRef: ViewRef) => void;
-  toggleCancelButton: (viewRef: ViewRef, flag: boolean) => void;
-  setText: (viewRef: ViewRef, text: string) => void;
-  cancelSearch: (viewRef: ViewRef) => void;
+  blur: (viewRef: NativeSearchBarRef) => void;
+  focus: (viewRef: NativeSearchBarRef) => void;
+  clearText: (viewRef: NativeSearchBarRef) => void;
+  toggleCancelButton: (viewRef: NativeSearchBarRef, flag: boolean) => void;
+  setText: (viewRef: NativeSearchBarRef, text: string) => void;
+  cancelSearch: (viewRef: NativeSearchBarRef) => void;
 };
 
-function ImplSearchBar(props: SearchBarProps, ref: Ref<SearchBarCommands>) {
+function SearchBar(props: SearchBarProps, ref: Ref<SearchBarCommands>) {
   const implRef = React.useRef<SearchBarCommands | null>(null);
 
   React.useImperativeHandle(ref, () => ({
@@ -80,5 +80,5 @@ function ImplSearchBar(props: SearchBarProps, ref: Ref<SearchBarCommands>) {
 }
 
 export default React.forwardRef<SearchBarCommands, SearchBarProps>(
-  ImplSearchBar
+  SearchBar
 );
--- a/src/components/SearchBar.tsx
+++ b/src/components/SearchBar.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode } from 'react';
+import React, { ReactNode, Ref } from 'react';
 import {
   isSearchBarAvailableForCurrentPlatform,
   SearchBarCommands,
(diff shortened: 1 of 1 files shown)


src/components/SearchBar.tsx
```diff
@@ -31,61 +33,65 @@ type SearchBarCommandsType = {
   cancelSearch: (viewRef: React.ElementRef<typeof NativeSearchBar>) => void;
 };
 
-class SearchBar extends React.Component<SearchBarProps> {
-  nativeSearchBarRef: React.RefObject<SearchBarCommands>;
+const SearchBar = React.forwardRef<SearchBarCommands, SearchBarProps>(
+  function SearchBar(props, ref) {
+    const nativeSearchBarRef = React.useRef<SearchBarCommands | null>(null!);
+    React.useImperativeHandle(ref, () => ({
```
kkafar, line 39: Nitpick: I wonder whether we can create `SearchBarInner` and then just write 

```ts
const SearchBar = React.forwardRef<SearchBarCommands, SearchBarProps>(SearchBarInner);
```

to avoid whole level of indentation?

kkafar: We could also do it for `Screen` in separate PR.

kkafar: Taking into consideration that there already exists `InnerScreen` we could use `Impl` suffix or something 😅

src/components/SearchBar.tsx
```diff
@@ -12,9 +12,11 @@ import SearchBarNativeComponent, {
 } from '../fabric/SearchBarNativeComponent';
 
 export const NativeSearchBar: React.ComponentType<SearchBarProps> &
-  typeof NativeSearchBarCommands = SearchBarNativeComponent as any;
+  typeof NativeSearchBarCommands =
+  SearchBarNativeComponent as unknown as React.ComponentType<SearchBarProps> &
+    SearchBarCommandsType;
 export const NativeSearchBarCommands: SearchBarCommandsType =
-  SearchBarNativeCommands as any;
+  SearchBarNativeCommands as SearchBarCommandsType;
```
kkafar, line 19: These types do scare me, I guess it works, but we need to cleanup this mess. Let's proceed now, we will do it in separate iteration on this file.

src/components/SearchBar.tsx
```diff
@@ -31,61 +33,65 @@ type SearchBarCommandsType = {
   cancelSearch: (viewRef: React.ElementRef<typeof NativeSearchBar>) => void;
 };
 
-class SearchBar extends React.Component<SearchBarProps> {
-  nativeSearchBarRef: React.RefObject<SearchBarCommands>;
+const SearchBar = React.forwardRef<SearchBarCommands, SearchBarProps>(
+  function SearchBar(props, ref) {
+    const nativeSearchBarRef = React.useRef<SearchBarCommands | null>(null!);
+    React.useImperativeHandle(ref, () => ({
+      ...nativeSearchBarRef.current,
```
kkafar, line 40: Why do we write `...nativeSearchBarRef.current` here? My guess was to expose usual `View` ref methods, however this is not reflected in the reference type so user would get an error any way. Is there different reason?

src/components/SearchBar.tsx
```diff
@@ -31,61 +33,65 @@ type SearchBarCommandsType = {
...
+    function setText(text: string) {
+      _callMethodWithRef(ref => NativeSearchBarCommands.setText(ref, text));
+    }
 
-  cancelSearch() {
-    this._callMethodWithRef(ref => NativeSearchBarCommands.cancelSearch(ref));
-  }
+    function cancelSearch() {
+      _callMethodWithRef(ref => NativeSearchBarCommands.cancelSearch(ref));
+    }
```
kkafar, line 84: Won't these get recreated on every re-render? I think we should move the declaration to imperative handle or (if that makes sense in this use case) use callback.

kkafar (changes requested):
Good job overall. I have few reflections & points to discuss, left them below.

src/components/SearchBar.tsx
alduzy: I thought it's a must have for useImperativehandle, but It seems to work fine without it

src/components/SearchBar.tsx
```diff
@@ -12,80 +12,73 @@ import SearchBarNativeComponent, {
...
 export const NativeSearchBar: React.ComponentType<SearchBarProps> &
-  typeof NativeSearchBarCommands = SearchBarNativeComponent as any;
+  typeof NativeSearchBarCommands =
+  SearchBarNativeComponent as unknown as React.ComponentType<SearchBarProps> &
+    SearchBarCommandsType;
 export const NativeSearchBarCommands: SearchBarCommandsType =
-  SearchBarNativeCommands as any;
+  SearchBarNativeCommands as SearchBarCommandsType;
+
+type ViewRef = React.ElementRef<typeof NativeSearchBar>;
```
kkafar, line 21:
```suggestion
type NativeSearchBarRef = React.ElementRef<typeof NativeSearchBar>;
```

Let's use more descriptive name, but this alias is very good idea.

I'm [nitpicking](https://dictionary.cambridge.org/pl/dictionary/english/nitpick) here hard, because our attention won't be back to this component for some time most likely 😄

src/components/SearchBar.tsx
```diff
@@ -12,80 +12,73 @@ import SearchBarNativeComponent, {
...
 };
 
-class SearchBar extends React.Component<SearchBarProps> {
-  nativeSearchBarRef: React.RefObject<SearchBarCommands>;
-
-  constructor(props: SearchBarProps) {
-    super(props);
-    this.nativeSearchBarRef = React.createRef();
-  }
+function ImplSearchBar(props: SearchBarProps, ref: Ref<SearchBarCommands>) {
```
kkafar, line 32: In my nitpicks I will insist here on using suffix instead of prefix.
 
```suggestion
function SearchBarImpl(props: SearchBarProps, ref: Ref<SearchBarCommands>) {
```

src/components/SearchBar.tsx
```diff
@@ -12,80 +12,73 @@ import SearchBarNativeComponent, {
...
+      if (ref) {
+        method(ref);
+      } else {
+        console.warn(
+          'Reference to native search bar component has not been updated yet'
+        );
+      }
+    },
+    [implRef]
+  );
```
kkafar, line 70: Yeah this looks good now.

kkafar (changes requested):
I'm [nitpicking](https://dictionary.cambridge.org/pl/dictionary/english/nitpick) here hard, because our attention won't be back to this component for some time most likely.

Have you tested this on our example apps (there is dedicated screen in Example & FabricExample for the search bar API I believe) whether everything works fine?

src/components/SearchBar.tsx
tboba: Let's use `React.Ref` instead of pure `Ref`.

tboba: Also, one thing about this line - should we go for `function` declaration, or a `const` as an arrow function? I saw we're going for an arrow function (e.g. in Screen.tsx component), but I'm leaving this into discussion 🤔

tboba: I don't think `ImplSearchBar` would be good name. How about just `SearchBar`?

src/components/SearchBar.tsx
```diff
@@ -12,80 +12,73 @@ import SearchBarNativeComponent, {
...
 
-class SearchBar extends React.Component<SearchBarProps> {
-  nativeSearchBarRef: React.RefObject<SearchBarCommands>;
-
-  constructor(props: SearchBarProps) {
-    super(props);
-    this.nativeSearchBarRef = React.createRef();
-  }
+function ImplSearchBar(props: SearchBarProps, ref: Ref<SearchBarCommands>) {
+  const implRef = React.useRef<SearchBarCommands | null>(null);
```
tboba, line 33: `searchBarRef` / `commandsRef`?

tboba (changes requested):
Good stuff! Just couple notes from me

src/components/SearchBar.tsx
tboba: @alduzy I saw your commit about the name change 😄 but what do you think about it?

alduzy: Good idea, especially since the export doesn't need name, previously the outer function declaration was SearchBar

alduzy: I am going to refactor the Screen once again after this PR so that the code matches - there is an 'outer' arrow function and the 'inner' one is declared as here. I would leave it like this and apply changes to the Screen component

kkafar (commented):
The code looks 🟢 

Have you tested the search bar API after these changes?

src/components/SearchBar.tsx
alduzy: [615eddf](https://github.com/software-mansion/react-native-screens/pull/2132/commits/615eddf4a654e4d97c8f99733ba962ae3d0b39af)

alduzy: https://github.com/software-mansion/react-native-screens/pull/2132/commits/351a18702deb8873ad48cd9cec3686c29c26741c

src/components/SearchBar.tsx
alduzy: https://github.com/software-mansion/react-native-screens/pull/2132/commits/351a18702deb8873ad48cd9cec3686c29c26741c

src/components/SearchBar.tsx
alduzy: https://github.com/software-mansion/react-native-screens/pull/2132/commits/351a18702deb8873ad48cd9cec3686c29c26741c

src/components/SearchBar.tsx
alduzy: https://github.com/software-mansion/react-native-screens/pull/2132/commits/33cc656967c6ab70262008ec3f79ac2c9ad1a550

src/components/SearchBar.tsx
alduzy: https://github.com/software-mansion/react-native-screens/pull/2132/commits/33cc656967c6ab70262008ec3f79ac2c9ad1a550

alduzy: https://github.com/software-mansion/react-native-screens/pull/2132/commits/6dcfb1841b98a3babbd19d679164b6cfea00fb9c

src/components/SearchBar.tsx
alduzy: https://github.com/software-mansion/react-native-screens/pull/2132/commits/6dcfb1841b98a3babbd19d679164b6cfea00fb9c

alduzy:
I've tested the changes on Example app both on android and iOS and it works just like before. I didn't find any usage of the searchbar in FabricExample.

kkafar (approved):
Looks like we're good to go then. Thank you 🎉

PR #2133, opened 2024-05-14
chore: bump Podfile.lock for example and test apps

maciekstosio:
## Description

Bump RNScreens version in Podfile.lock from 3.31.0-rc.1  to 3.31.1 in test and example apps.


kkafar (commented):
I think we re fine here. Let's open this PR

tboba (commented):
Have you checked if TVOSExample also needs to be updated?

maciekstosio:
> Have you checked if TVOSExample also needs to be updated?

I did run pod install there, but it doesn't seem to help a lot.

kkafar (approved):
🟢

PR #2142, opened 2024-05-17
fix: error when building paper after fabric

maciekstosio:
## Description

Currently when you build Fabric app and then Paper app Android Studio has errors about duplicated classes. This should resolve that issue. 

**Test:**
Packed with `npm pack` and tested on new expo app. Seems that everything is fine.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -152,7 +152,6 @@ android {
             } else {
                 srcDirs += [
                     "src/paper/java",
-                    "build/generated/source/codegen/java"
                 ]
             }
         }


WoLewicki (approved):
LGTM

kkafar (commented):
Yeah I think we are good here. 

Before merging I would recommend creating fresh RN application, installing `@react-navigation/native-stack`, `react-native-safe-area-context` & our library from sourcecode & build Paper Android.
I'm wary here after recent release issues.

maciekstosio:
> Yeah I think we are good here.
> 
> Before merging I would recommend creating fresh RN application, installing `@react-navigation/native-stack`, `react-native-safe-area-context` & our library from sourcecode & build Paper Android. I'm wary here after recent release issues.

Yeah, I packed the lib w `npm pack` and linked it to new expo app with all of those you mentioned. Everything seems to be working.

kkafar (approved):
Then we are good to go

tboba (approved):

PR #2146, opened 2024-05-18
Prevent iOS host object installation from getting the wrong bridge

gabrieldonadel:
## Description

Similar to https://github.com/software-mansion/react-native-reanimated/pull/5953,  using `[RCTBridge currentBridge]` may lead to the wrong bridge in apps that have multiple React instances, as is the case for expo-dev-client. This results in apps sometimes crashing when reloading the bundle after interacting with the DevMenu.

image

## Changes

Instead of using `[RCTBridge currentBridge]` we can just call `self.bridge` and ensure the correct bridge is always used

### DIFF
--- a/ios/RNSModule.mm
+++ b/ios/RNSModule.mm
@@ -138,8 +138,7 @@ RCT_EXPORT_MODULE()
    because depending on the selected architecture, only one method will be called.
    For `Paper`, it will be constantsToExport, and for `Fabric`, it will be getTurboModule.
 */
-  RCTBridge *bridge = [RCTBridge currentBridge];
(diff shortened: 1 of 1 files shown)


kkafar:
Hey @gabrieldonadel, thanks for noticing this and taking your time to fix this. 
I'll be asking @WoLewicki to take a look at these changes as he is well-versed with bridgeless mode.

WoLewicki (approved):
Hey, I made a PR changing it some time ago: https://github.com/software-mansion/react-native-screens/pull/2120 but it is stuck with some problems in core. I think we can merge that one to fix at least one of the issues already 🚀

PR #2149, opened 2024-05-20
feat: extract shared example and use in in both Example and FabricExa…

maciekstosio:
## Description

Currently we have separate apps for Example and FabricExample, Example covers most of the features and FabricExample only HeaderConfig, I think it would be better for to share the code base. The main reason is that it's easier to test if features fork the same on both platforms. 

FabricExample should look the same, as Example app.


kkafar (commented):
Let's wait for this one:

* https://github.com/software-mansion/react-native-screens/pull/2135

to be merged, and let's then update so that we are in sync. 

Besides that I think we're good.

kkafar (approved):
Provided that you have tested this on both architectures I think we're good to go.

tboba (approved):
LGTM aswell!

PR #2152, opened 2024-05-21
feat(Android): added navigationBarTranslucent option

### LINKED ISSUE #2122: Inconsistency with navigationBarColor Transparency on react-native-screens versions >= 3.30.0"
### Description

Before version `3.30.0`, when the `navigationBarColor` was set to transparent on a Stack navigator, the navigation bar would preserve transparency, adopting the color defined on SafeAreaView. However, in version `3.30.0` and onwards, this behavior has changed: the navigation bar no longer maintains transparency but instead defaults to a white color (or black if `navigationBarColor` is not explicitly defined).

### Steps to reproduce

Setting a stack navigator with `navigationBarColor` set as transparent:
```js

....

<Stack.Navigator
      screenOptions={{
        ...
        navigationBarColor: "transparent",
      }}
    >
...
</Stack.Navigator>
```

https://github.com/software-mansion/react-native-screens

### Screens version

`>= 3.30.0`

### React Native version

0.74.1

### Platforms

Android

### JavaScript runtime

None

### Workflow

Expo managed workflow

### Architecture

None

### Build type

None

### Device

Android emulator

### Device model

Pixel_3a_API_34

### Acknowledgements

Yes

alduzy:
## Description

Once upon a time there was an [issue](https://github.com/software-mansion/react-native-screens/issues/1719) regarding navigation bar. The problem was that the background color set to the navigation bar covered the content behind, see examples:

<table>
<tr>
<th>navigationBarColor: 'transparent'</th>
<th>navigationBarColor: 'yellow'</th>
</tr>
<tr>
<td>

Screenshot 2024-05-22 at 10 33 28

</td>
<td>

Screenshot 2024-05-22 at 10 34 08

</td>
</tr>
</table>

Then [PR 1988](https://github.com/software-mansion/react-native-screens/pull/1988) solved the issue, but introduced another problem: sometimes it may be intended to have content behind the navigation bar.

<table>
<tr>
<th>navigationBarColor: 'transparent' after PR 1988</th>
<th>navigationBarColor: 'transparent' before PR 1988</th>
</tr>
<tr>
<td>

Screenshot 2024-05-21 at 13 00 23

</td>
<td>

Screenshot 2024-05-21 at 13 07 19

</td>
</tr>
</table>

This PR intents to add a new screen option, enabling navigation bar translucency control, so that the developers can decide if they want to show content behind the navigation bar independently.

Fixes #2122 #1719

## Changes

- added navigationBarTranslucent option

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
@@ -197,6 +197,8 @@ object ScreenWindowTraits {
         val screenForNavBarTranslucent = findScreenForTrait(screen, WindowTraits.NAVIGATION_BAR_TRANSLUCENT)
         val translucent = screenForNavBarTranslucent?.isNavigationBarTranslucent ?: false
 
+        // Controls whether the decor view should fit root-level content views for insets based on translucent value:
+        // https://developer.android.com/reference/androidx/core/view/WindowCompat#setDecorFitsSystemWindows(android.view.Window,boolean)
         WindowCompat.setDecorFitsSystemWindows(window, !translucent)
     }
 
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
@@ -8,7 +8,6 @@ import android.app.Activity
 import android.content.pm.ActivityInfo
 import android.graphics.Color
 import android.os.Build
-import android.util.Log
 import android.view.ViewParent
 import androidx.core.graphics.Insets
 import androidx.core.view.ViewCompat
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
@@ -8,9 +8,11 @@ import android.app.Activity
 import android.content.pm.ActivityInfo
 import android.graphics.Color
 import android.os.Build
+import android.util.Log
 import android.view.ViewParent
 import androidx.core.graphics.Insets
 import androidx.core.view.ViewCompat
+import androidx.core.view.WindowCompat
 import androidx.core.view.WindowInsetsCompat
 import androidx.core.view.WindowInsetsControllerCompat
 import com.facebook.react.bridge.GuardedRunnable
@@ -185,6 +187,19 @@ object ScreenWindowTraits {
         window.navigationBarColor = color
     }
 
+    internal fun setNavigationBarTranslucent(screen: Screen, activity: Activity?) {
+        if (activity == null) {
+            return
+        }
+
+        val window = activity.window
+
+        val screenForNavBarTranslucent = findScreenForTrait(screen, WindowTraits.NAVIGATION_BAR_TRANSLUCENT)
+        val translucent = screenForNavBarTranslucent?.isNavigationBarTranslucent ?: false
+
+        WindowCompat.setDecorFitsSystemWindows(window, !translucent)
+    }
+
     internal fun setNavigationBarHidden(screen: Screen, activity: Activity?) {
         if (activity == null) {
             return
@@ -221,6 +236,7 @@ object ScreenWindowTraits {
         }
         if (didSetNavigationBarAppearance) {
             setNavigationBarColor(screen, activity)
+            setNavigationBarTranslucent(screen, activity)
             setNavigationBarHidden(screen, activity)
         }
     }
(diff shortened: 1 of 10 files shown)


kkafar (commented):
Hey, few initial remarks.

Why is this a problem since `3.30.1`, why was the code removed that was responsible for this behaviour.

If we go with new prop, we do not restore backward compatibility and we commit to introducing breaking change in behaviour.

I want to see some description in PR why this is the way to go instead of restoring old behaviour and keeping backward compatibility.

Another issue is that we want to have similar behaviour on both platforms. How does this look like on iOS, does the problem occur there? How do we plan to handle this prop there?

alduzy:
@kkafar Right before 3.30 release [this  commit](https://github.com/software-mansion/react-native-screens/commit/c2b68ba675781fcfa658ecbc7b9bf6cc0a3497a4) changed the navigation bar behavior.
As far as I know `WindowCompat.setDecorFitsSystemWindows()` caused problems with screen layout and/or jumps in some cases, so it's not a good idea to add the logic back in all cases when navigation bar is not hidden.
In certain situations users may want to achieve such appearance and I think it's best to enable it through a separate option.
@tboba feel free to share your thoughts.

kkafar:
I think we need some elaboration on reasons & whys of the situation in the PR's description.

This commit you refer to #1988 went under my radar & I'm not convinced it was good change.

By adopting the suggested approach we would commit to a small, but still breaking change in behaviour & I'm not convinced we do need this. 

So basically I want you to understand & advocate for why #1988 was needed / or it was not, and then we can continue here.

There is one more level to this change, how does this behave in context of edge-to-edge and not edge-to-edge applications?

Also my questions related to iOS are still open. 

I'm picky today in my reviews 😅

android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
```diff
@@ -185,6 +187,19 @@ object ScreenWindowTraits {
...
+        if (activity == null) {
+            return
+        }
+
+        val window = activity.window
+
+        val screenForNavBarTranslucent = findScreenForTrait(screen, WindowTraits.NAVIGATION_BAR_TRANSLUCENT)
+        val translucent = screenForNavBarTranslucent?.isNavigationBarTranslucent ?: false
+
+        WindowCompat.setDecorFitsSystemWindows(window, !translucent)
```
WoLewicki, line 200: I'd add a comment here of what this code does exactly.

WoLewicki (commented):
The code looks good overall, but we need to have it properly documented in the PR description why is it needed, so the users know how to correctly use it.

android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
WoLewicki: I meant something like It makes the content go under the navigationBar 😅 Does it make the content go under status bar also e.g. when there is no header displayed?

alduzy: https://github.com/software-mansion/react-native-screens/pull/2152/commits/950c8e34ba76f5a73fbf8a54690454de8b0b81ae
I updated the comment, let me know if it's clear now

tboba (approved):
That looks better, LGTM! Just one thing before merging 😄

android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
WoLewicki: Looks good now. What about the second part: `Does it make the content go under status bar also e.g. when there is no header displayed?`?

android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
```diff
@@ -8,9 +8,11 @@ import android.app.Activity
 import android.content.pm.ActivityInfo
 import android.graphics.Color
 import android.os.Build
+import android.util.Log
```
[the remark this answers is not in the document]
alduzy: https://github.com/software-mansion/react-native-screens/pull/2152/commits/5faba0aee298e6cb8d4baf4f9be3e2981612364f

kkafar (commented):
These changes look good, thank you! 🎉 

There is one more thing, however, namely have you tested what happens on iOS (especially Paper, but check Fabric too) when somebody uses this prop? Won't we get any undesired behaviour?

kkafar (approved):
Provided that iOS behaves nicely I think we're good to go

PR #2153, opened 2024-05-21
chore: Implement missing codegen interfaces 

maciekstosio:
## Description

Currently when we change native props autogenerated by codegen we need to manually update changed files. We're going to automate so everything is copied automatically, but as a first step we need to implement those that aren't yet. This is what this PR introduces. 

Go through Example app, nothing should change.


FabricExample/yarn.lock
kkafar, line 1: Can we remove lock file changes from the PR?

android/src/main/java/com/swmansion/rnscreens/SearchBarManager.kt
```diff
@@ -71,45 +84,33 @@ class SearchBarManager : ViewGroupManager<SearchBarView>() {
...
-            "focus" -> root.handleFocusJsRequest()
-            "blur" -> root.handleBlurJsRequest()
-            "clearText" -> root.handleClearTextJsRequest()
-            "toggleCancelButton" -> root.handleToggleCancelButtonJsRequest(false) // just a dummy argument
-            "setText" -> root.handleSetTextJsRequest(args?.getString(0))
-            "cancelSearch" -> root.handleCancelSearchJsRequest()
-            else -> throw JSApplicationIllegalArgumentException("Unsupported native command received: $commandId")
-        }
-    }
+    override fun setPlacement(view: SearchBarView, placeholder: String?) = Unit
```
kkafar, line 113: It seems this one is also iOS only 
```suggestion
    override fun setPlacement(view: SearchBarView, placeholder: String?) {
        logNotAvailable("placement")
    }
```

Idk whats better, truncating implementation down to noop or logging Android error (JS user won't see it anyway). We can discuss it shortly on weekly.

Also I'm not sure whether we can even trigger this function to be called when setting this prop from JS as it is not annotated with `@ReactProp` 😅 TODO: Test whether setting the non-supported prop in JS would trigger this function anyway. If not - we should remove all the "logNotAvailable" implementations (in separate PR).

android/src/main/java/com/swmansion/rnscreens/SearchBarManager.kt
```diff
@@ -71,45 +84,33 @@ class SearchBarManager : ViewGroupManager<SearchBarView>() {
...
+    override fun setHintTextColor(view: SearchBarView, color: Int?) {
         view.hintTextColor = color
     }
 
     @ReactProp(name = "shouldShowHintSearchIcon")
-    fun setShouldShowHintSearchIcon(view: SearchBarView, shouldShowHintSearchIcon: Boolean?) {
-        view.shouldShowHintSearchIcon = shouldShowHintSearchIcon ?: true
+    override fun setShouldShowHintSearchIcon(view: SearchBarView, value: Boolean) {
+        view.shouldShowHintSearchIcon = value ?: true
     }
```
kkafar, line 111:
```suggestion
    override fun setShouldShowHintSearchIcon(view: SearchBarView, shouldShowHintSearchIcon: Boolean) {
        view.shouldShowHintSearchIcon = shouldShowHintSearchIcon ?: true
    }
```

If we can, we should stay with more expressive parameter names here.

android/src/main/java/com/swmansion/rnscreens/SearchBarManager.kt
```diff
@@ -48,17 +61,17 @@ class SearchBarManager : ViewGroupManager<SearchBarView>() {
...
+    override fun setBarTintColor(view: SearchBarView, color: Int?) {
         view.tintColor = color
     }
 
     @ReactProp(name = "disableBackButtonOverride")
-    fun setDisableBackButtonOverride(view: SearchBarView, disableBackButtonOverride: Boolean?) {
-        view.shouldOverrideBackButton = disableBackButtonOverride != true
+    override fun setDisableBackButtonOverride(view: SearchBarView, value: Boolean) {
+        view.shouldOverrideBackButton = value != true
     }
```
kkafar, line 71:
```suggestion
    override fun setDisableBackButtonOverride(view: SearchBarView, disableBackButtonOverride: Boolean) {
        view.shouldOverrideBackButton = disableBackButtonOverride != true
    }
```

Same here

kkafar: Especially that it does not remain unused.

android/src/main/java/com/swmansion/rnscreens/SearchBarManager.kt
```diff
@@ -131,4 +132,56 @@ class SearchBarManager : ViewGroupManager<SearchBarView>() {
...
+        view?.handleToggleCancelButtonJsRequest(flag)
+    }
+
+    override fun setText(view: SearchBarView?, text: String?) {
+        view?.handleSetTextJsRequest(text)
+    }
+
+    override fun cancelSearch(view: SearchBarView?) {
+        view?.handleFocusJsRequest()
+    }
```
kkafar, line 164: Let's make sure that this approach with utilising generated `receiveCommand` implementation will still work on earlier React Native versions. Allegedly we support back to 0.68 but testing on 0.71 will be fine.

android/src/main/java/com/swmansion/rnscreens/SearchBarManager.kt
```diff
@@ -131,4 +132,56 @@ class SearchBarManager : ViewGroupManager<SearchBarView>() {
...
+        logNotAvailable("hideNavigationBar")
+    }
+
+    override fun setCancelButtonText(view: SearchBarView?, value: String?) {
+        logNotAvailable("cancelButtonText")
+    }
+
+    override fun setTintColor(view: SearchBarView?, value: Int?) {
+        logNotAvailable("tintColor")
+    }
```
kkafar, line 186: Same remarks as to `setPlacement` prop.

kkafar (commented):
Few initial remarks

android/src/main/java/com/swmansion/rnscreens/SearchBarManager.kt
maciekstosio: Changed in 3472960cb38ea20a6a963c47ec1fc76f0d8e4fd7

android/src/main/java/com/swmansion/rnscreens/SearchBarManager.kt
maciekstosio: Yeah, I added to todo and will change that in all places in separate PR.

android/src/main/java/com/swmansion/rnscreens/SearchBarManager.kt
maciekstosio: I tested in on clean project with RN 0.71.0, imperative api seems to be working fine.

android/src/main/java/com/swmansion/rnscreens/SearchBarManager.kt
maciekstosio: Yeah, actually those are logged 
image

kkafar:
I've sent some notes on private message 😅

maciekstosio:
Closing in favor of https://github.com/software-mansion/react-native-screens/pull/2163

Issue #2158, opened 2024-05-24
Importing RNScreens in Xcode iOS Swift file fails to compile


Brett-Best:
### Description

Attempting to `import RNScreens` in a swift file inside the main Xcode project where you'd define modules for example cause the build to fail.

I've attached a link to a new project I created using Expo.
The only customisation I made was to make CocoaPods use frameworks linking them statically.

The errors I'm getting are:
```shell
❌  (/Users/brettbest/Library/Developer/Xcode/DerivedData/Reproducer-favipxwwmbqszdfdwlhacyleizcj/Build/Products/Debug-iphonesimulator/RNScreens/RNScreens.framework/Headers/RNScreensTurboModule.h:1:10)

> 1 | #include <array>
    |          ^ 'array' file not found
  2 | #include <jsi/jsi.h>
  3 | 
  4 | #define JSI_HOST_FUNCTION(NAME)   \


❌  (/Users/brettbest/Library/Developer/Xcode/DerivedData/Reproducer-favipxwwmbqszdfdwlhacyleizcj/Build/Products/Debug-iphonesimulator/React-jsi/jsi.framework/Headers/decorator.h:10:10)

   8 | #pragma once
   9 | 
> 10 | #include <tuple>
     |          ^ 'tuple' file not found
  11 | 
  12 | #include <jsi/instrumentation.h>
  13 | #include <jsi/jsi.h>


❌  (ios/Reproducer/noop-file.swift:6:8)

  4 | //
  5 | 
> 6 | import RNScreens
    |        ^ could not build Objective-C module 'RNScreens'
  7 | 


❌  (/Users/brettbest/Library/Developer/Xcode/DerivedData/Reproducer-favipxwwmbqszdfdwlhacyleizcj/Build/Products/Debug-iphonesimulator/RNScreens/RNScreens.framework/Headers/RNScreensTurboModule.h:1:10)

> 1 | #include <array>
    |          ^ 'array' file not found
  2 | #include <jsi/jsi.h>
  3 | 
  4 | #define JSI_HOST_FUNCTION(NAME)   \


❌  (/Users/brettbest/Library/Developer/Xcode/DerivedData/Reproducer-favipxwwmbqszdfdwlhacyleizcj/Build/Products/Debug-iphonesimulator/React-jsi/jsi.framework/Headers/decorator.h:10:10)

   8 | #pragma once
   9 | 
> 10 | #include <tuple>
     |          ^ 'tuple' file not found
  11 | 
[...]

PR #2160, opened 2024-05-28
Fix: Importing of project into Swift projects

Brett-Best:
## Description

Fixes https://github.com/software-mansion/react-native-screens/issues/2158 by making C++ headers project instead of public.
Original code sample (in the issue) can be updated to include this change verifying the fix.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/RNScreens.podspec
+++ b/RNScreens.podspec
@@ -63,6 +63,7 @@ Pod::Spec.new do |s|
   s.platforms    = { :ios => platform, :tvos => "11.0", :visionos => "1.0" }
   s.source       = { :git => "https://github.com/software-mansion/react-native-screens.git", :tag => "#{s.version}" }
   s.source_files = source_files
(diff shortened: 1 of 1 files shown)


kkafar:
Hi @Brett-Best, thanks for taking your time to introduce these changes 🎉 We will test these changes in `reanimated` context, because they are main consumers of this API and if it goes smoothly we can proceed with merging. FYI next release should be coming next week (in positive scenario).

Brett-Best:
Thanks, it might be worth including `import RNScreens` in an example project or something that can be ran on CI to prevent breaking importing the library in the future.

kkafar (approved):
It seems we're fine on front of reanimated interop & also seems that it works well in runtime. Thanks!

PR #2168, opened 2024-06-03
feat: Add gradle task to automatically copy the codegen artifacts for paper

maciekstosio:
## Description

When changing native props on Fabric, codegen generates corresponding interfaces and delegates. To make sure both implementations are consistent, we implement those interfaces on Paper too. Currently, after generating interfaces using codegen, developer needs to copy corresponding files for paper manually. This task adds Gradle task, that automates this.

## Changes

Add new task to build Gradle and necessary properties: 
- codegen artifacts dir and paper dir 
- flag in both fabric apps that indicates that copying should be performed (we do want this task to be performed only when developing the library)

Remove `textColor` from `src/fabric/SearchBarNativeComponent.ts` and run ` ./gradlew generateCodegenArtifactsFromSchema` in `/react-native-screens/FabricExample/android`. That should automatically copy regenerated files to paper directory.


FabricExample/android/gradle.properties
```diff
@@ -39,3 +39,5 @@ newArchEnabled=true
 # Use this property to enable or disable the Hermes JS engine.
 # If set to false, you will be using JSC instead.
 hermesEnabled=true
+
+isScreensExampleApp=true
```
kkafar, line 43: We should add this flag to all of our example apps, so that it means what the name suggests.

android/build.gradle
```diff
@@ -192,3 +192,58 @@ dependencies {
         }
     }
 }
+
+def isScreensExampleApp() {
+    return project.hasProperty('isScreensExampleApp') && project.property('isScreensExampleApp')
```
kkafar, line 197: Also, doesn't `project.property` return a string? We might want to compare it to `"true"` or something.

android/build.gradle
```diff
@@ -192,3 +192,58 @@ dependencies {
...
+}
+
+task copyCodegenArtifacts(type: Copy) {
+    group 'After build tasks'
+    description 'Tasks which copy codegen artifacts to paper architecture'
+    
+    if (!isScreensExampleApp()) {
```
kkafar, line 204: And add check here for new architecture

```suggestion
    if (!isScreensExampleApp() || !isNewArchitectureEnabled()) {
```

android/build.gradle
```diff
@@ -192,3 +192,58 @@ dependencies {
...
+        return
+    }
+
+    dependsOn tasks.generateCodegenArtifactsFromSchema
```
kkafar, line 208: Isn't this one present only when running Fabric build? Won't we crash here on Paper?

android/build.gradle
```diff
@@ -192,3 +192,58 @@ dependencies {
...
+
+    def absoluteCodegenArtifactsPaperDestination = "${project.rootDir}/../../${project.property('codegenArtifactsPaperDestination')}"
+    def absoluteCodegenArtifactsSource = "${project.rootDir}/../../${project.property('codegenArtifactsSource')}"
```
kkafar: These lines assume that `codegenArtifactsPaperDestination` & `codegenArtifactsSource` properties do exist. I think it would be better to extract these two to separate functions and handle missing values gracefully e.g. with nice error messages.

android/build.gradle
```diff
@@ -192,3 +192,58 @@ dependencies {
...
+    }
+
+    generatedFiles.forEach { generatedFile -> 
+        if (!existingFilesMap.containsKey(generatedFile.name)) {
+            println generatedFile.name + " not found in paper dir"
+        }
+    }
+
+    if (existingFiles.size() <= 0) {
+        throw new Exception('Paper destination with codegen interfaces is empty.')
```
kkafar:
```suggestion
        throw new Exception('[RNScreens] Paper destination with codegen interfaces is empty.')
```

Nitpick

android/build.gradle
```diff
@@ -192,3 +192,58 @@ dependencies {
...
+    }
+
+    if (existingFiles.size() > generatedFiles.size()) {
+        throw new Exception('Number od generated artifacts should be greather then or equal to paper interfaces.')
```
kkafar:
```suggestion
        throw new Exception('[RNScreens] Number od generated artifacts should be greather then or equal to paper interfaces.')
```

nitpick

android/build.gradle
```diff
@@ -192,3 +192,58 @@ dependencies {
...
+
+    generatedFiles.forEach { generatedFile -> 
+        if (!existingFilesMap.containsKey(generatedFile.name)) {
+            println generatedFile.name + " not found in paper dir"
+        }
+    }
+
+    if (existingFiles.size() <= 0) {
+        throw new Exception('Paper destination with codegen interfaces is empty.')
+    }
```
kkafar: Why is this an error? Also it would be nice if error messages suggested an action to take (another nitpick)

android/build.gradle
```diff
@@ -192,3 +192,58 @@ dependencies {
...
+
+    def existingFilesMap = [:]
+
+    existingFiles.forEach { existingFile ->
+        existingFilesMap[existingFile.name] = 1
+    }
+
+    generatedFiles.forEach { generatedFile -> 
+        if (!existingFilesMap.containsKey(generatedFile.name)) {
+            println generatedFile.name + " not found in paper dir"
```
kkafar: Also I do not remember, but it would be worth checking whether there is a `warn` function of something, it seems that it would be appropriate to use it here.

kkafar (changes requested):
Looks good, thanks!

Left some remarks below

android/build.gradle
kkafar: And we should add a comment in code that they're defined in gradle.properties file or smth

android/build.gradle
maciekstosio: When I used project.property in project that do not have this I got an exception that property does not exist

android/build.gradle
maciekstosio: Tried logger.warn but it behaves the same. So I stick to println.

android/build.gradle
maciekstosio: I changed it to warning, cause this may indicate that the paper destination folder is incorrect. Changed in https://github.com/software-mansion/react-native-screens/pull/2168/commits/bd65c0c7bb6d6a12facbf79e40c1a04b89aebb4f

android/build.gradle
maciekstosio: Changed in https://github.com/software-mansion/react-native-screens/pull/2168/commits/bd65c0c7bb6d6a12facbf79e40c1a04b89aebb4f

android/build.gradle
maciekstosio: Changed in https://github.com/software-mansion/react-native-screens/pull/2168/commits/bd65c0c7bb6d6a12facbf79e40c1a04b89aebb4f

android/build.gradle
maciekstosio: First thing changed in https://github.com/software-mansion/react-native-screens/pull/2168/commits/bd65c0c7bb6d6a12facbf79e40c1a04b89aebb4f

android/build.gradle
maciekstosio: Shouldn't after checking newArchVariable from above, right?

android/build.gradle
maciekstosio: Changed in https://github.com/software-mansion/react-native-screens/pull/2168/commits/bd65c0c7bb6d6a12facbf79e40c1a04b89aebb4f

FabricExample/android/gradle.properties
maciekstosio: Changed in https://github.com/software-mansion/react-native-screens/pull/2168/commits/bd65c0c7bb6d6a12facbf79e40c1a04b89aebb4f

android/build.gradle
maciekstosio: Comment added in https://github.com/software-mansion/react-native-screens/pull/2168/commits/340de960ef878893869aa6f4c33bb9a0fdf730ac

android/build.gradle
kkafar: Yeah yeah, but my point is that `project.property('isScreensExampleApp')` does return a `String` (I think), thus no matter what string is present there, the whole expression will be resolved to true.

android/build.gradle
```diff
@@ -192,3 +192,79 @@ dependencies {
...
+        }
+
+        generatedFiles.forEach { generatedFile -> 
+            if (!existingFilesMap.containsKey(generatedFile.name)) {
+                println("[RNScreens] ${generatedFile.name} not found in paper dir, if it's used in Android you need to copy it manually and implement yourself before using auto-copy feature")
+            }
+        }
+
+        if (existingFiles.size() == 0) {
+            throw new Exception("[RNScreens] Paper destination with codegen interfaces is empty. This might be okay if you don't have any interfaces/delegates used in Android, if that's not the case please check if codegenArtifactsPaperDestination in android/gradle.properties is correct")
```
kkafar, line 253: If there are cases when this is ok, we should print a warning instead of throwing an exception.

android/build.gradle
```diff
@@ -192,3 +192,79 @@ dependencies {
...
+            from absoluteCodegenArtifactsSource
+            include existingFiles.collect { it.name }
+            into absoluteCodegenArtifactsPaperDestination
+        }
+    }
+}
+
+if (isScreensExampleApp() && isNewArchitectureEnabled()) {
+    tasks.generateCodegenArtifactsFromSchema.finalizedBy('copyCodegenArtifacts')
+}
```
kkafar, line 270:
```suggestion
}

```

Example/android/gradle.properties
```diff
@@ -49,3 +49,5 @@ hermesEnabled=true
 # This is necessary, because the code checking for multiple instances
 # detects package versions from other example applications in the repository.
 disableMultipleInstancesCheck=true
+
+isScreensExampleApp=true
```
kkafar, line 53:
```suggestion
isScreensExampleApp=true

```

TVOSExample/android/gradle.properties
```diff
@@ -26,3 +26,5 @@ android.enableJetifier=true
 
 # Version of flipper SDK to use with React Native
 FLIPPER_VERSION=0.75.1
+
+isScreensExampleApp=true
```
kkafar, line 30:
```suggestion
isScreensExampleApp=true

```

TestsExample/android/gradle.properties
```diff
@@ -41,3 +41,5 @@ newArchEnabled=false
 hermesEnabled=true
 
 disableMultipleInstancesCheck=true
+
+isScreensExampleApp=true
```
kkafar, line 45:
```suggestion
isScreensExampleApp=true

```

android/gradle.properties
```diff
@@ -21,3 +21,9 @@ android.useAndroidX=true
 android.enableJetifier=true
 
 kotlin.code.style=official
+
+# Path to codegen artifacts, used for task (copyCodegenArtifacts) that automates copying them to paper destination
+codegenArtifactsSource=android/build/generated/source/codegen/java/com/facebook/react/viewmanagers
+# Destination to peper path where codegen interfaces should stored and then implemented, so we have consistency in both architectures.
+# Used for task (copyCodegenArtifacts) that automates copying those interfaces/delegates when codegen is run
+codegenArtifactsPaperDestination=android/src/paper/java/com/facebook/react/viewmanagers/
```
kkafar, line 29:
```suggestion
codegenArtifactsPaperDestination=android/src/paper/java/com/facebook/react/viewmanagers/

```

kkafar (commented):
Answered you question, final notes & whitespaces.

android/build.gradle
maciekstosio: fixed in https://github.com/software-mansion/react-native-screens/pull/2168/commits/9a2a66e36f40345e800cb005bd69ce1827054051

android/build.gradle
kkafar: I doubt it behaves the same. It maybe looks the same, but it has different log level associated - WARN - instead of QUIET, I believe, for `println`.

Is is important? Definitely no. If you want to really stay with current approach it's ok.

android/build.gradle
kkafar: At the time of writing this comment there was no check 😅

android/gradle.properties
```diff
@@ -21,3 +21,9 @@ android.useAndroidX=true
 android.enableJetifier=true
 
 kotlin.code.style=official
+
+# Path to codegen artifacts, used for task (copyCodegenArtifacts) that automates copying them to paper destination
```
kkafar, line 25:
```suggestion
# Path to codegen output directory with this library view manager's interfaces & delegates. Used by `copyCodegenArtifacts` task that helps to synchronise newly generated files with their Paper conterparts.
```

android/gradle.properties
```diff
@@ -21,3 +21,9 @@ android.useAndroidX=true
...
+codegenArtifactsSource=android/build/generated/source/codegen/java/com/facebook/react/viewmanagers
```
kkafar, line 26:
```suggestion
codegenArtifactsSource=android/build/generated/source/codegen/java/com/facebook/react/viewmanagers

```

android/gradle.properties
```diff
@@ -21,3 +21,9 @@ android.useAndroidX=true
...
+# Destination to peper path where codegen interfaces should stored and then implemented, so we have consistency in both architectures.
```
kkafar, line 27:
```suggestion
# Path to directory with view manager's interfaces & delegates used while running on Paper architecture. This property is used as output path for `copyCodegenArtifacts` task.
```

kkafar (commented):
Some polishing and I think we're good to go

android/build.gradle
maciekstosio: You're right. Changed in https://github.com/software-mansion/react-native-screens/pull/2168/commits/5afd502c2b94a6035ac128860340695feaffc4e6

kkafar (approved):
I think we can proceed

PR #2169, opened 2024-06-04
fix(Android, Fabric): jumping content with native header

kkafar:
## Description

This PR intents to solve the *eternal* issue with "jumping content" on Android & Fabric. The issue itself is present on every platform / RN architecture combination, however this PR scope is to solve situation only on Android + Fabric. Android + Paper, and iOS will be solved in separate PRs.

> [!note]
> These videos are recorded with `topInsetEnabled: false`, as this prop implementation causes another series of issues that will be handled separately.

Here is before & after comparison (best way to see is to go frame-by-frame)

| Before | After |
|--------|--------|
| <video width="320" height="240" controls src="https://github.com/software-mansion/react-native-screens/assets/50801299/e1e995b5-885b-4bd4-941a-57cdc6b321b2"></video> | <video width="320" height="240" controls src="https://github.com/software-mansion/react-native-screens/assets/50801299/6ca87888-0f05-4dfc-b6db-bfd08e3735b3"></video> |

This will even work with irregular font sizes!

### Short issue genesis

> [!note]
> The flow described here below is a simplification, but should give you a good grasp on the issue.

Basically during the very first Yoga layout, that happens on secondary thread, React layout mechanism has no knowledge of the header size (there isn't even a node representing the header at appropriate tree-level present in shadow tree), thus the `Screen` content is layouted with more available space that it has in reality. These dimensions are then send to UI thread, and propagated bottom-up (children before parents) and `Screen` contents do receive too high frame. Then, when `ScreenContainer` / `ScreenStack` does receive its frame from RN, it triggers a fragmentary pass of native layout using `CoordinatorLayout` (the layout stops at `Screen`), offsetting the `Screen` by just-computed-header-height on Y axis, and in consequence pushing some of the `Screen`'s contents off the screen (exactly a strip of header height). The situation is then salvaged by sending state update from `Screen` to React Native with new frame size. 

### Implemented solution

During the first Yoga layout pass (when there is not state from UI thread received yet) we utilise the fact that RN clones our ShadowNode & calls `adapt` on it. In the adapt method we call into JVM where we have set up a dummy view hierarchy with coordinator layout & header, we layout it & return result to C++ layer where we set bottom padding on the Screen so that its contents do have right amount of space. 

> [!important]
> Downside of this solution is the fact, that the Yoga state / Shadow Tree still indicates that the `Screen`'s origin is at `(x, y) = (0, 0)` and it still will have wrong dimensions.
Setting dummy dimension on `HeaderConfig` shadow node will improve situation only slightly, as the `Screen` will still have wrong origin, but it will have appropriate size immediately, **hence `Screen`'s state update might not trigger follow-up transaction**. Thus I'm thinking now that I will update the solution.

### Yet ~un~tested approaches

* ~I want to try making custom descriptor for `ScreenStack`, and try to customise shadownode's layout method.~ <- tested this & I believe this will not work due to the fact, that `ShadowNode.layout` does not use `layoutContext.viewportOffset` at all (so we can not use this to offset our descendants). At the same time the `layout` method does not propagate layout metrics - they are extracted for each shadow node directly from it's yoga node and this process does not take into consideration parent's layout metrics. **However, if the `x, y` view origin coordinates determined by yoga are in parent node coordinate system** we can use `HeaderConfig` to ensure appropriate `Screens` size and at the same time set `frame.y` manually!

I've added `TestHeader` test case. It's best to run it with `FabricTestExample`, record it, and then see frame-by-frame that the content no longer jumps.

## Checklist

- [ ] Ensured that CI passes


FabricTestExample/android/settings.gradle
```diff
@@ -2,3 +2,12 @@ rootProject.name = 'FabricTestExample'
...
 includeBuild('../node_modules/@react-native/gradle-plugin')
+
+includeBuild('../node_modules/react-native') {
+    dependencySubstitution {
+        substitute(module("com.facebook.react:react-android")).using(project(":packages:react-native:ReactAndroid"))
+        substitute(module("com.facebook.react:react-native")).using(project(":packages:react-native:ReactAndroid"))
+        substitute(module("com.facebook.react:hermes-android")).using(project(":packages:react-native:ReactAndroid:hermes-engine"))
+        substitute(module("com.facebook.react:hermes-engine")).using(project(":packages:react-native:ReactAndroid:hermes-engine"))
+    }
+}
```
kkafar, line 13: Don't mind this, I'll remove it just before merging, as if there would be any necessity to introduce changes I want to have access to RN code in my Android Studio.

android/src/main/java/com/swmansion/rnscreens/RNScreensPackage.kt
```diff
@@ -51,4 +95,114 @@ class RNScreensPackage : TurboReactPackage() {
...
+        const val FONT_SIZE_UNSET = -1
+
+        // We access this field from C++ layer, through `getInstance` method.
+        // We don't care what instance we get access to as long as it has initialized
+        // dummy view hierarchy.
+        private var WEAK_THIS = WeakReference<RNScreensPackage>(null)
+
+        @JvmStatic
+        fun getInstance(): RNScreensPackage? {
+            return WEAK_THIS.get()
```
kkafar, line 204: I'm not sure of how JNI will work here. 

We return the reference here, assuming it is not null, we get it in C++, but who retains it? We hold only a weak ref to self, so if we're dropped by our owner (react package registry) the GC can come in and collect the object while C++ code is running? 

Not sure here on GC thread / the-thread-this-code-is-running-on interaction / execution of C++ code. 

Does the JNI guarantee validity of the returned reference we get in C++ layer for the scope of whole C++ function? My worries are that we do not use "JNI function" there, we just call into JVM from regular C++ code.

Anyway, *I do not think we should care here*, as **WEAK_THIS** lifecycle is tied directly to the application (react context) lifecycle, thus in any not *abnormal* scenario we should be completely fine.

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.kt
```diff
@@ -163,7 +163,7 @@ class ScreenStackHeaderConfig(context: Context) : ViewGroup(context) {
             screenFragment?.setToolbar(toolbar)
         }
 
-        if (isTopInsetEnabled) {
+        if (isTopInsetEnabled && false) {
```
kkafar, line 166: This is modified just for testing, will remove it just before merging.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
tboba, line 1: This file seems to be formatted by mistake. Revert?

android/src/main/jni/CMakeLists.txt
tboba, line 1: Hmm, revert?

WoLewicki:
> Downside of this solution is the fact, that the Yoga state / Shadow Tree still indicates that the Screen's origin is at (x, y) = (0, 0) and it still will have wrong dimensions.
Setting dummy dimension on HeaderConfig shadow node will improve situation only slightly, as the Screen will still have wrong origin, but it will have appropriate size immediately, hence Screen's state update might not trigger follow-up transaction. Thus I'm thinking now that I will update the solution.

So is there a way to tell `yoga` that the screen is offset by the header size so it has `origin` value and, at the same time, layout it natively without applying that additional offset since it is done by system itself?

kkafar:
Only thing, that I've not tried yet, that comes to my mind is attempt to directly manipulate `yogaNode_` in adapt method and set `marginTop` during tree configuration pass, as I have once tried to do on iOS.

However please note that we do not need the Screen to have correct origin in ShadowTree and I was told that neither does Reanimated, so this inadequacy is fine for now.

common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderConfigState.h
tboba, line 1: do we need to keep these C++ implementaions of ScreenStackHeaderConfig?

src/fabric/ScreenStackHeaderConfigNativeComponent.ts
```diff
@@ -50,5 +50,7 @@ export interface NativeProps extends ViewProps {
 
 export default codegenNativeComponent<NativeProps>(
   'RNSScreenStackHeaderConfig',
-  {}
+  {
+    interfaceOnly: true,
+  }
```
tboba, line 55: Why do we need this?

tboba (changes requested):
🔥🔥

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
```diff
@@ -26,13 +41,103 @@ class RNSScreenComponentDescriptor final
...
+      if (std::strcmp(
+              child->getComponentName(), "RNSScreenStackHeaderConfig") == 0) {
+        return {std::ref(child)};
+      }
+    }
+    return std::nullopt;
+  }
+
+#ifdef ANDROID
+  std::optional<float> findHeaderHeight(const int fontSize) const {
```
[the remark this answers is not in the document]
WoLewicki: Or maybe the calls between cpp and java are not so expensive and we should not care about it?

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
```diff
@@ -26,13 +41,103 @@ class RNSScreenComponentDescriptor final
             shadowNode.getState());
     auto stateData = state->getData();
 
+#ifdef ANDROID
     if (stateData.frameSize.width != 0 && stateData.frameSize.height != 0) {
       layoutableShadowNode.setSize(
```
WoLewicki, line 46: Could we have some comparison if the pre-calculated values were same as the ones in `stateData` and e.g. log it somehow if they weren't? That is a very valuable information that we are calculating something wrong and it needs a fix ASAP.

android/src/main/java/com/swmansion/rnscreens/RNScreensPackage.kt
```diff
@@ -51,4 +104,120 @@ class RNScreensPackage : TurboReactPackage() {
...
+    }
+
+    /**
+     * Triggers layout pass on dummy view hierarchy, taking into consideration selected
+     * ScreenStackHeaderConfig props that might have impact on final header height.
+     *
+     * @param fontSize font size value as passed from JS
+     * @return header height in dp as consumed by Yoga
+     */
+    private fun computeDummyLayout(fontSize: Int): Float {
```
WoLewicki, line 164: I think we could move the implementation of those methods to some utility file and only leave what's 100% needed here to keep the code cleaner.

android/src/main/java/com/swmansion/rnscreens/RNScreensPackage.kt
```diff
@@ -51,4 +104,120 @@ class RNScreensPackage : TurboReactPackage() {
...
+                CoordinatorLayout.LayoutParams.MATCH_PARENT,
+                CoordinatorLayout.LayoutParams.MATCH_PARENT
+            )
+        }
+
+        coordinatorLayout.apply {
+            addView(appBarLayout)
+            addView(dummyContentView)
+        }
+    }
```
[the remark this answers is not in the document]
kkafar: I must say I don't get exactly your point, cause I don't know what "render" & fragment you mean exactly, however from the gist of it:

1. there is really no fragment here,
2. we don't want to tie `Screen` instance to some object that lives throughout whole application lifecycle,
3. how do we know which screen to access when we're requested from C++ layer for header height? I guess we could have some kind of registry of all screens and lookup by tag but it seems bad,
4. we're calling the layout code from C++ layer from completely different thread (non main) -> we should not trigger any behaviour on NativeTree with possibly ongoing actions such as another native layout;
5. there are no views in native tree when we call the layout code for first time.

So as long as I agree that this setup is kinda artificial and can go out of sync with what's being created in `ScreenStackHeaderConfig` we can handle this by creating some kind of shared factory method that will be used in both places <- I've thought of this and I'll suggest waiting with this for the moment I have also prepared fix for Paper, but we should definitely do this.

WoLewicki (commented):
I left some comments, please answer them. Hopefully the approach is good enough and we will finally solve the problem :tada:

src/fabric/ScreenStackHeaderConfigNativeComponent.ts
kkafar: So that we can have custom descriptor for header config, otherwise (simplifying here a bit) RN will use the autogenerated one.

android/src/main/java/com/swmansion/rnscreens/RNScreensPackage.kt
kkafar: https://github.com/software-mansion/react-native-screens/pull/2169/commits/d3da35727ffaba855814cd6a8556087961c1a128

android/src/main/java/com/swmansion/rnscreens/RNScreensPackage.kt
```diff
@@ -1,21 +1,73 @@
...
+import com.google.android.material.appbar.AppBarLayout
+import java.lang.ref.WeakReference
+
+private class HeaderHeightCacheEntry(val cacheKey: Int, val headerHeight: Float) {
+    fun hasKey(key: Int) = cacheKey == key
+
+    companion object {
+        val EMPTY = HeaderHeightCacheEntry(Int.MIN_VALUE, 0f)
+    }
+}
```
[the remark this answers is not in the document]
kkafar: https://github.com/software-mansion/react-native-screens/pull/2169/commits/d3da35727ffaba855814cd6a8556087961c1a128

android/src/main/java/com/swmansion/rnscreens/RNScreensPackage.kt
```diff
@@ -1,21 +1,73 @@
...
 class RNScreensPackage : TurboReactPackage() {
-    override fun createViewManagers(reactContext: ReactApplicationContext) =
-        listOf<ViewManager<*, *>>(
+    // The state required to compute header dimensions. We want this on instance rather than on class
+    // for context access & being tied to instance lifetime.
+    private lateinit var coordinatorLayout: CoordinatorLayout
+    private lateinit var appBarLayout: AppBarLayout
+    private lateinit var dummyContentView: View
+    private lateinit var toolbar: Toolbar
+    private var defaultFontSize: Float = 0f
```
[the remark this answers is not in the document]
kkafar: I moved it to completely separate class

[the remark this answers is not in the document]
kkafar: https://github.com/software-mansion/react-native-screens/pull/2169/commits/d3da35727ffaba855814cd6a8556087961c1a128

android/src/main/java/com/swmansion/rnscreens/RNScreensPackage.kt
```diff
@@ -51,4 +104,120 @@ class RNScreensPackage : TurboReactPackage() {
...
+        coordinatorLayout = CoordinatorLayout(contextWithTheme)
+
+        appBarLayout = AppBarLayout(contextWithTheme).apply {
+            layoutParams = CoordinatorLayout.LayoutParams(
+                CoordinatorLayout.LayoutParams.MATCH_PARENT,
+                CoordinatorLayout.LayoutParams.WRAP_CONTENT,
+            )
+        }
+
+        toolbar = Toolbar(contextWithTheme).apply {
```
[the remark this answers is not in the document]
kkafar: Yeah, when we add additional top bar types we will need to adjust this logic for sure.
Right now though I'm willing to push this & let people test to see whether there are some fundamental issues with current approach.

common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderConfigShadowNode.cpp
```diff
@@ -0,0 +1,19 @@
...
+#include <react/renderer/mapbuffer/MapBuffer.h>
+#include <react/renderer/mapbuffer/MapBufferBuilder.h>
+#endif
+
+namespace facebook::react {
+
+const char RNSScreenStackHeaderConfigComponentName[] =
+    "RNSScreenStackHeaderConfig";
+
+} // namespace facebook::react
```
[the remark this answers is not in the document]
kkafar: Removed this file completely.

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.h
```diff
@@ -4,6 +4,8 @@
 #include <react/renderer/components/rnscreens/EventEmitters.h>
 #include <react/renderer/components/rnscreens/Props.h>
 #include <react/renderer/components/view/ConcreteViewShadowNode.h>
+#include <react/renderer/core/LayoutContext.h>
```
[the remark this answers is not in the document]
kkafar: I've been overriding `layout` method at some point, as it is done right now with https://github.com/software-mansion/react-native-screens/pull/2169/commits/e90f57d8fec72385035f680b6940febcf294b682. So the import stays for now 😅

apps/test-examples/src/TestHeader.tsx
```diff
@@ -0,0 +1,95 @@
...
+  navigation: NativeStackNavigationProp<ParamListBase>;
+};
+
+const Stack = createNativeStackNavigator();
+
+export default function App(): JSX.Element {
+  return (
+    <NavigationContainer>
+      <Stack.Navigator screenOptions={{
+        headerTopInsetEnabled: false,
```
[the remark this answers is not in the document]
kkafar: Was trying to get this prop working, but I didn't realise that `react-navigation` does not expose this prop in public API and you can not set it manually... Cleaned up in https://github.com/software-mansion/react-native-screens/pull/2169/commits/2c42d655c5e158c3e5825b67490ec171a05ef9e2.

common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderConfigState.h
kkafar: In the moment your comment was made - yes, we did. However now, I've undertaken completely different approach to the problem so I've removed these files related to `HeaderConfig` completely.

android/src/main/java/com/swmansion/rnscreens/RNScreensPackage.kt
```diff
@@ -1,21 +1,73 @@
...
+    }
+
+    override fun createViewManagers(reactContext: ReactApplicationContext): List<ViewManager<*, *>> {
+        // This is the earliest we lay our hands on react context.
+        // Moreover this is called before FabricUIManger has finished initializing, not to mention
+        // installing its C++ bindings - so we are safe in terms of creating this hierarchy
+        // before RN starts creating shadow nodes.
+        WEAK_INSTANCE = WeakReference(this)
+        this.reactContextRef = WeakReference(reactContext)
+        ensureDummyLayoutWithHeader(reactContext)
```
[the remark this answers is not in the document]
kkafar: Done in: https://github.com/software-mansion/react-native-screens/pull/2169/commits/d3da35727ffaba855814cd6a8556087961c1a128 and https://github.com/software-mansion/react-native-screens/pull/2169/commits/175f49c9bd196954e4b3545b83663fc479d78c09

android/src/main/java/com/swmansion/rnscreens/RNScreensPackage.kt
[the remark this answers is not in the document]
kkafar: I'll resolve this conversation now, as I aim to tackle this topic in separate PR

android/src/main/java/com/swmansion/rnscreens/RNScreensPackage.kt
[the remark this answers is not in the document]
kkafar: Resolving for now, let me know here or on DM if you disagree

android/src/main/java/com/swmansion/rnscreens/Screen.kt
kkafar: Ah, while I see why that's bothering, but... the file was formatted, because there were some logic changes there at some point that were later on removed. Right now restoring it would require *manually restoring bad formatting*. I'm not willing to do this :sweat_smile:

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
```diff
@@ -26,13 +41,103 @@ class RNSScreenComponentDescriptor final
...
+                headerConfigChild->getProps());
+        fontSize = headerProps.titleFontSize;
+        headerHidden = headerProps.hidden;
+      }
 
+      const auto headerHeight =
+          headerHidden ? 0.f : findHeaderHeight(fontSize).value_or(0.f);
+
+      // This will be consumed by header config node down in tree
+      rnscreens::gHeaderHeightCorrection = headerHeight;
```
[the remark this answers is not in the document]
kkafar: I resigned from this approach, but that worked, because **during layout** clones & adapts are propagated top-down similar to how layout flows.

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
```diff
@@ -26,13 +41,103 @@ class RNSScreenComponentDescriptor final
...
       layoutableShadowNode.setSize(
           Size{stateData.frameSize.width, stateData.frameSize.height});
-    }
+      rnscreens::gHeaderHeightCorrection = 0.f;
+    } else {
+      auto headerConfigChildOpt = findHeaderConfigChild(layoutableShadowNode);
+      int fontSize = kFontSizeUnset;
+      bool headerHidden = true;
+
+      if (headerConfigChildOpt) {
```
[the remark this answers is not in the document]
kkafar: This code is executed once / twice per screen. I do not think we should look too much into performance here.

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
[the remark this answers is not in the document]
kkafar: As to computing the header height in C++ layer: the header height depends on multiple multiple attributes defined in application resources & theme, font size, font style, font weight, etc. Moreover whole coordinator layout flow is not so straightforward and I would not want to replicate it here.

As to caching the value on C++ layer:

This should not be called more than twice. I don't think it is worth of introducing another complexity layer here to make a single JNI call less.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -66,7 +66,8 @@ class Screen(context: ReactContext?) : FabricEnabledViewGroup(context) {
             val height = b - t
 
             val headerHeight = calculateHeaderHeight()
-            val totalHeight = headerHeight.first + headerHeight.second // action bar height + status bar height
+            val totalHeight =
```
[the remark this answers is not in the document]
kkafar: I added a comment there and I wouldn't mind if it has stayed. If you are strongly opinionated here I'm willing to open separate PR with that single line comment 😄

android/src/main/java/com/swmansion/rnscreens/RNScreensPackage.kt
```diff
@@ -51,4 +104,120 @@ class RNScreensPackage : TurboReactPackage() {
...
+        toolbar = Toolbar(contextWithTheme).apply {
+            title = "FontSize123!#$"
+            layoutParams = AppBarLayout.LayoutParams(
+                AppBarLayout.LayoutParams.MATCH_PARENT,
+                AppBarLayout.LayoutParams.WRAP_CONTENT
+            ).apply { scrollFlags = 0 }
+        }
+
+        // We know the title text view will be there, cause we've just set title.
+        defaultFontSize = ScreenStackHeaderConfig.findTitleTextViewInToolbar(toolbar)!!.textSize
```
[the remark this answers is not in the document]
kkafar: I mean, I consider throwing NPE there adequate. The comment explains why using `!!` is safe there. If the function returns `null` somehow, we throw NPE.

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
kkafar: I mean, we could compare header height received from the native side & see whether it matches the value we computed in the very first layout - however right now we pass invalid value there (as `contentOffsetY`) and it requires separate fix that I believe is outside the scope of this PR. 

I can add a check inside this `if`, that checks whether the very first frame received from UI is the same as the frame computed during the very first layout. That's ok?

WoLewicki: Yeah, that's what I meant. If they are the same, then the views should physically be in the same place yeah?

android/src/main/jni/CMakeLists.txt
kkafar: Added changes to this file now

android/src/main/jni/CMakeLists.txt
```diff
@@ -23,15 +23,15 @@ file(GLOB LIB_CODEGEN_SRCS CONFIGURE_DEPENDS ${LIB_ANDROID_GENERATED_COMPONENTS_
 
 add_library(
   ${LIB_TARGET_NAME}
-  SHARED 
+  SHARED
```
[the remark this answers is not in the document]
kkafar: Added changes to this file now.

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
kkafar: Done in: https://github.com/software-mansion/react-native-screens/pull/2169/commits/e3c2fd89f531562c72dd5d62c150b5a28de90380

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
```diff
@@ -26,13 +38,139 @@ class RNSScreenComponentDescriptor final
...
 
+#ifdef ANDROID
     if (stateData.frameSize.width != 0 && stateData.frameSize.height != 0) {
+      // When we receive dimensions from JVM side we can remove padding used for
+      // correction, and we can stop applying height correction for the frame.
+      // We want to leave top offset correction though intact.
+      // TODO: In future, when we have dynamic header height we might want to
+      // update Y offset correction here.
+
+#ifdef REACT_NATIVE_DEBUG
```
WoLewicki, line 49: Is it a global flag that is always present?

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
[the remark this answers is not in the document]
WoLewicki: Still, some lines later you will set some paddings in all `Screen` components, event those without a `HeaderConfig` component, so every other navigator's ones. Why not put all this code inside the `if`?

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.h
```diff
@@ -18,8 +22,26 @@ class JSI_EXPORT RNSScreenShadowNode final : public ConcreteViewShadowNode<
                                                  RNSScreenState> {
  public:
   using ConcreteViewShadowNode::ConcreteViewShadowNode;
+  using StateData = ConcreteViewShadowNode::ConcreteStateData;
```
WoLewicki, line 25: This `StateData` seems to be only used on `Android`, so maybe we should wrap it with `ifdef` too?

common/cpp/react/renderer/components/rnscreens/utils/RectUtil.h
```diff
@@ -0,0 +1,35 @@
...
+#include <react/renderer/graphics/Float.h>
+#include <react/renderer/graphics/Size.h>
+#include <cmath>
+#include <concepts>
+
+namespace rnscreens {
+
+namespace react = facebook::react;
+
+template <typename T>
```
WoLewicki, line 12: Why did you make it extensible if there is only one usage of the method with exactly the same parameters? Or am I missing something?

common/cpp/react/renderer/components/rnscreens/utils/RectUtil.h
```diff
@@ -0,0 +1,35 @@
...
+ * @param first first frame size
+ * @param second second frame size
+ * @param eps comparison precision
+ * @return whether the frame dimensions are the same with respect to given
+ * epsilon
+ */
+inline bool compareFrameSizes(
+    const react::Size &first,
+    const react::Size &second,
+    react::Float eps = 0.01) {
```
WoLewicki, line 30: Is this `eps` a magic number or is it a standard for those measurements?

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.cpp
```diff
@@ -11,5 +14,51 @@ Point RNSScreenShadowNode::getContentOriginOffset() const {
...
+#ifdef ANDROID
+  applyFrameCorrections();
+#endif // ANDROID
+}
+
+#ifdef ANDROID
+void RNSScreenShadowNode::applyFrameCorrections() {
+  ensureUnsealed();
+
+  // On the very first layout we want to correct both Y offset and frame size.
```
WoLewicki, line 29: So here we try to mimic the values that will be eventually set on the shadow node so that there is no new layout passes after they come? Why is this needed here too and how does it differ from what we do in `adopt`?

WoLewicki: It's because `layoutMetrics_` are ignored and `reanimated` reads from them?

WoLewicki (commented):
Left some additional comments.

android/src/main/java/com/swmansion/rnscreens/RNScreensPackage.kt
```diff
@@ -7,15 +7,29 @@ import com.facebook.react.module.annotations.ReactModuleList
 import com.facebook.react.module.model.ReactModuleInfo
 import com.facebook.react.module.model.ReactModuleInfoProvider
 import com.facebook.react.uimanager.ViewManager
+import com.swmansion.rnscreens.utils.ScreenDummyLayoutHelper
```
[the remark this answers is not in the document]
kkafar: Because the symbol is invisible w/o the import.

Its in `.utils` package.

apps/test-examples/src/App.js
```diff
@@ -0,0 +1,111 @@
...
+import Test2048 from './Test2048';
+import Test2069 from './Test2069';
+import Test2118 from './Test2118';
+import TestScreenAnimation from './TestScreenAnimation';
+import TestHeader from './TestHeader';
+
+enableFreeze(true);
+
+export default function App() {
+  return <Test42 />;
```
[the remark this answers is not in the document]
kkafar: Good catch. `git` added it after merge - merge artifact.

https://github.com/software-mansion/react-native-screens/pull/2169/commits/c5e3c6ee572f93e9e1341852a1073016b603674c

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
kkafar: Available since 2021 in RN. It is defined with respect of `NDEBUG` <- this is a standardised macro by the language spec, so we can use that one if you prefer, but I think we should be fine with `REACT_NATIVE_DEBUG` also.

common/cpp/react/renderer/components/rnscreens/utils/RectUtil.h
kkafar: Just for produced code quality and I'll stand by that. Such util has potential for reusability and it is ready.

common/cpp/react/renderer/components/rnscreens/utils/RectUtil.h
kkafar: This is a reasonable default value, that's it. I selected it, so that we are below 1px threshold of inaccuracy, but you can compare your frame sizes with any accuracy you want. 

One thing to consider here is to make the inequality weak, instead of strict, so that `eps == 0.f` is possible and I'll do that in a sec.

kkafar: https://github.com/software-mansion/react-native-screens/pull/2169/commits/184a527d77f0dfc5419ae2d9a1c911748e2566ab

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.cpp
kkafar: In adopt we apply padding, so that during layout pass `Yoga` layouts `Screen`'s children with less available space. `Yoga` won't modify `Screen`'s size / frame origin itself. 
Here we do this manually, so when (e.g. reanimated) someone tries to read `Screen` dimensions from the shadow node they are possible accurate. If someone tries to hack around and read from `yogaNode_` private field, he will get the wrong value, but we don't care, if someone does this he's already doing something wrong.

kkafar: > It's because layoutMetrics_ are ignored and reanimated reads from them?

Yeah, changes to `layoutMetrics_` do not trigger layout pass. This is just a simple data field w/o any logic attached to setters / getters. ScrollView does the same. 

Also, this field is later read when sending updates to UI.

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.h
kkafar: That's compile time utility only, it does not allocate any space / do any actions during runtime. There isn't even a warning here when building iOS (just checked it out of curiosity).

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
```diff
@@ -0,0 +1,177 @@
...
+        coordinatorLayout = CoordinatorLayout(contextWithTheme)
+
+        appBarLayout = AppBarLayout(contextWithTheme).apply {
+            layoutParams = CoordinatorLayout.LayoutParams(
+                CoordinatorLayout.LayoutParams.MATCH_PARENT,
+                CoordinatorLayout.LayoutParams.WRAP_CONTENT,
+            )
+        }
+
+        toolbar = Toolbar(contextWithTheme).apply {
```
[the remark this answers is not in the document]
kkafar: Very good catch. I guess it won't behave properly. Testing and fixing right now.

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
WoLewicki: Just make sure to use a flag that won't be dropped soon 😄

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.h
WoLewicki: Ok, I thought rather about the code readability in this case.

common/cpp/react/renderer/components/rnscreens/utils/RectUtil.h
WoLewicki: So could you use a named value so it is clear why is it used here?

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
kkafar: No deprecation mark here

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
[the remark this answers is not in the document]
kkafar: https://github.com/software-mansion/react-native-screens/pull/2169/commits/cb6fe2b1718f9c966f2a0fa933b8aa955b20e254

works like charm now

common/cpp/react/renderer/components/rnscreens/utils/RectUtil.h
kkafar: I don't think it's good idea to introduce some random namespace scoped variable that is supposed to be default value for a function parameter, even if I mark it as static - that's a bit code smelly. I'll just add a comment.

common/cpp/react/renderer/components/rnscreens/utils/RectUtil.h
WoLewicki: So when the reusability is needed, you can just change the method, no? What better quality do you introduce with template for only one use case?

WoLewicki (approved):
LGTM! Just answer all the comments and apply changes if needed before merging.

FabricTestExample/android/settings.gradle
kkafar: Reverted in: https://github.com/software-mansion/react-native-screens/pull/2169/commits/a0d15bbfcda5b84814f217896e667fd516c81577

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.kt
kkafar: Reverted in: https://github.com/software-mansion/react-native-screens/pull/2169/commits/4f34895fbdf72ffad1ede6858b55c8340d664b8b

common/cpp/react/renderer/components/rnscreens/utils/RectUtil.h
kkafar: I'm coming from an angle that it is better to write general and reusable code in principle ([link to C++ core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#t143-dont-write-unintentionally-non-generic-code) - however this is general principle, not only applying to C++). 
If I'm right: your point is that there is no need here for generalised method, because it is overkill for our current use cases and maybe it introduces higher maintenance costs. I can see your point being valid, but I believe it is the same for my point 😄 

Taking these two into consideration I'll stand by current shape of the implementation, mostly because there is no some metaprogramming-wizadry going on here, it is almost as simple function as it could get. If that was some standard-library-type metaprogramming I think I would give you your due.

kkafar: Note: I don't consider this a blocker, so I'll proceed with merging, we can continue the discussion and take appropriate action if needed later.

PR #2173, opened 2024-06-06
refact(TVOS): using common example src

alduzy:
## Description

This PR intents to use common examples src directory in TVOS Example app.

The example screens that were removed with tvos specific src folder are all present in the common directory in similar or slightly more complex shape ready to be tested.

Some of the common examples that are being exposed to TVOS by this PR are not working properly as they were not intended to work on TV platform, those are:
- swipe back animation screen (working well, but without the gesture)
- animations playground
- status bar playground
- orientation playground
- search bar playground
- gestures playground
I decided to filter them out for TV devices.

There are also some issues worth pointing out I found while playing around with the TVOS example:
- header text and background colors are unchangeable (white on light grey in regular screens and gray on dark gray in modals)
- the value in the input is not visible until the input is hovered or focused (try it in Stack Presentation -> formSheet)
- select inputs are not selectable - just expandable, thus not accessible (try it in header options playground)

## Changes

- removed TVOS src directory
- linked apps/examples/App in TVOS App.tsx
- added TVOS Platform ternary in common App.tsx file

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


apps/examples/App.tsx
```diff
@@ -143,17 +152,21 @@ const MainScreen = ({ navigation }: MainScreenProps): React.JSX.Element => (
...
-      .filter(name => SCREENS[name].type === 'playground')
-      .map(name => (
-        <ListItem
-          key={name}
-          testID={`root-screen-playground-${name}`}
-          title={SCREENS[name].title}
-          onPress={() => navigation.navigate(name)}
-        />
-      ))}
+    {!Platform.isTV && (
```
maciekstosio, line 155: Is this if necessary? Maybe some playgrounds will be available for TVOS in the future. If you don't want to show the list if it's empty then we could check length.

apps/examples/App.tsx
```diff
@@ -107,6 +112,10 @@ const SCREENS: Record<
   },
 };
 
+const screens = Object.keys(SCREENS).filter(
```
maciekstosio, line 115: I wonder if it wouldn't be clearer to just do some method that returns check for each platform. What do you think?

apps/examples/App.tsx
```diff
@@ -107,6 +112,10 @@ const SCREENS: Record<
...
+  name => !Platform.isTV || !!SCREENS[name].isTVOSready
```
maciekstosio, line 116: I wonder if it wouldn't be clearer to just do some method that returns check for each platform. What do you think?

apps/examples/App.tsx
alduzy: https://github.com/software-mansion/react-native-screens/pull/2173/commits/4d12007c5b5c5c0ab84dace099e888d7f320c545

apps/examples/App.tsx
alduzy: Good point, looks a little cleaner now:
https://github.com/software-mansion/react-native-screens/pull/2173/commits/1e5856ebcafa28b580d97515df082694772496ee

apps/examples/App.tsx
alduzy: https://github.com/software-mansion/react-native-screens/pull/2173/commits/4d12007c5b5c5c0ab84dace099e888d7f320c545

apps/examples/App.tsx
```diff
@@ -133,27 +150,27 @@ const MainScreen = ({ navigation }: MainScreenProps): React.JSX.Element => (
     <Text style={styles.label} testID="root-screen-examples-header">
```
maciekstosio, line 150: I wonder about adding examples.length for consistency here

maciekstosio (approved):
One thing to consider rest looks ok 🚀

apps/examples/App.tsx
```diff
@@ -133,27 +150,27 @@ const MainScreen = ({ navigation }: MainScreenProps): React.JSX.Element => (
...
-      ))}
+    {examples.map(name => (
+      <ListItem
+        key={name}
+        testID={`root-screen-example-${name}`}
+        title={SCREENS[name].title}
+        onPress={() => navigation.navigate(name)}
+      />
+    ))}
+    {!!playgrounds.length && (
```
maciekstosio, line 161: Should do the same with less nesting. 
```
{!!playgrounds.length && <Text style={styles.label}>Playgrounds</Text>}
         {playgrounds.map(name => (
           <ListItem
             key={name}
             testID={`root-screen-playground-${name}`}
             title={SCREENS[name].title}
             onPress={() => navigation.navigate(name)}
           />
         ))}
     )}
```

alduzy: https://github.com/software-mansion/react-native-screens/pull/2173/commits/5654ac07ae8004ca2c8d3a88531a36b9e29bfaa1

apps/examples/App.tsx
alduzy: https://github.com/software-mansion/react-native-screens/pull/2173/commits/5654ac07ae8004ca2c8d3a88531a36b9e29bfaa1

kkafar (approved):
Yup, this looks nice. Only thing to consider here is to make screens not available on given platform not clickable instead of removing them, but this is also fine. I'm not testing runtime behaviour, just looking at the code. Please make sure it works as expected and I think we're good to go! Thank you! 🎉

alduzy:
Added minor changes like header icon suitable for the platform.
+ The unsupported screens are now disabled instead of hidden:
Screenshot 2024-06-11 at 15 16 19

apps/examples/App.tsx
```diff
@@ -164,7 +179,11 @@ const ExampleApp = (): React.JSX.Element => (
         <Stack.Navigator>
           <Stack.Screen
             name="Main"
-            options={{ title: '📱 React Native Screens Examples' }}
+            options={{
+              title: `${
+                Platform.isTV ? '📺' : '📱'
```
kkafar, line 184: Nice touch 😄

kkafar (approved):
yeah, this is great now 🙌🏻

PR #2113, opened 2024-04-26
fix: issues with presenting owned modals from foreign ones

kkafar:
## Description

Basically this is another edition of the issue #1829 (handled by #1912).
The issue comes down to the fact, that our `ScreenStack` is not aware of all modal view controllers being in presentation,
but this time it is not aware of third-party modal view controllers (I've named them "foreign" modals in opposite to "owned" modals).

This PR is not a comprehensive solution but rather just a patch aiming at fixing one particular interaction reported in #2048.

I've left verbose code comments explaining the issue and suggesting solution in the source code, including:

```
  // TODO: Find general way to manage owned and foreign modal view controllers and refactor this code. Consider building
  // model first (data structue, attempting to be aware of all modals in presentation and some text-like algorithm for
  // computing required operations).
```

Closes #2048
Closes #2085

## Changes

Trigger dissmisal of foreign modal if it is presented above `changeRoot` modal (last modal that is to stay on stack after the updates).

## Test code and steps to reproduce

`Test2048` in `TestsExample` & `FabricTestExample`.

## Checklist

- [x] Included code example that can be used to test this change
- [x] Ensured that CI passes

Issue #2125, opened 2024-05-09
react-navigation: App randomly freezes when react-native Modal is shown while preventing remove screen with usePreventRemove on iOS


BadLice:
### Description

I have a NativeStackNavigator with 2 screens; on the first screen i have a button to navigate to the second screen; on the second screen, i am using the `usePreventRemove` hook to show a confirmation modal to the user when i tries to navigate to the previous screen; 

```tsx
import React, {useCallback, useState} from 'react';
import {
  NavigationAction,
  useIsFocused,
  useNavigation,
} from '@react-navigation/native';
import usePreventRemove from '@react-navigation/core/src/usePreventRemove';
import {NativeStackNavigationProp} from '@react-navigation/native-stack';
import {Modal, Text, View, StyleSheet} from 'react-native';
import {Button} from './Button.tsx';
import {Routes} from './StackNavigator.tsx';

export const ExitConfirmModal = () => {
  const navigation =
    useNavigation<NativeStackNavigationProp<Routes, keyof Routes>>();
  const [pendingAction, setPendingAction] = useState<NavigationAction>();
  const isFocused = useIsFocused();

  const handlePressCancel = useCallback(() => {
    setPendingAction(undefined);
  }, []);

  const handlePressConfirm = useCallback(
    (action: NavigationAction) => {
      setPendingAction(undefined);
      navigation.dispatch(action);
    },
    [navigation],
  );

  usePreventRemove(isFocused && !pendingAction, event => {
    setPendingAction(event.data.action);
  });

  if (!pendingAction) {
    return null;
  }

  return (
    <Modal
      animationType="none"
      hardwareAccelerated
      transparent
      visible
      onRequestClose={handlePressCancel}>
       <View>
        <Button onPress={() => handlePressConfirm(pendingAction)}>
          <Text>Go back</Text>
        </Button>
        <Button onPress={handlePressCancel}>
          <Text>Cancel</Text>
        </Button>
      </View>
    </Modal>
  );
};
```

here it is an example of the bug behaviour:

https://github.com/react-navigation/react-navigation/assets/37150312/48ade50f-a6d9-4a3b-bf69-abdcb9736c5c

### Steps to reproduce

1. open the app
2. press "go to second" button to navigate to second screen
3. press the back button to go back to previous screen
4. when the modal open, press cancel
5. repeat step 3
6. sometimes the app freezes when the modal appears.

[...]

PR #2175, opened 2024-06-08
fix(Paper,iOS): dismiss all attached view controllers correctly on reload

hirbod:
## Description

This PR addresses an issue with react-native-screens where modals were not being dismissed correctly during app reloads when combined with foreign view controllers. The fix involves enhancing the invalidate method to recursively dismiss all presented view controllers, ensuring a clean state on reload.

This is not a development-only problem; this fix also addresses reloads from OTA updates.

## Changes

- Enhanced invalidate method in RNSScreenStack.mm to recursively dismiss all presented view controllers.
- Ensured _presentedModals is cleared and _controller is detached from its parent during invalidation.
- Added a helper method dismissAllPresentedViewControllersFrom to handle recursive dismissal logic.

## Screenshots / GIFs

| Before | After |
|--------|--------|
| <video width="320" height="240" controls src="https://github.com/software-mansion/react-native-screens/assets/504909/1476ab3a-4bd9-4ffa-9256-760467a108bc"></video> | <video width="320" height="240" controls src="https://github.com/software-mansion/react-native-screens/assets/504909/e6c5eef0-16c6-49a2-9124-86467feec7f2"></video> |

The red background is from a `transparentModal` by RNS, the sheet is a foreign view controller. Before the change, `react-native-screens` would break on reload if a foreign view controller was mounted on top. I came to the solution after finding [this PR](https://github.com/software-mansion/react-native-screens/pull/2113). The issue originally started [here](https://github.com/lodev09/react-native-true-sheet/issues/34). After my changes, RNS works correctly on reload with third-party controllers.

I have no experience with Fabric, so I can't help with that. Feel free to update the solution for Fabric if needed.

Test2175

## Checklist

- [x] Ensured that CI passes


kkafar:
Thank you! The code changes look good, however I'll need to test it & see what's going on on Fabric. Will be back with more insights once I do that (most likely Monday / Tuesday).

hirbod:
`dismissOnReload` is not implemented for the new architecture anyway, so this should not be blocker for now
https://github.com/software-mansion/react-native-screens/blob/5cfafbbf42a3b6ebcdb7772012f760c64bb305df/ios/RNSScreenStack.mm#L696-L704

hirbod:
I found an edge case with my PR. It will break the animation if you navigate from an open modal to a different tab, for example. It will hardly detach it. I will update the PR. We only need to kill recursively in `dismissOnReload`, not for the regular `invalidate`

hirbod:
@kkafar I have updated my approach, this seems solid now
https://github.com/software-mansion/react-native-screens/pull/2175/commits/b0dea33054a197cb0680061de64aca82539f8ae0

hirbod:
@kkafar friendly reminder :)

kkafar:
Yeah, thanks for the reminder. I'm convinced that GitHub needs a option of posting a periodic notification until something is done. 

I'll allow myself liberty of pushing onto your branch, hope you don't mind.

kkafar:
Ok, I'm thinking through this change right now & I see a **potential** issue. 

This goes through system chain of `presentedModals` and removes them one-by-one back-to-front.
Given that I can imagine a situation where there are modals presented from various stacks, then one stack is popped and all modals are popped all together, where desired behaviour would be to drop only ones owned by to-be-popped stack. This is kinda weird scenario, but I can imagine someone doing that. 

Second remark is that is should be sufficient to pop only bottom-most modal to dismiss whole chain, w/o the need for recursion, but that's implementation detail I'll correct later. 

Right know I want to ask you if you have throughput to provide me (& the PR) with issue reproduction so that I can play with it right now & someone has opportunity to test this behaviour later.

hirbod:
This PR aims to fix issues with reloading (pressing R) when RNS is used in conjunction with third-party UIViewControllers. We have a pretty complex setup, and I did not encounter any regressions with my proposed fix.

I’d suggest we merge this and address any regressions if they are reported (which I highly doubt). I also tested scenarios where you navigate from an open modal to a different bottom tab, and everything behaved exactly the same with the change.

kkafar:
I'll insist on proper reproduction here, so that I can play around & test this. I'll have some throughput to prepare one myself *most likely* early next week. 

> I don’t think this is a thing in iOS. Modals never really belong to a specific stack; they are always pushed on top (they can have an inline stack, though, which is unrelated in my opinion)

You are right with respect to the native platform. I meant by "modals owned by stack" these screens, that have modal presentation and are rendered directly as children of given `ScreenStack` in the JS layer.

hirbod:
I'll create a repro for you.

hirbod:
While working on your repro, I discovered that my PR fixes all related modal problems on reload, not just those by third parties. Currently, RNS closes all modals but keeps the last one mounted. This also makes the yellowbox log unclickable until all modals are closed. Little side bug I've discovered: when pulling on modals the header height gets totally borked. It's also visible in my video, but maybe something for a different PR. Repro coming in 10min.


https://github.com/software-mansion/react-native-screens/assets/504909/30775d37-b42d-407f-b006-2cfe535ab477

hirbod:
Hey @kkafar, here is the link to https://github.com/hirbod/RNSModalDismissRepro along with a video showcasing the issue. Additionally, this PR addresses #2125. Due to the way modals and UIPresentationController behave on iOS, I haven't found any other solution than forcibly closing all modals on reload.

Make sure to check the (i), tab one, and tab two. The video makes things clearer in my opinion. Whenever you see the "Metro reloading" bar, I pressed R.

https://github.com/software-mansion/react-native-screens/assets/504909/bcaea223-acb2-4011-b809-8c30902aa879

hirbod:
Hi @kkafar, did you get a chance to look into this? 😊

kkafar:
No, haven't sit down to yet, sorry. I'll put it much higher on my board rn

lucacaputo:
any update on this one? 👀

hirbod:
@tboba any chance you could look into this?

lucacaputo:
I found out that it works when setting `animation='none'`, although it's not going to look nice. My current workaround is the following:
```ts
import { useState, useEffect, useMemo, ComponentProps } from 'react';
import Animated, {
  useSharedValue,
  useAnimatedReaction,
  runOnJS,
  FadeOut,
} from 'react-native-reanimated';

const useModalVisibilityState = (isVisibleProp: boolean | undefined) => {
  const isSharedVisible = useSharedValue(isVisibleProp);
  const [isModalVisible, setIsModalVisible] = useState(isVisibleProp);

  useEffect(() => {
    if (isVisibleProp) {
      isSharedVisible.value = true;
    }
  }, [isSharedVisible, isVisibleProp]);

  useAnimatedReaction(
    () => isSharedVisible.value,
    (curr) => runOnJS(setIsModalVisible)(curr),
  );

  const fadeOutAnimation = useMemo<
    Exclude<ComponentProps<typeof Animated.View>['exiting'], undefined>
  >(
    () =>
      FadeOut.withCallback((finished) => {
        if (finished) {
          isSharedVisible.value = false;
        }
      }),
    [isSharedVisible],
  );

  return useMemo(
    () => ({
      isModalVisible,
      fadeOutAnimation,
    }),
    [fadeOutAnimation, isModalVisible],
  );
};

export default useModalVisibilityState;
```

then set the modal's animation prop to `'none'` and wrap your children in an `Animated.View`, use the `isModalVisible` as the `visible` prop of the modal, and add a prop to your component that toggles the content visibility (hence triggering the unmount animation of the modal contents).

lucacaputo:
@hirbod sorry I thought I was commenting on a related issue

kkafar:
Hey, I finally have time to look into it. I'm sorry for such long delay.

I've rebased the PR on main and tested the issue on a added reproduction (Test2175 in example apps). I can't reproduce the issue. 

I don't want to use Expo-based reproduction, because expo often runs its own logic which might interfere, thus I've prepared one based solely on react-navigation@v7 attempting to mimic structure you presented.

The example has stack navigator, with few screens: first contains tab navigator, the rest of them are push screen, transparent modal screen & modal screen. 

Inside tab navigator there are buttons to open the modals from outer stack navigator. I can't trigger the behaviour you described in any navigation scenario. Might it be possible, that this issue has been fixed some other way? We had merged few PRs related to foreign modals behaviour last few weeks.

hirbod:
@kkafar Current versions still break. We're using Expo Router, and my reproduction is still working. Upgrading to v7 is not an option for us right now. My PR fixes the issue. Honestly, it's tough to look into this again after almost 4 months.

kkafar:
I see, but I'm not willing to merge this, since I'm not able to reproduce the issue without Expo involved. I'll ask @adrianryt to take a look with fresh pair of eyes.

hirbod:
Expo Router is one of the most significant use cases for react-native-screens, and having Expo involved is a big deal.

adrianryt:
I allowed myself to push changes to test file. I want to note 4 scenarios related to the issue: 
### Scenario 1
Repro: 
- Start the Test2175 app and click show owned modal button.
- click "Push foreign modal" (not "Push foreign modal(inside Screen Component)") 
- Reload app
Video: 

https://github.com/user-attachments/assets/f4264fed-8a80-4c31-b41e-2c13d5b71a15


Summary: 
In this scenario, the behaviour with `oldInvalidateImpl` is **correct**.
### Scenario 2
Repro:
- Start the Test2175 app and click show owned modal button.
- click "Push foreign modal" then again click "Push foreign modal" and again.
- Reload app

Video:

https://github.com/user-attachments/assets/8c2639a8-3a8e-45be-b439-5b0f25205f3c

Summary: 
In this scenario, modal is not dismissed and changing ios code to use `newInvalidateImpl` solve the issue.

### Scenario 3
Repro: 
- Start the Test2175 app and click show owned modal button.
- click "Push foreign modal (inside Screen Component)" button
- Reload app
Video: 

https://github.com/user-attachments/assets/0109b658-85e0-400d-8522-34139f160ee2

Summary: 
In this scenario, modal is not dismissed and changing ios code to use `newInvalidateImpl` solve the issue.

### Scenario 4
Repro: 
- Start the Test2175 app and click show owned transparentModal button.
- click "Push another modal" button
- Reload app
Video: 


https://github.com/user-attachments/assets/6820ff58-60dd-4ee8-bdea-accbbc7a82e4


Summary: 
In this scenario, transparentModal is not dismissed, it is not possible to interact with the app anymore and changing ios code to use `newInvalidateImpl` solve the issue.

### Conclusions
- From scenario 2 and 3 I would say that whenever multiple modals wrapped in `Stack.Screen` component are pushed on the navigation stack, reloading the app will cause described issue.
- Solution provided by hirbod solves the issue, however, recursive removing the modals can be seen on the screen. 
- I also checked [#2125](https://github.com/software-mansion/react-native-screens/issues/2125), I managed to create the reproduction but unfortunately this solution did not work for me. _I will create a new PR with reproduction and link it here._ 

cc @kkafar

kkafar:
Thank you @adrianryt! I'll work on the implementation suggestion [(link)](https://github.com/software-mansion/react-native-screens/pull/2175#issuecomment-2194271620) and proceed with merging.

kkafar (approved):
Took a long time - sorry for that.  

I've changed the implementation - instead of implementing recursive dismissal by hand we now dismiss all controllers presented from `_controller` or "down" the presented view controllers hierarchy by letting system do its job. With these changes view controllers should be removed immediately on reload, w/o glitchy effect reported by @adrianryt.

If a controller was presented outside the screen stack (foreign modal not inside a screen) it should be responsible for its own dismissal.

PR #2176, opened 2024-06-08
chore(Android): fix lint warning of CustomToolbar component

kkafar:
## Description

Just noticed while working on #2169 that we got a warning in CustomToolbar.

Edit: followed review suggestions and suppressed lints for all our view, where this was requried.

## Changes

Suppressed lint on missing constructors. 

We're safe to miss these there, as this view is constructed only programatically (we do not inflate any of our views).

N/A

## Checklist

- [ ] Ensured that CI passes


tboba (approved):
LGTM! I'm just concerned if we should suppress all of these warnings with unused values like in this PR? What do you think?
Should we maybe do this in that PR?

kkafar:
Yeah, I'll look for other places with similar issue and update the PR

maciekstosio (approved):

PR #2182, opened 2024-06-12
fix: typescript paths

alduzy:
## Description

This PR fixes the errors showing on `yarn prepare` command. Turned out that custom path definition to react-navigation folder produced typescript errors when running the command. The default node_modules directory contains typescript definitions that do not cause any problems.
~~Some of the react-navigation packages were missing in node modules and the ts support was missing in examples so I added them as devDependencies - now there are no errors during the prepare command and there is full TS support in common examples directory.~~

~~The alternative approach would be to define compilerOptions paths pointing to react-navigation folder inside examples and test-examples tsconfigs individually instead of adding new devDependencies~~

Edit: decided to go with the second approach as it does not add any extra dependencies - the tsconfigs inside common example directories redefine paths.

## Changes

- removed custom paths to react-navigation types (node_modules is the default)
- added the paths to react-navigation types individually in apps/examples and apps/test-examples

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


kkafar (commented):
Let's make sure it works now 😅 Please test by running `yarn prepare`, runnning apps, looking for errors in code editor (we don't want "red" back.

Let me know ☝🏻 how does it look

alduzy:
@kkafar Now the `yarn prepare` command works as expected without any errors with typescript path pointing to node_modules/@react-navigation (default).
The custom path pointing to react-navigation folder is now only used inside example dirs to provide ts support in example files and it works as intended 🙂.

kkafar (approved):
I guess this is how this should work, because inside our native stack implementation (v5) we still use old types from `react-navigation` v5, thus it does not make sense to use types from our submodule. The goal of the submodule is to have easy time modifying v6 / v7 native stack code within our examples.

The changes look fine, make sure this works fine with code editors & our example do work just fine and we're good to go.

PR #2185, opened 2024-06-14
fix(iOS): back button does not respect I18nManager & improve RTL handling in header

### LINKED ISSUE #1884: headerLargeTitle & headerSearchBarOptions Doesn't Support RTL

alduzy:
## Description

This PR intents to fix back button direction in RTL mode forced by I18nManager for testing purposes on iOS.

When the application's language is changed using the `CFBundleDevelopmentRegion` the system is well aware of that, passes the information to React Native and Yoga. The result is a fully LTR / RTL compliant application.

However, when the `RTL` mode is set using the React Native's `I18nManager.forceRTL()` method (for testing purposes I believe) The change is applied to react views only and the system itself does not change it's direction.
In such case the only way for the native elements, such as `navigationBar` to adapt to the direction set by `I18nManager` is by manually applying the `semanticContentAttribute` based on `direction` prop passed by `react-navigation`.

> [!note]
In case there is another native `navigationBar` element apart from the one coming from `react-native-screens`, this change may override it's direction.

Tested successfully on iOS 17.5 and 16.4

Fixes #1884 .

## Changes

- removed faulty workaround
- set correct semanticContentAttribute for navigationBar and it's contents
- refactored how the searchbar direction is set

### Before
image

### After
image

- play around with LTR and RTL mode in the example app

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

## Short story on how RTL management works in RN apps on iOS by @kkafar

> [!note]
> Basically we have two RTL management systems: native one & ReactNative's I18nManager.
>
> Native one works most likely by applying semantic content attribute to all the views / classes and impacting layout constraints, thus whole native application is laid out properly.
> We can impact RTL on natively managed views by:
> * setting semantic content attributes for particular view
> * [setting semantic content attributes for particular class through appearance proxy](https://developer.apple.com/documentation/uikit/uiappearance?language=objc)
> * [setting semantic content attributes for particular class contained in instances of another class](https://developer.apple.com/documentation/uikit/uiappearance/1615006-appearancewhencontainedin?language=objc)
> 
> Topic's to research:
> * how parent's semantic content attribute (set directly on given view) does affect subviews
> * exact semantics of using appearance proxy object, their priority and resolve order.
> 
> 
> React Native has notion of internationalisation manager (I18n) which works as follows:
> 
> * The RTL is allowed by default.
> * If RTL is not forced from JS, and allowed, then system language writing direction is resolved based on system calls and such obtained interface direction is passed down to Yoga *impacting all views that are laid out by Yoga*, but not purely native ones such as our header. Thus leading to conformity between system options and RN app options.
> * If RTL is forced from JS and allowed, them Yoga managed views are laid out in RTL mode, while system managed ones respect the system settings.
> * ForceRTL / AllowRTL flags are stored in `NSUserDefaults` application persistent storage, thus living through application restarts. These flags are RN specific, and system has no knowledge of them. 
> 
> React Navigation takes the value from RN I18n system and passes it down to our screen / stack / header components.

### DIFF
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -494,10 +494,15 @@ namespace react = facebook::react;
        config.direction == UISemanticContentAttributeForceRightToLeft) &&
       // iOS 12 cancels swipe gesture when direction is changed. See #1091
       navctr.view.semanticContentAttribute != config.direction) {
+    // This is needed for swipe back gesture direction
     navctr.view.semanticContentAttribute = config.direction;
-    [[UIButton appearanceWhenContainedInInstancesOfClasses:@[ navctr.class ]]
+
+    // TODO: Some comment here please
+    navctr.navigationBar.semanticContentAttribute = config.direction;
+
+    [[UIButton appearanceWhenContainedInInstancesOfClasses:@[ navctr.navigationBar.class ]]
         setSemanticContentAttribute:config.direction];
-    [[UIView appearanceWhenContainedInInstancesOfClasses:@[ navctr.class ]]
+    [[UIView appearanceWhenContainedInInstancesOfClasses:@[ navctr.navigationBar.class ]]
         setSemanticContentAttribute:config.direction];
   }
 
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -495,8 +495,10 @@ namespace react = facebook::react;
       // iOS 12 cancels swipe gesture when direction is changed. See #1091
       navctr.view.semanticContentAttribute != config.direction) {
     navctr.view.semanticContentAttribute = config.direction;
-    [[UIButton appearance] setSemanticContentAttribute:config.direction];
-    [[UIView appearance] setSemanticContentAttribute:config.direction];
+    [[UIButton appearanceWhenContainedInInstancesOfClasses:@[ navctr.class ]]
+        setSemanticContentAttribute:config.direction];
+    [[UIView appearanceWhenContainedInInstancesOfClasses:@[ navctr.class ]]
+        setSemanticContentAttribute:config.direction];
   }
 
   if (shouldHide) {
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -495,7 +495,8 @@ namespace react = facebook::react;
       // iOS 12 cancels swipe gesture when direction is changed. See #1091
       navctr.view.semanticContentAttribute != config.direction) {
     navctr.view.semanticContentAttribute = config.direction;
-    navctr.navigationBar.semanticContentAttribute = config.direction;
+    [[UIButton appearance] setSemanticContentAttribute:config.direction];
+    [[UIView appearance] setSemanticContentAttribute:config.direction];
   }
 
   if (shouldHide) {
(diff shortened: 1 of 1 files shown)


ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -495,7 +495,8 @@ + (void)updateViewController:(UIViewController *)vc
       // iOS 12 cancels swipe gesture when direction is changed. See #1091
       navctr.view.semanticContentAttribute != config.direction) {
     navctr.view.semanticContentAttribute = config.direction;
-    navctr.navigationBar.semanticContentAttribute = config.direction;
+    [[UIButton appearance] setSemanticContentAttribute:config.direction];
+    [[UIView appearance] setSemanticContentAttribute:config.direction];
```
WoLewicki, line 499: It sets `RTL` on all buttons and views in the app this way?

alduzy: I am not sure about that. Probably not, as setting the UIView appearance here does not change the direction of search bar contents - it still required to set the direction for those subviews individually.

kkafar: I'm reading now into RTL support in UIKit & it seems it does. I'll post a review once I'm familiar a bit more with this mechanism in UIKit.

alduzy: Ended up narrowing targets to UIViews and UIButtons contained in `navctr` https://github.com/software-mansion/react-native-screens/pull/2185/commits/a2dcafc57f947d023341d1614f7e7fca3d4aaecc

WoLewicki: Since practically the whole App should be under the navigation structure, it does pretty much the same 😅

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -494,8 +494,17 @@ + (void)updateViewController:(UIViewController *)vc
        config.direction == UISemanticContentAttributeForceRightToLeft) &&
       // iOS 12 cancels swipe gesture when direction is changed. See #1091
       navctr.view.semanticContentAttribute != config.direction) {
+    // This is needed for swipe back gesture direction
     navctr.view.semanticContentAttribute = config.direction;
+
+    // This is responsible for the direction of the navigationBar nad its contents
```
kkafar, line 500:
```suggestion
    // This is responsible for the direction of the navigationBar and its contents
```

kkafar (commented):
A typo here, and I'll have some remarks to the PR description, so that we describe the whole mechanism in details, so that our debugging hours don't go to waste & we have good starting point for refactoring current header config code.

kkafar (approved):
Let's proceed 💪🏻

PR #2186, opened 2024-06-17
fix: bump gradle wrapper version to 7.5.1 & not use app node_modules when running tasks from repo

kkafar:
## Description

Basically tasks run from react-native-screens repo (such as linting & formating) would fail due to outdated gradle wrapper version (and the fact that we re using newer JDK now).

There was a second issue: when running tasks from react-native-screens project as gradle's `rootProject` and simultaneously having `react-native` repo side-by-side to `react-native-screens`
repo on disk - our current react native directory resolution algorithm would have found it and attempt to use to retrieve RN version, but:

1. the gradle.properties file is not there,
2. retrieved version could be out of sync with what's in project.

## Changes

* Bumped gradle wrapper version to 7.5.1 (we were still on major 6).
* Adjusted the react native directory algorithm just a bit, so that when running in context of screens repo we skip one code path.

There was also possibility to "reorder the ifs" in `resolveReactNativeDir` function, however this could possible cause issues in user projects - we want first to look for app's node modules
and later for project node modules.

`yarn lint-android`, `yarn format-android`, etc. should now work just fine when you re on JDK 17+ & have `react-native` repo side by side to `react-native-screens`.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -22,6 +22,10 @@ buildscript {
     }
 }
 
+def isRunningInContextOfScreensRepo() {
+    return project == rootProject
+}
+
 def isNewArchitectureEnabled() {
     // To opt-in for the New Architecture, you can either:
     // - Set `newArchEnabled` to true inside the `gradle.properties` file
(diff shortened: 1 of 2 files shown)


alduzy (approved):
Looks good 🚀

PR #2192, opened 2024-06-18
chore: reinstall pods in examples after release

kkafar:
## Description

So that this does not come up randomly in various PRs as unrelated changes.

## Changes

- **Bump in TE**
- **Bump in FTE**
- **Bump in Example**
- **Bump in FabricExample**
- **Bump in TVOSExample**

N/A

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/Example/ios/Podfile.lock
+++ b/Example/ios/Podfile.lock
@@ -1210,7 +1210,7 @@ PODS:
     - ReactCommon/turbomodule/bridging
     - ReactCommon/turbomodule/core
     - Yoga
-  - RNScreens (3.31.1):
+  - RNScreens (3.32.0):
     - DoubleConversion
     - glog
     - hermes-engine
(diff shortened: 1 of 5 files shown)


tboba:
I'm also observing a new type of error in our E2E tests. @kirillzyusko do you have similar behavior with your runners?

kirillzyusko:
@tboba no, because I switched to linux machine (which is more powerful than a Mac runner) and all these random crashes went away 🙂 

This is the first migration step that I made (https://github.com/kirillzyusko/react-native-keyboard-controller/pull/352/files#diff-7efeb6f32b8d7e3c36b551280577125c625c730580191805e78a3f2482a513bd) and this is the second change (https://github.com/kirillzyusko/react-native-keyboard-controller/pull/353/files#diff-7efeb6f32b8d7e3c36b551280577125c625c730580191805e78a3f2482a513bd)

After that Android tests are more stable than iOS 😂

alduzy (approved):

kkafar:
> What about FabricExample? Have you also yarned all the projects? 😄

you're absolutely right @tboba, I've somehow overlooked it - my bad. Fixed in https://github.com/software-mansion/react-native-screens/pull/2192/commits/54f53a4a88d50620487e62c2a2ede6fcacff78ff.

kkafar:
I'm merging, as there weren't more objections to handle. If you have some more remarks let me know, I'll fixup in followup.

PR #2193, opened 2024-06-18
fix(iOS, Fabric): add missing code for finding touch handler

kkafar:
## Description

`- [RNSScreen touchHandler]` method was missing implementation for Fabric, adding it here.

## Changes

- **Add category on UIView with code for touch handler finding**
- **Use the new category to simplify implementation**

`Test2118` - tested & works just fine

## Checklist

- [ ] Ensured that CI passes


ios/utils/UIView+RNSUtility.h
```diff
@@ -0,0 +1,23 @@
...
+#ifdef RCT_NEW_ARCH_ENABLED
+#import <React/RCTSurfaceTouchHandler.h>
+#else
+#import <React/RCTTouchHandler.h>
+#endif // RCT_NEW_ARCH_ENABLED
+
+#ifdef RCT_NEW_ARCH_ENABLED
+#define RNS_TOUCH_HANDLER_ARCH_TYPE RCTSurfaceTouchHandler
+#else
+#define RNS_TOUCH_HANDLER_ARCH_TYPE RCTTouchHandler
```
kkafar, line 12: I tried using `typedef`s instead of macros here, but stumbled upon [issues with nullability](https://developer.apple.com/swift/blog/?id=25) (section about "audited regions") and had unexpected issues doing any casting for typedef'd types -> went for macros.

kkafar (commented):
Just a note for reviewers.

WoLewicki (approved):
LGTM

ios/utils/UIView+RNSUtility.h
tboba, line 1: If we're keeping in this file only TouchHandler-specific code, I would call this file as `UIView+RNSTouchHandlers`.

ios/RNSScreenStack.mm
```diff
@@ -735,43 +737,8 @@ - (void)cancelTouchesInParent
...
-  UIView *parent = _controller.view;
-  while (parent != nil && ![parent respondsToSelector:@selector(touchHandler)]) {
-    parent = parent.superview;
-  }
-  if (parent != nil) {
-    RCTTouchHandler *touchHandler = [parent performSelector:@selector(touchHandler)];
-    [touchHandler rnscreens_cancelTouches];
-  }
-#endif // RCT_NEW_ARCH_ENABLED
+  [[self rnscreens_findTouchHandlerInAncestorChain] rnscreens_cancelTouches];
```
tboba, line 741: Can we do something like:

```obj-c
RNS_TOUCH_HANDLER_ARCH_TYPE *touchHandler = [self rnscreens_findTouchHandlerInAncestorChain];
[touchHandler rnscreens_cancelTouches];
```

to improve readability?

tboba (approved):
LGTM! Just left two comments out there 👍

ios/RNSScreenStack.mm
kkafar: I find it quite opposite 😄 I'm mean the redundant variable strikes me => I'll stay by current version as this is taste related, but thanks.

ios/utils/UIView+RNSUtility.h
kkafar: I used already existing category name - RNSUtility (we have already introduced this one), and I think it is better suited for future, as putting any more utility methods in `RNSTouchHandlers` wouldn't make sense. *It's again the matter of taste...* I'll keep current version, but thanks for suggestion!

PR #2194, opened 2024-06-18
chore: migrate examples to react-navigation v7

alduzy:
## Description

This PR intents to migrate examples to react-navigation v7.
Documentation: https://reactnavigation.org/docs/7.x/upgrading-from-6.x/

As for now I have only found ~~two~~ few files that needed changes. I ensured that nothing crashes in the example.

## Changes

- migrated navigating to nested screen
- replaced `navigate` with `popTo` when explicitly going back to a specific screen
- ensured nothing crashes

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/apps/examples/src/screens/Animations.tsx
+++ b/apps/examples/src/screens/Animations.tsx
@@ -113,7 +113,7 @@ const NavigateScreen = ({
 
   return (
     <View style={{ ...styles.container, backgroundColor: 'pink' }}>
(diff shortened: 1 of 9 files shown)


kkafar (approved):
I guess if these are only examples that you detected to have problems we should proceed with merging. 

We should just keep in mind, than when testing it might be possible that some tests require porting. 

Going through every single example, testing it & wondering whether something needs to be ported is a waste of time for now. I'll write this to backlog, but rn let's move forward.

PR #2196, opened 2024-06-19
chore: moved android e2e to ubuntu

alduzy:
## Description

This PR intents to move android e2e test to ubuntu to avoid random crashes. Inspired by changes made by @kirillzyusko in [react-native-keyboard-controller](https://github.com/kirillzyusko/react-native-keyboard-controller).

## Changes

- Changed workflow's `runs-on`
- Bump deps
- Added necessary extra steps

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/.github/workflows/android-e2e-test.yml
+++ b/.github/workflows/android-e2e-test.yml
@@ -54,10 +54,6 @@ jobs:
         name: Reinstall node dependencies
         working-directory: ${{ env.WORKING_DIRECTORY }}
         run: yarn
-      - name: Install AVD dependencies
-        run: |
-          sudo apt update
-          sudo apt-get install -y libpulse0 libgl1
       - name: Build app
         working-directory: ${{ env.WORKING_DIRECTORY }}
         run: yarn build-e2e-android
--- a/.github/workflows/android-e2e-test.yml
+++ b/.github/workflows/android-e2e-test.yml
@@ -52,6 +52,19 @@ jobs:
       - name: Build app
         working-directory: ${{ env.WORKING_DIRECTORY }}
         run: yarn build-e2e-android
+      - name: Enable KVM group perms
+        run: |
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
+      - name: AVD cache
+        uses: actions/cache@v4
+        id: avd-cache
+        with:
+          path: |
+            ~/.android/avd/*
+            ~/.android/adb*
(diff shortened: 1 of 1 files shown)


alduzy:
@kirillzyusko thanks! I saw that you added such step (Free Disk Space (Ubuntu) from: https://github.com/marketplace/actions/free-disk-space-ubuntu). I am going to add it later on, just curious to see how the workflow behaves without it 😄

kirillzyusko:
@alduzy I see that workflow is constantly failing - you may want to try to change `target` to `default` 👀 At least it's the only difference that I see that could cause CI failures 🤔 

I think I was the person who changed it to `aosp_atd` in RNS repository, but maybe on linux machine it should be again `default` 🤷‍♂️

alduzy:
Looks like the workflow is running stable after merging fixes from main branch.

.github/workflows/android-e2e-test.yml
```diff
@@ -51,20 +57,25 @@ jobs:
       - name: Build app
         working-directory: ${{ env.WORKING_DIRECTORY }}
         run: yarn build-e2e-android
+      - name: Enable KVM
+        run: |
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
```
kkafar: It would be best to understand why `kvm` is needed here and what do we do here 😄 

But if it works I guess we're just fine w/o it 🤷🏻‍♂️

kkafar (commented):
This looks really promising. Let's trigger the CI few times in a row by hand, so we can see whether it is more stable or not.

One thing I noticed is that you didn't setup AVD cache? Why don't we want this? Won't it speedup the CI?

.github/workflows/android-e2e-test.yml
alduzy: Testing the workflow without it right now 😄

alduzy: It is used by [android-emulator-runner](https://github.com/reactivecircus/android-emulator-runner?tab=readme-ov-file#running-hardware-accelerated-emulators-on-linux-runners) and it's necessary. I'm bringing it back

kkafar: Great find!

kkafar: What about the caches I asked in the main message?

kirillzyusko: In my case AVD cache helped to shave off 1-2 minutes when I boot the emulator. Curious to see the result in your project 👀

kkafar (approved):
Not gonna lie, 13 minutes seems awesome & it looks quite stable.

PR #2200, opened 2024-06-20
fix(examples): allow project node_modules to be used

kkafar:
## Description

The custom resolver implemented in our `metro.config.js` right now most likely does not conver all cases.
When working on my PR with modals for Android I've added `jotai` package and received errors from bundler that this package is not found,
because: 

1. it was not present in any nearby `node_modules` - the file that requries it resides now in `apps/test-examples`, thus application `node_modules` are not checked by default,
2. only our custom resolver looks into apps `node_modules` & fails to resolve this package. 

I haven't investigated why it does not work for `jotai` in particular, because more general solution is to just let the default module resolution algorithm look into
application `node_modules` as it should always do.

## Changes

Added application `node_modules` to the list of additional node modules in metro config of each example application.

Install `jotai` w/o this change, require it in any example and try to run the application.
Then apply changes from this PR and see it works well.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/Example/metro.config.js
+++ b/Example/metro.config.js
@@ -22,6 +22,8 @@ const modules = [
 
 const resolvedExts = ['.ts', '.tsx', '.js', '.jsx'];
 
+const projectNodeModules = path.join(__dirname, 'node_modules');
+
 const config = {
   projectRoot: __dirname,
   watchFolders: [rnsRoot],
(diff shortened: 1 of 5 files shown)


alduzy:
Looks good, if it fixes the crash then we should proceed.
I can still see `react-native-restart` import being underlined in `apps/examples/App.tsx` after I checkout to this branch, but I suppose it's a different problem to solve.

alduzy (approved):

kkafar:
Could you document it? Cause I just checked and don't see an issue on my end. I'm proceeding here. As you said, if there is some other probelm we will handle it separately.

PR #2207, opened 2024-06-24
feat(iOS): Implement blurEffect for new architecture

janicduplessis:
## Description

Implements blurEffect header config for new arch.

## Changes

 I kept this super simple and used string for codegen so we can simply reuse the RCTConvert implementation.

Tested that header blur now works in an app with new arch enabled.

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [x] Ensured that CI passes


src/fabric/ScreenStackHeaderConfigNativeComponent.ts
```diff
@@ -44,6 +44,7 @@ export interface NativeProps extends ViewProps {
   backButtonDisplayMode?: WithDefault<BackButtonDisplayMode, 'default'>;
   hideBackButton?: boolean;
   backButtonInCustomView?: boolean;
+  blurEffect?: string;
```
tboba, line 47: I don't know if I'm fond of setting blurEffect as a string. In our `types.tsx`, we're using a special [type](https://github.com/software-mansion/react-native-screens/blob/main/src/types.tsx#L42) for defining every possible string that could be a blurEffect. I'm concerned if there could be a situation, where invalid `blurEffect` entered here could just bring a crash on iOS. Some day ago, there was [that](https://github.com/software-mansion/react-native-screens/pull/1751) PR that was also adding a support for `blurEffect` on Fabric. Sadly, it has been abandoned, but maybe you could try to actually enter the type to the Spec and check if it'll just work the same as with `string`?

tboba (changes requested):
Hi, thanks for creating this PR! I left a comment about the way how we're defining that prop in a Spec.

janicduplessis:
@tboba Thanks for having a look. I will verify how incorrect values are handled. I liked the simplicity of this approach since it can use the same conversion code as the old arch, but I can also implement it using an enum, it will just require an extra function to convert from the codegen enum to the uikit one.

janicduplessis:
Looks like there is also an android build issue I will need to fix.

janicduplessis:
@tboba Update to use enum and fixed android build!

janicduplessis:
@tboba Friendly ping on this :)

tboba:
Hey hey @janicduplessis, yeah I'll try to check that today 👍 Thanks for the nudge :D

tboba (approved):
Everything looks good to me! 🫡 
Let's proceed when builds will pass 👍

PR #2208, opened 2024-06-24
chore(examples): use Material 3 in example applications

kkafar:
## Description

Following up review discussion:

* https://github.com/software-mansion/react-native-screens/pull/2045#discussion_r1650813540

to PR 

* #2045

This PR introduces usage of Material 3 in our example applications, so that it is easier to test new Sheets & top app bar APIs.

The 1.9.0 version was chosen, because:

1. This is the version I did all the testing on in #2045 
2. 1.10.0 introduces many [dependency bumps](https://github.com/material-components/material-components-android/releases/tag/1.10.0), including `appcompat` & `activity` libs, which are known to be problematic with updates (we already had some issues in the past related to that)

## Changes

- **Bump material to 1.9.0 in TestsExample**
- **Bump material to 1.9.0 in Example**
- **Bump material to 1.9.0 in FabricExample**
- **Bump material to 1.9.0 in FabricTestExample**

Application should work as normal & Material 3 components should be used. 

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/Example/android/app/build.gradle
+++ b/Example/android/app/build.gradle
@@ -132,6 +132,8 @@ dependencies {
     } else {
         implementation jscFlavor
     }
+
+    implementation("com.google.android.material:material:1.9.0")
 }
 
 apply from: file("../../node_modules/@react-native-community/cli-platform-android/native_modules.gradle"); applyNativeModulesAppBuildGradle(project)
--- a/Example/android/app/src/main/res/values/styles.xml
+++ b/Example/android/app/src/main/res/values/styles.xml
@@ -1,7 +1,7 @@
 <resources>
 
     <!-- Base application theme. -->
-    <style name="AppTheme" parent="Theme.AppCompat.DayNight.NoActionBar">
+    <style name="AppTheme" parent="Theme.Material3.DayNight.NoActionBar">
         <!-- Customize your theme here. -->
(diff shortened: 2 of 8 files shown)


Example/android/app/build.gradle
```diff
@@ -132,6 +132,8 @@ dependencies {
     } else {
         implementation jscFlavor
     }
+
+    implementation("com.google.android.material:material:1.9.0")
```
tboba, line 136: I'm worried if anyone that will try to install the Example app on Android will stumble upon the issue with too low version of Kotlin. Should we maybe add some comment here or create a README with prerequisites to install the app?

tboba (approved):
LGTM! Just left a small comment there. It's great to introduce this change now, so that we could check if setting Material 3 version is safe in RN apps for sure 👍

Example/android/app/build.gradle
ahmedawaad1804: iam working on a such integration but i think for new kotlin 2+ it works fine .

kkafar:
We wanna stay aligned with official app template & use ContextThemeWrapper where necessary to deploy material 3 styling with the lib.

PR #2214, opened 2024-06-26
fix(Android): missing negation in check for mismatched frames

kkafar:
## Description

The error check was bad - it reported error each time the frames were the same.
Obviously it shoud report frame mismatch.

## Changes

Added missing negation.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
+++ b/common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
@@ -51,7 +51,7 @@ class RNSScreenComponentDescriptor final
       // state update.
       if (screenShadowNode.getFrameCorrectionModes().check(
               FrameCorrectionModes::Mode::FrameHeightCorrection) &&
-          compareFrameSizes(
+          !compareFrameSizes(
               screenShadowNode.layoutMetrics_.frame.size,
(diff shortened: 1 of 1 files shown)


common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
```diff
@@ -51,7 +51,7 @@ class RNSScreenComponentDescriptor final
       // state update.
       if (screenShadowNode.getFrameCorrectionModes().check(
               FrameCorrectionModes::Mode::FrameHeightCorrection) &&
-          compareFrameSizes(
+          !compareFrameSizes(
```
alduzy, line 54: We could consider changing the name of this function to something like `checkIsFrameSizeEqual` so that it's obvious what it returns.

alduzy (approved):
LGTM 👍 
One refactor to consider

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
kkafar: I think this is good suggestion. `compare` may suggest inequality, right? 
I'll do it in separate PR, thanks!

PR #2116, opened 2024-04-29
fix(Android): fix build deprecations

alduzy:
## Description

This PR intents to fix some "Deprecated in Java" warnings when building on android. I decided not to suppress warnings manually when it is inescapable to avoid the warning.

Fixes #2018 

## Changes

- replaced deprecated targetElavation with elevation
- adjusted usage of porter duff color filter
- calculating header top inset using recent method if android API >= 30 (the warning persists as the previous APIs still require deprecated method)
- fixed guarded runnable deprecated warning
- removed unused import

### Before

```
> Task :react-native-screens:compileDebugKotlin
'compileDebugJavaWithJavac' task (current target is 11) and 'compileDebugKotlin' task (current target is 17) jvm target compatibility should be set to the same Java version.
Consider using JVM toolchain: https://kotl.in/gradle/jvm/toolchain

w: file:///Users/alexduzy/Projects/react-native-screens/Example/node_modules/react-native-screens/android/src/main/java/com/swmansion/rnscreens/Screen.kt:83:22 'constructor GuardedRunnable(ReactContext!)' is deprecated. Deprecated in Java
w: file:///Users/alexduzy/Projects/react-native-screens/Example/node_modules/react-native-screens/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt:65:27 'setter for targetElevation: Float' is deprecated. Deprecated in Java
w: file:///Users/alexduzy/Projects/react-native-screens/Example/node_modules/react-native-screens/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt:128:27 'setter for targetElevation: Float' is deprecated. Deprecated in Java
w: file:///Users/alexduzy/Projects/react-native-screens/Example/node_modules/react-native-screens/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.kt:86:34 'getter for systemWindowInsetTop: Int' is deprecated. Deprecated in Java
w: file:///Users/alexduzy/Projects/react-native-screens/Example/node_modules/react-native-screens/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.kt:231:37 'setColorFilter(Int, PorterDuff.Mode): Unit' is deprecated. Deprecated in Java
[...]

PR #2216, opened 2024-06-27
fix(Android): header shadow not hidden on go back

### LINKED ISSUE #2212: [Android] headers show strange artifacts after unfocus -> focus since v3.32
### Description

After updating to react-native 0.73 I also updated many packages that weren't locked to exact versions, including react-native-screens, which introduced a bug (android only) in how headers are rendered. It only occurs when focusing a screen for the second time (the initial render looks good, but navigating away and back leads to a 'ghost' header underneath the original header. It's mostly visible because the 'ghost' header appears to have shadows, even though my app does not have shadows below headers. Every screen that renders a native stack header is affected. 

Initial render (how it's supposed to look):
image

After switching to a different screen (tab or stack does not matter) and going back:
image

### Steps to reproduce

1. Run the following code on an android emulator/device with dependencies (or checkout the reproduction in https://github.com/arjovanramshorst/react-native-screens-bug-repro)

```
// package.json
{
    "react-native-screens": "3.32.0",
    "@react-navigation/native": "^6.1.6",
    "@react-navigation/native-stack": "^6.9.12",
}
```

```typescript
import React from 'react';
import {NavigationContainer} from '@react-navigation/native';
import {Button, Text, View} from 'react-native';

export const StorybookNavigation: React.FC<{}> = () => {
  return (
    <NavigationContainer>
      <Stack.Navigator>
        <Stack.Screen
          name="Screen1"
          options={{title: 'Screen 1', headerTransparent: true}}
          component={Screen1}
        />
        <Stack.Screen
          name="Screen2"
          options={{title: 'Screen 2', headerTransparent: true}}
          component={Screen2}
        />
      </Stack.Navigator>
    </NavigationContainer>
  );
};

const Screen1 = ({navigation}: any) => {
  return (
    <View
      style={{
        flex: 1,
        alignItems: 'center',
        justifyContent: 'center',
        backgroundColor: 'white',
      }}>
      <Text>Screen 1 </Text>
      <Button
        title={'Go to screen 2'}
        onPress={() => {
          navigation.push('Screen2');
        }}
      />
    </View>
  );
};

const Screen2 = () => {
  return (
    <View
      style={{
        flex: 1,
        backgroundColor: 'white',
        alignItems: 'center',
        justifyContent: 'center',
      }}>
      <Text>Screen 2 </Text>
    </View>
  );
};
```

2. Navigate to screen 2 and back to screen 1, and notice the difference between before and after navigating. It is very apparent with `headerTransparent: true`, however, it is also visible with other changes to options.

https://github.com/arjovanramshorst/react-native-screens-bug-repro

### Screens version

3.32.0

### React Native version

0.73.8

### Platforms

Android

### JavaScript runtime

None

### Workflow

React Native (without Expo)

### Architecture

None

### Build type

Release mode

### Device

None

### Device model

Every device (emulator/real/debug/release build)

### Acknowledgements

Yes

alduzy:
## Description

This PR intents to fix header shadow not being hidden after navigating back to a screen. The bug was visible when using `headerTransparent: true` option.

The origin of this bug is [This PR](https://github.com/software-mansion/react-native-screens/pull/2116) fixing build depracations. Setting the `elevation` on appBarLayout requires resetting the `stateListAnimator` to work properly, as opossed to deprecated `targetElevation` used before.

For more info visit: https://developer.android.com/reference/com/google/android/material/appbar/AppBarLayout#setTargetElevation(float)

Fixes #2212 .

## Changes

- added missing `appBarLayout?.stateListAnimator` reset.

## Screenshots / GIFs

### Before
Screenshot 2024-06-27 at 11 39 50

### After
Screenshot 2024-06-27 at 11 39 29

1. use `Test1097.tsx` repro
2. navigate into second or third screen
3. go back to the first screen

## Checklist

- [x] Ensured that CI passes


kkafar (approved):
I think this change is ok, because it improves the situation. 

However:

> [!important]
We should research whether setting stateListAnimator to null here and in the other setters do not remove some system/material defined appbar animations. Let's put a task on our board for that.

tboba:
Just a small comment from me that's a great change! Good catch, Alex 💪 I was wondering why header has such an ugly shadow after setting its style to transparent.

PR #2224, opened 2024-07-02
chore: change gradle task for copying to new archs into JS scripts

maciekstosio:
## Description

This task is rewrite to JS of https://github.com/software-mansion/react-native-screens/pull/2168 with some improvements. General point of creating those tasks is: 

> When changing native props on Fabric, codegen generates corresponding interfaces and delegates. To make sure both implementations are consistent, we implement those interfaces on Paper too. Currently, after generating interfaces using codegen, developer needs to copy corresponding files for paper manually. This task adds Gradle task, that automates this.

## Changes
Current assumption: 
Two scripts: `check-archs-consistency` and `sync-archs`. The first one generates codegen interfaces and compares them with what we have for paper, the second generates and copies for paper to sync the archs.
- sync is run pre build on example app
- sync is run when staged on changes to `src/paper`
- check is run on CI when `src/paper` or `android/src/paper/java/com/facebook/react/viewmanagers` changes

What it improves:
- JS tasks are much faster, as codegen is JS script anyway, we skip gradle and java setup all together (CI task down from 7min to 30s), 
- we do not put code to library, so it shouldn't be possible to mess up something for end users,
- instead of syncing archs when running codegen we do that on paper example build and when staged, so: when developer didn't touch the code it will have changes after commit, when developer switched to working on paper interfaces should be always up to date when building the app.

Open `src/fabric/ScreenStackHeaderConfigNativeComponent.ts` and remove any proper form interface. Now: 
- when building paper, interface should be updated
- when committing, interface should be updated
- if committed and pushed, Test consistency between Paper & Fabric  should fail :)
Brining back the prop and repeating up should cause the interface back and CI green.
Posting changes in other places should cause CI task to run. 

You can also run those commands yourself using `yarn check-archs-consistency` and `yarn sync-archs`


.github/workflows/check-paper-integrity.yml
```diff
@@ -0,0 +1,20 @@
+name: Test Paper Architecture integrity
+on: [pull_request]
```
kkafar, line 2: I would suggest running this CI on PRs to main & when files inside `src/fabric` or `android/src/paper/java/.../viewmanagers` are edited.

.github/workflows/check-paper-integrity.yml
```diff
@@ -0,0 +1,20 @@
...
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: Use Node.js 18
+        uses: actions/setup-node@v2
+        with:
+          node-version: 18
+          cache: 'yarn'
+      - name: Install node dependencies
+        run: yarn
+      - name: Install node dependencies
```
kkafar, line 19:
```suggestion
      - name: Check Android Paper & Fabric generated interfaces consistency
```

kkafar: or something like that

.github/workflows/check-paper-integrity.yml
```diff
@@ -0,0 +1,20 @@
...
+        run: yarn check-integrity
```
kkafar, line 20:
```suggestion
        run: yarn check-integrity
        
```

package.json
```diff
@@ -18,7 +18,9 @@
     "lint-android": "./android/gradlew -p android spotlessCheck -q",
     "lint": "yarn lint-js && yarn lint-android",
     "release": "yarn prepare && npm login && release-it",
-    "prepare": "bob build && husky install"
+    "prepare": "bob build && husky install",
+    "check-integrity": "node ./scripts/codegen-check-integrity.js",
```
kkafar, line 22: "integrity" word does not make much sense in context of what this script is doing. "cohesion" or even better - "consistency" is the word you're looking for.

kkafar (commented):
Just initial remarks - I'll have more once the PR is opened & ready for review. The approach is good.

package.json
maciekstosio: Fixed in https://github.com/software-mansion/react-native-screens/pull/2224/commits/8e4f7a0e7df8b0768a9fc33eecae16c4be3f7a50

.github/workflows/check-paper-integrity.yml
maciekstosio: Fixed in https://github.com/software-mansion/react-native-screens/pull/2224/commits/8e4f7a0e7df8b0768a9fc33eecae16c4be3f7a50

scripts/codegen-check-consistency.js
```diff
@@ -0,0 +1,8 @@
+const { checkCodegenIntegrity } = require('./codegen-utils');
+
+if (require.main === module) {
+  async function main() {
+    await checkCodegenIntegrity();
+  }
+  main();
```
j-piasecki, line 7: Cannot it be called directly?

scripts/codegen-utils.js
```diff
@@ -0,0 +1,133 @@
...
+  );
+
+  fixOldArchJavaForRN72Compat(`${GENERATED_DIR}/source/codegen/java/`);
+}
+
+async function generateCodegenJavaOldArch() {
+  await generateCodegen();
+
+  const generatedFiles = fs.readdirSync(CODEGEN_FILES_DIR);
+  const oldArchFiles = fs.readdirSync(OLD_ARCH_FILES_DIR);
```
j-piasecki, line 73: Just to be sure, it does deep traversal, right?

scripts/codegen-utils.js
```diff
@@ -0,0 +1,133 @@
...
+
+  const existingFilesMap = Object.fromEntries(
+    oldArchFiles.map(fileName => [fileName, 1])
+  );
```
j-piasecki, line 77: Why not `Set`?

scripts/codegen-utils.js
maciekstosio: Yeah `(...) reads the content of a folder, both files and subfolders, and returns their relative path:`

scripts/codegen-utils.js
maciekstosio: Fixed in https://github.com/software-mansion/react-native-screens/pull/2224/commits/1b2a000a3c5238fd142f95ddb7119e0a216f0ed8

scripts/codegen-check-consistency.js
maciekstosio: Fixed in https://github.com/software-mansion/react-native-screens/pull/2224/commits/1b2a000a3c5238fd142f95ddb7119e0a216f0ed8

scripts/codegen-utils.js
```diff
@@ -0,0 +1,123 @@
...
+  await generateCodegen();
+
+  const generatedFiles = fs.readdirSync(CODEGEN_FILES_DIR);
+  const oldArchFiles = fs.readdirSync(OLD_ARCH_FILES_DIR);
+  const existingFilesSet = new Set(oldArchFiles.map(fileName => fileName));
+
+  generatedFiles.forEach(generatedFile => {
+    if (!existingFilesSet.has(generatedFile)) {
+      console.warn(
+        `[RNScreens] ${generatedFile} not found in paper dir, if it's used on Android you need to copy it manually and implement yourself before using auto-copy feature.`
```
WoLewicki, line 79:
```suggestion
        `[<package_name_as_variable>] ${generatedFile} not found in paper dir, if it's used on Android you need to copy it manually and implement yourself before using auto-copy feature.`
```

Would be better for easy porting? Same in other places with `RNSCREENS`

Example/android/app/build.gradle
```diff
@@ -135,3 +135,13 @@ dependencies {
 }
 
 apply from: file("../../node_modules/@react-native-community/cli-platform-android/native_modules.gradle"); applyNativeModulesAppBuildGradle(project)
+apply plugin: 'com.github.node-gradle.node'
+
+task syncArchs(type: NodeTask) {
```
WoLewicki, line 140: I don't get it why it's still here. Don't we do everything on the JS side now?

WoLewicki (commented):
Added 2 comments

Example/android/app/build.gradle
maciekstosio: It's in the app so it shouldn't interfere with the lib. I understood that @kkafar wanted to have it run automatically while developing. It needs to be added per example/test app, so I can just not add it in other packages.

j-piasecki (approved):
I don't think I have anything else to add here 😄

Example/android/app/build.gradle
kkafar: Yeah & I want this in screens.

scripts/codegen-utils.js
```diff
@@ -0,0 +1,124 @@
...
+
+  oldArchFiles.forEach(file => {
+    exec(`cp -rf ${CODEGEN_FILES_DIR}/${file} ${OLD_ARCH_FILES_DIR}/${file}`);
+  });
+}
+
+function compareFileAtTwoPaths(filename, firstPath, secondPath) {
+  const fileA = fs.readFileSync(`${firstPath}/${filename}`, 'utf-8');
+  const fileB = fs.readFileSync(`${secondPath}/${filename}`, 'utf-8');
+
```
kkafar, line 107: Yeah, I guess it is fine, as our files are up to 2k lines long. 

In case of any problems we can use md5sum here, but that's not necessary for now.

kkafar (approved):
Looks good now, thanks 👍🏻

scripts/codegen-utils.js
maciekstosio: I assumed that you need to get the file to hash it anyway, so don't know where the benefit would be? Comparison itself?

.github/workflows/check-archs-consistency.yml
```diff
@@ -0,0 +1,26 @@
...
+  check:
+    runs-on: ubuntu-latest
+    concurrency:
+      group: check-archs-consistency-${{ github.ref }}
+      cancel-in-progress: true
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: Use Node.js 18
+        uses: actions/setup-node@v2
```
tboba, line 19: Let's use `v4` script versions.

```suggestion
        uses: actions/checkout@v4
      - name: Use Node.js 18
        uses: actions/setup-node@v4
```

scripts/codegen-utils.js
tboba, line 1: Great job with this one!

tboba (approved):
Just a one suggestion from me. Great change!

kkafar:
Just leaving a note that the scripts are failing on CI, let's make sure they work consistently before we merge here.

kkafar (approved):
Looks solid. We'll sort out if there is something to improve while using it. We can afford such approach as these changes should not affect end users in any way. Thanks! 🎉

.github/workflows/check-archs-consistency.yml
tboba: :(

PR #2225, opened 2024-07-02
refact: merge example and test example apps

alduzy:
## Description

This PR intents to merge together the example and test-example apps resulting in less configuration files.
TestsExample and FabricTestExample apps are removed so there's only Example, FabricExample and TVOSExample app to take care of.

it is more convenient to use Example and FabricExample apps to run tests as there is less configuration files to take care of, less building and less metro restarting. Simply replace the returned component in `apps/App.tsx` with the selected test and you're ready for testing. For example replace `<Example/>` with `<Test.Test42/>`.

## Changes

- removed TestsExample and FabricTestExample apps
- allowed for source change in `apps/App.tsx` file to switch between example and tests

Run the Example or FabricExample app and change source in `apps/App.tsx` file.

## Checklist

- [x] Updated documentation: 
  - [x] https://github.com/software-mansion/react-native-screens/blob/main/guides/CONTRIBUTING.md
- [x] Ensured that CI passes


apps/TestApp.tsx
maciekstosio, line 1: Wouldn't it be better to add some TestExamples list like with normal examples?

guides/CONTRIBUTING.md
```diff
@@ -73,14 +71,12 @@ Submitting Pull Requests that resolve issues is great way to contribute to Scree
...
+We have two types of sources: pure examples (apps/App for source code) and examples, dedicated for tests from issues / pull requests (apps/TestApp for source code). The first work as a showcase of the library, the latter contains specific test cases that corresponds to GitHub issues. For example, `Test1864.tsx` corresponds to issue [#1864](https://github.com/software-mansion/react-native-screens/issues/1864). Our developer flow usually consists of creating new `Test*.tsx` file with code example that we try to fix or add. For new features we try to prepare showcases in Example app. Example and FabricExample applications are used as wrappers for those applications.
```
maciekstosio, line 74: I'd rewrite last 3 sentences now they're missing the point imo. We create testing example in TestApp source and showcases in example app. The wrapper depends on arch, don't think if that's clear now for someone who didn't use it before

apps/TestApp.tsx
alduzy: I thought about it. Although it will look well polished I think it'll be a pain to navigate to the current Test each time the app rebuilds during debugging and development. WDYT?

alduzy: https://github.com/software-mansion/react-native-screens/pull/2225/commits/1afa70a1a14943e67aaaba5d766b25e210ceb183

guides/CONTRIBUTING.md
alduzy: https://github.com/software-mansion/react-native-screens/pull/2225/commits/a415eca814b16a2cf4bde49fd087c80fb2d24973

apps/App.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { enableFreeze } from 'react-native-screens';
+import Example from './Example';
+import * as Test from './src/tests';
+
+enableFreeze(true);
+
+export default function App() {
+  return <Example />;
```
maciekstosio, line 9: Maybe add here something like
```
// return <Test.Test42 />
```
and some comments. So it's possible to use it without sorcery 101.

alduzy: @maciekstosio That's a great idea! https://github.com/software-mansion/react-native-screens/pull/2225/commits/393b1cc263c368569821d8b0ed5149c572c418b8

maciekstosio (approved):
lgtm 🎉

apps/src/tests/index.ts
```diff
@@ -0,0 +1,109 @@
+// @ts-nocheck
+
+export { default as Test42 } from './Test42';
```
kkafar, line 3:
```suggestion
export { Test42 } from './Test42';
```

Shouldn't it be enough?

kkafar (commented):
Yeah, this looks nice. Thank you!

Just few remarks & we'll be good to go.

apps/src/tests/index.ts
kkafar: Yeah, I've tested it, should be enough.

kkafar: Oh, you're right, we export each test as `default` from each test module & we need to reexport it in such way. Okey, that's fine then. 👍🏻

kkafar (approved):
Okay, I've answered my concerns myself 😄 Looks good 💪🏻

PR #2226, opened 2024-07-02
refact: remove unnecessary workflow step

alduzy:
## Description

I noticed there's an unnecessary step in the worklfow both for iOS and android e2e tests. 
The step is supposed to install any dependencies inside `examples/app` but there's no `package.json` file.
This PR removes the step.

## Changes

- removed workflow step

## Checklist

- [x] Ensured that CI passes


kkafar (approved):
Looks like a good catch. Thanks.

PR #2229, opened 2024-07-03
Fix bug with header hide prop on the screen under the modal

kuczi55:
## Description
When we push multiple screens with different configuration of `headerShown` prop (e.g. 4 screens with visible header and one screen with hidden header), open a modal and trigger rerenders on screens (e.g. by changing the theme) there's a high chance that value of header hide prop will be incorrect on the screen under the modal.

Big kudos to @WoLewicki for helping with debugging!

## Changes
`isPresentingVC` flag wasn't checking if current view controller was on the top, which was causing header updates for all screens in the stack (moreover it was done in non-deterministic order as context updates can cause non-deterministic rerenders of particular screens), so check `&& vc == nav.topViewController` was added to `isPresentingVC` flag in `ios/RNSScreenStackHeaderConfig.mm` file. After that change, we still update the screen which is directly under the modal, so fix merged in https://github.com/software-mansion/react-native-screens/pull/1228 is still valid.

### Before

https://github.com/software-mansion/react-native-screens/assets/44415389/cf9b07fc-4e95-4362-95c7-ab387a94148e

### After

https://github.com/software-mansion/react-native-screens/assets/44415389/94a5dd72-498b-4959-a2de-1e6008d63895

1. `Test2229` was added, so just render it in `TestsExample` app
2. Push a few screens with header
3. Push screen without header
4. Open modal
5. Background and open the app or change theme
6. Close the modal
7. Before the fix header should be visible on the screen although it should be a screen without a header

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


WoLewicki (approved):
LGTM!

kkafar (approved):
Looks good, thanks!

PR #2234, opened 2024-07-05
fix(iOS): push and pop transitions change after full screen back swipe 

### LINKED ISSUE #2227: When enabled fullScreenGestureEnabled, the animation changed from Native one to React Navigation animation once goes back triggered
### Description

Basically the title.
In my project, I enabled `fullScreenGestureEnabled` for stack options. I went on using the app with no problem but once I go back then go into it again or any other screens, the animation changes from Native behaviour to React Navigation one.

### Steps to reproduce

1. set fullScreenGestureEnabled to true.
2. App must have more than one stack screen.
3. Navigate to one, then swipe back.
4. Navigate again, the animation looks different.

https://github.com/xiaomei7/ReactScreenBugExample

### Screens version

3.27.0

### React Native version

0.73.8

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

maksg:
## Description

Doing the swipe also caused programmatic push to use manual transition and have no shadow.
Fixes #2227.

## Changes

Set `_isFullWidthSwiping` to `NO` when swipe gesture ends in `handleSwipe:` so programmatic push and pop can use OS logic by default.

### Before
https://github.com/software-mansion/react-native-screens/assets/11507974/0f710aec-c8f0-4a3a-82df-d3edb5f0fb19

### After
https://github.com/software-mansion/react-native-screens/assets/11507974/51119c74-b2c3-4c32-9202-6954c3c70218

1. Run `Test2227` in TestsExample app.
2. Push screen with `Go to Details`.
3. Swipe back with full screen swipe gesture.
4. Push screen with `Go to Details` again.
7. Before this fix there would be no shadow during transition after first full screen swipe back.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStack.mm
+++ b/ios/RNSScreenStack.mm
@@ -872,6 +872,7 @@ namespace react = facebook::react;
         [_interactionController cancelInteractiveTransition];
       }
       _interactionController = nil;
+      _isFullWidthSwiping = NO;
     }
     default: {
       break;
--- a/apps/test-examples/App.js
+++ b/apps/test-examples/App.js
@@ -104,6 +104,7 @@ import Test2069 from './src/Test2069';
 import Test2118 from './src/Test2118';
 import Test2184 from './src/Test2184';
 import Test2223 from './src/Test2223';
+import Test2227 from './src/Test2227';
 import Test2229 from './src/Test2229';
 import TestScreenAnimation from './src/TestScreenAnimation';
 import TestHeader from './src/TestHeader';
--- a/apps/test-examples/src/Test2227.tsx
+++ b/apps/test-examples/src/Test2227.tsx
@@ -1,0 +1,57 @@
+import * as React from 'react';
+import {NavigationContainer} from '@react-navigation/native';
+import {
+  createNativeStackNavigator,
+  NativeStackScreenProps,
+} from '@react-navigation/native-stack';
+import {Button, View, Text} from 'react-native';
+
+type RootStackParamList = {
+  Home: undefined;
+  Details: undefined;
+};
+
+type HomeScreenProps = NativeStackScreenProps<RootStackParamList, 'Home'>;
+type DetailsScreenProps = NativeStackScreenProps<RootStackParamList, 'Details'>;
+
+const Stack = createNativeStackNavigator<RootStackParamList>();
+
+function HomeScreen({navigation}: HomeScreenProps) {
+  return (
+    <View style={{flex: 1, justifyContent: 'center', alignItems: 'center'}}>
+      <Text>Home Screen</Text>
+      <Button
+        title="Go to Details"
+        onPress={() => navigation.navigate('Details')}
+      />
+    </View>
+  );
+}
+
+function DetailsScreen({navigation}: DetailsScreenProps) {
+  return (
+    <View style={{flex: 1, justifyContent: 'center', alignItems: 'center'}}>
+      <Text>Details Screen</Text>
+      <Button
+        title="Go to Details Again"
+        onPress={() => navigation.push('Details')}
+      />
+      <Button title="Go Back" onPress={() => navigation.goBack()} />
+      <Button title="Go to Home" onPress={() => navigation.navigate('Home')} />
+    </View>
+  );
+}
+
+export default function App() {
+  return (
+    <NavigationContainer>
+      <Stack.Navigator
+        screenOptions={{
+          fullScreenGestureEnabled: true,
+        }}>
+        <Stack.Screen name="Home" component={HomeScreen} />
+        <Stack.Screen name="Details" component={DetailsScreen} />
+      </Stack.Navigator>
+    </NavigationContainer>
+  );
+}


ios/RNSScreenStack.mm
```diff
@@ -752,7 +752,7 @@ - (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
   [self cancelTouchesInParent];
   return YES;
 #else
-  // RNSPanGestureRecognizer will receive events iff topScreen.fullScreenSwipeEnabled == YES;
+  // RNSPanGestureRecognizer will receive events if topScreen.fullScreenSwipeEnabled == YES;
```
kkafar, line 755: The `iff` is intentional - it is widely used, especially in mathematics & stands for `if and only if` (`wtedy i tylko wtedy`) 😄 

```suggestion
  // RNSPanGestureRecognizer will receive events iff topScreen.fullScreenSwipeEnabled == YES;
```

ios/RNSScreenStack.mm
```diff
@@ -872,6 +872,7 @@ - (void)handleSwipe:(UIPanGestureRecognizer *)gestureRecognizer
         [_interactionController cancelInteractiveTransition];
       }
       _interactionController = nil;
+      _isFullWidthSwiping = NO;
```
kkafar, line 875: Okay, so the main issue (change of transition style after popping a screen with swipe for the very first time) was caused by not properly reset state. Great finding!

ios/RNSScreenStackAnimator.mm
```diff
@@ -105,28 +105,40 @@ - (void)animateSimplePushWithTransitionContext:(id<UIViewControllerContextTransi
...
     toViewController.view.transform = rightTransform;
     [[transitionContext containerView] addSubview:toViewController.view];
+    [[transitionContext containerView] insertSubview:shadowView belowSubview:toViewController.view];
+    shadowView.alpha = 0.0;
+
     [UIView animateWithDuration:[self transitionDuration:transitionContext]
         animations:^{
           fromViewController.view.transform = leftTransform;
           toViewController.view.transform = CGAffineTransformIdentity;
+          shadowView.alpha = 0.1;
```
kkafar, line 121: My only suggestion here would be to extract this target alpha value to a constant, see the top of the `RNSScreenStackAnimator.mm` file (maybe use `constexpr` instead of static const) & if there is any reasoning behind this value (e.g. "Looked best in experimentation") - leave a comment behind.

kkafar (commented):
Left a tine remark, other than that - great job!

ios/RNSScreenStack.mm
maksg: Ah, haven't heard about that one before. That's good to know, thanks for enlightening me! 😃 I will revert that change

ios/RNSScreenStackAnimator.mm
maksg: Sure thing, I'll extract it and I'll add a comment a comment with explanation

ios/RNSScreenStackAnimator.mm
```diff
@@ -9,6 +9,9 @@
 static const float RNSSlideCloseTransitionDurationProportion = 0.25 / 0.35;
 static const float RNSFadeCloseTransitionDurationProportion = 0.15 / 0.35;
 static const float RNSFadeCloseDelayTransitionDurationProportion = 0.1 / 0.35;
+// same value is used in other projects using similar approach for transistions
+// and it looks the most similar to the value used by Apple
+constexpr float RNSShadowViewMaxAlpha = 0.1;
```
kkafar, line 14: Idk why I suggested sole `constexpr` here, let's make it `static` also, so that the definition is visible only in this translation unit.

```suggestion
static constexpr float RNSShadowViewMaxAlpha = 0.1;
```

kkafar (approved):
One last thing, and we're good to go.

WoLewicki (commented):
I am afraid doing it like this is a breaking change. When people use `fullScreenSwipe` right now, they don't have the shadow included. With those changes, it will appear which can be not expected by the users. Shouldn't we make this an optional prop of screen/navigator instead?

kkafar:
I considered it rather a fix, than breaking the behaviour. Although I can see your point.

One thing for sure: the change responsible for using default animation for programming push / pop *after* popping a screen with a gesture is a fix & it should be merged as is. 

The code responsible for adding a dimming view for swipe-transition might be hidden behind a prop, as opt-in.

@maksg @WoLewicki Let's split this PR into two separate ones. Does that sound right?

maksg:
I moved shadow view implementation to #2239

kkafar:
@jwoodmansey Can you post a video where you trigger the buggy behaviour? Maybe we can figure out the scenario in which in happens together somehow. 

Despite the fact, this PR improves the situation. We should proceed @maksg.

kkafar (approved):

PR #2239, opened 2024-07-08
feat(iOS): add shadow to custom push pop transitions

maksg:
## Description

When using full screen swipe back there was no shadow under the view.
Related PR #2234 with discussion about this change.

Corresponding PR in `react-navigation`:

* https://github.com/react-navigation/react-navigation/pull/12053

This PR add `fullScreenSwipeShadowEnabled` prop.

## Changes

Added shadow to transition in animateSimplePushWithTransitionContext:toVC:fromVC:.

1. Run `Test2227` in TestsExample app.
2. Push screen with Go to Details.
3. Swipe back with full screen swipe gesture.
5. Before this fix there would be no shadow during transition with full screen swipe back.

## Checklist

- [ ] Updated TS types
- [x] Updated documentation: 
- [ ] Ensured that CI passes


guides/GUIDE_FOR_LIBRARY_AUTHORS.md
```diff
@@ -50,6 +50,12 @@ Defaults to `false`. When `enableFreeze()` is run at the top of the application
 
 Boolean indicating whether the swipe gesture should work on whole screen. Swiping with this option results in the same transition animation as `simple_push` by default. It can be changed to other custom animations with `customAnimationOnSwipe` prop, but default iOS swipe animation is not achievable due to usage of custom recognizer. Defaults to `false`.
 
+### `fullScreenSwipeShadowEnabled` (iOS only)
+
+Boolean indicating whether the full screen dismiss gesture has shadow under view during transition. The gesture uses custom transition and thus
+doesn't have a shadow by default. When enabled a custom shadow view is added during the transition which tries to mimick the
```
WoLewicki, line 56:
```suggestion
doesn't have a shadow by default. When enabled, a custom shadow view is added during the transition, which tries to mimick the
```

WoLewicki: Same for other occourences of this comment.

src/native-stack/views/NativeStackView.tsx
```diff
@@ -216,6 +217,9 @@ const RouteView = ({
     if (fullScreenSwipeEnabled === undefined) {
       fullScreenSwipeEnabled = true;
     }
+    if (fullScreenSwipeShadowEnabled === undefined) {
+      fullScreenSwipeShadowEnabled = true;
```
WoLewicki, line 221: If you do it like this, isn't it `true` by default?

WoLewicki (commented):
Left some comments.

src/native-stack/views/NativeStackView.tsx
maksg: I was thinking the same but `fullScreenSwipeEnabled` does it this way and it also is `false` by default. I think it is set elsewhere as `false` so this doesn't matter. But I can change that and maybe `fullScreenSwipeEnabled` also.

kkafar: But the `false` value must be set somewhere right? Better to make this explicit.

kkafar (commented):
The things we lack here:

1. Update codegen interfaces for Android: run build on FabricExample & copy interfaces for Paper **or** wait for merge of [this PR](https://github.com/software-mansion/react-native-screens/pull/2224) and then trigger build on Android / add a commit.
2. Add a stub implementation for Android as done [here](https://github.com/software-mansion/react-native-screens/blob/151c1c74cfcc4ae4e6cbf987b64472be8d17dc9c/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt#L259-L262), for example.

src/native-stack/views/NativeStackView.tsx
maksg: Maybe it would be better to remove this altogether? Now I see that this is only set when `swipeDirection === 'vertical'` so it doesn't make sense to keep it there if it's not set for other swipe directions. Unless I should move it out of this `if`.

WoLewicki: I think this prop is not dependent on any other props so we can remove it?

src/native-stack/views/NativeStackView.tsx
```diff
@@ -202,6 +202,7 @@ const RouteView = ({
   let {
     customAnimationOnSwipe,
     fullScreenSwipeEnabled,
+    fullScreenSwipeShadowEnabled,
```
kkafar, line 205:
```suggestion
    fullScreenSwipeShadowEnabled = false,
```

src/native-stack/views/NativeStackView.tsx
```diff
@@ -224,6 +225,10 @@ const RouteView = ({
     }
   }
 
+  if (fullScreenSwipeShadowEnabled === undefined) {
+    fullScreenSwipeShadowEnabled = false;
+  }
```
kkafar, line 230:
```suggestion
```

react-navigation
kkafar, line 1: The problem with bumping `react-navigation` in such way is that it will be overridden by next PR that tries to do the same. We need a PR to `react-navigation`, merge it there to main & bump `react-navigation` here to new main revision. 

This change should be reverted & we should wait for progress on `react-navigation` PR.

kkafar:
Hey, if we are to proceed this one, you need to bump react navigation to it's recent main & incorporate these changes here. Obviously also resolve the merge conflicts.

kkafar (approved):
Okay, I think we're good here. Great job! 🎉

PR #2241, opened 2024-07-09
chore: remove unused code in podspec

WoLewicki:
## Description

Since we support RN from 0.71, and `install_modules_dependencies` is already available there, we can remove the unused code.


maciekstosio (approved):

tboba (approved):
That's good! Have you tested if project with RN 0.71 installed works correctly on Paper/Fabric with this change?

WoLewicki:
Nope

tboba:
Let me check that

tboba:
Everything seems to work fine! Unless there are other changes, I think we can proceed.

tboba:
Everything seems to work fine there! I think we can proceed.

kkafar (approved):
Yeah, the code changes do look ok. If we're good down to 0.71 (or even 0.72, as that will be the last version we support with release of 0.75) then we're good here.

PR #2245, opened 2024-07-11
fix: removed deprecated test that caused fabric bundle failure

alduzy:
## Description

This PR removes `Test898`. Importing this test in `apps/src/tests/index.ts` caused FabricExample bundling failure.
The test itself was introduced by this [PR](https://github.com/software-mansion/react-native-screens/pull/898) in order to test reanimated v2. Idk why the import causes bundle to fail, but it seems reasonable to just remove it rather than debug further as the test is redundant since there is a newer version of reanimated.

## Changes

- remove `Test898`

- try running any test using FabricExample when the import is there and when it's not

## Checklist

- [x] Ensured that CI passes


tboba (approved):

tboba:
ok then

kkafar (approved):
We don't have dependency on rea2 anyway.

PR #2247, opened 2024-07-15
fix(iOS): flickering custom header items

### LINKED ISSUE #2243: Header views flicker when navigating back on new arch
### Description

When navigating back via `navigation.goBack` the `headerTitle` view from the top screen disappears during the transition. Note this only happens when navigating from JS, not when navigating via the native back button. This also only happens on new arch.

You can observe the behaviour in the following video:

https://github.com/software-mansion/react-native-screens/assets/2677334/c3e404a0-06b9-402b-a214-3416c48b5000

When click "Go back" the header switches to fully blank view, then the previous screen headers fades in.

image

image

During the 2nd back navigation when using the native back button instead, you can see the flicker does not happen and the transition is smooth.

### Steps to reproduce

- In the fabric example app using [these changes](https://github.com/janicduplessis/react-native-screens/commit/2e08fc6f77c43e2cd2b397782cd2c34fe0ef4073)
- Navigate to "Simple Native Stack" example
- Click "Go to detail"
- Click "Go back"
- Notice the header flicker
- Click "Go to detail"
- Click "< Main"
- Notice the header transition is fine

https://github.com/janicduplessis/react-native-screens/tree/%40janic/header-flick

### Screens version

main@151c1c74cfcc4ae4e6cbf987b64472be8d17dc9c

### React Native version

0.73

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

_No response_

### Acknowledgements

Yes

alduzy:
## Description

This PR intents to fix flickering custom header items when going to a previous screen on `fabric` architecture.
The items are unmounted before the transition happens when `POP` action is dispatched on navigation from JS causing the items to vanish for a moment.
The adopted solution uses snapshots of the custom items to be used until the transition's done.

Fixes #2243.

## Changes

- added snapshots of the custom header items
- modified `Test556` for repro

### Before

https://github.com/user-attachments/assets/9da060ed-b65e-4b32-9ab1-debfc2bfd02d

### After

https://github.com/user-attachments/assets/0413fab0-05f6-4e55-adab-f283e01bc551

- Use `Test556` repro

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/apps/src/tests/Test556.tsx
+++ b/apps/src/tests/Test556.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Button, Text } from 'react-native';
+import { Button, StyleSheet, Text, View } from 'react-native';
 import { NavigationContainer } from '@react-navigation/native';
 import { createNativeStackNavigator } from '@react-navigation/native-stack';
 
@@ -11,8 +11,16 @@ export default function App() {
       <Stack.Navigator
         screenOptions={{
           animation: 'fade',
-          headerTitle: () => <Text>Simple Native Stack</Text>,
-          headerRight: () => <Text>Right</Text>,
+          headerTitle: () => (
+            <View style={styles.container}>
+              <Text>Simple Native Stack</Text>
+            </View>
+          ),
+          headerRight: () => (
+            <View style={styles.container}>
+              <Text>Right</Text>
+            </View>
+          ),
         }}>
         <Stack.Screen name="First" component={First} />
         <Stack.Screen name="Second" component={Second} />
@@ -38,3 +46,10 @@ function Second({ navigation }: any) {
     />
   );
 }
+
+const styles = StyleSheet.create({
+  container: {
+    padding: 3,
+    backgroundColor: 'cyan',
+  },
+});
--- a/apps/src/tests/Test556.tsx
+++ b/apps/src/tests/Test556.tsx
@@ -12,6 +12,7 @@ export default function App() {
         screenOptions={{
           animation: 'fade',
           headerTitle: () => <Text>Simple Native Stack</Text>,
+          headerRight: () => <Text>Right</Text>,
         }}>
         <Stack.Screen name="First" component={First} />
         <Stack.Screen name="Second" component={Second} />
--- a/apps/src/tests/Test556.tsx
+++ b/apps/src/tests/Test556.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Button } from 'react-native';
+import { Button, Text } from 'react-native';
 import { NavigationContainer } from '@react-navigation/native';
 import { createNativeStackNavigator } from '@react-navigation/native-stack';
 
@@ -8,7 +8,11 @@ const Stack = createNativeStackNavigator();
 export default function App() {
   return (
     <NavigationContainer>
-      <Stack.Navigator>
+      <Stack.Navigator
+        screenOptions={{
+          animation: 'fade',
+          headerTitle: () => <Text>Simple Native Stack</Text>,
+        }}>
         <Stack.Screen name="First" component={First} />
         <Stack.Screen name="Second" component={Second} />
       </Stack.Navigator>
@@ -16,7 +20,7 @@ export default function App() {
   );
 }
 
-function First({ navigation }) {
+function First({ navigation }: any) {
   return (
     <Button
       title="Tap me for second screen"
@@ -25,10 +29,10 @@ function First({ navigation }) {
   );
 }
 
-function Second({ navigation }) {
+function Second({ navigation }: any) {
   return (
     <Button
-      title="Tap me for second screen"
+      title="Tap me for first screen"
       onPress={() => navigation.popTo('First')}
     />
   );
--- a/apps/src/tests/Test556.tsx
+++ b/apps/src/tests/Test556.tsx
@@ -1,9 +1,13 @@
 import * as React from 'react';
 import { Button, StyleSheet, Text, View } from 'react-native';
-import { NavigationContainer } from '@react-navigation/native';
-import { createNativeStackNavigator } from '@react-navigation/native-stack';
+import { NavigationContainer, ParamListBase } from '@react-navigation/native';
+import { NativeStackNavigationProp, createNativeStackNavigator } from '@react-navigation/native-stack';
 
 const Stack = createNativeStackNavigator();
+
+type ScreenBaseProps = {
+  navigation: NativeStackNavigationProp<ParamListBase>;
+}
 
 export default function App() {
   return (
@@ -41,7 +45,7 @@ export default function App() {
   );
 }
 
-function First({ navigation }: any) {
+function First({ navigation }: ScreenBaseProps) {
   return (
     <Button
       title="Tap me for second screen"
@@ -50,7 +54,7 @@ function First({ navigation }: any) {
   );
 }
 
-function Second({ navigation }: any) {
+function Second({ navigation }: ScreenBaseProps) {
   return (
     <Button
       title="Tap me for first screen"
--- a/apps/src/tests/Test556.tsx
+++ b/apps/src/tests/Test556.tsx
@@ -11,19 +11,31 @@ export default function App() {
       <Stack.Navigator
         screenOptions={{
           animation: 'fade',
+        }}>
+        <Stack.Screen name="First" component={First} options={{
           headerTitle: () => (
-            <View style={styles.container}>
-              <Text>Simple Native Stack</Text>
+            <View style={[styles.container, { backgroundColor: 'goldenrod' }]}>
+              <Text>Hello there!</Text>
             </View>
           ),
           headerRight: () => (
-            <View style={styles.container}>
-              <Text>Right</Text>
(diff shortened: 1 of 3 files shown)


ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -756,9 +756,22 @@ - (void)mountChildComponentView:(UIView<RCTComponentViewProtocol> *)childCompone
 - (void)unmountChildComponentView:(UIView<RCTComponentViewProtocol> *)childComponentView index:(NSInteger)index
 {
   [_reactSubviews removeObject:(RNSScreenStackHeaderSubview *)childComponentView];
+  [self setViewToSnapshot:(RNSScreenStackHeaderSubview *)childComponentView];
```
tboba, line 759: I don't know if you should set the view to snapshot after removing childComponentView from `_reactSubviews`. Because of the side-effects, have you tried setting view to snapshot before that?

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -756,9 +756,22 @@ - (void)mountChildComponentView:(UIView<RCTComponentViewProtocol> *)childCompone
...
+- (void)setViewToSnapshot:(RNSScreenStackHeaderSubview *)childComponentView
+{
+  UINavigationItem *navitem = _screenView.controller.navigationItem;
+  if (childComponentView.type == RNSScreenStackHeaderSubviewTypeLeft) {
+    navitem.leftBarButtonItem.customView = [navitem.leftBarButtonItem.customView snapshotViewAfterScreenUpdates:NO];
+  } else if (childComponentView.type == RNSScreenStackHeaderSubviewTypeCenter) {
+    navitem.titleView = [navitem.titleView snapshotViewAfterScreenUpdates:NO];
+  } else if (childComponentView.type == RNSScreenStackHeaderSubviewTypeRight) {
+    navitem.rightBarButtonItem.customView = [navitem.rightBarButtonItem.customView snapshotViewAfterScreenUpdates:NO];
+  }
```
tboba, line 772: Do you know if search bar could also flicker, since it's also a subview of a header?

alduzy: Only custom react subviews flicker by disapearing before the screen transition's done. The searchbar does change color for a moment (like it's focused) but It's not related.

ios/RNSScreenStackHeaderConfig.mm
alduzy: It looks like setting the view to snapshot before works fine. I have not observed any side-efffects in the current solution but I'll trust you on this 😄

alduzy: https://github.com/software-mansion/react-native-screens/pull/2247/commits/a2882895f5a5bd6b3ebba3c8b50aa55d6280cc11

ios/RNSScreenStackHeaderConfig.mm
WoLewicki: Is it not too late to call it here? Shouldn't the `unmountChildComponentView` here be called after the subviews are already removed?

apps/src/tests/Test556.tsx
```diff
@@ -8,15 +8,20 @@ const Stack = createNativeStackNavigator();
...
+        }}>
         <Stack.Screen name="First" component={First} />
         <Stack.Screen name="Second" component={Second} />
       </Stack.Navigator>
     </NavigationContainer>
   );
 }
 
-function First({ navigation }) {
+function First({ navigation }: any) {
```
WoLewicki, line 24: Do you need to provide the return type at all? If so, `any` doesn't seem too good :D

alduzy: It's the type of the props, not the return type. + I don't think we need to define the types of the navigation prop in such minimal repro, that's why `any`.

ios/RNSScreenStackHeaderConfig.mm
alduzy: I haven't noticed any problems with current order.

alduzy:
I added a container to both header elements in repro to make sure the nested views are considered during snapshot.

apps/src/tests/Test556.tsx
tboba: @alduzy But is it this `any` type required here? I believe that you can skip the `any` here and leave destructurized property without any type 😄

kkafar (commented):
I have tested these changes together with @alduzy and I'm not sure why this change works and whether the fix will be stable. 

On Fabric the views are unmounted bottom-up, implying that from PoV of given component while removing a child view component A, the children of A are already unmounted. We have positively verified that this is the case. in `- (void)unmountChildComponentView:index:` method `UIView.subviews` array of `RNSScreenStackHeaderSubview` is already empty.

*However when making a snapshot inside this method the just-removed-children are still captured and visible on the resulting snapshot.*

We take the snapshot with the `- (UIView *)snapshotViewAfterScreenUpdates:(BOOL)afterUpdates;` which [documentation states](https://developer.apple.com/documentation/uikit/uiview/1622531-snapshotviewafterscreenupdates#parameters) following (I emphasised important parts): 

> **afterUpdates**
A Boolean value that specifies whether the snapshot should be taken after recent changes have been incorporated. Pass the value NO to capture the screen in its current state, **which might not include recent changes**.

> [Return Value](https://developer.apple.com/documentation/uikit/uiview/1622531-snapshotviewafterscreenupdates#return_value)
A new view object based on a snapshot of the **current view’s rendered contents**.

> [...] **Because the content is captured from the already rendered content, this method reflects the current visual appearance of the view and is not updated to reflect animations that are scheduled or in progress.** However, calling this method is faster than trying to render the contents of the current view into a bitmap image yourself.

The documentation indicates that that the snapshot is created based on lower-level internal structure (some kind of bitmap), when `afterUpdates == NO` the snapshot is created without taking into account ongoing / pending changes. Hence we can conclude, and that is my hypothesis, that the state with already unmounted child views has not been rendered yet (by render I mean OS level render to the screen) and thus we can still see them. 

I do not understand fully how rendering loop, transition start / end is realised by iOS and when we can rely on the unmounted views being still rendered and available for snapshotting. Thus I'm not sure how stable the patch (this PR changes) is. *I think we should test this in **debug and release** mode* @alduzy. 

This patch also gives a nudge, that we might be able to do something similar on the level of the screenstack, since all mounting code is being done synchronously on the UI thread  (but I'm not sure how rendering relates to that) @WoLewicki. I'll try to test this today & tomorrow alongside the `useLayoutEffect` approach on RN 0.75.

kkafar:
Okay, after research I think I have understanding why it works and can conclude that my hypothesis was correct. I'll try to describe the mechanism behind it in #2261

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -755,10 +755,33 @@ - (void)mountChildComponentView:(UIView<RCTComponentViewProtocol> *)childCompone
...
+      navitem.titleView = snapshot;
+      break;
+    case RNSScreenStackHeaderSubviewTypeRight:
+      navitem.rightBarButtonItem.customView = snapshot;
+      break;
+    case RNSScreenStackHeaderSubviewTypeSearchBar:
+    case RNSScreenStackHeaderSubviewTypeBackButton:
+      break;
+  }
+}
```
kkafar, line 783: I wanted to suggest to refactor out this code, so that it is shared with the `+ [RNSScreenStackHeaderConfig updateViewController:]` method, but it seems that we need only a subset of the functionality there. 

I would like you to add comments here, and in appropriate fragment of the `+ [RNSScreenStackHeaderConfig updateViewController:]` method that these two must be kept in sync.

kkafar (changes requested):
Few change requests, but overall solution is good. I think we'll proceed with merging this since we know understand why does it work. See my previous comments on this PR & description of #2261

ios/RNSScreenStackHeaderConfig.mm
kkafar: https://github.com/software-mansion/react-native-screens/pull/2247/commits/1f55775e0778460448b70773e6675a6e7a19bb88

apps/src/tests/Test556.tsx
```diff
@@ -25,11 +38,18 @@ function First({ navigation }) {
   );
 }
 
-function Second({ navigation }) {
+function Second({ navigation }: any) {
```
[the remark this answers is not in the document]
kkafar: https://github.com/software-mansion/react-native-screens/pull/2247/commits/779fd7c2571dfc6ffaf8aa054379c01dea991de4

apps/src/tests/Test556.tsx
kkafar: https://github.com/software-mansion/react-native-screens/pull/2247/commits/779fd7c2571dfc6ffaf8aa054379c01dea991de4

ios/RNSScreenStackHeaderConfig.mm
kkafar: The search bar animates nicely and does not disappear.

Issue #2232, opened 2024-07-05
Wrong headerTitle layout when pushing screen with keyboard on new arch


janicduplessis:
### Description

When pushing a new screen that has `headerTitle`, `headerRight` and an autoFocus TextInput the layout of the `headerTitle` is incorrect the 2nd time the screen is opened.

Workaround:

Disable view recycling in RNSScreenStackHeaderSubview

```objc
// RNSScreenStackHeaderSubview.mm

+ (BOOL)shouldBeRecycled
{
  return NO;
}
``` 

Not sure if it would be an acceptable fix, but disabling view recycling does fix the problem.

### Steps to reproduce

Using [this code](https://gist.github.com/janicduplessis/f33697c78a8aa8701eba94356e8648ca) in FabricExample app HeaderOptions.tsx

Note: If using iOS simulator make sure to toggle the software keyboard.

- Go to "Header Options" example
- Click the "Header large item" button
- Click "Go to next screen"
- See that header title layout is correct
- Click on the navbar back icon
- Click "Go to next screen"
- See that the header title layout is not incorrect and offset to the right

First open:

<img width="380" alt="image" src="https://github.com/software-mansion/react-native-screens/assets/2677334/5421fc68-e0d5-4398-bc76-160ae804bc44">

2nd open:

<img width="377" alt="image" src="https://github.com/software-mansion/react-native-screens/assets/2677334/3b8964ae-d478-4598-9117-4ab587647676">

Weirdly this doesn't repro if no headerRight is set or if the keyboard is not opened.

### Snack or a link to a repository

https://github.com/janicduplessis/react-native-screens/tree/%40janic/header-bug-2

### Screens version

main@e5a6220

### React Native version

0.74

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Acknowledgements

Yes

tboba:
Hi @janicduplessis, thanks for creating this issue!
@alduzy Could you check if PR #2248 that you've created fixes this issue? I believe that these layout things might be similar.

alduzy:
@janicduplessis I can confirm it is reproducible even without opening the keyboard. @tboba Unfortunately the PR you mentioned does not fix this issue. I'll take a look at it

PR #2248, opened 2024-07-16
fix(iOS): header left and right layout on fabric

### LINKED ISSUE #2231: Wrong layout when updating headerLeft and headerRight on new arch
### Description

When the headerLeft is removed and headerRight has updated layout at the same time in native-stack, it results in the headerRight having incorrect layout.

Workaround:

If I update them in separate ticks (using setTimeout) layout is correct.

### Steps to reproduce

Using [this code](https://gist.github.com/janicduplessis/9e3e056aa2fb2744936615ce81dc72be) in FabricExample app HeaderOptions.tsx

- Go to "Header Options" example
- Click the "Header large item" button
- See that the header right item green square is off screen
- Comment out [line 59](https://gist.github.com/janicduplessis/9e3e056aa2fb2744936615ce81dc72be#file-headeroptions-tsx-L59) where we set `headerLeft`
- Click the "Header large item" button
- See that the header right item green square has now correct layout

Updating both headerLeft and headerRight (bug):

image

image

Updating only headerRight (correct)

image

image

https://github.com/janicduplessis/react-native-screens/tree/%40janic/header-bug-1

### Screens version

main@e5a6220

### React Native version

0.74

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

_No response_

### Acknowledgements

Yes

alduzy:
## Description

This PR fixes the header layout mismatch when updating both `headerLeft` and `headerRight` options simultaneously using `navigation.setOptions`.

Fixes #2231 .

## Changes

- Forced subview to re-layout when updating layoutMetrics
- added `Test2231.tsx` repro

### Before
Screenshot 2024-07-16 at 12 05 14

### After
Screenshot 2024-07-16 at 12 04 49

- Use `Test2231.tsx` repro

## Checklist

- [x] Ensured that CI passes


ios/RNSScreenStackHeaderSubview.mm
```diff
@@ -72,6 +72,7 @@ - (void)updateLayoutMetrics:(const react::LayoutMetrics &)layoutMetrics
         self);
   } else {
     self.bounds = CGRect{CGPointZero, frame.size};
+    [self.superview layoutIfNeeded];
```
kkafar, line 75: Okay, so we now trigger layout in the native superview (this is one of system views, beyond control of react) every time we receive a new frame from react. 

The solution itself is ok & if it helps I'm for it, but do we have understanding why does it not work without this change? Especially that we do not have such code on Paper & I believe it works fine there?

kkafar (commented):
Looks promising, I just left a question on "whys". 

If you're to spend too much time investigating then I think it is not worth it & feel free to say so. 

One general note: add link to this PR in comment above the line you added, because it is definitely not obvious why the line should be there.

ios/RNSScreenStackHeaderSubview.mm
kkafar: Did you check what happens if you disable view recycling on header subviews in this PR? Im speaking with context of this issue on which we debated earlier today: https://github.com/software-mansion/react-native-screens/issues/2232

kkafar: This might be the same issue of the views being swapped due to recycling pool working in stack manner.

alduzy: Turning off recycling does not fix this issue

kkafar (approved):
Okay, after playing around with this & seeing no other solution I think we should proceed.

PR #2249, opened 2024-07-16
fix(iOS): Crash while pushing n different screens at the same time

### LINKED ISSUE #2235: Crash when navigating to 2 screens at the same time on new arch
### Description

When navigating to 2 screens at the same time on new arch it crashes with the following error:

```ts
navigation.navigate('a');
navigation.navigate('b');
```

```
"<RNSNavigationController> is pushing the same view controller instance (<RNSScreen>) more than once which is not supported and is most likely an error in the application"
```

### Steps to reproduce

Using [this code](https://gist.github.com/janicduplessis/8750f9017d8e28c795e6dd1df8469ba9) in FabricExample app SimpleNativeStack.tsx

- Go to "Simple Native Stack" example
- Click on "Go to detail"
- App crashes

https://github.com/janicduplessis/react-native-screens/tree/%40janic/double-push

### Screens version

main@e5a6220

### React Native version

0.74

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

_No response_

### Acknowledgements

Yes

tboba:
## Description

On Paper, while user tries to navigate to 2 or more screens at the same time, updates were batched (on async thread), resulting rerendering stack hierarchy on the next layout. In the same scenario, on Fabric, updates are not being batched, which causes crash:

```
"<RNSNavigationController> is pushing the same view controller instance (<RNSScreen>) more than once which is not supported and is most likely an error in the application"
```

since `maybeAddToParentAndUpdateContainer` is being called to early. This PR fixes this bug by moving the logic of calling that method on `mountingTransactionDidMount` method.

I've also checked if this change does not call `maybeAddToParentAndUpdateContainer` too frequent and I can't see more than one correct call there (every doubled call is being catched by `return` in conditions).

Fixes #2235.

## Changes

- Moved updating containers to `mountingTransactionDidMount` method.

### Before

https://github.com/user-attachments/assets/24daf575-d06c-4972-9166-89454be60126

### After

https://github.com/user-attachments/assets/d377c527-0b2c-4850-b0b5-8e93dd5f25ac

You can use `Test2235.tsx` to test this change.

## Checklist

- [x] Ensured that CI passes


ios/RNSScreenStack.mm
```diff
@@ -1178,6 +1174,25 @@ - (void)mountingTransactionWillMount:(react::MountingTransaction const &)transac
...
 }
 
+- (void)mountingTransactionDidMount:(const facebook::react::MountingTransaction &)transaction
+               withSurfaceTelemetry:(const facebook::react::SurfaceTelemetry &)surfaceTelemetry
+{
+  for (auto &mutation : transaction.getMutations()) {
+    if (mutation.parentShadowView.componentName != nil &&
+        strcmp(mutation.parentShadowView.componentName, "RNSScreenStack") == 0) {
+      if (mutation.type == react::ShadowViewMutation::Type::Update ||
+          mutation.type == react::ShadowViewMutation::Type::Remove) {
```
alduzy, line 1184: I would avoid nesting conditions if not necessary

ios/RNSScreenStack.mm
```diff
@@ -1117,9 +1117,8 @@ - (void)mountChildComponentView:(UIView<RCTComponentViewProtocol> *)childCompone
 
   [_reactSubviews insertObject:(RNSScreenView *)childComponentView atIndex:index];
   ((RNSScreenView *)childComponentView).reactSuperview = self;
-  dispatch_async(dispatch_get_main_queue(), ^{
-    [self maybeAddToParentAndUpdateContainer];
-  });
+  // Child update operation is performed in the mountingTransactionDidMount method to prevent error, while
+  // navigating to `n` different screens at the same time.
```
alduzy, line 1121: I think this comment isn't very helpful outside the scope of this PR since the deleted lines won't be visible. WDYT?

alduzy (commented):
Looks good, left 2 comments

ios/RNSScreenStack.mm
tboba: I did it intentionally, as four conditions in one if are kinda smudging in the eyes 😄

ios/RNSScreenStack.mm
tboba: I just made it for the sake of questions like "why we cannot move this update method to the mount and unmount methods, if they're doing the same?" (especially, that removing this method won't be visible in git blame), but if you think this comment is unnecessary, I can remove it.

ios/RNSScreenStack.mm
WoLewicki: For me it looks like separating such `if` means that there should be some `else` for the nested condition. Also, why do we call it on `update` and `remove` and not on `insert` and `remove` ?

tboba: Because from my observation, `update` operations are being prepared after all `insert`/`remove` operations, which means that we should have ideal state for updating containers, since we're at the moment where all mutations have been performed and we're ready to update. However, I just noticed that it's only the case, when we're navigating to different screens, but not popping from some of them - I'll also check how this method behaves :)

tboba: @WoLewicki Everything still works fine with `update` and `remove` operations, even if I'm doing some mayhem scenarios :smile:

tboba: Original suggestion fixed in https://github.com/software-mansion/react-native-screens/pull/2249/commits/fb9e60ff78c9bb7f7f851f2ff9dea52e10dbc5d1.

WoLewicki:
Now I can see that my PR with observer does it much better since it operates on `tag` props, this way we can get the direct `Screen` it is concerning.

kkafar:
I wonder whether we should pursue reporting this to RN team & trying to fix this upstream. Since there is a discrepancy between Paper & Fabric updates batching I believe this could be considered as an unintended change, similarly to [this one](https://github.com/facebook/react-native/issues/44111).

tboba:
@kkafar Do you believe that's a batching issue? Between Paper and Fabric there are different methods for inserting and removing views and I believe that mistake lies on us and that we didn't fully covered when we should update parent containers 🤔

kkafar:
That was my impression after we talked in the office ~2 weeks ago, that on Paper these updates are batched together & on Fabric these come separated

kkafar:
Note: I recently merged #2261 which removed `RCTMountingTransactionObserving` from list of implemented protocols of `RNSScreenStackView`. If we're to merge this PR, we have to restore this otherwise the delegate methods won't be called.

ios/RNSScreenStack.mm
```diff
@@ -1178,6 +1174,24 @@ - (void)mountingTransactionWillMount:(react::MountingTransaction const &)transac
...
+        strcmp(mutation.parentShadowView.componentName, "RNSScreenStack") == 0 &&
+        (mutation.type == react::ShadowViewMutation::Type::Update ||
+         mutation.type == react::ShadowViewMutation::Type::Remove)) {
+      // we need to wait until children have their layout set. At this point they don't have the layout
+      // set yet, however the layout call is already enqueued on ui thread. Enqueuing update call on the
+      // ui queue will guarantee that the update will run after layout.
+      dispatch_async(dispatch_get_main_queue(), ^{
+        [self maybeAddToParentAndUpdateContainer];
+      });
+    }
```
kkafar, line 1191:
```suggestion
    if (mutation.parentShadowView.componentName != nil &&
        strcmp(mutation.parentShadowView.componentName, "RNSScreenStack") == 0 &&
        (mutation.type == react::ShadowViewMutation::Type::Insert ||
         mutation.type == react::ShadowViewMutation::Type::Remove)) {
      // we need to wait until children have their layout set. At this point they don't have the layout
      // set yet, however the layout call is already enqueued on ui thread. Enqueuing update call on the
      // ui queue will guarantee that the update will run after layout.
      dispatch_async(dispatch_get_main_queue(), ^{
        [self maybeAddToParentAndUpdateContainer];
      });
      break;
    }
```

So the `Update` here worked just by accident, because there was only single `Update` per transaction per Screen (dunno why) & when you put `Insert` here there were two consecutive inserts one by one thus leading to the same crash as calling `updateContainer` in `mountChildComponentView:index:`. I propose using `Insert` here, as that makes much more sense & just breaking after a update is scheduled, as it doesn't seem to make sense to schedule the block multiple (possibly a big number) of times per every transaction.

ios/RNSScreenStack.mm
kkafar, line 1121: I would rather have this comment removed or limited to following

```suggestion
  // Container update is done after all mount operations are executed in `- [RNSScreenStackView mountingTransactionDidMount: withSurfaceTelemetry:]`
```

kkafar: As when reading the code it is not really helpful and rather confusing.

kkafar (changes requested):
Great job on this one 💪🏻 

Just short summary so that error mechanism is noted down & not lost for future generations 😄 

Okay, so the crash happens because we do schedule container update each time a component is mounted under stack and when two components come in consecutively, one after another in single transaction the UIKit protests - to be honest I do not get why this is the case as the updates we schedule should be executed in serial & synchronous manner on main thread, but it is the case. 

Thus mimicking logic on Paper, this patch moves container updates to a point after the transaction is completed. 

> [!caution] 
One thing that bothers me is that during testing of this PR I observed that the `Create` / `Insert` mutations are coming in separate transaction from `Update` mutations when navigating, however I can not observe why this is the case in the native code. **Most likely** this is caused by setting props / state in useEffects and thus triggering consecutive renders, but I wanted this to be noted down here.

The fix makes sense & we will proceed with this approach.

The `updateContainer` call is moved to `mountingTransactionDidMount: withSurfaceTelemetry:` and I believe there is no better place to do it, since on Fabric there is no counterpart of `didUpdateReactSubviews` (we have `finalizeUpdates`, however it does not inform on changes in subviews - this would be something nice to add to `RCTComponentViewProtocol`). Thus we're good on this front. 

The call to `updateContainer` is enqueued on main queue and not executed synchronously - and this is good, because we want to update container (present our view controllers) after they receive native frame (the one computed by native layout). I do not know what action triggers the native layout precisely, however it seems that it is already scheduled during react mounting. Thus we're good on this front also.

Left few remarks to consult.

> [!note]
Before merging this PR we need to merge main & add the `RCTMountingTransactionObserving` protocol back to the list of implemented protocols of `RNSScreenStackView`

ios/RNSScreenStack.mm
tboba: Fixed in https://github.com/software-mansion/react-native-screens/pull/2249/commits/3d89cfd66f04044c2fb3062765f7becd6dbe3f62.

ios/RNSScreenStack.mm
tboba: Fixed in https://github.com/software-mansion/react-native-screens/pull/2249/commits/3d89cfd66f04044c2fb3062765f7becd6dbe3f62.

ios/RNSScreenStack.mm
```diff
@@ -1178,6 +1174,24 @@ - (void)mountingTransactionWillMount:(react::MountingTransaction const &)transac
   }
 }
 
+- (void)mountingTransactionDidMount:(const facebook::react::MountingTransaction &)transaction
+               withSurfaceTelemetry:(const facebook::react::SurfaceTelemetry &)surfaceTelemetry
+{
+  for (auto &mutation : transaction.getMutations()) {
```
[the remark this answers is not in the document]
tboba: Fixed in https://github.com/software-mansion/react-native-screens/pull/2249/commits/3d89cfd66f04044c2fb3062765f7becd6dbe3f62.

kkafar (approved):
I think we're good to go 🎉

kkafar:
> I'm starting to think if it is safe what we do in those observers. What I mean is that we don't check if the mutations concern this exact `ScreenStack`, but we check all of them. Fortunately the actions dispatched here are rather safe, since we have checks with early returns if the updates are not meant for this particular `ScreenStack`. Nevertheless, it seems risky to do it that way in general.

Absolute agree here. We can protect ourselves temporarily by filtering mutations by the tag <- we do have access to it.

kkafar:
Please withhold merging right now, I'm preparing & testing a patch here

kkafar:
With clearance from @tboba I allowed myself to push following changes:

1. updated example to test nested stack view,
2. filter mutations by component tag, not by component name.

Both changes aim to handle what @WoLewicki has noticed in https://github.com/software-mansion/react-native-screens/pull/2249#issuecomment-2239473307

By filtering by component tag we will dispatch update block only when there was a mutation concerning this particular stack view in given transaction.

The nested stack screen was added to allow for testing this change (you can do so by adding a native log in line before block dispatch).

PR #2250, opened 2024-07-17
refact: unified prettier config across repository

alduzy:
## Description

This PR intents to unify prettier config across example apps. Inititally I wanted to copy the configuration but It's more convenient to just use the one from the root (by not overriding it).

## Changes

- removed per project `.prettierrc` files
- fixed eslint export syntax in Example
- formatted affected files

## Checklist

- [x] Ensured that CI passes


kkafar (approved):
I like the changes. 

I'm just leaving a link to prettier docs with description of config file resolution algorithm, so that it is clear why removing `.prettierrc.js` files from applications works.

One note, I personally like `trailingComma: 'all'` option, as it makes adding new parameter / call / element less awkward. Won't insist here.

alduzy:
@kkafar Here you go 😉

kkafar (approved):
lgtm

PR #2251, opened 2024-07-17
fix: changed the minsdk from 21 to 23 to support RN 0.74.3

victorsilva-i:
## Description

Fixes the minSdkVersion from 21 to 23 to support RN 0.74.3

## Changes

Updated the build.gradle file on the android level from rnsDefaultMinSdkVersion = 21 to rnsDefaultMinSdkVersion = 23

### Before
Captura de Tela 2024-07-17 às 16 13 01

### After

Captura de Tela 2024-07-17 às 16 12 37

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -5,7 +5,7 @@ buildscript {
     ext {
         rnsDefaultTargetSdkVersion = 34
         rnsDefaultCompileSdkVersion = 34
-        rnsDefaultMinSdkVersion = 21
+        rnsDefaultMinSdkVersion = 23
         rnsDefaultKotlinVersion = '1.8.0'
     }
     ext.safeExtGet = {prop, fallback ->


kkafar:
Hi @victorsilva-i, thanks for opening this PR!

Could you elaborate what problem do you solve with this PR? 

The line you changed determines the fallback value of `minSdk` in case it is not set by the root project (application).

victorsilva-i:
Hey @kkafar! Happy to be helping this amazing open-source project. 

A couple of days ago I tried using it in a brand new RN 0.74.3 project using the new arch, with turbomodules and integrating custom kotlin features on the android side.

I had some compilation warnings and a few errors because React Native stopped supporting the SDK 21 starting on the 0.74 version. So it seems it was react-native-screens that was causing this issues because of this line, I was able to fix it using patch-package, but I decided to open this PR so this lib would follow the new minimum SDK supported. Let me know if you need any additional changes to the PR. I'm happy to help.

kkafar:
Ok I see, so there are few things. 
 
First thing is how it works: the library tries to read `midSdkVersion` from the root project (it should be application) and only if it fails (most likely the `minSdkVersion` is not specified) it uses the default value you changed here. So my guess would be that you would be able to solve your build issue with specifying the `minSdkVersion` by your self explicitly in yours application `build.gradle`.

Second thing is that from the library perspective we support react-native versions down to 0.71, so bumping it here to 23 is so so, unless we check that for RN 0.71 `minSdkVersion` was already 23 -> then we're fine.

Third thing is that I recall something that `minSdkVersion` was changed to just `minSdk` or something? Idk if I recall correctly, but it is worth checking & it might be the reason the `minSdkVersion` discovery mechanism did not work in your application, because we look `minSdkVersion` just by looking for a property with that key in root project. Worth checking out.

kkafar:
Yeah, I've just checked and it seems I have good memory :D Indeed the name was changed from `minSdkVersion` to just `minSdk` at some point. Could you try out changing the name of prop in this line of code: 

https://github.com/software-mansion/react-native-screens/blob/c70f23b090247e17fcdb897043b803cd136b886d/android/build.gradle#L105

and checking whether your project starts to work (obviously without bumping the minsdk from 21 to 23)

PR #2260, opened 2024-07-22
fix(iOS): non-interactive screen while switching between bottom-tab and native-stack navigators

### LINKED ISSUE #2252: Tab non-interactive after remount on new arch
### Description

When changing from tab navigation to stack navigation then back to tab navigation again the tab will be frozen. This seems to happen only in the very specific scenario provided in the repro example.

Some things I noticed:

- If there is no `SafeAreaProvider` it works fine
- If there is no `headerShown: false` on the stack navigators it works fine

### Steps to reproduce

Using the Fabric example [with these changes](https://github.com/janicduplessis/react-native-screens/commit/cbe041b28606acf4d3ed9e29440700483babddea):

1. Click on "Test", see the alert opens
2. Click on "Toggle"
3. Click on "Toggle" again
4. Click on "Test", see the button does not work anymore

https://github.com/janicduplessis/react-native-screens/tree/%40janic/tab-freeze

### Screens version

main@8013dc87c18c0367b2fb5a2ce99b5f1e84dff0bb

### React Native version

0.73

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

_No response_

### Acknowledgements

Yes

tboba:
## Description
Currently, while switching between BottomTabNavigator and NativeStackNavigator, user cannot interact with screens from BottomTabNavigator. That's because while we're coming back to the previous screen, original screen frame has frame (0, 0, 0, 0). I've investigated that even if we're calling updateLayoutMetrics on a view, it's not being updated, since the RNSScreen implementation of that method firstly checks if `parentVC` is not RNSNavigationContainer and if it's not nil -> Because it's always nil during the insertion (as we're calling for screen updates too late), we're never updating layout metrics.

Fixes #2252.

## Changes

- Moved calling for parent updates before updating layout metrics.

### Before

https://github.com/user-attachments/assets/8da01059-e488-43e4-ab44-2286b7cd3078

### After

https://github.com/user-attachments/assets/b7336425-bb01-4c66-80e7-4e8d83ca995d

You can test `Test2252.tsx` to check these changes.

## Checklist

- [x] Ensured that CI passes


kkafar (commented):
Great catch! I've few questions that require answering before we can proceed, and possibly few objections to proposed solution. 

> [...]  I've investigated that even if we're calling updateLayoutMetrics on a view, it's not being updated, since the RNSScreen implementation of that method firstly checks if parentVC is not RNSNavigationContainer and if it's not nil -> Because it's always nil during the insertion (as we're calling for screen updates too late), we're never updating layout metrics.

I do not understand a couple of things here and need some explanation. 

Let me present you with my thought process first:

1. On Fabric, children are initialised first with props, frame, state, etc. and mounted to the parent view component before the parent is updated with its frame & prop information.
2. In `- [RNSScreenView updateLayoutMetrics: newLayoutMetrics:]`, thus when receiving Yoga frame, [we're setting the frame if and only if the `RNSScreenView`'s `RNSScreen` view controller has parent view controller and it is not a stack view controller](https://github.com/software-mansion/react-native-screens/blob/de39af6d2ab39da74e246f9ba520960316e10606/ios/RNSScreen.mm#L789). 
3. Points (1) & (2) & the fact that it is [responsibility of screen container to attach `RNSScreen` into view controller hierarchy ](https://github.com/software-mansion/react-native-screens/blob/de39af6d2ab39da74e246f9ba520960316e10606/ios/RNSScreenContainer.mm#L143-L152) imply that the initial frame for a `RNSScreenView` under a *screen container* **is always ignored**, because at [the time when the initial frame is set, the `RNSScreenView` is not yet mounted under `RNSScreenContainerView`](https://github.com/facebook/react-native/blob/700b403e06fdcbcde2a4ade9570eb572431487ea/packages/react-native/React/Fabric/Mounting/RCTMountingManager.mm#L85-L92). This is interesting fact on it's own & I think it is not intended or at least I do not know the reason to ignore initial frame in screen container. We should definitely look into this at some later point. 
4. [Just few calls later](https://github.com/facebook/react-native/blob/700b403e06fdcbcde2a4ade9570eb572431487ea/packages/react-native/React/Fabric/Mounting/RCTMountingManager.mm#L85-L92) after ignoring initial frame for `RNSScreenView` because it's view controller has not yet been mounted under screen container, we finally mount the `RNSScreenView` under `RNSScreenContainerView` & in the `mountChildComponentView:index:` method we try to set `RNSScreenView`'s frame before we effectively mount it by calling `- markUpdated`, which fails exactly due to the same reasoning as in point (3).

Do I get that correctly?

Now, provided that we do not set the screen's frame in any other place after the screen is mounted it might be likely that bugs happens cause we did not effectively set a frame for a screen. However we do it in `- [RNScreenContainerView layoutSubviews]` method <- thus here comes my next question: doesn't the system trigger the layout once the child view of container is mounted? Cause if it does not trigger the layout, then we would not set the screen's frame ever 😅 How does it happen then? 

Another questions relate to the #2252 issue description. How does the presence of `SafeAreaProvider` or `headerShown: false` on stack navigators impact the behaviour of the screen stack? Why there is no bug when `SafeAreaProvider` is not present? 

And at last I have a remark to the current solution, as mounting the child before it has a frame set could possibly lead to reintroducing some bugs. I can't point to what exactly could break, but we break [assumption we make in `- [RNSScreenContainerView attachScreen: atIndex:]` method](https://github.com/software-mansion/react-native-screens/blob/de39af6d2ab39da74e246f9ba520960316e10606/ios/RNSScreenContainer.mm#L146-L148). Thus my suggestion would be to remain with current calling order but, for example create a separate method `- [RNSScreenView updateLayoutMetrics: oldLayoutMetrics: force:]` which would set the frame forcibly if `force == YES` or apply `- [RNSScreenView updateLayoutMetrics: oldLayoutMetrics]` otherwise. You could also try to modify condition in `- [RNSScreenView updateLayoutMetrics: oldLayoutMetrics]`, so that when `parent` is `nil` we allow for setting the frame.

tboba:
Hey hey @kkafar, about your reasoning - yep, you're absolutely right! 😄 
So, about your question - we're indeed updating the layout metrics inside the `layoutSubviews` method, but what's interesting is that subviews are being layouted only on the initial render of bottom tab views (at least in this scenario, with conditional rendering) - this means that after switching what should be rendered conditionally, original component (here, bottomtabnavigator) is being recycled and after coming back again to this view, it is taken from recycling pool (I just confirmed that). So one option is to disable recycling here (I'm going to do this in a separate PR) and remove updating layout metrics from `mountChildComponentView`, but for now I'm against it, as I'm worried about content jumping while layouting subviews and not knowing initial values from `updateLayoutMetrics` in `mountChildComponentView` 🤔 

Frankly, I don't know why having `SafeAreaProvider` impacts the layout 😅 But without SafeAreaProvider or/and with `headerShown` set to true, layoutSubviews method is being called -> really hard to say what calls this layout, since it comes from `UIView` from iOS natives. Possibly overriden method of `layoutSubviews` in RNCSafeAreaProviderCompat blocks further method calls (even if it has `[super layoutSubviews]` method called there?).

Sure, I'll think about another approach at today's morning!

kkafar (approved):
I've played around with this diff on different examples & haven't seen any issue it could have caused, so I think we're good to go here. Great job!

PR #2268, opened 2024-07-30
fix: added missing icon and launch screen to fabric example

alduzy:
## Description

This PR adds missing icon and launch screen to `FabricExample` app.
After copying the icons from paper example i noticed there are icons for iPad and the example itself supports iPad platform. Thus I decided to add the iPad support to fabric example as well.

## Changes

- added iOS icons and launch screen
- added iPad support
- added android icons and launch screen

### Before
Screenshot 2024-07-30 at 11 13 17
Screenshot 2024-07-30 at 11 15 08

### After
Screenshot 2024-07-30 at 13 20 29
Screenshot 2024-07-30 at 12 54 53

## Checklist

- [x] Ensured that CI passes


tboba:
Great change! Just leaving one suggestion here.
In Reanimated, Example app and FabricExample app have different app icons. This allows the developer to visually distinguish app with old and new arch. Can we do something similar there? I would go for having an outline version of logo in FabricExample app, something like in our docs (maybe with gradient kept?) - you can find the logo [here](https://github.com/software-mansion/react-native-screens/blob/6e8380aa9a05ff90515a7346296fe17b75a86ba0/docs/static/img/logo-dark.svg).

image

For reference, I'm also leaving links for reanimated logos:
https://github.com/software-mansion/react-native-reanimated/blob/main/apps/paper-example/android/app/src/main/ic_launcher-playstore.png
https://github.com/software-mansion/react-native-reanimated/blob/main/apps/fabric-example/android/app/src/main/ic_launcher-playstore.png

kkafar:
I've also thought of that @tboba, but I don't think we have such graphic ready. Unless @alduzy can rustle one up in half an hour, otherwise I'm for merging this PR as is.

alduzy:
I've replaced the icon on fabric as @tboba suggested 😉

kkafar:
Murarz tynkarz akrobata

Nice

kkafar (approved):
A load of new kilobytes for our repository (that hurts me), but fine :D 🟢

PR #2270, opened 2024-07-30
fix(Fabric): not working animations on second-top screen

WoLewicki:
## Description

PR fixing animations on Fabric. 

The problem started in RN `0.74`, from which suspending the component triggers its `useLayoutEffect`. It has the consequences in `Animated` since detaching the native node happens then: https://github.com/facebook/react-native/blob/a5b84b925891fa509bec2e5ee5988bcb0326e198/packages/react-native/Libraries/Animated/useAnimatedProps.js#L221-L238. 

Check `Test887.tsx` to see that on fabric, when going forward or back, without this change the screen below is not animating.

### DIFF
--- a/src/components/ScreenStack.tsx
+++ b/src/components/ScreenStack.tsx
@@ -7,6 +7,10 @@ import ScreenStackNativeComponent from '../fabric/ScreenStackNativeComponent';
 const NativeScreenStack: React.ComponentType<ScreenStackProps> =
   ScreenStackNativeComponent as any;
 
+function isFabric() {
+  return 'nativeFabricUIManager' in global;
+}
+
 function ScreenStack(props: ScreenStackProps) {
   const { children, gestureDetectorBridge, ...rest } = props;
   const ref = React.useRef(null);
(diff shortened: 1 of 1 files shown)


src/components/ScreenStack.tsx
```diff
@@ -19,8 +23,14 @@ function ScreenStack(props: ScreenStackProps) {
     const isFreezeEnabled =
       descriptor?.options?.freezeOnBlur ?? freezeEnabled();
 
+    // On Fabric, when screen is frozen, animated and reanimated values are not updated
+    // due to component being unmounted. To avoid this, we don't freeze the previous screen there
+    const freezePreviousScreen = isFabric()
+      ? size - index > 2
+      : size - index > 1;
```
kkafar, line 30: We need to keep in mind that this is only a heuristics anyway and animations won't work on third-top screen etc. 

For now we can proceed.

src/components/ScreenStack.tsx
```diff
@@ -7,6 +7,10 @@ import ScreenStackNativeComponent from '../fabric/ScreenStackNativeComponent';
 const NativeScreenStack: React.ComponentType<ScreenStackProps> =
   ScreenStackNativeComponent as any;
 
+function isFabric() {
+  return 'nativeFabricUIManager' in global;
+}
```
kkafar, line 12: IIRC this should be reliably available during the very first render, so we're safe on that front (the value won't differ on first & consecutive renders).

kkafar (approved):
I don't think we have better option right now. We must do some research on why exactly this is different between Fabric & Paper and most likely try to patch this in core.

src/components/ScreenStack.tsx
WoLewicki: I can only see this as a problem when we have multiple transparent modals, in which case I wouldn't enable freeze on that navigator anyway.

PR #2274, opened 2024-07-31
fix(android): disappearing search icon on opening transparent modal

### LINKED ISSUE #2271: HeaderSearchBar Android hides on navigation
### Description

Search bar hides on navigation. I use bottom sheet as navigation screen

https://github.com/iliapnmrv/rn-navigation-SearchBar-android-bug

https://github.com/user-attachments/assets/8258fbe3-4ef2-4a44-8655-58bdf341b937

I've also created an issue in rn-navigation repository https://github.com/react-navigation/react-navigation/issues/12075

### Steps to reproduce

1. create basic screen with react-navigation search bar
2. create bottom sheet as screen
3. navigate back and forth

https://snack.expo.dev/@iliapnmrv/rnnavigationandroidsearchiconbug

### Screens version

3.32.0

### React Native version

0.74.3

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

None

### Architecture

Paper (Old Architecture)

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

alduzy:
## Description

This PR fixes disappearing search icon of the previous screen when navigating to a transparent modal with hidden header.

When navigating to a screen with `presentation: "transparentModal"` and `headerShown: false` the toolbar menu was 
unnecessarily updated with new screen's options. The update caused the search icon of the previous screen (still visible in the background) to disappear.

Fixes #2271 .

## Changes

- added `Test2271.tsx` repro
- updating the toolbar menu conditionally

### Before

https://github.com/user-attachments/assets/390f054b-b003-4b7f-b6f5-9aa9fd936dca

### After

https://github.com/user-attachments/assets/cb1fc795-c7a3-4514-9bb5-b7e1995d0b90

- use `Test2271.tsx` for repro

## Checklist

- [x] Ensured that CI passes


kkafar (commented):
Let's have some explanation what was the error mechanism & why the previous code does not work.

alduzy:
@kkafar I added a brief explanation to the description

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
```diff
@@ -154,7 +154,12 @@ class ScreenStackFragment :
     }
 
     override fun onPrepareOptionsMenu(menu: Menu) {
-        updateToolbarMenu(menu)
+        // If the screen is a transparent modal with hidden header we don't want to update the toolbar
+        // menu because it may erase the menu of the previous screen (which is still visible in these
+        // circumstances). See here: https://github.com/software-mansion/react-native-screens/issues/2271
+        if (screen.stackPresentation != Screen.StackPresentation.TRANSPARENT_MODAL || screen.headerConfig?.isHeaderHidden == false) {
+            updateToolbarMenu(menu)
+        }
```
kkafar, line 162: There is a private `isTransparent` method in `ScreenStack.Companion`. I would like you to move this method to the `Screen`, make it public and use it here.

I want you to do this, because we will have more transparent presentations introduced with #2045 and I want this logic to stay working.

kkafar (changes requested):
Okay, so the `onPrepareOptionsMenu` method is called for every screen on the stack. 

Hypothesis: it seems that Android looks for closest menu in view hierarchy to update, hence if a screen does not have a header we try to update our sibling's header.

Usually this works fine because we unmount the screen below on the stack. When transparent stack presentation is in use we try to update the sibling's toolbar with transparent screen's `null` header config, thus effectively clearing it, leading to the bug.

The logic fix is ok, however we need to tune it up a bit. See below.

Good job overall!

kkafar (approved):
I think we're good now!

PR #2276, opened 2024-07-31
fix(Android,Fabric,bridge-mode): patch crash with context detached from activity

kkafar:
## Description

When running on Fabric + "bridgefull" combination, react packages are initialised very early in application lifetime (much earlier than in Fabric + bridgeless), when there is no activity injected to `ReactContext` yet. Thus when creating packages they do not have access to the activity and we relied on this access up to this moment to setup `ScreenDummyLayoutHelper` object, throwing exception if activity was not available. 

## Changes

This diff delays initialisation of the dummy layout in case there is no access to activity when creating the object up to the point of invocation of `onHostResume` callbacks of `ReactContext` (this is the very moment activity is injected into the context), avoiding the crash. There is also fallback mechanism added in `computeDummyLayout` method, so that when accessed from C++ layer it has another chance of initialising the dummy layout, before performing computations. If it fails (see below for reasons why it might fail) `0f` is returned as header height causing content jump, but not crashing the application.

> [!important]
**Most likely** there is a race condition in this code. `onHostResume` is called from UI thread at the execution point, when JS thread & first native render & thus commit & layout is  already in progress on background thread. In case JS thread proceeds to layout & `computeDummyLayout` is called from our Screen component descriptor before `onHostResume` is called on UI thread & the dummy layout is initialised, we hit the case when computing header height will not be possible due to uninitialised dummy layout. 
I failed to trigger this behaviour even once for ~30 min of testing with trying to put UI thread to sleep etc., however I've also failed to find a proof that it won't happen because of some synchronisation / execution order.

What's also important is that there is no good way to synchronise these threads, because it is not the matter of exclusive access to some critical section, but rather a order of access between UI & background thread. Some barrier mechanism would be required here, however we do not have thread references accessible from our code. 

> [!note] 
One possible solution would be to synchronise access to `maybeInitDummyLayoutWithHeader` method & in case the background thread hits it before UI, we could wait for UI in some kind of loop ("slow spinlock"). This could guarantee correctness, however it is crazy bad, because we would impede whole render process for possibly long time. Despite the flaws of this approach this might be something to consider for the future.

> [!caution]
One more thing to note is that I rely on [JVM atomicity guarantees](https://docs.oracle.com/javase/tutorial/essential/concurrency/atomic.html) when reading / writing `isDummyLayoutInitialized` variable. There is danger that both threads hit the layout initialisation code at the same time and thus leading to corruption. TODO: try to handle this case more gracefully. A lock for initialisation code should be enough. 

Run `FabricExample` with `load(bridgelessEnabled = false)` set in `MainApplication` and see that it now works. 

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/FabricExample/android/app/src/main/java/com/fabricexample/MainApplication.kt
+++ b/FabricExample/android/app/src/main/java/com/fabricexample/MainApplication.kt
@@ -37,7 +37,7 @@ class MainApplication : Application(), ReactApplication {
     SoLoader.init(this, false)
     if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
       // If you opted-in for the New Architecture, we load the native entry point for this app.
-      load()
+      load(bridgelessEnabled = false)
     }
   }
 }
--- a/FabricExample/android/app/src/main/java/com/fabricexample/MainApplication.kt
+++ b/FabricExample/android/app/src/main/java/com/fabricexample/MainApplication.kt
@@ -37,7 +37,7 @@ class MainApplication : Application(), ReactApplication {
     SoLoader.init(this, false)
     if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
       // If you opted-in for the New Architecture, we load the native entry point for this app.
-      load(bridgelessEnabled = false)
+      load()
     }
   }
 }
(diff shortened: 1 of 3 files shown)


WoLewicki (approved):
LGTM

PR #2288, opened 2024-08-07
chore: remove unused code for finding rn version

WoLewicki:
## Description

We support RN >= 0.72 so this code is no longer needed.


kkafar (approved):
You're absolutely right. Please test this on examples / fresh app to ensure we don't break the runtime (cause the artifacts of the modified code are loaded in runtime).

tboba (approved):
LGTM aswell!

PR #2289, opened 2024-08-07
fix(Android,Fabric,bridgeless): crash on RN hot reload in dev mode when redbox in presentation

kkafar:
## Description

There was a crash on Android + Fabric + bridgeless:

1. trigger a redbox by triggering some JS error (e.g. call a method that does not exist),
2. fast refresh react-native,
3. see the crash in `onHostResume`.

The application crashed because we had a check for not null activity in reactContext. 
However after such reload ReactContext is recreated with null activity!!!.

In such case, when we don't have access to the activity, we're currently unable to create the dummy layout,
so the best we can do is not fail. The fix for jumping layout won't work, however.

This PR also adds synchornization on `synchornied` block in `maybeInitDummyLayoutWithHeaderMethod`, which should prevent
a possible data race, where both threads could try to create dummy layout simmultaneously. 

## Changes

* Do not fail in `onHostResume` when there is no activity present in the provided context,
* only log a warning in the described case,
* synchronize layout initialisation, so that there is no data race <- I'm not sure it was real due to JVM atomicity guarantees, however this time I prefer the "better safe than sorry" approach,
* mark `isLayoutInitialized` variable as `Volatile`, to ensure that updates from one thread are visible on the other thread.

Desribed above ^ in the "Description" section.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
@@ -237,7 +237,7 @@ internal class ScreenDummyLayoutHelper(
     }
 
     // This value is fetched / stored from UI and background thread. Volatile here ensures
-    // that updates are visible to the other thread. This
+    // that updates are visible to the other thread.
     @Volatile
     private var isLayoutInitialized = false
 
--- a/android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
@@ -47,7 +47,7 @@ internal class ScreenDummyLayoutHelper(
         try {
             System.loadLibrary(LIBRARY_NAME)
         } catch (e: UnsatisfiedLinkError) {
-            Log.w(TAG, "Failed to load $LIBRARY_NAME")
+            Log.w(TAG, "[RNScreens] Failed to load $LIBRARY_NAME library.")
         }
 
         weakInstance = WeakReference(this)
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
```diff
@@ -57,25 +58,47 @@ internal class ScreenDummyLayoutHelper(
...
+        // Possible data race here - activity is injected into context on UI thread.
         if (!reactContext.hasCurrentActivity()) {
             return false
         }
 
         // We need to use activity here, as react context does not have theme attributes required by
         // AppBarLayout attached leading to crash.
         val contextWithTheme =
-            requireNotNull(reactContext.currentActivity) { "[RNScreens] Attempt to use context detached from activity" }
+            requireNotNull(reactContext.currentActivity) { "[RNScreens] Attempt to use context detached from activity. This could happen only due to race-condition." }
```
WoLewicki, line 83: Why aren't you using `reactContextRef` everywhere?

kkafar: This method takes `reactContext: ReactApplicationContext` as a parameter, there is no point of using the `reactContextRef` field here. It makes more sense for me to take this as parameter, because:

1. this method requires nonnull react context to do anything,
2. moving context validation out of the method allows to throw better error messages which are tailored to specific code path & context, see e.g. `onHostResume` method;

kkafar: I do not see any advantages of using `reactContextRef` directly, because I do need to sanitise the reference anyway an almost all cases, because I do want to throw / log an message everytime `reactContextRef.get() == null` & it is unexpected.

PR #2292, opened 2024-08-08
fix(Android,Fabric): WIP Pressable / Touchable press issues

kkafar:
- **Add reproduction**
- **TOREVERT: changes in App.tsx**
- **TOREVERT: build RN from source**

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/FabricExample/android/settings.gradle
+++ b/FabricExample/android/settings.gradle
@@ -4,3 +4,13 @@ extensions.configure(com.facebook.react.ReactSettingsExtension) { ex -> ex.autol
 rootProject.name = 'FabricExample'
 include ':app'
 includeBuild('../node_modules/@react-native/gradle-plugin')
+
+includeBuild('../node_modules/react-native') {
+    dependencySubstitution {
+        substitute(module("com.facebook.react:react-android")).using(project(":packages:react-native:ReactAndroid"))
+        substitute(module("com.facebook.react:react-native")).using(project(":packages:react-native:ReactAndroid"))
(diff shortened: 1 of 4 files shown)


WoLewicki:
I'd still wait for @kkafar to come back and elaborate why he wanted to leave the `FrameOriginCorrection` set.

kkafar:
@WoLewicki I've left `FrameOriginCorrection` set after receiving initial update from JVM side, "to ensure correct values in layout metrics of shadownode" <- without this, the Screen's origin position would be (0, 0) - which would not take the header into account, however this was not used by our code anywhere.

WoLewicki:
So it was only for libs like `reanimated` to correctly read those values from `layoutMetrics`?

kkafar (commented):
Okay, this fixes the issue with pressables on screen (but not in header), thus I'm for merging it. 

The issue with pressables in header will be handled separately.

kkafar:
@WoLewicki Exactly, but right now we should prioritise making it work & defer integration with reanimated.

hosseinmd:
Please release this changes.

Kaveh-ap:
I'm facing the same issue!
Are there any plans to release these changes?

WKampel:
The onPress event does not work for me (Android), but only in a specific situation: Inside a scrollview inside a drawer's content, AND scrolled down more than a few pixels. But onPressIn does work. I'm assuming this has the same root cause.

kkafar:
@WKampel, if you mind providing a reproduction we might be quicker to handle this, thanks!

PR #2301, opened 2024-08-12
fix: do not force set `translucent` nav bar (until it's explicitly specified)

kirillzyusko:
## Description

I have next navigator structure:
- JS
   - native-stack

`native-stack` navigator customizes options as:

```ts
{
    headerShown: false,
    statusBarTranslucent: true,
    navigationBarColor: "#FFFFFF",
    navigationBarTranslucent: true,
},
```

When I go to `native-stack` - everything works well: the nav bar changes color. However, when I go back, then I'm getting a gray space in the bottom of my screen.

It happens because we disable mode `edge-to-edge` by calling `WindowCompat.setDecorFitsSystemWindows(window, true)` (the gray space appears because before we were already in edge-to-edge mode, because I had `KeyboardProvider` mounted in `App.tsx`).

So to fix this problem I decided explicitly check for boolean value for `navigationBarTranslucent` and set `decorFitsSystemWindows` only when we have an actual boolean value.

If you think that it's a problem in my project, then, please, let me know the way to fix it 😊 

## Changes

- call `WindowCompat.setDecorFitsSystemWindows(window, true)` only if `isNavigationBarTranslucent` has a boolean value;

### Before

telegram-cloud-photo-size-2-5321328488650760881-y

### After

telegram-cloud-photo-size-2-5321328488650760882-y

I tested in `react-native-keyboard-controller` example app, but if you need to test it in your code - let me know, and I'll try to prepare a reproduction code.

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
@@ -225,11 +225,7 @@ object ScreenWindowTraits {
         val window = activity.window
 
         val screenForNavBarTranslucent = findScreenForTrait(screen, WindowTraits.NAVIGATION_BAR_TRANSLUCENT)
-        val translucent = screenForNavBarTranslucent?.isNavigationBarTranslucent
-
-        if (translucent == null) {
-            return
-        }
+        val translucent = screenForNavBarTranslucent?.isNavigationBarTranslucent ?: return
 
         // Following method controls whether to display edge-to-edge content that draws behind the navigation bar
         WindowCompat.setDecorFitsSystemWindows(window, !translucent)
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
@@ -225,7 +225,11 @@ object ScreenWindowTraits {
         val window = activity.window
 
         val screenForNavBarTranslucent = findScreenForTrait(screen, WindowTraits.NAVIGATION_BAR_TRANSLUCENT)
-        val translucent = screenForNavBarTranslucent?.isNavigationBarTranslucent ?: false
+        val translucent = screenForNavBarTranslucent?.isNavigationBarTranslucent
+
+        if (translucent == null) {
+            return
+        }
 
         // Following method controls whether to display edge-to-edge content that draws behind the navigation bar
         WindowCompat.setDecorFitsSystemWindows(window, !translucent)


kirillzyusko:
@alduzy since you implemented https://github.com/software-mansion/react-native-screens/pull/2152 - may I ask you to review these changes? 👀

alduzy (changes requested):
@kirillzyusko Thanks for the PR. I could not achieve the grey stripe at the bottom of the screen while trying to reproduce it, but I believe I can understand the problem.

I have following concerns:
1. With the proposed solution edge-to-edge content mode in a native stack won't change until we encounter a screen with this option set to `false` and then it won't change until we encounter a screen with this option set to `true`. The screens in between would then have different layout depending on the navigation history, which is undesired.

2. The JS stack navigator does not enable this option and with proposed solution we are able to change it by setting it in the nested native stack.

I think we should seek another approach to fixing your problem as we want the `navigationBarTranslucent` to be `false` by default. The example below demonstrates how setting the options only in `Nested First` screen affects other screens and the root JS stack.
```
          navigationBarTranslucent: true,
          navigationBarColor: 'transparent',
 ```
https://github.com/user-attachments/assets/2cf92b9b-e46b-44e8-a42e-99357ecaf8af

**Comment by kirillzyusko** [comment]
Okay @alduzy I see

Well, the problem actually comes from the fact, that before I used `navigationBarColor`. And prior to `3.31.0` everything was fine

But starting from `3.32.0` it looks like if I use `navigationBarColor` then we automatically specify `navigationBarTranslucent=false` (and it'll move app not to edge-to-edge mode).

But what if I enable edge-to-edge mode globally in my app and I don't want to disable it when I go to nested stack navigator?.. (but just want to change `navigationBarColor` as it was prior to `3.32.0`) 🤔

**Comment by alduzy** [comment]
@kirillzyusko prior to https://github.com/software-mansion/react-native-screens/pull/2152 it was impossible to draw behind the navigation bar (for some time at least, more info in the PR's description). You probably shouldn't use `navigationBarTranslucent=true` at all and it should work just like before `3.32.0`. Please let me know if it works.

Btw.
- How is `navigationBarColor` affecting the translucency?
- How do you enable edge-to-edge mode globally?

**Comment by kirillzyusko** [comment]
> Please let me know if it works.

It doesn't work, because `navigationBarTranslucent` will be `undefined` if we don't specify it. And then the line `val translucent = screenForNavBarTranslucent?.isNavigationBarTranslucent ?: false` automatically cast it to `false` and next line disables the edge-to-edge mode (which shouldn't do that, because I don't specify `navigationBarTranslucent`, i. e. I don't want to change `edge-to-edge` mode at all).

That was the idea behind my fix - if it's `null` (undefined), then do not modify edge-to-edge layout 😅 

> How is navigationBarColor affecting the translucency?

This is what I don't understand fully yet, but:
- when I specify `navigationBarColor="#FFFFFF"`, then method `setNavigationBarTranslucent` gets called;
- if I don't specify `navigationBarColor` prop, then method `setNavigationBarTranslucent` is not getting called and everything works as it was working before 🤷‍♂️ 

Based on your words I assume the correct behavior would be to specify `navigationBarColor="#FFFFFF"` but method `setNavigationBarTranslucent` shouldn't be called?

~~I don't understand all transitive dependencies yet (in terms of functions call) in `react-native-screens` codebase, but if you can explain me why and who calls `setNavigationBarTranslucent` when we specify `navigationBarColor` I would highly appreciate that!~~

Ignore that, I found it:

```kt
        if (didSetNavigationBarAppearance) {
            setNavigationBarColor(screen, activity)
            setNavigationBarTranslucent(screen, activity)
            setNavigationBarHidden(screen, activity)
        }
```

So it looks like if we set any of three properties: `color`, `translucent`, `hidden` -> we'll automatically force set `setNavigationBarTranslucent` and even if it's not defined for current screen we anyway will cast the value to boolean and will set that.

Curious to know what is the correct fix in this case then 🤔 

> How do you enable edge-to-edge mode globally?

Many packages moves the app to edge-to-edge mode globally, for example `react-native-bars`, `react-native-keyboard-provider`, `react-native-unistyles`, `react-native-avoid-softinput`.

In my case I simply mount `KeyboardProvider` from `react-native-keyboard-controller` and it moves the app to `edge-to-edge` mode: https://github.com/kirillzyusko/react-native-keyboard-controller/blob/0d7ef4373a279eb66166fe5d53285f198762995f/android/src/main/java/com/reactnativekeyboardcontroller/views/EdgeToEdgeReactViewGroup.kt#L123-L128

alduzy:
@kirillzyusko Ok I see

I'm trying to think of a solution that doesn't introduce the problems I mentioned earlier.
Separating the `didSetNavigationBarAppearance` into smaller pieces, so that the translucency is not set along with the color, as you discovered, would probably solve the problem. However, when the translucency is eventually set the problem will reappear.

I've just now discovered, that the `navigationBarColor` does introduce the exact problems mentioned.
After setting it in one screen it's going to affect all the other screens that don't have this option set, because the line `val color = screenForNavBarColor?.navigationBarColor ?: window.navigationBarColor` cast it to current color.

Thus I'm starting to lose my objection to your proposal, as it's kind of consistent to have it this way 😄

alduzy (approved):
Despite my initial concerns about the proposal, it now appears consistent with the other options. I think we can proceed.

android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
```diff
@@ -225,7 +225,11 @@ object ScreenWindowTraits {
         val window = activity.window
 
         val screenForNavBarTranslucent = findScreenForTrait(screen, WindowTraits.NAVIGATION_BAR_TRANSLUCENT)
-        val translucent = screenForNavBarTranslucent?.isNavigationBarTranslucent ?: false
+        val translucent = screenForNavBarTranslucent?.isNavigationBarTranslucent
```
kirillzyusko, line 228: Actually it can be collapsed into one line:

`val translucent = screenForNavBarTranslucent?.isNavigationBarTranslucent ?: return`

Let me know if you are okay with these changes and I'll rework the code

WoLewicki: Wouldn't it return when `translucent` is `false` too then?

kirillzyusko: I think safe-operator + elvis operator will execute "else" only if predicate is null (not false)

At least Android studio offers such refacoring + according to the docs it looks like it's indeed truth: https://kotlinlang.org/docs/null-safety.html#elvis-operator

WoLewicki: Ok so if you could check if it really works that way (seems counterintuitive to me 😅 ) than I'm okay with it.

kirillzyusko: @WoLewicki I double checked and it works as expected - false/trues passed down, but `null` is filtered out 👍 

Below is the image from Android Studio that offered refactoring 😅 

image

WoLewicki (approved):
If it works as we would want it to work then I'm OK with it

kirillzyusko:
Are we waiting for a review from @tboba ? 👀

tboba (approved):
Nah, let's proceed 😄

kirillzyusko:
Does anything prevent this from being merged? 👀

tboba:
@kirillzyusko Nah, I thought you could do that and I was waiting if you could have anything more to commit here 😄 
Let me merge this PR then 👍

PR #2303, opened 2024-08-13
docs: Fix layout of screen sequence animation

tboba:
## Description

So previously, when there was a production build, screens from screen sequence in hero were wrongly moving into incorrect div. I've wrapped whole sequence into an additional div, so to disallow screens from moving there (somehow it fixes this issue).

## Changes

- Wrapped page with additional divs

### Before

CleanShot 2024-08-13 at 16 40 24@2x

### After

CleanShot 2024-08-13 at 16 39 23@2x

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/docs/src/pages/index.js
+++ b/docs/src/pages/index.js
@@ -14,7 +14,9 @@ import styles from './styles.module.css';
 export default function Home() {
   return (
     <Layout description="Native Navigation primitives for your React Native app.">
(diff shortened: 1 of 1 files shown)


patrycjakalinska (approved):
image

tboba:
One must imagine sisyphus happy :shipit: :shipit:

PR #2311, opened 2024-08-21
Fixing Android keyboard adjustResize not working with status bar tran…

### LINKED ISSUE #2308: Android keyboard adjustResize not working with status bar translucent
### Description

In android, the adjustResize soft input mode is not working as expected when status bar is set as translucent. 
react-native version: 0.74.3
react-native-screens version: 3.34.0

### Steps to reproduce

1. Create a basic native stack navigator and add a single screen
2. In screen options, set statusBarTranslucent as true.
3. Observe the issue with input box rendered in the screen.

With status bar translucent: (issue repro)

https://github.com/user-attachments/assets/34720672-c723-439a-9bda-98b473cef5ec

Without status bar translucent: (no issue)

https://github.com/user-attachments/assets/67289ce4-f3a1-4b53-9275-f5024733370c

https://github.com/1880akshay/RN_NewArch_TextInput/tree/android-keyboard

### Screens version

3.34.0

### React Native version

0.74.3

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

Android emulator

### Device model

_No response_

### Acknowledgements

Yes

nikhiltekwani09:
…slucent

## Description

In android, the adjustResize soft input mode is not working as expected when status bar is set as translucent.
react-native version: 0.74.3
react-native-screens version: 3.34.0

Fixes #2308

## Changes

- Fixed an issue with window insets where the keyboard (IME) height was not being accurately calculated.
- Replaced the old handling of WindowInsetsCompat.Type.statusBars() with a new implementation that includes the keyboard inset (WindowInsetsCompat.Type.ime()).
- This ensures correct bottom insets handling when the keyboard is open by taking the maximum of the system bars and IME insets.

Example where issue is reproducible - https://github.com/1880akshay/RN_NewArch_TextInput/tree/android-keyboard

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenWindowTraits.kt
@@ -19,6 +19,7 @@ import com.facebook.react.bridge.GuardedRunnable
 import com.facebook.react.bridge.ReactContext
 import com.facebook.react.bridge.UiThreadUtil
 import com.swmansion.rnscreens.Screen.WindowTraits
+import kotlin.math.max
 
 object ScreenWindowTraits {
     // Methods concerning statusBar management were taken from `react-native`'s status bar module:
@@ -39,6 +40,7 @@ object ScreenWindowTraits {
                 if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
                     val windowInsets =
                         defaultInsets.getInsets(WindowInsetsCompat.Type.statusBars())
+                    val imeInsets = defaultInsets.getInsets(WindowInsetsCompat.Type.ime())
 
                     return WindowInsetsCompat
                         .Builder()
@@ -48,7 +50,7 @@ object ScreenWindowTraits {
                                 windowInsets.left,
                                 0,
                                 windowInsets.right,
-                                windowInsets.bottom,
+                                max(windowInsets.bottom, imeInsets.bottom),
                             ),
                         ).build()
                 } else {


kirillzyusko:
@nikhiltekwani09 I don't think it's a great fix. Adding keyboard insets (i. e. keyboard height) as bottom padding doesn't sound right to me (but it's my personal opinion - I'm not a maintainer of RNS, but I added the code that you are modifying).

Another question - why do we add keyboard insets only to `Android R`? Prior to that version everything works fine?

I also opened a separate issue in another repository: https://github.com/Expensify/react-native-live-markdown/issues/346

The issue may be different but maybe it can give you some insights into where to search for the problem additionally!

nikhiltekwani09:
@kirillzyusko the things were working perfectly prior to android R
The commit that generated the issue: https://github.com/software-mansion/react-native-screens/commit/52f13c69af80e92fdf7c58aaa9645c4d20aee297

And I have considered the keyboard height using WindowInsetsCompat.Type.ime() as this gives a non 0 value only when the keyboard is opened and we want the height to decrease equivalent to value of keyboard height when it is opened

kirillzyusko:
Well, my opinion is that it's not a correct approach, because prior to https://github.com/software-mansion/react-native-screens/commit/52f13c69af80e92fdf7c58aaa9645c4d20aee297 we were using only `getSystemWindowInsetBottom`, right?

If we open a documentation we'll clearly see, that new approach is to use `getInsets(WindowInsetsCompat.Type.statusBars())`:

image

Now we add ime insets, which I think is not a correct solution, because `systemWindowInsetBottom` didn't include `.ime()` insets, but in this PR we start to mix different insets and apply them to a decorView 🤷‍♂️ (so these changes do not look to me as a backward compatible).

Here is the PR that introduced new API for Android 30+: https://github.com/software-mansion/react-native-screens/pull/1868

But again, it's only my opinion - I would wait for a review from maintainers to see what they say 👀

nikhiltekwani09:
Yes but that is always returning a 0 value and earlier it was returning the keyboard height
Do you suggest any other workaround to fix the issue?

kirillzyusko:
@nikhiltekwani09 well, I checked and it looks like `systemWindowInsetBottom` indeed includes `ime` insets. I'm still slightly concerned because the documentation says the opposite (that it includes StatusBar only). And also curious why it happens on Fabric only?

nikhiltekwani09:
Yes that is why I manually added ime insets. Also the issue is happening without fabric as well.

kirillzyusko:
> Also the issue is happening without fabric as well.

Honestly - I didn't test Fabric, but in my paper example project (keyboard-controller, without `KeyboardProvider` and `StatusBar`) everything works well:

image

`statusBarTranslucent` is managed from `native-stack`

RN 0.75.2 + RNS 3.34.0 🤷‍♂️

kirillzyusko:
@nikhiltekwani09 oh, yeah, I see it now. I think potentially a better option would be to combine insets together without involving math operations:

```kt
val windowInsets = defaultInsets.getInsets(WindowInsetsCompat.Type.systemBars() or WindowInsetsCompat.Type.ime())
WindowInsetsCompat
    .Builder()
    .setInsets(
        WindowInsetsCompat.Type.systemBars(),
        Insets.of(
            windowInsets.left,
            0,
            windowInsets.right,
          windowInsets.bottom,
        ),
    ).build()
```

But this is really strange, that Android claims to use `systemBars()` as a replacement for `systemWindowInset*`, but `systemWindowInset*` includes ime insets and `systemBars()` are not 🤷‍♂️ 

```java
        /**
         * @return All system bars. Includes {@link #statusBars()}, {@link #captionBar()} as well as
         * {@link #navigationBars()}, but not {@link #ime()}.
         */
        @InsetsType
        public static int systemBars() {
            return STATUS_BARS | NAVIGATION_BARS | CAPTION_BAR;
        }
```

Anyway I would wait for a review from SWM team 😅

nikhiltekwani09:
Yes I would also wait for the comments from the maintainers now

nikhiltekwani09:
Can we please merge this PR

PR #2313, opened 2024-08-22
fix(Android): jumping content on fabric

alduzy:
## Description

This PR applies modifications to a previous fix: https://github.com/software-mansion/react-native-screens/pull/2169 for fabric only, which has stopped working since RN `0.75`.

In RN `0.74` the `adopt` in `RNSScreenComponentDescriptior.h` was once called without `stateData` but with children and we could then check if the `ScreenStackHeaderConfig` is present among them and make adjustments based on it.

When working on https://github.com/software-mansion/react-native-screens/pull/2292 it became clear that the fix does not work anymore. Now the `adopt` is called either with no children and no `stateData` or with both.
The solution is to move the code to `appendChild` in `RNSScreenShadowNode.cpp` so we can perform the adjustments as soon as the children append.

## Changes

- moved code from `adopt` in `RNSScreenComponentDescriptior.h` to newly added `appendChild` override in `RNSScreenShadowNode.cpp`

### Before

https://github.com/user-attachments/assets/6b76864b-58bb-4c6e-9f5b-a01bb0c88d2a

### After

https://github.com/user-attachments/assets/98931e77-3877-4f67-8b28-f49d2e0f42ff

- Use `TestHeader` to test this change

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


tboba:
Why do we want to merge this into `@kkafar/fabric-pressables`? Is this PR some part of ongoing another one?

WoLewicki:
Is there a reason you removed the checks for if the calculated values are the same as the ones returned by the system later? It is important source of information for if we do the calculations right.

alduzy:
@tboba You're right, we can merge it independently.

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.cpp
```diff
@@ -15,6 +15,110 @@ Point RNSScreenShadowNode::getContentOriginOffset(
...
+                        << "[RNScreens] Failed to retrieve packageInstance or the package instance was null on JVM side";
+                return {};
+            }
+
+            jfloat headerHeight = env->CallFloatMethod(
+                    packageInstance, computeDummyLayoutID, fontSize, isTitleEmpty);
+
+            return {headerHeight};
+        }
+#endif // ANDROID
```
tboba, line 85: Let's move these methods below `appendChild`

tboba:
Are we sure this is correct solution? Currently, my emulator on `TestHeader` hides header behind the status bar. Is it just my case, or this correction should be reconsidered?

CleanShot 2024-08-23 at 11 48 08@2x

tboba (approved):
LGTM! Just one remark about method ordering 😄

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.cpp
alduzy: @tboba I'm afraid those methods have to be declared before they're used in cpp

alduzy:
@tboba EDIT: the correct layout is actually achieved based on the changes made in https://github.com/software-mansion/react-native-screens/pull/2292 so I had to rebase back to it.

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.cpp
tboba: @alduzy But didn't we achieve different order, but in RNSScreenComponentDescriptor.h? 😄 
It's not a big deal to have them here though, so if it's really not possible, I think we can proceed with what we have right now.

kkafar (approved):
Looks good! Thanks!

PR #2316, opened 2024-08-23
fix(iOS): right header incorrect position

### LINKED ISSUE #432: headerRight randomly position the component incorrectly

alduzy:
## Description

This PR fixes the incorrect position of the custom header items when updating more than one option at the same time. The proposed solution let us remove the previous fix for a similar problem: https://github.com/software-mansion/react-native-screens/pull/2248 which only fixed the issue on fabric.

Fixes #432, #2231.

## Changes

- forced re-layout of the navigation controller when subviews are updated
- removed previous fix
- updated `Test432` repro

- use `Test432` and `Test2231` to test this fix on both architectures.

## Checklist

- [x] Ensured that CI passes


ios/RNSScreenStackHeaderSubview.mm
```diff
@@ -72,9 +72,6 @@ - (void)updateLayoutMetrics:(const react::LayoutMetrics &)layoutMetrics
         self);
   } else {
     self.bounds = CGRect{CGPointZero, frame.size};
-    // We're forcing the parent view to layout this subview with correct frame size,
-    // see: https://github.com/software-mansion/react-native-screens/pull/2248
-    [self.superview layoutIfNeeded];
```
thomasttvo, line 77: I'm not super familiar with this code, but will this take away the ability to position the layout when the headerRight view changes internally (not through a `setOptions`)?

```ts
useEffect(() => {
	navigation.setOptions({
		headerRight: () => <HeaderRight />
	})
}, []) 


const HeaderRight = () => {
 const [x, setX] = useState(false)
 useTimeout(() => setX(true), 2000)
 return <View style={{backgroundColor:'green', height:20, width: x ? 40 : 20}} />
}
```

alduzy: @thomasttvo It'll be fine 😉. If you have any concerns, feel free to test this change within your project and let me know if it works as intended.

kkafar (changes requested):
Hey, native code changes look ok, if you claim this solves the issues. What I want you to do right now is:

1. Add some inline comment around the line with `layoutIfNeeded` you added explaining why it is there.
2. Rebase the PR on main & make sure that it still works, since there were some changes interfering with header layout logic since this PR has been opened. 


Great job overall!

alduzy:
@kkafar Rebased and added a brief comment. I can confirm the fix is still doing it's job 😉

Jonoh89:
patched this into our project and fixes our issue with the right header 👍

kkafar (approved):
Excellent @alduzy, we can proceed with merging then. Thank you!

kkafar:
@Jonoh89, thanks for confirmation!

PR #2325, opened 2024-08-29
fix(Android)!: overflowing text in native header

### LINKED ISSUE #1946: Long titles don't get truncated on native stack screens
### Description

on iOS:
- When the `title` option is used on `ScreenStackHeaderConfig`, then the long title works great. image

- However, when we use `ScreenStackHeaderCenterView`, the long title overlaps the right element. image

on Android:
- The long title overlaps the right element in both cases

### Steps to reproduce

mentioned in description

--

### Screens version

3.25.0

### React Native version

0.72.5

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

kkafar:
## Description

> [!caution]
This PR includes **BREAKING CHANGES**

Corresponding PR in `react-navigation`:

* https://github.com/react-navigation/react-navigation/pull/12125

Fixes #1946

This PR refactors the header config component to use flex-box model instead of absolute positioning in Yoga layer. This is required so that the Yoga layouts children of header config with respect to one another (not absolutely), so that the title can be properly truncated. 

In the end, the subviews are laid out by system anyway, thus we send additional information, such as padding / margins from native side to the shadow tree, so that we can adjust for these in Yoga layout.

> [!important]
This PR introduces a bug in very first few frames - before the information from HostTree is propagated to ShadowTree the header title might not be truncated properly. This is known issue and based on the same mechanism and "jumping content" due to not including header dimensions in first Yoga layout.

I've tried to test these changes carefully, however there still might be some bugs. It would be nice if we could get rid of them during beta-phase of 4.0. Below I paste my test matrix I conducted on `Test1649` on both platforms and architectures.

image

## Checklist

- [ ] Ensured that CI passes


Example/android/settings.gradle
```diff
@@ -1,6 +1,16 @@
 pluginManagement { includeBuild("../node_modules/@react-native/gradle-plugin") }
 plugins { id("com.facebook.react.settings") }
-extensions.configure(com.facebook.react.ReactSettingsExtension){ ex -> ex.autolinkLibrariesFromCommand() }
+extensions.configure(com.facebook.react.ReactSettingsExtension) { ex -> ex.autolinkLibrariesFromCommand() }
 rootProject.name = 'ScreensExample'
 include ':app'
 includeBuild('../node_modules/@react-native/gradle-plugin')
+
+
+includeBuild('../node_modules/react-native') {
```
maciekstosio, line 9: I don't think we want to commit that 👀

src/components/ScreenStackHeaderConfig.tsx
```diff
@@ -68,11 +89,19 @@ export const ScreenStackHeaderSearchBarView = (
...
+  headerSubviewCenter: {
+    flexDirection: 'row',
+    alignItems: 'center',
+    justifyContent: 'center',
+    flexShrink: 1,
+  },
+  headerConfig: {
+    flexDirection: 'row',
+    width: '100%',
+    justifyContent: 'space-between',
```
alduzy, line 105: Don't we need to center subviews horizontally?

```suggestion
    justifyContent: 'space-between',
    alignItems: 'center',
```

alduzy (commented):
Looks great! 🎉 
Just a question from me

src/components/ScreenStackHeaderConfig.tsx
kkafar: `justifyContent` centers the elements horizontally in this case.

Note that `justifyContent` centers the elements with respect to the "main axis". Due to `flexDirection: 'row'`, main axis is a horizontal axis.

PR #2328, opened 2024-08-30
feat(Android): add ios like `ios_from_left` slide animation 

chrispader:
## Description

This PR is based on https://github.com/software-mansion/react-native-screens/pull/1945 and implements the `ios` animation in the reverse direction (slide in from left to right, slide out from right to left)

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [x] Ensured that CI passes


chrispader:
We should discuss how we want to name these two "ios-like" animations. Before there was only `ios`, but now that there are two animations - one for each direction - i renamed `ios` to `ios_from_right` and created `ios_from_left`.

I feel like this is the more consistent approach, compared to the other animations that exist. This would be a breaking change for apps that currently use `ios`, and therefore would need to be communicated in a new update.

kkafar:
Hey @chrispader, thanks a lot for a contribution 🙌🏻 I think we might want to take a backwardcompatible approach. Let's keep the old `ios` stack animation as is, and add two new: `ios_from_right` & `ios_from_left`. This way we won't break the compatibility. We can mark the `ios` one as deprecated & I'll remove it from screens 4.0. What do you think?

chrispader:
sounds good! @kkafar

chrispader:
Just added back the `ios` animation option. This just links to the `ios_from_right` animation on Android.

I also added a depreaction warnings to the comments for `ios` animation.

kkafar:
Thanks @chrispader, we just waiting for me to have time to test this & I believe we can merge. Thank you!

kkafar (approved):
Thanks for contributing 🎉

PR #2330, opened 2024-09-02
fix(Android): getChildDrawingOrder when using refreshcontrol

### LINKED ISSUE #2329: RefreshControl IndexOutOfBoundsException: getChildDrawingOrder() crash
### Description

In Screen.kt we have this [check](https://github.com/software-mansion/react-native-screens/blob/main/android/src/main/java/com/swmansion/rnscreens/Screen.kt#L366)
`if (child.javaClass.simpleName.equals("CircleImageView")) {`

However, when we minify our code the class name is obfuscated and this check never returns true and we run into the same IndexOutOfBoundsException this check and corresponding code was added to avoid ([link](https://github.com/software-mansion/react-native-screens/pull/2134#issuecomment-2120520896))

CircleImageView is a private class within `androidx.swiperefreshlayout.widget` therefore we cannot do a direct class comparison. A work around is to stop minification of this class by adding the following to pro guard rules

`-keepnames class androidx.swiperefreshlayout.widget.CircleImageView`

This is obviously a hacky workaround, and we should instead change the check. CircleImageView extends ImageView therefore we can add a safe, minification safe check to resolve the original issue, and this bug.

`if (parent is SwipeRefreshLayout && child is ImageView) {`

### Steps to reproduce

1. Build example app in release mode
2. Navigate between screens
3. Observe IndexOutOfBoundsException due to the CircleImageView child of SwipeRefreshLayout

https://github.com/BenIrving/refreshcontrolrepro/tree/main

### Screens version

3.24.0

### React Native version

0.74.5

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

Fabric (New Architecture)

### Build type

Release mode

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

BenIrving:
## Description
Fixes #2329

## Changes
- Changes how we check if the child is a CircleImageView to be minification safe

Build `[this](https://github.com/BenIrving/refreshcontrolrepro/tree/main)` repro using this changeset, observe no crash when navigating between screens.

You can also check `Test640.tsx` test by going into the second screen (by button), then try to manually refresh list by swiping down (until the refresh control will show). During the refresh, try to go back to the first screen - application shouldn't crash.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -9,9 +9,11 @@ import android.view.View
 import android.view.ViewGroup
 import android.view.WindowManager
 import android.webkit.WebView
+import android.widget.ImageView
 import androidx.coordinatorlayout.widget.CoordinatorLayout
 import androidx.core.view.children
 import androidx.fragment.app.Fragment
+import androidx.swiperefreshlayout.widget.SwipeRefreshLayout
 import com.facebook.react.bridge.GuardedRunnable
 import com.facebook.react.bridge.ReactContext
 import com.facebook.react.uimanager.UIManagerHelper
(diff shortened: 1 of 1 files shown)


tboba (approved):
LGTM! Thanks 😄

renyufei8023:
It doesn't work for me
react native@0.75.3
react native screens@3.34.0

PR #2332, opened 2024-09-03
fix(Android): Request layout manually for CustomToolbar below Android API 29

tboba:
## Description

On Android API 29, while using `windowSoftInputMode` with `adjustPan` option, requestLayout is not being called while subviews are being added. That's because while ScreenStackHeaderConfig adds view to the toolbar, onMeasure is being called and even if we're calling `requestLayout` on parent, Android is returning from requesting the layout, as there's somehow ongoing layout. This is not the case for Android API 30 and higher.

The solution is to request another layout via ReactChoreographer (same as in ScreenContainer) to call our own layout callback on the next frame.

## Changes

- Request layout via ReactChoreographer on `requestLayout` call in CustomToolbar class

## Screenshots / GIFs

<details><summary>BEFORE</summary>

CleanShot 2024-09-03 at 17 59 13

</details>

<details><summary>AFTER</summary>

CleanShot 2024-09-03 at 17 51 14

</details>

Use `Test2332.tsx` test case to check whether this PR works properly.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
@@ -2,11 +2,50 @@ package com.swmansion.rnscreens
 
 import android.annotation.SuppressLint
 import android.content.Context
+import android.os.Build
 import androidx.appcompat.widget.Toolbar
+import com.facebook.react.modules.core.ChoreographerCompat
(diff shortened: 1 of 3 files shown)


kkafar (approved):
👍🏻 We came out with this solution on a call, tested it, can proceed.

PR #2335, opened 2024-09-05
fix(iOS): modal not presenting when deep in stack

maksg:
## Description

When 3 screens with a ‘modal’ presentation mode are stacked, navigating to a screen outside of the current stack is not functioning correctly.

## Changes

Removed dismissing logic for `lastModalVc`.

### Before
https://github.com/user-attachments/assets/d4991253-52f1-4f35-8698-242d98b218d1

### After
https://github.com/user-attachments/assets/571ebd45-7319-4225-ab5b-4f68027f4f09

`TestModalNavigation` was added to tests.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStack.mm
+++ b/ios/RNSScreenStack.mm
@@ -513,13 +513,6 @@ namespace react = facebook::react;
       [changeRootController dismissViewControllerAnimated:shouldAnimate completion:finish];
       return;
     }
-
-    UIViewController *lastModalVc = [self lastFromPresentedViewControllerChainStartingFrom:firstModalToBeDismissed];
-
-    if (lastModalVc != firstModalToBeDismissed) {
-      [lastModalVc dismissViewControllerAnimated:shouldAnimate completion:finish];
-      return;
-    }
   }
 
   // changeRootController does not have presentedViewController but it does not mean that no modals are in presentation;
--- a/apps/src/tests/TestModalNavigation.tsx
+++ b/apps/src/tests/TestModalNavigation.tsx
@@ -1,0 +1,116 @@
+import * as React from 'react';
+import { View, Text, Button } from 'react-native';
+import { NavigationContainer } from '@react-navigation/native';
+import { NativeStackScreenProps, createNativeStackNavigator } from '@react-navigation/native-stack';
+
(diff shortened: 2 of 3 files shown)


ios/RNSScreenStack.mm
```diff
@@ -513,13 +513,6 @@ - (void)setModalViewControllers:(NSArray<UIViewController *> *)controllers
       [changeRootController dismissViewControllerAnimated:shouldAnimate completion:finish];
       return;
     }
-
-    UIViewController *lastModalVc = [self lastFromPresentedViewControllerChainStartingFrom:firstModalToBeDismissed];
-
-    if (lastModalVc != firstModalToBeDismissed) {
-      [lastModalVc dismissViewControllerAnimated:shouldAnimate completion:finish];
-      return;
-    }
```
tboba, line 522: Have you checked if this doesn't break anything else? According to blame, this check has been here added in purpose. See: https://github.com/software-mansion/react-native-screens/commit/471127e26aa23eed4d14388a9646bf83de5e8699
Could you check if `Test1829.tsx` works properly with changes from this PR?

maksg: Yes, we had a call with @kkafar trying to figure out the proper fix prior to making this PR and we checked a few scenarios including `Test1829`.
Unless there is some hidden scenario which breaks by this, there shouldn't be any issues with the removal of those lines.

kkafar (commented):
Hey, I confirm that we tested this code and it seems that some other changes in code made the to-be-removed lines obsolete or even "misbehaving" (they introduce a bug now, instead of improving the situation). 
We've also tested common cases and navigation scenarios & it seems we're good on this front. 

Let's solve the merge conflicts and we can proceed.

kkafar:
Also for some reason iOS e2e CI is failing - I've restarted it. Let's make sure it passes.

maksg:
@kkafar iOS tests seem to be failing since last week so I'm not sure what's exactly the cause of that.

Anyway, I've resolved conflicts.

kkafar (approved):
Let's go. Thanks for these changes.

PR #2337, opened 2024-09-06
Monorepo Support: Use node resolver for react native location

morganick:
## Description

Monorepos are configured and structured on what works best for the project. Since this can vary from project to project, having a dynamic way to find dependencies will make this work for more projects. Currently, the repo only supports certain configurations of Monorepo structures. This PR adopts a similar approach as Expo to locate the react native package within the project. [Expo usage for reference]( https://github.com/expo/expo/blob/9821f2f8c8927d930a15041f22cd55e28f443274/packages/expo-modules-core/README.md?plain=1#L65)

## Changes

- Updates `build.gradle` to have node resolve the react native package directory
- Updates `CONTRIBUTING.md` to correct setup step as `yarn prepare` does not exist inside of `react-navigation` submodule.

1. Put a `println` statement in the `build.gradle` file to examine the output of `reactNativePackage.parentFile`
2. Build Example app for android
3. Build FabricExample app for android
4. Both apps build successfully.

## Checklist

- [ ] ~Included code example that can be used to test this change~
- [ ] ~Updated TS types~
- [x] Updated documentation: 
  - [x] https://github.com/software-mansion/react-native-screens/blob/main/guides/CONTRIBUTING.md
- [ ] Ensured that CI passes

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -150,13 +150,13 @@ repositories {
         // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
 
         // Use node resolver to locate react-native package
-        def reactNativePackage = file(["node", "--print", "require.resolve('react-native/package.json')"].execute(null, projectDir).text.trim())
+        def reactNativePackage = file(["node", "--print", "require.resolve('react-native/package.json')"].execute(null, rootDir).text.trim())
         if (reactNativePackage.exists()) {
             url "$reactNativePackage.parentFile/android"
         }
         // Fallback to react-native package colocated in node_modules
         else {
-            url "$projectDir/../node_modules/react-native/android"
+            url "$rootDir/../node_modules/react-native/android"
         }
     }
 
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -148,10 +148,18 @@ android {
 repositories {
     maven {
         // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
-        // Matches the RN Hello World template
-        // https://github.com/facebook/react-native/blob/1e8f3b11027fe0a7514b4fc97d0798d3c64bc895/local-cli/templates/HelloWorld/android/build.gradle#L21
-        url "$projectDir/../node_modules/react-native/android"
+
+        // Use node resolver to locate react-native package
+        def reactNativePackage = file(["node", "--print", "require.resolve('react-native/package.json')"].execute(null, projectDir).text.trim())
+        if (reactNativePackage.exists()) {
+            url "$reactNativePackage.parentFile/android"
+        }
+        // Fallback to react-native package colocated in node_modules
+        else {
+            url "$projectDir/../node_modules/react-native/android"
+        }
     }
+
     mavenCentral()
     mavenLocal()
     google()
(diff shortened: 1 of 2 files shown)


android/build.gradle
```diff
@@ -148,10 +148,18 @@ android {
 repositories {
     maven {
         // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
-        // Matches the RN Hello World template
-        // https://github.com/facebook/react-native/blob/1e8f3b11027fe0a7514b4fc97d0798d3c64bc895/local-cli/templates/HelloWorld/android/build.gradle#L21
-        url "$projectDir/../node_modules/react-native/android"
+
+        // Use node resolver to locate react-native package
+        def reactNativePackage = file(["node", "--print", "require.resolve('react-native/package.json')"].execute(null, projectDir).text.trim())
+        if (reactNativePackage.exists()) {
```
morganick, line 154: Inspect with:
```
println reactNativePackage.parentFile
```

You'll then see this output in the terminal when you run `yarn android` again.

morganick:
I'm able to run the same command locally that is failing in CI. I've also dropped back to node 18.20.4 locally to see if that made a difference. Everything continues to build normally.

It appears that for some reason `projectDir` is `null` on CI. I'd like to try running this with `rootDir` as it continues to work locally with that. `rootDir` resolves to the `android` directory in each of the `Example` and `FabricExample` projects.

kkafar (changes requested):
Hello, I've finally found time to test this, thank you for the contribution, we will surely merge this!

First, however, I need to articulate my confusion regarding differences between old and new approach. 

Honestly I do not understand, why does the current main code even work. In context of `react-native-screens`, `$projectDir` in "regular client react-native application" resolves to `App/node_modules/react-native-screens/android`, thus `"$projectDir/../node_modules/react-native/android"` resolves to `App/node_modules/react-native-screens/node_modules/react-native/android` => This look like we've been pointing to wrong react-native module over the last couple of years!? What's even more baffling is that when you inspect `App/node_modules/react-native-screens/node_modules/react-native` it surely does not contain `android` subdirectory... I'm sure there are mechanisms I'm just now aware now as of right now, but that's just crazy ;D 

Okay, so looking at the solution you propose I think we're fine on the first part, and the fallback to "new-old" approach with `rootDir` instead of `projectDir` also does make more sense than starting point old approach, at least for now. 

I've tested this on:

1. newly created standard react native app & it seems to be working,
2. our examples in the library repository - they work just fine.

Only remark that my purist brain brings up is that maybe we should first look for "standard" application structure and if it fails, just then try to handle other cases with fallback code. Basically swap the cases around, so that we do not run whole search algorithm if it is not needed.

I'm thinking this through right now & I believe such swap won't break any behaviour you expect in monorepo, hence I'll allow myself to change it. **Let me know, if I'm in wrong** and the order is important.

There is one more thing to do: revert changes to `guides/CONTRIBUTING.md` file. You are right - these guidelines are wrong, however your suggestion is also wrong, despite the fact that `yarn prepare` commands exists in our repo :D I'll fix this in separate PR. 

Since I can't push to your fork I'll close this PR, cherry-pick your changes and open a new one.

kkafar:
Surrogate PR: https://github.com/software-mansion/react-native-screens/pull/2352

PR #2338, opened 2024-09-10
fix(iOS): `extraLight` blur not working

### LINKED ISSUE #2320: [NativeStack] headerBlurEffect: "extraLight" has no effect
### Description

I want to use headerBlurEffect in NativeStack with value "extraLight" , but it seems to have no effect

https://github.com/user-attachments/assets/e8f3908b-3ecd-4682-b1c7-20701c111894

Same issue in `react-navigation` [repo](https://github.com/react-navigation/react-navigation/issues/12096)

### Steps to reproduce

1. clone the repo provided below
2. npm install && npm start
3. change the value of headerBlurEffect to "extraLight" as shown on the video 

https://github.com/stanleyoos/native-stack-reproduction

### Screens version

3.31.1

### React Native version

0.74.3

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

iPhone 15 Pro (iOS 17.0)

### Acknowledgements

Yes

maksg:
## Description

Blur style `UIBlurEffectStyleExtraLight` equals `0` so using it in `if (config.blurEffect)` makes it that the code for applying `backgroundEffect` doesn't execute.

With this fix the navigation bar can no longer be fully transparent because blur effect is always set as `extraLight` by default. So it's up for a discussion if it should stay that way or if we should make this value optional.

Fixes #2320.

Added `Test2320` to tests.

## Checklist

- [ ] Ensured that CI passes


kkafar (commented):
Good catch, but we rather don't want to go this way. 
Let's add a option unspecified, but only for internal API usage (do not include this option in public API), and when `blurEffect` has undefined / nullish value, in bottom-level `Screen` component (just in the wrapper for `NativeScreenComponent`) let's make use of this `Unspecified` / `Undefined` enum variant and pass it to native side. Then in headerConfig check for variant value. 

Does that sound right?

maksg:
@kkafar I added custom enum value `UIBlurEffectStyleUndefined` so it looks pretty clean.
Let me know if this approach is good enough or I should change it.

kkafar (approved):
Thank you @maksg! 

I've changed `undefined` for `none` and intorduced proxy type `RNSBlurEffectStyle` so that we can properly differentiate when the blur effect is not set.

kkafar:
I'll merge this once CI passes.

PR #2351, opened 2024-09-25
fix!: always set Zindex to undefined for InnerScreen

### LINKED ISSUE #2345: No view found for id 0x20 (unknown) for fragment ScreenFragment Android in React-Native 0.75
### Description

If you change repeatedly between tabs you get the following error:

https://github.com/user-attachments/assets/6437a2af-cb27-4715-8bad-6c1ed4238e1f

Note: I have to use the 3.35.0-rc.1 version, because of this issue: https://github.com/software-mansion/react-native-screens/issues/2324#issue-2493613911

Seems to work if you add the `detachInactiveScreens={false}` prop to the Tab.Navigator

Dependencies:

```
{
    "@react-navigation/bottom-tabs": "6.6.1",
    "@react-navigation/native": "6.1.18",
    "@react-navigation/native-stack": "6.11.0",
    "@react-navigation/stack": "6.4.1",
    "react": "18.3.1",
    "react-native-gesture-handler": "2.19.0",
    "react-native-safe-area-context": "4.11.0",
    "react-native-screens": "3.35.0-rc.1",
}
```

### Steps to reproduce

1. Create a TabNavigator with React-Navigation
2. Change between the tabs until get the error

nel

### Screens version

3.35.0-rc.1

### React Native version

0.75.0

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

Android emulator

### Device model

_No response_

### Acknowledgements

Yes

adrianryt:
## Description
Together with @kkafar we noticed that when switching back between BottomTabsView (A -> B -> A) , React is calling removeScreenAt and addScreen for tab we are leaving - in the same transaction! Because of that and asynchronous nature of react-native-screens, error described in #2345 exist. We also noticed that this weir behaviour is related to setting ZIndex in InnerScreen.

Fixes #2345

Test 2232 was failing due to change in headerBackTitleVisible property.

> [!Caution]
@kkafar:
Note that this change might be potentially breaking since we're effectively removing possibility of managing `Screen` components through "z indices", breaking public API. 

@kkafar:
The error mechanism is as follows:

1. `zIndex` being set causes RN diffing mechanism to include two mutations on the same view in the same transaction - effectively detaching and attaching it,
2. Thus when we navigate from tab B to A, react first detaches B from screen container and in the same transaction it attaches it again - however at the moment of reattach react expects B to be detached. This is not case due to the fact, that we don't execute updates synchronously, but rather we just schedule them in another block on UI thread;
3. React asserts the invariant from point 2., and when we violate it, its internal state gets corrupted later leading to crash. 

We detected that getting rid of setting `zIndex` on screens prevents the two consecutive operations on the same screen to appear, thus effectively solving the problem. 

Note, however, that we still won't support such cases with multiple mount/unmount mutations related to the same component in single transaction. 

## Changes

- For every InnerScreen created we set/override ZIndex style to undefined.
- Use headerBackButtonDisplayMode instead of headerBackTitleVisible.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/src/components/Screen.tsx
+++ b/src/components/Screen.tsx
@@ -95,6 +95,7 @@ export const InnerScreen = React.forwardRef<View, ScreenProps>(
         isNativeStack,
         gestureResponseDistance,
         onGestureCancel,
+        style,
         ...props
       } = rest;
 
@@ -125,6 +126,7 @@ export const InnerScreen = React.forwardRef<View, ScreenProps>(
         <DelayedFreeze freeze={freezeOnBlur && activityState === 0}>
           <AnimatedScreen
             {...props}
+            style={[style, { zIndex: undefined }]}
             activityState={activityState}
             sheetAllowedDetents={sheetAllowedDetents}
             sheetLargestUndimmedDetent={sheetLargestUndimmedDetent}
--- a/src/components/Screen.tsx
+++ b/src/components/Screen.tsx
@@ -126,6 +126,10 @@ export const InnerScreen = React.forwardRef<View, ScreenProps>(
         <DelayedFreeze freeze={freezeOnBlur && activityState === 0}>
           <AnimatedScreen
             {...props}
+            // Hierarchy of screens is handled on the native side and setting zIndex value causes this issue:
+            // https://github.com/software-mansion/react-native-screens/issues/2345
+            // With below change of zIndex, we force RN diffing mechanism to NOT include detaching and attaching mutation in one transaction.
+            // Detailed information can be found here https://github.com/software-mansion/react-native-screens/pull/2351
             style={[style, { zIndex: undefined }]}
             activityState={activityState}
             sheetAllowedDetents={sheetAllowedDetents}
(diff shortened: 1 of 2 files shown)


src/components/Screen.tsx
```diff
@@ -125,6 +126,7 @@ export const InnerScreen = React.forwardRef<View, ScreenProps>(
         <DelayedFreeze freeze={freezeOnBlur && activityState === 0}>
           <AnimatedScreen
             {...props}
+            style={[style, { zIndex: undefined }]}
```
WoLewicki, line 129: I'd add a comment that we are handling the order of screens on the native side and changing it here causes issues, with the link to the issue.

kkafar: +1

kkafar (changes requested):
Yeah, beside the comment and testing on react-navigation v6 I think we're good. Let me know once the comment is added (mark me for another review).

I've enhanced the PR description, so we can better reconstruct the error mechanism in case its needed in few months.

kkafar (approved):
Okay, I think we're good to go. Thank you very much :tada:

PR #2360, opened 2024-09-30
fix: change @deprecated tag into comment

maciekstosio:
## Description

Fixes https://github.com/orgs/software-mansion/projects/3/views/1?pane=issue&itemId=81360291.
`@deprecated` in JSDoc works only for the entire symbol https://stackoverflow.com/questions/29763919/jsdoc-param-together-with-deprecated
I didn't find a straightforward way to deprecate one option of union type, so I replaced it with a comment. 

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [ ] Ensured that CI passes


kkafar:
Hey, I think we're good now with our deprecation. I'm closing the PR. Thanks!

PR #2363, opened 2024-10-01
fix(Android): onLayout calls in bottom tabs with freezeOnBlur

### LINKED ISSUE #2349: Android freeze not working if layout changes since >= 3.30.0
### Description

Since `>= 3.30.0`, using `freezeOnBlur: true` in `@react-navigation/bottom-tabs` or using `enableFreeze` directly from `react-native-screens`, tabs that are not in focus still trigger `onLayout` changes for the following scenarios:
* rotating the device
* changing status bar visibility
* changing fullscreen mode

I would expect that a tab which is not in focus to not undergo a layout change. And this is causing some issues in our app.

Notice in the video, when rotating the device, the home page, which is not in focus, still has a layout change.

https://github.com/user-attachments/assets/076da81c-2d1c-417c-bfb7-9416f6f34f5b

### Steps to reproduce

1. Use `freezeOnBlur: true` in `Tab.Navigator` `screenOptions` from `@react-navigation/bottom-tabs` or directly call `enableFreeze` from `react-native-screens`
2. Switch to a new tab
3. Rotate the device OR have the change status bar visibility OR change fullscreen mode
4. Notice how layout changes occur for a view that is not in focus

https://github.com/adxzhang1/react-native-screens/tree/unfocus-freeze

### Screens version

3.34.0

### React Native version

0.75.3

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

alduzy:
## Description

This PR fixes `freezeOnBlur` option not working for `onLayout` events with `@react-navigation/bottom-tabs` navigator.

In bottom tabs `onLayout` is called inside frozen screens. Re-renders don't happen so it seems like freeze itself is working correctly, except for native events. In native stack nothing is called inside frozen screens as expected (note: native stack with fabric keeps one previous screen unfrozen).

Comparing both it seems like calling `updateScreenSizeFabric` or `updateScreenSizePaper` makes the difference - effectively stopping native event calls in non-active screens.

Fixes #2349

## Changes

- removed check for ScreenStack
- updated `TestFreeze.tsx` repro

## Checklist

- [x] Ensured that CI passes


kkafar (commented):
I don't remember from the top of my head why we have checked there for `container is ScreenStack`. Maybe @WoLewicki will have more context here. I'll be able to look into it todays afternoon.

kkafar (commented):
I'm reading this through & came to conclusion that we need better description of the error mechanism here. Why the frame is wrong (what part of it is wrong), why does this interfere with `freezeOnBlur` mechanism?

apps/src/tests/Test2349.tsx
```diff
@@ -0,0 +1,41 @@
...
+export default function App() {
+  return (
+    <NavigationContainer>
+      <Tab.Navigator screenOptions={{freezeOnBlur: true}}>
+        <Tab.Screen name="Home" component={HomeScreen} />
+        <Tab.Screen name="Settings" component={SettingsScreen} />
+      </Tab.Navigator>
+    </NavigationContainer>
+  );
+}
```
WoLewicki, line 41: newline

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -137,7 +137,7 @@ class Screen(
         r: Int,
         b: Int,
     ) {
-        if (container is ScreenStack && changed) {
+        if (changed) {
```
WoLewicki, line 140: It was added since the logic inside is intended only for `native-stack`. So I have no idea why removing it would change anything. Can we have more information?

apps/src/tests/Test2349.tsx
alduzy: https://github.com/software-mansion/react-native-screens/pull/2363/commits/e48378000422c86f8c18aad04e1f88ac3f623d92

android/src/main/java/com/swmansion/rnscreens/Screen.kt
alduzy: @WoLewicki I updated the description with some more information. That's all I got, I'll keep on investigating.

PR #2367, opened 2024-10-01
feat(iOS): sheetInitialDetent support

alduzy:
## Description

This PR adds `sheetInitialDetent` support to iOS and fixes detents logic on iOS < 16.

This functionality depends on changes made in this PR: https://github.com/react-navigation/react-navigation/pull/12032

## Changes

- supporting `sheetInitialDetent` on iOS
- modified `Test2002.tsx` repro
- updated README definition to match types
- fixed `sheetAllowedDetents` on iOS < 16

- Use `Test2002.tsx` repro

## Checklist

- [x] Updated documentation: 
- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -1158,6 +1158,10 @@ constexpr NSInteger SHEET_LARGEST_UNDIMMED_DETENT_NONE = -1;
     [self setSheetAllowedDetents:[RNSConvert detentFractionsArrayFromVector:newScreenProps.sheetAllowedDetents]];
   }
 
+  if (newScreenProps.sheetInitialDetent != oldScreenProps.sheetInitialDetent) {
+    [self setSheetInitialDetent:newScreenProps.sheetInitialDetent];
+  }
+
   if (newScreenProps.sheetLargestUndimmedDetent != oldScreenProps.sheetLargestUndimmedDetent) {
     [self setSheetLargestUndimmedDetent:[NSNumber numberWithInt:newScreenProps.sheetLargestUndimmedDetent]];
   }
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -902,13 +902,6 @@ constexpr NSInteger SHEET_LARGEST_UNDIMMED_DETENT_NONE = -1;
           [self setAllowedDetentsForSheet:sheet
                                        to:[self detentsFromMaxHeightFractions:_sheetAllowedDetents]
                                   animate:NO];
-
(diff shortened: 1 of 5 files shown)


kkafar (approved):
Looks solid, good job

kkafar (commented):
I've not tested the runtime though, make sure it works as expected & we're good.

ios/RNSScreen.mm
```diff
@@ -902,6 +902,13 @@ - (void)updateFormSheetPresentationStyle
           [self setAllowedDetentsForSheet:sheet
                                        to:[self detentsFromMaxHeightFractions:_sheetAllowedDetents]
                                   animate:NO];
+
+          if (_sheetInitialDetent > 0 && _sheetInitialDetent < _sheetAllowedDetents.count) {
+            UISheetPresentationControllerDetent *detent = sheet.detents[_sheetInitialDetent];
+            [self setSelectedDetentForSheet:sheet to:detent.identifier animate:YES];
+          } else if (_sheetInitialDetent != 0) {
+            RCTLogError(@"[RNScreens] sheetInitialDetent out of bounds for sheetAllowedDetents array");
+          }
```
kkafar, line 911: Maybe one thing that strikes me, what will happen on iOS 15? This code won't execute right and we still want to allow for choice there. Can we make is somehow work?

kkafar (commented):
I gave premature approval - have one question

alduzy:
@kkafar I updated the PR. Tested on **iOS 15.5** and **iOS 18** successfully

ios/RNSScreen.mm
```diff
@@ -889,8 +889,8 @@ - (void)updateFormSheetPresentationStyle
     sheet.delegate = self;
 #if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && defined(__IPHONE_16_0) && \
     __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_16_0
-    if (_sheetAllowedDetents.count > 0) {
-      if (@available(iOS 16.0, *)) {
+    if (@available(iOS 16.0, *)) {
+      if (_sheetAllowedDetents.count > 0) {
```
kkafar, line 893: Why is this swap done?

kkafar (commented):
Just a question

ios/RNSScreen.mm
alduzy: _sheetAllowedDetents.count > 0 is true in most cases. Nothing would happen on iOS < 16 as such case isn't handled inside the condition. After swapping them both versions work correctly

kkafar: okay, thanks!

kkafar (approved):
Looks solid 💪🏻

PR #2370, opened 2024-10-02
feat!: clamp `sheetInitialDetent` value

kkafar:
## Description

In case `sheetInitialDetent` value is not in range `[0, sheetAllowedDetents.length - 1]` I decided to clamp it. 

Alternatives:

* throw an error,
* clamp it and print warning in dev mode that value has been clamped.

Open for discussion.

`Test1649`

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/src/components/Screen.tsx
+++ b/src/components/Screen.tsx
@@ -93,6 +93,15 @@ function resolveSheetLargestUndimmedDetent(
   }
 }
 
+function clamp(value: number, lowerBound: number, upperBound: number): number {
+  if (value > upperBound) {
+    return upperBound;
+  } else if (value < lowerBound) {
+    return lowerBound;
+  }
+  return value;
+}
+
 export const InnerScreen = React.forwardRef<View, ScreenProps>(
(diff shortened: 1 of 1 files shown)


alduzy (changes requested):
Good idea!
Let's remove the `out of bounds` errors from [ios/RNSScreen.mm](https://github.com/software-mansion/react-native-screens/blob/606cd4bb495c9403138514ab5dc0042c16048b3f/ios/RNSScreen.mm#L947)  then.

OR change them to warnings, which will give a clue about the misuse

kkafar:
I think the checks on native side should stay as they are. This causes less coupling (I know we have a ton of assumptions on what our JS code does, but IMHO it should not legitimise us to create more coupling) and in case someone changes JS code in the future, there will be check for this invariant on native side.

kkafar:
On meeting we decided to go with different way of error handling & different shape of API.

PR #2372, opened 2024-10-02
fix: bring back headers when using "modal" presentation on Android

kkafar:
## Description

Recently, when adding formSheets I've also merged some code for future modals implementation.
When detaching the yet-unused code for modals from being used I've forgotten about one code place,
causing the header to not be created.

`modal` presentation is rather not widely used on Android, because it has no real differences from `push`,
however I'm restoring the old behavior for the sake of backward compatibility.

Kudos to @alduzy for noticing the issue.

## Changes

* Fixed the condition that prevents the header from being created on Android, right now header 
    is not created only for `formSheet` stack presentation.

`Test1649` - set 'modal' stack presentation to any screen and do not disable the header. See that it now works. 

## Checklist

- [ ] Ensured that CI passes


alduzy (approved):
🚀

PR #2378, opened 2024-10-03
fix: align units of sheet corner radius in different callsites

kkafar:
## Description

Before this PR, when updating corner size after initial render different unit (pixel) has been used opposed to
`dp` used during initial render. This PR align this behaviour so that `dp` is always used.

## Changes

* Calculate appropriate value in `Screen.onSheetDetentChange` function

`Test1649`

## Checklist

- [ ] Ensured that CI passes


kkafar:
Approved here: https://github.com/software-mansion/react-native-screens/pull/2377#event-14500463912

Could not reopen that PR, hece new one.

Issue #2341, opened 2024-09-16
Android RN 0.75.3, app crashes when returning to the previous screen


DinhHoangNgoc:
### Description

When I return to the previous screen, the app crashes. I am using version RN 0.75.3. I have tested with other RN versions, and this issue occurs from version RN 0.75 onwards.

https://github.com/user-attachments/assets/97f32a60-3a97-4b37-af4e-388d4db08553

![7a0f880c8e5c2802714d](https://github.com/user-attachments/assets/865d1fb4-55de-4080-9c88-1cc4a8813668)

### Steps to reproduce

app crashes when returning to the previous screen

### Snack or a link to a repository

https://github.com/Ngoca1k15PT/AppNew

### Screens version

3.34.0

### React Native version

0.75.3

### Platforms

Android

### JavaScript runtime

None

### Workflow

React Native (without Expo)

### Architecture

None

### Build type

Debug mode

### Device

Real device

### Device model

Xiaomi

### Acknowledgements

Yes

hosseinmd:
I have the same issue.

tamacroft:
same issue here

DorianMazur:
@tamacroft @hosseinmd @DinhHoangNgoc 
```
diff --git a/android/src/main/java/com/swmansion/rnscreens/Screen.kt b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
index 1ab8594ef35bf43821a46c97ca77292c87190850..8e9d712718f385638d0d58a1f3ef144e3de6eb55 100644
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -10,13 +10,16 @@ import android.view.View
 import android.view.ViewGroup
 import android.view.WindowManager
 import android.webkit.WebView
+import android.widget.ImageView
 import androidx.core.view.children
 import androidx.fragment.app.Fragment
+import androidx.swiperefreshlayout.widget.SwipeRefreshLayout
 import com.facebook.react.bridge.GuardedRunnable
 import com.facebook.react.bridge.ReactContext
 import com.facebook.react.uimanager.PixelUtil
 import com.facebook.react.uimanager.UIManagerHelper
 import com.facebook.react.uimanager.UIManagerModule
+import com.facebook.react.views.scroll.ReactScrollView
 import com.swmansion.rnscreens.events.HeaderHeightChangeEvent
 
[...]

PR #2383, opened 2024-10-03
fix(Android): going back on fabric with nested list

alduzy:
## Description

This PR intents to fix crash happening on Android (fabric) when navigating back from a screen with nested list.

### Before 
The crash happens whenever the nested list is visible on the screen, as shown below:

| ✅  | ❌ | ❌ | ❌  | ✅  |
|----------|----------|----------|----------|----------|
| Screenshot 2024-10-03 at 16 12 16 | Screenshot 2024-10-03 at 16 12 25 | Screenshot 2024-10-03 at 16 12 34 | Screenshot 2024-10-03 at 16 12 40 | Screenshot 2024-10-03 at 16 12 47 |

### After

https://github.com/user-attachments/assets/fb36d030-ef53-493d-a8cc-38be670ee504

Nested lists are rendered as ViewGroups, thus the check for ReactScrollView fails. This PR fixes the issue by checking wether the view is an indirect child of a ScrollView with [removeClippedSubviews](https://reactnative.dev/docs/optimizing-flatlist-configuration#removeclippedsubviews) enabled.

Fixes: #2341

## Changes

- added `isInsideScrollViewWithRemoveClippedSubviews` check to `startTransitionRecursive`
- modified `Test2282.tsx` repro

- use `Test2282.tsx` repro

## Checklist

- [x] Ensured that CI passes


WoLewicki (approved):
This logic keeps on growing and I am not happy about it at all. But until we find better way of handling it, I think we have to try and fix those edge cases.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -371,6 +372,20 @@ class Screen(
         }
     }
 
+    private fun isInsideScrollViewWithRemoveClippedSubviews(child: ViewGroup): Boolean {
+        if (child is ReactHorizontalScrollView || child is ReactScrollView) {
+            return false
+        }
+        var parentView = child.parent
+        while (parentView is ViewGroup && parentView !is ScreenStack) {
```
kkafar, line 380: It might be enough to check for `Screen`, right? `Screen` will be direct descendant of `ScreenStack` and we don't want to look further up. Am I right here?

```suggestion
        while (parentView is ViewGroup && parentView !is Screen) {
```

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -371,6 +372,20 @@ class Screen(
         }
     }
 
+    private fun isInsideScrollViewWithRemoveClippedSubviews(child: ViewGroup): Boolean {
```
kkafar, line 375: It would be nicer for this method to live in `ViewExt.kt` for example. I don't like this code in `Screen` implementation.

kkafar (changes requested):
I really don't like the idea of iterating up the whole tree for each child for each transition, but I guess that is better than crashing 🤷🏻‍♂️ 

We should add a comment above this function that it serves purposes of a workaround, refer to the `startTransitionRecursive` function and link this PR. 

Also: do we actually know what changed that this started crashing (maybe in react-native)? Because it looks like the root cause of the issue is on react-native side.

WoLewicki:
> What has changed? @WoLewicki do you have some context?

Nope, but would be nice to find out what changed and maybe iterate on it to handle it in core instead.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
alduzy: The loop actually finishes quicker with this check for `ScreenStack`. It's because the `ActionMenuView` goes through the loop as well and there's no `Screen` between it and the `ScreenStack`. I didn't want to add yet another check for that.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
alduzy: https://github.com/software-mansion/react-native-screens/pull/2383/commits/c1e59b14253c64ffc65c0d628ed5f9c17359fdad

kkafar (approved):
I think we're good. Thank you!

PR #2385, opened 2024-10-04
fix(iOS): header subviews layout on tab change

alduzy:
## Description

This PR intents to fix header subviews incorrect layout when changing tabs.

The previous solution did layout the subviews correctly in the test cases, but triggered an undesirable `layoutIfNeeded` when going back from tab to tab. In such case the navigation layout happened without updating subview's layout metrics.

Moving the logic to subview resolves the issue as the re-layout is now triggered only when subview's layout metrics are updated.

Related fixes from the past: https://github.com/software-mansion/react-native-screens/pull/2316, https://github.com/software-mansion/react-native-screens/pull/2248

## Changes

- combined `Test2231.tsx` with `Test432.tsx` to create comprehensive test case
- moved re-layout logic to subview

### Before
Screenshot 2024-10-04 at 10 10 15

### After
Screenshot 2024-10-04 at 10 09 37

- Use `Test432.tsx` repro

## Checklist

- [x] Ensured that CI passes


ios/RNSScreenStackHeaderSubview.mm
```diff
@@ -22,6 +22,15 @@ @implementation RNSScreenStackHeaderSubview
 
 #pragma mark - Common
 
+// We're forcing the navigation controller's view to re-layout
+// see: https://github.com/software-mansion/react-native-screens/pull/2385
+- (void)ensureNavigationLayout
+{
+  UIViewController *vc = self.reactSuperview.reactSuperview.reactViewController;
+  UINavigationController *navctr = vc.navigationController;
```
kkafar, line 30: Let's rewrite this, so the code is more readable.

`self.reactSuperview` should be a `RNSScreenStackHeaderConfig` right? Create a method that gets it. `HeaderConfig` does have `screenView` field, from which you can access `reactViewController` and then the navigation controller.

```suggestion
  UINavigationController *navctr = self.headerConfig.screenView.reactViewController.navigationController;
```

It would be also ok, to extract it to separate method.

ios/RNSScreenStackHeaderSubview.mm
```diff
@@ -22,6 +22,15 @@ @implementation RNSScreenStackHeaderSubview
...
+  [navctr.view layoutIfNeeded];
```
kkafar, line 31: Do we need to layout whole view of the navigation controller? Won't it be enough to call `navctr.navigationBar layoutIfNeeded]`?

ios/RNSScreenStackHeaderSubview.mm
```diff
@@ -22,6 +22,15 @@ @implementation RNSScreenStackHeaderSubview
 
 #pragma mark - Common
 
+// We're forcing the navigation controller's view to re-layout
+// see: https://github.com/software-mansion/react-native-screens/pull/2385
+- (void)ensureNavigationLayout
```
kkafar, line 27: `ensureNavigationBarLayout`? or `layoutNavigationBarIfNeeded`? - but only if you limit layouting to the navigation bar and not whole navigation controller content view. Otherwise I suggest `ensureNavigationControllerContentLayout` or `layoutNavigationControllerIfNeeded`.

kkafar (changes requested):
Great job overall, few remarks below

ios/RNSScreenStackHeaderSubview.mm
alduzy: https://github.com/software-mansion/react-native-screens/pull/2385/commits/9179b6b6aa8afe9626c453d4d99d27b84bbd3595

ios/RNSScreenStackHeaderSubview.mm
alduzy: `navctr.navigationBar` seems enough, thanks! https://github.com/software-mansion/react-native-screens/pull/2385/commits/9179b6b6aa8afe9626c453d4d99d27b84bbd3595

ios/RNSScreenStackHeaderSubview.mm
alduzy: Referencing headerConfig makes sense! Added a getter method for the header, but I don't think we'll be reusing those methods tbh. https://github.com/software-mansion/react-native-screens/pull/2385/commits/282b86d642a49dad764db493d5f651a9d0b6ddb7

ios/RNSScreenStackHeaderSubview.mm
```diff
@@ -22,6 +23,21 @@ @implementation RNSScreenStackHeaderSubview
 
 #pragma mark - Common
 
+- (RNSScreenStackHeaderConfig *)getHeaderConfig
+{
+  RNSScreenStackHeaderConfig *headerConfig = (RNSScreenStackHeaderConfig *)self.reactSuperview;
+  return headerConfig;
+}
```
kkafar, line 30:
```suggestion
- (nullable RNSScreenStackHeaderConfig *)getHeaderConfig
{
  return (RNSScreenStackHeaderConfig * _Nullable)self.reactSuperview;
}
```

kkafar: It might be even worth adding a check whether the `reactSuperview` is of appropriate type, e.g. in debug mode, using `NDEBUG` symbol and `isKindOfClass`

kkafar (changes requested):
Few more notes.

ios/RNSScreenStackHeaderSubview.mm
alduzy: https://github.com/software-mansion/react-native-screens/pull/2385/commits/0667d88d9a698627459506459b0094e281de58e0

kkafar (approved):
Looks good, great job :tada:

PR #2389, opened 2024-10-04
feat: allow preload using activityState

maciekstosio:
## Description

This PR is a rewrite of https://github.com/software-mansion/react-native-screens/pull/2278 using already existing `activityState` -> to avoid introducing redundant prop. 

Connected PR to react-navigation:  https://github.com/react-navigation/react-navigation/pull/12189

You can play with the preload feature in `TestPreload.tsx`

To test assertion of activityState progress go to `TestActivityStateProgression.tsx` - keep in mind that if the Screen has `isNativeStack` it will just run JS validation, remove the prop to test native checks. 

## Checklist

- [ ] Updated TS types
- [x] Updated documentation: 
- [ ] Ensured that CI passes


src/components/Screen.tsx
```diff
@@ -165,6 +166,7 @@ export const InnerScreen = React.forwardRef<View, ScreenProps>(
   function InnerScreen(props, ref) {
     const innerRef = React.useRef<ViewConfig | null>(null);
     React.useImperativeHandle(ref, () => innerRef.current!, []);
+    const prevActivityState = usePrevious(props.activityState);
```
kkafar, line 169: I don't like running another useEffect every render, however this should be ok for now, since we want to relax this constraint in future.

kkafar (approved):
Great job!

PR #2230, opened 2024-07-04
fix(iOS): update HeaderConfig view controller after unmounting subviews

janicduplessis:
## Description

When using a header left view and then removing it the default back button doesn't reappear on new arch. This is because we do not call `updateViewControllerIfNeeded` when removing views.

## Changes

Call `updateViewControllerIfNeeded` in `unmountChildComponentView`.

## Test code and steps to reproduce

In the FabricExample HeaderOptions example:

- Open Header item setting
- Choose left
- Choose right
- See the back button is gone.

image

image

image

When toggling state the default back button doesn't reappear.

## Checklist

- [x] Included code example that can be used to test this change
- [x] Ensured that CI passes

PR #2393, opened 2024-10-07
fix(iOS): header snapshots not working

alduzy:
## Description

The `updateViewControllerIfNeeded` call introduced by https://github.com/software-mansion/react-native-screens/pull/2230 forces the view controller to rebuild the subviews with the recent config.
When the screen is being unmounted it replaces the subviews with nil values as the `reactSubviews` are removed from the config one by one. Snapshots made earlier get discarded in the process.

This PR adds a condition that prevents updating the view controller when the screen is being changed + stops unnecessary snapshots when the screen is not changed.

## Changes

- updated `Test556.tsx` repro
- added `isGoingToBeRemoved` property to `RNSScreenView`
- making snapshots / updating the view controller conditionally

 

- Use `Test556.tsx` repro

## Checklist

- [x] Ensured that CI passes


kkafar (commented):
Hey, I need you to confirm that this PR does not introduce regression on  https://github.com/software-mansion/react-native-screens/pull/2230. 
You can test specifically on @tboba's reproduction: https://github.com/software-mansion/react-native-screens/pull/2230#issuecomment-2255733264 

If we don't have a regression I need you to explain what other change fixed the issue mentioned in #2230. Otherwise I'm a bit reluctant, cause it might be working only by accident.

alduzy:
@kkafar Ok, I see. Thanks to @tboba's reproduction I was able to test it and it does in fact introduce regression. I'm going to keep working on it

ios/RNSScreen.h
```diff
@@ -100,6 +100,7 @@ namespace react = facebook::react;
 @property (nonatomic) react::LayoutMetrics newLayoutMetrics;
 @property (weak, nonatomic) RNSScreenStackHeaderConfig *config;
 @property (nonatomic, readonly) BOOL hasHeaderConfig;
+@property (nonatomic) BOOL isGoingToBeRemoved;
```
kkafar, line 103: We don't need this in public API of Screen component

```suggestion

```

ios/RNSScreen.h
```diff
@@ -124,6 +125,7 @@ namespace react = facebook::react;
 - (void)updateBounds;
 - (void)notifyDismissedWithCount:(int)dismissCount;
 - (instancetype)initWithFrame:(CGRect)frame;
+- (void)onWillReceiveRemoveMutation;
```
kkafar, line 128:
```suggestion
/// Tell `Screen` that it will be unmounted in next transaction.
/// The component needs this so that we can later decide whether to 
/// replace it with snapshot or not. 
- (void)willBeUnmountedInUpcomingTransaction
```

I know I've suggested the current name, but when I'm thinking of it now: the Screen won't receive mutation, it will be only subject of one.

ios/RNSScreen.mm
```diff
@@ -123,6 +123,9 @@ - (void)initCommonProps
 #endif // !TARGET_OS_TV
   _sheetsScrollView = nil;
   _didSetSheetAllowedDetentsOnController = NO;
+#ifdef RCT_NEW_ARCH_ENABLED
+  _isGoingToBeRemoved = NO;
```
kkafar, line 127:
```suggestion
  _markedForUnmountInCurrentTransaction = NO;
```

kkafar (changes requested):
Few remarks. The logic looks good, I would only want you to change the naming.

ios/RNSScreen.mm
alduzy: https://github.com/software-mansion/react-native-screens/pull/2393/commits/29ab1b7c665260f0764d5e130ec4a403e667e9a5

PR #2395, opened 2024-10-09
fix: `HeaderConfig` impacts layout of `Screen`'s contents

alduzy:
## Description

This PR fixes headerConfig's incorrect layout with custom subviews after recent changes in https://github.com/software-mansion/react-native-screens/pull/2325.

> [!note]
@kkafar: 
Previously, before #2325, all children of the headerConfig component have been positioned absolutely, thus the headerConfig was always of height 0, not impacting layout of other components. After #2325, headerConfig's children are positioned using flexbox. This implies that it has no longer height of 0, thus it impacts the layout of other other elements, in particular `ScreenContentWrapper`, which is offset by the height of the highest header config subview.
>
> The initial idea to solve this was to set `height: 0; overflow: visible`, however, for some yet unknown reason the subviews become invisible with such styles set of headerConfig. Note that if you set the `height: 1` it works as expected.
>
> Due to above hindrance we decided to position the headerConfig approximately at the position of native header, by setting `top: -100%`. To prevent the headerConfig from blocking gestures we set `pointerEvents: 'box-none'`. 
>
> In the end I want to note, that it would be best if we came out with solution that excludes headerConfig from layout as it was before #2325.

## Changes

- added `Test2395.tsx` repro
- adjusted headerConfig's styles

### Before
Screenshot 2024-10-09 at 12 48 32

### After
Screenshot 2024-10-09 at 12 47 57

- use `Test2395.tsx` repro

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/apps/src/tests/Test2395.tsx
+++ b/apps/src/tests/Test2395.tsx
@@ -1,13 +1,5 @@
 import * as React from 'react';
-import {
-  FlatList,
-  Image,
-  Pressable,
-  StyleSheet,
-  Text,
-  View,
-  ViewProps,
-} from 'react-native';
+import { FlatList, Pressable, StyleSheet, Text, View } from 'react-native';
 import { NavigationContainer } from '@react-navigation/native';
 import { createNativeStackNavigator } from '@react-navigation/native-stack';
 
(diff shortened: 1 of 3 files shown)


kkafar (approved):
From initial testing looks good, we can release it in beta.

PR #2402, opened 2024-10-14
fix(iOS,Fabric): prevent memory leak by calling `invalidate` on deleted screens

kkafar:
## Description

On new architecture there is no callback that notifies components that they're deleted
and won't be used anymore. `RCTComponentViewProtocol#prepareForRecycle` was supposed to fulfill
this role, however after it became possible to disable view recycling for given class of components,
it is not always called. 

In our case, we need such callback in `Screen` to break the retain (strong reference) cycle between `ScreenView` and its `Screen` (controller),
otherwise we leak the `RNSScreenView` and `RNSScreen` instances.

## Changes

Overrode the `mountingTransactionWillMount:withSurfaceTelemetry:` in `RNSScreenStack`, where screens that are meant to receive
`Delete` mutation are retained, and later in `mountingTransactionDidMount:withSurfaceTelemetry:` we dispatch a series of `invalidate` calls & release the components.

## Before

image

## After

image

> [!note]
> Please note, that these screenshots are done with a patch presented below 👇🏻, w/o it, the memory leak is not that big.

Added `TestMemoryLeak` test screen to our example app. The easiest way to test this is to apply following patch:

<details>
<summary>See bigFatMemoryChunk</summary>

```c
diff --git a/ios/RNSScreen.mm b/ios/RNSScreen.mm
index 6d069499f..0e3a572b5 100644
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -61,6 +61,7 @@ constexpr NSInteger SHEET_LARGEST_UNDIMMED_DETENT_NONE = -1;
 @implementation RNSScreenView {
   __weak ReactScrollViewBase *_sheetsScrollView;
   BOOL _didSetSheetAllowedDetentsOnController;
+  NSMutableArray<UIView *> *bigFatMemoryChunk;
 #ifdef RCT_NEW_ARCH_ENABLED
   RCTSurfaceTouchHandler *_touchHandler;
   react::RNSScreenShadowNode::ConcreteState::Shared _state;
@@ -89,7 +90,12 @@ constexpr NSInteger SHEET_LARGEST_UNDIMMED_DETENT_NONE = -1;
     _props = defaultProps;
     _reactSubviews = [NSMutableArray new];
     _contentWrapper = nil;
+    bigFatMemoryChunk = [[NSMutableArray alloc] init];
+    for (int i = 0; i < 1024 * 5; ++i) {
+      [bigFatMemoryChunk addObject:[[UIView alloc] initWithFrame:frame]];
+    }
     [self initCommonProps];
+//    NSLog(@"[ALLOC][%ld] %p, memory chunk at %p, %@", self.tag, self, bigFatMemoryChunk, bigFatMemoryChunk[1023]);
   }
   return self;
 }
@@ -753,6 +759,7 @@ constexpr NSInteger SHEET_LARGEST_UNDIMMED_DETENT_NONE = -1;
 {
   _controller = nil;
   [_sheetsScrollView removeObserver:self forKeyPath:@"bounds" context:nil];
+  [bigFatMemoryChunk removeAllObjects];
 }
 
 #if !TARGET_OS_TV && !TARGET_OS_VISION
```
</details>

Try to remove call to `[strongSelf invalidate]` in `mountingTransactionDidMount:withSurfaceTelemetry:` and see that the screens are indeed retained indefinitely.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStack.mm
+++ b/ios/RNSScreenStack.mm
@@ -152,6 +152,13 @@ namespace react = facebook::react;
   UIPercentDrivenInteractiveTransition *_interactionController;
   __weak RNSScreenStackManager *_manager;
   BOOL _updateScheduled;
+#ifdef RCT_NEW_ARCH_ENABLED
+  /// Screens that are subject of `ShadowViewMutation::Type::Delete` mutation
+  /// in current transaction. This vector should be populated when we receive notification via
+  /// `RCTMountingObserving` protocol, that a transaction will be performed, and should
+  /// be cleaned up when we're notified that the transaction has been completed.
+  std::vector<__strong RNSScreenView *> _toBeDeletedScreens;
+#endif // RCT_NEW_ARCH_ENABLED
 }
 
 #ifdef RCT_NEW_ARCH_ENABLED
@@ -1151,6 +1158,16 @@ namespace react = facebook::react;
   // `- [RNSScreenStackView mountingTransactionDidMount: withSurfaceTelemetry:]`
 }
 
+- (nullable RNSScreenView *)childScreenForTag:(react::Tag)tag
+{
+  for (RNSScreenView *child in _reactSubviews) {
+    if (child.tag == tag) {
+      return child;
+    }
+  }
+  return nil;
+}
+
 - (void)unmountChildComponentView:(UIView<RCTComponentViewProtocol> *)childComponentView index:(NSInteger)index
 {
   RNSScreenView *screenChildComponent = (RNSScreenView *)childComponentView;
@@ -1184,6 +1201,19 @@ namespace react = facebook::react;
   [screenChildComponent removeFromSuperview];
 }
 
+- (void)mountingTransactionWillMount:(const facebook::react::MountingTransaction &)transaction
+                withSurfaceTelemetry:(const facebook::react::SurfaceTelemetry &)surfaceTelemetry
+{
+  for (const auto &mutation : transaction.getMutations()) {
+    if (mutation.type == react::ShadowViewMutation::Delete) {
+      RNSScreenView *_Nullable toBeRemovedChild = [self childScreenForTag:mutation.oldChildShadowView.tag];
+      if (toBeRemovedChild != nil) {
+        _toBeDeletedScreens.push_back(toBeRemovedChild);
+      }
+    }
+  }
+}
+
 - (void)mountingTransactionDidMount:(const facebook::react::MountingTransaction &)transaction
                withSurfaceTelemetry:(const facebook::react::SurfaceTelemetry &)surfaceTelemetry
 {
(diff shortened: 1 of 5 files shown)


ios/RNSScreenStack.mm
```diff
@@ -1200,6 +1230,21 @@ - (void)mountingTransactionDidMount:(const facebook::react::MountingTransaction
       break;
     }
   }
+
+  if (!self->_toBeDeletedScreens.empty()) {
+    __weak RNSScreenStackView *weakSelf = self;
+    // We want to run after container updates are performed (transitions etc.)
+    dispatch_async(dispatch_get_main_queue(), ^{
```
WoLewicki, line 1237: How are you sure that it will run in the correct moment? Maybe it would be better to do it in some platform methods from `UINavigationController` or in `viewDidDisappear`? Not sure though.

kkafar: `viewDidDisappear` is called e.g. for views that are no longer on top of the stack (e.g. new screen has been pushed in), so it does not serve our need. 

Only method that would make sense here is `didMoveToParentViewController:` called with `nil`, however I'm not sure of its exact timing with respect to react methods & I believe (haven't tested it though), that it has the same drawback as `viewDidDisappear`. 

Only concern here is that the transitions started in `maybeAddToParentAndUpdateContainer` will break in result of `screenView._controller` being set to `nil` during or before the transition. It seems that this could happen, if the transitions:

1. were asynchronous,
2. held only weak reference to the `RNSScreen`'s used in the transition,
3. or referenced the `RNSScreen`'s only via `screenView->_controller`.

I assumed that 3rd point is unlikely, because `_controller` is our private field, and AFAIK (just checked in UIKit docs) `UIView` does not expose any accessor to its owning view controller in UIKit API, thus system does not do that. 
Our animation code uses only reversed reference: it obtains `RNSScreenView` instance  from `RNSScreen`, hence we're safe on this front. 

2nd point *seems* to be false, haven't found any information on this, saying only empirically.

1st point: 🤷🏻‍♂️ Not really sure how transitions work & haven't managed to find any information on threading etc. 

Summary: This surely can be implemented in other ways, however did not see a better approach at this moment, while the suggested one seems to solve the issue w/o introducing regressions (at least my testing suggests so).

alduzy (approved):

ios/RNSScreenStack.mm
WoLewicki: > viewDidDisappear is called e.g. for views that are no longer on top of the stack (e.g. new screen has been pushed in), so it does not serve our need.

I think we have some heuristics already for when it is the top screen that is dismissed but yeah, if it works then we can stick with that.

PR #2404, opened 2024-10-14
fix(iOS): `activityState` regression check false-positive

kkafar:
## Description

We had an issue introduced in #2389 that added the preload. We temporarily
have an assumption in native stack, that the `activityState` can only progress when 
`Screen` component is rendered (directly) inside native stack navigator. This invariant
was enforced with appropriate checks on native side. It turns out that we were wrong in our 
way of checking whether the `Screen` belongs to native stack or not - we have forgotten about
`RNSContainerNavigationController` class that is used when rendering `ScreenContainer` with `hasTwoStates: true`
and which inherits from `RNSNavigationController`.

## Changes

~Added a `isNativeStackViewController` method to `RNSNavigationController` and overrode it in subclass.~

Added `RNSScreenView#isNativeStackScreen` method, which checks whether our direct react superview is a `RNSScreenStackView`.

I've found checking for `_controller.navigationController` not reliable, as `navigationController` can be found arbitrarily high in the view hierarchy.

Reported by Expo. If you want to reproduce this: create a JS stack navigator with bottom tabs inside (with `hasTwoStates: true`),
and then a regular JS stack inside & try to navigate to any screen. 

We have not caught that earlier, because we're not testing regular JS stack at all.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -304,13 +304,24 @@ constexpr NSInteger SHEET_LARGEST_UNDIMMED_DETENT_NONE = -1;
 {
   int activityState = [activityStateOrNil intValue];
   if (activityStateOrNil != nil && activityState != -1 && activityState != _activityState) {
-    if ([_controller.navigationController isKindOfClass:RNSNavigationController.class] &&
-        _activityState < activityState) {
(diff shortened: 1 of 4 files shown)


maciekstosio (approved):
LGTM 🎉

maciekstosio (approved):

PR #2407, opened 2024-10-15
chore: add support for 0.76-rc.6

maciekstosio:
## Description

This PR bumps example apps to 0.76.0-rc.6.
Removed outdated patches for react-native-reanimated and react-native-gesture-handler.
Add patch for react-native-safe-area-context.
Reanimated is provided as `.tgz` it should be released with the necessary fixes any day now. 
Tested also on 0.75.4 both paper and fabric seems to work fine.

To run it go to `Example` / `FabricExample`:

```
#  rm -rf node_modules/ android/build android/app/build # might not be necessary 
yarn
npx patch-package

```


kkafar (commented):
Let's bump to 0.76.0-rc.6, it has been released few hours ago.

.github/workflows/ios-build-test-fabric.yml
```diff
@@ -54,7 +54,7 @@ jobs:
         id: install_pods
         continue-on-error: true
         working-directory: ${{ env.WORKING_DIRECTORY }}/ios
-        run: RCT_NEW_ARCH_ENABLED=1 pod install
+        run: pod install
```
kkafar, line 57: IMO we should keep this explicit, especially that we've had this already in place.

The same for `RCT_NEW_ARCH_ENABLED=0`

Example/ios/Podfile
```diff
@@ -7,6 +7,7 @@ require Pod::Executable.execute_command('node', ['-p',
 
 platform :ios, min_ios_version_supported
 prepare_react_native_project!
+ENV['RCT_NEW_ARCH_ENABLED'] = '0'
```
kkafar, line 10: Add comment that since 0.76.0 new arch is disabled by default and we need to explicitly turn it off here.

Example/package.json
```diff
@@ -39,13 +39,16 @@
...
-    "@babel/runtime": "^7.20.0",
-    "@react-native/babel-preset": "0.75.4",
-    "@react-native/eslint-config": "0.75.4",
-    "@react-native/metro-config": "0.75.4",
-    "@react-native/typescript-config": "0.75.4",
+    "@babel/core": "^7.25.2",
+    "@babel/preset-env": "^7.25.3",
+    "@babel/runtime": "^7.25.0",
+    "@react-native-community/cli": "15.0.0-alpha.2",
+    "@react-native-community/cli-platform-android": "15.0.0-alpha.2",
```
kkafar, line 46: 🤔 Why do we need dependency for `rncli` here? 

I see it in diff, but that's interesting. 

Don't waste time on research here, I'm just posting open question if someone knows already.

FabricExample/ios/Podfile
```diff
@@ -7,7 +7,6 @@ require Pod::Executable.execute_command('node', ['-p',
 
 platform :ios, min_ios_version_supported
 prepare_react_native_project!
-ENV['RCT_NEW_ARCH_ENABLED'] = '1'
```
kkafar, line 10: I would leave it explicit - but won't insist here.

android/src/main/jni/CMakeLists.txt
```diff
@@ -38,25 +38,34 @@ target_include_directories(
...
+    react_nativemodule_core
+    react_render_core
+    react_render_debug
+    react_render_graphics
+    react_render_mapbuffer
+    react_render_componentregistry
+    react_utils
+    rrc_view
+    turbomodulejsijni
+    yoga
```
kkafar, line 66: Yoga now comes with `libreactnative.so`? Interesting.

kkafar (commented):
The changes seems good! I have some suggestions / nitpicks.

.github/workflows/ios-build-test-fabric.yml
maciekstosio: From what I checked Podfile value overrides command anyway

Example/package.json
maciekstosio: Just fyi, this is error I was having without it https://github.com/software-mansion/react-native-screens/actions/runs/11347709614/job/31559764615

FabricExample/ios/Podfile
maciekstosio: So I would go with explicte in podfile and skip the command one (I think it's overrided anyway)

alduzy (approved):

FabricExample/ios/Podfile
maciekstosio: Changed in: https://github.com/software-mansion/react-native-screens/pull/2407/commits/5d6718bcd725abf90c185d5a83f114fc6db98725

Example/ios/Podfile
maciekstosio: Changed in: https://github.com/software-mansion/react-native-screens/pull/2407/commits/5d6718bcd725abf90c185d5a83f114fc6db98725

maciekstosio:
Changed in: https://github.com/software-mansion/react-native-screens/pull/2407/commits/7765e6e76f55db63d0624f8ecab9857794dcbe9b

Example/package.json
kkafar: I see! Thanks for finding this out.

kkafar (approved):
I think we're looking good here! Thanks!

Example/package.json
alduzy: https://github.com/software-mansion/react-native-screens/pull/2432 possibly related

PR #2410, opened 2024-10-16
fix: prevent Require cycles warning

### LINKED ISSUE #2408: Require cycles break Metro import support
### Description

The require cycles in files like https://github.com/software-mansion/react-native-screens/blob/3ca34864e9d4e67bedef94079516b9d1e5a5b97a/src/components/ScreenContainer.tsx#L5C39-L5C59 where a module imported by `react-native-screens` imports `react-native-screens` leads to a require cycle. 

This is generally not great, but it leads to a break in Metro when using built-in import support:
```js
config.transformer.getTransformOptions = () => ({
  transform: {
    experimentalImportSupport: true
    inlineRequires: false,
  },
});
```

This should be fixed upstream in Metro (cc @robhogan) but figure it'd be good to also resolve it in `react-native-screens`.

### Steps to reproduce

1. Expo Router project on iOS, or Android
2. Configure Metro to use experimentalImportSupport (`npx expo customize metro.config.js`):
```js
config.transformer.getTransformOptions = () => ({
  transform: {
    experimentalImportSupport: true
    inlineRequires: false,
  },
});
```
3. Open project, it should crash with something like `Warning: TypeError: freezeEnabled is not a function (it is undefined)`

https://snack.expo.io

### Screens version

react-native-screens@4.0.0-beta.9

### React Native version

react-native@0.76.0-rc.4

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

alduzy:
## Description

This PR fixes `require cycles` warning caused by importing from `react-native-screens` in a `react-native-screens` module by changing the imports to relative.

Fixes #2408  .

## Changes

- changed imports to relative

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/gesture-handler/GestureDetectorProvider.tsx
+++ b/src/gesture-handler/GestureDetectorProvider.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { GHContext } from 'react-native-screens';
+import { GHContext } from '../native-stack/contexts/GHContext';
 import ScreenGestureDetector from './ScreenGestureDetector';
 import type { GestureProviderProps } from '../native-stack/types';
 
--- a/src/components/Screen.tsx
+++ b/src/components/Screen.tsx
@@ -5,7 +5,7 @@ import { Animated, View, Platform } from 'react-native';
 
 import TransitionProgressContext from '../TransitionProgressContext';
 import DelayedFreeze from './helpers/DelayedFreeze';
-import { ScreenProps } from 'react-native-screens';
+import { ScreenProps } from '../types';
 
 import {
   freezeEnabled,
--- a/src/components/Screen.web.tsx
+++ b/src/components/Screen.web.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { ScreenProps } from 'react-native-screens';
+import { ScreenProps } from '../types';
 import { Animated, View } from 'react-native';
 import React from 'react';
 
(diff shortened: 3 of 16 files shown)


maciekstosio (approved):

src/gesture-handler/GestureDetectorProvider.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { GHContext } from 'react-native-screens';
+import { GHContext } from '../native-stack/contexts/GHContext';
```
WoLewicki, line 2: I would be afraid of doing it in the separate package, same for `native-stack`. But if you checked it outside the repo and it works, then I'm fine with that 🎉

kkafar: What are the possible dangers here?

kkafar (approved):
I think we should be good here. 

One thing we might wanna do, as mentioned by @WoLewicki is to test the lib in external fresh app. After we've done that I think we're good to go. Thank you!

src/gesture-handler/GestureDetectorProvider.tsx
WoLewicki: I'm not sure but we followed this convention for some reason I think 😅

alduzy: It might have been auto-imported like that during development, that's what eslint suggests.
I built a clean app with screens installed from this branch just to make sure it works, it does 😄

PR #2411, opened 2024-10-16
feat!: expose library `compatibilityFlags` with new flag regarding header implementation

kkafar:
## Description

This PR introduces new `compatiblityFlags` object to the library public API. 
It is meant as a source of truth for the library consumers (especially `react-navigation`) to allow them to adjust to breaking changes
done in this library.

I've decided to go with single object with fields instead of individual constants, because it makes it easier for handling,
we don't have to change public exports every time we add something, it makes it easier to add/remove a flag.

## Changes

1. Created & exposed `compatiblityFlags` object,
2. added `usesHeaderFlexboxImplementation`,
2. moved old flags into the object.

N/A

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -72,7 +72,7 @@ export { GHContext } from './native-stack/contexts/GHContext';
  */
 export {
   isSearchBarAvailableForCurrentPlatform,
-  isNewBackTitleImplementation,
+  compatibilityFlags,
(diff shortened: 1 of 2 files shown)


satya164 (approved):

alduzy (approved):
LGTM

PR #2413, opened 2024-10-16
fix: native transition for native-like animations

kirillzyusko:
## Description

According to description:
- - `"simple_push"` – performs a default animation, but without shadow and native header transition (iOS only)
- - `"slide_from_left"` - slide in the new screen from right to left (Android only, resolves to default transition on iOS)

Both statements say that these transitions should act as a default one, but:
- default duration for these transitions is 350ms (by default, but it's customizable)
- these transitions looks similar to native transitions, but don't exactly match each other

Below is a comparison table comparing a `default` (native) animation with `simple_push`:

|Default|Simple push|
|--------|------------|
|<video src="https://github.com/user-attachments/assets/f9baae5f-443f-41e0-af2b-86428cf8b573">|<video src="https://github.com/user-attachments/assets/a7fefe55-9ced-40a5-a7d7-2a27a508448a">|

As you can see they are quite different. In this PR I'd like to propose to make them matching native transition (even though it will be managed by `UIView animateWithDuration` code).

For that I:
- changed default duration from 350ms to 500ms
- changed animation curve to match native transition.

In current PR shape it's a breaking change, because 500ms will be spreaded to `fade_from_bottom` animations, which are currently based on Android animations (and they rely on custom duration, like 250ms, if I understand the code correctly).

Anyway, I'll be glad to hear your feedback on whether you want to have something like this in the code in upcoming `v4` 🙌 

## Changes

- change `transitionDuration` to 500ms from 350ms;
- added `UIViewAnimationOptionCurveDefaultTransition`;
- use `UIViewAnimationOptionCurveDefaultTransition` in `simple_push`, `slide_from_left`, `slide_from_bottom`, `fade` animations;

### Before

https://github.com/user-attachments/assets/45b4a2fd-75f2-4d33-8e2c-4dba75fc1de3

### After

https://github.com/user-attachments/assets/782a625b-514f-4865-ae5a-e8555f17e858

You can use example app to test these changes. And you can take iOS Settings app as a reference for transitions.

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [x] Ensured that CI passes


android/src/paper/java/com/facebook/react/viewmanagers/RNSScreenManagerDelegate.java
```diff
@@ -91,7 +91,7 @@ public void setProperty(T view, String propName, @Nullable Object value) {
         mViewManager.setStackAnimation(view, (String) value);
         break;
       case "transitionDuration":
-        mViewManager.setTransitionDuration(view, value == null ? 350 : ((Double) value).intValue());
+        mViewManager.setTransitionDuration(view, value == null ? 500 : ((Double) value).intValue());
```
tboba, line 94: Out of curiosity, why we want to have `500ms` there? Is it taken from your own observations, or is it supported by iOS documentation?

ios/RNSScreenStackAnimator.mm
```diff
@@ -5,13 +5,14 @@
...
-static const float RNSFadeCloseTransitionDurationProportion = 0.15 / 0.35;
-static const float RNSFadeCloseDelayTransitionDurationProportion = 0.1 / 0.35;
+static const float RNSFadeOpenTransitionDurationProportion = 0.2 / 0.5;
+static const float RNSSlideCloseTransitionDurationProportion = 0.25 / 0.5;
+static const float RNSFadeCloseTransitionDurationProportion = 0.15 / 0.5;
+static const float RNSFadeCloseDelayTransitionDurationProportion = 0.1 / 0.5;
 // same value is used in other projects using similar approach for transistions
 // and it looks the most similar to the value used by Apple
 static constexpr float RNSShadowViewMaxAlpha = 0.1;
+static const int UIViewAnimationOptionCurveDefaultTransition = 7 << 16;
```
tboba, line 15: Should we add a comment why is it `7 << 16`?

tboba (commented):
Nice work! I haven't tested it, but just got some questions 😄 
Also, I believe this PR introduces breaking changes, as we're changing the default value of animation? So it's ideal to push it for `4.x` version of screens and mark it with `!` (I think).

ios/RNSScreenStackAnimator.mm
```diff
@@ -5,13 +5,14 @@
...
 // proportions to default transition duration
 static const float RNSSlideOpenTransitionDurationProportion = 1;
-static const float RNSFadeOpenTransitionDurationProportion = 0.2 / 0.35;
-static const float RNSSlideCloseTransitionDurationProportion = 0.25 / 0.35;
-static const float RNSFadeCloseTransitionDurationProportion = 0.15 / 0.35;
-static const float RNSFadeCloseDelayTransitionDurationProportion = 0.1 / 0.35;
+static const float RNSFadeOpenTransitionDurationProportion = 0.2 / 0.5;
+static const float RNSSlideCloseTransitionDurationProportion = 0.25 / 0.5;
+static const float RNSFadeCloseTransitionDurationProportion = 0.15 / 0.5;
+static const float RNSFadeCloseDelayTransitionDurationProportion = 0.1 / 0.5;
```
kkafar, line 11: Where these values come from? Do you have some kind of reference here, or are these found empirically?

ios/RNSScreenStackAnimator.mm
```diff
@@ -22,7 +23,7 @@ - (instancetype)initWithOperation:(UINavigationControllerOperation)operation
 {
   if (self = [super init]) {
     _operation = operation;
-    _transitionDuration = 0.35; // default duration in seconds
+    _transitionDuration = 0.5; // default duration in seconds
```
kkafar, line 26: Also does it not differ from one iOS version to another? Value of 0.35s is around for longer that I'm in this project 😅 but I'm **almost** sure that when it was introduced it was meant to resemble native animation as close as possible, hence my impression that this differs between various iOS versions. It would be nice to have some kind of reference, because changing it here to e.g. 0.5s might look "more natively" on iOS 18 but "less natively" on iOS 16 - this is my main objection right now.

android/src/paper/java/com/facebook/react/viewmanagers/RNSScreenManagerDelegate.java
kirillzyusko: @tboba I would say both 😅 I measured duration of animation in `QuickTime` player and it was showing ~480ms (which is very close to 500ms and 20ms could be due to incorrect cut out of a video).
And second observation is that `7 << 16` animation takes 500ms to complete 🙃

kkafar (commented):
Hey, I have few threads here.

First of all - thank you noticing this and taking your time to prepare the PR.

The code changes look solid, besides few remarks raised by @tboba and now by me.

My only objection is that I do not know, how "native slide animation" evolved between various iOS versions & I'm afraid that fixing it e.g. for iOS 18 will make matters worse for other iOS versions. 

Overall I like the changes & will want to proceed in this direction, however I need to do some research on how the animation has evolved between various iOS versions.

ios/RNSScreenStackAnimator.mm
kirillzyusko: The first time when I found it was in https://stackoverflow.com/questions/18870447/how-to-use-the-default-ios7-uianimation-curve
Should I add this as a comment above?

ios/RNSScreenStackAnimator.mm
kirillzyusko: Are you talking about 0.5? I captured a transition and measured the length of the video 🙈 

I think there is no any notes about that in official documentation, but it's 500ms and many iOS developers aware about that 🙃

kirillzyusko:
> Also, I believe this PR introduces breaking changes, as we're changing the default value of animation? So it's ideal to push it for 4.x version of screens and mark it with ! (I think).

Thank you for a comment @tboba 🙌 Actually I was trying to introduce as less as possible breaking changes there, and animation with 500ms and `7 << 16` curve looks and feels like the animation with 350ms duration. 150ms just spending in the end for 20-40 pixels and it's just for smooth stopping of the animation.

However what I noticed now is that it looks like my changes introduce breaking changes, because looks like `animationDuration` is not taking into consideration and you can not change the duration of the animation.

I think it's because `7 << 16` curve is a spring animation and in spring animation you can not control the duration of the animation because duration calculated only based on a physics rules 😔 

So we have kind of this:
- `7 << 16` is used natively on iOS to animate screen transitions;
- `7 << 16` doesn't allow to change duration since it's a spring-animation, so this PR introduce breaking changes 🙈 

Maybe it's worth to ask @WoLewicki (I think he was a person who introduced the concept of `simple_push` animation) why he decided to use default curves and 350ms animation? And what is the best way to handle changes from this PR?

I believe if we merge the PR in its current state, then we will introduce many breaking changes (especially for people who relied on `animationDuration`/`transitionDuration` property). Maybe we can create another param called `curved`/`interpolator` and pass it from JS?

Let me know what do you think 🙌

WoLewicki:
As for 350ms, I don't remember it exactly too, I think it was there already (yep: https://github.com/software-mansion/react-native-screens/blob/2.0/ios/RNSScreenStack.m#L336).

kirillzyusko:
@kkafar you reacted with emoji that scares me 😅 Do you have any ideas in your head how this PR can be adopted to introduce as less as possible breaking changes and at the same time to give a native transitions feeling if they are needed?

I think additional `curved`/`interpolation` properties would give that flexibility without introducing breaking changes, so the API could look like:

```ts
animation: 'simple_push',
interpolator: Interpolators.Default, // or Interpolators.EaseInEaseOut 
```

kkafar:
@kirillzyusko I admit I am not decided what is the best option here yet. I'll give it more consideration this afternoon and let you know how we want to proceed.

ios/RNSScreenStackAnimator.mm
kkafar: I was looking & playing around with this PR yesterday and came to conclusion that `7<< 16` value does not exist as valid animation options. The `option` parameter takes values of type `UIViewAnimationOptions` (mask combined from the options) & if you take a look on how it is defined:

```objective-c
typedef NS_OPTIONS(NSUInteger, UIViewAnimationOptions) {
    UIViewAnimationOptionLayoutSubviews            = 1 <<  0,
    UIViewAnimationOptionAllowUserInteraction      = 1 <<  1, // turn on user interaction while animating
    UIViewAnimationOptionBeginFromCurrentState     = 1 <<  2, // start all views from current value, not initial value
    UIViewAnimationOptionRepeat                    = 1 <<  3, // repeat animation indefinitely
    UIViewAnimationOptionAutoreverse               = 1 <<  4, // if repeat, run animation back and forth
    UIViewAnimationOptionOverrideInheritedDuration = 1 <<  5, // ignore nested duration
    UIViewAnimationOptionOverrideInheritedCurve    = 1 <<  6, // ignore nested curve
    UIViewAnimationOptionAllowAnimatedContent      = 1 <<  7, // animate contents (applies to transitions only)
    UIViewAnimationOptionShowHideTransitionViews   = 1 <<  8, // flip to/from hidden state instead of adding/removing
    UIViewAnimationOptionOverrideInheritedOptions  = 1 <<  9, // do not inherit any options or animation type
    
    UIViewAnimationOptionCurveEaseInOut            = 0 << 16, // default
    UIViewAnimationOptionCurveEaseIn               = 1 << 16,
    UIViewAnimationOptionCurveEaseOut              = 2 << 16,
    UIViewAnimationOptionCurveLinear               = 3 << 16,
    
    UIViewAnimationOptionTransitionNone            = 0 << 20, // default
    UIViewAnimationOptionTransitionFlipFromLeft    = 1 << 20,
    UIViewAnimationOptionTransitionFlipFromRight   = 2 << 20,
    UIViewAnimationOptionTransitionCurlUp          = 3 << 20,
    UIViewAnimationOptionTransitionCurlDown        = 4 << 20,
    UIViewAnimationOptionTransitionCrossDissolve   = 5 << 20,
    UIViewAnimationOptionTransitionFlipFromTop     = 6 << 20,
    UIViewAnimationOptionTransitionFlipFromBottom  = 7 << 20,

    UIViewAnimationOptionPreferredFramesPerSecondDefault     = 0 << 24,
    UIViewAnimationOptionPreferredFramesPerSecond60          = 3 << 24,
    UIViewAnimationOptionPreferredFramesPerSecond30          = 7 << 24,
    
} API_AVAILABLE(ios(4.0)) API_UNAVAILABLE(watchos);
```

using `7 << 16` rather does not make any sense (or am I getting this wrong)?

kkafar: Second thing is that we would rather construct our mask explicitly from available value in lieu of using a magic constant.

kkafar (commented):
First I have few remarks on `7 << 16` value, which seem to be incorrect (see below). Or maybe there are some hidden values not represented in public UIKit types?

Second thing is that I don't think we would want to completely disable opportunity to specify `animationDuration` prop. This is fine for `default` value, because it is supposed to look like the native one, however for the library-defined set we would want to keep the duration configurable. 

I'll be playing more with this PR right now, because it seems inadequate that specifying animation curve disables the `animationDuration`. Animation curve & duration are supposed to work tightly together and this should be doable, even if we need to do a little math. I'll get back to you once I have more deets.

kirillzyusko:
> First I have few remarks on 7 << 16 value, which seem to be incorrect (see below). Or maybe there are some hidden values not represented in public UIKit types?

Yes, this value is not available in public UIKit types.

> Second thing is that I don't think we would want to completely disable opportunity to specify animationDuration prop. This is fine for default value, because it is supposed to look like the native one, however for the library-defined set we would want to keep the duration configurable.

I agree with you, but it looks like `7 << 16` uses a different animation (`CASpringAnimation` vs `CABasicAnimation`) and in `CASpringAnimation` you can not specify duration. That's why I suggested to add `interpolator` property, so that people can use additional property if they want to have a closer animation to a native one (but they will loose an opportunity to customize duration - for me it's a fair alternative).

> Second thing is that we would rather construct our mask explicitly from available value in lieu of using a magic constant.

Not sure I understand how we can derive the value from existing options 🙈 Would you mind to give a code snippet or a hint? 👀

kkafar:
Okay, my conclusions after playing around a bit are as follows: 

1. I think I haven't understood you correctly initially. For some reason I've thought that you want to just disable animation duration and rely only on transition curves, but the idea of yours is that animation duration stops working only if you opt-in for specifying transition curve. I guess we could got with that, however I think that with a little more effort we could achieve the desired effect without any tradeoff necessity 👇🏻 
2. We want the default animation to resemble native-default-animation as closely as possible;
3. `UIView animateXXX` API does not allow for usage of custom timing curves thus to achieve desired effect we most likely need to refactor whole `RNSStackAnimator` to make use of Core Animation directly (animate the layers) and then just mimic default UIKit animation closely.

I think I have some throughput now.

I have to attend to sheetFooter a bit & various scrollview interactions in formSheets, however I might be able to do that this week.

kkafar:
@kirillzyusko I've updated implementation of simple_push & I would like to get your opinion whether you think this will be good enough now. Let me know, please. If we can adjust parameters to get satisfying result we'll go with the approach I've proposed in [39b599f](https://github.com/software-mansion/react-native-screens/pull/2413/commits/39b599fde70360b6ed0ecbaa19b280ff08e585e9)

kirillzyusko:
> I've updated implementation of simple_push & I would like to get your opinion whether you think this will be good enough now. Let me know, please. If we can adjust parameters to get satisfying result we'll go with the approach I've proposed in [39b599f](https://github.com/software-mansion/react-native-screens/pull/2413/commits/39b599fde70360b6ed0ecbaa19b280ff08e585e9)

@kkafar Cool, I'm going to check that implementation tomorrow or the day after tomorrow and will let you know! The approach looks promising, however I'm still concerned that we can control duration and use spring animation simultaneously... Will see! Thank you for the code!

kkafar:
I want to update you, that using the API of `UIViewPropertyAnimator` ruined the full screen swipe gesture, for some reason. I'm doing some more research now in order to fix it, however it might not work in current state of the PR, heads up.

kkafar:
I figured this out ☝🏻 !

ios/RNSScreenStackAnimator.mm
```diff
@@ -128,22 +132,33 @@ - (void)animateSimplePushWithShadowEnabled:(BOOL)shadowEnabled
...
-          }
-          fromViewController.view.transform = CGAffineTransformIdentity;
-          toViewController.view.transform = CGAffineTransformIdentity;
-          [transitionContext completeTransition:![transitionContext transitionWasCancelled]];
-        }];
+    // Default curve provider is as defined below, however defined this way it ignores the `animationDuration` prop.
+    // Damping of 1.0 seems close enough.
+    // id<UITimingCurveProvider> timingCurveProvider = [[UISpringTimingParameters alloc] init];
+
+    id<UITimingCurveProvider> timingCurveProvider = [[UISpringTimingParameters alloc] initWithDampingRatio:1.0];
```
kirillzyusko, line 139: I think such config is definitely better, but animation begins very slowly (unlike native iOS animation). Compare it yourself (I took 3rd frame for a reference):

image

I think next config gives a better results:

```objc
id<UITimingCurveProvider> timingCurveProvider = [[UISpringTimingParameters alloc] initWithMass:3 stiffness:1000 damping:500 initialVelocity:CGVectorMake(0, 0)];
```

image

Let me see if such config supports custom duration 👀

kirillzyusko: Okay, even though a new config is matching a native animation more closely - it doesn't allow to customize duration in this case 😔

kkafar: Most likely it does not, but let me know.

The spring curve for default animation is just `[[UISpringTimingParameters alloc] init]` (according to "iOS Programming Guide" by Matt Neuburg, there is no need to guess the params or use some magic constants 😅 )

kkafar: I've opened https://github.com/software-mansion/react-native-screens/pull/2477, as making this work requires more changes that I've thought initially. I'm in process of writing PR description but have some meetings currently.

kirillzyusko: @kkafar but if we are using `damping: 1`, then we have following transition:

|Variant         |1 frame|2 frame|3 frame|4 frame|5 frame|
|------------|-------|--------|--------|--------|--------|
|damping: 1 |image|image|image|image|image|
|Native iOS  |image|image|image|image|image|

For me it looks like a native iOS animation (`default`) is smoother than `simple_push` (with `damping: 1` it feels like it's moving slightly faster than I expect to).

kirillzyusko:
Closing in favour of https://github.com/software-mansion/react-native-screens/pull/2477

PR #2414, opened 2024-10-17
feat!: remove `shouldUseActivityState` flag

kkafar:
## Description

This PR removes `shouldUseActivityState` flag from public API (and in general).

The flag was initially introduced in #624, as a mean for downstream library developers
to differentiate between different screen versions. This was introduced over 4 years ago 
and currently does not serve any role in `react-navigation` implementation. 

## Changes

* Removed `shouldUseActivityState` flag from public API,
* added comment for `compatibilityFlags` object.

N/A

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -2,9 +2,6 @@
 
 import { Platform, UIManager } from 'react-native';
 
-// const that tells if the library should use new implementation, will be undefined for older versions
-export const shouldUseActivityState = true;
-
 export const isNativePlatformSupported =
   Platform.OS === 'ios' ||
   Platform.OS === 'android' ||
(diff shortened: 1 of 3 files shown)


alduzy (approved):
LGTM!
I found a commit that removed it from navigation: https://github.com/react-navigation/react-navigation/commit/64bcd58b416fddb3329cdf83d4f0d0dfb3f8eec0

maciekstosio (approved):
LGTM

WoLewicki (approved):
It was only needed for `react-navigation`, so if it's no longer used in the supported versions, then let's merge it.

PR #2419, opened 2024-10-18
refactor: replace unstable_screenStyle with backgroundColor from contenStyle

maciekstosio:
## Description

This PR removes of `unstable_screenStyle`

## Changes

Instead of adding `unstable_screenStyle` that directly adds background on `Screen` component, we can pick the `backgroundColor` from `contentStyle` and apply it on `Screen` component. Corresponding PR in react-navigation: https://github.com/react-navigation/react-navigation/pull/12218

## Screenshots / GIFs

N/A

See test Test1649.tsx 

## Checklist

- [x] Updated TS types
- [x] Ensured that CI passes


kkafar:
Hey https://github.com/software-mansion/react-native-screens/pull/2433 has been merged, we should be able to proceed now

maciekstosio:
Moved the changes from react-navigation to react-native-screens and created new PR in react-navigation: https://github.com/react-navigation/react-navigation/pull/12218

kkafar (approved):
I think we're looking good here!

src/components/ScreenStackItem.tsx
```diff
@@ -86,17 +101,16 @@ function ScreenStackItem({
...
       {isHeaderInModal ? (
         <ScreenStack style={styles.container}>
           <Screen
             enabled
             isNativeStack
             activityState={activityState}
             hasLargeHeader={headerConfig?.largeTitle ?? false}
-            style={StyleSheet.absoluteFill}
-          >
+            style={[StyleSheet.absoluteFill, internalScreenStyle]}>
```
alduzy, line 113: Do we need this style here?

kkafar: I think you're right. I haven't added it there initially (in the code that introduced the formsheets), because this Screen here is not the bottom sheet (it is not responsible for carrying out the presentation)!

I think this is good catch, but I'm only 95% sure. @maciekstosio can we test this?

maciekstosio: Yeah, I played with it and it seems that's not necessary. Removed in: https://github.com/software-mansion/react-native-screens/pull/2419/commits/40b78a8ad30ce56e74202ecead2248f7f92a5d34

kkafar (approved):
Okay! Looks like we're good now, thanks!

alduzy (approved):
🚀

PR #2420, opened 2024-10-18
refactor(Android): cleanup sheet behavior configuration

kkafar:
## Description

I plant to extract most of the form sheet API related code out of `ScreenStackFragment`, 
so that it is not polluted that much and more readable. Having experienced the #2045, where
big changes were rolled out at once, now I want to do it in couple of separate PRs.

This one simplifies sheet behaviour configuration logic in ScreenStackFragment by implementing 
series of helper extension functions on BottomSheetBehavior.

Test1649

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/bottomsheet/BottomSheetBehaviorExt.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/bottomsheet/BottomSheetBehaviorExt.kt
@@ -1,0 +1,42 @@
+package com.swmansion.rnscreens.bottomsheet
+
+import android.view.View
+import com.google.android.material.bottomsheet.BottomSheetBehavior
+
+internal fun <T : View> BottomSheetBehavior<T>.useSingleDetent(height: Int? = null, forceExpandedState: Boolean = true): BottomSheetBehavior<T> {
+    this.skipCollapsed = true
+    this.isFitToContents = true
+    if (forceExpandedState) {
+        this.state = BottomSheetBehavior.STATE_EXPANDED
+    }
+    height?.let { maxHeight = height }
+    return this
+}
+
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/swmansion/rnscreens/bottomsheet/BottomSheetBehaviorExt.kt
```diff
@@ -0,0 +1,42 @@
...
+        this.state = BottomSheetBehavior.STATE_EXPANDED
+    }
+    height?.let { maxHeight = height }
+    return this
+}
+
+internal fun <T : View> BottomSheetBehavior<T>.useTwoDetents(
+    @BottomSheetBehavior.StableState state: Int? = null,
+    firstHeight: Int? = null,
+    secondHeight: Int? = null
```
alduzy, line 19: Wouldn't it be better to stick with the original "peekHeight" and "maxHeight"?
Same applies to `useSingleDetent` and `useThreeDetents`.

kkafar: My intention was to abstract the `BottomSheetBehavior` properties away as much as possible, because `peekHeight` is not the "first detent height` and `maxHeight` is not `secondHeight` in general. They serve these functions only when other properties are set appropriately. 

I've yet failed to abstract it away from useThreeDetents, because situation is a little bit more complicated there & requires another refactor, which will come in the future.

alduzy (approved):
I am not 100% convinced with the names, but not against them either 😅 Great change overall, let's proceed

PR #2421, opened 2024-10-20
RFC + WIP: FullWindowOverlay on android

mfazekas:
## Description

FullWindowOverlay implementation for android

TODO:
- [ ] EventHandling

## Changes

- Added FullWindowOverlay implementation for android

The implementation is adding a FullWindowOverlayRootView to the window for every  FullWindowOverlay, FullWindowOverlay on native side is just a placeholder.

image

## Screenshots / GIFs

|*Android*|*iOS*|
|---|---|
|android|android|

See:
https://github.com/mfazekas/react-native-screens/blob/6f081fc33f1b8165df194eb3dd9ee42504c0d415/apps/src/tests/TestFullWindowOverlay.tsx#L1C1-L30C2

## Checklist

- [x] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/FullWindowOverlayViewManager.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/FullWindowOverlayViewManager.kt
@@ -1,0 +1,124 @@
+package com.swmansion.rnscreens
+
+import android.content.Context
+import android.graphics.PixelFormat
+import android.view.MotionEvent
+import android.view.View
+import android.view.WindowManager
+import com.facebook.react.bridge.ReactContext
+import com.facebook.react.bridge.UiThreadUtil
+import com.facebook.react.module.annotations.ReactModule
+import com.facebook.react.uimanager.JSPointerDispatcher
+import com.facebook.react.uimanager.JSTouchDispatcher
+import com.facebook.react.uimanager.RootView
+import com.facebook.react.uimanager.ThemedReactContext
+import com.facebook.react.uimanager.UIManagerHelper
+import com.facebook.react.uimanager.ViewGroupManager
+import com.facebook.react.uimanager.events.EventDispatcher
+import com.facebook.react.views.view.ReactViewGroup
+
+class FullWindowOverlayRootViewGroup(val reactContext: ThemedReactContext): ReactViewGroup(reactContext), RootView {
+    internal var eventDispatcher: EventDispatcher? = null
+
+    internal val jSTouchDispatcher: JSTouchDispatcher = JSTouchDispatcher(this)
+    internal var jSPointerDispatcher: JSPointerDispatcher? = null
+
+    override fun onChildStartedNativeGesture(childView: View, ev: MotionEvent) {
+        eventDispatcher?.let {
+            jSTouchDispatcher.onChildStartedNativeGesture(ev, it)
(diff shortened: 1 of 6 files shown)


maciekstosio:
Hi, thanks for this PR! The problem is a little bit more complex, I started to work on it but needed to switch focus to something else. Correct me if I'm wrong: 
- If you click in the overlay the the app will crash
- Overlay it's not edge to edge 
- Opening RN Modal after overlay will cover the overlay 

We need to try to tackle those problems because in this state it doesn't give us any advantage over RN Modal. I've created a draft with my changes - the PR https://github.com/software-mansion/react-native-screens/pull/2425 is messy, but it may give you some insights - I was able to solve some of the problems: 

**As far as I remember crashes can be solved by hardcoding the ID for the view:**
https://github.com/software-mansion/react-native-screens/blob/2bd05f41dd28a1cad58c537131e24fd9f772d3ad/android/src/main/java/com/swmansion/rnscreens/fullwindowoverlay/FullWindowOverlayRootViewGroup.kt#L51

**Passing events could be solved by dispatching the event to RootView of the FullWinowOverlay (the placeholder):**
https://github.com/software-mansion/react-native-screens/blob/2bd05f41dd28a1cad58c537131e24fd9f772d3ad/android/src/main/java/com/swmansion/rnscreens/fullwindowoverlay/FullWindowOverlayRootViewGroup.kt#L173

⚠️ Sometimes touches were translated in the underlying view (I think just in RN modals) or didn't work (depending on LayoutParams). The translation might be related to changes in the shadow tree for top Header: 

https://github.com/software-mansion/react-native-screens/blob/2bd05f41dd28a1cad58c537131e24fd9f772d3ad/common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.cpp#L27

**To go edge to edge I used different LayoutParams and setting width height in ShadowNode:** 
https://github.com/software-mansion/react-native-screens/blob/2bd05f41dd28a1cad58c537131e24fd9f772d3ad/android/src/main/java/com/swmansion/rnscreens/fullwindowoverlay/FullWindowOverlay.kt#L75
https://github.com/software-mansion/react-native-screens/blob/2bd05f41dd28a1cad58c537131e24fd9f772d3ad/android/src/main/java/com/swmansion/rnscreens/fullwindowoverlay/FullWindowOverlayViewManager.kt#L60

**Being on top of everything:** 
https://github.com/software-mansion/react-native-screens/blob/2bd05f41dd28a1cad58c537131e24fd9f772d3ad/android/src/main/java/com/swmansion/rnscreens/fullwindowoverlay/FullWindowOverlay.kt#L78 

⚠️ This was the closest to the desired outcome, but it requires granting "draw over other apps" permission, which we can't force users to enable - so we'll probably put it behind some flag for devs to decide. So, probably, the best for us will be to not support RN models and opt-in support for the permission behind the flag. We could try to reopen the window after each modal, but there isn't a clean way of observing if the RN modal is open: https://github.com/facebook/react-native/pull/45534, https://github.com/software-mansion/react-native-reanimated/pull/5916#issuecomment-2180751937

Btw. I'm not saying any of those solutions are the best approaches to tackle the issues - I'm open for discussion. Once again thanks for spending the time, and I'd be thrilled to work with you on getting it done.

kkafar:
Thanks for your work! I'll continuing implementation in #2733.

PR #2425, opened 2024-10-21
[WIP] Full window overlay for android

maciekstosio:
## Description

This PR intends to introduce FWO for Android. For now PR is a bit messy, posted just to showcase solution that could be useful for #2421.

Status: 
 - edge to edge ✅ 
 - passing events to overlayed content ⚠️ (depending on WindowManager.LayoutParams not always reliable or the clickes were shifted)
 - Covering content ⚠️ (it is always on top when using `WindowOverlayCompat.TYPE_SYSTEM_OVERLAY`, but requires granting permission "draw over other apps" - which we can't require from the user as a library, the workaround here can be putting `WindowOverlayCompat.TYPE_SYSTEM_OVERLAY` behind flag and using `TYPE_APPLICATION` otherwise)
 
## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


kkafar:
Thanks for your work! I'll continuing implementation in https://github.com/software-mansion/react-native-screens/pull/2733.

PR #2430, opened 2024-10-22
fix(iOS): FullWindowOverlay layout height

### LINKED ISSUE #1247: FullWindowOverlay not fitting the window
I m trying to use  `FullWindowOverlay`  to show a custom action sheet in the modal screen. but FullWindowOverlay has some spaces at the bottom.
In our app, I edited react-native-screens/ios/RNSFullWindowOverlay.m  ( version: 3.10.1) like this way. it seems working for me. but I m not sure this is the right way to do 
```
- (void)reactSetFrame:(CGRect)frame
{
  _reactFrame = frame;
  CGRect rect = [[UIScreen mainScreen] bounds];
  [_container setFrame:rect];
}
```
https://snack.expo.dev/@nazrdogan/suspicious-coffee

Image from iOS

alduzy:
## Description

This PR intents to fix `FullWindowOverlay`s height issue.

The component is given wrong frame size during layout because of it's placement in the react tree. Although in the iOS view hierarchy it is displayed independently its height is still reduced by the height of the header.

Correct frame can be achieved by utilising `useWindowDimensions` hook and forcing correct width and height on the JS side.

Fixes #1247 

## Changes

- modified `Test1096` repro 

### Before
Simulator Screenshot - iPhone 16 Pro - 2024-10-22 at 10 08 00

### After
Simulator Screenshot - iPhone 16 Pro - 2024-10-22 at 10 08 09

- use `Test1096` repro

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/components/FullWindowOverlay.tsx
+++ b/src/components/FullWindowOverlay.tsx
@@ -1,5 +1,12 @@
 import React, { PropsWithChildren, ReactNode } from 'react';
-import { Platform, StyleProp, View, ViewStyle } from 'react-native';
+import {
+  Platform,
+  StyleProp,
+  StyleSheet,
+  View,
+  ViewStyle,
+  useWindowDimensions,
+} from 'react-native';
 
 // Native components
 import FullWindowOverlayNativeComponent from '../fabric/FullWindowOverlayNativeComponent';
@@ -10,13 +17,14 @@ const NativeFullWindowOverlay: React.ComponentType<
 > = FullWindowOverlayNativeComponent as any;
 
 function FullWindowOverlay(props: { children: ReactNode }) {
+  const { width, height } = useWindowDimensions();
   if (Platform.OS !== 'ios') {
     console.warn('Using FullWindowOverlay is only valid on iOS devices.');
     return <View {...props} />;
   }
   return (
     <NativeFullWindowOverlay
-      style={{ position: 'absolute', width: '100%', height: '100%' }}>
+      style={[StyleSheet.absoluteFill, { width, height }]}>
       {props.children}
     </NativeFullWindowOverlay>
   );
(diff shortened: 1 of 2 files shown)


src/components/FullWindowOverlay.tsx
```diff
@@ -10,13 +17,14 @@ const NativeFullWindowOverlay: React.ComponentType<
...
 function FullWindowOverlay(props: { children: ReactNode }) {
+  const { width, height } = useWindowDimensions();
   if (Platform.OS !== 'ios') {
     console.warn('Using FullWindowOverlay is only valid on iOS devices.');
     return <View {...props} />;
   }
   return (
     <NativeFullWindowOverlay
-      style={{ position: 'absolute', width: '100%', height: '100%' }}>
+      style={[StyleSheet.absoluteFill, { width, height }]}>
```
kkafar, line 27: Do we need `absoluteFill` here, wasn't `position: 'absolute', width, height` enough?

kkafar (commented):
The code changes look very promising! I have only single remark and I think we can proceed. Good job 🎉

src/components/FullWindowOverlay.tsx
alduzy: We still need `top: 0` to ensure correct origin so I decided to utilise the `absoluteFill`.
simulator_screenshot_1A032E77-EDC4-40F0-A30E-6B934C7ACF78

kkafar: Okay, I see. That's fine with me. Just make sure this works as expected - buttons, pressables on full window overlay work correctly. I do not remember whether the buttons under FWO should be clickable, but if so - let's make sure the behaviour is unchanged. 

Provided that above assertions are upheld, I believe we're good to go.

kkafar (approved):
I have few test recommendations before merging. Beside this I think we're good to go.

maciekstosio (approved):
🚀

PR #2431, opened 2024-10-22
refactor: sync props

maciekstosio:
## Description

Sync ScreenNative and ModalScreenNative props.


kkafar (approved):
Yeah, thanks! This looks legit. 

I think we might want a CI that checks this ;D Let's create a ticket on our board for that.

PR #2433, opened 2024-10-22
feat: add a ScreenStackItem component

satya164:
Currently there is a lot of logic to make `Screens` work properly in native-stack, including:

- Making sure screens is enabled and mark it for native-stack
- Making sure screens knows that it contains a large title
- Rendering a `DebugContainer` for `LogBox`
- Making sure the `DebugContainer` contains appropriate logic and wrapper
- Rendering a nested stack to be able to show header in modals

This diff consolidates this logic to a new component `ScreenStackItem`. This makes it possible to move the above logic to the `react-native-screens` package by moving this component. Moving the logic will make it easier to fix bugs in screens.

Extracted from https://github.com/react-navigation/react-navigation/commit/fbc8635d22acadd2487f6bfc94338cbdac0d5a60

### DIFF
--- a/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
+++ b/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
@@ -30,20 +30,23 @@ When `activityState` is set to `0`, the parent container will detach its views f
 </ScreenContainer>
 ```
 
-When used in `<ScreenStack />` `activityState` can only be increased. The checks are added (in both native sides and JS part) to prevent situation when it's being removed, but still exists in in React Tree or if someones tries to preload already displayed screen. 
-
+When used in `<ScreenStack />` `activityState` can only be increased. The checks are added (in both native sides and JS part) to prevent situation when it's being removed, but still exists in in React Tree or if someones tries to preload already displayed screen.
 
 ## `<ScreenStack>`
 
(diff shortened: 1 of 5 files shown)


kkafar (commented):
I have one main concern:

this radically changes the API of screens - We'll now have `Screen` and `ScreenStackContent` components, which are "the same", however the latter should be used with `ScreenStack` and the former should be avoided or provided with additional fixes. 

I feel this should be avoided, but at the same time using `Screen` right now requires adding all these fixes on consumer end...

My second concern is naming: shouldn't we use like `StackScreen` or `NativeStackScreen` so that we're consistent? 

Also now we should add mention of this new component in the `GUIDE_FOR_LIBRARY_AUTHORS` - if you want I can take it from here, just let me know.

kkafar (commented):
I've just tested the runtime & it seems we're fine after these changes.

kkafar:
Also I think we'll want to bump react-navigation submodule to its current `main` in this PR also. If you want me to takeover from here just let me know.

satya164:
I have updated the guide for library author. Though seems the info regarding `activityState` is out of date there.

I'll let you updated the react-navigation submodule.

kkafar:
Ok, thanks. I'll do it in a moment.

PR #2434, opened 2024-10-22
fix: header config missing ref

alduzy:
## Description

This PR changes  `ScreenStackHeaderConfig` into a [forwardRef](@alduzy/reanimated-patch).

This is needed if you want to attach `ref` to it, which **react-native-reanimated** needs to do [here](https://github.com/software-mansion/react-native-reanimated/blob/1d417a5c711d60822d25bd8b823f1f5968214bde/apps/common-app/src/examples/ScreenStackHeaderConfigBackgroundColorExample.tsx#L24) in order to migrate to v4, since the header config is now a function component and cannot be used directly anymore.

Without this change the app crashes with `Cannot read property 'getScrollableNode' of null`.

Related PR: https://github.com/software-mansion/react-native-reanimated/pull/6622

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/components/ScreenStackHeaderConfig.tsx
+++ b/src/components/ScreenStackHeaderConfig.tsx
@@ -1,18 +1,12 @@
 'use client';
 
-import React, { Component, LegacyRef } from 'react';
+import React from 'react';
 import {
   HeaderSubviewTypes,
   ScreenStackHeaderConfigProps,
   SearchBarProps,
 } from '../types';
-import {
-  Image,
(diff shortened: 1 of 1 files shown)


kkafar (commented):
I'm not really into typescript magic, however isn't it main way to use `forwardRef` for function components?

maciekstosio (approved):

kkafar (approved):
I think we are good. Thanks!

PR #2437, opened 2024-10-24
fix: update getShadowNodeWrapperAndTagFromRef

maciekstosio:
## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/src/gesture-handler/fabricUtils.ts
+++ b/src/gesture-handler/fabricUtils.ts
@@ -1,13 +1,7 @@
-import { NativeStackNavigatorProps } from '../native-stack/types';
+'use strict';
+/* eslint-disable */
 
-interface HostInstance {
-  _internalInstanceHandle: {
-    stateNode: {
-      node: Record<string, unknown>;
-    };
-  };
-  _nativeTag: number;
-}
+import { NativeStackNavigatorProps } from '../native-stack/types';
 
 type LocalGlobal = typeof global & Record<string, unknown>;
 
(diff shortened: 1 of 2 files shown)


apps/src/screens/SwipeBackAnimation.tsx
```diff
@@ -27,12 +28,36 @@ interface ScreenBProps {
...
-  </View>
-);
+const ScreenB = ({ navigation }: ScreenBProps): React.JSX.Element => {
+  const [gestureType, setGestureType] = React.useState<NonNullable<NativeStackNavigationOptions['gestureType']>>('twoDimensionalSwipe');
+
+  React.useEffect(() => {
+    navigation.setOptions({
+      gestureType,
+    });
+  }, [gestureType]);
```
kkafar, line 38: Why do we move this into `setOptions` instead of leaving it set directly at screen level? 

Another thing is that me might use to opportunity to use `useLayoutEffect` here. I know it does not work as expected, but once it does we'll have this in place already.

src/gesture-handler/fabricUtils.ts
```diff
@@ -1,39 +1,84 @@
...
 
+let findHostInstance_DEPRECATED: (ref: unknown) => void;
+
+let getInternalInstanceHandleFromPublicInstance: (ref: unknown) => {
+  stateNode: { node: unknown };
+};
+
+// Taken and modifies from reanimated
 export function getShadowNodeWrapperAndTagFromRef(
   ref: React.Ref<NativeStackNavigatorProps> | React.Component,
```
kkafar, line 24: This is bad here. We need to somehow decouple it from native-stack implementation or move the whole implementation to react-navigation. But that is story for another set of PRs.

kkafar: Till the native-stack v5 is in the repo we're fine, however we'll have to move it at some point.

src/index.tsx
```diff
@@ -29,6 +29,8 @@ export {
   ScreenStackHeaderSearchBarView,
 } from './components/ScreenStackHeaderConfig';
 
+// NativeModules are not imported anywhere, but need to stay here so the TurboModule is registered
+export { default as NativeModules } from './fabric/NativeScreensModule';
```
kkafar, line 33: So this ended up being necessary as I've suggested, or this is just a left over? 

If this is needed - we don't want this in public API. It should be enough to import it in some other module loaded on any import from `react-native-screens/`

kkafar (commented):
My initial thoughts 👇🏻

src/gesture-handler/fabricUtils.ts
kkafar: I'm talking about using `NativeStackNavigatorProps` here.

apps/src/screens/SwipeBackAnimation.tsx
maciekstosio: I updated the PR description with a preview. I extended the example with the ability to switch between all possible gestures.

From what I researched for SearchBar, `useLayoutEffect` works as expected, and it won't solve our problems here.

src/gesture-handler/fabricUtils.ts
maciekstosio: I replaced it with generic Ref<React.Component> I don't think it's necessary to make it more precise as we're using it just to retrieve nativeTag and HostObject

kkafar (approved):
Looks good

src/index.tsx
```diff
@@ -1,3 +1,5 @@
+require('./fabric/NativeScreensModule'); // It needs to stay here so the TurboModule is registered
```
kkafar, line 1: This won't be needed anymore, we can settle with this: https://github.com/software-mansion/react-native-screens/commit/699c13edddc0a20737f244fc8c35a93c68e3057e

kkafar (changes requested):

src/index.tsx
kkafar: Just merge in the main

maciekstosio: Changed in: https://github.com/software-mansion/react-native-screens/pull/2437/commits/4df1fc2408102e323ef6a8c9dc4c731e7d4a785e

kkafar (approved):
We're good I think

PR #2439, opened 2024-10-24
fix(Android): center headerTitleAlign below Android API 29

### LINKED ISSUE #2435: headerTitleAlign set to "center" is broken on Android 4.0.0-beta.14
### Description

Hi,

I am using react-native-screens 4.0.0-beta.14 and I would like to center my title, can be always reproduced within the example of this repo:

https://github.com/user-attachments/assets/36cf5cab-6845-4b52-a671-392ec9c326a9

### Steps to reproduce

Either have a Navigation with headerTitleAlign set to center or clone my fork, run the example app, and follow the video

https://github.com/itsramiel/react-native-screens/tree/chore/reproduce

### Screens version

4.0.0-beta.14

### React Native version

0.74.5

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

Android emulator

### Device model

_No response_

### Acknowledgements

Yes

alduzy:
## Description

This PR intents to fix `headerTitleAlign: 'center'` breaking layout on android API <= 29.

I found changes added in https://github.com/software-mansion/react-native-screens/pull/2332, which fixed a related toolbar problem on android API <= 29, affect the correct layout of the centered title.

The `EXACTLY` [MeasureSpec](https://developer.android.com/reference/android/view/View.MeasureSpec) causes the title not to obtain as much space as it needs. Fortunately changing the spec to `AT_MOST` in the custom `layoutCallback` fixes the issue without introducing regression.

I also included a check for the current `softInputMode` to abandon the action if it is not set to `adjustPan`.

Fixes #2435 

## Changes

- updated `Test2332.tsx` repro
- changed measureSpec used in `layoutCallback`
- added `softInputMode` check

## Screenshots / GIFs

| Before https://github.com/software-mansion/react-native-screens/pull/2332 | Before | After |
| --- | --- | --- |
|  Screenshot_20241024_114026 | Screenshot_20241024_113943 | Screenshot_20241024_113853 |

- use `Test2332.tsx` repro

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
@@ -21,8 +21,8 @@ open class CustomToolbar(
             override fun doFrame(frameTimeNanos: Long) {
                 isLayoutEnqueued = false
                 measure(
-                    MeasureSpec.makeMeasureSpec(width, MeasureSpec.EXACTLY),
-                    MeasureSpec.makeMeasureSpec(height, MeasureSpec.EXACTLY),
+                    MeasureSpec.makeMeasureSpec(width, MeasureSpec.AT_MOST),
+                    MeasureSpec.makeMeasureSpec(height, MeasureSpec.AT_MOST),
                 )
                 layout(left, top, right, bottom)
             }
--- a/android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
@@ -3,9 +3,11 @@ package com.swmansion.rnscreens
 import android.annotation.SuppressLint
 import android.content.Context
 import android.os.Build
+import android.view.WindowManager
 import androidx.appcompat.widget.Toolbar
 import com.facebook.react.modules.core.ChoreographerCompat
 import com.facebook.react.modules.core.ReactChoreographer
+import com.facebook.react.uimanager.ThemedReactContext
 
 // This class is used to store config closer to search bar
 @SuppressLint("ViewConstructor") // Only we construct this view, it is never inflated.
@@ -28,7 +30,13 @@ open class CustomToolbar(
 
     override fun requestLayout() {
         super.requestLayout()
-        if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.Q) {
+        val softInputMode =
+            (context as ThemedReactContext)
+                .currentActivity
+                ?.window
+                ?.attributes
+                ?.softInputMode
+        if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.Q && softInputMode == WindowManager.LayoutParams.SOFT_INPUT_ADJUST_PAN) {
             // Below Android API 29, layout is not being requested when subviews are being added to the layout,
             // leading to having their subviews in position 0,0 of the toolbar (as Android don't calculate
             // the position of each subview, even if Yoga has correctly set their width and height).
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
```diff
@@ -19,16 +21,22 @@ open class CustomToolbar(
             override fun doFrame(frameTimeNanos: Long) {
                 isLayoutEnqueued = false
                 measure(
-                    MeasureSpec.makeMeasureSpec(width, MeasureSpec.EXACTLY),
-                    MeasureSpec.makeMeasureSpec(height, MeasureSpec.EXACTLY),
+                    MeasureSpec.makeMeasureSpec(width, MeasureSpec.AT_MOST),
+                    MeasureSpec.makeMeasureSpec(height, MeasureSpec.AT_MOST),
```
kkafar, line 25: 1. This is fine as long as it does not break any other behaviour. Please check header behaviours carefully, especially on newer Android API levels (34, 35).

2. Please add a comment here why this is done. 
3. Maybe we should select the measure spec depending on the API level being in use? So that we're sure we do not change the behaviour for newer API levels?

android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
```diff
@@ -19,16 +21,22 @@ open class CustomToolbar(
...
     override fun requestLayout() {
         super.requestLayout()
-        if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.Q) {
+        val softInputMode =
+            (context as ThemedReactContext)
+                .currentActivity
+                ?.window
+                ?.attributes
+                ?.softInputMode
+        if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.Q && softInputMode == WindowManager.LayoutParams.SOFT_INPUT_ADJUST_PAN) {
```
kkafar, line 39: I like introduction of this check, however I need you to confirm that issue from #2332 is not reintroduced.

kkafar (changes requested):

android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
alduzy: From what I've tested the issue mentioned only appeared when mode was set to `adjustPan`. I am not 100% certain though. @tboba Do you remember if it was an issue with other modes?

kkafar: I was involved in fixing it & can confirm that it occurred only for `adjustPan`.

However have you tested that? You do set this in `AndroidManifest` file of the application under the `softInputMode` key (or something similar).

android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
alduzy: @kkafar I believe this layoutCallback is not triggered in any other circumstances. It is called conditionally when API <= 29 and softInputMode is set to "adjustPan" in the modified `requestLayout` in the same file. Thus I don't think we need to add conditions/comment here.

android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
alduzy: I have tested it and didn't found any problems with softInputModes other than `adjustPan`, which is handled by this check.

android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
kkafar: Yes, I think you're right. I wrote this code but I've forgotten about it.

kkafar (approved):
Thanks for answering my concerns. I think we look good, beside one thing:

I'll insist you add a comment shortly explaining that these measure specs are selected to work for Android APIs level <= 29. 

Other than that - 🟢

PR #2447, opened 2024-10-25
fix(iOS): remove leftover _sheetsScrollView reference after cherry-picking from main

alduzy:
## Description

This PR fixes iOS build after recent cherry-picks. It removes the _sheetsScrollView leftover reference.

## Checklist

- [ ] Ensured that CI passes


kkafar (approved):
Makes sense

PR #2448, opened 2024-10-25
fix: removed leftover test import after cherry-picking from main

alduzy:
## Description

This PR removed leftover test import


kkafar (approved):
+1

PR #2450, opened 2024-10-25
refactor: Separate API props and NativeProps for ScreenStack 

maciekstosio:
## Description

Continuation of https://github.com/software-mansion/react-native-screens/pull/2423. Separate public `ScreenStackProps` and `NativeProps` by not casting NativeComponent as `React.ComponentType<ScreenStackProps> `. 

## Checklist

- [ ] Ensured that CI passes


kkafar (approved):
I think we're good here, thanks!

PR #2456, opened 2024-10-28
fix(Android): title invisible with searchbar enabled

alduzy:
## Description

This PR fixes title being invisible when searchbar is enabled.

In https://github.com/software-mansion/react-native-screens/pull/2395 I added `alignItems: 'center'` style to the header config component. It aimed to fix iOS subviews floating mid-screen when a larger subview is provided:

| (iOS) without `alignItems: 'center'` | (iOS) with `alignItems: 'center'`|
| --- | --- |
|  Simulator Screenshot - iPhone 16 Pro - 2024-10-28 at 09 36 53 |  Simulator Screenshot - iPhone 16 Pro - 2024-10-28 at 09 37 07 |

This change unintentionally introduced a regression on Android. When searchabar is enabled, the `ScreenStackHeaderConfig` components height is equal 0 and it does not 'stretch' to accomodate the native title. Ensuring it's height is at least 100% with `minHeight: '100%'` style fixes the issue.

| (Android) without `minHeight: '100%'` | (Android) with `minHeight: '100%'`|
| --- | --- |
| Screenshot_1730106776 | Screenshot_1730106784 |

## Changes

- updated `Test2395.tsx` repro
- ensured header config's height is at least filling the available space in the native header.

- use `Test2395.tsx` repro

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/src/components/ScreenStackHeaderConfig.tsx
+++ b/src/components/ScreenStackHeaderConfig.tsx
@@ -109,5 +109,6 @@ const styles = StyleSheet.create({
     flexDirection: 'row',
     justifyContent: 'space-between',
     alignItems: 'center',
+    minHeight: '100%',
   },
 });
--- a/apps/src/tests/Test2395.tsx
+++ b/apps/src/tests/Test2395.tsx
@@ -1,7 +1,17 @@
 import * as React from 'react';
-import { FlatList, Pressable, StyleSheet, Text, View } from 'react-native';
-import { NavigationContainer } from '@react-navigation/native';
-import { createNativeStackNavigator } from '@react-navigation/native-stack';
+import {
+  Button,
+  FlatList,
+  Pressable,
+  StyleSheet,
+  Text,
+  View,
+} from 'react-native';
+import { NavigationContainer, ParamListBase } from '@react-navigation/native';
+import {
+  createNativeStackNavigator,
+  NativeStackScreenProps,
+} from '@react-navigation/native-stack';
 
 const Stack = createNativeStackNavigator();
 
@@ -10,8 +20,8 @@ export default function App() {
     <NavigationContainer>
       <Stack.Navigator>
         <Stack.Screen
-          name="Title"
-          component={Screen}
+          name="First"
+          component={First}
           options={{
             headerRight: () => (
               <View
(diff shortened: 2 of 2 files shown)


src/components/ScreenStackHeaderConfig.tsx
```diff
@@ -109,5 +109,6 @@ const styles = StyleSheet.create({
     flexDirection: 'row',
     justifyContent: 'space-between',
     alignItems: 'center',
+    minHeight: '100%',
```
kkafar, line 112: Doesn't it set the header config height to the height of its parent? We don't want the header to be of full screen height, do we?

kkafar (commented):
My initial question 👇🏻

src/components/ScreenStackHeaderConfig.tsx
alduzy: I thought It's taking the height available in the AppBarLayout, but after opening it in the layout inspector it really is the height of the screen, just positioned strangely:
Screenshot 2024-10-28 at 11 41 35

Got CI failures because of it as well, I'm gonna change it 😉

src/components/ScreenStackHeaderConfig.tsx
```diff
@@ -108,6 +115,6 @@ const styles = StyleSheet.create({
     width: '100%',
     flexDirection: 'row',
     justifyContent: 'space-between',
-    alignItems: 'center',
+    alignItems: Platform.OS === 'ios' ? 'center' : undefined,
```
kkafar, line 118: Yeah, I think we might be better off with this approach.

`minHeight` / `height` set to percents sets the value basing on the dimensions of the containing box in the css model **in the `ShadowTree`**. `AppBarLayout` does not exist in `ShadowTree`. `Screen` is the parent of the `ScreenStackHeaderConfig` in the `ShadowTree`.

kkafar (approved):
Yeah, we might be better off with this approach.

src/components/ScreenStackHeaderConfig.tsx
satya164: I think it'd be good to add a comment describing why this is done like this for people who edit this code in the future

kkafar:
Note: line of comment and link to the PR might be useful somewhere close to these styles. Otherwise this code is unreadable.

src/components/ScreenStackHeaderConfig.tsx
alduzy: https://github.com/software-mansion/react-native-screens/pull/2456/commits/a6849f6a62729eda480300b014e135135d633254

PR #2462, opened 2024-10-29
fix(Android): formSheet flex 1 support

alduzy:
This PR was originally posted in react-navigation: https://github.com/react-navigation/react-navigation/pull/12196
Since https://github.com/software-mansion/react-native-screens/pull/2433 the changes should be applied here.

This change relies on replacing `ScreenStackContent` in react-navigation with this `ScreenStackItem` component imported from react-native-screens. ~Until it's done this PR is a draft.~ Keep that in mind if you want to test this change.

## Description

This PR intents to add `flex: 1` support to android formSheets. Currently when you set `flex: 1` style to a wrapper component (quite usual thing to do) inside a screen with `presentation: 'formSheet'` it doesn't show up on android. On iOS it works fine.

This adjustment prevents the android sheet from breaking when given the `flex: 1` style and ensures that its behaviour aligns better with iOS.

While iOS is given sheet specific styles, Android receives `null`. Passing same styles to both platforms makes the `flex: 1` work as expected. To ensure that the scrollview on Android knows it's height and scrolls properly, a `maxHeight: '100%'` style is added.

Additionally, on Android there's no header in `formSheet` presentation. Thus we want to explicitly set it to hidden, as it may influence the content's  layout.

## Changes

- adjusted layout styles for `formSheet` presentation.
- ensured header is hidden on android sheet
- added `Test2462.tsx` repro

### Before

| without backgroundColor | with backgroundColor |
|---|---|
|  Screenshot 2024-10-29 at 11 44 38 | Screenshot 2024-10-29 at 11 44 47 |

### After

| without backgroundColor | with backgroundColor |
|---|---|
|  Screenshot 2024-10-29 at 11 47 03 | Screenshot 2024-10-29 at 11 43 57 |

- use `Test2462.tsx` repro

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/components/ScreenStackItem.tsx
+++ b/src/components/ScreenStackItem.tsx
@@ -77,7 +77,13 @@ function ScreenStackItem(
        * See https://github.com/software-mansion/react-native-screens/pull/1825
        * for detailed explanation.
        */}
-      <ScreenStackHeaderConfig {...headerConfig} />
+      <ScreenStackHeaderConfig
+        {...headerConfig}
+        hidden={
+          headerConfig?.hidden ||
+          (stackPresentation === 'formSheet' && Platform.OS === 'android')
+        }
+      />
     </>
   );
 
--- a/src/components/ScreenStackItem.tsx
+++ b/src/components/ScreenStackItem.tsx
@@ -22,14 +22,17 @@ type Props = Omit<
   contentStyle?: StyleProp<ViewStyle>;
 };
 
-function ScreenStackItem({
-  children,
-  headerConfig,
-  activityState,
-  stackPresentation,
-  contentStyle,
-  ...rest
-}: Props, ref: React.ForwardedRef<View>) {
+function ScreenStackItem(
+  {
+    children,
+    headerConfig,
+    activityState,
+    stackPresentation,
+    contentStyle,
+    ...rest
+  }: Props,
+  ref: React.ForwardedRef<View>,
+) {
   const isHeaderInModal =
     Platform.OS === 'android'
       ? false
@@ -42,7 +45,7 @@ function ScreenStackItem({
       Platform.OS !== 'android' &&
         stackPresentation !== 'push' &&
         headerHiddenPreviousRef.current !== headerConfig?.hidden,
-      `Dynamically changing header's visibility in modals will result in remounting the screen and losing all local state.`
+      `Dynamically changing header's visibility in modals will result in remounting the screen and losing all local state.`,
     );
 
     headerHiddenPreviousRef.current = headerConfig?.hidden;
@@ -53,14 +56,14 @@ function ScreenStackItem({
       <DebugContainer
         style={[
           stackPresentation === 'formSheet'
-            ? Platform.OS === 'ios'
-              ? styles.absolute
-              : null
+            ? {
+                ...styles.sheet,
+                maxHeight: Platform.OS === 'android' ? '100%' : undefined,
+              }
             : styles.container,
           contentStyle,
         ]}
-        stackPresentation={stackPresentation ?? 'push'}
-      >
+        stackPresentation={stackPresentation ?? 'push'}>
         {children}
       </DebugContainer>
       {/**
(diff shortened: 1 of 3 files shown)


satya164:
> This change relies on replacing ScreenStackContent in react-navigation with this ScreenStackItem component imported from react-native-screens. Until it's done this PR is a draft.

The bugfix needs React Navigation to change this import, but this bugfix doesn't depend on React Navigation doing it first. Even if you merge it first here and React Navigation bumps the version afterwards, the outcome is the same. So please go ahead with merging and releasing this first. That way I can bump the version in React Navigation and make releases once.

satya164 (commented):
I'd suggest moving/copying explanation from PR to code comments - that way when someone is reading the code it'll be more obvious what's happening without having to visit PRs.

src/components/ScreenStackItem.tsx
```diff
@@ -74,7 +77,13 @@ function ScreenStackItem({
        * See https://github.com/software-mansion/react-native-screens/pull/1825
        * for detailed explanation.
        */}
-      <ScreenStackHeaderConfig {...headerConfig} />
+      <ScreenStackHeaderConfig
+        {...headerConfig}
+        hidden={
+          headerConfig?.hidden ||
+          (stackPresentation === 'formSheet' && Platform.OS === 'android')
```
satya164, line 84: What does this change do?

src/components/ScreenStackItem.tsx
```diff
@@ -53,14 +56,14 @@ function ScreenStackItem({
       <DebugContainer
         style={[
           stackPresentation === 'formSheet'
-            ? Platform.OS === 'ios'
-              ? styles.absolute
-              : null
+            ? {
+                ...styles.sheet,
+                maxHeight: Platform.OS === 'android' ? '100%' : undefined,
```
satya164, line 61: Does this break anything on iOS?

src/components/ScreenStackItem.tsx
alduzy: We want to ensure the header is not considered when layouting android formSheet, since it doesn't have one. It prevents sheet content height from being reduced by header height, like so:

Screenshot_1730203067

src/components/ScreenStackItem.tsx
alduzy: I can't remember if there were any problems, can't find any right now. I believe it's just extra precaution.

alduzy: Having a single style would look neat though

src/components/ScreenStackItem.tsx
satya164: It'd be good to add a comment here

kkafar (commented):
I won't approve this until we understand why this works. I don't get why setting `maxHeight` this way would impact the layout as it did. 

If you're able - please provide us with some insights - otherwise I'll be able to dive into it ~todays afternoon tomorrow morning

alduzy:
@kkafar I added few more lines to description to make it more clear.

Basically changing the styles on android from `null` to those used on iOS is the most important step that adds `flex: 1 ` support.

Then, as I was testing the changes with various sheet contents, I had to add `maxHeight` and ensure the header is hidden on Android for the scrollview to layout and scroll correctly.

src/components/ScreenStackItem.tsx
alduzy: https://github.com/software-mansion/react-native-screens/pull/2462/commits/b110017917e3bc6e76e94d9d036c5305d5a085a5

RodolfoSilva:
@alduzy do you have any workaround suggestion for this?

dustinvforsyth:
I was having issues with this even without using `flex: 1`. I tried adding `height: '100%'` and a bunch of other changes. Sometimes the model would pop up and look great other times it would be sized wrong. The worst part was it wasn't just the `formSheet` it was `modal` and `fullscreenModal` as well that were broken. 

Weirdly enough disabling the new arch solved all my modal problems. 

That being said it seemed like as long as the content filled the screen everything was fine but if I was trying to make the `formSheet`, `modal` or `fullscreenModal` fill the screen it was inconsistent.

Seemed like a race condition to me if I opened the modal 10 times it might layout correctly 8 of the 10, with two being messed up.

PR #2464, opened 2024-10-29
feat(Android,**unstable**): integration with `react-native-edge-to-edge`

zoontek:
> [!note]
> note by @kkafar: 
> Please note that I merge this as unstable behaviour, since I might want to remove the dependency & code once we land https://github.com/facebook/react-native/pull/47554 in core or endup with only edge-to-edge supported by Android.

## Description

This follow a discussion started [on the react-navigation repository](https://github.com/react-navigation/react-navigation/pull/12217).

## Motivation

The future of Android is [edge-to-edge](https://developer.android.com/about/versions/15/behavior-changes-15#edge-to-edge), and to make the React Native developer experience seamless in this regard, the ecosystem needs to transition from “opaque system bars by default” to “edge-to-edge by default.”

To prevent library authors from implementing their own edge-to-edge solutions—which could interfere with other libraries—and because it’s not possible to reliably detect if edge-to-edge is already enabled on Android, we have collaborated with [Expo](https://expo.dev/) to create a library that handles this functionality and is detectable using a simple helper: [`react-native-is-edge-to-edge`](https://github.com/zoontek/react-native-edge-to-edge/tree/main/react-native-is-edge-to-edge).

This approach allows you to bypass certain options and props (in this case, `Screen` `statusBarTranslucent`, `navigationBarTranslucent`, `statusBarColor` and `navigationBarColor` - [setting background color is deprecated](https://developer.android.com/about/versions/15/behavior-changes-15#deprecated-apis)). All are Android only props and are now obsolete / deprecated (at least, when running in edge-to-edge mode)

## Changes

- Add `react-native-is-edge-to-edge` to bypass `statusBarTranslucent`, `navigationBarTranslucent`, `statusBarColor` and `navigationBarColor` values
- Add a warning about their usage when running in edge-to-edge mode.

- Install [react-native-edge-to-edge](https://github.com/zoontek/react-native-edge-to-edge) in the Example app.
- Force `react-navigation` to use local version of `react-native-screens`.

> [!WARNING]  
> For the `StatusBar` example, `react-native` `headerTopInsetEnabled` might interfere when `statusBarTranslucent` is explicitely set to `false`, but the user will receive a warning inviting him to remove the option / prop, so that's acceptable.

## More

If you want to bypass some Kotlin code directly, consider this util:

```kotlin
object EdgeToEdgeUtil {
  val ENABLED: Boolean
    get() = try {
      // we cannot detect edge-to-edge, but we can detect react-native-edge-to-edge install
      Class.forName("com.zoontek.rnedgetoedge.EdgeToEdgePackage")
      true
    } catch (exception: ClassNotFoundException) {
      false
    }
}
```

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/Example/package.json
+++ b/Example/package.json
@@ -28,6 +28,7 @@
     "nanoid": "^4.0.2",
     "react": "19.0.0",
     "react-native": "0.78.0-rc.5",
+    "react-native-edge-to-edge": "^1.6.0",
     "react-native-gesture-handler": "2.22.0",
     "react-native-reanimated": "3.17.1",
     "react-native-restart": "^0.0.27",
--- a/Example/yarn.lock
+++ b/Example/yarn.lock
@@ -3101,6 +3101,7 @@ __metadata:
     react: "npm:19.0.0"
     react-native: "npm:0.78.0-rc.5"
     react-native-codegen: "npm:^0.71.3"
+    react-native-edge-to-edge: "npm:^1.6.0"
     react-native-gesture-handler: "npm:2.22.0"
     react-native-reanimated: "npm:3.17.1"
     react-native-restart: "npm:^0.0.27"
@@ -8856,6 +8857,16 @@ __metadata:
     jscodeshift: "npm:^0.14.0"
     nullthrows: "npm:^1.1.1"
   checksum: 10c0/7daca7b2317dda277a4f8f230ac141ef575c0e6376eeb3cd333480fea9dc41b0b4e31b8521c1f9d496f7cd887e47c4885f2d330c902ddb07539e7af764a4109c
+  languageName: node
+  linkType: hard
+
+"react-native-edge-to-edge@npm:^1.6.0":
+  version: 1.6.0
+  resolution: "react-native-edge-to-edge@npm:1.6.0"
+  peerDependencies:
+    react: "*"
+    react-native: "*"
+  checksum: 10c0/6373cc1b447eae31689a9b62e38b15621e9273626e2324700c4c3eb58c02ce489236a4b9e3e0dc1187e062defd8316195c5b1213facd718706b79b92127a05a3
   languageName: node
   linkType: hard
 
--- a/FabricExample/package.json
+++ b/FabricExample/package.json
@@ -27,6 +27,7 @@
     "nanoid": "^4.0.2",
     "react": "19.0.0",
     "react-native": "0.78.0-rc.5",
+    "react-native-edge-to-edge": "^1.6.0",
     "react-native-gesture-handler": "2.22.0",
     "react-native-reanimated": "^4.0.0-nightly-20250218-e5a0cdf69",
     "react-native-restart": "^0.0.27",
--- a/FabricExample/yarn.lock
+++ b/FabricExample/yarn.lock
@@ -2952,6 +2952,7 @@ __metadata:
     prettier: "npm:2.8.8"
     react: "npm:19.0.0"
     react-native: "npm:0.78.0-rc.5"
+    react-native-edge-to-edge: "npm:^1.6.0"
     react-native-gesture-handler: "npm:2.22.0"
     react-native-reanimated: "npm:^4.0.0-nightly-20250218-e5a0cdf69"
     react-native-restart: "npm:^0.0.27"
@@ -8522,6 +8523,16 @@ __metadata:
   version: 19.0.0
   resolution: "react-is@npm:19.0.0"
   checksum: 10c0/d1be8e8500cf04f76df71942a21ef3a71266397a383d7ec8885f35190df818d35c65efd35aed7be47a89ad99aaff2c52e0c4e39e8930844a6b997622e50625a8
+  languageName: node
+  linkType: hard
+
+"react-native-edge-to-edge@npm:^1.6.0":
+  version: 1.6.0
+  resolution: "react-native-edge-to-edge@npm:1.6.0"
+  peerDependencies:
+    react: "*"
+    react-native: "*"
+  checksum: 10c0/6373cc1b447eae31689a9b62e38b15621e9273626e2324700c4c3eb58c02ce489236a4b9e3e0dc1187e062defd8316195c5b1213facd718706b79b92127a05a3
   languageName: node
   linkType: hard
 
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
   "homepage": "https://github.com/software-mansion/react-native-screens#readme",
   "dependencies": {
     "react-freeze": "^1.0.0",
-    "react-native-is-edge-to-edge": "^1.1.6",
+    "react-native-is-edge-to-edge": "^1.1.7",
     "warn-once": "^0.1.0"
   },
   "peerDependencies": {
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
   "homepage": "https://github.com/software-mansion/react-native-screens#readme",
   "dependencies": {
     "react-freeze": "^1.0.0",
+    "react-native-is-edge-to-edge": "^1.1.6",
     "warn-once": "^0.1.0"
   },
   "peerDependencies": {
--- a/src/components/Screen.tsx
+++ b/src/components/Screen.tsx
@@ -194,15 +194,6 @@ export const InnerScreen = React.forwardRef<View, ScreenProps>(
       ...rest
     } = props;
 
-    if (__DEV__) {
-      controlEdgeToEdgeValues({
-        navigationBarColor,
-        navigationBarTranslucent,
-        statusBarColor,
-        statusBarTranslucent,
-      });
-    }
-
     // To maintain default behavior of formSheet stack presentation style and to have reasonable
     // defaults for new medium-detent iOS API we need to set defaults here
     const {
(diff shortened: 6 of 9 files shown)


kkafar:
Hey @zoontek I took first glance I want to reiterate to make sure I get this right: 

the goal here is to extract status bar / navigation bar management in terms of its visibility from all packages beside `react-native-is-edge-to-edge`, so that there is one, centralised solution they (other packages) can use to detect, whether the app is in edge-to-edge. Moreover -  manipulation of visibility of these elements should be done only through API of this package.

Note that changes you suggest are **breaking**.

I like the direction in general, assuming that we get broader adoption this might be a good thing. Do yo have a list of packages that need to undergo these changes prepared? If so, please share it. I'm asking for it, because it is related to my another point below 👇🏻 

I'm reluctant to introduce any third party dependency, especially one with native code. `react-native-screens` is a upstream package and such dependencies become liabilities especially when it comes to adopting to breaking changes rolled out in react-native between versions. We have this situation already partially (because only in example-apps, not in library directly) with different package, which requires us often to provide patches and upstream them, adding maintenance burden. Having this is mind I wonder whether it would be maybe better direction to get rid of this set of functionalities all together from `react-native-screens`? I haven't made my mind yet. I'm willing to change it. I'll forward this PR internally in order to get some comments on whether we should do it or not.

satya164:
> I'm reluctant to introduce any third party dependency, especially one with native code.

afaik it's a JS only and lightweight dep that checks if the `react-native-edge-to-edge` module is available.

kkafar:
> As Android 15+ (targetting SDK 15) doesn't allow opaque system bars and enforces apps to draw under, the library enforces edge-to-edge and actually don't allow switching back to opaque bars anymore (because it will be impossible in the future - same for setting bars background color) - it doesn't want to handle their visibility, it aligns with (future) Android enforced behaviour.

I see, makes even more sense now. 

> I understand, but there's currently no native code involved in the PR, it just detect if another package is installed.

Yeah, I've missed the difference between `react-native-is-edge-to-edge` and `react-native-edge-to-edge` and inspected only the code of the latter. Dependency on `react-native-is-edge-to-edge` is much more acceptable.

> [...] we had only two options: a common library + lib detection or the feature being included in RN core (we asked for it, it might arrive at some point, but not fot the moment).

It would be good to upstream it to core at some point, definitely. I see your point though.


Thank you for detailed explanation! I'm leaning towards accepting this change, but need some more time to think this through (maybe we want to go more radically and get rid of some future-scope from this library entirely).

zoontek:
@kkafar If it helps take a decision, a new library added support: [react-native-bottom-tabs](https://okwasniewski.github.io/react-native-bottom-tabs/docs/guides/edge-to-edge-support.html) (which plays nice with `react-navigation)`

zoontek:
@kkafar FYI, we continue pushing for it. The react native team just merged [edge-to-edge for Modal into core](https://github.com/facebook/react-native/pull/47254#issuecomment-2461335718).

satya164:
@zoontek I'm curious whether these also apply to React Native's `StatusBar` component/module since it also changes status bar's color/translucency like the code here.

satya164:
@zoontek the other thing you mentioned here is that user needs to manually add these options until this PR is merged https://github.com/zoontek/react-native-edge-to-edge/issues/45#issuecomment-2472863904

But by default these values are `undefined` iirc, and they'll also be `undefined` after this PR is merged. So unless user is explicitly overriding somewhere, then there shouldn't be any change in the default behavior, right? Or was this a misunderstanding?

zoontek:
@satya164 Yes, there's no need to manually set them indeed, only to avoid that those are not set to `undefined`. My first though here was that something set`statusBarTranslucent` / `navigationBarTranslucent` to `false` somewhere in the dependencies chain or other were set to `false` somewhere (in `react-navigation`? in `expo-router`? I didn't checked yet at the time I wrote the reply), but the issue was with builtin `StatusBar` indeed.

If those stay `undefined`, there's indeed no issue. But it's very easy to imagine that a dependency, somewhare use:

```tsx
type Props = {
  statusBarTranslucent?: boolean
}

// Here, defaulting to `false` breaks edge-to-edge, it's vicious
const MyComponent = ({ statusBarTranslucent = false }: Props) => {
  return <Screen statusBarTranslucent={statusBarTranslucent} />
}
```

This kind of issues are super hard to debug if you don't understand the topic perfectly, so a nightmare for users that only want to enable it and stop thinking about it.

> @zoontek I'm curious whether these also apply to React Native's StatusBar component/module since it also changes status bar's color/translucency like the code here. Does it make sense to prevent the user from changing statusbar color if they enable edge to edge? What if they want to do it on one screen?

`StatusBar` is the elephant in the room. They put a warning about `backgroundColor` in the documentation:

Screenshot 2024-11-18 at 14 04 23

But there's still in the module [some code that breaks edge-to-edge](https://github.com/facebook/react-native/blob/v0.76.2/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/view/WindowUtil.kt#L59): we saw it with the `expo-router` issue, that just mounted a `StatusBar` in `ExpoRoot`. For the moment, `react-native-edge-to-edge` `SystemBars` component is the safe way to go.

To tackle the edge-to-edge issue, Meta team asked me for a proposal, it started [here](https://github.com/facebook/react-native/pull/47554) (very WIP). The idea is similar:

- Having a way to enable edge-to-edge easily (`enableEdgeToEdge=true`)
- Give to library authors an easy way to detect if it's enabled in native code (`BuildConfig.IS_EDGE_TO_EDGE_ENABLED`)
- Give to non-native libraries authors a way to detect if it's enabled in JS (`Appearance.isEdgeToEdge()`)

And of course, this PR takes care of [fixing the StatusBar module behavior](https://github.com/facebook/react-native/pull/47554/files#diff-b56ad144ce3d7787344bc9a51c568eafaa0e377f193390a02739723ea0dfebcd).

zoontek:
> Does it make sense to prevent the user from changing statusbar color if they enable edge to edge?

For me, yes. As it will not work in the future (when targetSDK = 35 and on Android 15+):

Screenshot 2024-11-18 at 15 17 55

Note that the users can create a `<View />` with `height: insets.top` and a background color. This still works on Android 15+ and it also works on iOS.

satya164:
I think if edge-to-edge will be part of core, in addition to these, it'll be better to mark these props as deprecated as well so they can be removed entirely in the future.

kkafar:
*Note: List of resources is for future me* 😅 

I've familiarised myself with pretty much all the resources you guys linked:

1. [Android behaviour changes when targeting API Level 35+](https://developer.android.com/about/versions/15/behavior-changes-15#window-insets)
2. @zoontek's [Proposal to core](https://github.com/facebook/react-native/pull/47554)
3. [Last year's timeline of enforcing minimal `targetSdk` for PlayStore apps](https://support.google.com/googleplay/android-developer/answer/11926878?sjid=11853000253346477363-EU#zippy=%2Care-there-any-exceptions-for-existing-apps-targeting-api-or-below)
4. Once again with all the discussion above ☝🏻 
5. Source code of both [`react-native-edge-to-edge` & `react-native-is-edge-to-edge`](https://github.com/zoontek/react-native-edge-to-edge?tab=readme-ov-file)

Conclusions: 

the arguments made here are compelling. We'll proceed with merging this PR, after I'm done with testing (should be able to do this by the end of the week). **My only concerns right now are Fabric-related**, as there is no such thing as `undefined` value for props on new architecture - codegened component's code always passes some non-null (non-undefined) default value, therefore we need to ensure that these defaults are appropriate. 

Another thing I want to address is handling of `topInsetEnabled` prop on `HeaderConfig` [(link)](https://github.com/react-navigation/react-navigation/blob/113cea5770fbca7cc86594d7dac1f00c2c392eec/packages/native-stack/src/views/NativeStackView.native.tsx#L229-L238). It would be nice if we had knowledge of whether the app is in edge-to-edge in react-navigation code, to be able to set this prop and prevent content jumping appropriately - I'm thinking whether we should not consider adding `react-native-is-edge-to-edge` as dependency downstream as well @satya164. Alternative approach would be to remove this `topInsetEnabled` property entirely - but this is not feasible before the changes land in core, I believe. 

WIP WIP (I ned to write few more paragraphs)

zoontek:
> My only concerns right now are Fabric-related, as there is no such thing as undefined value for props on new architecture - codegened component's code always passes some non-null (non-undefined) default value, therefore we need to ensure that these defaults are appropriate.

@kkafar  We could also update the code to be something like this:

```tsx
const edgeToEdgeProps = EDGE_TO_EDGE
  ? {
      navigationBarTranslucent: true,
      statusBarTranslucent: true,
    }
  : {
      ...(navigationBarColor != null && { navigationBarColor }),
      ...(navigationBarTranslucent != null && { navigationBarTranslucent }),
      ...(statusBarColor != null && { statusBarColor }),
      ...(statusBarTranslucent != null && { statusBarTranslucent }),
    };
    
// …

return <AnimatedScreen
  {...props}
  {...edgeToEdgeProps}
```

zoontek:
@kkafar @satya164 I rebased against main and updated the PR to avoid passing `statusBarTranslucent` / `navigationBarTranslucent` / `statusBarColor` / `navigationBarColor` completly (not even when `undefined`)

Screenshot 2024-11-26 at 11 09 32

zoontek:
Hi @kkafar, I hate to do this kind of comments, but do you have any news? 😅

kkafar:
I'll be rolling out these changes in either 4.7.0 beta or 4.8.0 beta, depending on few other PRs I want to land & having this change released in separation.

kkafar:
Also note for the future: I've decided on [deprecating all edge-to-edge related props](https://github.com/software-mansion/react-native-screens/pull/2638). They will be removed in future versions of screens.

zoontek:
@kkafar Perfect! I will rebase to resolve the conflict.

zoontek:
@kkafar Rebased! As there were some linting issues, I extracted the logic in a helper function. This way, it will have no extra cost for non-users of edge-to-edge.

kkafar:
I do not remember from the top of my head the exact interaction there. I'll update the PR when doing testing before the release.

kkafar (approved):
I've added `react-native-edge-to-edge` to our example apps & added enforced values for `topInsetEnabled` when `EDGE_TO_EDGE===true`.

Did some sanity manual testing & it looks that we're good. I'll merge once the CI passes. Took a while, I know 😅 

Thanks a lot @zoontek.

PR #2468, opened 2024-10-30
refactor: use side-effect import declaration to ensure TurboModule gets loaded

kkafar:
## Description

We want to ensure that the tubomodule spec gets loaded & executed
and at the same time we don't want it exported. 

This syntax should ensure it! [MDN docs link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#import_a_module_for_its_side_effects_only)

## Changes

Change export to side effect import!

Run FabricExample on Android & see that the underneath screen does not disappear when popping a screen.

## Checklist

- [ ] Ensured that CI passes


alduzy (approved):
LGTM 🚀

PR #2476, opened 2024-11-04
refactor!: use context instead of screenRefs

maciekstosio:
## Description

This PR replaces passing screenRefs into using context. This avoids unnecessary logic on react-navigation side. 

## Changes

Create Context that holds screen refs and add this ref in ScreenStackItem instead of doing same on react-navigation part. See the changes there: https://github.com/react-navigation/react-navigation/pull/12225

## Testing

See Example App > Swipe Back Animation (changes from react-navigation required: https://github.com/react-navigation/react-navigation/pull/12225)
OR
react-navigation: TestScreenAnimation.tsx
native-stack v5: TestScreenAnimationV5.tsx

|native-stack v5|react-navigation|
|-|-|
|<video src="https://github.com/user-attachments/assets/0af1422c-4e61-4c7c-ba6d-1c6a852d4e13" />|<video src="https://github.com/user-attachments/assets/03af7230-198f-4081-ba6d-2b78325ebaa7" />|

## Checklist

- [x] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


src/components/ScreenStackItem.tsx
```diff
@@ -95,7 +120,7 @@ function ScreenStackItem(
 
   return (
     <Screen
-      ref={ref}
+      ref={screenRef}
```
satya164, line 123: Instead of the `useLayoutEffect`, you should be able to assign the ref here using a ref callback:

```suggestion
      ref={(node) => {
        screenRef.current = node;
        screenRefs.current[screenId] = node;
      }}
```

Or

```suggestion
      ref={(node) => {
        screenRef.current = node;
        
        if (node == null) {
          delete screenRefs.current[screenId];
        } else {
          screenRefs.current[screenId] = node;
        }
      }}
```

maciekstosio: I wonder if it's possible that `ref` will be `null`. I like this approach more - it's cleaner, I just wonder if the unmounted views won't stay in `screenRefs`. It shouldn't cause any bugs, but sounds like memory leak.

satya164: when the component is unmounted, the ref callback will be called with `null`.

maciekstosio: Changed in: https://github.com/software-mansion/react-native-screens/pull/2476/commits/8d7fddb9eb79c86032d063c190526daa1f29266a

PR #2477, opened 2024-11-04
feat(iOS)!: change default animation curve & duration

kkafar:
- **Use 0.5 for default transition duration value __but__ keep old animation proportions**
- **Write Push transition using property animator and spring timing curve**
- **Linter**
- **Write Pop transition using property animator and spring timing curve**
- **Implement non-interactive simple push using UIViewPropertyAnimator [PRB]**
- **Implement interactive simple-push using UIViewPropertyAnimator**
- **Make RNSScreenStackAnimator a source of truth on animation properties [PRB]**
- **Update documentation: default transition lasts 500ms**
- **Update default transition duration in Android spec**
- **Implement both interactive and not slide_from_left using UIViewPropertyAnimator**
- **Move comment on default spring animation to helper function**

## Description

WIP

## Changes

WIP

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSPercentDrivenInteractiveTransition.mm
+++ b/ios/RNSPercentDrivenInteractiveTransition.mm
@@ -1,0 +1,69 @@
+#import "RNSPercentDrivenInteractiveTransition.h"
+
+@implementation RNSPercentDrivenInteractiveTransition {
+  RNSScreenStackAnimator *_animationController;
+}
+
+#pragma mark - UIViewControllerInteractiveTransitioning
+
+- (void)startInteractiveTransition:(id<UIViewControllerContextTransitioning>)transitionContext
+{
+  [super startInteractiveTransition:transitionContext];
+}
+
+#pragma mark - UIPercentDrivenInteractiveTransition
+
+// `updateInteractiveTransition`, `finishInteractiveTransition`,
+// `cancelInteractiveTransition` are forwared by superclass to
+// corresponding methods in transition context. In case
+// of "classical CA driven animaitons", such as UIView animation blocks
+// or direct utilization of CoreAnimation API, context drives the animation,
+// however in case of UIViewPropertyAnimator it does not. We need
+// to drive animation manually and this is exactly what happens below.
+
+- (void)updateInteractiveTransition:(CGFloat)percentComplete
+{
+  if (_animationController != nil) {
+    [self.animationController.inFlightAnimator setFractionComplete:percentComplete];
+  }
+  [super updateInteractiveTransition:percentComplete];
+}
+
+- (void)finishInteractiveTransition
+{
+  [self finalizeInteractiveTransitionWithAnimationWasCancelled:NO];
+  [super finishInteractiveTransition];
+}
+
+- (void)cancelInteractiveTransition
+{
+  [self finalizeInteractiveTransitionWithAnimationWasCancelled:YES];
+  [super cancelInteractiveTransition];
+}
+
+#pragma mark - Helpers
+
+- (void)finalizeInteractiveTransitionWithAnimationWasCancelled:(BOOL)cancelled
+{
+  if (_animationController == nil) {
+    return;
+  }
+
+  UIViewPropertyAnimator *_Nullable animator = self.animationController.inFlightAnimator;
+  if (animator == nil) {
+    return;
+  }
+
+  BOOL shouldReverseAnimation = cancelled;
+
+  id<UITimingCurveProvider> timingParams = [[UISpringTimingParameters alloc] initWithDampingRatio:1.0];
+
+  [animator pauseAnimation];
+  [animator setReversed:shouldReverseAnimation];
+  [animator continueAnimationWithTimingParameters:timingParams durationFactor:(1 - animator.fractionComplete)];
+
+  // System retains it & we don't need it anymore.
+  _animationController = nil;
+}
+
+@end
--- a/ios/RNSPercentDrivenInteractiveTransition.mm
+++ b/ios/RNSPercentDrivenInteractiveTransition.mm
@@ -56,7 +56,7 @@
 
   BOOL shouldReverseAnimation = cancelled;
 
-  id<UITimingCurveProvider> timingParams = [[UISpringTimingParameters alloc] initWithDampingRatio:1.0];
+  id<UITimingCurveProvider> timingParams = [_animationController timingParamsForAnimationCompletion];
 
   [animator pauseAnimation];
   [animator setReversed:shouldReverseAnimation];
--- a/ios/RNSScreenStackAnimator.h
+++ b/ios/RNSScreenStackAnimator.h
@@ -2,6 +2,10 @@
 
 @interface RNSScreenStackAnimator : NSObject <UIViewControllerAnimatedTransitioning>
 
+/// This property is filled only when there is an property aniamator associated with an ongoing interactive transition.
+/// In our case this is when we're handling full swipe or edge back gesture.
+@property (nonatomic, strong, nullable, readonly) UIViewPropertyAnimator *inFlightAnimator;
+
 - (nonnull instancetype)initWithOperation:(UINavigationControllerOperation)operation;
 + (BOOL)isCustomAnimation:(RNSScreenStackAnimation)animation;
 
--- a/ios/RNSScreenStackAnimator.h
+++ b/ios/RNSScreenStackAnimator.h
@@ -7,6 +7,15 @@
 @property (nonatomic, strong, nullable, readonly) UIViewPropertyAnimator *inFlightAnimator;
 
 - (nonnull instancetype)initWithOperation:(UINavigationControllerOperation)operation;
+
+/// In case of interactive / interruptible transition (e.g. swipe back gesture) this method should return
+/// timing parameters expected by animator to be used for animation completion (e.g. when user's
+/// gesture had ended).
+///
+/// @return timing curve provider expected to be used for animation completion or nil,
+/// when there is no interactive transition running.
+- (nullable id<UITimingCurveProvider>)timingParamsForAnimationCompletion;
+
 + (BOOL)isCustomAnimation:(RNSScreenStackAnimation)animation;
 
 @end
--- a/ios/RNSScreenStackAnimator.h
+++ b/ios/RNSScreenStackAnimator.h
@@ -2,7 +2,7 @@
 
 @interface RNSScreenStackAnimator : NSObject <UIViewControllerAnimatedTransitioning>
 
-- (instancetype)initWithOperation:(UINavigationControllerOperation)operation;
+- (nonnull instancetype)initWithOperation:(UINavigationControllerOperation)operation;
 + (BOOL)isCustomAnimation:(RNSScreenStackAnimation)animation;
 
 @end
--- a/ios/RNSScreenStackAnimator.mm
+++ b/ios/RNSScreenStackAnimator.mm
@@ -164,10 +164,9 @@ static constexpr float RNSShadowViewMaxAlpha = 0.1;
     // Damping of 1.0 seems close enough and we keep `animationDuration` functional.
     // id<UITimingCurveProvider> timingCurveProvider = [[UISpringTimingParameters alloc] init];
 
-    id<UITimingCurveProvider> timingCurveProvider = [[UISpringTimingParameters alloc] initWithDampingRatio:1.0];
     UIViewPropertyAnimator *animator =
         [[UIViewPropertyAnimator alloc] initWithDuration:[self transitionDuration:transitionContext]
-                                        timingParameters:timingCurveProvider];
+                                        timingParameters:[RNSScreenStackAnimator defaultSpringTimingParametersApprox]];
 
     [animator addAnimations:^{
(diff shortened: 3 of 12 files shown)


hirbod:
Interesting, will this change address the hard animation cancel when the fullScreenGesture swipe to pop was aborted?

kkafar:
@hirbod, could you elaborate a bit? What do you mean exactly by "hard animation cancel"?

ios/RNSScreenStackAnimator.mm
```diff
@@ -442,4 +508,14 @@ - (void)animateTransitionWithStackAnimation:(RNSScreenStackAnimation)animation
...
 }
 
++ (UISpringTimingParameters *)defaultSpringTimingParametersApprox
+{
+  // Default curve provider is as defined below, however spring timing defined this way
+  // ignores the requested duration of the animation, effectively impairing our `animationDuration` prop.
+  // Damping of 1.0 seems close enough and we keep `animationDuration` functional.
+  // id<UITimingCurveProvider> timingCurveProvider = [[UISpringTimingParameters alloc] init];
+
+  return [[UISpringTimingParameters alloc] initWithDampingRatio:1.0];
```
kirillzyusko, line 518: I think this PR is a continuation of https://github.com/software-mansion/react-native-screens/pull/2413

I left a comment about this config: https://github.com/software-mansion/react-native-screens/pull/2413/files#r1829111470 :)

kkafar: I've answered there ;D

hirbod:
Here is a video showing how it looks like with the custom fullScreenGestureEnabled cancel

https://github.com/user-attachments/assets/fcdd92cc-899e-4e7f-9b88-da5529fb89c8

If you compare this with the default one (which is smoothed), this looks pretty bad.

kkafar:
Then yes - these changes will impact how the "completion animation after gesture release" looks. It'll use spring curve (with very high damping) instead of straight up linear curve.

hirbod:
Yes, I think having this for simple_push is good enough. Damping for fade doesn't seem necessary

hirbod:
The before/after video for simple_push appears to break the header. The back button transition looks terrible in the after video. :/

MaxAst:
I have to second @hirbod, the navigation header transition looks quite abrupt, imo worse than the current version

ios/RNSScreenStackAnimator.mm
```diff
@@ -3,30 +3,48 @@
...
+  UIViewPropertyAnimator *_Nullable _inFlightAnimator;
 }
 
 - (instancetype)initWithOperation:(UINavigationControllerOperation)operation
 {
   if (self = [super init]) {
     _operation = operation;
-    _transitionDuration = 0.35; // default duration in seconds
+    _transitionDuration = RNSDefaultTransitionDuration; // default duration in seconds
+    _inFlightAnimator = nil;
```
WoLewicki, line 41: Is it necessary to set it to `nil` in constructor?

ios/RNSPercentDrivenInteractiveTransition.mm
```diff
@@ -0,0 +1,69 @@
...
+// or direct utilization of CoreAnimation API, context drives the animation,
+// however in case of UIViewPropertyAnimator it does not. We need
+// to drive animation manually and this is exactly what happens below.
+
+- (void)updateInteractiveTransition:(CGFloat)percentComplete
+{
+  if (_animationController != nil) {
+    [self.animationController.inFlightAnimator setFractionComplete:percentComplete];
+  }
+  [super updateInteractiveTransition:percentComplete];
```
WoLewicki, line 29: Not sure why header looks so bad in those transitions, I think this line should control the state of animation of header too. Are those values not changing pretty linearly ?

WoLewicki (commented):
Left some comments! Nice to see progress in that thing!

ios/RNSPercentDrivenInteractiveTransition.mm
```diff
@@ -0,0 +1,69 @@
...
+{
+  [super startInteractiveTransition:transitionContext];
+}
+
+#pragma mark - UIPercentDrivenInteractiveTransition
+
+// `updateInteractiveTransition`, `finishInteractiveTransition`,
+// `cancelInteractiveTransition` are forwared by superclass to
+// corresponding methods in transition context. In case
+// of "classical CA driven animaitons", such as UIView animation blocks
```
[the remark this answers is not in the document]
kkafar: I believe `classical` is the right word here 😄 Typos fixed

ios/RNSScreenStackAnimator.h
```diff
@@ -2,7 +2,20 @@
 
 @interface RNSScreenStackAnimator : NSObject <UIViewControllerAnimatedTransitioning>
 
-- (instancetype)initWithOperation:(UINavigationControllerOperation)operation;
+/// This property is filled only when there is an property aniamator associated with an ongoing interactive transition.
+/// In our case this is when we're handling full swipe or edge back gesture.
```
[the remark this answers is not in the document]
kkafar: This is a doc-style comment in objective-c (XCode?), same as `/** */` Idk your keybindings, but when using vim plugin, in normal mode you can press `shift + k` to display the docs for given symbol. I'm sure w/o Vim there is some other key combination to trigger this.

ios/RNSScreenStackAnimator.mm
kkafar: I guess it is not required, however I want it to be explicit.

ios/RNSPercentDrivenInteractiveTransition.mm
kkafar: Header looks bad only in non-interactive transitions. TBH I'm surprised that it does not look bad in old implementation - the [docs suggest that it should look bad](https://github.com/software-mansion/react-native-screens/blob/960873aa03b9e22136d2f7b3b272e407011568d2/src/types.tsx#L395)

kkafar: I do not know what code drives the header transition in non-interactive transitions (notice that I'm not swiping on these recordings). Need to spend some time on this.

kkafar:
Yeah, I agree. I'll give my self an hour to figure this out, as I need to release 4.0.0. If I don't make it on time I'll release it with next minor behind some kind of feature flag.

ios/RNSPercentDrivenInteractiveTransition.mm
WoLewicki: The docs mention the native header transition where the back button title travels to the place of title when navigating back/forward. It works only with the default animations. For `simple_push` it was the fade animation in header iirc.

ios/RNSPercentDrivenInteractiveTransition.mm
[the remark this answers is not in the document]
kkafar: https://github.com/software-mansion/react-native-screens/pull/2477/commits/7b0bdca9d3ff1ab21b7439c695092651ccb209b1

ios/RNSPercentDrivenInteractiveTransition.mm
kkafar: I guess you're right. I've just confirmed that it has not been a fluke - on main there is nice fade transition. Do you happen to remember where this is animated? Or is this handled entirely by UIKit?

kkafar: Figured this out in: https://github.com/software-mansion/react-native-screens/pull/2477/commits/7a2dfc948eebc056700e9281d496cfc215a0a174

kkafar:
@WoLewicki @hirbod @kirillzyusko @MaxAst I've fixed the header animation. Take a look when you have a moment. I've updated the "after" videos. 

I'll proceed with merging these changes, as I want to include these in 4.0.0 which will be released any minute from now. If we'll want to make some minor adjustments I'll release them as fixes in next minor.

kirillzyusko:
Great job @kkafar 💪 

I tested the example app and I can not spot any issues: header works fine, animation feels native, navigation after gesture also feels like a native, transition duration is still customizable 🔥 Awesome stuff! And congrats with `4.0` release!

kkafar:
Glad to hear that!

> [...] congrats with 4.0 release!

Thanks!

hirbod:
Thanks for the updates—it’s definitely better, but the header is still not quite right and differs from the native stack crossfade. It feels delayed and it’s using a spring animation, but the header should animate linearly to avoid harsh cross fades. (Cross fades are always linear)

Also, the cancel animation is too long. It should maybe be 50% of the animation duration, or it should have its own prop. I also believe that the distance is important (a short swipe followed by cancel shouldn’t take 2000ms). Both the header and cancel need adjustments. In my opinion, this PR was rushed to get merged, considering it’s a significant change that affects many apps.

kirillzyusko:
> Also, the cancel animation is too long. It should maybe be 50% of the animation duration, or it should have its own prop. I also believe that the distance is important (a short swipe followed by cancel shouldn’t take 2000ms).

@hirbod I was comparing cancel animation side by side with iOS Settings app and in my opinion the native app uses the same transition as native-stack does now (but I used default 500ms transition). Of course, if you specify 2000ms it feels slow, but then plain navigation also works slow (and in this case it's kind of consistency?).

> It feels delayed and uses a spring animation, but the header should animate linearly to avoid harsh crossfades. (Crossfades are always linear, and it becomes very noticeable when going back, as the back button hides abruptly.)

Probably you are right here - header animation indeed feels delayed, but i think it can be fixed in subsequent releases 👀 

> In my opinion, this PR was rushed to get merged, considering it’s a significant change that affects many apps.

I think the main reason to include this PR was because SWM team had to release RNS for `react-navigation@v7`. And this PR brings breaking changes, so it was very reasonable to include it in 4.0 (maybe it does not have an ideal state of things, but it can be always fixed later - main thing here is to highlight breaking changes into a single release).

hirbod:
I tested the change, and indeed, the cancel animation works fine as long as the animationDuration hasn’t been modified. I’m not sure if canceling should take that long when animationDuration is set to a high value—it would be consistent, but I think having an option to override this would be beneficial.

The header is definitely off. After testing it more thoroughly myself, I really don’t like it; it stands out quite strongly for me. I understand the reasoning for shipping this with a major (and breaking) release, but I was hoping for more community testing and feedback before releasing it, as SDK 52 will ship with Expo Router v4, which will use v7 and RNS 4. This means the impact will be larger, as people are likely to upgrade without even knowing.

That said, it’s still a fresh release, and a subsequent update could quickly address the header issue. I really appreciate both your work @kirillzyusko and @kkafar

kkafar:
> [...] the header still isn't quite right and differs from the native stack crossfade. It feels delayed and uses a spring animation, but the header should animate linearly to avoid harsh crossfades. (Crossfades are always linear, and it becomes very noticeable when going back, as the back button hides abruptly.)

& 

> Probably you are right here - header animation indeed feels delayed, but i think it can be fixed in subsequent releases 👀

Can agree here, but I'm not sure what should be the direction. Before this merge it has looked as follows:

https://github.com/user-attachments/assets/935386e1-aa2c-44d7-b702-208816c4630e

That is, in case of:

* non-interactive transition we have indeed the crossfade
* interactive transition the being-popped navbar item first fades out completely, and just then the entering one starts fading in (I guess that's "sequenced fade"). This results in few frames when the navigation bar is completely blank I don't like this personally, but I'm curious of your opinion.

We could:

1. restore old behaviour here (I see one technical problem, but might be able to workaround it, will turn out when I'll start coding),
2. make it always linearly crossfade (I think that's better option)


> [...]  I’m not sure if canceling should take that long when animationDuration is set to a high value—it would be consistent, but I think having an option to override this would be beneficial. 

Yep, I've highlighted this effect in `important` section in PR description & suggested that it might be reasonable to allow to control this also in PR description. I will expose a flag to control whether animation duration should impact the "completion animation". 


Thanks for your input

hirbod:
I believe make it crossfade linearly _could_ be enough, but would appreciate a recording.

kkafar:
|Linear|EaseInOut|DefaultSpring|
|--------|-------|------|
|<video width="454" alt="image" src="https://github.com/user-attachments/assets/12a4479e-6c63-4498-b7ae-0d011f57f136">|<video width="454" alt="image" src="https://github.com/user-attachments/assets/dab0bcf2-4a09-42f1-97a9-c96e4b5aead2">|<video width="454" alt="image" src="https://github.com/user-attachments/assets/2122104c-2142-42af-83be-7b310f4e91f5">|


wdyt?

kirillzyusko:
I like ease-in-ease-out - I think it was a default transition before?

Correct me if I'm wrong @kkafar 🙈

hirbod:
Linear is the winner for me.

hirbod:
While dragging, all look fine but I'd say Linear is best. But when you press the back button, it feels like it is kind of delays and "flashes", before it cross fades. Something is off, but it's not too bad. Almost there

kkafar:
I've created #2492 improving the header animation (ease-in-out). 

> [...] it feels like it is kind of delays and "flashes", before it cross fades [...]

☝🏻 also explaining this and requesting opinions

PR #2479, opened 2024-11-05
chore: compatibility table README update

alduzy:
## Description

This PR updates the compatibility table in `README` with support for React Native 0.76.

I updated the table and added a brief description for Paper architecture since it's not the default for RN 0.76+.

## Tests

I ran the following tests in order to see if the compatibility changed since the last check.

### Paper

- RN `0.72` builds and run with screens `3.35.0` ✅
    - iOS ✅
    - Android ✅
- RN `0.76` builds and run with screens `3.33.0` ✅
    - iOS ✅
    - Android ✅

### Fabric

- RN `0.75` builds and run with screens `3.35.0` ✅
    - iOS ✅
    - Android ✅
- RN `0.76` builds and run with screens `3.33.0` ❌ 
    - iOS ✅
    - Android ❌

## Checklist

- [x] Updated documentation: 

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,9 +106,13 @@ Screens are already integrated with the React Native's most popular navigation l
 
 ## Supported react-native version
 
+### Support for Paper
+
+Paper is the default rendering system for React Native versions prior to 0.76.
+
 | library version | react-native version |
 | --------------- | -------------------- |
-| 3.33.0+         | 0.72.0+
+| 3.33.0+         | 0.72.0+              |
 | 3.32.0+         | 0.71.0+              |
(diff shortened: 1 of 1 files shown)


kkafar (approved):
The changes look ok ✅ 

But just a note so that we're on the same page: the supported versions table is not about what versions work with `react-native-screens`, it is about what versions of react-native we support (fixing issues, etc.).

alduzy:
Ok I see @kkafar. Should we add the new row (since it's not react-native-screens that require RN 0.76 but the opposite) then?

PR #2481, opened 2024-11-06
feat(iOS)!: change default of `fullScreenSwipeShadowEnabled` to true

kkafar:
## Description

This PR changes default value for `fullScreenSwipeShadowEnabled` prop to `true` - meaning that the shadow will be present by default
in `simple_push` animation on iOS (this is the platform default).

> [!important]
> When changing this value I've noticed that the `fullScreenSwipeShadowEnabled` does something different
> than it suggests in type documentation & our GUIDE_FOR_LIBRARY_AUTHORS. Namely, it does effect `simple_push` animation
> in every scenario, not only in full screen swipe gesture interactive transition. I think we should change the name
> of the prop eventually. 

## Changes

Updated all the docs & code.

I've played around with `Test1072`. Just set stack animation to simple push.

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [ ] Ensured that CI passes


alduzy (commented):
The changes look good, but your note is concerning. 
We could indeed distinguish the prop name from fullScreenSwipe since it's not that closely related. Even `shadowEnabled` with current description would be less confusing in my opinion. It may be a good time for such change

PR #2483, opened 2024-11-06
chore: workflow for updating Podfile.lock files

alduzy:
## Description

This PR adds a workflow that updates `Podfile.lock` files in the example apps automatically based on dependency changes.

In the current form the update is triggered by any change inside the `package.json` file in the root or in the corresponding example app.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


kkafar:
We won't proceed here right now. Will reopen if we decide to do so.

PR #2485, opened 2024-11-07
fix: remove unnecessary checks for creating snapshot

WoLewicki:
## Description

Based on https://github.com/Expensify/App/pull/49937#issuecomment-2451993625 and the comments above, PR fixes the unnecessary checks for creating snapshot on the view on dismiss. It was refactored already in https://github.com/software-mansion/react-native-screens/pull/2134 and https://github.com/software-mansion/react-native-screens/pull/2261 (👏  to @kkafar). Those checks do nothing now since each screen is responsible for making its own snapshot. Having those checks can only lead to problems.


ios/RNSScreenStack.mm
```diff
@@ -1176,16 +1176,7 @@ - (nullable RNSScreenView *)childScreenForTag:(react::Tag)tag
...
-  // We should only do a snapshot of a screen that is on the top.
-  // We also check `_presentedModals` since if you push 2 modals, second one is not a "child" of _controller.
-  // Also, when dissmised with a gesture, the screen already is not under the window, so we don't need to apply
-  // snapshot.
-  if (screenChildComponent.window != nil &&
-      ((screenChildComponent == _controller.visibleViewController.view && _presentedModals.count < 2) ||
-       screenChildComponent == [_presentedModals.lastObject view])) {
-    [screenChildComponent.controller setViewToSnapshot];
-  }
+  [screenChildComponent.controller setViewToSnapshot];
```
kkafar, line 1179: I see that this simplifies the logic, but I don't think this is quite right. 

Imagine you have Stack with three screens: "A, B, C". You pop from C to top, thus resulting stack is "A". This code would do snapshots of both C and B, while we only want to do a snapshot of C.

If you describe what problem exactly you intend to solve we might be able to figure out better solution.

kkafar (changes requested):
I think we need to work on improving this code 👇🏻

ios/RNSScreenStack.mm
WoLewicki: Will this code: https://github.com/software-mansion/react-native-screens/blob/46147026c708eab19de8082a2d2a1c7fd680b30c/ios/RNSScreen.mm#L1842 return `true` for `B`? I think its view is not in the window since it is detached. Would have to check though.

WoLewicki: I checked. It is `nil` then so it should work correctly.

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -847,27 +847,29 @@ - (void)unmountChildComponentView:(UIView<RCTComponentViewProtocol> *)childCompo
...
-      break;
-    case RNSScreenStackHeaderSubviewTypeRight:
-      navitem.rightBarButtonItem.customView = snapshot;
-      break;
-    case RNSScreenStackHeaderSubviewTypeSearchBar:
-    case RNSScreenStackHeaderSubviewTypeBackButton:
-      break;
-    default:
-      RCTLogError(@"[RNScreens] Unhandled subview type: %ld", childComponentView.type);
+  if (childComponentView.window != nil) {
```
WoLewicki, line 850: Also added this check since we got many warning of snapshotting a view that is not in hierarchy when dismissing many screens with header items at once.

ios/RNSScreenStack.mm
kkafar: +1, I've forgotten that UINavigationController will keep all the controllers lower in the stack in detached state and that's it why it will work correctly.

kkafar (approved):
We're good then. Thanks for answering my concerns.

PR #2487, opened 2024-11-08
fix: get imageLoader from contextContainer

WoLewicki:
## Description

PR getting rid of the last occurrence of `currentBridge`. As suggested by Meta team, we get the needed `RCTImageLoader` from `contextContainer`, which is available in `componentDescriptor`.

### DIFF
--- a/common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderSubviewComponentDescriptor.h
+++ b/common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderSubviewComponentDescriptor.h
@@ -2,7 +2,7 @@
 
 #ifdef ANDROID
 #include <fbjni/fbjni.h>
-#endif
+#endif // ANDROID
 #include <react/debug/react_native_assert.h>
 #include <react/renderer/components/rnscreens/Props.h>
 #include <react/renderer/components/rnscreens/utils/RectUtil.h>
@@ -21,6 +21,15 @@ class RNSScreenStackHeaderSubviewComponentDescriptor final
 
   void adopt(ShadowNode &shadowNode) const override {
     ConcreteComponentDescriptor::adopt(shadowNode);
+#ifndef NDEBUG
+    std::weak_ptr<void> imageLoader =
+        contextContainer_->at<std::shared_ptr<void>>("RCTImageLoader");
+    react_native_assert(
+        dynamic_cast<RNSScreenStackHeaderSubviewShadowNode *>(&shadowNode));
+    auto &subviewShadowNode =
+        static_cast<RNSScreenStackHeaderSubviewShadowNode &>(shadowNode);
+    subviewShadowNode.setImageLoader(imageLoader);
+#endif // NDEBUG
   }
 };
 
--- a/common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderSubviewShadowNode.cpp
+++ b/common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderSubviewShadowNode.cpp
@@ -5,4 +5,21 @@ namespace facebook::react {
 extern const char RNSScreenStackHeaderSubviewComponentName[] =
     "RNSScreenStackHeaderSubview";
 
+#ifdef ANDROID
+#else // ANDROID
+#ifndef NDEBUG
+void RNSScreenStackHeaderSubviewShadowNode::setImageLoader(
+    std::weak_ptr<void> imageLoader) {
+  getStateDataMutable().setImageLoader(imageLoader);
 }
+RNSScreenStackHeaderSubviewShadowNode::StateData &
+RNSScreenStackHeaderSubviewShadowNode::getStateDataMutable() {
+  // We assume that this method is called to mutate the data, so we ensure
+  // we're unsealed.
+  ensureUnsealed();
+  return const_cast<RNSScreenStackHeaderSubviewShadowNode::StateData &>(
+      getStateData());
+}
+#endif // NDEBUG
+#endif // ANDROID
+} // namespace facebook::react
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -378,8 +378,12 @@ namespace react = facebook::react;
         // in DEV MODE we try to load from cache (we use private API for that as it is not exposed
         // publically in headers).
         RCTImageSource *imageSource = [RNSScreenStackHeaderConfig imageSourceFromImageView:imageView];
-        RCTImageLoader *imageLoader = [subview.bridge moduleForClass:[RCTImageLoader class]];
-
+        RCTImageLoader *imageLoader =
+#ifdef RCT_NEW_ARCH_ENABLED
+            [subview imageLoader];
+#else
+            [subview.bridge moduleForClass:[RCTImageLoader class]];
+#endif
         image = [imageLoader.imageCache
             imageForUrl:imageSource.request.URL.absoluteString
                    size:imageSource.size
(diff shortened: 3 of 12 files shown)


ios/RNSScreenStackHeaderSubview.mm
```diff
@@ -129,17 +143,11 @@ - (void)reactSetFrame:(CGRect)frame
...
-#endif // RCT_NEW_ARCH_ENABLED
-
 - (RCTBridge *)bridge
 {
-#ifdef RCT_NEW_ARCH_ENABLED
-  return [RCTBridge currentBridge];
-#else
   return _bridge;
-#endif // RCT_NEW_ARCH_ENABLED
 }
```
j-piasecki, line 149: Is this still necessary when there's no custom logic?

ios/RNSScreenStackHeaderSubview.mm
```diff
@@ -17,11 +18,13 @@
...
-@interface RCTBridge (Private)
-+ (RCTBridge *)currentBridge;
-@end
-
 @implementation RNSScreenStackHeaderSubview
+#ifdef RCT_NEW_ARCH_ENABLED
+{
+  react::RNSScreenStackHeaderSubviewShadowNode::ConcreteState::Shared _state;
+  RCTImageLoader *imgLoader;
+}
```
j-piasecki, line 26: You gated almost every part of logic related to it with `NDEBUG` but left a few parts here. Why?

WoLewicki:
tagging @cipolleschi

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -378,8 +378,12 @@ + (UIImage *)loadBackButtonImageInViewController:(UIViewController *)vc withConf
         // in DEV MODE we try to load from cache (we use private API for that as it is not exposed
         // publically in headers).
         RCTImageSource *imageSource = [RNSScreenStackHeaderConfig imageSourceFromImageView:imageView];
-        RCTImageLoader *imageLoader = [subview.bridge moduleForClass:[RCTImageLoader class]];
-
+        RCTImageLoader *imageLoader =
```
cipolleschi, line 381: it's a bit weird that the config extracts the imageLoader from the subview instead of receiving the `contextContainer` in the init. It's ok for now, if it works, but this approach is not super clean. I expect that the config are created before the view is created, no?

cipolleschi (commented):
I left a comment but the code looks good overall.
The navigation submodule has some changes, should I review those as well?

common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderSubviewShadowNode.cpp
```diff
@@ -5,4 +5,21 @@ namespace facebook::react {
 extern const char RNSScreenStackHeaderSubviewComponentName[] =
     "RNSScreenStackHeaderSubview";
 
+#ifdef ANDROID
+#else // ANDROID
```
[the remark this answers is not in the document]
kkafar: +1

common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderSubviewComponentDescriptor.h
```diff
@@ -21,6 +21,15 @@ class RNSScreenStackHeaderSubviewComponentDescriptor final
...
     ConcreteComponentDescriptor::adopt(shadowNode);
+#ifndef NDEBUG
+    std::weak_ptr<void> imageLoader =
+        contextContainer_->at<std::shared_ptr<void>>("RCTImageLoader");
+    react_native_assert(
+        dynamic_cast<RNSScreenStackHeaderSubviewShadowNode *>(&shadowNode));
+    auto &subviewShadowNode =
+        static_cast<RNSScreenStackHeaderSubviewShadowNode &>(shadowNode);
+    subviewShadowNode.setImageLoader(imageLoader);
+#endif // NDEBUG
```
kkafar, line 32:
```suggestion
#endif // !NDEBUG
```

ios/RNSScreenStackHeaderSubview.mm
kkafar: +1

ios/RNSScreenStackHeaderSubview.mm
```diff
@@ -17,11 +18,13 @@
...
 
-@interface RCTBridge (Private)
-+ (RCTBridge *)currentBridge;
-@end
-
 @implementation RNSScreenStackHeaderSubview
+#ifdef RCT_NEW_ARCH_ENABLED
+{
+  react::RNSScreenStackHeaderSubviewShadowNode::ConcreteState::Shared _state;
+  RCTImageLoader *imgLoader;
```
kkafar, line 25:
```suggestion
  RCTImageLoader *_imgLoader;
```

ios/RNSScreenStackHeaderSubview.mm
```diff
@@ -68,9 +71,9 @@ - (instancetype)initWithFrame:(CGRect)frame
 
 #pragma mark - RCTComponentViewProtocol
 
-- (void)prepareForRecycle
+- (RCTImageLoader *)imageLoader
 {
-  [super prepareForRecycle];
+  return imgLoader;
 }
```
kkafar, line 77: Naaa, even if the recycling is disabled, please do not mix up all the changes in this single PR. Podfile locks, this. Let's clean this up before merging.

react-navigation
kkafar, line 1: This **must** be reverted.

ios/RNSScreenStackHeaderConfig.mm
kkafar: I don't think there is a reason. Most likely, because the we had access to bridge already there and someone just used it. 

Let's init header config with bridge and use it on Paper. 

Same thing applies to holding the weak pointer to the module in subview shadow node. We should keep it in header config's state.

kkafar (changes requested):
👇🏻

ios/RNSScreenStackHeaderSubview.mm
WoLewicki: I needed them since the code with trying to get `imgLoader` was not explicitly surrounded with `NDEBUG` macro till now.  I changed it.

common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderSubviewShadowNode.cpp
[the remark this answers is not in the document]
WoLewicki: I did it since in `ifndef` you can easily miss that there is `n` in there leading to confusion. But if you don't care, I'll change it.

ios/RNSScreenStackHeaderConfig.mm
kkafar: I was not around back then 😅 

My (educated) guess would be that `loadBackButtonImageInViewController` was class method in order to match the pattern in header config - most of other methods are class-bounded as well. I seriously doubt that migrating here to instance method will give us any hiccup. We should be fine on this front.

common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderConfigShadowNode.h
```diff
@@ -27,6 +27,20 @@ class JSI_EXPORT RNSScreenStackHeaderConfigShadowNode final
 #pragma mark - ShadowNode overrides
 
 #pragma mark - Custom interface
+#ifdef ANDROID
+#else // ANDROID
+#ifndef NDEBUG
+  void setImageLoader(std::weak_ptr<void> imageLoader);
+#endif // NDEBUG
+#endif // ANDROID
```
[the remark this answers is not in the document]
kkafar: There are several places with this issue

react-navigation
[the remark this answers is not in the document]
kkafar: I guess you might be missing refspecs from remote on you local brach. Try out running `git fetch origin main` in `react-navigation` submodule.

kkafar (approved):
We should be good.

PR #2494, opened 2024-11-12
fix: Remove dangling global reference to `ScreenModule`

lukmccall:
## Description

Removes a dangling global reference to the `ScreenModule`.

## Changes

When the application reloads, `nativeInstall` is called, which overwrites the reference to `globalThis` without deleting it. This PR deletes the reference before creating a new one.

I've tested it in a new project created using expo (SDK 52) and with fabric turn on.


kkafar (approved):
Makes sense, thanks!

PR #2495, opened 2024-11-12
fix(Android): going back on fabric with removeClippedSubviews

### LINKED ISSUE #2491: Cannot remove child at index X from parent ViewGroup X, only X children in parent. Warning: childCount may be incorrect!
### Description

[Android Only] Application crashes when unmounting a screen with `removeClippedSubviews` set to `true`.

### Steps to reproduce

Run application from reproduction repository (`rnscreens` branch).
Press login, then logout, see the application crash:

[rnscreens_demo.webm](https://github.com/user-attachments/assets/dcc3b3ba-bb06-460b-a3ec-ce9722b77e37)

https://github.com/maciekBudzinski/rnp-material-bottom-tabs-repro-issue/tree/rnscreens

### Screens version

4.0.0

### React Native version

0.76.1

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

Android emulator

### Device model

_No response_

### Acknowledgements

Yes

alduzy:
## Description

This PR intents to fix going back on fabric issue when using a `View` with [removeClippedSubviews](https://reactnative.dev/docs/view#removeclippedsubviews) prop set to true.

Previous bug fixes addressed this issue primarily for FlatLists, where it's set to true by default on Android. See https://github.com/software-mansion/react-native-screens/pull/2383.

Additionally, this PR greatly improves the performance of `startTransitionRecursive` as it does not climb up the tree in search for a parent with `removeClippedSubview` set to true anymore.

Fixes #2491 .

## Changes

- removed redundant code
- updated `Test2282.tsx` repro

- use `Test2282.tsx` repro

## Checklist

- [x] Ensured that CI passes


android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -378,7 +380,10 @@ class Screen(
         }
     }
 
-    private fun startTransitionRecursive(parent: ViewGroup?) {
+    private fun startTransitionRecursive(
+        parent: ViewGroup?,
+        isPossiblyRemovedClippedSubview: Boolean = true,
```
WoLewicki, line 385: Why is the default `true`? The only time we call it without this argument is at the start in `startRemovalTransition` yes? If so, maybe instead of giving the default, just use the proper value there too?

kkafar: +1

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -396,19 +401,19 @@ class Screen(
...
+                    // set to true (which is the default for FlatLists).
+                    // We add a simple view for each possibly clipped item to make it work as expected.
+                    // See https://github.com/software-mansion/react-native-screens/pull/2495
+                    if (isPossiblyRemovedClippedSubview && child !is ReactScrollView && child !is ReactHorizontalScrollView) {
                         for (j in 0 until child.childCount) {
                             child.addView(View(context))
                         }
                     }
-                    startTransitionRecursive(child)
+                    startTransitionRecursive(child, isPossiblyRemovedClippedSubview || it.removeClippedSubviews)
```
kkafar, line 416: same here

android/src/main/java/com/swmansion/rnscreens/Screen.kt
alduzy: @WoLewicki @kkafar you're absolutely right, it should be false by default or don't have a default value at all.
Eventually this mistake got me thinking and it looks like assuming it's true is a safe option here. Thus we can get rid of the redundant logic.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -396,19 +401,19 @@ class Screen(
                 if (child is ScreenStackHeaderConfig) {
                     // we want to start transition on children of the toolbar too,
                     // which is not a child of ScreenStackHeaderConfig
-                    startTransitionRecursive(child.toolbar)
+                    startTransitionRecursive(child, isPossiblyRemovedClippedSubview || it.removeClippedSubviews)
```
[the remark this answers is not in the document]
alduzy: Removed in https://github.com/software-mansion/react-native-screens/pull/2495/commits/b4416e22d2a3caecb182f101354582636884e7c3

kkafar (commented):
I have not tested the runtime. Let me know that you've tested this (some casual navigation, flatlists, etc.) or that you need me to test this. When you're ready, please mark the PR as open.

Right now the code changes look good, my only concern is runtime behaviour, as I do not understand the issue at hand (leaving it up to you).

kkafar (approved):
The code is looking good. If you tested and the logic makes sense, then I think we should proceed.

kkafar:
Please do not merge it yet. I just thought of something, but I'll be able to test my hypothesis tomorrow / early next week.

kkafar:
I've debugged this for a while now & I have good understanding what's going on. This bug is caused by our usage of `startViewTransition` and its implications. We use it well, however React does not account for case that some view might be in transition. Error mechanism is as follows: 

1. Let's have initially simple stack with two screens: "A, B". This is component rendered under "B":

```javascript
    <View <-- ContainerView (CV)
      removeClippedSubviews
      style={{ flex: 1, backgroundColor: 'slateblue', overflow: 'hidden' }}>
      <View removeClippedSubviews style={{ height: '100%' }}> <--- IntermediateView (IV)
        <View removeClippedSubviews style={{ backgroundColor: 'pink', width: '100%', height: 50 }} /> <--- ChildView (ChV)
      </View>
    </View>
```

2. We press the back button.
3. We're on Fabric, therefore subtree of B gets destroyed before B itself is unmounted -> in our commit hook we detect that the screen B will be unmounted & we mark every node under B as transitioning by calling `startViewTransition`.
4. React Mounting stage starts, view hierarchy is disassembled in bottom-up fashion (leafs first).
5. ReactViewGroupManager receives MountItem to detach ChV from IV.
6. A call to [`IV.removeView(ChV)` is made](https://github.com/facebook/react-native/blob/9c11d7ca68c5c62ab7bab9919161d8417e96b28b/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactClippingViewManager.kt#L58-L73), which effectively removes ChV from `IV.children`, ***HOWEVER*** it does not clear `ChV.parent`, meaning that after the call, `ChV.parent == IV`. This happens, due to view being marked as in-transition by our call to `startViewTransition`. If the view is not marked as in-transition this parent-child relationship is removed.
7. IV has `removeClippedSubviews` enabled, therefore a [call to `IV.removeViewWithSubviewsClippingEnabled(ChV)` is made](https://github.com/facebook/react-native/blob/9c11d7ca68c5c62ab7bab9919161d8417e96b28b/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactClippingViewManager.kt#L68). [This function](https://github.com/facebook/react-native/blob/9c11d7ca68c5c62ab7bab9919161d8417e96b28b/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewGroup.java#L726-L744) does effectively two things:
    1. if the ChV has parent (interpretation: it has not yet been detached from parent), we compute it's index in `IV.children` (Android.ViewGroup's state) and remove it from the array,
    2. remove the ChV from `mAllChildren` array (this is state maintained by ReactViewGroup for purposes of implementing the "subview clipping" mechanism".

The crash happens in 7.1, because ChV has been removed from `IV.children` in step 6, but the parent-child relationship has not been broken up there. Under usual circumstances (this is my hypothesis now, yet unconfirmed) 7.1 does not execute, because `ChV.parent` is nulled in step no. 6.

Possible solutions: WIP

kkafar:
Related PR in react-native repo: 

* https://github.com/facebook/react-native/pull/47634

kkafar (approved):
I think we should proceed with merging to salvage the situation before (and if) the fix for this lands in core.

PR #2501, opened 2024-11-13
feat: add support for transparent large header on iOS

EvanBacon:
## Description

- requires this bug fix https://github.com/software-mansion/react-native-screens/pull/2500
- make it possible to support the stock Apple behavior where the large header is transparent and the collapsed header is blurry with a shadow. This is currently not possible because we inherit the blur effect from the collapsed title, and because there's no way to set the headerLargeTitle blur effect.

With this change, the default iOS behavior can more closely be matched by using the following props:

```js
{
  headerLargeTitle: true,
  headerTransparent: true,
  headerBlurEffect: "systemChromeMaterial",
  headerLargeTitleShadowVisible: false,
  headerShadowVisible: true,
  headerLargeStyle: {
    // NEW: Make the large title transparent to match the background.
    backgroundColor: "transparent",
  },
}
```

Without the shadow changes, users also need to add this (and adjust for dark mode):

```
{
  headerStyle: {
    // Hack to ensure the collapsed small header shows the shadow / border.
    backgroundColor: "rgba(255,255,255,0.01)",
  },
}
```

### Before

This is the best we can currently do.

IMG_5976

### After

https://github.com/user-attachments/assets/b1e8999d-14ce-40ef-aef6-2478a9e6326c

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -671,6 +671,17 @@ RNS_IGNORE_SUPER_CALL_END
     UINavigationBarAppearance *scrollEdgeAppearance =
         [[UINavigationBarAppearance alloc] initWithBarAppearance:appearance];
     if (config.largeTitleBackgroundColor != nil) {
+      // Add support for using a fully transparent bar when the backgroundColor is set to transparent. 
+      if (CGColorGetAlpha(config.largeTitleBackgroundColor.CGColor) == 0.) {
+      // This will also remove the background blur effect in the large title which is otherwise inherited from the standard appearance.
+        [scrollEdgeAppearance configureWithTransparentBackground];
+        // This must be set to nil otherwise a default view will be added to the navigation bar background with an opaque background.
+        scrollEdgeAppearance.backgroundColor = nil;
+      } else {
+        scrollEdgeAppearance.backgroundColor = config.largeTitleBackgroundColor;
+      }
+    }
+    if (config.largeTitleBackgroundColor != nil) {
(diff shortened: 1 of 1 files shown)


ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -670,6 +678,17 @@ + (void)updateViewController:(UIViewController *)vc
...
+        [scrollEdgeAppearance configureWithTransparentBackground];
+        // This must be set to nil otherwise a default view will be added to the navigation bar background with an opaque background.
+        scrollEdgeAppearance.backgroundColor = nil;
+      } else {
+        scrollEdgeAppearance.backgroundColor = config.largeTitleBackgroundColor;
+      }
+    }
     if (config.largeTitleBackgroundColor != nil) {
       scrollEdgeAppearance.backgroundColor = config.largeTitleBackgroundColor;
     }
```
kkafar, line 694: It looks like something is off here. You now have two "if" statements checking exactly for the same condition `if (config.largeTitleBackgroundColor != nil)`, while in first one you set `scrolleEdgeAppearance.backgroundColor` to nonnull value only when alpha is not 0, and then you override the background color no matter the alpha. Are you sure this works as you expected?

kkafar (changes requested):
Thanks for the changes!

I have one question that needs to be answered before we can proceed.

ios/RNSScreenStackHeaderConfig.mm
EvanBacon, line 694:
```suggestion
```

kkafar (approved):
This version makes more sense, thanks!

PR #2527, opened 2024-11-20
fix: do not add views on transition to ViewPager2

radko93:
## Description

There is an issue in `react-native-pager-view` that happens when you unmount a screen with ViewPager2 in it. As you cannot add views to `ViewPager2` directly. This change omits it. I can reproduce it on paper but Fabric seems also to be affected.

Fixes https://github.com/callstack/react-native-pager-view/issues/906.
## Changes

- Fixes crash when transitioning with ViewPager2 mounted

The repro is linked in the original issue.
## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -405,7 +405,7 @@ class Screen(
                     // Unless the child is a ScrollView it's safe to assume that it's true
                     // and add a simple view for each possibly clipped item to make it work as expected.
                     // See https://github.com/software-mansion/react-native-screens/pull/2495
-                    if (child !is ReactScrollView && child !is ReactHorizontalScrollView) {
+                    if (child !is ReactScrollView && child !is ReactHorizontalScrollView && child !is ViewPager2) {
                         for (j in 0 until child.childCount) {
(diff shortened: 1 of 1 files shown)


radko93:
I also had an issue with expo-video crashing in the same scenario.
telegram-cloud-photo-size-4-5904763039482234137-y

kkafar:
Hey, while this is good patch for your particular case we won't go this way. 
I've opened a PR to core in order to fix the real issue cause there: https://github.com/facebook/react-native/pull/47634.

Here I'll patch this in a bit different manner

radko93:
@kkafar thanks for making the effort for fixing the cause of the problem, much appreciated. I will of course leave it to your discretion, but I would argue that it will impact many developers before it lands in core and they manage to update. So while this is not ideal, as temporary measure it might work to fix issues with pager view and expo video.

kkafar:
If you have capacity to patch your project with https://github.com/software-mansion/react-native-screens/pull/2531 and let me know whether it resolves it issue it would be great.

kkafar:
I'll close the PR as it should be superseded and the issue should be handled by #2531

Thank you for investigating this!

In case something still does not work for you let me know

Issue #2536, opened 2024-11-22
Header jump when animation is set for bottom tab navigator


matinzd:
### Description

Moving this issue from https://github.com/react-navigation/react-navigation/issues/12273.

> I can't repro this on JS stack, and the headerTopInsetEnabled prop for screens is always set to true. So it seems to be an issue with react-native-screens.

The header will have an initial jump when animation is set for tab navigator. It doesn't matter if it's set to `fade` or `shift`.

https://github.com/user-attachments/assets/8435fe72-3fad-4e48-a27e-ce92c11d1ce9

### Expected behavior

The header should not flicker or have jumps.

https://github.com/user-attachments/assets/b13aeb15-2d52-4955-aeba-87ae63f6ac84

### Steps to reproduce

Set bottom tab animation to `fade` or `shift`.

### Snack or a link to a repository

https://github.com/matinzd/ReactNavigationHeaderFlicker

### Screens version

4.1.0

### React Native version

0.76.2

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Acknowledgements

Yes

lucasmrl:
+1

matinzd:
That is the same issue but it's not caused by the same thing. Now the new tab animation is causing this issue which is not the same as that one.

The workaround may work but it's not a correct solution. I also usually use scroll view with automatic inset adjustment so that I don't need to take care of the header height.

lucasmrl:
I agree with you.  And that hack actually doesn't fix the issue.

hirbod:
Actually this is not related to animation or bottom tabs. I can reproduce this bug super easily. If one screen has a header and the one you’re navigating to not (or vice versa), the bug happens. Not always, but like 90% of times.

Our App is visually pretty much broken now, since we upgraded to SDK52.

cc @kkafar

hirbod:
Alright, I could pretty much narrow it down. It is indeed related to the new animation I guess, and in my case it was triggered because we enabled `fullScreenGestureEnabled`, which under the hood switches to simple_push.

It's now also reproducible in the Expo Snack using stock react navigation v7. https://snack.expo.dev/@hirbod/native-stack-navigator-header-to-non-header

I suspect this PR caused it: [https://github.com/software-mansion/react-native-screens/pull/2477](https://t.co/dPkvFtowmM)
[...]

PR #2565, opened 2024-12-10
fix(iOS): restore behaviour of RNSScreenStackAnimationNone

kkafar:
## Description

WIP WIP

There is still some header animation noticable for some reason.

Note: Must be implemented with old animation API, because `UIViewPropertyAnimator` does not allow for 0 duration (it uses default if the specified animation duration is below some undocumented treshold).

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStackAnimator.mm
+++ b/ios/RNSScreenStackAnimator.mm
@@ -59,7 +59,7 @@ static constexpr float RNSShadowViewMaxAlpha = 0.1;
   }
 
   if (screen != nil && screen.stackAnimation == RNSScreenStackAnimationNone) {
-    return 0;
+    return 0.0;
   }
 
   if (screen != nil && screen.transitionDuration != nil && [screen.transitionDuration floatValue] >= 0) {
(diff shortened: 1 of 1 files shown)


matinzd:
Is this a fix for https://github.com/software-mansion/react-native-screens/issues/2536?

hirbod:
@matinzd no, this is for supporting animation: none again, which is broken since v4, as it always fades.
Try to install  react-native-screens `4.4.0-rc.0` to see if #2536 is fixed.

hirbod:
Can we get the fixed none behavior shipped with another rc? Or do I have to patch?

kkafar:
I'll release it today's evening

kkafar:
I'm tracking now the "header animation", when the `animation: none` is set, but it seems that this behaviour was also present before v4 & must be fixed separately.

https://github.com/user-attachments/assets/f3099dd7-c2c5-48d0-ba13-fc781db7e5db

hirbod:
Yeah, that is great. Thanks for fixing the header mess, was always very unfortunate

Issue #2557, opened 2024-12-07
navigation.preload() with enableFreeze breaks app, stopping rendering


hirbod:
### Description

React Navigation v7 introduces `navigation.preload()`.

If you preload a screen without navigating to it, the entire app stops rendering because `react-freeze` is triggered, halting the rendering process. To resolve this, you must navigate to the preloaded screen and then pop it to unfreeze the app.

I discovered this issue by calling `preload()` in an effect and during `onPressIn`, then aborting the press (by moving my finger away). In both cases, the app stopped rendering updates.

However, once I navigated to the preloaded screen and returned, the app resumed working as expected.

React Native Screens should avoid applying any freezing logic when a screen is pre-rendered using `preload()`.

React Navigation already passes activityState, so react-freeze should read it:
https://github.com/react-navigation/react-navigation/blob/8209a855b312db1104270774810f6d9a7e0395b6/packages/native-stack/src/views/NativeStackView.native.tsx#L283

### Steps to reproduce

Everything said already, the Snack is fake, sorry! The description is clear enough.

### Snack or a link to a repository

https://snack.expo.dev/@hirbod/fake

### Screens version

4.3.0

### React Native version

0.76.3

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

kkafar:
Reproducible! Thanks

hirbod:
I think the cause is pretty clear now: https://github.com/software-mansion/react-native-screens/blob/main/src/components/Screen.tsx#L275

@satya164 do you think it would make sense to introduce a new state number for preloaded screens?

WoLewicki:
I'm working on it right now. Does it happen only in native-stack?

satya164:
I'm fine with anything, new state or separate prop

hirbod:
I was only able to reproduce it on native-stack, didn't freeze on web for me

WoLewicki:
I guess it is this logic then: https://github.com/software-mansion/react-native-screens/blob/8b9b9f217a85710ea0b91354f396b2e1180a61e2/src/components/ScreenStack.tsx#L90-L97 for native-stack, which just checks children, not looking at their `activityState`.

WoLewicki:
I guess you only saw it on old arch @hirbod ?

hirbod:
[...]

PR #2566, opened 2024-12-10
fix: take preload into account when freezing children

WoLewicki:
## Description

Fixed https://github.com/software-mansion/react-native-screens/issues/2557. We have to take preloaded screens into account when freezing screens. Till now, the assumption was that the newest screens are the ones rendered, right now top children can be the preloaded ones, so we have to filter them from calculating the size of frozen screens. We could also think of freezing the preloaded screens, wdyt about it @satya164 ?

TestFreeze now includes preloading of screens, even more than 1 to see that the current one still gets the updates.

## Checklist

### DIFF
--- a/src/components/Screen.tsx
+++ b/src/components/Screen.tsx
@@ -177,6 +177,7 @@ export const InnerScreen = React.forwardRef<View, ScreenProps>(
     const {
       enabled = screensEnabled(),
       freezeOnBlur = freezeEnabled(),
+      shouldFreeze,
       ...rest
     } = props;
 
@@ -271,8 +272,12 @@ export const InnerScreen = React.forwardRef<View, ScreenProps>(
         }
       };
 
+      const freeze =
+        freezeOnBlur &&
+        (isNativeStack ? Boolean(shouldFreeze) : activityState === 0);
+
       return (
-        <DelayedFreeze freeze={freezeOnBlur && activityState === 0}>
+        <DelayedFreeze freeze={freeze}>
           <AnimatedScreen
             {...props}
             /**
--- a/src/components/Screen.tsx
+++ b/src/components/Screen.tsx
@@ -274,7 +274,7 @@ export const InnerScreen = React.forwardRef<View, ScreenProps>(
 
       const freeze =
         freezeOnBlur &&
-        (isNativeStack ? Boolean(shouldFreeze) : activityState === 0);
+        (shouldFreeze !== undefined ? shouldFreeze : activityState === 0);
 
       return (
         <DelayedFreeze freeze={freeze}>
--- a/src/components/ScreenStack.tsx
+++ b/src/components/ScreenStack.tsx
@@ -77,6 +77,12 @@ function ScreenStack(props: ScreenStackProps) {
     },
   });
 
+  const preloadedScreensCount = React.Children.toArray(children).filter(
+    // @ts-expect-error it's either SceneView in v6 or RouteView in v5
+    child => child.props.isPreloaded,
+  ).length;
+
+  const renderedScreensSize = size - preloadedScreensCount;
   // freezes all screens except the top one
   const childrenWithFreeze = React.Children.map(children, (child, index) => {
     // @ts-expect-error it's either SceneView in v6 or RouteView in v5
@@ -88,8 +94,8 @@ function ScreenStack(props: ScreenStackProps) {
     // On Fabric, when screen is frozen, animated and reanimated values are not updated
     // due to component being unmounted. To avoid this, we don't freeze the previous screen there
     const freezePreviousScreen = isFabric()
-      ? size - index > 2
-      : size - index > 1;
+      ? renderedScreensSize - index > 2
+      : renderedScreensSize - index > 1;
 
     return (
       <DelayedFreeze freeze={isFreezeEnabled && freezePreviousScreen}>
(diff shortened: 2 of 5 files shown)


src/components/ScreenStack.tsx
```diff
@@ -77,6 +77,12 @@ function ScreenStack(props: ScreenStackProps) {
     },
   });
 
+  const preloadedScreensCount = React.Children.toArray(children).filter(
+    // @ts-expect-error it's either SceneView in v6 or RouteView in v5
+    child => child.props.isPreloaded,
+  ).length;
```
satya164, line 83: I think it's a bit fragile to read it like this, and can break if we reorganize our components. It'd be better to have something more explicit that can be type checked.

satya164:
> We could also think of freezing the preloaded screens

I think we need to leave them unfrozen so that effects etc. can run on that screen to load everything required.

WoLewicki:
> I think we need to leave them unfrozen so that effects etc. can run on that screen to load everything required.

I mean there is one render before the screen is frozen, but it might not be enough so it makes sense.

src/components/ScreenStack.tsx
WoLewicki: Yeah, it was done quickly to check if the logic is correct. Still, we get the direct children of `ScreenStack`, so how would you see it done better?

hirbod:
Freezing usually doesn’t affect effects, right? It only stops re-rendering the UI, no? I think preloaded screens can be frozen—it’s just their UI not being updated. But I think that’s kind of against the whole purpose of preloading or pre-rendering a screen. So, it makes sense not to freeze preloaded screens.

WoLewicki:
We could also add prop `freezeOnPreload` or something like that 🤷‍♂️ I'm okay with doing the logic on the `react-navigation` side and just pass a prop to children deciding if the screen should be frozen or not. Or even do the freezing on the navigation side too to keep it always in sync?

satya164:
> I'm okay with doing the logic on the react-navigation side and just pass a prop to children deciding if the screen should be frozen or not. Or even do the freezing on the navigation side too to keep it always in sync

Either is fine but I think the second one will be breaking, so for now I think it maybe good to add a new prop like `freeze` and whether a screen would be frozen or not can be decided by react navigation.

src/components/ScreenStack.tsx
satya164: @WoLewicki, it may not always be the direct child in how `React.Children` sees it. `React.Children` only sees it if you have written it as a direct child, but it could also be a direct child in terms of the React tree, but in a sub component.

in addition, this also relies on a prop called `isPreloaded` - an internal name, passed to an internal component `SceneView`. any refactor that changes how `isPreloaded` is determined or passed, or any change to the name of the prop or component signature can break this code.

using `React.Children` is only ok if you control the code that you access via the children API, or if your library's API already enforces this (like `Stack.Navigator` requires `Stack.Screen` as direct children with specific props passed to it).

since `ScreenStack`'s public API doesn't enforce what direct child its direct child is, and whether it should have an `isPreloaded` prop or not, accessing children here will always be fragile.

src/components/Screen.tsx
```diff
@@ -271,8 +272,12 @@ export const InnerScreen = React.forwardRef<View, ScreenProps>(
         }
       };
 
+      const freeze =
+        freezeOnBlur &&
+        (isNativeStack ? Boolean(shouldFreeze) : activityState === 0);
```
satya164, line 277: Maybe should always use `shouldFreeze` if a boolean is passed? I think for JS stack `activityState` will be `0` for preloaded screens as well so we will need a way for preloaded screens to opt-out.

WoLewicki: Ok, looking at it: https://github.com/react-navigation/react-navigation/blob/b70823fa58ea36217534e646f172516665b684a0/packages/stack/src/views/Stack/CardStack.tsx#L699-L701 it makes sense. But then we should add the `shouldFreeze` logic to stack too and remove relying on `activityState` at all.

satya164: @satya164 removing the logic would be breaking when using screens with react navigation 6 (same for native stack), so it's upto you

WoLewicki: So what is the official version of support? I heard that RNScreens v4 should be working only with navigation v7.

WoLewicki: And v6 is expected to not work correctly with v4

satya164: There's no breaking change in screens v4 that should affect React Navigation 6 JS Stack. There are some changes in header for native stack intended to fix some bugs but from my testing it doesn't break anything either - though I guess the bugs it's supposed to fix will still be there, but it's just keeping same behavior.

satya164 (approved):

kkafar (approved):
Both https://github.com/react-navigation/react-navigation/pull/12332/files and this one look reasonable.

 Only thing that missing is whether we want to document somewhere that we don't freeze preloaded screens or leave it "unstable". Same for fabric behaviour (top most two screens not frozen) we introduced a while back.

WoLewicki:
Yeah, it would be nice to inform people that behavior differs on old/new arch. As for preloaded screens, I would add a prop on e.g. navigator to decide if we should freeze preloaded screens too, but it is on the side of `react-navigation`. cc @satya164 for follow-ups on this one.

satya164:
@WoLewicki PRs to docs is welcome :)

hirbod:
I [started a discussion ](https://github.com/react-navigation/react-navigation/discussions/12337) about re-thinking react-freeze a little bit. They are definitely some flaws/inconsistencies that should be tackled in the future.

PR #2567, opened 2024-12-10
feat: add codegen `ios.componentProvider` to package.json

jakex7:
## Description

Add a newly introduced field `ios.componentProvider` to the `codegen` configuration to create an association map between JS components and their native implementations. 

See more information here: https://github.com/software-mansion/react-native-svg/pull/2572

## Changes

- added `codegenConfig.ios.componentProvider` field to package.json

### Before

```objc

Class<RCTComponentViewProtocol> RCTThirdPartyFabricComponentsProvider(const char *name) {
  static std::unordered_map<std::string, Class (*)(void)> sFabricComponentsClassMap = {
    // ...
    {"RNSFullWindowOverlay", RNSFullWindowOverlayCls}, // 3
    {"RNSModalScreen", RNSModalScreenCls}, // 3
    {"RNSScreenContainer", RNSScreenContainerCls}, // 3
    {"RNSScreenContentWrapper", RNSScreenContentWrapperCls}, // 3
    {"RNSScreenFooter", RNSScreenFooterCls}, // 3
    {"RNSScreen", RNSScreenCls}, // 3
    {"RNSScreenNavigationContainer", RNSScreenNavigationContainerCls}, // 3
    {"RNSScreenStackHeaderConfig", RNSScreenStackHeaderConfigCls}, // 3
    {"RNSScreenStackHeaderSubview", RNSScreenStackHeaderSubviewCls}, // 3
    {"RNSScreenStack", RNSScreenStackCls}, // 3
    {"RNSSearchBar", RNSSearchBarCls}, // 3
    // ...
```
### After

```objc
@implementation RCTThirdPartyComponentsProvider

+ (NSDictionary<NSString *, Class<RCTComponentViewProtocol>> *)thirdPartyFabricComponents
{
  return @{
		@"RNSFullWindowOverlay": NSClassFromString(@"RNSFullWindowOverlay"), // react-native-screens
		@"RNSModalScreen": NSClassFromString(@"RNSModalScreen"), // react-native-screens
		@"RNSScreenContainer": NSClassFromString(@"RNSScreenContainer"), // react-native-screens
		@"RNSScreenContentWrapper": NSClassFromString(@"RNSScreenContentWrapper"), // react-native-screens
		@"RNSScreenFooter": NSClassFromString(@"RNSScreenFooter"), // react-native-screens
		@"RNSScreen": NSClassFromString(@"RNSScreen"), // react-native-screens
		@"RNSScreenNavigationContainer": NSClassFromString(@"RNSScreenNavigationContainer"), // react-native-screens
		@"RNSScreenStackHeaderConfig": NSClassFromString(@"RNSScreenStackHeaderConfig"), // react-native-screens
		@"RNSScreenStackHeaderSubview": NSClassFromString(@"RNSScreenStackHeaderSubview"), // react-native-screens
		@"RNSScreenStack": NSClassFromString(@"RNSScreenStack"), // react-native-screens
		@"RNSSearchBar": NSClassFromString(@"RNSSearchBar"), // react-native-screens
  };
}
```

Run `pod install` in React Native 0.77 app and see `RCTThirdPartyFabricComponentsProvider` to check the content

### DIFF
--- a/package.json
+++ b/package.json
@@ -153,6 +153,21 @@
     "jsSrcsDir": "./src/fabric",
     "android": {
       "javaPackageName": "com.swmansion.rnscreens"
+    },
+    "ios": {
+      "componentProvider": {
+        "RNSFullWindowOverlay": "RNSFullWindowOverlay",
+        "RNSModalScreen": "RNSModalScreen",
+        "RNSScreenContainer": "RNSScreenContainer",
+        "RNSScreenContentWrapper": "RNSScreenContentWrapper",
+        "RNSScreenFooter": "RNSScreenFooter",
+        "RNSScreen": "RNSScreen",
+        "RNSScreenNavigationContainer": "RNSScreenNavigationContainer",
+        "RNSScreenStackHeaderConfig": "RNSScreenStackHeaderConfig",
+        "RNSScreenStackHeaderSubview": "RNSScreenStackHeaderSubview",
+        "RNSScreenStack": "RNSScreenStack",
+        "RNSSearchBar": "RNSSearchBar"
+      }
     }
   },
   "packageManager": "yarn@4.1.1"


kkafar:
Why don't we just use [`react-native.config.js`](https://github.com/software-mansion/react-native-screens/blob/main/react-native.config.js)? Do we need to have all our components declared in both of these places :smile: I remember that the mechanism you described was just recently introduced, maybe we can do something to unify these before 0.77 is released. 

I'm thinking right now - it is possible that these fields in `react-native.config.js` are no longer utilised & we just haven't removed them, but it would be quite nice to get confirmation.

jakex7:
I'm not sure if `react-native.config.js` is still necessary to generate `autolinking.cpp` on android, however, the changes mentioned above only impact iOS and have been introduced to remove those functions in the future:
```objc
#ifdef RCT_NEW_ARCH_ENABLED
Class<RCTComponentViewProtocol> RNSModalScreenCls(void)
{
  return RNSModalScreen.class;
}
#endif
```

kkafar (approved):
Okay, this looks solid then. I've glanced at the docs & see that there should be no backward compat issue.

PR #2568, opened 2024-12-10
fix(iOS, Paper): prevent double modal dismissal

### LINKED ISSUE #2525: `ScreenStack.onDismissed` incorrectly called after a navigating back, causing double-back navigation in old architecture
### Description

In a navigator with 2 card modals in the stack, calling `navigation.goBack()` causes a double-pop (navigating back 2 screens, instead of 1). This only occurs on the old architecture, switching to the new architecture has the correct behaviour.

It appears that the second `POP` action is being dispatched from `ScreenStack.onDismissed` here: https://github.com/react-navigation/react-navigation/blob/5da190167882859bb3356a84de42be3ca540ef99/packages/native-stack/src/views/NativeStackView.native.tsx#L538-L546

### Steps to reproduce

1. Press `Open modal`
2. Expect screen `One`
3. Press `Open second modal`
4. Expect screen `Two`
5. Press `Back`
6. Expect screen `One` 🚫 

https://github.com/levibuzolic/model-navigation-repro

[Same code as a Expo Snack](https://snack.expo.dev/@levibuzolic/react-navigation---nested-go_back-action), but *does not* reproduce the issue now that Expo Snack is running the new architecture.

### Screens version

4.2.0

### React Native version

0.76.2

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

_No response_

### Acknowledgements

Yes

# Old Architecture

Demonstrates broken behaviour, two models are removed with a single `goBack()` action.

https://github.com/user-attachments/assets/d090d001-98d7-4392-a2b7-14c1494c60fd

# New Architecture

Shows correct behaviour, only one screen removed.

https://github.com/user-attachments/assets/8c1af769-a05d-48fd-91b3-a7a4e6635b65

kkafar:
## Description

Fixes #2525

View hierarchy for modal with header looks kinda like this:

`UITransitionView` -> `Screen` -> `Stack` (`UINavigationController`) -> `Screen` -> content

calling `[_controller dismissViewControllerAnimated:YES completion:nil]` on the `Stack` :point_up:, according to [documentation](https://developer.apple.com/documentation/uikit/uiviewcontroller/dismiss(animated:completion:)?language=objc)
should dismiss the view controller itself and any view controllers presented from that view controller. However in practive it dismisses not only itself and "above" modal, but also a single modal "below". I'm not sure
why it is the case. Removing this line of code introduces a bug: during development, when multiple modals are opened and you reload the react-native, the modals could be left in stuck state. I'll proceed, as this bug 
bug in development is of much lesser severity. 

I've decided to move the call "one up" the view controller hierarchy, to avoid calling dismiss directly on UINavigationController as it looks like it causes the problem. This fixes the problem & keeps the development behaviour working as intended. 

## Changes

Described above :point_up:

I've enhanced `TestModalNavigation` test case to cover this issue. 

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/apps/src/tests/TestModalNavigation.tsx
+++ b/apps/src/tests/TestModalNavigation.tsx
@@ -7,10 +7,12 @@ type StackParamList = {
   Home: undefined;
   NestedStack: undefined;
   MainStackScreen: undefined;
+  MainStackScreen2: undefined;
   Screen1: undefined;
   Screen2: undefined;
   Screen3: undefined;
 };
+
 
 function HomeScreen({
   navigation,
@@ -35,6 +37,22 @@ function MainStackScreen({ navigation }: NativeStackScreenProps<StackParamList, 
     <View style={{ flex: 1, alignItems: 'center', justifyContent: 'center' }}>
       <Text>Main stack screen</Text>
       <Button
+        title="Go to second modal"
+        onPress={() => navigation.navigate('MainStackScreen2')}
+      />
+      <Button
+        title="goBack"
+        onPress={() => navigation.goBack()}
+      />
+    </View>
+  );
+}
+
+function MainStackScreen2({ navigation }: NativeStackScreenProps<StackParamList, 'MainStackScreen2'>) {
+  return (
+    <View style={{ flex: 1, alignItems: 'center', justifyContent: 'center' }}>
+      <Text>Main stack screen 2</Text>
+      <Button
         title="goBack"
         onPress={() => navigation.goBack()}
       />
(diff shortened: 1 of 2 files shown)


gusthavosouza:
@kkafar Amazing to see this live already. Thank you for the good work on this!

I just tested it on our app and it works like a charm!

PR #2581, opened 2024-12-18
feat: add support for `react-native@0.77`

kkafar:
## Description

Adding support for react-native 0.77. 
After this PR is merged the library will support 0.74+ Paper and 0.77 Fabric.

Please note, that I haven't notice any not-handled breaking changes - the library should work with versions down to 0.72 on old architecture and at least down to 0.76 on Fabric, however we won't extend issue resolving to these versions. 

## Changes

* adjusted example apps on both architectures with changes from upgrade helper,
* bumped `safe-area-context` to 5.0.0 (it comes with support for Paper 0.74+) in examples (not lib!),
* patched `safe-area-context`:
    * https://github.com/th3rdwave/react-native-safe-area-context/pull/566
* Removed patches for and dependencies on reanimated and gesture-handler - awaiting for support, see below 👇🏻 (comment) <-- **this must be restored for e2e to work!!!**
* Reanimated uses [the patch](https://github.com/software-mansion/react-native-reanimated/pull/6768#issuecomment-2579236084)
* Gesture handler has been set to [current main](https://github.com/software-mansion/react-native-gesture-handler/tree/c3ced78e1608e63b5b3a9bca0c66cc11f4029605)

When doing some quick checks I haven't noticed any obvious regressions.

## Checklist

- [ ] Ensured that CI passes


Example/patches/react-native-safe-area-context+5.0.0.patch
kkafar, line 1: This is not required for Fabric, as these files are auto generated there.

kkafar (commented):
Ci fails because I've temporarily removed `react-native-gesture-handler` and `react-native-reanimated` from dependencies of example apps. I'll restore them later, once we get closer to 0.77 being supported in these libs.

Reanimated PR:

* https://github.com/software-mansion/react-native-reanimated/pull/6768

Gesture handler PR:

* https://github.com/software-mansion/react-native-gesture-handler/pull/3278
* https://github.com/software-mansion/react-native-gesture-handler/pull/3277

kkafar:
The CI does not pass due to hardened mode, despite the fact that I've updated the locks in every way possible (I'm aware of) locally & pushed the changes out.

kkafar:
Blocker: locally, when I enable reanimated I get following error:

image

(3.16.1 + patch)

vonovak:
@kkafar maybe this helps: https://github.com/software-mansion/react-native-gesture-handler/pull/3278/files#diff-cd4c9ce6c3a5ed03b8eba2b60a6acca30452aed7e8f89210c5b4b6338eb91ec6

kkafar:
@vonovak Yeah, this was the culprit: https://github.com/facebook/react-native/pull/47551/files

kkafar:
Verified that both Fabric & Paper do build & applications are functional. Let's ignore the CI.

Issue #2591, opened 2024-12-27
Failed to create android build due to cmake debug|arm64-v8a : com.google.prefab.api.NoMatchingLibraryException: No compatible library found for //ReactAndroid/hermestooling. Rejected the following libraries


Muhammad-Hassam:
### Description

i've updated react native version from 0.73.4 to 0.76.5 using react native upgrade helper and update other libraries also. IOS build is successfully running but giving error on android

Task :app:compileDebugKotlin FAILED

Task :react-native-screens:configureCMakeDebug[arm64-v8a] FAILED
C/C++: /Users/Downloads/node_modules/react-native-screens/android/CMakeLists.txt debug|arm64-v8a : com.google.prefab.api.NoMatchingLibraryException: No compatible library found for //ReactAndroid/hermestooling. Rejected the following libraries:
425 actionable tasks: 44 executed, 381 up-to-date

### Steps to reproduce

1. install react-native-screen^4.4.0
2. cd android && ./gradlew --stop
3. yarn android

### Snack or a link to a repository

https://github.com/

### Screens version

^4.4.0

### React Native version

0.76.5

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

None

### Acknowledgements

Yes

fynally:
`react-native-screens:buildCMakeDebug[arm64-v8a] FAILED` I keep getting this very issue no matter the version of react native i use.

atifaziz1:
I am getting below error even for new project as well.

`2 files found with path 'lib/arm64-v8a/libfbjni.so' from inputs:
 - /Users/userName/Documents/sme/SmeApp/node_modules/react-native-screens/android/build/intermediates/library_jni/debug/copyDebugJniLibsProjectOnly/jni/arm64-v8a/libfbjni.so
 
 - /Users/userName/.gradle/caches/8.10.2/transforms/95f6f58ffe80f07573bbc4f1f4f0ecd0/transformed/jetified-react-android-0.76.5-debug/jni/arm64-v8a/libfbjni.so`

kkafar:
[...]

PR #2612, opened 2025-01-10
fix(Android): should not set cmake path in gradle when new architectu…

sleepym09:
# Summary:
Android build will fail when newArchEnabled=false, and trhow below exception:

```
Caused by: com.android.builder.errors.EvalIssueException: [CXX1210] /Users/sleepym09/code/project_test/node_modules/react-native-screens/android/CMakeLists.txt debug|arm64-v8a : No compatible library found
	at com.android.builder.errors.IssueReporter.reportError(IssueReporter.kt:116)
	at com.android.builder.errors.IssueReporter.reportError$default(IssueReporter.kt:112)
```

Actually we needn't to set cmake path if newArchEnabled is false.

Maybe related issues:
- [#48399](https://github.com/facebook/react-native/issues/48399)
- [#47729](https://github.com/facebook/react-native/issues/47729)

## Test Plan:
React Native Android builds with this change without newArchEnabled and no more build error.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -130,9 +130,11 @@ android {
         prefab true
         buildConfig true
     }
-    externalNativeBuild {
-        cmake {
-            path "CMakeLists.txt"
+    if (isNewArchitectureEnabled()) {
+        externalNativeBuild {
+            cmake {
+                path "CMakeLists.txt"
+            }
         }
     }
     lintOptions {
@@ -159,7 +161,7 @@ android {
                 // Some of above lib* names could be removed after we remove support for 0.76.
                 // https://github.com/facebook/react-native/pull/43909
                 // https://github.com/facebook/react-native/pull/46059
-                "**/libfbjni.so", 
+                "**/libfbjni.so",
                 "**/libreactnative.so"
         ]
     }


kkafar:
Hey, thanks for the patch, however I do not think this is accurate. We have a [turbomodule which is build also on old architecture](https://github.com/software-mansion/react-native-screens/blob/b4fe2c13d1d4a7e7802ae5169fd888c80baf438d/android/CMakeLists.txt#L14-L20). The library will build, but some of the functionalities won't be available.

sleepym09:
Sorry to bother you.

I found the root cause, [This issue](https://android.googlesource.com/platform/tools/base/+/84342aae383e0b29039c741884ac17545c993578) is caused by how prefab is handled in AGP. It was introduced in AGP 8.4 and fixed in AGP 8.7. Therefore, upgrading or downgrading RN can resolve the problem.

Also I made a commit to solve this issue in the [minimal project](https://github.com/sleepym09/rn-playground), hope this helps others as well.

sleepym09:
> @sleepym09 thanks for investigating this. Does your solution apply to #2591?

The error message is slightly different, but I think it’s worth trying to modify `classpath("com.android.tools.build:gradle")` to `classpath("com.android.tools.build:gradle:8.7.0")` in the `root_project/build.gradle` to test it. This is because the AGP version used in RN 0.76.5 is 8.6.0, which still contains the issue.

PR #2466, opened 2024-10-29
fix: clicking on Pressable located in screen header

coado:
## Description

This PR fixes clicking on Pressables that are added to the native header. Previously, Yoga had incorrect information about the location of the content in the header. 

The first step was to remove `top: "-100%"` style from the `ScreenStackHeaderConfig` which made Yoga think, that the content is pushed up by the size of its parent (Screen size). 

The second step involved updating `layoutMetrics` of the `RNSScreenStackHeaderConfigShadowNode`. The entire app content is pushed down by the size of the header, so it also has an impact on the header config in Yoga metrics. To mitigate this, the origin of `RNSScreenStackHeaderConfigShadowNode` is decreased by the size of the header which will zero out eventually leaving the header content in the desired position. On iOS this position is actually moved by the top inset size, so we also have to take it into account when setting header config layout metrics.

## Changes

Updated `ScreenShadowNode` to decrease `origin.y` of the `HeaderConfigShadowNode` by the size of the header. Added `paddingTop` to `HeaderConfigState` and set it as origin offset on iOS.

### Before

| iOS       | Android        |
|--------------|--------------|
| <video src="https://github.com/user-attachments/assets/aa054d20-58e5-4153-924b-a1b74879ae6b">  | <video src="https://github.com/user-attachments/assets/dd9d2732-10c5-40ca-b384-1570c08410cf">  |

### After

| iOs       | Android        |
|--------------|--------------|
| <video src="https://github.com/user-attachments/assets/6953c9e5-b418-41b1-8b99-27f70c5df688">  | <video src="https://github.com/user-attachments/assets/f5762f46-36b2-423a-93da-9f61032b54bf">  |

## Test code and steps to reproduce

Tested on this example:

<details>
 <summary>code</summary>

```ts
import { NavigationContainer } from "@react-navigation/native";
import { createNativeStackNavigator, NativeStackNavigationProp } from "@react-navigation/native-stack";
import React, { ForwardedRef, forwardRef } from "react";
import { findNodeHandle, Pressable, PressableProps, StyleSheet, Text, View, } from "react-native";

type StackParamList = {
  Home: undefined,
}

type RouteProps = {
  navigation: NativeStackNavigationProp<StackParamList>;
}

type PressableState = 'pressed-in' | 'pressed' | 'pressed-out'


const Stack = createNativeStackNavigator<StackParamList>();


const PressableWithFeedback = forwardRef((props: PressableProps, ref: ForwardedRef<View>): React.JSX.Element => {
[...]

PR #2623, opened 2025-01-16
refactor(iOS): update view controller once per transaction when adding header subviews

kkafar:
## Description

In #2466 I've introduced `RCTMountingTransactionObserving` for `RNSScreenStackHeaderConfig` component.
Now we can use this to reduce amount of calls to `updateViewControllerIfNeeded` (causing layout passes) when adding subviews.

I'm not changing the `unmount` path of the code as we have some more additional logic there which requires some more 
careful consideration. 

This also aligns us with Paper implementation. Note that it has been only implemented this way, because at the implementation time there was no way
to run this code on transaction completion.

## Changes

`- [RNSScreenStackHeaderConfig updateViewControllerIfNeeded]` is now called only once per transaction when adding subviews to header config.

Test432, TestHeaderTitle, Test2552 - see that there are no regressions. 

> [!note]
During testing I've found that there is a bug with subview layout when modifying subviews after first render. This is not a regression however. Notice the wrong position of header right on video below 👇🏻 (Test432)

https://github.com/user-attachments/assets/7f8653e8-a7d9-4fb8-a875-182e7deb0495

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -883,9 +883,6 @@ RNS_IGNORE_SUPER_CALL_BEGIN
   [self insertReactSubview:(RNSScreenStackHeaderSubview *)childComponentView atIndex:index];
 
   _addedReactSubviewsInCurrentTransaction = true;
-
-  // TODO: This could be called only once per transaction.
-  [self updateViewControllerIfNeeded];
 }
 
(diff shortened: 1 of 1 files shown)


kkafar:
Lint started failing w/o any changes to the PR (before force push it passed). Moreover it has started failing on every single PR 🤷🏻‍♂️ indicating separate issue.

PR #2640, opened 2025-01-21
fix(Android): Restore focus on page transitions

### LINKED ISSUE #1706: [ANDROID TV] Focus gets lost between page transitions

micheleb:
## Description

After #1894 was merged, a refactoring of the native source happened, and the call to `lastFocusedChild.requestFocus()` that happened on the overridden `onStart()` method in `ScreenStackFragment` got lost along the way, making focus disappear when popping a screen from the stack.

Fixes #1706 

## Changes

- Updated `ScreenStackFragment.kt`, resurrecting the old `onStart()` override (`lastFocusedChild` is only set on Android TV, so this change only affects that platform)

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
@@ -478,6 +478,11 @@ class ScreenStackFragment :
         screen.background = shape
     }
 
+    override fun onStart() {
+        lastFocusedChild?.requestFocus()
+        super.onStart()
+    }
+
     override fun onStop() {
         if (DeviceUtils.isPlatformAndroidTV(context)) {
             lastFocusedChild = findLastFocusedChild()


kkafar:
Yeah, the code was accidentally removed in #2045. Thanks a lot!

kkafar (approved):
Accidentally removed in #2045. Thanks!

PR #2660, opened 2025-01-29
fix(Android): missing background color for formSheets on RN 0.77+

kkafar:
## Description

* https://github.com/facebook/react-native/pull/45688 introduced `BackgroundStyleApplicator`
* https://github.com/facebook/react-native/pull/47906 removed our current accessor
* somewhere along the way the `ColorDrawable` used previously by `ReactViewGroup` has been exchanged for `CompositeBackgroundDrawable` added in https://github.com/facebook/react-native/pull/45688 

> [!caution]
~This PR breaks compatibility with older versions of react-native. While this is fine on Fabric, this also breaks things for Paper.~
> ~Possible solution is to detect react native version in gradle and add versioned sourcesets with implementations for given react native versions.~
>
> Not up to date. I've added versioned source files to ensure appropriate backward compatibility down to 0.74.

> [!note]
> We need CI to ensure the projects do build on all versions we support.

## Changes

Migrated to `BackgroundStyleApplicator API` to resolve background color of `contentWrapper`.  

`TestAndroidTransitions` - the form sheet should no longer be cut.

## Checklist

- [ ] Ensured that CI passes


RohovDmytro:
Could this PR led to https://github.com/software-mansion/react-native-screens/issues/2738 ?

PR #2678, opened 2025-02-09
[Question] Is crash on RNSNavigationController when pushing non-RNSScreen ViewController intended?

e-sung:
## Description

This is basically a question PR. 

Given that RNSNavigationController is just another UINavigationController, I thought that I could push my ViewController to it.

But when I did, the app crashed on file, where this is PR is addressing.

RNSScreenStack tried to call a method `screen` on my ViewController, and because my ViewController had nothing to do with RNSScreen, the app crashed.

But, as I changed the code as this PR, the crash didn't happen, and navigation happend smoothly without any side-effects.

So, here's the question.
Is crash on RNSNavigationController when pushing non-RNSScreen ViewController expected, or intended?

If not, I think it'd be great to allow it...!

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStack.mm
+++ b/ios/RNSScreenStack.mm
@@ -751,6 +751,9 @@ RNS_IGNORE_SUPER_CALL_END
                                                fromViewController:(UIViewController *)fromVC
                                                  toViewController:(UIViewController *)toVC
 {
+  if (![toVC.view isKindOfClass:[RNSScreenView class]]) {
+    return nil;
+  }
   RNSScreenView *screen;
   if (operation == UINavigationControllerOperationPush) {
     screen = ((RNSScreen *)toVC).screenView;


RohovDmytro:
Interesting.

@e-sung , I'm experiencing plenty of crashes in recent version. Can it have something to do with your question?

image

kkafar:
Hey @e-sung, 

`ScreenStack` (natively represented on iOS by `RNSNavigationController` + `RNSScreenStackView`) accepts only children of type `Screen`. This is an assumption that we rely on in numerous places in our codebase and in several places we assert it. 

At this moment we do not plan to relax this assumption. 

If you wan't to play around you could always inherit from `RNSScreen` though.

e-sung:
> Hey @e-sung,
> 
> `ScreenStack` (natively represented on iOS by `RNSNavigationController` + `RNSScreenStackView`) accepts only children of type `Screen`. This is an assumption that we rely on in numerous places in our codebase and in several places we assert it.
> 
> At this moment we do not plan to relax this assumption.
> 
> If you want to play around you could always inherit from `RNSScreen` though.

Got it! Thank you for the quick response!

PR #2682, opened 2025-02-11
fix(Android): missing transition events with formSheet presentation

kkafar:
## Description

This PR aligns* events received in element tree between Android and iOS for screens with `presentation: 'formSheet'`.

\* **almost aligns** - on iOS when natively dismissed `onDismissed` and `onDisappear/transitionEnd` will be received, whereas on Android it will be only `onDismissed` for now. 
Adding support for `onDisappear/transitionEnd` requires wider changes in logic which I don't want in incoming 4.7.0 release.

## Changes

`formSheet` animations based on custom animators now have listener set, which emmits appropriate events **bypassing** "regular" view emitting infrastructure.
I've found that `ScreenFragment.onViewAnimation{Start,End}` methods do not work for some reason (`isResumed` is false for form sheet) with the sheets, therefore I bypassed it for now.

`TestAndroidTransition` + add logs in `NativeStackView` in `react-navigation` to see when events are received (or attach appropriate listeners).

## Checklist

- [ ] Ensured that CI passes


android/src/main/java/com/swmansion/rnscreens/events/ScreenEventEmitter.kt
```diff
@@ -0,0 +1,36 @@
...
+    fun dispatchOnWillDisappear() =
+        reactEventDispatcher?.dispatchEvent(ScreenWillDisappearEvent(reactSurfaceId, screen.id))
+
+    fun dispatchOnDisappear() =
+        reactEventDispatcher?.dispatchEvent(ScreenDisappearEvent(reactSurfaceId, screen.id))
+
+    fun dispatchOnDismissed() =
+        reactEventDispatcher?.dispatchEvent(ScreenDismissedEvent(reactSurfaceId, screen.id))
+
+    fun dispatchTransitionProgress(progress: Float, isExitAnimation: Boolean, isGoingForward: Boolean) {
```
kkafar, line 31: Not entirely sure of `isGoingForward` semantics here. This is something to retrospect later.

kkafar (commented):
Notes for future me

android/src/main/java/com/swmansion/rnscreens/events/ScreenEventEmitter.kt
kkafar: Created ticket on board for this

PR #2683, opened 2025-02-11
fix(iOS): look through whole ancestor chain when looking for screenview from content wrapper

kkafar:
## Description

In #2670 I've added a check that logged RN error in case `contentWrapper.reactSuperview` is not a `RNSScreenView`. 
Now I got reports (but no reproduction :/) from two independent sources that this error is triggered pretty often,
which is unexpected from my perspective. I blindly blame view flattening / usage of legacy navigators but can not be sure.

This PR is an attempt to patch this behaviour.

## Changes

contentWrapper now searches whole view ancestor chain before logging an error.

No reproducer :/ 

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenContentWrapper.mm
+++ b/ios/RNSScreenContentWrapper.mm
@@ -87,7 +87,9 @@ namespace react = facebook::react;
   UIView *currentView = self;
 
   // In standard scenario this should do only a single iteration.
-  // Haven't got repro, but we got
+  // Haven't got repro, but we got reports that there are scenarios
+  // when there are intermediate views between screen view & the content wrapper.
+  // https://github.com/software-mansion/react-native-screens/pull/2683
   do {
     currentView = currentView.reactSuperview;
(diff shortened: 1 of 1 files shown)


RohovDmytro:
It did start to work in most places, so you seem to be right. Cool! 

At least one place is broken, when the app gets logged out and I remove native stack container for authorized users with a container for guests.

image

I'll let you know if I can pinpoint any useful details.

kkafar:
@RohovDmytro thanks for testing & feedback. The best for me would be to get a reproduction of the issue. Then I'll be able to patch this correctly.

RohovDmytro:
@kkafar I've created a bunch of issues that are highlighting my findings of the last week. I hope that's more helpful as they have reproductions. Thanks for looking at them, kkafar! Ping me if you will need any details on any issue.

PR #2685, opened 2025-02-12
fix(Android,Fabric): inifinite state update loop in header subview

kkafar:
## Description

Small oversight (:sweat-smile:) but we have infinite state update loop in header subview caused by not comparing 
current value to previous one (check out code changes, it will be obvious).

We have similar checks in `FabricEnabledViewGroup` & `FabricEnabledHeaderConfigViewGroup`.

## Changes

Added a check that prevents the loop.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/fabric/java/com/swmansion/rnscreens/FabricEnabledHeaderSubviewViewGroup.kt
+++ b/android/src/fabric/java/com/swmansion/rnscreens/FabricEnabledHeaderSubviewViewGroup.kt
@@ -14,8 +14,8 @@ abstract class FabricEnabledHeaderSubviewViewGroup(
 ) : ViewGroup(context) {
     private var mStateWrapper: StateWrapper? = null
 
-    private var lastWidth = -1f
-    private var lastHeight = -1f
+    private var lastWidth = 0f
+    private var lastHeight = 0f
 
     fun setStateWrapper(wrapper: StateWrapper?) {
         mStateWrapper = wrapper
--- a/android/src/fabric/java/com/swmansion/rnscreens/FabricEnabledHeaderSubviewViewGroup.kt
+++ b/android/src/fabric/java/com/swmansion/rnscreens/FabricEnabledHeaderSubviewViewGroup.kt
@@ -7,11 +7,15 @@ import com.facebook.react.bridge.WritableMap
 import com.facebook.react.bridge.WritableNativeMap
 import com.facebook.react.uimanager.PixelUtil
 import com.facebook.react.uimanager.StateWrapper
+import kotlin.math.abs
 
 abstract class FabricEnabledHeaderSubviewViewGroup(
     context: Context?,
 ) : ViewGroup(context) {
     private var mStateWrapper: StateWrapper? = null
+
+    private var lastWidth = -1f
+    private var lastHeight = -1f
 
     fun setStateWrapper(wrapper: StateWrapper?) {
         mStateWrapper = wrapper
@@ -38,6 +42,17 @@ abstract class FabricEnabledHeaderSubviewViewGroup(
         val offsetXDip: Float = PixelUtil.toDIPFromPixel(offsetX.toFloat())
         val offsetYDip: Float = PixelUtil.toDIPFromPixel(offsetY.toFloat())
 
+        // Check incoming state values. If they're already the correct value, return early to prevent
+        // infinite UpdateState/SetState loop.
+        if (abs(lastWidth - realWidth) < DELTA &&
+            abs(lastHeight - realHeight) < DELTA
+        ) {
+            return
+        }
+
+        lastWidth = realWidth
+        lastHeight = realHeight
+
         val map: WritableMap =
             WritableNativeMap().apply {
                 putDouble("frameWidth", realWidth.toDouble())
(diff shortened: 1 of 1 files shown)


sbeigel:
Hi, this fix solves the height problems in the Android header, but it introduces a flickering of the left and/or right header components. This is hard to reproduce though, at least in the emulator. Any idea what's happening here?

https://github.com/user-attachments/assets/be77d0b9-2c78-46f3-8cc8-bbfafc91b6e0

sbeigel:
If I comment line 32 in

`//updateState(width, height, offsetX, offsetY)`

The flickering disappears but the header is (again) too high (using native stack and material top bar):

image

LesterWeng:
> Hi, this fix solves the height problems in the Android header, but it introduces a flickering of the left and/or right header components. This is hard to reproduce though, at least in the emulator. Any idea what's happening here?
> 
>  android-header-flicker.mov

I encountered the same issue. After debugging, I found that the solution is to prevent changes in the width and height of HeaderRight
try it~

sbeigel:
> I encountered the same issue. After debugging, I found that the solution is to prevent changes in the width and height of HeaderRight try it~

What do you mean? I've added the header items the documented way like this and don't change their dimensions dynamically or afterwards:

```
useEffect(() => {
    navigation.setOptions({
        headerRight: () => (
            <Pressable onPressIn={...}><Ionicon size={30} ... /></Pressable>
        ),
    })
}, [...])


```
This works in iOS and has worked for years in Android as well :) The flickering only appears using the latest RNS update.

sbeigel:
Updating to RNS 4.9.0 did not solve the problem. Neither did setting explicit dimensions in the header item.

I tested with just one native stack navigator with just one screen without any screen config (i.e. custom styling) and even the most basic header item (headerRight: () => <Text>Foo</Text>) is flickering.

kkafar:
I guess I've got some another state update loop, however I can't identify it, because I do not reproduce your issue.

kkafar:
Thanks a lot! I'll get back to you once I try it

sbeigel:
@kkafar It seems that going back to RNS 4.5.0 (but leaving all other deps at their current versions), the problem disappears. Even the "jumping" header (vertical size changes) we've experienced in our "real" app (using RNS 4.6.0 iirc) are gone.

kkafar:
@abulenok is looking into the issue in this very moment

sbeigel:
@abulenok any news?

sbeigel:
I just updated all the deps (including RN 0.78) in the reproducer and the issue still exists.

RNS 4.10.0-beta.2 doesn't solve it either.

kkafar:
https://github.com/software-mansion/react-native-screens/blob/2f7b5f6f55071511f0f48eef2d453c3f256bb9f1/common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderSubviewComponentDescriptor.h#L33-L36

For quick workaround you can swap `getMostRecentState()` with `getState()` & it should improve the situation (prevent update loop). 

However the soultion is not 100% stable - we encountered the issue that the subviews disappear (they receive 0 width) after fast refreshing the app multiple times (in dev mode). We have never experienced it after regular app start. We have tested only dev mode.

sbeigel:
Thank you! This seems to fix the flickering but we are back to the wrong header height when using @react-navigation/material-top-tabs (see my comment above: https://github.com/software-mansion/react-native-screens/pull/2685#issuecomment-2669073154 )

kkafar:
I'm working on it in this very moment. You can workaround the issue by setting the header element directly on `Screen` instead of doing it in `useEffect`.

sbeigel:
That would be an option in our simple test case, but not in the real world applications where you need access to all the data and/or functions *in* the actual screen. So it's not possible to setup those elements in App.tsx where the navigation is configured.

kkafar:
I see @sbeigel, but it was always worth recommending. I think I'm onto something, but reading Yoga code is not that easy. Please try patching these two lines: https://github.com/software-mansion/react-native-screens/blob/29c3ae2dd08e0b550685a791544f772f008d30a4/common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderConfigComponentDescriptor.h#L37-L38

to 

```c++
    if (stateData.frameSize.width != 0) {
      layoutableShadowNode.setSize({stateData.frameSize.width, YGUndefined});
```

sbeigel:
I reverted the change in RNSScreenStackHeaderSubviewComponentDescriptor.h (back to getMostRecentState()) and changed the 2 lines in RNSScreenStackHeaderConfigComponentDescriptor.h. The header's height is correct now, but the right header element is flickering again.

sbeigel:
Applying both of your suggested changes (getState() instead of getMostRecentState() AND the 2 lines in RNSScreenStackHeaderConfigComponentDescriptor.h, the header height is correct and the flickering seems gone. But sometimes (quite often), the right header elements are missing / not visible.

kkafar:
Yeah I know, I'm investigating it right now.  
Do you experience these missing / not visible header subviews on earlier versions? I'm testing react-native-screens 4.9.2 right now, wondering whether this is a regression or not.

sbeigel:
I'm not sure tbh.

Issue #2677, opened 2025-02-08
Can you take a look at the issue that I created on react navigation.


Pnlvfx:
### Description

 I tough that was their code but I've seen that they are just reexporting from here the header animation.

Link to the issue:

https://github.com/react-navigation/react-navigation/issues/12427

### Steps to reproduce

I've already explained above

### Snack or a link to a repository

none

### Screens version

4.6.0

### React Native version

0.77

### Platforms

iOS

### JavaScript runtime

None

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

Real device

### Device model

iphone 12 pro

### Acknowledgements

Yes

PR #2696, opened 2025-02-13
fix(Android,Fabric): prevent another infinite state update loop for header subviews with zero size

### LINKED ISSUE #2675: Android Header expands unexpectedly when headerSearchBarOptions are set in v4.6.0+
### Description

After upgrading from v4.5.0 to v4.6.0+, the header height on Android unexpectedly increases to a large value when navigating to a page with the search icon enabled. This issue does not occur on iOS, where the behavior remains normal.

Image

In v4.5.0, the header height functions correctly:
Image

Stack:

```
"react-native": "0.77.0",
"react-native-screens": "4.6.0",
"@react-navigation/native": "7.0.14",
"@react-navigation/native-stack": "7.2.0",
"react-native-reanimated": "4.0.0-beta.2",
...
```

### Steps to reproduce

I was able to reproduce the issue in a slightly different way using a minimal example. While navigating to a new page with search works correctly, setting headerSearchBarOptions on the initial screen causes the header height to display incorrectly in v4.6.0. In v4.5.0, the height briefly flickers from the incorrect value to the correct one.

1. npm i
2. npx expo prebuild --clean
3. npx expo run:android -d

https://github.com/marcel-happyfloat/android-screen-issue

### Screens version

4.6.0

### React Native version

0.77.0

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

None

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

kkafar:
## Description

Closes #2675

### Error mechanism

This one was tricky to figure out.

Few concurrent facts first:

1. When `statusBarTranslucent` is `true` we apply `paddingTop` to native `Toolbar` to heighten the header.
2. Since 4.6.0 we update frames of header subviews in ShadowTree to the values from HostTree, everytime the frame changes in HostTree.
3. We had a check in header subview shadow node that would prevent node size update if the frame was (0, 0).

Header subview that represents the search bar has no content. Therefore native layout sets it frame to exactly (0, 0), but with origin `(x, toolbar.paddingTop)`.
This frame - `((x, toolbar.paddingTop), (0, 0))` is send to shadow node, **where it is ignored**, but the layout is triggered by subsequent *header config* update.
Therefore Yoga resolves height of the subview to full height of the parent (usually `154 px = 40 dip`). New layout metrics are sent to HostTree, where next native toolbar layout
determines toolbar size to be its content height + padding == subview height + paddingTop. This gets send to ShadowTree, then back to HostTree, native layout is triggered and another
`paddingTop` value is added to the overall header height, and so on. 

This is a regression introduced in 4.6.0 and limited to Fabric. 

The fix is simple - accept the (0, 0) size for the subview in ShadowTree. We just need to use more reasonable value to denote the uninitialized frame - `{-1, -1}` seems like a better choice as 
it is an invalid frame.

## Changes

- **Use different initial value in state so that we can unambiguously distinguish between un- and initialized state**

`Test2675`

Tested on all combinations:

* Android SDK 34 (w/o edge-to-edge) and 35 (w/ edge-to-edge enabled),
* `statusBarTranslucent: true` and `false`,
* search bar present, not present
* other header elements present / not present.

Also tested iOS on the same example, because the code changes affect shared C++ code.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderSubviewComponentDescriptor.h
+++ b/common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderSubviewComponentDescriptor.h
@@ -8,6 +8,7 @@
 #include <react/renderer/components/rnscreens/utils/RectUtil.h>
 #include <react/renderer/core/ConcreteComponentDescriptor.h>
 #include "RNSScreenStackHeaderSubviewShadowNode.h"
+#include "utils/RectUtil.h"
 
 namespace facebook::react {
 
@@ -35,7 +36,7 @@ class RNSScreenStackHeaderSubviewComponentDescriptor final
         shadowNode.getState());
     auto stateData = state->getData();
 
-    if (stateData.frameSize.width != -1 && stateData.frameSize.height != -1) {
+    if (!isSizeEmpty(stateData.frameSize)) {
       layoutableShadowNode.setSize(
           Size{stateData.frameSize.width, stateData.frameSize.height});
     }
--- a/common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderSubviewComponentDescriptor.h
+++ b/common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderSubviewComponentDescriptor.h
@@ -33,9 +33,15 @@ class RNSScreenStackHeaderSubviewComponentDescriptor final
 
     auto state = std::static_pointer_cast<
         const RNSScreenStackHeaderSubviewShadowNode::ConcreteState>(
-        shadowNode.getState());
+        shadowNode.getMostRecentState());
     auto stateData = state->getData();
 
+    const auto frameSize = stateData.frameSize;
+    LOG(WARNING) << "HEADER_SUBVIEW (" << frameSize.width << ", "
+                 << frameSize.height << ")";
+
+    //    if (stateData.frameSize.width != 0 && stateData.frameSize.height != 0)
+    //    {
     if (!isSizeEmpty(stateData.frameSize)) {
       layoutableShadowNode.setSize(
           Size{stateData.frameSize.width, stateData.frameSize.height});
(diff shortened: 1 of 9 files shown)


kkafar (commented):
Currently blocked by regression detected in testing

Pnlvfx:
This seems related to the issue that I opened: #2677, I hope so.

Pnlvfx:
This seems related to the issue I opened: #2677

common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderSubviewComponentDescriptor.h
```diff
@@ -32,12 +33,11 @@ class RNSScreenStackHeaderSubviewComponentDescriptor final
 
     auto state = std::static_pointer_cast<
         const RNSScreenStackHeaderSubviewShadowNode::ConcreteState>(
-        shadowNode.getState());
```
kkafar, line 35: The bug I've reported in earlier comment was caused by accessing mounted state and not the most recent one.

marcel-happyfloat:
@kkafar  Thanks for the quick fix, unfortunately I can't install v4.7.0 on IOS now. 
Getting this error (`npx expo prebuild --clean` and `npx expo run:ios --no-build-cache)` : 

⚠️  ld: ignoring duplicate libraries: '-lc++'
❌  ld: framework 'RNScreens' not found

Android works now, but it looks like it is adding some safe area insets again (as a black bar) to the bottom of my screen, when I'm navigating (default presentation) to some other root screen.`modal` and screens within the current Tab Stack are  working fine... 

Just to let you quickly know...

kkafar:
@marcel-happyfloat thanks for the report, however I do need more details to investigate any of the issues you mentioned. 

Could you provide a recording of the bug and preferably a snippet / detailed description of navigation hierarchy where I can reproduce the issue?

PR #2706, opened 2025-02-18
Remove incorrect import

jankosecki:
4.7.0 comes with an incorrect import:
```
#include "utils/RectUtil.h"
``` 
which duplicates import of the same file in line 8:
```
#include <react/renderer/components/rnscreens/utils/RectUtil.h>
```


Zao721:
I suppose it is fixed in 
https://github.com/software-mansion/react-native-screens/pull/2711/files
And 4.8.0 contains this changes

kkafar:
Yeah, thanks for taking your time to open the PR, however it is already fixed & released.

Thanks for the effort, anyway!

PR #2723, opened 2025-02-20
fix(Android): fix sheet transition when there is no dimming applied

kkafar:
## Description

Shout out to @kligarski for spotting this out!

I've introduced a regression in 4.7.0 when refactoring sheet animation - basically it always animated to the resolved `maxAlpha`
for given sheet dimming view & did not take into account `sheetLargestUndimmedDetentIndex` prop, effectively disabling this prop.

This affects both architectures.

## Changes

When configuring enter animation we now take the prop into account.

`TestFormSheet`

## Checklist

- [ ] Ensured that CI passes


kligarski (approved):
lgtm

PR #2724, opened 2025-02-20
fix: Allow undefined ref for SearchBarCommands in SearchBarProps

Pnlvfx:
## Description

This change updates the type definition for SearchBarProps to accept a ref that can be undefined. Previously, the ref was forced to be non-nullable, which required developers to use workarounds (like non-null assertions) even though React handles null/undefined refs gracefully. With React 19, the framework ensures that refs are managed correctly regardless of their empty initial state. 

By allowing n undefined ref, we align the type definitions with React's natural behavior and improve the developer experience by eliminating unnecessary TypeScript errors. This change has no runtime impact, as the component lifecycle guarantees that the ref will be properly set before it is used.

Example that trigger the Typescript Error below:

  ```
useLayoutEffect(() => {
    navigation.setOptions({
      headerSearchBarOptions: {
        tintColor: colors.text,
        ref: searchBarRef,
      } satisfies SearchBarProps,
    });
  }, [navigation, colors.text]);
```
  
  Type 'RefObject<SearchBarCommands | undefined>' is not assignable to type 'RefObject<SearchBarCommands>'.
  Type 'SearchBarCommands | undefined' is not assignable to type 'SearchBarCommands'.
    Type 'undefined' is not assignable to type 'SearchBarCommands'.ts(2322)

### DIFF
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -687,7 +687,7 @@ export interface SearchBarProps {
    * * `cancelSearch` - cancel search in search bar.
    * * `toggleCancelButton` - depending on passed boolean value, hides or shows cancel button (iOS only)
    */
-  ref?: React.RefObject<SearchBarCommands>;
+  ref?: React.RefObject<SearchBarCommands | undefined>;
 
   /**
    * The auto-capitalization behavior
--- a/src/components/SearchBar.tsx
+++ b/src/components/SearchBar.tsx
@@ -16,7 +16,9 @@ import SearchBarNativeComponent, {
 import { DirectEventHandler } from 'react-native/Libraries/Types/CodegenTypes';
 
 const NativeSearchBar: React.ComponentType<
-  SearchBarNativeProps & { ref?: React.RefObject<SearchBarCommands> }
+  SearchBarNativeProps & {
+    ref?: React.RefObject<SearchBarCommands | undefined>;
+  }
 > &
   typeof NativeSearchBarCommands =
   SearchBarNativeComponent as unknown as React.ComponentType<SearchBarNativeProps> &
@@ -36,7 +38,7 @@ type SearchBarCommandsType = {
 };
 
 function SearchBar(props: SearchBarProps, ref: React.Ref<SearchBarCommands>) {
-  const searchBarRef = React.useRef<SearchBarCommands | null>(null);
+  const searchBarRef = React.useRef<SearchBarCommands>(undefined);
 
   React.useImperativeHandle(ref, () => ({
     blur: () => {

### CODE REVIEW

**Comment by Pnlvfx** [comment]
> ## Description
> This change updates the type definition for SearchBarProps to accept a ref that can be undefined. Previously, the ref was forced to be non-nullable, which required developers to use workarounds (like non-null assertions) even though React handles null/undefined refs gracefully. With React 19, the framework ensures that refs are managed correctly regardless of their empty initial state.
> 
> By allowing n undefined ref, we align the type definitions with React's natural behavior and improve the developer experience by eliminating unnecessary TypeScript errors. This change has no runtime impact, as the component lifecycle guarantees that the ref will be properly set before it is used.
> 
> Example that trigger the Typescript Error below:
> 
> ```
> const searchBarRef = useRef<SearchBarCommands>(undefined); // or null
> 
> useLayoutEffect(() => {
>   navigation.setOptions({
>     headerSearchBarOptions: {
>       tintColor: colors.text,
>       ref: searchBarRef,
>     } satisfies SearchBarProps,
>   });
> }, [navigation, colors.text]);
> ```
> 
> Type 'RefObject<SearchBarCommands | undefined>' is not assignable to type 'RefObject'. Type 'SearchBarCommands | undefined' is not assignable to type 'SearchBarCommands'. Type 'undefined' is not assignable to type 'SearchBarCommands'.ts(2322)

**Comment by Pnlvfx** [comment]
Test passed successfully.

**Comment by kkafar** [comment]
hey! thanks for the PR. I took a quick glance I'm not convinced on backward compat of your solution with react 18 yet.
I'll try to find some time this week to test this.

**Comment by Pnlvfx** [comment]
> hey! thanks for the PR. I took a quick glance I'm not convinced of backward compat of your solution with react 18 yet. I'll try to find some time this week to test this.

Ehy no problem. Do you want me to do it using null instead of undefined? I usually use undefined with imperative ref as I prefer it, but yes, maybe react 18 doesn't support undefined, but should definetely support null. If you want I can do it, actually all react 19 users will have this issue, and as far as I know the only way to fix it is provide a nullable (initial) value to the ref.

File: src/types.tsx [hunk 1]
```diff
@@ -687,7 +687,7 @@ export interface SearchBarProps {
    * * `cancelSearch` - cancel search in search bar.
    * * `toggleCancelButton` - depending on passed boolean value, hides or shows cancel button (iOS only)
    */
-  ref?: React.RefObject<SearchBarCommands>;
+  ref?: React.RefObject<SearchBarCommands | undefined>;
```
sharplet, line 690: This should really use `React.Ref<SearchBarCommands>`, which takes care of all the different shapes of react ref that could be passed here.

Pnlvfx: this or mine are both good,the actually used for me is wrong!

kkafar:
This will be done here: https://github.com/software-mansion/react-native-screens/pull/3509

Thanks!

PR #2730, opened 2025-02-21
chore: restore backward compat for new & old architecture

kkafar:
## Description

Restoring backward compatibility for old architecture - now we require 76. 75 could work, but I've not tested this. 

On new architecture we require 77. It should compile on 76, however we need the fix for the removal of transitioning views that landed in 77. See here: https://github.com/software-mansion/react-native-screens/pull/2596.

## Changes

Described mostly down below 👇🏻 in review comments.

## Checklist

I've tested this manually on fresh RN 76 & 78 app, new/old architecture, iOS/Android.

- [ ] Ensured that CI passes


android/src/main/java/com/swmansion/rnscreens/bottomsheet/DimmingView.kt
```diff
@@ -17,13 +16,24 @@ import com.swmansion.rnscreens.ext.equalWithRespectToEps
...
 ) : ViewGroup(context),
     ReactCompoundViewGroup,
-    ReactPointerEventsView {
-    private val blockGestures
+    ReactPointerEventsView by pointerEventsProxy {
+
+    constructor(context: Context, initialAlpha: Float = 0.6F) : this(
+        context, initialAlpha,
+        DimmingViewPointerEventsProxy(null)
+    )
```
kkafar, line 30: Doing it it such convoluted way, because I want the interface implementation to be delegated, but the implementation requires reference to `this@DimmingView` which can not be passed while calling other constructor.

This should not cause any timing problems since I'm initialising the implementation field just below in the `init` block.

Delegation of the interface implementation allows me to not put whole logic under versioning but rather only its fragments related to pointer events.

android/src/main/java/com/swmansion/rnscreens/bottomsheet/GestureTransparentViewGroup.kt
kkafar, line 1: Unused code that would required to be adjusted similarly to DimmingView & ScreensCoordinatorLayout.

ios/utils/RNSDefines.h
```diff
@@ -5,3 +5,18 @@
...
 #define RNS_IGNORE_SUPER_CALL_END _Pragma("clang diagnostic pop")
+
+#if defined __has_include
+#if __has_include( \
+    <React-RCTAppDelegate/RCTReactNativeFactory.h>) // added in 78
+#define RNS_REACT_NATIVE_VERSION_MINOR_BELOW_78 0
+#else
+#define RNS_REACT_NATIVE_VERSION_MINOR_BELOW_78 1
+#endif
+#endif
```
kkafar, line 16: This is not pretty, but good enough for now. 

If we need to have more RN version dependent code in the future we should define `RNS_REACT_NATIVE_VERSION_MINOR` during compilation or something. 

I've also seen that `REACT_NATIVE_VERSION_MINOR` macro is defined in `ReactNativeVersion.h` since RN 0.79.

scripts/codegen-utils.js
```diff
@@ -83,6 +83,40 @@ function fixOldArchJavaForRN72Compat(dir) {
...
+        );
+
+        console.log(' => fixOldArchJava77 applied to:', filePath);
+        fs.writeFileSync(filePath, newFileContent, 'utf-8');
+      }
+    } else if (fs.lstatSync(filePath).isDirectory()) {
+      fixOldArchJavaForRN77Compat(filePath);
+    }
+  });
+}
```
kkafar, line 118: 78 introduced `ViewManagerWithGeneratedInterface` that codegened interfaces extend.

This causes issues on RN versions <= 77. However it seems to not be required for Paper to work properly on 78, therefore I decided to remove it.

kkafar (commented):
Notes for future me.

kkafar:
e2e fails because atm reanimated v4 is set in example apps (as there is no reanimated v3 with support for 78 yet) and it does not work on old architecture.

Issue #2726, opened 2025-02-21
[react-native-screens] No member named 'parentShadowView' in 'facebook::react::ShadowViewMutation' (Build Failure)


shubhanshubb:
### Description

I am encountering a build error when trying to compile my React Native project with react-native-screens. The error occurs in the RNSScreenRemovalListener.cpp file inside the react-native-screens module.

### Steps to reproduce

Steps to Reproduce:
Create a new React Native project or use an existing one.

npx @react-native-community/cli init MovieApp
cd MovieApp
Install react-native-screens.

yarn add @react-navigation/native @react-navigation/bottom-tabs
yarn add react-native-screens react-native-safe-area-context react-native-vector-icons react-native-gesture-handler react-native-reanimated

cd ios
pod-install

yarn start
The build fails with the above error.

Open Xcode tried to build the project

### Snack or a link to a repository

shubhanshubb/MovieApp

### Screens version

4.8.0

### React Native version

0.78.0

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

Iphone 16

### Acknowledgements

Yes

kkafar:
4.8.0 does not support RN 0.78. Only version that supports it right now is 4.9.0-beta.0.

shubhanshubb:
@kkafar so which version react native satble now so I don't get any error as in RN 0.77 also same issue is there

kkafar:
when running on RN 0.77 you **can't** get this error, because `parentShadowView` member was removed from mutation types in RN 0.78

shubhanshubb:
Yes, you are correct with RN 0.77 , It's working fine

kkafar:
Would you ming opening `RNSDefines.h` and taking a look at what definition of `MUTATION_PARENT_TAG` macro is chosen?

irmakcosarsahna:
same problem rn v0.78

kkafar:
@irmakcosarsahna could you provide me with information I've asked for above?

irmakcosarsahna:
@kkafar  I solved the problem using [4.9.0-beta.0](https://www.npmjs.com/package/react-native-screens/v/4.9.0-beta.0), but I downgraded react native version because many packages do not support version 78 :(

kkafar:
This is not really a solution. 

[...]

Issue #2716, opened 2025-02-19
iOS compile failes on react native 0.78.0


jonluca:
### Description

When attempting to install react-native-screens in a new react-native project we're seeing a compilation error

- no member named 'parentShadowView' in 'facebook::react::ShadowViewMutation'
Refer to "Xcode Logs" below for additional, more detailed logs.

```
CompileC /Users/expo/Library/Developer/Xcode/DerivedData/Weights-dqqzqcrfobomtkdpsjeqmpbzsjar/Build/Intermediates.noindex/ArchiveIntermediates/Weights/IntermediateBuildFilesPath/Pods.build/Release-iphoneos/RNScreens.build/Objects-normal/arm64/RNSScreenRemovalListener.o /Users/expo/workingdir/build/node_modules/react-native-screens/cpp/RNSScreenRemovalListener.cpp normal arm64 c++ com.apple.compilers.llvm.clang.1_0.compiler (in target 'RNScreens' from project 'Pods')
```

### Steps to reproduce

Attempt to build with react native 0.78.0 on v4.8.0

### Snack or a link to a repository

https://expo.dev/accounts/weights/projects/weights/builds/0922c63e-e6cb-42df-99ad-0a01faf27ea7

### Screens version

4.8.0

### React Native version

0.78.0

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

kkafar:
Hey, support for 0.78 comes with 4.9.0 (currently 4.9.0-beta.0 is released).

Issue #2594, opened 2024-12-30
No member named 'parentShadowView' in 'facebook::react::ShadowViewMutation'


mrousavy:
### Description

I'm trying to build react-native 0.78 nightly, but react-native-screens doesn't seem to support RN 78 yet because there were some changes to Yoga/Fabric APIs which RNScreens seems to rely on:

```diff
- No member named 'parentShadowView' in 'facebook::react::ShadowViewMutation'
```

![image](https://github.com/user-attachments/assets/51cf4b9f-d123-4a1c-849c-6f8ebd4a129a)

### Steps to reproduce

1. Create a new react native app with RN 78 nightly / main
2. Install react-native-screens
3. Build

### Snack or a link to a repository

https://github.com/mrousavy/nitro/pull/454

### Screens version

4.4.0

### React Native version

0.78.0@nightly

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Acknowledgements

Yes

mrousavy:
Same also happens here in `RNSScreenStack.mm` line 1220:

```diff
- No member named 'parentShadowView' in 'facebook::react::ShadowViewMutation'
```

![Screenshot 2024-12-30 at 13 47 00](https://github.com/user-attachments/assets/e03e2cad-03b1-48fe-a448-ea9cd0cf7010)


But I think this can be replaced with `parentTag`.

kkafar:
Thanks for reporting this. We won't be working on 0.78 support for at least few more weeks now, but this is great heads up. Thanks

kkafar:
I've merged https://github.com/software-mansion/react-native-screens/pull/2595 into PR with 0.78 support. I'll close the issue as this is not a "regular issue".

gagtic:
> ### Description
> 
> I'm trying to build react-native 0.78 nightly, but react-native-screens doesn't seem to support RN 78 yet because there were some changes to Yoga/Fabric APIs which RNScreens seems to rely on:
> 
> - No member named 'parentShadowView' in 'facebook::react::ShadowViewMutation'
> ![image](https://github.com/user-attachments/assets/51cf4b9f-d123-4a1c-849c-6f8ebd4a129a)
> 
> ### Steps to reproduce
> 
> 1. Create a new react native app with RN 78 nightly / main
> 2. Install react-native-screens
> 3. Build
> 
[...]

PR #2736, opened 2025-02-24
Fix MUTATION_PARENT_TAG to Handle Parent Tag Across React Native Versions

AliRezaBeigy:
## Description

The old `MUTATION_PARENT_TAG` code couldn’t always find the right property (`parentTag` or `parentShadowView.tag`) because its version check wasn’t reliable across setups. This PR fixes that by switching to a smarter way to pick the correct property without needing version checks, making it work for all React Native versions (before and after 78).

## Changes

- Replaced the version-based `#if` check with a new `get_parent_tag` function that automatically uses `parentTag` or `parentShadowView.tag` based on what’s available.

## Screenshots / GIFs

This is the previous state of the code in React Native 78.

## Related Issue
https://github.com/software-mansion/react-native-screens/issues/2726
https://github.com/software-mansion/react-native-screens/issues/2716
https://github.com/software-mansion/react-native-screens/issues/2594

@kkafar

### DIFF
--- a/ios/utils/RNSDefines.h
+++ b/ios/utils/RNSDefines.h
@@ -6,20 +6,28 @@
 
 #define RNS_IGNORE_SUPER_CALL_END _Pragma("clang diagnostic pop")
 
-#if defined __has_include
-#if __has_include(<React-RCTAppDelegate/RCTReactNativeFactory.h>) ||\
-    __has_include(<React_RCTAppDelegate/RCTReactNativeFactory.h>) // added in 78; underscore is used in dynamic frameworks
-#define RNS_REACT_NATIVE_VERSION_MINOR_BELOW_78 0
-#else
-#define RNS_REACT_NATIVE_VERSION_MINOR_BELOW_78 1
-#endif
-#else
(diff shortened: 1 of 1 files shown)


kkafar:
I would be happy to work on the issue, however I need a repro first. You say that current solution does not work in certain setups - what are these setups? Could you provide me with enough information so that I can reproduce it?

kkafar:
I'd recommend you check out whether these changes do not solve the issue in your case: https://github.com/software-mansion/react-native-screens/pull/2737

AliRezaBeigy:
my bad! I see what you meant now. I was only testing the case where parentTag existed, so I totally missed the issue when it wasn’t there. I’ve fixed it by using std::void_t properly, hopefully, it should be all good now.

I also checked your latest revision, and it looked like it worked fine too. So whatever you think makes the most sense is cool with me.

Thanks for catching that and for the review! 😅 Appreciate it.

kkafar:
@AliRezaBeigy big thanks for bringing this issue up to my attention & the PR.

For the sake of keeping thing maintainable I'll go with #2737 for now. 

Thanks a lot!

PR #2754, opened 2025-03-04
refactor(iOS): split big classes into multiple files

kligarski:
WIP


ios/RCTConvert+RNSScreen.mm
```diff
@@ -0,0 +1,149 @@
...
+#import "RNSScreenStackView.h"
+
+#import "RNSDefines.h"
+#import "UIView+RNSUtility.h"
+
+#ifdef RCT_NEW_ARCH_ENABLED
+namespace react = facebook::react;
+#endif // RCT_NEW_ARCH_ENABLED
+
+@implementation RCTConvert (RNSScreen)
```
kligarski, line 41: I moved RCTConvert+RNSScreen from RNSScreen, but some implementation is also in RNSSearchBar.mm - I'm not quite sure how to handle it. Also, I considered splitting RNSSearchBar into seperate files as well but decided to wait before I know how to handle the RCTConvert.

PR #2756, opened 2025-03-06
fix(Android): adapt header insets when interface orientation changes

kkafar:
## Description

We did not adapt header insets when interface orientation changes, leading to cases like on screenshots below 👇🏻 

image

image

We want it to adapt to changing window insets properly. 

What needs to be tested / answered:

1. [x] Pressables in header - insets change is not observed by ShadowTree -> ~this might lead to issues with pressables~ - checked on `Test2466` in e2e and not e2e - seems to work ok
2. [x] String truncation - tested on `TestHeaderTitle` - seems to work ok (tested both e2e and not e2e & in any screen orientation)

This is how it works now (look only at header):

### SDK 34 (no edge-to-edge) + no cutout

https://github.com/user-attachments/assets/347d0aaf-fef2-4689-98f8-5def8264c4cc

### SDK 34 (no edge-to-edge) + cutout

https://github.com/user-attachments/assets/6a9fd9c6-aaa4-4c92-a0b4-f4e96d1e245f

### SDK 35 (edge-to-edge) + no cutout

https://github.com/user-attachments/assets/75e4b4af-d504-48c8-86d3-e9ba603f88cf

### SDK 35 (edge-to-edge) + cutout

https://github.com/user-attachments/assets/6254f209-ad8d-4188-81e1-5d499b69adb4

---

Let it be noted that I've also tested that it work's fine with button navigation.

## Changes

I've moved all inset/padding related logic to `CustomToolbar`, because it is in different place in view hierarchy than `HeaderConfig` & the insets it receives are sometimes different & I wanted the logic manipulating `CustomToolbar` layout to live in `CustomToolbar`.

`CustomToolbar` now listens for dispatches of window insets & updates its horizontal paddings accordingly.

Currently I've decided that this behaviour will be controlled by `isTopInsetEnabled` prop on `HeaderConfig` together with the topInset.

1. Use device with cutout / add cutout to emulator (in developer options)
2. Run any example with header, e.g. `TestFormSheet`
3. Change screen orientation & observe the header behaviour.

## Checklist

- [ ] Ensured that CI passes


kkafar:
There is stil issue when the app start s in landscape mode -> the header font is weirdly small, but this is not a regression and a separate problem. I'll tackle it in separate PR.

kligarski (approved):
Looks good!

PR #2760, opened 2025-03-07
fix(CI): update Example, bump reanimated

kligarski:
## Description

Fixes CI on both platforms.

## Changes

- change Example to use tab bar button test ID option available in react-navigation 6 (previous one was correct for react-navigation 7 but example apps use version 6)
- bump to `react-native-reanimated@3.17.2` in order to fix the issue with Android CI (https://github.com/software-mansion/react-native-reanimated/pull/7196)

CI

## Checklist

- [x] Ensured that CI passes


kligarski:
@kkafar

kkafar (approved):
Looks solid. Thanks a lot.

PR #2766, opened 2025-03-09
fix(iOS) pushing a removing screen to controller

### LINKED ISSUE #2559: [iOS] Unexpected Navigation Behavior When Swiping Back Quickly
### Description

When using iOS devices with the following dependencies, swiping back quickly during stack navigation leads to unexpected behavior:

Expected behavior: stack a -> stack b -> stack c -> stack b -> stack a
Actual behavior: stack a -> stack b -> stack c -> stack b -> stack a -> stack b -> stack a
The issue occurs only when swiping back quickly. When swiping back slowly, the navigation works as expected.

## Expected Behavior

https://github.com/user-attachments/assets/ed515fe7-b876-4b45-b1c0-e036575bb62a

## Actual Behavior

https://github.com/user-attachments/assets/017bf04d-14dd-49dd-8910-c1f79327c54d

### Steps to reproduce

1. Enable fullScreenGestureEnabled in the navigation options.
2. Set up stack navigation with three screens: stack a, stack b, and stack c.
3. Navigate through the stack in order: stack a -> stack b -> stack c.
4. Swipe back quickly from stack c to stack b and observe the behavior.
5. Swipe back quickly from stack c to stack b, and then quickly swipe back again from stack b to stack a.
6. Observe the unexpected behavior.

https://github.com/Unuuuuu/react-native-screens-repro

### Screens version

4.3.0

### React Native version

0.76.3

### Platforms

iOS

### JavaScript runtime

None

### Workflow

Expo managed workflow

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

Real device

### Device model

iphone 15 pro

### Acknowledgements

Yes

wvq:
## Description

Fixes #2559 

Bug track:
The logic waiting for the transition complete and then call `updateContainer` has problem. 
An old screen may be about to be removed but has not yet been removed.
which not in _controllers, but _reactSubviews not remove it yet.
`updateContainer` iterates _reactSubviews and call `setPushViewControllers`.
The old screen will push again.

This change add a property **`disappeared`** to RNSScreen, defaults to `NO`.
- on `viewWillAppear`/`viewDidAppear`/`viewWillDisappear` set `NO`
- on `viewDidDisappear` set `YES`

In `RNSScreenStack`'s `setPushViewControllers`,  check if it's not an new screen.

``` typescript
if(!((RNSScreen *)top).disappeared) {
  [_controller pushViewController:top animated:YES];
}
```

### DIFF
--- a/ios/RNSScreen.h
+++ b/ios/RNSScreen.h
@@ -40,6 +40,7 @@ namespace react = facebook::react;
 - (void)calculateAndNotifyHeaderHeightChangeIsModal:(BOOL)isModal;
 - (void)notifyFinishTransitioning;
 - (RNSScreenView *)screenView;
+@property (nonatomic, readonly) BOOL disappeared;
 #ifdef RCT_NEW_ARCH_ENABLED
 - (void)setViewToSnapshot;
 - (CGFloat)calculateHeaderHeightIsModal:(BOOL)isModal;
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -1383,6 +1383,7 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
     self.view = view;
     _fakeView = [UIView new];
     _shouldNotify = YES;
+    _disappeared = NO;
 #ifdef RCT_NEW_ARCH_ENABLED
     _initialView = (RNSScreenView *)view;
 #endif
@@ -1394,6 +1395,7 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
 - (void)viewWillAppear:(BOOL)animated
 {
   [super viewWillAppear:animated];
+  _disappeared = NO;
   if (!_isSwiping) {
     [self.screenView notifyWillAppear];
     if (self.transitionCoordinator.isInteractive) {
@@ -1421,6 +1423,7 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
 - (void)viewWillDisappear:(BOOL)animated
 {
   [super viewWillDisappear:animated];
+  _disappeared = NO;
   // self.navigationController might be null when we are dismissing a modal
   if (!self.transitionCoordinator.isInteractive && self.navigationController != nil) {
     // user might have long pressed ios 14 back button item,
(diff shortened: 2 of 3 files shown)


kkafar (commented):
Hey, that looks great, thanks. I'll test that when I have a moment & we'll see how do we proceed here.

wvq:
@kkafar  This fix  conflicts with `react-navigation`'s `usePreventRemove`, should not be approved.
The logic of `usePreventRemove` is to add disappeared screen again, 
This bug is also because add disappeared screen again.
In this repo, there is no way to check if the screen is disappeared and should't add it.
It's need `react-navigation` passing some neccessary information to check it.

JustJoostNL:
@wvq So do I understand correctly that, to fix the issue, changes in react-navigation are needed rather than in react-native-screens?

If so, are you planning to do a PR there? I would love to see this fixed. 😄

wvq:
@JustJoostNL No, the bug is on this project, I just can't find a proper way to mark the expired screen.
my PR is marking expired screen and prevent it add again,  it work for my case, not for all.
react-navigation usePreventRemove just an example which not suite.
So I close this PR.

wvq:
For now, my temporary solution is run hook after npm install.

```bash
#!/bin/bash

#
# temporary fix react-native-screens
# 
# https://github.com/software-mansion/react-native-screens/issues/2559
#

check='disappeared'

screen_header='./node_modules/react-native-screens/ios/RNSScreen.h'
screen_mm='./node_modules/react-native-screens/ios/RNSScreen.mm'
screen_stack_mm='./node_modules/react-native-screens/ios/RNSScreenStack.mm'


if ! grep -qw $check $screen_header; then
    sed -i '' '43i\
@property (nonatomic, readonly) BOOL disappeared;\
' $screen_header
fi

if ! grep -qw '_disappeared' $screen_mm; then
sed -i '' '1353i\
    _disappeared = NO;\
' $screen_mm
sed -i '' '1365i\
  _disappeared = NO;\
' $screen_mm
sed -i '' '1393i\
  _disappeared = NO;\
' $screen_mm
sed -i '' '1431i\
  _disappeared = NO;\
' $screen_mm
sed -i '' '1448i\
  _disappeared = YES;\
' $screen_mm
fi


if ! grep -qw $check $screen_stack_mm; then
    sed -i '' '680i\
      if(!((RNSScreen *)top).disappeared) {\
' $screen_stack_mm
sed -i '' '682i\
      }\
' $screen_stack_mm
fi

echo 'hook finished'
```

wvq:
This bug is caused by asynchrony, I think some logic may need to be refactored, which is not something that a simple solution can handle.

PR #2767, opened 2025-03-10
fix(Android): `formSheet` not visible when going back from `push` in parent stack

kkafar:
## Description

1. We have two stacks A & B. 
2. B is nested in A. 
3. We open a Sheet in B.
4. We navigate from the sheet to a `push` screen in A.
5. We execute (back button click) `navigation.goBack()`.

Notice that the see is not visible (but the dimming view is).

Problem: at step 4. the sheet is dismissed & its exit animation plays (not visible, but it plays out)
affecting its `translateY` property. When we remount the `Screen` we recycle the component, but in no place
we reset the `translateY` -> the sheet is not visible.

## Changes

`View.recycle()` extension method resets `translateY` now.

## Recordings

| before | after |
| -- | -- |
| <video src="https://github.com/user-attachments/assets/d80c25bf-5ae9-436f-932b-484c8fbf9876" alt="before" /> | <video src="https://github.com/user-attachments/assets/70d2d12a-a590-4464-b082-1fa0f18b31d2" alt="after" /> |

Added `Test2767` to test this scenario (and other navigation scenarios including sheet).

## Checklist

- [ ] Ensured that CI passes


kligarski (commented):
Great catch!

PR #2781, opened 2025-03-14
fix(Android,Paper): header config blocks gestures close to the top of the screen

### LINKED ISSUE #2758: ScreenStackHeaderConfig masks the SceneScreen area when customizing the header component
### Description

customizing the header component `headerLeft` or `headerCenter` or `headerRight` .

Image

### Steps to reproduce

1. Customizing the header component `headerLeft` or `headerCenter` or `headerRight` .

https://github.com/b1ngx/rns-issue

### Screens version

4.9.1

### React Native version

0.78.0

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

newArchEnabled=false

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

kkafar:
## Description

Fixes #2758

#2466 removed old workaround for header config blocking gestures - we just set top: `-100%` 
to place the headerconfig in the top of the screen effectively preventing blocking. 

#2466 removed these styles & frame correction is now applied directly in shadow node. This is done fine
on Fabric, however the solution was not replicated on Paper.

## Changes

Beside padding information we now send native toolbar height to header config shadow node on Paper.
This information is used there to offset the header config position by this value.

Should solve the problem.

### Screenshots

| before | after |
| - | - |
| before | after |

`Example` -> `Header options` -> click on the very top button (very top of it). Previously the button was not effectively 
clicked. Now it works.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/fabric/java/com/swmansion/rnscreens/FabricEnabledHeaderConfigViewGroup.kt
+++ b/android/src/fabric/java/com/swmansion/rnscreens/FabricEnabledHeaderConfigViewGroup.kt
@@ -23,24 +23,20 @@ abstract class FabricEnabledHeaderConfigViewGroup(
         mStateWrapper = wrapper
     }
 
-    fun updatePaddings(
-        paddingStart: Int,
-        paddingEnd: Int,
-    ) {
-        // Do nothing on Fabric. This method is used only on Paper.
-    }
-
     fun updateHeaderConfigState(
         width: Int,
         height: Int,
         paddingStart: Int,
         paddingEnd: Int,
     ) {
+        // Implementation of this method differs between Fabric & Paper!
+        Log.i("ConfigShadowNode", "FABRIC updateHeaderConfigState")
         updateState(width, height, paddingStart, paddingEnd)
     }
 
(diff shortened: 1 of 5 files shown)


b1ngx:
`headerShown: false` still has this issue.

kkafar:
@b1ngx thanks, looking into it

kkafar:
@b1ngx I'm not sure whether this is a bug. The only behaviour I reproduce is as follows:

https://github.com/user-attachments/assets/db2c0d9f-1c74-40c4-852e-2c874d5cfae9
(old architecture)

All clicks are accepted, except ones that go through status bar + this behaviour is consistent across old & new architecture. 

Even if I disable the header config completely, the area covered by status bar is still not clickable. 

Ping me, in case you meant some different behaviour.

b1ngx:
repo: https://github.com/b1ngx/rns-issue

image

kkafar:
That's great, thanks! I'll look into it shortly

PR #2785, opened 2025-03-18
fix(CI): change order of events in Fabric iOS events e2e test

kligarski:
## Description

Changed order of events in Fabric iOS events e2e test.

The order of events in this test differs from the Paper version of the same test. When screen is removed after using iOS native back button, native side notifies JS via:

1. onDismissed => in reaction to this event, a navigation action is dispatched, which causes emission of beforeRemove event;
2. onDisappear => in reaction to this event, transitionEnd with closing=true is emitted.

On Fabric, these events are handled in the same order as the order of notifications from the native side. On Paper, this is not the case - onDismissed is executed after two transitionEnd events are already handled. onDismissed is dispatched asynchronously from the native side (it is wrapped in `dispatch_async`!) and this probably delays handling it on JS side but I am not sure of the details why it is dispatched asynchronously.

Currently this divergence between architectures seems to not lead to any issues. Just let it be noted & not forgotten. 

## Changes

- change order of assertions
- add comment with explanation

Run Fabric iOS CI.

## Checklist

- [x] Ensured that CI for Fabric iOS passes

### DIFF
--- a/FabricExample/e2e/examplesTests/events.e2e.ts
+++ b/FabricExample/e2e/examplesTests/events.e2e.ts
@@ -13,14 +13,7 @@ const pressBack = async () => {
 const awaitClassicalEventBehavior = async () => {
   if (device.getPlatform() === 'ios') {
     // The order of events in this test differs from the Paper version of the same test.
-    // When screen is removed after using iOS native back button, native side notifies JS via:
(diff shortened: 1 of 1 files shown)


kkafar (approved):
Thanks a lot!

Issue #263, opened 2020-01-03
Collapsible header / modal swipe gesture on inverted ScrollView


Hirbod:
If you have a ScrollView with inverted property, the LargeTitleHeader behaves accordingly. 

If the ScrollView is set to inverted (often the case with chats where messages appear at the bottom) and the whole thing takes place in a modal, you have to swipe backwards to close the modal. This is not an error in itself, but there should be the possibility or a prop for these cases to explicitly reverse the behavior.

WoLewicki:
Can you create a repo with the minimal configuration that shows the issue?

Hirbod:
Just open a modal in native-stack with a <FlatList inverted={true}> and try to close the modal by swiping down. It wont work, since the direction is reversed. So I need to slide up to move the modal down.

(The FlatList ofc needs to have a height bigger then the viewport)

Hirbod:
@WoLewicki this is how facebook fixed it for the status bar tab:
https://github.com/facebook/react-native/commit/7a4753d76aab1c52a09f26ec6f7fd43a68da8a97

it would be nice if you guys could also implement in the same fashion so a modal, which contains a inverted Scrollview or FlastList, works as expected (and not reversedd)

berkaey:
Yes, this is a real issue for chat views inside modals on iOS 13. Any suggested workarounds for this type of use case. Do we even stop propagation between nested scroll views?

berkaey:
@Hirbod did you find any solution?

Hirbod:
There is no solution currently @berkaey. react-native-screens needs to fix this. I would love to help but my obj-c/ios native development skills are super limited. I think the listener values for the swipe down gesture have to be reversed (or listen for negative)

berkaey:
Can we implement something similar to this https://stackoverflow.com/a/59462475? @WoLewicki Otherwise, we and probably many more can't use native modal patterns in a chat app or something similar. We need a programmatic control of the gesture recognizers somehow.

WoLewicki:
It might not be possible to change the direction of dragging to dismiss a modal from the native side. If you find an example where it is done on the native side, please share an example.
About disabling the gesture on the modal and the thread from SO you sent @berkaey, you can use RNGH to catch the gesture events. You can look at the gist:  https://gist.github.com/WoLewicki/7a6f5ceebc228ad5ac8aaeb7bd5a6a36 where the box takes over the gestures and stops the screen from dismissing.

Hirbod:
[...]

PR #2787, opened 2025-03-18
test(e2e,Fabric): add e2e tests for issue/PR examples 42..528

kligarski:
## Description

Check which example screens from issues/PRs can be used in e2e testing for tests `Test42, ..., Test528` and implement them if possible for Fabric. In the future, each non-skipped test will be in a seperate PR.

### Test42
Skipped because we can’t really check what is the orientation of the screen with Detox.

### Test111
Skipped because we can’t test flickering with Detox.

### Test263
Skipped because example differs from the problem mentioned in https://github.com/software-mansion/react-native-screens/issues/263. Even when I recreated the issue, it still occurs. [This comment from the PR](https://github.com/software-mansion/react-native-screens/issues/263#issuecomment-930127450) explains why it is the case. 

### Test349
Skipped because we can’t test autofill with Detox. The test also uses outdated prop name.

### Test364
Skipped because the test uses `tabBarVisible` prop that no longer exists. `react-navigation` documentation suggests other [solution](https://reactnavigation.org/docs/hiding-tabbar-in-screens/) to this problem.

### Test 432
Test created. It checks whether correct squares are fully visible. For some reason, on Android, these squares are not `100%` visible from the start (but I don’t see the problem in manual testing).  For the test, `waitingFor` the squares to become `100%` visible works.

### Test 528
Test created but only for iOS because Detox supports changing orientation of the device [only on iOS](https://wix.github.io/Detox/docs/api/device/#devicesetorientationorientation). Test checks if the headerRight button is visible after changing screens/orientation.

## Changes

- add `Test432` and `Test528`
- add comments for every test screen in `apps/src/tests/index.ts` with the reason for (not) implementing e2e test for it

CI

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/FabricExample/Gemfile.lock
+++ b/FabricExample/Gemfile.lock
@@ -95,6 +95,7 @@ PLATFORMS
 DEPENDENCIES
   activesupport (>= 6.1.7.5, != 7.1.0)
   cocoapods (>= 1.13, != 1.15.1, != 1.15.0)
+  xcodeproj (< 1.26.0)
 
 RUBY VERSION
    ruby 3.2.1p31
--- a/FabricExample/e2e/examplesTests/tests.e2e.ts
+++ b/FabricExample/e2e/examplesTests/tests.e2e.ts
@@ -1,7 +1,7 @@
 import { device } from 'detox';
 
-describe('E2E', () => {
-  beforeAll(async () => {
+describe('Tests from PRs and issues', () => {
+  beforeEach(async () => {
     await device.reloadReactNative();
   });
 
--- a/FabricExample/e2e/examplesTests/tests.e2e.ts
+++ b/FabricExample/e2e/examplesTests/tests.e2e.ts
@@ -9,19 +9,66 @@ describe('Test432', () => {
     await waitFor(element(by.id('root-screen-tests-Test432')))
     .toBeVisible()
     .whileElement(by.id('root-screen-examples-scrollview'))
-    .scroll(100, 'down', NaN, 0.85);
+    .scroll(600, 'down', NaN, 0.85);
 
     await expect(element(by.id('root-screen-tests-Test432'))).toBeVisible();
     await element(by.id('root-screen-tests-Test432')).tap();
   });
 
+  it('home-square should be fully visible', async () => {
+    await expect(element(by.id('home-square'))).toBeVisible(100);
+  });
 
-  // home home-square
-  // details details-red-square | click | details-green-square | click | details-red-square | back
-  // info info-green-square-1 | click | info-red-square info-green-square-1 info-green-square-2 | click | info-green-square-1 | back
-  // settings settings-square
-  // za kazdym powrotem do home sprawdzic czarny square
+  it('squares from details screen should be fully visible', async () => {
+    await element(by.id('go-to-details')).tap();
+    await expect(element(by.id('details-red-square'))).toBeVisible(100);
 
+    await element(by.id('details-toggle-subviews')).tap();
+    // On Android, we need to wait for some elements (e.g. at first, this square is only 25% visible)
+    waitFor(element(by.id('details-green-square'))).toBeVisible(100);
 
+    await element(by.id('details-toggle-subviews')).tap();
+    await expect(element(by.id('details-red-square'))).toBeVisible(100);
 
+    if (device.getPlatform() === 'ios') {
+      await element(by.id('BackButton')).tap();
+    } else {
+      await device.pressBack();
+    }
+
+    await expect(element(by.id('home-square'))).toBeVisible(100);
+  });
+
+  it('squares from info screen should be fully visible', async () => {
+    await element(by.id('go-to-info')).tap();
+    await expect(element(by.id('info-green-square-1'))).toBeVisible(100);
+
+    await element(by.id('info-toggle-subviews')).tap();
+    waitFor(element(by.id('info-green-square-1'))).toBeVisible(100);
+    waitFor(element(by.id('info-green-square-2'))).toBeVisible(100);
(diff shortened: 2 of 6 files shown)


FabricExample/e2e/examplesTests/tests.e2e.ts
```diff
@@ -0,0 +1,111 @@
+import { device, expect, element, by } from 'detox';
+
+describe('Test432', () => {
+  beforeAll(async () => {
+    await device.reloadReactNative();
+  });
+
+  it('should Test432 exist', async () => {
```
kkafar, line 8:
```suggestion
  it('Test432 should exist', async () => {
```

FabricExample/e2e/examplesTests/tests.e2e.ts
```diff
@@ -0,0 +1,111 @@
...
+    await waitFor(element(by.id('root-screen-tests-Test432')))
+    .toBeVisible()
+    .whileElement(by.id('root-screen-examples-scrollview'))
+    .scroll(600, 'down', NaN, 0.85);
+
+    await expect(element(by.id('root-screen-tests-Test432'))).toBeVisible();
+    await element(by.id('root-screen-tests-Test432')).tap();
+  });
+
+  it('home-square should be fully visible', async () => {
```
kkafar, line 18:
```suggestion
  it('headerRight should be fully visible', async () => {
```

I think we should change naming convention to be more role-specific. This is something I'm coming up as I write this message, so I'm very open for discussion. 

What I mean is - instead of `home-square` we should describe the role of the element in contention (is that a good word here? 😄). For example: `headerRight element should be fully visible`. That makes it much easier to reason about what goes on in the test only by skimming its description.

FabricExample/e2e/examplesTests/tests.e2e.ts
```diff
@@ -0,0 +1,111 @@
...
+    await expect(element(by.id('home-square'))).toBeVisible(100);
+  });
+
+  it('squares from details screen should be fully visible', async () => {
```
kkafar, line 22:
```suggestion
  it('headerRight elements should stay visible after toggling', async () => {
```

Same here & in many other places. I'm certain you see the pattern.

FabricExample/e2e/examplesTests/tests.e2e.ts
```diff
@@ -0,0 +1,111 @@
...
+    await element(by.id('go-to-details')).tap();
+    await expect(element(by.id('details-red-square'))).toBeVisible(100);
+
+    await element(by.id('details-toggle-subviews')).tap();
```
kkafar, line 26:
```suggestion
    await element(by.id('details-button-toggle-subviews')).tap();
```

As for assigning testID I do also think that we should have a clear convention. Open for discussion there. I would suggest: `<where>-<role-of-the-element>-<action>`

* `where` might be relative to the *test* root screen (not app)
* `role-of-the-element` I mean button / view / header item / header left / etc.
* `action` - in case there is any, e.g. what the button click does (if can be put in 2-4 words, otherwise some generic phrase).

kkafar: This applies widely to this PR.

FabricExample/Gemfile.lock
```diff
@@ -95,6 +95,7 @@ PLATFORMS
 DEPENDENCIES
   activesupport (>= 6.1.7.5, != 7.1.0)
   cocoapods (>= 1.13, != 1.15.1, != 1.15.0)
+  xcodeproj (< 1.26.0)
```
kkafar, line 98: I don't think we need that. Let's restore if feasible.

kkafar (changes requested):
Thanks! This looks kinda good initially. I have left few important remarks down below, please answer them.

FabricExample/Gemfile.lock
kligarski: On `main`, in FabricExample's Gemfile `xcodeproj` is specified (https://github.com/software-mansion/react-native-screens/blob/4d044dab8c387c6f3988434276594a2d917f5022/FabricExample/Gemfile#L9C1-L9C28) but in the lock file it is not - after `bundle install` it is added to the lock so I think we should probably leave it.

kkafar: makes sense, thanks for explanation

FabricExample/e2e/examplesTests/tests.e2e.ts
kligarski: 2ed530f9171e919d9ad3722006170f54d0977fd7

FabricExample/e2e/examplesTests/tests.e2e.ts
kligarski: Agreed, it makes sense.

kligarski: 2ed530f9171e919d9ad3722006170f54d0977fd7

FabricExample/e2e/examplesTests/tests.e2e.ts
kligarski: 2ed530f9171e919d9ad3722006170f54d0977fd7

FabricExample/e2e/examplesTests/tests.e2e.ts
kligarski: Makes sense as well.

kligarski: d2afd9a8d2acb7e43f6ad65571025287402cb770

kkafar (approved):
Good job. I think we're good enough now.

PR #2788, opened 2025-03-18
fix(Android,Fabric): pressables losing focus on screens with `formSheet` presentation

kkafar:
## Description

This PR relates to new architecture only.

Currently, the pressables on form sheet lose focus on `move` action. This is the same problem we had with `Pressables` in screen & header on new architecture. 
See: https://github.com/software-mansion/react-native-screens/pull/2466

Basically the information about sheet position is different between `ShadowTree` (ST) and `HostTree` (HT), which leads to losing focus due to how pressables
are now handled on new architecture.

**Simplified description** of basically what happens when you click a pressable on new arch is as follows (Android):

1. The gesture is detected after the host platform dispatches touch event (touch events on Android are dispatched in top-down manner) [(link)](https://github.com/facebook/react-native/blob/d6ca25b0c1a0aeed3507ec3c2f65e453e2700dc2/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/uimanager/JSTouchDispatcher.java#L81-L105)
2. The JS responder is set & touch event dispatched - therefore pressable is always clickable,
3. Moreover, the JS responder is [measured **IN THE ST**](https://github.com/facebook/react-native/blob/d6ca25b0c1a0aeed3507ec3c2f65e453e2700dc2/packages/react-native/Libraries/Pressability/Pressability.js#L805-L810)
3. When you start moving your finger, the platform is still the source of motion events, and target [coordinates are **read from HOST TREE**](https://github.com/software-mansion/react-native-screens/pull/2466)
4. Motion event (with platform measurement (HT)!!!) is then [compared in JS with responder measurements based on ST](https://github.com/facebook/react-native/blob/d6ca25b0c1a0aeed3507ec3c2f65e453e2700dc2/packages/react-native/Libraries/Pressability/Pressability.js#L831-L8750)

Therefore, any inconsistency here leads to responder losing focus.

Reference:

1. [`TouchesHelper.createPointersArray`](https://github.com/facebook/react-native/blob/ac97177651cf369783ca93fac50c2824b484abef/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/uimanager/events/TouchesHelper.kt#L37)
2. [Responder measurement](https://github.com/facebook/react-native/blob/d6ca25b0c1a0aeed3507ec3c2f65e453e2700dc2/packages/react-native/Libraries/Pressability/Pressability.js#L805-L810)
3. [Gesture start detection (Android)](https://github.com/facebook/react-native/blob/d6ca25b0c1a0aeed3507ec3c2f65e453e2700dc2/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/uimanager/JSTouchDispatcher.java#L81-L105)
4. [Gesture move handling (Android)](https://github.com/facebook/react-native/blob/d6ca25b0c1a0aeed3507ec3c2f65e453e2700dc2/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/uimanager/JSTouchDispatcher.java#L137-L147)
5. [Checking whether native touch is in responder region](https://github.com/facebook/react-native/blob/d6ca25b0c1a0aeed3507ec3c2f65e453e2700dc2/packages/react-native/Libraries/Pressability/Pressability.js#L831-L875)

What is bewildering is that it works on iOS, despite the fact, that e.g. when using React inspector the views are completely out of place in ShadowTree. 
I haven't debugged the iOS part to the very bottom, but my suspicion is as follows:

1. The form sheet on Android is mounted in subtree under react root view, while on iOS it is mounted under separate `UITransitionView` (different subtree of window),
2. additionally we use `RNSModalScreen` component on iOS, which has shadow node with `RootNodeKind` (measurements in subtree of its shadow node are done relative to it),
3. There is no root view / any react view above it -> native measurement (done by react-native) might also think that its positioned at (0, 0) (same as in shadow tree).

This is something to verify in the future. Opened ticked on the board for it.

## Changes

The sheet now sends updates to the shadow tree at following moments:

1. layout (including initial one),
2. sheet detent change to any stable state (could consider here not sending the update when the sheet is hidden, added to project board),

I'm not sending updates on every sheet move (`View.top` change), to avoid clogging the JS thread (and later UI), whole advantage of our sheet is that it feels smooth (given sensible Android device).
However, it could be considered in case of any further issues.

Please note, that between the event syncs or in case the JS thread is blocked / clogged, this still will be a problem. However, any JS would lag, not only pressables and it's not down to us.

Commits:

- **Prevent modal content from disappearing**
- **Add comment that createShadowNode is used only on old arch**
- **Add comment on threading on NativeProxy mechanisms**
- **Simplify code in RNSModalScreenShadowNode.cpp**
- **Make modal screen component selection more clear in Screen.tsx**
- **Add `headerHeight` to diff prop list when determining need for shadow state update**
- **Add pressables to form sheet example!**
- **Send updates to shadow tree on form sheet layout changes in appropriate moments**

### Recordings

| before | after |
| -- | -- |
| <video src="https://github.com/user-attachments/assets/b5bbb149-61da-41d1-b6a4-73dd89eb1f92" alt="before" /> | <video src="https://github.com/user-attachments/assets/ce5c89c0-7215-4ebb-a347-fa50946308f4" alt="after" /> |

`TestFormSheet`, open the sheet, play with it & see that pressables work!

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -103,6 +103,9 @@ class Screen(
             field = value
         }
 
+    private val isNativeStackScreen: Boolean
+        get() = container is ScreenStack
+
     init {
         // we set layout params as WindowManager.LayoutParams to workaround the issue with TextInputs
         // not displaying modal menus (e.g., copy/paste or selection). The missing menus are due to the
@@ -173,7 +176,9 @@ class Screen(
         r: Int,
         b: Int,
     ) {
-        if (container is ScreenStack && changed) {
+        // In case of form sheet we get layout notification a bit later, in `onBottomSheetBehaviorDidLayout`
+        // after the attached behaviour laid out this view.
+        if (changed && isNativeStackScreen && !usesFormSheetPresentation()) {
             val width = r - l
             val height = b - t
 
@@ -182,14 +187,21 @@ class Screen(
             } else {
                 updateScreenSizePaper(width, height)
             }
-
-            footer?.onParentLayout(changed, l, t, r, b, container!!.height)
+            // FormSheet has no header in current model.
             notifyHeaderHeightChange(t)
-
-            if (!BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
-                maybeTriggerPostponedTransition()
-            }
         }
+    }
+
+    internal fun onBottomSheetBehaviorDidLayout(coordinatorLayoutDidChange: Boolean) {
+        if (coordinatorLayoutDidChange && usesFormSheetPresentation() && isNativeStackScreen && BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
+            updateScreenSizeFabric(width, height, top)
+        }
+        if (!BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
+            updateScreenSizePaper(width, height)
+            maybeTriggerPostponedTransition()
+        }
+
+        footer?.onParentLayout(coordinatorLayoutDidChange, left, top, right, bottom, container!!.height)
     }
 
     private fun maybeTriggerPostponedTransition() {
@@ -487,6 +499,11 @@ class Screen(
         isStable: Boolean,
     ) {
         dispatchSheetDetentChanged(detentIndex, isStable)
+        // There is no need to update shadow state for transient sheet states -
+        // we are unsure of the exact sheet position anyway.
+        if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED && isStable) {
+            updateScreenSizeFabric(width, height, top)
+        }
     }
 
     private fun dispatchSheetDetentChanged(
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -182,22 +182,21 @@ class Screen(
             val width = r - l
             val height = b - t
 
-            if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
-                updateScreenSizeFabric(width, height, t)
-            } else {
-                updateScreenSizePaper(width, height)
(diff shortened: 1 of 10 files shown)


android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -173,7 +176,9 @@ class Screen(
         r: Int,
         b: Int,
     ) {
-        if (container is ScreenStack && changed) {
+        // In case of form sheet we get layout notification a bit later, in `onBottomSheetBehaviorDidLayout`
+        // after the attached behaviour laid out this view.
+        if (changed && isNativeStackScreen && !usesFormSheetPresentation()) {
```
kkafar, line 181: In case of `formSheet`, when `CoordinatorLayout` layouts it calls to the `BottomSheetBehavior` to layout the child (Screen), which in turn calls `CoordinatorLayout.layoutChild(Screen)` leading to a call to the overridden `onLayout` method here - and this is where we would update the shadow state before. 

This was not good, because at the moment of `onLayout` being called, the behaviour hadn't finished its layout process yet - it later would modify `top` property of the Screen, which we are interested in.

Therefore I've moved the "onlayout" callback to new method `onBottomSheetBehaviorDidLayout` which is called from our `ScreensCoordinatorLayout` after it has completely finished computing the layout.

I decided to not move all the code to the mentioned new callback, because the semantics are a bit different, especially the `changed` parameter - in case of `Screen.onLayout` it determines whether the bounds of the screen changed. In case of `onBottomSheetBehaviorDidLayout` the `coordinatorLayoutDidChange` informs whether coordinator layout bounds did change. I don't know whether this (moving all the logic to the new callback) would lead to some issues or not, therefore I've decided to leave regular screen behaviour intact for now.

kkafar (commented):
Notes for future me

kligarski (approved):
Great!

kkafar (commented):
CI passes locally, here it fails most likely due to slow runner.

PR #2789, opened 2025-03-18
fix(Android): `formSheet` with `fitToContents` does not have correct height after goBack

kkafar:
## Description

When fragment is reattached (on navigation back) `onCreateView` is called which tries to configure the form sheet (that has `fitToContents` set).
On view recreation the subtree is not laid out (the flag is cleared) & therefore current code fails.

## Changes

We now utilise the fact, that the Screen's subtree is retained by react-native and the components are not recreated themselves -> old height is stored in the member variable. 
We reuse it. 

Added Test2789

## Checklist

- [x] Ensured that CI passes


kkafar (commented):
CI failure is unrelated - started failing in ~60% of times, most likely to slow runner.

Passes just fine locally.

kligarski (commented):
Nice.

PR #2795, opened 2025-03-21
fix(Android): drawing order for multiple disappearing screens

kligarski:
## Description
WIP

## Changes
WIP

WIP

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
@@ -97,7 +97,7 @@ class ScreenStack(
     /**
      * Marks given fragment as to-be-dismissed and performs updates on container
      *
-     * @param fragmentWrapper to-be-dismissed wrapper
+     * @param screenFragment to-be-dismissed wrapper
      */
     fun dismiss(screenFragment: ScreenStackFragmentWrapper) {
         dismissedWrappers.add(screenFragment)
@@ -259,6 +259,8 @@ class ScreenStack(
                 childDrawingOrderStrategy =
                     ReverseOrderInRange(max(stack.lastIndex - dismissedTransparentScreenApproxCount + 1, 0)..stack.lastIndex)
             }
+        } else if (newTop != null && !newTopAlreadyInStack && topScreenWrapper?.screen?.isTransparent() == true) {
+            childDrawingOrderStrategy = SwapLastTwo()
         }
 
         createTransaction().let { transaction ->
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
@@ -92,7 +92,6 @@ class ScreenStack(
     private val drawingOpPool: MutableList<DrawingOp> = ArrayList()
     private var drawingOps: MutableList<DrawingOp> = ArrayList()
     private var topScreenWrapper: ScreenStackFragmentWrapper? = null
-    private var previousChildrenCount = 0
     private var removalTransitionChildrenCount = 0
 
     private var childDrawingOrderStrategy: ChildDrawingOrderStrategy? = null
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
```diff
@@ -124,20 +129,22 @@ class ScreenStack(
...
             dispatchOnFinishTransitioning()
         }
     }
 
     fun onViewAppearTransitionEnd() {
-        if (!removalTransitionStarted) {
+        if (removalTransitionChildrenCount == 0) {
             dispatchOnFinishTransitioning()
         }
     }
```
kligarski, line 150: I'm not sure whether this properly mirrors logic from before.

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
```diff
@@ -124,20 +129,22 @@ class ScreenStack(
...
+        removalTransitionChildrenCount += 1
+        // println("+1, now: " + removalTransitionChildrenCount)
     }
 
     override fun endViewTransition(view: View) {
         super.endViewTransition(view)
-        childDrawingOrderStrategy?.disable()
-        if (removalTransitionStarted) {
-            removalTransitionStarted = false
+        removalTransitionChildrenCount = max(0, removalTransitionChildrenCount - 1)
```
kligarski, line 138: Without this, the count would go negative - in the formSheet example (TestFormSheet -> Open Sheet -> Open Second) it prints:
```
+1, now: 1
+1, now: 2
-1, now: 1
-1, now: 0
-1, now: 0
-1, now: 0
-1, now: 0
```

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
```diff
@@ -229,16 +236,27 @@ class ScreenStack(
...
+                .asReversed()
+                .asSequence()
+                .filter { !dismissedWrappers.contains(it)
+                        && it.screen.activityState !== Screen.ActivityState.INACTIVE }
+                .drop(1)
+                .takeWhile { it.screen.isTransparent() }
+                .count() + 1
+
+            childDrawingOrderStrategy = ReverseOrderInRange(
+                screenWrappers.lastIndex - disappearingCount..screenWrappers.lastIndex)
```
kligarski, line 259: I'm not sure what reference should I use here to count disappearing screens - I mirrored logic from `notDismissedWrappers` but I guess you can also use the `stack`?

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
```diff
@@ -354,12 +372,6 @@ class ScreenStack(
     override fun dispatchDraw(canvas: Canvas) {
         super.dispatchDraw(canvas)
 
-        // check the view removal is completed (by comparing the previous children count)
-        if (drawingOps.size < previousChildrenCount) {
-            childDrawingOrderStrategy = null
-        }
-        previousChildrenCount = drawingOps.size
-
```
kligarski, line 362: I think with disable() method on childDrawingOrderStrategy this probably isn't needed anymore but I am not 100% sure.

kkafar (commented):
In general I think that what we might want to do is:

1. Mark the to-be-removed children for removal & make this state `internal`, so that the screen stack has access to it.

When screenstack starts view **removing** transition on some coordinator layout (SCL), we should store reference to this SCL in screen stack.

When screenstack ends view transition we should remove the reference.

Only when the reference store is empty, we should disable child drawing order strategy. In case the number of transitioning children mutates & single child drawing order strategy is not enough, we should find a way to update it or write adaptive drawing order strategy. I'll add a draft of the solution soon.

kligarski:
Continuation & new approach in https://github.com/software-mansion/react-native-screens/pull/2806.

PR #2796, opened 2025-03-21
fix(Android): implement ReactPointerEventsView to prevent header config from intercepting touches

### LINKED ISSUE #2794: ScreenStackHeaderConfig masks the top of Header area
### Description

The header config blocks gestures close to the top of the screen. This issue was supposed to be fixed with https://github.com/software-mansion/react-native-screens/pull/2781, but I am still facing the issue with the old architecture. 
In case I try following [this comment](https://github.com/software-mansion/react-native-screens/issues/2758#issuecomment-2720300998) (adding `top: '-100%'`), it does get resolved.

The GitHub repository link is given to reproduce the issue (The code given is extracted out from an internal react native project)

Screenshot of Element Inspect View:
Image

Dev Tools component shown when that particular part is selected:
Image

### Steps to reproduce

1. Follow the GitHub repository link given below.
2. Use old architecture & latest (beta) version of the rnscreens packages.

https://github.com/bhavberi/rnscreen-error

### Screens version

^4.10.0-beta.3

### React Native version

^0.78.0

### Platforms

Android

### JavaScript runtime

None

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

Real device

### Device model

redmi note 10 pro max

### Acknowledgements

Yes

kkafar:
## Description

Closes #2794

We set `visibility = GONE` for the header config in HostTree (HT), however this only prevents the Android framework from setting header config frame,
and does not prevent `ReactNative` from setting non-zero frame. When it has non-zero frame, react recognizes it as a touch target.

## Changes

I've implemented `ReactPointerEventsView` (returning `box-none`) making sure that RN won't treat it as touch target.

I considered another approach - setting the frame to zero or not letting react native to set the frame on the view - it also works, 
however requires cumbersome implementation on Android API level < 29, since `suppressLayout` method is available since 29.
On lower API levels you can not really ignore the frame - you can call `layout` with 0 width in `onLayout` of the header config...

`HeaderOptions` screen in `Example` app.

Tested it on both architectures & with `headerShown: false/true`.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
@@ -480,7 +480,7 @@ class ScreenStackFragment :
         constructor(context: Context, fragment: ScreenStackFragment) : this(
             context,
             fragment,
-            ScreensCoordinatorLayoutPointerEventsImpl(),
+            PointerEventsBoxNoneImpl(),
         )
 
         override fun onApplyWindowInsets(insets: WindowInsets?): WindowInsets = super.onApplyWindowInsets(insets)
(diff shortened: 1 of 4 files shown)


kligarski (commented):
Great!

Issue #2549, opened 2024-12-04
presentation: 'modal' Screens are tiny on iPad / iOS18


jforaker:
### Description

Greetings. 

(Firstly I'm always a little confused about who owns what in terms of `react-navigation` vs `react-navite-screens` so hopefully I'm in the right place. If I'm not please let me know!)

Ever since iOS 18 came out, our users have pointed that several screens now appear really tiny on IOS.

Note in the screenshot the screen titled "Adjust Study Weights" is a `createNativeStackNavigator()` screen pushed with `presentation: 'modal' `. I believe this to be a bug somewhere down the native screen chain, but its hard for me as an outsider to isolate the source. 

`"react-native-screens": "4.3.0"`

Note this is reproducible in the Snack: https://snack.expo.dev/TrJplgYWOFa8TGOILcUb6 (You might have to choose IOS > "iPad Pro 12.9" manually) 

![image](https://github.com/user-attachments/assets/86e9d207-f99a-42af-ab32-2886b4c6db99)

side by side ios 17/18

![image](https://github.com/user-attachments/assets/0431f684-7103-4f24-a9a4-61289a74f5ef)

### Steps to reproduce

1. make some screens using `createNativeStackNavigator()`
2. use `presentation: 'modal'` and test on an iPad

### Snack or a link to a repository

https://snack.expo.dev/O78qXxcUm1WbmoQuoTM9A

### Screens version

4.3.0

### React Native version

0.73.9

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Release mode

### Device

iOS simulator

### Device model

iPad Pro 13-inch (M4), iOS 18.0

### Acknowledgements

Yes

kkafar:
Unfortunately it seems that this is not a bug. This is change of behaviour introduced with iOS 18 & when I did research on it ~1.5 weeks ago I haven't found **UIKit** API to restore older behaviour. Apple added **SwiftUI** API [(link)](https://developer.apple.com/documentation/swiftui/presentationsizing/proposedsize(for:context:)) to allow to restore old behaviour, however I haven't found equivalent in UIKit. Right now I do not know what should be the way forward with this issue. I'm open for suggestions.

jforaker:
[...]

PR #2797, opened 2025-03-24
fix(iOS): use prefersPageSizing for modal on iOS 18.0+

kligarski:
## Description

Brings back `stackPresentation: 'modal'`'s size on iOS 18.0+.

Since the release of iOS 18, screens with `stackPresentation: 'modal'` changed size on devices with a bigger screen, such as an iPad - they became much smaller, which impacted applications of the library's users (see issue https://github.com/software-mansion/react-native-screens/issues/2549). This happens because `UIModalPresentationAutomatic`, which is used by `react-native-screens` for modal on iOS, has been changed with the release of iOS 18. `UIModalPresentationAutomatic` is now mapped to `UIModalPresentationFormSheet` for iOS >=18 and `UIModalPresentationPageSheet` for earlier versions (this was the default before).

For more context please see PR https://github.com/software-mansion/react-native-screens/pull/2793 but the TL;DR is that we consider this a breaking change for our users and therefore we decided to bring back the old behavior for `react-native-screens` v4 using [`prefersPageSizing`](https://developer.apple.com/documentation/uikit/uisheetpresentationcontroller/preferspagesizing?language=objc) property of the `UISheetPresentationController` which has been available since iOS 17.0+.

Resolves https://github.com/software-mansion/react-native-screens/issues/2549.

## Changes

- use `prefersPageSizing` for modal on iOS 18.0+

### Before
Modal without `prefersPageSizing` on iOS 18

### After
Modal with `prefersPageSizing` on iOS 18

Open `Modals` example screen and use `Open modal` button.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
+++ b/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
@@ -247,7 +247,10 @@ Using `containedModal` and `containedTransparentModal` with other types of modal
 
 For iOS:
 
-- `modal` will use [`UIModalPresentationAutomatic`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationautomatic?language=objc) on iOS 13 and later, and will use [`UIModalPresentationFullScreen`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationfullscreen?language=objc) on iOS 12 and earlier. For iOS 18 and later, additional prop `prefersPageSizing` will be set to `true` to ensure consistent behavior between iOS versions (`UIModalPresentationAutomatic` was mapped to `UIModalPresentationPageSheet` before iOS 18 but now it is mapped to `UIModalPresentationFormSheet`).
+- `modal` will use:
+   *  on iOS 18 and later [`UIModalPresentationAutomatic`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationautomatic?language=objc). However, since the iOS 18 changed the default behaviour from `UIModalPresentationPageSheet` to `UIModalPresentationFormSheet` (it looks vastly different on regular size classes devices, e.g. iPad), for the sake of backward compatibility, we keep the default behaviour from before iOS 18. *This might change in future major release of `react-native-screens`.
+   *  on iOS 13 and later [`UIModalPresentationAutomatic`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationautomatic?language=objc)
+   * on iOS 12 and earlier will use [`UIModalPresentationFullScreen`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationfullscreen?language=objc). 
 - `fullScreenModal` will use [`UIModalPresentationFullScreen`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationfullscreen?language=objc)
 - `formSheet` will use [`UIModalPresentationFormSheet`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationformsheet?language=objc)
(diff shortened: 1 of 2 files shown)


ios/RNSScreen.mm
```diff
@@ -246,6 +246,17 @@ - (void)setStackPresentation:(RNSScreenStackPresentation)stackPresentation
...
+#if !TARGET_OS_TV
+      if (@available(iOS 18.0, *)) {
+        UISheetPresentationController *sheetController = _controller.sheetPresentationController;
+        if (sheetController != nil) {
+          sheetController.prefersPageSizing = true;
+        } else {
+          RCTLogError(
+              @"[RNScreens] sheetPresentationController is null when attempting to set prefersPageSizing for modal");
+        }
+      }
```
kkafar, line 259: This check for iOS 18 won't be enough. If we leave this in current shape it'll fail to compile when somebody uses XCode / SDK that does not support iOS 17 or greater. Sometimes people are limited due to hardware limitations or business reasons.

What we should do here is to wrap it in

```
#if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && defined(__IPHONE_17_0) && \
    __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_17_0
```

kkafar (changes requested):
Thanks!

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
```diff
@@ -247,7 +247,7 @@ Using `containedModal` and `containedTransparentModal` with other types of modal
 
 For iOS:
 
-- `modal` will use [`UIModalPresentationAutomatic`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationautomatic?language=objc) on iOS 13 and later, and will use [`UIModalPresentationFullScreen`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationfullscreen?language=objc) on iOS 12 and earlier.
+- `modal` will use [`UIModalPresentationAutomatic`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationautomatic?language=objc) on iOS 13 and later, and will use [`UIModalPresentationFullScreen`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationfullscreen?language=objc) on iOS 12 and earlier. For iOS 18 and later, additional prop `prefersPageSizing` will be set to `true` to ensure consistent behavior between iOS versions (`UIModalPresentationAutomatic` was mapped to `UIModalPresentationPageSheet` before iOS 18 but now it is mapped to `UIModalPresentationFormSheet`).
```
kkafar, line 250: Let's rewrite this.

```suggestion
- `modal` will use:
   *  on iOS 18 and later [`UIModalPresentationAutomatic`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationautomatic?language=objc). However, since the iOS 18 changed the default behaviour from `UIModalPresentationPageSheet` to `UIModalPresentationFormSheet` (it looks vastly different on regular size classes devices, e.g. iPad), for the sake of backward compatibility, we keep the default behaviour from before iOS 18. *This might change in future major release of `react-native-screens`.
   *  on iOS 13 and later [`UIModalPresentationAutomatic`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationautomatic?language=objc)
   * on iOS 12 and earlier will use [`UIModalPresentationFullScreen`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationfullscreen?language=objc). 
```

kkafar (changes requested):
Thanks for remembering that the documentation must also change.

I think we should rephrase the description.

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
kligarski: A lot more clear than before, thanks! 5dd11198e9abb7217ca77a3ee8085449e2cd891c

kkafar (approved):
✅

PR #2800, opened 2025-03-24
fix(iOS): prevent back button icon from "jumping" during pop animation when display mode minimal is set

kkafar:
## Description

See: https://github.com/react-navigation/react-navigation/pull/12500 
for issue description with videos.

We can use the backbuttondisplaymode to hide the backbuttontitle avoiding setting the title to `nil`
and therefore preventing the UIKit buggy back button icon animation.

@maciekstosio:
We agreed to leave `backTitleVisible`, so the API doesn't change in a minor version. The expected behavior is that `backTitleVisible: false` will work as "kill switch" and ignore other back button properties. When `backTitleVisible: true` other properties will work (including backButtonDisplayMode), but note that iOS omits `backButtonDisplayMode` when custom back button is provided (so when any of  `backTitleFontSize`, `backTitleFontFamily`, `headerBackButtonMenuEnabled`, `headerBackTitle` is used).

## Changes

:point_up:

You can use `Test1084`.

1. Simply set the `headerBackButtonDisplayMode: 'minimal'` on the second push screen.
2. Set `headerLargeTitle: true` on the first push screen.

## How this has been tested

@maciekstosio:

I tested RNScreens properties by modifying props in `react-navigation/packages/native-stack/src/views/useHeaderConfigProps.tsx`.

I used Test1084, with the first screen having `headerLargeTitle: true`.

**Tests:**

When `backTitleVisible: false`:
- backButtonDisplayMode is ignored
- backTitleFontSize is ignored
- backTitle is ignored (but it changes title in backButtonMenu)
- disableBackButtonMenu is ignored 

When `backTitleVisible: true | undefined`:
- `backButtonDisplayMode: minimal` works
- `backButtonDisplayMode: generic` works
- `backButtonDisplayMode: default` works
- `backTitle` works and overrides `backButtonDisplayMode`
- `headerBackTitleStyle.fontSize` works and overrides `backButtonDisplayMode`
- `headerBackButtonMenuEnabled` works and overrides `backButtonDisplayMode` 

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
+++ b/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
@@ -502,6 +502,8 @@ Allows for customizing font size to be used for back button title on iOS.
 
 Whether the back button title should be visible. Defaults to `true`.
 
+When set to `false` it works as a "kill switch": it enforces `backButtonDisplayMode=minimal` and ignores `backButtonDisplayMode`, `backTitleFontSize`, `backTitleFontFamily`, `disableBackButtonMenu`, and `backTitle` works only for back button menu.
+
 ### `blurEffect` (iOS only)
 
 Blur effect to be applied to the header. Works with `backgroundColor`'s alpha < 1.
@@ -520,7 +522,9 @@ Boolean indicating whether to show the menu on longPress of iOS >= 14 back butto
 
 ### `backButtonDisplayMode` (iOS only)
 
-Enum value indicating display mode of **default** back button. It works on iOS >= 14, and is used only when none of: `backTitleFontFamily`, `backTitleFontSize`, `disableBackButtonMenu` or `backTitle` is set. Otherwise, when the button is customized, under the hood we use iOS native `backButtonItem` which overrides `backButtonDisplayMode`. Read more [#2123](https://github.com/software-mansion/react-native-screens/pull/2123). Possible options:
+Enum value indicating display mode of back button. It is used only when none of: `backTitleFontFamily`, `backTitleFontSize`, `disableBackButtonMenu`, `backTitle` and `backTitleVisible=false` is set. When the button is customized, under the hood we use iOS native `backButtonItem` which overrides `backButtonDisplayMode`. Read more [#2123](https://github.com/software-mansion/react-native-screens/pull/2123). The `backTitleVisible` forces `backButtonDisplayMode: minimal` and omits other values. Read more [#2800](https://github.com/software-mansion/react-native-screens/pull/2800)
+
+Possible options:
 
 - `default` – show given back button previous controller title, system generic or just icon based on available space
 - `generic` – show given system generic or just icon based on available space
--- a/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
+++ b/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
@@ -522,7 +522,7 @@ Boolean indicating whether to show the menu on longPress of iOS >= 14 back butto
 
 ### `backButtonDisplayMode` (iOS only)
 
-Enum value indicating display mode of back button. It is used only when none of: `backTitleFontFamily`, `backTitleFontSize`, `disableBackButtonMenu`, `backTitle` and `backTitleVisible=false` is set. When the button is customized, under the hood we use iOS native `backButtonItem` which overrides `backButtonDisplayMode`. Read more [#2123](https://github.com/software-mansion/react-native-screens/pull/2123). The `backTitleVisible` forces `backButtonDisplayMode: minimal` and omits other values. Read more [#2800](https://github.com/software-mansion/react-native-screens/pull/2800)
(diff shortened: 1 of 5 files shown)


kkafar (commented):
Notes for future us

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
```diff
@@ -520,7 +522,9 @@ Boolean indicating whether to show the menu on longPress of iOS >= 14 back butto
 
 ### `backButtonDisplayMode` (iOS only)
 
-Enum value indicating display mode of **default** back button. It works on iOS >= 14, and is used only when none of: `backTitleFontFamily`, `backTitleFontSize`, `disableBackButtonMenu` or `backTitle` is set. Otherwise, when the button is customized, under the hood we use iOS native `backButtonItem` which overrides `backButtonDisplayMode`. Read more [#2123](https://github.com/software-mansion/react-native-screens/pull/2123). Possible options:
+Enum value indicating display mode of back button. It is used only when none of: `backTitleFontFamily`, `backTitleFontSize`, `disableBackButtonMenu`, `backTitle` and `backTitleVisible=false` is set. When the button is customized, under the hood we use iOS native `backButtonItem` which overrides `backButtonDisplayMode`. Read more [#2123](https://github.com/software-mansion/react-native-screens/pull/2123). The `backTitleVisible` forces `backButtonDisplayMode: minimal` and omits other values. Read more [#2800](https://github.com/software-mansion/react-native-screens/pull/2800)
```
kkafar, line 525:
```suggestion
Enum value indicating display mode of back button. It is used only when none of: `backTitleFontFamily`, `backTitleFontSize`, `disableBackButtonMenu`, `backTitle` and `backTitleVisible=false` is set. When these props are set, under the hood we use customise `backButtonItem` which disables `backButtonDisplayMode` behavior. Read more [#2123](https://github.com/software-mansion/react-native-screens/pull/2123). The `backTitleVisible` forces `backButtonDisplayMode: minimal` and omits other values. Read more [#2800](https://github.com/software-mansion/react-native-screens/pull/2800)
```

native-stack/README.md
```diff
@@ -82,7 +82,9 @@ Boolean indicating whether to show the menu on longPress of iOS >= 14 back butto
 
 #### `backButtonDisplayMode` (iOS only)
 
-Enum value indicating display mode of **default** back button. It works on iOS >= 14, and is used only when none of: `backTitleFontFamily`, `backTitleFontSize`, `disableBackButtonMenu` or `backTitle` is set. Otherwise, when the button is customized, under the hood we use iOS native `backButtonItem` which overrides `backButtonDisplayMode`. Read more [#2123](https://github.com/software-mansion/react-native-screens/pull/2123). Possible options:
+Enum value indicating display mode of back button. It is used only when none of: `backTitleFontFamily`, `backTitleFontSize`, `disableBackButtonMenu`, `backTitle` and `backTitleVisible=false` is set. When the button is customized, under the hood we use iOS native `backButtonItem` which overrides `backButtonDisplayMode`. Read more [#2123](https://github.com/software-mansion/react-native-screens/pull/2123). The `backTitleVisible` forces `backButtonDisplayMode: minimal` and omits other values. Read more [#2800](https://github.com/software-mansion/react-native-screens/pull/2800)
+
+Possible options:
```
kkafar, line 87: Let's apply my previous comment here also

src/types.tsx
```diff
@@ -570,11 +574,13 @@ export interface ScreenStackHeaderConfigProps extends ViewProps {
    */
   disableBackButtonMenu?: boolean;
   /**
-   * How the back button behaves by default (when not customized). Available on iOS>=14, and is used only when none of: `backTitleFontFamily`, `backTitleFontSize`, `disableBackButtonMenu` or `backTitle` is set.
-   * The following values are currently supported (they correspond to https://developer.apple.com/documentation/uikit/uinavigationitembackbuttondisplaymode?language=objc):
-   * - "default" – show given back button previous controller title, system generic or just icon based on available space
-   * - "generic" – show given system generic or just icon based on available space
-   * - "minimal" – show just an icon
+   * How the back button behaves. It is used only when none of: `backTitleFontFamily`, `backTitleFontSize`, `disableBackButtonMenu`, `backTitle` and `backTitleVisible=false` is set.
+   * To read more see [#2800](https://github.com/software-mansion/react-native-screens/pull/2800) and [#2800](https://github.com/software-mansion/react-native-screens/pull/2800).
```
kkafar, line 578: I don't think we should link these here. Previous link to apple docs should stay.

kkafar: This is not useful when you trigger signature help.

kkafar (commented):
Looks good, few last remarks. Thanks.

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
maciekstosio: > When these props are set
Don't like this part, cause backTitleVisible is separate case. What about:

```suggestion
Enum value indicating display mode of back button. It is used only when none of: `backTitleFontFamily`, `backTitleFontSize`, `disableBackButtonMenu`, `backTitle` and `backTitleVisible=false` is set. The `backTitleVisible` forces `backButtonDisplayMode: minimal` and omits other values. Read more [#2800](https://github.com/software-mansion/react-native-screens/pull/2800). The other props, under the hood, customize `backButtonItem` which overrides `backButtonDisplayMode`. Read more [#2123](https://github.com/software-mansion/react-native-screens/pull/2123).
```

kkafar: I think your suggestion is ✅

kkafar:
Let it be noted that CI fails for unrelated reasons (seems lik slow runner)

PR #2806, opened 2025-03-25
fix(Android): drawing order for multiple disappearing screens

kligarski:
## Description

Fixes drawing order to account for multiple disappearing screens (e.g. formSheets).

## Changes

- move `ScreensCoordinatorLayout`, `ScreensAnimation` to seperate package
- move `ChildDrawingOrderStrategy` interface to seperate file
- track disappearing transitioning children & disable `ChildDrawingOrderStrategy` when the transition is over
- add new, more general `ReverseOrderStartingFrom` strategy to replace `SwapLastTwo` and use it in `ScreenStack`'s `onUpdate`

Currently, the transition from a `formSheet` to `push` is not working properly (two animation mechanisms don't work correctly together) so it is difficult to verify drawing order. To temporarily disable formSheet animation, you can `return null` at the beginning of `ScreenStackFragment`'s `onCreateAnimator` method. Then go to `TestFormSheet` example sheet, open the sheet and then click `Open second`. New screen will be drawn over disappearing screens during the transition.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
@@ -89,7 +89,7 @@ class ScreenStack(
     private var drawingOps: MutableList<DrawingOp> = ArrayList()
     private var topScreenWrapper: ScreenStackFragmentWrapper? = null
     private var removalTransitionStarted = false
-    private var previousChildrenCount = 0
+//    private var previousChildrenCount = 0
 
     private var childDrawingOrderStrategy: ChildDrawingOrderStrategy? = null
     private var disappearingTransitioningChildren: MutableList<View> = ArrayList()
@@ -126,7 +126,7 @@ class ScreenStack(
     override fun startViewTransition(view: View) {
         check(view is ScreensCoordinatorLayout) { "[RNScreens] Unexpected type of ScreenStack direct subview ${view.javaClass}"}
         super.startViewTransition(view)
-        if (view.fragment.screen.isInRemovalTransition()) {
+        if (view.fragment.isRemoving) {
             disappearingTransitioningChildren.add(view)
         }
         if (disappearingTransitioningChildren.isNotEmpty()) {
@@ -367,11 +367,11 @@ class ScreenStack(
     override fun dispatchDraw(canvas: Canvas) {
         super.dispatchDraw(canvas)
 
-        // check the view removal is completed (by comparing the previous children count)
-        if (drawingOps.size < previousChildrenCount) {
-            childDrawingOrderStrategy = null
-        }
-        previousChildrenCount = drawingOps.size
+//        // check the view removal is completed (by comparing the previous children count)
+//        if (drawingOps.size < previousChildrenCount) {
+//            childDrawingOrderStrategy = null
+//        }
+//        previousChildrenCount = drawingOps.size
 
         childDrawingOrderStrategy?.apply(drawingOps)
 
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
@@ -49,21 +49,13 @@ internal class ReverseOrderInRange(
     }
 }
 
-internal class ReverseOrderStartingFrom(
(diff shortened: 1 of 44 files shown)


kkafar (commented):
This looks good!
I'll test it later in the day & we'll merge it after 4.10.0 release.

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
```diff
@@ -175,21 +174,21 @@ class ScreenStack(
...
-                    .dropWhile { it.screen.isTransparent() }
-                    .firstOrNull()
-                    ?.takeUnless { it === newTop }
-        }
+        newTop = notDismissedWrappers.firstOrNull()
+        visibleBottom =
+            notDismissedWrappers
+                .dropWhile { it.screen.isTransparent() }
+                .firstOrNull()
+                ?.takeUnless { it === newTop }
```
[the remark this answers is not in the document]
kligarski: No, I used `notDismissedWrappers` sequence to count disappearing views later in the code (line 244) and I didn't want to repeat the same logic there with reversing and filtering the sequence. Was the `run` scope important?

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
```diff
@@ -226,16 +228,28 @@ class ScreenStack(
...
             // Atm needsDrawReordering implementation guards that assuming that first screen on stack uses `NONE` animation.
-            childDrawingOrderStrategy = SwapLastTwo()
+
+            // Count how many screens are disappearing
+            // (all transparent screens beneath newTop + one screen under them)
+            val disappearingCount =
+                notDismissedWrappers
+                    .drop(1)
+                    .takeWhile { it.screen.isTransparent() }
+                    .count() + 1
```
kkafar, line 247: I'm reading the code once again now & my initial impression is that it won't work in general case. I need to come up with counter example -- will try to do that next week.

kkafar: The problem is that `notDismissedWrappers` sequence is computed basing on `screenWrappers` - these represent **current** react view state. We don't have guarantees that this is consistent with **currently HostTree state** represented by `stack` list. 

I think we need to find some time in upcoming weeks & design this algorithm properly from the ground up.

kkafar (commented):
Left some notes for future us

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
kkafar: Notation: 

* uppercase letters, e.g. A, B - **opaque** screens,
* lowercase letters, e.g. a, b - **translucent** screens.

`stack`: `Abc`
`screenWrappers`: `AD`

This case will have issue, where the count of dismissed screens will be 0 -> therefore we won't change drawing order, while we should.

One thing to note that current solution might be good enough to work in simple stack operations (push, pop, replace). The one I've just described is not a stack operation. 

My aspiration is to make the screen stack work with arbitrary state changes, but I guess we can postpone this plant until we have more robust update algorithm.

kkafar (changes requested):
I've also spotted problem while testing on Example app. 

https://github.com/user-attachments/assets/688fe7af-f65f-4f86-bc4f-bc9f4e442b34

I do not know whether this is a regression or not. We must check before proceeding.

kkafar:
Yeah, [this](https://github.com/software-mansion/react-native-screens/pull/2806#pullrequestreview-2726319340) is a regression. Works fine on main.

kkafar (approved):
Okay, I think that current approach is reasonable.

When we decide to swap:

1. we want to draw entering screen after exiting ones -> therefore we need to put entering screen at the end of the list,
2. when we have multiple exiting screens we want to reverse the order among them, because framework draws them reversed (disappearing child with greater index in parent is drawn before one with smaller index).

Therefore reversing whole drawing operation list seems 🆗

kkafar:
I'll merge it once we manage to convince the CI to pass

kkafar:
No idea why CI fails. It passes locally.

image

PR #2808, opened 2025-03-25
fix(Android): fix backward compat with K1 compiler

kligarski:
## Description

Fixes backward compat with K1 compiler

## Changes

- remove `run` scope

## Checklist

- [ ] Ensured that CI passes


kkafar (approved):
Thanks!

PR #2816, opened 2025-03-28
test(Fabric): add e2e test for issue 577

kligarski:
## Description

Check which example screens from issues/PRs can be used in e2e testing for tests `Test550, ..., Test577` and implement them if possible for Fabric. 

### Test550
Skipped because the issue is present on the current version (custom native header back button image). When it is resolved, we might create an e2e test for it.

### Test556
Skipped because we can't check flickering nor non-deterministic white screen bug.

### Test564
Skipped because the issue is still present on the current version (headers overlapping when one navigatior has `headerTransparent: true`). When it is resolved, we might create an e2e test for it.

### Test577
Test created. It checks whether attempting to close a `gestureEnabled: false` modal on iOS doesn't break the modal's height allowing the content underneath to be visible. The test is present only on iOS because of differences in `modal` between iOS and Android.

## Changes

- add `Test577`
- add comments for test screen in `apps/src/tests/index.ts` with the reason for (not) implementing e2e test for it
- add `e2e-utils.ts` with `decribeIfiOS` function that will skip running the tests if they are run on Android

CI

## Checklist

- [x] Ensured that CI passes


kligarski:
After splitting tests into seperate files, CI fails on Android because some tests are iOS-only. We need to figure out how to handle it.

kkafar:
I don't know the details yet, but maybe lets add some `else` clause & add some empty test on Android?

kligarski:
We might consider using [`--passWithNoTests`](https://jestjs.io/docs/cli#--passwithnotests) Jest flag to avoid wasting resources for testing only that a tests exists in the menu on Android (considering how much time scrolling takes in these tests).

FabricExample/e2e/issuesTests/Test577.e2e.ts
```diff
@@ -0,0 +1,28 @@
...
+  });
+
+  // This issue is related to iOS modal
+  if (device.getPlatform() === 'ios') {
+    it('does not display content underneath modal with gesture disabled when attempting to close it', async () => {
+      await element(by.text('Open modal')).tap();
+      for (let i = 0; i < 5; ++i) {
+        await element(by.text('Modal')).swipe('down', 'fast');
+        await expect(element(by.text('Open modal'))).not.toBeVisible();
+      }
```
kkafar, line 25: Single attempt was not enough? 😄

kkafar (commented):
Looks good initially

FabricExample/e2e/issuesTests/Test577.e2e.ts
kligarski: Original bug seems to be happening after a few consecutive attempts to close the modal: https://github.com/software-mansion/react-native-screens/issues/577#issue-666185758. I wasn't sure how to test it - doing it once seems not enough and choosing a big number of attempts seems unnecessary. I guess that we can skip this issue or do a single attempt - let me know what you think is suitable.

kkafar: I think it is fine. But this explanation should be present in code & we're good.

kligarski:
It turns out that `--passWithNoTests` works only if there are no files at all - this is not our case. I created an utility that allows us to use `describe` or `describe.skip` based on platform (Android or iOS).

kkafar (approved):
✅

Issue #2611, opened 2025-01-10
useHeaderHeight does not update after cancelling search


C2thehris:
### Description

See https://github.com/react-navigation/react-navigation/issues/12331

### Current behavior

useHeaderHeight() does not update its value when the search bar is closed

https://github.com/user-attachments/assets/7108eaa9-1ba1-4806-ac59-9b8d4910f704

### Expected behavior

useHeaderHeight() should update to provide the adjusted height after the search bar is dismissed.

### Steps to reproduce

1. Click search bar
2. Click "Cancel"

### Snack or a link to a repository

https://snack.expo.dev/@c2thehris/useheaderheight-with-searchbar-ios-bug

### Screens version

4.4.0

### React Native version

0.76

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

None

### Acknowledgements

Yes

iliapnmrv:
Hi! I think you're missing this part from [docs](https://reactnavigation.org/docs/native-stack-navigator/#headersearchbaroptions)

>You also need to specify contentInsetAdjustmentBehavior="automatic" in your ScrollView, FlatList etc. If you don't have a ScrollView, specify headerTransparent: false.

kkafar:
Thanks for the report! We'll look into this

JustJoostNL:
Can confirm this issue.

Pnlvfx:
Same issue with reanimatedHeader, 

I opened an issue on react navigation but I think I should have opened it here.

https://github.com/react-navigation/react-navigation/issues/12427

PR #2817, opened 2025-03-31
fix useHeaderHeight hook

maciekstosio:
## Description

This PR fixes https://github.com/software-mansion/react-native-screens/issues/2611.

After this changes the hook will provide header height values after focusing search bar. 

## Changes

It seems that we have a logic, that disables `useHeaderHeight` updates when the modal is being closed. The condition assumes that only if there is a [presentedViewController](https://developer.apple.com/documentation/uikit/uiviewcontroller/presentedviewcontroller?language=objc), this is a modal. But it looks like it can also be a `UISearchController`, so we exclude that from the condition. 

## Screenshots / GIFs

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/928fdd00-a27e-4ff4-82d3-adf9f9e69cba" />|<video src="https://github.com/user-attachments/assets/b5c9dd7b-570c-4a78-b244-ffcff26a7032"/>| 

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStack.mm
+++ b/ios/RNSScreenStack.mm
@@ -77,7 +77,8 @@ namespace react = facebook::react;
     RNSScreen *screenController = (RNSScreen *)self.topViewController;
     BOOL isNotDismissingModal = screenController.presentedViewController == nil ||
         (screenController.presentedViewController != nil &&
-         ![screenController.presentedViewController isBeingDismissed]);
+         ![screenController.presentedViewController isBeingDismissed]) ||
+        ([screenController.presentedViewController isKindOfClass:[UISearchController class]]);
 
     // Calculate header height during simple transition from one screen to another.
     // If RNSScreen includes a navigation controller of type RNSNavigationController, it should not calculate


ios/RNSScreenStack.mm
```diff
@@ -77,7 +77,8 @@ - (void)viewDidLayoutSubviews
     RNSScreen *screenController = (RNSScreen *)self.topViewController;
     BOOL isNotDismissingModal = screenController.presentedViewController == nil ||
         (screenController.presentedViewController != nil &&
-         ![screenController.presentedViewController isBeingDismissed]);
+         ![screenController.presentedViewController isBeingDismissed]) ||
+        ([screenController.presentedViewController isKindOfClass:[UISearchController class]]);
```
kkafar, line 81: The condition is ok, but naming is misleading. 

Either rename the variable to `isNotDismissingModalOrPresentingSearchController` or 
split this condition - leave `isNotDismissingModal` as is and add new variable `isNotPresentingSearchController` and use it below.

kkafar (changes requested):
Left single comment.

ios/RNSScreenStack.mm
maciekstosio: Hmmm, I think the naming is correct, although it's unnecessarily (?) complicated. 

```
BOOL isDismissingModal = (screenController.presentedViewController != nil && [screenController.presentedViewController isBeingDismissed]);
```

and then

```
if (![screenController hasNestedStack] && !isDismissingModal) {
```

maciekstosio: changed in: https://github.com/software-mansion/react-native-screens/pull/2817/commits/97e4b0d4043fd1422d9bd0bb0f71a4addea16028

ios/RNSScreenStack.mm
```diff
@@ -78,11 +78,13 @@ - (void)viewDidLayoutSubviews
     BOOL isNotDismissingModal = screenController.presentedViewController == nil ||
         (screenController.presentedViewController != nil &&
          ![screenController.presentedViewController isBeingDismissed]);
+    BOOL isPresentingSearchController =
+        ([screenController.presentedViewController isKindOfClass:[UISearchController class]]);
```
kkafar, line 82:
```suggestion
        [screenController.presentedViewController isKindOfClass:UISearchController.class];
```

kkafar (approved):
This one & we're cacy

PR #2819, opened 2025-03-31
fix(Android): pressables lose focus on fast movement when `gesture-handler` is present

kkafar:
## Description

When `GestureHandler` & `GestureHandlerRootView` are present, the pressables
in subtree of `GestureHandler` will lose focus on fast initial movement. 
See the video below :point_down:.

> [!note]
> Edit: the *fast movement* requirement most likely comes from the fact that we use `fling` for the `EmptyGestureHandler`. In case you use e.g. `Pan`, the movement does not have to fast.

> [!note]
> The issue is reproducible **w/o** `react-native-screens`.

The issue affects us, because #1913 added custom-screen-transitions based 
on integration with `gesture-handler` & `reanimated`. Each `ScreenStack` is since then 
wrapped in a `GestureDetector`, which uses either [`EmptyGestureHandler`](https://github.com/software-mansion/react-native-screens/blob/4d044dab8c387c6f3988434276594a2d917f5022/src/gesture-handler/ScreenGestureDetector.tsx#L27)
or [one defined based on gesture config](https://github.com/software-mansion/react-native-screens/blob/4d044dab8c387c6f3988434276594a2d917f5022/src/gesture-handler/ScreenGestureDetector.tsx#L247).

### Issue mechanism

Basically it seems like bug in `gesture-handler`. The `RNGestureHandlerRootHelper` [cancels gesture](https://github.com/software-mansion/react-native-gesture-handler/blob/4232fc3127949c2869769542bfc7c3710b1348ee/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.kt#L87)
on the root view, causing RN pressable gesture handling to be cancelled. Instead, the attached gesture-handler
intercepts the touch events.

> [!warning]
> I was able to reproduce the issue 100% reliably earlier w/o presence of the `Screen` component, however
> testing it now it seems that the `Screen` is required (w/o `Screen` in view hierarchy it seems to work fine...).
> Dunno what is going on :shrug:

### Issue recording

https://github.com/user-attachments/assets/d0e9de2e-7d77-4379-b3f8-5da9e8b028ac

## Changes

It looks to me that it has to be handled on the side of `gesture-handler` / `reac-native` core.

Currently, I've decided to `disable` the `EmptyGestureHandler` we wrap the `ScreenStack`.

> [!caution]
> **This is only partial workaround**. There still will be issue in case someone setups the 
`goBackGesture` & we pass `enabled` `GestureHandler` to `GestureDetector`.

`Test2819`, may require you to wrap `HomeOne` inside a `Screen`, depending on the device/RN/other code mood I guess.

## Checklist

- [ ] Ensured that CI passes


m-bert (approved):

kligarski (commented):
👍

PR #2820, opened 2025-04-01
fix(Android,Paper): fix flicker during formsSheet enter transition on slow Android devices - not preloaded formSheet

kkafar:
## Description

On old architecture we have "enter transition postponing" mechanism, however it has been applied only to formsheets with fit to contents. It turned out that on **very low-end** Android devices the issue affects also not `fitToContents` sheets. 

## Changes

This PR extends the enter transition postponing mechanism to all `sheetAllowedDetents` values.

> [!important]
> There still remains problem though, when the sheet is preloaded before navigating to it. The content wrapper receives the layout from RN, however the screen never receives the frame, because it is not initially attached to the container (and native layout triggered from screen stack does not reach it).
>
> Then, when when the transition starts, content wrapper is already laid out, however the screen is not & **FOR SOME WEIRD** reasons it receives many frame updates during the transition - animated value updates are interleaved with `Screen.onLayout` calls - resulting in flicker effect.

Was not able to reproduce the issue on our end :/

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetUtils.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetUtils.kt
@@ -6,6 +6,7 @@ import com.google.android.material.bottomsheet.BottomSheetBehavior.STATE_COLLAPS
 import com.google.android.material.bottomsheet.BottomSheetBehavior.STATE_EXPANDED
 import com.google.android.material.bottomsheet.BottomSheetBehavior.STATE_HALF_EXPANDED
 import com.google.android.material.bottomsheet.BottomSheetBehavior.STATE_HIDDEN
+import com.swmansion.rnscreens.BuildConfig
 import com.swmansion.rnscreens.Screen
 
 object SheetUtils {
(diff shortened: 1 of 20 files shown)


android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetUtils.kt
```diff
@@ -133,3 +135,24 @@ fun Screen.isSheetFitToContents(): Boolean =
...
+    // This is used only for formSheet presentation, because we use value animators
+    // there. Tween animations have some magic way to make this work (maybe they
+    // postpone the transition internally, dunno).
+
+    if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED || !this.usesFormSheetPresentation()) {
+        return false
+    }
+    // Assumes that formSheet uses content wrapper
+    return !this.isLaidOutOrHasCachedLayout() || this.contentWrapper?.isLaidOutOrHasCachedLayout() != true
+}
```
kkafar, line 152: This fails in case the formsheet has been preloaded. Described on internal board.

kkafar (commented):
Left a comment for future me.

PR #2825, opened 2025-04-02
feat: support for RN 0.79

kkafar:
- **FE: update js deps**
- **FE: Reinstall gems**
- **FE: reinstall pods**
- **FE: update AppDelegat**
- **FE: update gradle wrapper**
- **CHECK IN .ruby-version**

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


kkafar:
Nightly reanimated should already work -> @kkafar

kkafar:
I've disabled `Test1649` for a while, because it causes crash during bundle loading.

image


The problem is resolved once the `jotai` dependency is removed.

We need to get rid of the lib or find a way to fix the issue.

kkafar:
I've reported the issue on detox repo here: https://github.com/wix/Detox/issues/4760

Issue #2541, opened 2024-11-26
Native Stack is not working with react-native-windows@0.76


pcprinz:
### Description

After a clean `react-native-windows@0.76` installation followed by the recommended installation for the native stack from the [Getting Started guide](https://reactnavigation.org/docs/getting-started), I get the following build error:

> × Building Solution: C:\dev\RNWTests\node_modules\react-native-screens\windows\RNScreens\RNScreens.vcxproj(184,5...
 × Build failed with message 7:10>C:\dev\RNWTests\node_modules\react-native-screens\windows\RNScreens\RNScreens.vcxproj(184,5): error : This project references NuGet package(s) that are missing on this computer. Use NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is C:\dev\RNWTests\windows\\packages\Microsoft.Windows.CppWinRT.2.0.200316.3\build\native\Microsoft.Windows.CppWinRT.props.. Check your build configuration.

### Steps to reproduce

1. install clean react-native@0.76
2. install and init react-native-windows@0.76
3. install `@react-navigation/native`,  `react-native-screens`, `react-native-safe-area-context`, `@react-navigation/native-stack`, `@react-navigation/elements`
4. run `npx @react-native-community/cli run-windows`

### Snack or a link to a repository

https://github.com/pcprinz/react-native-windows-tests/tree/0.76-failing-native-stack

### Screens version

4.3.0

### React Native version

0.76.2

### Platforms

Windows

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

None

### Device

None

### Acknowledgements

Yes

kkafar:
Thanks for reporting! Noted. I'll try to get someone to look into this.

dppo:
It works fine using 4.2.0

pedro-w:
I had this and solved it by editing `NuGet.Config` in the `windows\` sub-directory. I moved the Nuget source above the react-native one, and added the protocol version.
i.e. from
```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <config>
    <add key="repositoryPath" value="packages" />
  </config>
  <packageSources>
    <clear />
    <add key="react-native" value="https://pkgs.dev.azure.com/ms/react-native/_packaging/react-native-public/nuget/v3/index.json" />
[...]

PR #2826, opened 2025-04-03
fix(Windows): fix building and crashing issues

kligarski:
## Description

Fixes problems with building and crashes in Windows apps. Some components are still WIP and we'll finish work on them in the future.

Resolves https://github.com/software-mansion/react-native-screens/issues/2541.

## Changes

- add Windows-specific implementations of `ScreenContentWrapper` and `ScreenFooter`,
- start work on stack header and subviews, searchbar,  stack animations - these components are not implemented yet
- change minimum version of `WindowsTargetPlatform` to match `WinUI` requirements,

Create new `react-native-windows` app with components from `react-native-screens`, build and run (it might be necessary to run `npx react-native run-windows` twice in order to install all required packages).

### DIFF
--- a/windows/RNScreens/ScreenStackHeaderConfig.cpp
+++ b/windows/RNScreens/ScreenStackHeaderConfig.cpp
@@ -15,5 +15,29 @@ using namespace Windows::UI::Xaml::Controls;
 namespace winrt::RNScreens::implementation {
 ScreenStackHeaderConfig::ScreenStackHeaderConfig(
     winrt::Microsoft::ReactNative::IReactContext reactContext)
-    : m_reactContext(reactContext) {}
+    : m_reactContext(reactContext),
+     m_children(
+                    {winrt::single_threaded_vector<Windows::UI::Xaml::UIElement>()}) {}
+
+void ScreenStackHeaderConfig::addView(winrt::Windows::UI::Xaml::UIElement element) {
+  Children().Append(element);
+}
+
+void ScreenStackHeaderConfig::removeAllChildren() {
+  Children().Clear();
+}
+
+void ScreenStackHeaderConfig::removeChildAt(int64_t index) {
+  Children().RemoveAt(static_cast<uint32_t>(index));
+}
+
+void ScreenStackHeaderConfig::replaceChild(
+    winrt::Windows::UI::Xaml::UIElement oldChild,
+    winrt::Windows::UI::Xaml::UIElement newChild) {
+  uint32_t index;
+  if (!Children().IndexOf(oldChild, index))
+    return;
+
+  Children().SetAt(index, newChild);
+}
 } // namespace winrt::RNScreens::implementation
--- a/windows/RNScreens/ScreenStackHeaderConfigViewManager.cpp
+++ b/windows/RNScreens/ScreenStackHeaderConfigViewManager.cpp
@@ -26,7 +26,48 @@ ScreenStackHeaderConfigViewManager::CreateView() noexcept {
 
 // IViewManagerRequiresNativeLayout
 bool ScreenStackHeaderConfigViewManager::RequiresNativeLayout() {
-  return true;
+  return false;
+}
+
+// IViewManagerWithChildren
+void ScreenStackHeaderConfigViewManager::AddView(
+    FrameworkElement parent,
+    UIElement child,
+    int64_t index) {
+  auto screenStackHeaderConfig = parent.as<ScreenStackHeaderConfig>();
+  if (!screenStackHeaderConfig)
+    return;
+
+  screenStackHeaderConfig->addView(child);
+}
+
+void ScreenStackHeaderConfigViewManager::RemoveAllChildren(FrameworkElement parent) {
+  auto screenStackHeaderConfig = parent.as<ScreenStackHeaderConfig>();
+  if (!screenStackHeaderConfig)
+    return;
+
+  screenStackHeaderConfig->removeAllChildren();
+}
+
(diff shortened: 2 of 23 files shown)


windows/RNScreens/ScreenStackHeaderConfig.cpp
```diff
@@ -15,5 +15,29 @@ using namespace Windows::UI::Xaml::Controls;
...
+
+void ScreenStackHeaderConfig::replaceChild(
+    winrt::Windows::UI::Xaml::UIElement oldChild,
+    winrt::Windows::UI::Xaml::UIElement newChild) {
+  uint32_t index;
+  if (!Children().IndexOf(oldChild, index))
+    return;
+
+  Children().SetAt(index, newChild);
+}
```
kkafar, line 42: I believe the super implementation is exactly the smae & we can remove this code. It has been most likely added just to simplify debugging process.

kkafar: We need to verify whethert the super::xxx implementation is the same. If not, we'll leave this, otherwise its subject for removal.

windows/RNScreens/ScreenStackHeaderConfigViewManager.cpp
```diff
@@ -26,7 +26,48 @@ ScreenStackHeaderConfigViewManager::CreateView() noexcept {
...
+
+void ScreenStackHeaderConfigViewManager::ReplaceChild(
+    FrameworkElement parent,
+    UIElement oldChild,
+    UIElement newChild) {
+  auto screenStackHeaderConfig = parent.as<ScreenStackHeaderConfig>();
+  if (!screenStackHeaderConfig)
+    return;
+
+  screenStackHeaderConfig->replaceChild(oldChild, newChild);
```
kkafar, line 70: Similar story here. One thing to check is why the returned value in `RequiresNativeLayout` has changed.

windows/RNScreens/ScreenStackHeaderSubview.cpp
```diff
@@ -0,0 +1,43 @@
...
+    : m_reactContext(reactContext),
+      m_children(
+               {winrt::single_threaded_vector<Windows::UI::Xaml::UIElement>()}) {}
+
+void ScreenStackHeaderSubview::addView(winrt::Windows::UI::Xaml::UIElement element) {
+  Children().Append(element);
+}
+
+void ScreenStackHeaderSubview::removeAllChildren() {
+//   Children().Clear();
```
kkafar, line 27: This surely should not be commented out.

windows/RNScreens/ScreenStackHeaderSubview.cpp
```diff
@@ -0,0 +1,43 @@
...
+}
+
+void ScreenStackHeaderSubview::removeChildAt(int64_t index) {
+//   Children().RemoveAt(static_cast<uint32_t>(index));
```
kkafar, line 31: Same here

kkafar (commented):
Please also check which source files are required by metor by default (e.g. Screen.tsx or `Screen.web.tsx`)

windows/RNScreens/SearchBarViewManager.cpp
```diff
@@ -0,0 +1,88 @@
...
+    IJSValueReader const &propertyMapReader) noexcept {
+  (void)view;
+  const JSValueObject &propertyMap = JSValue::ReadObjectFrom(propertyMapReader);
+  for (auto const &pair : propertyMap) {
+    auto const &propertyName = pair.first;
+    auto const &propertyValue = pair.second;
+    (void)propertyName;
+    (void)propertyValue;
+  }
+}
```
kkafar, line 50: This also seems like added only for debugging purposes, we should not commit that.

kkafar (changes requested):
Okay, so we need to get rid of unused code / code introduced only to simplify debug. 

We need to check whether search bar has any functionality here. 

Andddd, we'll what do we do later.

windows/RNScreens/ScreenStackHeaderConfig.cpp
kligarski: I think we're implementing interface methods here.

windows/RNScreens/SearchBarViewManager.cpp
kligarski: Same kind of implementations were used in initial commit with support for `react-native-windows`, e.g. in `ScreenStackHeaderConfigViewManager.cpp`. https://github.com/software-mansion/react-native-screens/pull/989

windows/RNScreens/ScreenStackHeaderSubview.cpp
kligarski: 454fd44ae532d28565220e0fe8df146bc50487d5

windows/RNScreens/ScreenStackHeaderSubview.cpp
kligarski: 454fd44ae532d28565220e0fe8df146bc50487d5

kligarski:
> Please also check which source files are required by metor by default (e.g. Screen.tsx or `Screen.web.tsx`)

Windows apps use `Screen.tsx`.

windows/RNScreens/SearchBarViewManager.cpp
kkafar: Okay, it is **almost** noop, should not affect performance & we'll handle this on demand (when someone reports an issue)

kkafar (approved):
Let's proceed.

PR #2832, opened 2025-04-07
test(Windows): add Windows CI

kligarski:
## Description

Adds build test CI for Windows.

## Changes

- add Windows build test workflow for GitHub Actions

CI

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/.github/workflows/windows-build-test.yml
+++ b/.github/workflows/windows-build-test.yml
@@ -1,0 +1,72 @@
+name: Test Windows build
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'windows/**'
+      - '.github/workflows/windows-build-test.yml'
+  push:
+    branches:
+      - main
+    paths:
+      - 'windows/**'
+      - '.github/workflows/windows-build-test.yml'
+  workflow_dispatch:
+jobs:
+  build:
+    runs-on: windows-latest
+    timeout-minutes: 60
+    concurrency:
+      group: windows-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


kkafar (commented):
+1, it's not every elegant, but it is ok for now. 

I wanted it to run a bit more rarely than on every commit to main, however this is also fine. 

It would be great if we could inspect the file list changed by the particular commit and run the workflow only if the `windows/**` files were modified. Is there a option for that?

.github/workflows/windows-build-test.yml
```diff
@@ -0,0 +1,80 @@
...
+    paths:
+      - 'windows/**'
+      - '.github/workflows/windows-build-test.yml'
+
+  push:
+    branches:
+      - main
+    paths:
+      - 'windows/**'
+      - '.github/workflows/windows-build-test.yml'
```
kkafar, line 16: Seems, that I've overlooked this part for some reason 😅 This looks good, idk whether this works or not, but we can tune it once it lands. Let's just make sure in GitHub Actions docs that this makes sense & we're good.

kkafar (commented):
Follow up comment, I overlooked something last time.

.github/workflows/windows-build-test.yml
kligarski: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

PR #2838, opened 2025-04-07
fix(iOS,Fabric): fix sporadic exiting screen content jump on goBack action

### LINKED ISSUE #2538: Back Navigation displays bug while transitioning backwards
### Description

**Note**: Originally posted [here](https://github.com/react-navigation/react-navigation/issues/12292) but was redirected to this repo.
### Current behavior

I am experiencing a bug when using my personal iPhone but not on iOS simulator. The Simulator shown is the iPhone 15 Pro while my personal is the iPhone 12. The bug is that sometimes, when the backwards functionality is used: (1) either by verbosely calling navigation.goBack(); or (2) Pressing the icon on the bottom tab for the current stack that I'm on a bug is shown in the middle of the backwards transition/navigation. The bug is some black square on the bottom of the screen.

Personal iPhone (With bug):
https://github.com/user-attachments/assets/36f0213f-0c39-48e8-b128-fcd87178a3bf

iOS Simulator (Without the bug):
https://imgur.com/a/TuhXjzZ

### Expected behavior

I expect the backwards navigation/transition animation to appear the same as the forward navigation does.

### Environment

| @react-navigation/native               | 7.0.4
| @react-navigation/bottom-tabs          | 7.0.0
| @react-navigation/native-stack         | 7.1.1
| react-native-safe-area-context         | 4.12.0
| react-native-gesture-handler           | 2.20.2
| react-native-reanimated                | 3.16.1
| expo                                   | 52.0.11
| node                                   | 23.2.0
| npm                            | 10.9.0

### Steps to reproduce

1. Click the first bottom tab icon 
2. Click the screen on the initial page, this should redirect you to the second page
3. Click the screen on the second page, this should redirect you to the third page
4a. Click the first bottom tab icon to go back to the top of the stack (every few clicks this creates the bug)
4b. Click the screen of the third page to verbosely call navigation.goBack()  (every few clicks this creates the bug)

https://github.com/nathanembaye/test

### Screens version

4.1.0

### React Native version

0.76.3

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

Real device

### Device model

iPhone 12

### Acknowledgements

Yes

kkafar:
## Description

Fixes #2538

From time to time the **snapshot** of exiting screen would receive bad frame from the OS during the transition
resulting in the visual bug depicted in #2538 and on video down below :point_down:.

Basically what happens is described by logs present on the recording below.
In JS initiated back transition we exchange the `screenView` with its snapshot &
the snapshot **sometimes** receives UIKit-originated frame with origin at `(0, 0)`.

I guess it happens only sporadically, because there is some race condition, between
animation & layout depending on whether the navigation bar is still attached in the model
tree or not.

There is no issue in natively initiated back transition.

https://github.com/user-attachments/assets/d9d90cac-4507-44ff-bda1-619b71b5da56

## Changes

I've decided to go with setting the snapshot frame directly on creation. 
Given that the `screenView` has correct frame set (with header height taken into account),
the snapshot will have correct frame.

I've added simple `Test2538`, which you can test by going back and forth between screens.
With the patch applied I haven't seen the issue even once.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/apps/src/tests/Test2538.tsx
+++ b/apps/src/tests/Test2538.tsx
@@ -1,0 +1,41 @@
+import { NavigationContainer, ParamListBase } from '@react-navigation/native';
+import { NativeStackNavigationProp, createNativeStackNavigator } from '@react-navigation/native-stack';
+import React from 'react';
+import { Button, Text, View } from 'react-native';
+
+const Stack = createNativeStackNavigator();
+
+type RouteNavProps = {
+  navigation: NativeStackNavigationProp<ParamListBase>;
+}
+
+function Home({ navigation }: RouteNavProps) {
+  return (
+    <View style={{ flex: 1, backgroundColor: 'lightgreen' }}>
+      <Text>Home screen</Text>
+      <Button title="Go second" onPress={() => navigation.navigate('Second')} />
+    </View>
+  );
+}
+
+function Second({ navigation }: RouteNavProps) {
+  return (
+    <View style={{ flex: 1, backgroundColor: 'lightblue' }}>
+      <Text>Second screen</Text>
+      <Button title="Go back" onPress={() => navigation.goBack()} />
+    </View>
+  );
+}
+
+function App() {
+  return (
+    <NavigationContainer>
+      <Stack.Navigator>
+        <Stack.Screen name="Home" component={Home} />
+        <Stack.Screen name="Second" component={Second} />
+      </Stack.Navigator>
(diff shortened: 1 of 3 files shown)


maciekstosio (approved):
LGTM

kligarski (commented):
Looks good.

PR #2846, opened 2025-04-08
fix(Android): fix animations between formSheet and new screen

kligarski:
## Description

Fixes missing animation between formSheet and push screen.

FormSheets use `Animator` but entering screens use tween animations - these animation systems don't work together very well. We rewrote tween animations using `Animator`.

## Things to consider

- should we change all animations to `Animator` (current version) or rather detect a transition from formSheet to push screen and apply `Animator` only in that case, leaving tween animations for regular transistions (there might be difference in performance on older devices)
- should we implement these transitions using XML Resource files if they don't allow to define the whole animation (e.g. we can't set translation by percentage of width/height) and we have to do modify them in code

## Changes

- rewrite animations with Animators for every stack animation,
- create `CustomAnimatorSetProvider` to add translation animators to animators read from XML files  (translation by percentage of width/height is not possible in XML Animator file, we need to set exact values based on screen's dimensions)

### Default
| before | after |
| -- | -- |
| <video src="https://github.com/user-attachments/assets/2f57ea9d-14dc-42da-b75e-468302f3581f" alt="default before" /> | <video src="https://github.com/user-attachments/assets/fd2f02f1-d645-4f7a-9925-e5a4a6be0c9a" alt="default after" />  |

### None

| before | after |
| -- | -- |
| <video src="https://github.com/user-attachments/assets/b05b6faa-1969-4234-ad12-c350b51ec89c" alt="none before" /> | <video src="https://github.com/user-attachments/assets/0b24ddb4-d26d-4d75-b58e-4e6ec0ca2b15" alt="none after" />  |

There is a problem with the formSheet disappearing for one frame in the new version. We also need to consider whether the formSheet should animate at all.

### Fade

| before | after |
| -- | -- |
| <video src="https://github.com/user-attachments/assets/08478e3c-c206-4d72-bab9-d08c357a0ca5" alt="fade before" /> | <video src="https://github.com/user-attachments/assets/827cd2ad-bd94-40c0-b2e9-58d13ea1bb81" alt="fade after" />  |

### Slide from right (left)

| before | after |
| -- | -- |
| <video src="https://github.com/user-attachments/assets/15a4a20c-e02e-4ba4-aeae-49f43c736e9f" alt="slide_from_right before" /> | <video src="https://github.com/user-attachments/assets/911c0c86-2075-4c10-abca-c8a7ba70e85a" alt="slide_from_right after" />  |

There is a problem with the formSheet not moving horizontally with the screen beneath it. We need to decide whether we want to change it or leave it.

### Fade from bottom

| before | after |
| -- | -- |
| <video src="https://github.com/user-attachments/assets/a091812f-5e38-4155-9ec6-0e7355d69cf7" alt="fade_from_bottom before" /> | <video src="https://github.com/user-attachments/assets/c46432f0-289c-45a6-ab0a-2bab4bfe5ccd" alt="fade_from_bottom after" />  |

### iOS from right (left)

| before | after |
| -- | -- |
| <video src="https://github.com/user-attachments/assets/2966b52b-e48d-4c9b-b352-4b6652a34d3c" alt="ios_from_right before" /> | <video src="https://github.com/user-attachments/assets/01fc59e6-d42f-4672-b604-d498029bc230" alt="ios_from_right after" />  |

There is a problem with the formSheet not moving horizontally with the screen beneath it. We need to decide whether we want to change it or leave it.

Enter `TestFormSheet` test screen and compare animation when clicking `Open Second` and `Open Sheet` -> `Open Second`.

## Checklist

- [ ] Check animations on Paper
- [ ] Ensured that CI passes


kkafar:
Leaving a note for future us: the tween animations differ between API levels (below 33 e.g.), we need to make sure that we support this properly. 

I haven't read the code yet, just read the PR description.

android/src/main/res/base/animator/rns_default_exit_out.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8"?>
+<set xmlns:android="http://schemas.android.com/apk/res/android">
+    <objectAnimator
+        android:propertyName="alpha"
+        android:valueFrom="1"
+        android:valueTo="1"
+        android:duration="50"
+        android:interpolator="@android:interpolator/accelerate_decelerate" />
```
kligarski, line 8: Sometimes it was necessary to add animators that are doing nothing but filling the time - I am not sure if this impacts performance in significant way.

android/src/main/java/com/swmansion/rnscreens/animator/CustomAnimatorSetProvider.kt
```diff
@@ -0,0 +1,239 @@
...
+        ): AnimatorSet {
+            val finalAnimatorSet = AnimatorSet()
+            val screenParent = screen.parentAsView()
+
+            // TODO: verify that these values can be used to get width/height
+            val screenWidth = screen.measuredWidth
+
+            // TODO: first value available when entering, second when exiting; last fallback
+            //       is off by header's height
+            val screenHeight = screen.container?.height ?: screenParent?.height ?: screen.measuredHeight
```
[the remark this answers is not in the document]
kkafar: I dunno whether these dimensions are accurate or not, depending on whether the header is present. I'll look into it when testing.

android/src/main/java/com/swmansion/rnscreens/animator/CustomAnimatorSetProvider.kt
```diff
@@ -0,0 +1,239 @@
...
+                        ObjectAnimator
+                            .ofFloat(screenParent, "translationX", 0.1f * screenWidth, 0.0f)
+                            .apply {
+                                duration = 450
+                                interpolator = AnimationUtils.loadInterpolator(context, android.R.interpolator.fast_out_extra_slow_in)
+                            },
+                    )
+                }
+                R.animator.rns_default_enter_out -> {
+                    finalAnimatorSet.playTogether(
```
kkafar, line 42: Why is some places do you use `playTogether` and `play` in the others?

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
```diff
@@ -263,57 +265,97 @@ class ScreenStackFragment :
...
+                R.animator.rns_default_exit_in,
+                R.animator.rns_default_enter_out,
+                R.animator.rns_default_enter_in,
+                R.animator.rns_no_animation_20,
+                R.animator.rns_fade_out,
+                R.animator.rns_fade_in,
+                R.animator.rns_slide_out_to_left,
+                R.animator.rns_slide_in_from_left,
+                R.animator.rns_slide_out_to_right,
+                R.animator.rns_slide_in_from_right,
```
kkafar, line 282: I'm also not convinced that this is the place to customise this. What we need is cohesive approach to setting up animations. Right now we set them both here & in `ScreenStack.onUpdate`. This is kinda bad & we need to come up with some clear abstraction here. I don't have clear idea in mind yet. Will come back here once I think this over.

android/src/main/res/base/animator/rns_ios_from_left_background_close.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<set xmlns:android="http://schemas.android.com/apk/res/android">
+
+</set>
```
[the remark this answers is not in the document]
kkafar: Let's get rid of the empty files here. Also -> let's abstract the animator code implementations to separate files / or single file but separate type for each animator.

kkafar (commented):
> should we change all animations to Animator (current version in this PR) or rather detect a transition from formSheet to push screen and apply Animator only in that case, leaving tween animations for regular transistions (there might be difference in performance on older devices)

No. We want only the push over sheet / return from push to sheet transitions to use the animators for now. 

The aniamtor-based animations have numerous issues in our setup (talking about sheet enter animation here), I'm pretty sure the same issues apply to implementations introduced by this PR. Therefore we want to keep behaviour impact at minimal level & apply it only to the behaviour where it is really needed.

> should we implement these transitions using XML Resource files if they don't allow to define the whole animation (e.g. we can't set translation by percentage of width/height) and we have to do modify them in code

That is a good question. I would prefer the whole animator definition in single place & would like to avoid situations where part of configuration is contained in XML & we apply more logic after loading. Unless we introduce some abstraction over it. 

I don't have good suggestions atm though. Will try to come up with something by next week.

android/src/main/java/com/swmansion/rnscreens/animator/CustomAnimatorSetProvider.kt
kkafar: okay, I see.

The way to go here, would be to keep definition of an animation in single place. Let's not split them into XML & code part.

android/src/main/java/com/swmansion/rnscreens/animator/CustomAnimatorSetProvider.kt
```diff
@@ -0,0 +1,239 @@
...
+                    )
+                }
+                R.animator.rns_default_exit_out -> {
+                    finalAnimatorSet.playTogether(
+                        initialAnimatorSet,
+                        ObjectAnimator
+                            .ofFloat(screenParent, "translationX", 0.0f, 0.1f * screenWidth)
+                            .apply {
+                                duration = 450
+                                interpolator = AnimationUtils.loadInterpolator(context, android.R.interpolator.fast_out_extra_slow_in)
```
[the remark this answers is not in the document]
kligarski: It is used only in `v33` animations so I think we're good.

kligarski:
In previous version, when we used new Animators for all animations, `events.e2e.test` failed - some events might be missing. It passes now because there are no tests for events with formSheet but we should take a look whether this PR introduces any changes for emitting events.

android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt
```diff
@@ -297,7 +297,7 @@ open class ScreenFragment :
             // onViewAnimationStart/End is triggered from View#onAnimationStart/End method of the fragment's root
             // view. We override an appropriate method of the StackFragment's
             // root view in order to achieve this.
-            if (isResumed) {
+            if (!fragment.isRemoving) {
```
kligarski, line 300: We'll need to test that this does not change any behavior but it seems reasonable - in `ScreensCorrdinatorLayout`'s `startAnimation` we use `fragment.isRemoving` as well.

android/src/main/java/com/swmansion/rnscreens/events/ScreenAnimationDelegate.kt
kligarski, line 1:
```kotlin
private val eventEmitter: ScreenEventEmitter?
private val animationType: AnimationType
```

These properties are unused now but I'm not sure if we want to remove them.

kligarski (commented):
This commit fixes problems with `onViewAnimation{Start,End}` mentioned in https://github.com/software-mansion/react-native-screens/pull/2682. 

`isResumed` is `false` for formSheets and now for regular screens with Animator-based animation. This happens because Animators impact Fragment's lifecycle but Animations do not. Here are some references I found:

> When the fragment is visible, all [Animator](https://developer.android.com/reference/android/animation/Animator) and [Transition](https://developer.android.com/reference/androidx/transition/Transition) effects have finished, and the fragment is ready for user interaction. The fragment's Lifecycle moves to the RESUMED state, and the [onResume()](https://developer.android.com/reference/androidx/fragment/app/Fragment#onResume()) callback is invoked.

https://developer.android.com/guide/fragments/lifecycle#fragment_and_view_resumed

> Entering Animations will not be part of the fragment lifecycle. So your fragment will not wait until an entering Animation is complete to move to RESUME.

https://issuetracker.google.com/issues/167579557?hl=it

kligarski:
There are 2 more problems that need investigating, they do not happen on main.

| Nested stack flash | Bottom tabs disappearing |
| -- | -- |
| <video src="https://github.com/user-attachments/assets/73f667d8-d2d3-43c4-bb0d-5e10617186f1" alt="nested stack flash" /> | <video src="https://github.com/user-attachments/assets/5fda05d1-ebac-4a4d-becd-1b625f7c6e8e" alt="bottom tabs disappearing" />  |

kligarski (commented):
When we're going back from nested stack (from Nested Home to Home):

| Stack1 | Stack 2 |
| ------  | -------  |
| Home | |
| NestedStackHost | NestedHome | 

Stack2 has `stackAnimation == null`, because animation is handled by Stack1. 

Previously `stackAnimation == null` would mean that we did not use [setCustomAnimations()](https://developer.android.com/reference/androidx/fragment/app/FragmentTransaction#setCustomAnimations(int,%20int)) on FragmentTransaction. Now we're setting regular animations using `onCreateAnimation`. 

Before this commit we would return default animation from `ScreenStackAnimationManager` which was `StackAnimation.NONE` but we're supposed to return `null` to match previous behaviour.

Issue #2809, opened 2025-03-25
backButtonDisplayMode is not working when headerBackTitle is used


maciekstosio:
### Description

While working on `headerBackButtonDisplayMode`, I noticed that the `headerBackButtonDisplayMode` behavior is skipped, when `headerBackTitle` is provided. I tested it on plan UIKit (without RN) and it suppose to work.

https://github.com/user-attachments/assets/3130b9df-2557-449e-9d23-cf569cd76ca7

### Steps to reproduce

Open Test1084 and set on second screen: 
```
            headerBackTitle: 'Banana',
            headerBackButtonDisplayMode: 'generic',
```
when using `generic` "Back" or just an arrow should be displayed instead of "Banana", when using `default` the back title should change between "Banana", "Back", and just an icon depending of the size of the title. Currently, when the `backTitle` is to large it moves the title. 

### Snack or a link to a repository

☝️

### Screens version

4.10.0-beta.3

### React Native version

0.78.0-rc.5

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

PR #2848, opened 2025-04-08
chore(iOS, Fabric): add tests for header interactions

maciekstosio:
## Description

This PR adds tests for #2809, #1589 and https://github.com/software-mansion/react-native-screens/pull/2800.

## Changes

The main assumption is to test all permutation of headerBackButtonMenuEnabled, headerBackButtonDisplayMode and content of button (default - taken from previous screen headerTitle, custom from headerBackTitle and styled when there are styles applied). I hope this will help us tracking regression as this part of the code is a bit confusing and caused us problem in the past. 
Additionally there are added "custom" scenarios that test default default behavior with default text and custom text (if the back button changes it's appearance depending on available space and [headerBackButtonDisplayMode](https://developer.apple.com/documentation/uikit/uinavigationitem/backbuttondisplaymode-swift.property) specification). 
The some of the last cases are set to failing, and this is known disparity reported in #2809, we plan to fix it soon after, but wonted to make sure we don't cause a regression.

**Remark:**
I did it in a way that after each test we're returning to the screen with all test cases for 2809, this is due to the fact that scrolling through list of all Test* takes a lot and reloading RN after each test would cause huge overhead. The downside of it is that if any of the test fails if will cascade fail all other. 

You can run test with: 
```
detox build --configuration ios.sim.debug
detox test --configuration ios.sim.debug e2e/issuesTests/Test2809.e2e.ts -l debug
````

## Checklist

- [ ] Ensured that CI passes

### CODE REVIEW

File: FabricExample/e2e/issuesTests/Test2809.e2e.ts [hunk 1]
```diff
@@ -0,0 +1,267 @@
...
+     .whileElement(by.id('root-screen-examples-scrollview'))
+     .scroll(600, 'down', NaN, 0.85);
+
+   await expect(element(by.id('root-screen-tests-Test2809'))).toBeVisible();
+   await element(by.id('root-screen-tests-Test2809')).tap();
+ });
+
+   describe('backButtonMenuEnabled: true', () => {
+      describe('backButtonDisplayMode: default', () => {
+         it('with default text', async () => {
```
kkafar, line 55:
```suggestion
         it('default back title stays visible', async () => {
```
Just one note to all the tests - would be nice if the string in `it` better described our expectations regarding the test, so that it is more readable from terminal what the test does. I'm not claiming that my suggestion is even good, but would be nice to come up with something more readable.

FabricExample/e2e/issuesTests/Test2809.e2e.ts
```diff
@@ -0,0 +1,267 @@
...
+            await expectBackButtonMenuToNotExistOnLabel('First'); // Check if backButtonMenu is disabled
+            await element(by.text('Pop to top')).tap(); // Go back
+         });
+
+         it('with custom text', async () => {
+            await element(by.text('DisabledDefaultCustomText')).tap();
+            await element(by.text('Open screen')).tap();
+            await expect(element(by.text('Custom'))).toBeVisible();
+            await expectBackButtonMenuToNotExistOnLabel('Custom'); // Check if backButtonMenu is disabled
+            await element(by.text('Pop to top')).tap(); // Go back
```
kkafar, line 157: I think this can be refactored out  to a function with some parameters, right? We do not need to repeat it every single time.

FabricExample/e2e/issuesTests/Test2809.e2e.ts
```diff
@@ -0,0 +1,267 @@
...
+            await expect(element(by.id('chevron.backward'))).toBeVisible();
+            await expect(element(by.text('First'))).not.toBeVisible();
+            await expectBackButtonMenuIconAndLabel('chevron.backward', 'First'); // Check if backButtonMenu works
+            await element(by.text('Pop to top')).tap(); // Go back
+         });
+      });
+   });
+
+   // Custom
+   it('Default long back label should be truncated to generic by buckButtonDisplayMode', async () => {
```
kkafar, line 230: These titles are very nice

kkafar (changes requested):
The initiative is great. We definitely need this.

This looks very solid, I've just left few style-related remarks. 

Great job.

FabricExample/e2e/issuesTests/Test2809.e2e.ts
maciekstosio: changed in: https://github.com/software-mansion/react-native-screens/pull/2848/commits/d1c0127fc0c3b102537b45e840f539d3f7acc3c6

FabricExample/e2e/issuesTests/Test2809.e2e.ts
maciekstosio: changed in: https://github.com/software-mansion/react-native-screens/pull/2848/commits/d1c0127fc0c3b102537b45e840f539d3f7acc3c6

kkafar (approved):
Great job!

PR #2850, opened 2025-04-08
test(e2e, Fabric): add e2e tests for issue/PR examples 619..640 

kligarski:
## Description

Check which example screens from issues/PRs can be used in e2e testing for tests `Test619, ..., Test640` and implement them if possible for Fabric. 

### Test619
Skipped because we can't check components "jumping".

### Test624
Skipped because the change in PR https://github.com/software-mansion/react-native-screens/pull/624 is connected to internal JS prop. The test screen seems to be repurposed.

### Test640
Test created. It checks if `RefreshControl` works correctly with `modal` (refresh instead of the modal closing).

## Changes

- add `Test640`

CI

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/FabricExample/e2e/issuesTests/Test640.e2e.ts
+++ b/FabricExample/e2e/issuesTests/Test640.e2e.ts
@@ -1,0 +1,30 @@
+import { device, expect, element, by } from 'detox';
+
+describe('Test640', () => {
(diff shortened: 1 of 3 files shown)


kkafar (approved):
This looks solid 💪🏻 Good job.

PR #2853, opened 2025-04-09
chore(CI): Update node version on CI, bump react-navigation and fix windows pipeline

maciekstosio:
## Description

Update node version on CI, bump react-navigation to latest main and fix windows pipeline (https://github.com/software-mansion/react-native-screens/pull/2857).

## Changes
- Fix windows pipeline (https://github.com/software-mansion/react-native-screens/pull/2857)
- Bump node version to latest stable (23.11) as 18 used on CIs was outdated with what's necessary react-native-buildr-bob used by react-navigation 
- Move node version to `.nvmrc` and use it as source of truth for GitHub workflows 
- additionally: cd FabricExample/ios && pod install

## Checklist

- [ ] Ensured that CI passes


FabricExample/ios/Podfile.lock
```diff
@@ -2013,72 +2013,72 @@ SPEC CHECKSUMS:
...
+  RNReanimated: bc7be90dec8df273a2d14da13a64f9da4b694e59
+  RNScreens: e33d663f09210e3b2ad85969fd71d6b35c05231b
   SocketRocket: d4aabe649be1e368d1318fdf28a022d714d65748
-  Yoga: 330be28eee1242da875db9e851b19a4df496b999
+  Yoga: 76eca883712bbdd12fcc754f2342f3ca307adb44
 
 PODFILE CHECKSUM: 9368f39644a8576a848701c298cb4a4fd39a41bf
 
-COCOAPODS: 1.15.2
+COCOAPODS: 1.14.3
```
kkafar, line 2084: I think you need to update cocoapods 😄

kkafar (changes requested):
I see on CI that the `yarn submodules` repository step is failing after the update. Do we know why is it the case? 

We don't want to merge it it current state.

.github/workflows/android-build-test-fabric.yml
```diff
@@ -32,10 +32,10 @@ jobs:
           java-version: '17'
           distribution: 'zulu'
           cache: 'gradle'
-      - name: Use Node.js 18
+      - name: Setup Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: 18
+          node-version-file: '.nvmrc'
```
kkafar, line 38: That is only kinda ok. I'm worried of cases, where different workflows require different node versions. But I guess we'll handle these cases separately then.

.github/workflows/android-build-test.yml
```diff
@@ -30,10 +30,10 @@ jobs:
           java-version: '17'
           distribution: 'zulu'
           cache: 'gradle'
-      - name: Use Node.js 18
+      - name: Setup Node.js
```
kkafar, line 33:
```suggestion
      - name: Setup Node.js (version from .nvmrc)
```

just a nit pick

react-navigation
kkafar, line 1: I don't see this commit:

* https://github.com/react-navigation/react-navigation/commit/df49c8a3a154fc1308a2cad71c6e2d08a2c089b7

here & we want it

kkafar (changes requested):
I want this commit to be included:

* https://github.com/react-navigation/react-navigation/commit/df49c8a3a154fc1308a2cad71c6e2d08a2c089b7


& please change PR title to align with conventional commit naming system.

.github/workflows/android-build-test-fabric.yml
maciekstosio: Yeah, I wouldn't worry about it in advance, we have been using one hardcoded version for long time and it was ok.

react-navigation
maciekstosio: changed in: https://github.com/software-mansion/react-native-screens/pull/2853/commits/95fa1180f0ba26c12ae7cb32da51f1b20e40fab1

.github/workflows/android-build-test.yml
maciekstosio: changed in: https://github.com/software-mansion/react-native-screens/pull/2853/commits/95fa1180f0ba26c12ae7cb32da51f1b20e40fab1

.github/workflows/windows-build-test.yml
```diff
@@ -57,8 +57,23 @@ jobs:
...
+          if ($rnwVersion -match '(\d+)\.(\d+)\.') {
+              $major = $matches[1]
+              $minor = $matches[2]
+              $desiredVersion = "$major.$minor"
+              echo "Installing $desiredVersion version."
+              echo "rn=$desiredVersion" >> $env:GITHUB_OUTPUT
+          } else {
+              echo "Failed to parse version from npm. Installing latest version."
+              echo "rn=latest" >> $env:GITHUB_OUTPUT
+          }
```
kkafar, line 73: I see what it does, but I have idea what language is this 😅

It is fine though, asking out of curiosity.

kkafar (commented):
Okay, we're looking good now! Let's wait for the CI before merging.

.github/workflows/windows-build-test.yml
kligarski: PowerShell

kkafar: If only bash had looked like this...

maciekstosio:
I confirmed that old arch android passes locally 
image

PR #2861, opened 2025-04-11
chore(iOS): refactor back button logic

maciekstosio:
## Description / Changes

This is continuation of https://github.com/software-mansion/react-native-screens/pull/2860 that proposes small refactor around back button logic. This area is very confusing and in my opinion this is due to flag `shouldUseCustomBackBarButtonItem` that breaks the flow of other if's and causes other props to not work. Not we contained this flag in `if (visible)` which currently is the only case when backButtonItem might be used after the changes from #2800.

see `FabricExample/e2e/issuesTests/Test2809.e2e.ts` and `apps/src/tests/Test2809/index.tsx`

## Checklist
- [ ] Ensured that CI passes


maciekstosio:
Closed in favor of https://github.com/software-mansion/react-native-screens/pull/2867

PR #2867, opened 2025-04-11
refactor(iOS): limit scope of `shouldUseCustomBackBarButtonItem` flag

maciekstosio:
## Description / Changes

This is continuation of https://github.com/software-mansion/react-native-screens/pull/2860 that proposes small refactor around back button logic. This area is very confusing and in my opinion this is due to flag `shouldUseCustomBackBarButtonItem` that breaks the flow of other if's and causes other props to not work. Not we contained this flag in `if (visible)` which currently is the only case when backButtonItem might be used after the changes from #2800.

see `FabricExample/e2e/issuesTests/Test2809.e2e.ts` and `apps/src/tests/Test2809/index.tsx`

## Checklist
- [ ] Ensured that CI passes


kkafar (changes requested):
I don't think this works.

Consider scenario: `config.disableBackButtonMenu == true && config.isBackTitleVisible == false`

You won't use custom back button then -> you will break the behaviour of `disableBackButtonMenu` prop. We can't merge this.

maciekstosio:
But it didn't work before the this changes too. When  `backTitleVisible == false`, then in else statement we set `shouldUseCustomBackBarButtonItem = NO` and `backButtonItem` won't be set, thus we use default `UIBarButtonItem` implementation **with** backButtonMenu

kkafar (approved):
I think we're good. Thanks!

PR #2875, opened 2025-04-15
test(e2e, Fabric): add e2e tests for issue/PR examples 642..645

kligarski:
## Description

Check which example screens from issues/PRs can be used in e2e testing for tests `Test642, Test645` and implement them if possible for Fabric. 

### Test642
Skipped because we can't check status bar visibility/style.

### Test645
Test created, only on iOS (`headerLargeTitle` is supported only on iOS). It checks if `headerLargeTitle` isn't collapsed when changing screens. I changed `Test645` screen to use `ScrollView` because it was used in original issue.

## Changes

- add `Test645`
- change regular `View` to `ScrollView` in test screen `Test645`

CI

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/FabricExample/e2e/issuesTests/Test645.e2e.ts
+++ b/FabricExample/e2e/issuesTests/Test645.e2e.ts
@@ -1,0 +1,48 @@
+import { device, expect, element, by } from 'detox';
+import { describeIfiOS } from '../e2e-utils';
+
+// headerLargeTitle is supported only on iOS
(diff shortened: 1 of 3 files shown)


kligarski:
Need to check whether assertions will fail on collapsed header. Also, the original issue was connected with `ScrollView`s but they are not used in the test screen - we should probably add them.

kkafar (approved):
Great, thanks!

Issue #2688, opened 2025-02-12
fitToContents does not fit to contents


RohovDmytro:
### Description

Layout changes does not reflect the height of the form sheet on iOS.

### Steps to reproduce

1. https://github.com/RohovDmytro/react-native-screams

2. Open example "fofrm-sheet-fit-to-content-not"

3. Wait 1000 ms.

### Snack or a link to a repository

https://github.com/RohovDmytro/react-native-screams

### Screens version

4.7.0-beta.4

### React Native version

0.76.7

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Acknowledgements

Yes

gremech:
This is the condition that prevents the frame from being resized:
```diff --git a/node_modules/react-native-screens/ios/RNSScreen.mm b/node_modules/react-native-screens/ios/RNSScreen.mm
index 8481d21..54a70c8 100644
--- a/node_modules/react-native-screens/ios/RNSScreen.mm
+++ b/node_modules/react-native-screens/ios/RNSScreen.mm
@@ -437,7 +437,7 @@ - (BOOL)registerContentWrapper:(RNSScreenContentWrapper *)contentWrapper content
 /// This is RNSScreenContentWrapperDelegate method, where we do get notified when React did update frame of our child.
 - (void)contentWrapper:(RNSScreenContentWrapper *)contentWrapper receivedReactFrame:(CGRect)reactFrame
 {
-  if (self.stackPresentation != RNSScreenStackPresentationFormSheet || _didSetSheetAllowedDetentsOnController == YES) {
+  if (self.stackPresentation != RNSScreenStackPresentationFormSheet) {
     return;
   }
 
```
It might solve some other issues. When I removed it, I noticed an unexpected header animation on iPads. React frame width value alternates between correct and incorrect:
```
reactFrame	CGRect	(origin = (x = 0, y = 0), size = (width = 1133, height = 151.5)) ### Incorrect
reactFrame	CGRect	(origin = (x = 0, y = 0), size = (width = 580, height = 151.5)) ### Correct
reactFrame	CGRect	(origin = (x = 0, y = 0), size = (width = 1133, height = 151.5))) ### Incorrect
reactFrame	CGRect	(origin = (x = 0, y = 0), size = (width = 580, height = 151.5)) ### Correct
```

[...]

PR #2877, opened 2025-04-17
feat(iOS): make `fitToContents` sheet height react to dynamic content 

kligarski:
## Description

Allows `fitToContents` sheet's height to adapt to dynamic content on iOS.

`contentWrapper:(RNSScreenContentWrapper *)contentWrapper receivedReactFrame:(CGRect)reactFrame` receives many frames, some of them contain the same dimensions. Calling `[self setAllowedDetentsForSheet:sheetController to:detents animate:YES]` mutliple times, even with the same `detents`, causes visual glitches (dimming behing formSheet on iPhone/iPad and header on iPad have very odd-looking animations).

Before, we would set `_didSetSheetAllowedDetentsOnController` to `YES` after setting detents for the first time and then ignore all incoming frames - including those with new height after content changed its height. Now, we remember frame height we received, and when we receive a new frame, we compare new height to the previous value. If it is different, we update detents using `setAllowedDetentsForSheet`.

Supersedes https://github.com/software-mansion/react-native-screens/pull/2741.

Fixes https://github.com/software-mansion/react-native-screens/issues/2688.

## Changes

- replace boolean property `_didSetSheetAllowedDetentsOnController` with float `_sheetFrameHeight`
- add `Test2877` test screen

## Screenshots / GIFs

| Before | After |
| ------ | ------ |
| <video src="https://github.com/user-attachments/assets/52895e55-483d-4766-bca2-7550757f1868" alt="before" /> | <video src="https://github.com/user-attachments/assets/d3bf33c0-5186-40e5-9b41-77aa31936cec" alt="after" /> |

Enter `Test2877` screen and open form sheet. Its height should adapt to displayed content every second.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/FabricExample/e2e/issuesTests/Test2877.e2e.ts
+++ b/FabricExample/e2e/issuesTests/Test2877.e2e.ts
@@ -1,0 +1,37 @@
+import { device, expect, element, by } from 'detox';
+import { describeIfiOS } from '../e2e-utils';
+
+// issue fixed only on iOS at the moment
+describeIfiOS('Test2877', () => {
+  beforeAll(async () => {
+    await device.reloadReactNative();
+  });
+
+  it('Test2877 should exist', async () => {
+    await waitFor(element(by.id('root-screen-tests-Test2877')))
+      .toBeVisible()
+      .whileElement(by.id('root-screen-examples-scrollview'))
+      .scroll(600, 'down', NaN, 0.85);
+
+    await expect(element(by.id('root-screen-tests-Test2877'))).toBeVisible();
+    await element(by.id('root-screen-tests-Test2877')).tap();
+  });
+
+  it('formSheet should open without hidden content', async () => {
+    await element(by.id('home-button-open-formsheet')).tap();
+
+    await expect(element(by.text('Like this.'))).not.toBeVisible();
+  });
+
+  it('formSheet should adapt height to appearing content, making it visible after 1 second', async () => {
+    await waitFor(element(by.text('Like this.')))
+      .toBeVisible(100)
+      .withTimeout(1200);
+  });
+
+  it('additional content in the formSheet should hide after 1 second', async () => {
+    await waitFor(element(by.text('Like this.')))
+      .not.toBeVisible(100)
+      .withTimeout(1200);
+  });
+});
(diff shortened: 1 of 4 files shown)


kligarski:
In the future, it might be necessary to investigate the issue in more detail: why updating detents causes the glitches to appear and why new frames have same height but different width on an iPad (it sometimes alternates between 2 values, e.g. `1300 -> 500 -> 1300 -> 500`).

kkafar (commented):
This is great.

I will want to put some time into debugging here, so that we understand the changes 😅 

I'll try to find some time next week.

lodev09:
> In the future, it might be necessary to investigate the issue in more detail: why updating detents causes the glitches to appear and why new frames have same height but different width on an iPad (it sometimes alternates between 2 values, e.g. 1300 -> 500 -> 1300 -> 500).

This might not be related but I found a bug while working on new features for the `formSheet` presentation. In New Arch, every layout changes triggers `updateFormSheetPresentationStyle` and causing visual glitches. This is due to the method being called in `finalizeUpdates` block.

If you move the prop update to `updateProps`, it fixes the issue.

kkafar:
@lodev09 thanks! Do you have any more specific information, e.g. when this issue is triggered? How can we observe the visual glitches?

kkafar (approved):
This looks solid. I've debugged it for few minutes now & seems that all frame updates are sensible & for a known reason. 

I've allowed myself to improve example & do a small variable renaming.

lodev09:
@kkafar my fork has some added features like the `translateY` event among others. while working on this, I found a performance issue in New Arc in this block:
https://github.com/software-mansion/react-native-screens/blob/d1e9aaffe3d3e2848ad8fc7cd600a3a84b2ace71/ios/RNSScreen.mm#L1326-L1332

Every time the sheet is dragged, this block is being called. I have a fix for it in my fork, together with some other features.
I will submit PR soon, currently trying to find solution on the footer issue on IOS.

https://github.com/lodev09/react-native-screens

For the context, had to implement these changes due to requirement for our app. Feel free to check it out in the meantime if get a chance.

https://github.com/user-attachments/assets/823440f2-545e-4956-b58b-bb246072b200

lodev09:
Created draft PR here:
https://github.com/software-mansion/react-native-screens/pull/2902

PR #2883, opened 2025-04-18
refactor: simplify props in `HeaderConfig` as `ViewProps` already include children

kkafar:
## Description

Simplify props in header config as ViewProps already include children.

Exactly the same type of props should be accepted by `ScreenStackHeaderSubviews`.

## Checklist

- [ ] Ensured that CI passes


kligarski (commented):
Looks good but shouldn't we apply these changes in `ScreenStackHeaderConfig.web.tsx` as well?

PR #2886, opened 2025-04-23
refactor(web): cleanup types in ScreenStackHeaderConfig

kkafar:
## Description

Follow up to #2833, review: https://github.com/software-mansion/react-native-screens/pull/2883#pullrequestreview-2786094325

## Changes

`ViewProps` already do have children, there is no need to specify it for the second time. 

Web should work as before.

## Checklist

- [ ] Ensured that CI passes


kligarski (approved):
Looks good.

PR #2887, opened 2025-04-23
chore(CI): add nightly publish workflow

arthwood:
## Description

The workflow is inspired by `react-native-reanimated`.

N/A

## Checklist

- [ ] Ensured that CI passes

## Description
The goal is to have nightly build workflow.

## Changes
Added all the scripts required in order to run the workflow.

It should be possible to trigger workflow manually. It will also run on schedule.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


src/platform-specific/jsVersion.ts
tomekzaw, line 1: This file is not needed in react-native-screens.

.github/workflows/npm-screens-build.yml
```diff
@@ -0,0 +1,123 @@
+name: NPM screens build
```
tomekzaw, line 1:
```suggestion
name: npm package build
```

.github/workflows/npm-screens-publish-nightly.yml
```diff
@@ -0,0 +1,25 @@
+name: NPM Screens publish [Nightly]
```
tomekzaw, line 1:
```suggestion
name: Build nightly and publish on npm
```

.github/workflows/npm-screens-publish-nightly.yml
```diff
@@ -0,0 +1,25 @@
...
+    secrets: inherit
+
+  npm-screens-3-publish-nightly:
+    if: github.repository == 'software-mansion/react-native-screens'
+    uses: ./.github/workflows/npm-screens-build.yml
+    with:
+      option: --nightly
+      publish_on_npm: true
+      ref: 3.x
+    secrets: inherit
```
tomekzaw, line 25: No need to publish nightlies for 3.x

createNPMPackage.sh
```diff
@@ -0,0 +1,19 @@
...
+    exit 1
+  fi
+fi
+
+yarn build
+
+npm pack
+
+if [ $# -ge 1 ]; then
+  node scripts/set-screens-version.js "$CURRENT_VERSION" >/dev/null
```
tomekzaw, line 16: I don't think we need to include package name in the script name
```suggestion
  node scripts/set-version.js "$CURRENT_VERSION" >/dev/null
```

scripts/releasing.js
```diff
@@ -0,0 +1,100 @@
...
+function getVersion(rawArgs, packageJsonPath) {
+  const args = rawArgs.slice(2).map((arg) => ({
+    value: arg,
+    handled: false,
+  }));
+
+  let IS_HELP = false;
+  let IS_NIGHTLY = false;
+  let IS_FRESH = false;
+  let IS_SET_CUSTOM = false;
```
tomekzaw, line 17: This logic doesn't need to be that complex, we can hardcode it to support only the nightly version.

scripts/set-screens-version.js
```diff
@@ -0,0 +1,29 @@
...
+const jsVersionPath = path.resolve(
+  __dirname,
+  '../src/platform-specific/jsVersion.ts'
+);
+const before = fs.readFileSync(jsVersionPath, 'utf-8');
+const after = before.replace(
+  /jsVersion = '(.*)';/g,
+  `jsVersion = '${newVersion}';`
+);
+fs.writeFileSync(jsVersionPath, after, 'utf-8');
```
tomekzaw, line 26:
```suggestion
```

.github/workflows/npm-screens-publish-nightly.yml
```diff
@@ -0,0 +1,65 @@
...
+      - name: Upload npm package to GitHub
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ env.PACKAGE_NAME }}
+          path: './${{ env.PACKAGE_NAME }}'
+
+      - name: Publish npm package
+        run: npm publish $PACKAGE_NAME --tag nightly --provenance
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NODE_AUTH_TOKEN }}
```
kkafar, line 65: Where does this secret come from? I guess we have to generate it & put into GitHub secrets system, right?

createNPMPackage.sh
kkafar, line 1: Can we put this script into `scripts/` directory? I don't think there is a need to keep it in top level dir. 

Just remember to adjust relative paths wherever needed.

createNPMPackage.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+yarn install --immutable
+
+if ! CURRENT_VERSION=$(node scripts/set-version.js); then
+  exit 1
+fi
+
+yarn build
```
kkafar, line 9: I don't think we even have `build` script here in the repo. We use `yarn prepare`.

createNPMPackage.sh
```diff
@@ -0,0 +1,15 @@
...
+
+npm pack
+
+node scripts/set-version.js "$CURRENT_VERSION" >/dev/null
```
kkafar, line 13: If we pack, before changing version in `package.json` we'll use old version for the artifact, right?

scripts/releasing.js
```diff
@@ -0,0 +1,37 @@
+const { cat, exec } = require('shelljs');
```
kkafar, line 1: I wasn't aware that we have dependency for this package present in our repo. I've inspected `yarn.lock` & it seems that it is transitive dependency of our package through `react-native-windows` & `release-it`. To be honest, these are pretty stable dependencies, but I don't think we should rely on their presence. We should specify the dependency on things we're using directly.

kkafar (changes requested):
Thanks a lot for the contribution here. We'll surely want to land it . I have some remarks that need to be addressed before we can proceed though 👇🏻 

Also, I have questions about secrets management: 

1. Do we need to add some dummy user to npm & generate secrets for it, so that the package can be effectively published? 
2. Same with GitHub secrets - what else do we need to do, to make package publish to GitHub successful? 

I have no prior experience with ☝🏻 and want to understand ins and outs before we land it.

scripts/releasing.js
kkafar: image

.github/workflows/npm-screens-publish-nightly.yml
tomekzaw: You need to generate a npm token on npm website, it can be either yours or we can create `softwaremansion` account on npm for this case.

arthwood: Yes, that's the action repository secret. And the value is npm access token. I think we can use our "official" npm user and not the "dummy" one.

kkafar (commented):
Okay, I think it looks good now. Great job. We'll test & land this once I get appropriate secrets setup (WIP).

kkafar (changes requested):
Hey @arthwood! One last thing before we proceed. Beside that, I think we're good.

kkafar (commented):
On paper this looks good, let's test it in action now.

kkafar:
Since it is on your fork, & not directly in the repo, it seems I can not trigger the workflow manually.

I'll merge the PR to some other branch belonging directly to the repo & then test it.

PR #2891, opened 2025-04-29
fix(CI): publish nightly - incorrect relative paths to scripts & missing build.log

kkafar:
## Description

Fixing issues making the CI not work!

## Changes

* fixed relative paths to `set-version` script from `create-npm-package` script
* removed job steps of `npm-screens-publish-nightly` workflow, since we do not generate `build.log` file at any point; we're also not including any `node_module` files into a release due to `whitelist` placed in `package.json`.

Let's see whether CI passes

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/.github/workflows/npm-screens-publish-nightly.yml
+++ b/.github/workflows/npm-screens-publish-nightly.yml
@@ -35,18 +35,6 @@ jobs:
         id: build
         run: ./scripts/create-npm-package.sh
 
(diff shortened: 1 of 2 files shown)


kkafar (commented):
Seems to work now! 

https://github.com/software-mansion/react-native-screens/actions/runs/14831670504

PR #2897, opened 2025-04-30
test(e2e, Fabric): add e2e tests for issue/PR examples 648..649

kligarski:
## Description

Check which example screens from issues/PRs can be used in e2e testing for tests `Test648, Test649` and implement them if possible for Fabric. 

### Test648
Skipped because we can't check new animation in any meaningful way.

### Test649
Test created, only on iOS (`headerLargeTitle` is supported only on iOS). It checks if `headerLargeTitle` isn't collapsed when coming back from a screen without `headerLargeTitle`. I changed `Test649` screen to match original issue.

## Changes

- add `Test649`
- change header to normal size on Second screen in `Test649` to match original issue

CI

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/FabricExample/e2e/issuesTests/Test649.e2e.ts
+++ b/FabricExample/e2e/issuesTests/Test649.e2e.ts
@@ -1,0 +1,51 @@
+import { device, expect, element, by } from 'detox';
+import { describeIfiOS } from '../e2e-utils';
+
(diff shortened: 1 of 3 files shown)


kkafar (approved):
LGTM ✅ Thanks!

PR #2900, opened 2025-05-05
test(e2e, Fabric): add e2e test for issue/PR example 658

kligarski:
## Description

Create e2e test for `Test658`.

### Test658
Test created. It checks that opening multiple modals/transparent modals works correctly.

## Changes

- modify `Test658` screen to allow testing:
  - change `dangerouslyGetState()` to `getState()` ([details](https://reactnavigation.org/docs/6.x/upgrading-from-5.x/#dropped-dangerously-from-dangerouslygetparent-and-dangerouslygetstate))
  - move buttons from under status bar
  - add `Text` with screen number
  - add testIDs
- add e2e test for `Test658`
- add comment in `apps/src/tests/index.ts`

CI

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/FabricExample/e2e/issuesTests/Test658.e2e.ts
+++ b/FabricExample/e2e/issuesTests/Test658.e2e.ts
@@ -1,0 +1,197 @@
+import { device, expect, element, by } from 'detox';
+
+const checkScreenVisibility = async function (visible: boolean[]) {
+  for (const [index, shouldBeVisible] of visible.entries()) {
+    const assertionElement = expect(
+      element(
+        by.id('screen-text-added-routes-number').and(by.text(`${index}`)),
+      ),
+    );
+
+    if (shouldBeVisible) {
+      await assertionElement.toBeVisible();
+    } else {
+      await assertionElement.not.toBeVisible();
+    }
+  }
+};
+
+describe('Test658', () => {
+  beforeAll(async () => {
+    await device.reloadReactNative();
+  });
+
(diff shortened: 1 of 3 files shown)


FabricExample/e2e/issuesTests/Test658.e2e.ts
```diff
@@ -0,0 +1,197 @@
+import { device, expect, element, by } from 'detox';
+
+const checkScreenVisibility = async function (visible: boolean[]) {
```
kkafar, line 3:
```suggestion
async function checkScreenVisibility(visible: boolean[]) {
```

kkafar (changes requested):
😄 This one is a bit tedious. I haven't checked the runtime behaviour, trusting you here to get it right!

FabricExample/e2e/issuesTests/Test658.e2e.ts
kligarski: 197879e551537601f5997b3408a7d36a2feaeac2

kligarski:
Yeah, this one is not the most interesting. The test seems to be working fine, I will wait for the CI to pass. We could probably add a lot of different combinations of transparent modals and regular modals but the original issue was mainly about transparent modals - only 2 of them were visible - so I focused on simpler examples in order not to overwhelm our CI in the future :D

kkafar (approved):
That is fine for a moment. Thanks! The lib really needs this!

Issue #2895, opened 2025-04-30
[4.10.0] formSheet slides from top, when it contains Input with autoFocus


plneple:
### Description

When `TextInput` with `autoFocus` is present inside the `Screen` with presentation `formSheet`, the screen slides in from top instead of bottom

### Steps to reproduce

1. Add `TextInput` with `autoFocus`

### Snack or a link to a repository

https://github.com/plneple/react-native-screens/tree/repro/formSheet-autoFocus

### Screens version

4.10.0

### React Native version

0.78.0

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

https://github.com/user-attachments/assets/7092c6eb-6c2c-4df9-b0eb-9812a527ba6d

PR #2909, opened 2025-05-08
fix(Android): fix formSheet sliding from top when it contains input with autofocus

kligarski:
## Description

On API 36, formSheet that contains input with autoFocus would sometimes slide from the top. 

Even though `react-native-screens` used `com.google.android.material:material:1.6.1`, `InsetsAnimationCallback` (which is not present in `1.6.1`) is called (it probably comes from `com.google.android.material:material:1.12.0` dependency in `react-native-edge-to-edge`). This callback interferes with our formSheet entering/closing animations:

1. Values used to calculate start and end position in `InsetsAnimationCallback`'s `onPrepare` and `onStart` are usually incorrect. This has become a significant issue since API 36: for some reason, animation created in `ScreenStackFragment`'s `onCreateAnimator` is sometimes incorrecly accounted for in calculating `startTranslationY`. The value becomes negative which means that the formSheet slides from the top.
2. Both `InsetsAnimationCallback` and formSheet animations use `setTranslationY` and they are overwriting each other's changes. `InsetsAnimationCallback`'s `onProgress` is called just after animation's `setTranslationY` so it takes precedence. 

We decided to replace the window insets animation callback which is set in `BottomSheetBehavior`'s `onLayoutChild` with our own implementation. Currently it is empty in order to bring back previous behavior.

FormSheet that contains input with autoFocus still does not behave as intented - only rarely it gets expanded with the keyboard. This seems to be a timing issue and needs more investigation. It might be necessary to implement logic in new window insets animation callback.

Fixes https://github.com/software-mansion/react-native-screens/issues/2895.

## Changes

- change `com.google.android.material:material` version to `1.12.0`
- replace `BottomSheetBehavior`'s `InsetsAnimationCallback` with empty implementation
- add test screen `Test2895`

Open `Test2895` screen in the example app and open the formSheet.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -12,6 +12,9 @@ import android.view.WindowManager
 import android.webkit.WebView
 import android.widget.ImageView
 import androidx.coordinatorlayout.widget.CoordinatorLayout
+import androidx.core.view.ViewCompat
+import androidx.core.view.WindowInsetsAnimationCompat
+import androidx.core.view.WindowInsetsCompat
 import androidx.core.view.children
 import androidx.fragment.app.Fragment
 import androidx.swiperefreshlayout.widget.SwipeRefreshLayout
@@ -188,6 +191,19 @@ class Screen(
         if (!usesFormSheetPresentation() || !isNativeStackScreen) {
             return
         }
+
+        ViewCompat.setWindowInsetsAnimationCallback(
+            this,
+            object : WindowInsetsAnimationCompat.Callback(
+                DISPATCH_MODE_STOP,
+            ) {
+                override fun onProgress(
+                    insets: WindowInsetsCompat,
+                    runningAnimations: MutableList<WindowInsetsAnimationCompat>,
+                ): WindowInsetsCompat = insets
+            },
+        )
+
         if (coordinatorLayoutDidChange) {
             dispatchShadowStateUpdate(width, height, top)
         }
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -188,6 +191,19 @@ class Screen(
         if (!usesFormSheetPresentation() || !isNativeStackScreen) {
             return
         }
+
+        ViewCompat.setWindowInsetsAnimationCallback(
```
kkafar, line 195: The `onBottomSheetBehaviourDidLayout` function should be considered hot, as it can be called many many times throughout sheet lifecycle & therefore setting new listener object every single time seems like an overkill. Especially that you're creating new callback object every time. 

Let's verify when the listener coming from `Material` library is attached & override it only once.

kkafar (changes requested):
Left some notes that need to be taken into account. Great job overall.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
kligarski: You're right, that was a major oversight. Now I moved the code to `ScreenStackFragment`'s `onCreateView` so it happens only after the first layout (which is when `BottomSheetBehavior` attaches its callback).

7d15be7f909ac69b24d108a6359d3c205db18e3a

kkafar (approved):
Haven't tested the runtime this time. Provided that you did - let's proceed.

PR #2911, opened 2025-05-08
fix(Android): fix formSheet not adjusting for keyboard when it contains input with autofocus

kligarski:
## Description

Sometimes, formSheet that contains input with autoFocus would not account for the keyboard and would remain behind it. This is because `SheetDelegate` is set as a `OnApplyWindowInsetsListener` when fragment is resumed - this happens after finishing formSheet entering animations (animators impact fragment lifecycle). Sometimes, IME insets were applied before this handler was set. Now, we can use `WindowInsetsAnimationCallback` added in [this PR](https://github.com/software-mansion/react-native-screens/pull/2909). Its `onPrepare` method is called before `onApplyWindowInsets` therefore making it possible to set `OnApplyWindowInsetsListener` in time.

## Changes

- change `insetsObserverProxy`'s `listeners` to `HashSet` (we might try to add the same listener multiple times)
- override `onPrepare` method in `WindowInsetsAnimationCallback`
- change visibility of `SheetDelegate` to `internal`

Open `Test2895` screen in the example app, open & close the formSheet (multiple times).

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -527,6 +527,23 @@ class Screen(
         }
     }
 
+    override fun onAttachedToWindow() {
+        super.onAttachedToWindow()
+
+        // Insets handler for formSheet is added onResume but it is often
+        // too late if we use input with autofocus. It happens because onResume is called after
+        // finishing animator animation. onAttachedToWindow is called before onApplyWindowInsets
+        // so we use it to set the handler earlier.
+        // More details: https://github.com/software-mansion/react-native-screens/pull/2911
+        if (usesFormSheetPresentation()) {
+            (fragment as ScreenStackFragment?)?.sheetDelegate?.let {
+                InsetsObserverProxy.addOnApplyWindowInsetsListener(
+                    it,
+                )
+            }
+        }
+    }
+
     private fun dispatchSheetDetentChanged(
         detentIndex: Int,
         isStable: Boolean,
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -544,12 +544,11 @@ class Screen(
     override fun onAttachedToWindow() {
         super.onAttachedToWindow()
 
-        // Insets handler for formSheet is added onResume but it is often
-        // too late if we use input with autofocus. It happens because onResume is called after
-        // finishing animator animation. onAttachedToWindow is called before onApplyWindowInsets
-        // so we use it to set the handler earlier.
-        // More details: https://github.com/software-mansion/react-native-screens/pull/2911
-        if (usesFormSheetPresentation()) {
+        // Insets handler for formSheet is added onResume but it is often too late if we use input
+        // with autofocus - onResume is called after finishing animator animation.
+        // onAttachedToWindow is called before onApplyWindowInsets so we use it to set the handler
+        // earlier. More details: https://github.com/software-mansion/react-native-screens/pull/2911
+        if (usesFormSheetPresentation() && isNativeStackScreen) {
             (fragment as ScreenStackFragment?)?.sheetDelegate?.let {
                 InsetsObserverProxy.addOnApplyWindowInsetsListener(
                     it,
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -31,6 +31,7 @@ import com.swmansion.rnscreens.bottomsheet.useSingleDetent
 import com.swmansion.rnscreens.bottomsheet.usesFormSheetPresentation
 import com.swmansion.rnscreens.events.HeaderHeightChangeEvent
 import com.swmansion.rnscreens.events.SheetDetentChangedEvent
+import com.swmansion.rnscreens.ext.asScreenStackFragment
 import com.swmansion.rnscreens.ext.parentAsViewGroup
 
 @SuppressLint("ViewConstructor") // Only we construct this view, it is never inflated.
@@ -548,8 +549,8 @@ class Screen(
         // with autofocus - onResume is called after finishing animator animation.
         // onAttachedToWindow is called before onApplyWindowInsets so we use it to set the handler
         // earlier. More details: https://github.com/software-mansion/react-native-screens/pull/2911
-        if (usesFormSheetPresentation() && isNativeStackScreen) {
-            (fragment as ScreenStackFragment?)?.sheetDelegate?.let {
+        if (usesFormSheetPresentation()) {
+            fragment?.asScreenStackFragment()?.sheetDelegate?.let {
                 InsetsObserverProxy.addOnApplyWindowInsetsListener(
                     it,
                 )
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
@@ -249,18 +249,6 @@ class ScreenStackFragment :
                 object : WindowInsetsAnimationCompat.Callback(
                     DISPATCH_MODE_STOP,
                 ) {
-                    override fun onPrepare(animation: WindowInsetsAnimationCompat) {
-                        // FormSheet's onResume is called after finishing animator animation but adding
-                        // insets handler then is often too late if we use input autofocus.
-                        // This callback's onPrepare is called before onApplyWindowInsets so we use it
-                        // to set the handler earlier.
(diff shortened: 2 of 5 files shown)


android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
```diff
@@ -249,6 +249,18 @@ class ScreenStackFragment :
                 object : WindowInsetsAnimationCompat.Callback(
                     DISPATCH_MODE_STOP,
                 ) {
+                    override fun onPrepare(animation: WindowInsetsAnimationCompat) {
+                        // FormSheet's onResume is called after finishing animator animation but adding
+                        // insets handler then is often too late if we use input autofocus.
+                        // This callback's onPrepare is called before onApplyWindowInsets so we use it
+                        // to set the handler earlier.
+                        (fragment as ScreenStackFragment?)?.sheetDelegate?.let {
```
kkafar, line 257:
```suggestion
                        this@ScreenStackFragment.sheetDelegate?.let {
```

right? Or do we use some different fragment here?

kkafar (changes requested):
I see that it improves the situation, however, I'm not convinced taht onPrepare is good palce to do so. My initial impression is that  registering observer proxy in `onAttachedToWindow` would make more sense. Can you check, whether it's feasible timing-wise to do so?

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -527,6 +527,23 @@ class Screen(
...
+    override fun onAttachedToWindow() {
+        super.onAttachedToWindow()
+
+        // Insets handler for formSheet is added onResume but it is often
+        // too late if we use input with autofocus. It happens because onResume is called after
+        // finishing animator animation. onAttachedToWindow is called before onApplyWindowInsets
+        // so we use it to set the handler earlier.
+        // More details: https://github.com/software-mansion/react-native-screens/pull/2911
+        if (usesFormSheetPresentation()) {
+            (fragment as ScreenStackFragment?)?.sheetDelegate?.let {
```
kligarski, line 539: I'm not sure if it is the most elegant that in `Screen` we check if we're in a ScreenStack by checking if fragment is ScreenStackFragment.

kligarski (commented):
`onAttachedToWindow` works as well.

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
kligarski: Right, but I moved the code to `Screen` now so I can't really use it there.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
kkafar: there is `isNativeStackScreen` property on screen available.

Just check whether it will work or not, because it has some specific semantics - it checks whether the `screen.container is ScreenStack`, therefore checking the type of a container - note that this might fail when the `screen.container == null`

kkafar (commented):
Answering your question here.

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
```diff
@@ -175,12 +175,20 @@ class SheetDelegate(
...
+
+                // val newMaxHeight =
+                //     if (behavior.maxHeight - keyboardState.height > 1) {
+                //         behavior.maxHeight - keyboardState.height
+                //     } else {
+                //         behavior.maxHeight
+                //     }
+
+                val newMaxHeight = behavior.maxHeight
+
```
kligarski, line 191: Decided to comment out this part because it does not work currently (even before this PR, sheet would extend to previous maxHeight) but it causes visual glitch when **closing** the formSheet.

| Before | After |
| -- | -- |
| <video src="https://github.com/user-attachments/assets/7f3ced6e-fc7f-494d-8c86-66539782a064" alt="before" /> | <video src="https://github.com/user-attachments/assets/c305abc1-49de-4581-b121-e5805e620977" alt="after" />  |

android/src/main/java/com/swmansion/rnscreens/Screen.kt
kligarski: It works, thanks.

78a2576a66a6a3c7baa5551fc4869a6bdf141ee7

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
```diff
@@ -175,12 +175,20 @@ class SheetDelegate(
...
+                // but it creates visual glitch when closing formSheet with keyboard still open.
+                // Decided to temporarily comment it out.
+                // More details: https://github.com/software-mansion/react-native-screens/pull/2911
+
+                // val newMaxHeight =
+                //     if (behavior.maxHeight - keyboardState.height > 1) {
+                //         behavior.maxHeight - keyboardState.height
+                //     } else {
+                //         behavior.maxHeight
+                //     }
```
kkafar, line 188: Yeah, remove this code. If you claim id doesn't serve any purpose. Just, please make sure that you test on 1, 2, & 3 detents.

kkafar (commented):
I see one issue, however: you say that we can rely on `WindowInsetsAnimationCallback`. I say: it is supported since API level 30, while min sdk is lower than that. How does the current apporach look on e.g. SDK 29?

kligarski:
> I see one issue, however: you say that we can rely on WindowInsetsAnimationCallback. I say: it is supported since API level 30, while min sdk is lower than that. How does the current apporach look on e.g. SDK 29?

As we're using `ViewCompat.setWindowInsetsAnimationCallback` to register the callback, it handles the case when using API 21 <= ... < 30 for us.

```java
    static void setCallback(@NonNull View view, @Nullable Callback callback) {
        if (Build.VERSION.SDK_INT >= 30) {
            Impl30.setCallback(view, callback);
        } else if (Build.VERSION.SDK_INT >= 21) {
            Impl21.setCallback(view, callback);
        }
        // Do nothing pre 21
    }
```

But I now noticed that [in the docs](https://developer.android.com/reference/kotlin/androidx/core/view/ViewCompat#setWindowInsetsAnimationCallback(android.view.View,androidx.core.view.WindowInsetsAnimationCompat.Callback)), there is a note about using API < 30:

> Prior to API 30, if an [OnApplyWindowInsetsListener](https://developer.android.com/reference/kotlin/androidx/core/view/OnApplyWindowInsetsListener) is used on the same view, be sure to always use the [ViewCompat](https://developer.android.com/reference/kotlin/androidx/core/view/ViewCompat) version of [setOnApplyWindowInsetsListener](https://developer.android.com/reference/kotlin/androidx/core/view/ViewCompat#setOnApplyWindowInsetsListener(android.view.View,androidx.core.view.OnApplyWindowInsetsListener)), otherwise the listener will be overridden by this method.

I'll try to make sure that everything works on API < 30 as well, because we're sometimes overriding `onApplyWindowInsets` instead of using `setOnApplyWindowInsetsListener`.

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/2911/commits/da28da21eac9ef4d88f1c7cbb1274d5bfd10b62f

kligarski:
With `setOnApplyWindowInsetsListener`, we set the listener on `Screen` which does not override `onApplyWindowInsets` so we're good here. I also didn't find any case where we would use `View.setOnApplyWindowInsetsListener` instead of `ViewCompat` version. The animation looks on API 29 as good as on 36.

https://github.com/user-attachments/assets/ee04e8a1-bfab-4c4a-9873-3a45c493a3e1

kkafar (changes requested):
Once https://github.com/software-mansion/react-native-screens/pull/2925#discussion_r2097640161 this is answered we can proceed

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -548,8 +549,8 @@ class Screen(
         // with autofocus - onResume is called after finishing animator animation.
         // onAttachedToWindow is called before onApplyWindowInsets so we use it to set the handler
         // earlier. More details: https://github.com/software-mansion/react-native-screens/pull/2911
-        if (usesFormSheetPresentation() && isNativeStackScreen) {
-            (fragment as ScreenStackFragment?)?.sheetDelegate?.let {
+        if (usesFormSheetPresentation()) {
+            fragment?.asScreenStackFragment()?.sheetDelegate?.let {
```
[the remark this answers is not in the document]
kkafar: yep, thats it

kkafar (approved):
Looks good now!

PR #2914, opened 2025-05-09
fix: remove global ref when invalidating module

WoLewicki:
## Description

Right now, when removing React Native instance in brownfield apps, the `ScreensModule` cpp side is kept in memory until the whole app is killed. This code removes global ref when invalidating the module too. It fixes memleaks e.g. in Expensify App.

You'd need brownfield app so no repro attached. Try remove and add RN instance to native android app a couple of times, then do a heap dump in AS and check for `ScreensModule` instances with 0 depth in its references.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


kkafar (approved):
makes sense, thanks.

PR #2922, opened 2025-05-15
chore(example): add shared, reusable color values

kkafar:
## Description

Adding shared, reusable color values here. I get a bit frustrated when I have to comeup with colors each time I made an example 
& for some reason `ColorValue` does not have typing for color-literals. 

I've stolen the values from reanimated docs (they use swm color palette), thought it might look nicely & consistent.

I would recommend using these in any future examples.

## Changes

:point_up:

N/A

## Checklist

- [ ] Ensured that CI passes


maciekstosio (approved):

kligarski (approved):
Great idea, struggled with choosing the colors as well.

kligarski (approved):

Issue #2896, opened 2025-04-30
[4.10.0] formSheet does not respect safe area


plneple:
### Description

The full-screen formSheet modal ignores window insets and opens beneath the Status Bar. I haven't found a way to address this from userland.

https://github.com/user-attachments/assets/918e9d97-3825-4284-8f58-6df0fb9cdac7

### Steps to reproduce

1. Open formSheet

### Snack or a link to a repository

https://github.com/plneple/react-native-screens/tree/repro/formSheet-safeArea

### Screens version

4.10.0

### React Native version

0.78.0

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

LeviWilliams:
Confirming there does appear to be some sort of Android status bar safe area issue introduced in 4.10, downgrading is fine. In our case our custom headers set via react navigation render under the status bar until navigating to another screen and back.

PR #2925, opened 2025-05-16
fix(Android): status bar insets for formSheet

kligarski:
## Description

Status bar insets were not handled properly when opening the formSheet with initial height that was greater than safe content area. The bug is directly connected with how `BottomSheetBehavior` works. For unknown reasons, `onChildLayout` sometimes isn't called after `BottomSheetBehavior`'s insets handler runs. In this PR, we override `Screen`'s `setOnApplyWindowInsetsListener` (used by `BottomSheetBehavior` to register its handler), to make sure to request the layout after the listener runs.

### The problem
On first layout, `BottomSheetBehavior` registers the inset listener which updates `insetsTop` field. In `onLayoutChild`, `BottomSheetBehavior` uses `insetsTop` value to calculate `childHeight` which is then used to calculate `fitToContentsOffset`. `fitToContentsOffset` is returned from `getExpandedOffset()` if `isFitToContents == true` (which is the case for single and two detents in `screens`). If `isFitToContents == false` (i.e. we use 3 detents), `getExpandedOffset()` uses `insetsTop` directly. When layout wasn't called after the insets handler, previous value was used which caused the sheet to go under the status bar.

Fixes https://github.com/software-mansion/react-native-screens/issues/2896.

## Changes

- add custom setter for `onApplyWindowInsetsListener`
- add `Test2896` test screen (unable to create e2e test because content behind formSheet still has `visibility = VISIBLE`)

## Screenshots / GIFs
| Nested stack flash | Bottom tabs disappearing |
| -- | -- |
| <video src="https://github.com/user-attachments/assets/527ec68c-e292-4610-b2cb-9066b9d500ef" alt="before" /> | <video src="https://github.com/user-attachments/assets/ac035057-f0b6-4a8d-b844-42eab202927b" alt="after" />  |

Run example app and open `Test2896` test screen.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/InsetsObserverProxy.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/InsetsObserverProxy.kt
@@ -2,6 +2,8 @@ package com.swmansion.rnscreens
 
 import android.util.Log
 import android.view.View
+import androidx.collection.ArraySet
+import androidx.collection.arraySetOf
 import androidx.core.view.OnApplyWindowInsetsListener
 import androidx.core.view.ViewCompat
 import androidx.core.view.WindowInsetsCompat
@@ -10,7 +12,7 @@ import com.facebook.react.bridge.ReactApplicationContext
 import java.lang.ref.WeakReference
 
 object InsetsObserverProxy : OnApplyWindowInsetsListener, LifecycleEventListener {
-    private val listeners: ArrayList<OnApplyWindowInsetsListener> = arrayListOf()
+    private val listeners: ArraySet<OnApplyWindowInsetsListener> = arraySetOf()
     private var eventSourceView: WeakReference<View> = WeakReference(null)
 
     // Please note semantics of this property. This is not `isRegistered`, because somebody, could unregister
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -527,6 +527,23 @@ class Screen(
         }
     }
 
+    override fun onAttachedToWindow() {
+        super.onAttachedToWindow()
+
+        // Insets handler for formSheet is added onResume but it is often
+        // too late if we use input with autofocus. It happens because onResume is called after
+        // finishing animator animation. onAttachedToWindow is called before onApplyWindowInsets
+        // so we use it to set the handler earlier.
+        // More details: https://github.com/software-mansion/react-native-screens/pull/2911
+        if (usesFormSheetPresentation()) {
+            (fragment as ScreenStackFragment?)?.sheetDelegate?.let {
+                InsetsObserverProxy.addOnApplyWindowInsetsListener(
+                    it,
+                )
+            }
+        }
+    }
+
     private fun dispatchSheetDetentChanged(
         detentIndex: Int,
         isStable: Boolean,
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -544,12 +544,11 @@ class Screen(
     override fun onAttachedToWindow() {
         super.onAttachedToWindow()
 
-        // Insets handler for formSheet is added onResume but it is often
-        // too late if we use input with autofocus. It happens because onResume is called after
-        // finishing animator animation. onAttachedToWindow is called before onApplyWindowInsets
-        // so we use it to set the handler earlier.
-        // More details: https://github.com/software-mansion/react-native-screens/pull/2911
-        if (usesFormSheetPresentation()) {
+        // Insets handler for formSheet is added onResume but it is often too late if we use input
+        // with autofocus - onResume is called after finishing animator animation.
+        // onAttachedToWindow is called before onApplyWindowInsets so we use it to set the handler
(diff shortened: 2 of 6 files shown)


android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
```diff
@@ -175,12 +175,20 @@ class SheetDelegate(
...
-                        behavior.maxHeight - keyboardState.height
-                    } else {
-                        behavior.maxHeight
-                    }
+                // This code currently does not work (changing maxHeight requires relayout)
+                // but it creates visual glitch when closing formSheet with keyboard still open.
+                // Decided to temporarily comment it out.
+                // More details: https://github.com/software-mansion/react-native-screens/pull/2911
+
+                // val newMaxHeight =
```
maciekstosio, line 183: Imo if it didn't work, and after removing this part you can't see any problems I wouldn't leave commented code. We have repository to revert back if needed.

maciekstosio (changes requested):

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
kligarski: I started this branch from previous PR that hasn't been merged yet and this change comes from it: https://github.com/software-mansion/react-native-screens/pull/2911#discussion_r2084467656, so I'll handle it there.

android/src/main/java/com/swmansion/rnscreens/InsetsObserverProxy.kt
```diff
@@ -10,7 +12,7 @@ import com.facebook.react.bridge.ReactApplicationContext
 import java.lang.ref.WeakReference
 
 object InsetsObserverProxy : OnApplyWindowInsetsListener, LifecycleEventListener {
-    private val listeners: ArrayList<OnApplyWindowInsetsListener> = arrayListOf()
+    private val listeners: ArraySet<OnApplyWindowInsetsListener> = arraySetOf()
```
kkafar, line 15: > [!note]
> This has been changed, because we had scenarios where some listeners were registered multiple times.

kkafar: I was wrong about semantics of this structure, just looked up the docs - https://developer.android.com/reference/androidx/collection/ArraySet

Let's go with simple `HashSet`.

kkafar (commented):
Beside my wrong suggestion to use ArraySet last time, this one looks good!

android/src/main/java/com/swmansion/rnscreens/InsetsObserverProxy.kt
kligarski: This PR is based on changes from https://github.com/software-mansion/react-native-screens/pull/2911. I'll fix it there and rebase.

maciekstosio (approved):

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -541,6 +542,37 @@ class Screen(
...
+        // with autofocus - onResume is called after finishing animator animation.
+        // onAttachedToWindow is called before onApplyWindowInsets so we use it to set the handler
+        // earlier. More details: https://github.com/software-mansion/react-native-screens/pull/2911
+        if (usesFormSheetPresentation() && isNativeStackScreen) {
+            (fragment as ScreenStackFragment?)?.sheetDelegate?.let {
+                InsetsObserverProxy.addOnApplyWindowInsetsListener(
+                    it,
+                )
+            }
+        }
```
kkafar, line 558: One more thing - form sheet works exclusively with native stack. Do you rely on `isNativeStack` returning false in some case?

kkafar (commented):
I have follow up question that just popped in my head

android/src/main/java/com/swmansion/rnscreens/Screen.kt
kligarski: Finally changed the base branch of this PR, as this part was also from https://github.com/software-mansion/react-native-screens/pull/2911, sorry for making it confusing. 

I'll answer here though: I wasn't sure if checking that a fragment is `ScreenStackFragment` was an elegant solution for verifying if we are a `Screen` in `ScreenStack` and added `isNativeStackScreen` following your suggestion [here](https://github.com/software-mansion/react-native-screens/pull/2911#discussion_r2084329524) but if form sheet works only with native stack, I guess that it is redundant - I don't rely on its value at the moment. Please let me know if we should remove it then.

kkafar: yeah let's clean it up. The beset way would be here to add extension method on fragment (see examples in ViewExt) `fun Fragment.asStackFragment` & just use it here. 

The method should not have safe-guards. I prefer to throw there when an invariant is violated.

kligarski: Continuation in https://github.com/software-mansion/react-native-screens/pull/2911#discussion_r2097670875

ilteoood:
Any chance that we get this PR released soon?

t0maboro:
The work is continued here: https://github.com/software-mansion/react-native-screens/pull/3336

PR #758, opened 2020-12-21
feat: add IOS searchBar

WoLewicki:
## Description

Added native IOS `SearchBar`.

https://user-images.githubusercontent.com/39658211/117150430-709f2e00-adb8-11eb-8993-9bcc2b6900de.mp4

## Test code and steps to reproduce

See `Test758.tsx` in `TestsExample/` project
or
`SearchBar` playground in `Example/` app

## Checklist

- [x] Included code example that can be used to test this change
- [x] Updated TS types
- [x] Updated documentation: 
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/native-stack/README.md
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/createNativeStackNavigator/README.md
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/src/types.tsx
  - [x] https://github.com/software-mansion/react-native-screens/blob/master/src/native-stack/types.tsx
- [x] Ensured that CI passes

PR #2926, opened 2025-05-16
fix(iOS): allow to hide the search bar

kligarski:
## Description

Search bar on iOS would not disappear after setting `headerSearchBarOptions` to `undefined`. It seems like this issue was present [for some time now](https://github.com/software-mansion/react-native-screens/pull/758#issuecomment-852534970) but somehow it wasn't reported in a seperate issue.

## Changes

- set `navItem.searchController` to `nil` if there is no `RNSScreenStackHeaderSubviewTypeSearchBar` anymore
- add `Test2926` test screen and e2e test

Open `Test2926` in Example app, try to hide the search bar.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -615,9 +615,11 @@ RNS_IGNORE_SUPER_CALL_END
   navitem.rightBarButtonItem = nil;
   navitem.titleView = nil;
 
+#if !TARGET_OS_TV
   // We want to set navitem.searchController to nil only if we are sure
   // that we are removing the search bar from the header.
   bool searchBarPresent = false;
+#endif /* !TARGET_OS_TV */
 
   for (RNSScreenStackHeaderSubview *subview in config.reactSubviews) {
     // This code should be kept in sync on Fabric with analogous switch statement in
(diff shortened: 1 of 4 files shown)


ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -667,6 +674,13 @@ + (void)updateViewController:(UIViewController *)vc
       }
     }
   }
+#if !TARGET_OS_TV
+  if (@available(iOS 11.0, *)) {
+    if (!searchBarPresent) {
+      navitem.searchController = nil;
+    }
+  }
```
kkafar, line 682:
```suggestion
  if (!searchBarPresent) {
    navitem.searchController = nil;
  }
```

This check is not needed anymore -> 15.1 is the last supported iOS version.

kkafar (changes requested):
One small remark. Looks great otherwise, thanks!

ios/RNSScreenStackHeaderConfig.mm
kligarski: b559d17048cec4bfd8db209b500318a706f64900

kkafar (approved):
Looks good. Let's wait for CI & proceed.

Issue #713, opened 2020-11-16
iOS transitionStart - transitionEnd not firing when calling navigation.goBack().


Cal-um:
## Description

When calling navigation.goBack()/ pop() the transitionStart/End events are not fired when in stackPresentation "modal" in iOS.

## Steps To Reproduce

1. Subscribe to the transitionStart and transitionEnd listeners on a modal screen on iOS that has a header close button with an navigation.goBack() onPress function.
2. Open the screen and notice the call-backs firing.
3. tap the close button and notice no transitionStart and transitionEnd events firing.
4. drag to close does have the events.

### Expected behaviour

on calling navigation.goBack() the transitionStart and transitionEnd call backs are triggered.

### Actual behaviour

on calling navigation.goBack() the transitionStart and transitionEnd call backs are not triggered.

## Snack or minimal code example

https://github.com/Cal-um/iOS13-screens-bug

## Package versions

latest versions.

Cal-um:
Events added in  #https://github.com/software-mansion/react-native-screens/pull/499 @janicduplessis

Cal-um:
Having an issue similar to #https://github.com/software-mansion/react-native-screens/issues/513 where the alert is firing before a transition is finished causing the app to lock up

Cal-um:
Seems that updating navigation state via the navigation prop causes these events to not fire. So tapping the back button on a stack triggers a ViewController dismiss which in turn updates navigation state so we get these events. I wonder if we can expose something on the native side to trigger a dismiss so we can get these events when dismissing with react navigation?

WoLewicki:
Isn't it the same for screens with `stackPresentation: 'push'`? The callbacks are not fired for the screen we are leaving when it is done by doing it from the JS side. Am I right?

Cal-um:
Yes that is correct

Cal-um:
I dont think this is going to be a big issue for me because in the next react-native release UIAlerts are going to be presented from the window but it would be nice if we could find a fix for this. I think it would involve building a new StackRouter to handle it from removing the screen from state until the transition is finished.

WoLewicki:
Can you elaborate on what is the use-case for these events that cannot be handled by e.g. `useEffect` cleanups?

Cal-um:
I believe it is helpful to have a way of knowing when a screen transition has finished. useEffect cleanups do not happen when transitions are complete.

Cal-um:
[...]

PR #2928, opened 2025-05-19
test(e2e, Fabric): add e2e tests for issue/PR examples 702..726

kligarski:
## Description

Check which example screens from issues/PRs can be used in e2e testing for tests `Test702, ..., Test726` and implement them if possible for Fabric. 

### Test702
Skipped because we can't check animation in any meaningful way.

### Test706
Skipped because we can't check font weight.

### Test713
Skipped because the issue is still present (`transitionStart/End` events are not fired when leaving the screen due to a call from the JS side; the decision was that [this will not be fixed](https://github.com/software-mansion/react-native-screens/issues/713#issuecomment-732209901) if there is no important use case for it).

### Test726
Test created, only on iOS (original issue was related to iOS). It checks if the app freezes after going back from screen with autofocus input. It is important to note that the issue was reopened with similar, but not the same, bug.

## Changes

- add `Test726` e2e test
- add testIDs to `Test726` test screen

CI

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/apps/src/tests/index.ts
+++ b/apps/src/tests/index.ts
@@ -23,8 +23,8 @@ export { default as Test654 } from './Test654';     // [E2E created](iOS): issue
 export { default as Test658 } from './Test658';     // [E2E created]
 export { default as Test662 } from './Test662';     // [E2E skipped]: can't check animation in a meaningful way
 export { default as Test691 } from './Test691';     // [E2E created](iOS): issue related to iOS modal behavior
-export { default as Test702 } from './Test702';
-export { default as Test706 } from './Test706';
+export { default as Test702 } from './Test702';     // [E2E skipped]: can't check animation in a meaningful way
+export { default as Test706 } from './Test706';     // [E2E skipped]: can't check font weight
 export { default as Test713 } from './Test713';
 export { default as Test726 } from './Test726';
 export { default as Test748 } from './Test748';
--- a/apps/src/tests/index.ts
+++ b/apps/src/tests/index.ts
@@ -26,7 +26,7 @@ export { default as Test691 } from './Test691';     // [E2E created](iOS): issue
 export { default as Test702 } from './Test702';     // [E2E skipped]: can't check animation in a meaningful way
 export { default as Test706 } from './Test706';     // [E2E skipped]: can't check font weight
 export { default as Test713 } from './Test713';     // [E2E skipped]: issue still present
-export { default as Test726 } from './Test726';
+export { default as Test726 } from './Test726';     // [E2E created](iOS): issue related to iOS
 export { default as Test748 } from './Test748';
 export { default as Test750 } from './Test750';
 export { default as Test758 } from './Test758';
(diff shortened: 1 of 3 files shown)


apps/src/tests/index.ts
```diff
@@ -23,10 +23,10 @@ export { default as Test654 } from './Test654';     // [E2E created](iOS): issue
 export { default as Test658 } from './Test658';     // [E2E created]
 export { default as Test662 } from './Test662';     // [E2E skipped]: can't check animation in a meaningful way
 export { default as Test691 } from './Test691';     // [E2E created](iOS): issue related to iOS modal behavior
-export { default as Test702 } from './Test702';
-export { default as Test706 } from './Test706';
-export { default as Test713 } from './Test713';
-export { default as Test726 } from './Test726';
+export { default as Test702 } from './Test702';     // [E2E skipped]: can't check animation in a meaningful way
+export { default as Test706 } from './Test706';     // [E2E skipped]: can't check font weight
+export { default as Test713 } from './Test713';     // [E2E skipped]: issue still present
```
kkafar, line 28: Please add it to internal board then!

kkafar (approved):
Looks good 💪🏻

apps/src/tests/index.ts
kligarski: In the original issue, the decision was that [it will not be fixed](https://github.com/software-mansion/react-native-screens/issues/713#issuecomment-732209901) unless there is some important use case because it would require a lot of changes to the logic. It was 5 years ago though - if you think we should have another look at it, I'll add it to the board, just let me know.

kkafar: I had not looked into the issue before, now I do understand. What is the `Test713` for if the issue was not resolved?

kkafar: If it does not have any purpose (please take a look at `blame`), we can safely remove it.

kligarski: The tests seems unchanged ever since adding it to the repo, but it contains also some other problem with a view "wobbling" at the bottom of the modal. It was not mentioned in the issue discussion. 

https://github.com/user-attachments/assets/7f5b42b6-300a-4561-b580-9a0e6abb45da

It seems like the original author created the reproduction for the wobble first (but I am unable to find any issue related to it in screens/react-navigation repos), then added an issue with the events later and reported it as an issue to our repo (here is the original reproduction repo: https://github.com/Cal-um/iOS13-screens-bug/commits/master/).

kkafar: I know where the "wobbling" / "flickering" comes from, however we won't be able to fix it. I can explain it to you when we have a call. 

So it seems that this can be removed, right?

kkafar (approved):
I think we can proceed. The challenged test can be removed separately

apps/src/tests/index.ts
kligarski: You're right, it can be removed. I'll do it in seperate PR, as you suggested.

PR #2932, opened 2025-05-21
test(e2e, Fabric): add e2e tests for issue/PR examples 748..758, remove Test713

kligarski:
## Description

Check which example screens from issues/PRs can be used in e2e testing for tests `Test748, ..., Test758` and implement them if possible for Fabric. 

### Test713
Test screen removed because the issue is not planned to be fixed, full discussion [here](https://github.com/software-mansion/react-native-screens/pull/2928#discussion_r2095277442).

### Test748
Skipped because the issue description does not contain enough information about the original bug. It is most likely connected with `headerRight` appearing/disappearing in the right moments when navigating between screens and `Test432` already covers even more complex cases.

### Test750
Skipped because we can't check whether insets are respected.

### Test758
Test created, only on iOS (original PR introduced `SearchBar` to iOS). It checks interactions with the search bar.

## Changes

- add `Test758` e2e test
- modify`Test758` test screen (change placeholder and cancel button text, fix TS issue)

CI

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/FabricExample/e2e/issue-tests/Test758.e2e.ts
+++ b/FabricExample/e2e/issue-tests/Test758.e2e.ts
@@ -1,5 +1,18 @@
 import { device, expect, element, by } from 'detox';
 import { describeIfiOS, selectIssueTestScreen } from '../e2e-utils';
+import isVersionEqualOrHigherThan from '../helpers/isVersionEqualOrHigherThan';
+
+const { getIOSVersionNumber } = require('../../../scripts/e2e/ios-devices.js');
+
+// On iOS 26+ cancel button does not contain any text.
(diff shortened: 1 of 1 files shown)


kkafar:
Is this something we should proceed with @kligarski?

kligarski:
I've been waiting for https://github.com/software-mansion/react-native-screens/pull/3430 but after it was merged, I didn't have time to revisit this.

I guess that if the test already exist then we can merge it, I just need to check whether referencing the search bar by class works on iOS 26. I'll try to look at it soon.

kkafar:
This is a low priority. I just revisited some of the older PRs on the repo.

PR #2934, opened 2025-05-21
fix: remove javapart ref from cpp nativeproxy

WoLewicki:
## Description

PR analogous to https://github.com/software-mansion/react-native-reanimated/pull/7515 regarding the `javaPart_`. 
Right now when you do a reload in a simple app with screens installed, you will see in AS profiler that instances of all NativeProxys are still in the memory. It is caused by `javaPart_` never being released thus keeping java part after reload.

Make simple app with screens and do a couple of reloads and check in profiler if the NativeProxy leaks.

### DIFF
--- a/android/src/fabric/java/com/swmansion/rnscreens/NativeProxy.kt
+++ b/android/src/fabric/java/com/swmansion/rnscreens/NativeProxy.kt
@@ -20,6 +20,8 @@ class NativeProxy {
 
     external fun nativeAddMutationsListener(fabricUIManager: FabricUIManager)
 
+    external fun invalidateCpp()
+
     companion object {
         // we use ConcurrentHashMap here since it will be read on the JS thread,
         // and written to on the UI thread.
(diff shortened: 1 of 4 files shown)


tjzel (approved):
LGTM, but make sure that RNScreens doesn't actually try to use anything from `javaPart_` once it was nulled.

In the case of Reanimated and Worklets some (concurrent) code always expected that it's not nulled.

WoLewicki:
But isn't it prohibited behavior for the code to run anything after `invalidate` was called on Java side?

tjzel:
It is prohibited - but the question is, if it's actually respected. For instance, in Worklets and Reanimated we have to make sure that other threads do gracefully abort when `invalidate` is called on the JS thread.

kkafar (commented):
Makes sense to me.

Regarding what @tjzel has said: we **should** be safe on that front, because:

1. the lifecycle of a native proxy is tied & bound by lifecycle of `RNScreensModule`, which in turn is managed by React,
2. If React calls our `RNSScreenRemovalListener` after the screen module has been invalidated it would mean that something went seriously wrong - am I right here @WoLewicki? Or is it possible for native modules to be invalidated && before the view hierarchy is torn down?

Issue #2543, opened 2024-11-27
Jumpy behaviour when setting sheetInitialDetentIndex to other then 0. 


blytung:
### Description

When im using sheetInitialDetentIndex and set it to another value then 1, the formSheet is behave jumpy, it hard to stick to the anchor. 

 
Here how it is
https://github.com/user-attachments/assets/c79b61cd-22e3-406d-8139-e485fa01f18e

`
<RootStack.Screen
            name="Favorites"
            component={Favorites}
            options={{
              presentation: 'formSheet',
              headerShown: true,
              title: '',
              sheetAllowedDetents: [0.2, 0.5, 0.99],
              gestureEnabled: true,
              sheetInitialDetentIndex: 2,
              sheetLargestUndimmedDetentIndex: 'last',
              sheetCornerRadius: 24,

              contentStyle: {
                backgroundColor: COLORS.primaryBackground,
              },
              unstable_screenStyle: {
                backgroundColor: COLORS.primaryBackground,
              },
            }}
          />
`

Expected behaviour: 

https://github.com/user-attachments/assets/51474625-bdb8-4a24-a2a2-f7055f6dd29a

`
<RootStack.Screen
            name="Favorites"
            component={Favorites}
            options={{
              presentation: 'formSheet',
              headerShown: true,
              title: '',
              sheetAllowedDetents: [0.2, 0.5, 0.99],
              gestureEnabled: true,
              sheetInitialDetentIndex: 0,
              sheetLargestUndimmedDetentIndex: 'last',
              sheetCornerRadius: 24,

              contentStyle: {
                backgroundColor: COLORS.primaryBackground,
              },
              unstable_screenStyle: {
                backgroundColor: COLORS.primaryBackground,
              },
            }}
          />
`

### Steps to reproduce

1. set the sheetInitialDetentIndex to a value other then 0. 
2. using formSheet 

### Snack or a link to a repository

[snack](https://snack.expo.dev/@blytung123/react-navigation---nested-go_back-action)

### Screens version

4.0.0

### React Native version

0.75.0

### Platforms

iOS

### JavaScript runtime

None

### Workflow

Expo bare workflow

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

Real device

### Device model

iPhone 14 Pro 

### Acknowledgements

Yes

blytung:
Do you need any more info about this or do you understand the issue? :)

kkafar:
[...]

PR #2935, opened 2025-05-22
fix(iOS): fix detents not changing if sheetInitialDetent > 0

kligarski:
## Description

Changing detents on iOS form sheets would not work correctly if `sheetInitialDetent` was set to value other than `0`. For some reason, initial detent was set on **every** update instead of being set only once.

Fixes https://github.com/software-mansion/react-native-screens/issues/2543.

## Changes

- add `_hasInitialDetentSet` flag and use it to determine whether to set initial detent
- add `Test2543` test screen and e2e test

## Screenshots / GIFs

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/626b4dc8-52ac-454d-a223-d311c0a7a08d" /> | <video src="https://github.com/user-attachments/assets/f7ca38ca-c0b5-4455-b04e-384621bfecbd" /> |

First example works on both videos, second one only after fix.

Run example app, open `Test2543`.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -67,7 +67,7 @@ struct ContentWrapperBox {
   /// Up-to-date only when sheet is in `fitToContents` mode.
   CGFloat _sheetContentHeight;
   ContentWrapperBox _contentWrapperBox;
-  bool _hasInitialDetentSet;
+  bool _sheetHasInitialDetentSet;
 #ifdef RCT_NEW_ARCH_ENABLED
   RCTSurfaceTouchHandler *_touchHandler;
   react::RNSScreenShadowNode::ConcreteState::Shared _state;
@@ -1049,7 +1049,7 @@ RNS_IGNORE_SUPER_CALL_END
     }
 
     // Handle initial detent on the first update.
-    if (!_hasInitialDetentSet) {
+    if (!_sheetHasInitialDetentSet) {
       if (_sheetInitialDetent > 0 && _sheetInitialDetent < _sheetAllowedDetents.count) {
 #if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && defined(__IPHONE_16_0) && \
     __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_16_0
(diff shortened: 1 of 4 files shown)


ios/RNSScreen.mm
```diff
@@ -1047,24 +1048,28 @@ - (void)updateFormSheetPresentationStyle
...
+            RCTLogError(
+                @"[RNScreens] sheetInitialDetent out of bounds, on iOS versions below 16 sheetAllowedDetents is ignored in favor of an array of two system-defined detents");
+          }
         }
+      } else if (_sheetInitialDetent != 0) {
+        RCTLogError(@"[RNScreens] sheetInitialDetent out of bounds for sheetAllowedDetents array");
       }
-    } else if (_sheetInitialDetent != 0) {
-      RCTLogError(@"[RNScreens] sheetInitialDetent out of bounds for sheetAllowedDetents array");
+      _hasInitialDetentSet = true;
```
kligarski, line 1072: I think setting it even though an error could've occured (and the value was not set in any way) is a good solution in order to avoid spamming with errors. Also, this fixed some crashes on iOS 15.5 I encountered when changing the detent very quickly.

kligarski: I'm only wondering if in that case we should change the name to something along the lines of `_initialDetentHandled`.

kligarski: Good point.

https://github.com/software-mansion/react-native-screens/pull/2935/commits/f2c173f1271758db12f14d44563fd8d5dc4f027c

kkafar (approved):
Looks good, thanks!

lodev09:
@kligarski
> For some reason, initial detent was set on every update instead of being set only once.

I think the main culprit here is this block:

https://github.com/software-mansion/react-native-screens/blob/c188d35077848fddf89dc645f9c055050407afac/ios/RNSScreen.mm#L1334-L1340

This gets called everytime you drag, causing performance issues. hence the jumpy glitch. Moving it to `updateProps` should fix the issue.

I've added this fix to my PR https://github.com/software-mansion/react-native-screens/pull/2902 since it caused an issue with the translation value I wanted to add. Also, we wanted to actually use `updateFormSheetPresentationStyle` when `sheetInitialDetentIndex` is changed in order for us to change detent index programatically.

PR #2937, opened 2025-05-23
feat: support `gestureEnabled` for sheets on android

intergalacticspacehighway:
## Description
Adds support for `gestureEnabled` prop on Android for form sheet presentation style. Previously this prop was iOS-only, but now it can control whether bottom sheets can be dismissed by dragging on Android. iOS is doing something similar by setting [modalInPresentation](https://github.com/software-mansion/react-native-screens/blob/c188d35077848fddf89dc645f9c055050407afac/ios/RNSScreen.mm#L345-L346) and overriding [presentationControllerShouldDismiss](https://github.com/software-mansion/react-native-screens/blob/c188d35077848fddf89dc645f9c055050407afac/ios/RNSScreen.mm#L756).

## Changes
- Added `gestureEnabled` support for Android form sheets - controls `isDraggable` and `sheetClosesOnTouchOutside` properties
- Updated `SheetDelegate` to respect `isGestureEnabled` when configuring bottom sheet behavior
- Removed Android platform check in `NativeStackView` to allow the `gestureEnabled` prop on Android
- Updated documentation

Pass `gestureEnabled` false in formSheet `presentation` style and test `FabricExample` on android.
```jsx
<Stack.Screen
  name="FormSheet"
  component={FormSheetScreen}
  options={{
    presentation: 'formSheet',
    gestureEnabled: false, <- this now works on Android too
  }}
/>
```

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [x] Ensured that CI passes


intergalacticspacehighway:
Note: we need to be careful in case we end up merging this PR. This [line](https://github.com/react-navigation/react-navigation/blob/821e08473ca191eec137de26fe03674a22f0bda2/packages/native-stack/src/views/NativeStackView.native.tsx#L296) in React navigation passes explicit `false` for android platform, so it will disable gestures for bottom sheets with the changes in this PR. I am not sure if it is needed in react navigation since react native screens is already [passing](https://github.com/software-mansion/react-native-screens/blob/d017e7ba966ceeb76a49f4b0660b9bb9c2403e51/src/native-stack/views/NativeStackView.tsx#L341) it as `false` for android. Also I don't think screens is handling the hardware back button pop, it is only using back button handler to [update](https://github.com/software-mansion/react-native-screens/blob/d017e7ba966ceeb76a49f4b0660b9bb9c2403e51/android/src/main/java/com/swmansion/rnscreens/CustomSearchView.kt#L24) some config in search view. cc - @satya164 🙏

PR #2938, opened 2025-05-23
refactor(Android): replace inset listeners on decor with multiple insets listeners on screen

maciekstosio:
## Description

Currently we have one singleton proxy listener that is attached to decor and we use that to listen for keyboard events. That sometimes may leads to problems, that some underlying view will consume insets and we won't get notified about this. This PR fixes is by allowing to add multiple events listeners on screen and using it for foresheets. 

- Test2774.tsx
- Test2895.tsx
- TestFormSheet.tsx

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/InsetsObserver.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/InsetsObserver.kt
@@ -1,0 +1,66 @@
+package com.swmansion.rnscreens
+
+import android.view.View
+import android.view.WindowInsets
+import androidx.core.view.OnApplyWindowInsetsListener
+import androidx.core.view.ViewCompat
+import androidx.core.view.WindowInsetsCompat
+import com.facebook.react.bridge.LifecycleEventListener
+import com.facebook.react.bridge.ReactApplicationContext
+
+class InsetsObserver(
+    view: View,
+) {
+    private val observedView = view
+    private var systemListener: View.OnApplyWindowInsetsListener? = null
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/swmansion/rnscreens/InsetsObserver.kt
```diff
@@ -0,0 +1,66 @@
...
+        return rollingInsets
+    }
+
+    fun setOnApplyWindowListener(listener: View.OnApplyWindowInsetsListener?) {
+        systemListener = listener
+    }
+
+    fun addOnApplyWindowInsetsListener(listener: OnApplyWindowInsetsListener) {
+        ownListeners.add(listener)
+        println("InsetsObserver addOnApplyWindowInsetsListener count=${ownListeners.count()}")
```
maciekstosio, line 54:
```suggestion
```

kkafar:
Found a bug on API 29, that does not happen on API 35. Dunno on what version it starts. I also don't know whether this is a regression or it also occurs in main. Will check this. 

Video recorded on `TestFormSheet` screen.


https://github.com/user-attachments/assets/ab37298c-d6b1-41cd-911a-af1bd0b3e49b

kkafar:
Found another bug on API 29, that does not happen on API 35

https://github.com/user-attachments/assets/e5c93a54-b1c9-4423-a65f-463fd6dbd525


Video recorded on `TestFormSheet`

kkafar (changes requested):
Found two pretty big regressions ☝🏻 

I'll try to look into them now.

PR #2952, opened 2025-05-29
refactor(iOS, Fabric): move updateFormSheetPresentationStyle to updateProps

kligarski:
## Description

Previously, `updateFormSheetPresentationStyle` was called from `finalizeUpdates`, which would run many times when the sheet was being dragged. This impacted performance and was the cause of the bug which was fixed in https://github.com/software-mansion/react-native-screens/pull/2935 by introducing a flag for setting initial detent. Calling `updateFormSheetPresentationStyle` in `updateProps` instead of `finalizeUpdates` fixes the bug and eliminates redundant calls to `updateFormSheetPresentationStyle` while allowing to dynamically change initial detent (which was not possible with the flag solution).

Thanks to @lodev09 for [suggesting the change](https://github.com/software-mansion/react-native-screens/pull/2935#issuecomment-2905783835).

## Changes

- remove `_sheetHasInitialDetentSet`
- move calling `updateFormSheetPresentationStyle` from `finalizeUpdates` to `updateProps`

Run formSheet-related test screens in the example app, such as `Test2543`, `TestFormSheet` (you can change `sheetInitialDetent` in the file and save to see the update), `Test2877`.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -67,7 +67,6 @@ struct ContentWrapperBox {
   /// Up-to-date only when sheet is in `fitToContents` mode.
   CGFloat _sheetContentHeight;
   ContentWrapperBox _contentWrapperBox;
-  bool _sheetHasInitialDetentSet;
 #ifdef RCT_NEW_ARCH_ENABLED
   RCTSurfaceTouchHandler *_touchHandler;
   react::RNSScreenShadowNode::ConcreteState::Shared _state;
@@ -966,7 +965,7 @@ RNS_IGNORE_SUPER_CALL_END
  * Updates settings for sheet presentation controller.
  * Note that this method should not be called inside `stackPresentation` setter, because on Paper we don't have
  * guarantee that values of all related props had been updated earlier. It should be invoked from `didSetProps`.
- * On Fabric we have control over prop-setting process but it might be reasonable to run it from `finalizeUpdates`.
+ * On Fabric we run it from `updateProps`.
  */
 - (void)updateFormSheetPresentationStyle
 {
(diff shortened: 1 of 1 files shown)


ios/RNSScreen.mm
```diff
@@ -1305,6 +1300,10 @@ - (void)updateProps:(react::Props::Shared const &)props oldProps:(react::Props::
     [self setReplaceAnimation:[RNSConvert RNSScreenReplaceAnimationFromCppEquivalent:newScreenProps.replaceAnimation]];
   }
 
+#if !TARGET_OS_TV && !TARGET_OS_VISION
+  [self updateFormSheetPresentationStyle];
+#endif // !TARGET_OS_TV && !TARGET_OS_VISION
```
kkafar, line 1305: I would not clutter up this method.

`finalizeUpdates` if a right place to run such side effects. Just limit calling it to when the props have changed. You can achieve this by checking the udpate mask (you have access to it as a parameter of `finalizeUpdates:` method.

kkafar (changes requested):

ios/RNSScreen.mm
lodev09: @kkafar I'm wondering why dragging the sheet triggers that method? Dunno how you could limit that. It is the size changes that's causing it to trigger?

ios/RNSScreen.mm
```diff
@@ -1048,28 +1047,24 @@ - (void)updateFormSheetPresentationStyle
...
+        if (_sheetInitialDetent < 2) {
+          [self setSelectedDetentForSheet:sheet to:UISheetPresentationControllerDetentIdentifierLarge animate:YES];
+        } else {
+          RCTLogError(
+              @"[RNScreens] sheetInitialDetent out of bounds, on iOS versions below 16 sheetAllowedDetents is ignored in favor of an array of two system-defined detents");
         }
-      } else if (_sheetInitialDetent != 0) {
-        RCTLogError(@"[RNScreens] sheetInitialDetent out of bounds for sheetAllowedDetents array");
       }
-      _sheetHasInitialDetentSet = true;
```
kkafar: We should not get rid of that flag. Filtering out the calls to `updateFormSheetPresentationStyle` only to react transactions that modify the props is not enough & the bug we fixed with this flag could be reintroduced. 

Let's keep the behaviour here.

kkafar (changes requested):
I don't want us to remote the `_sheetHasInitialDetentSet` flag.

ios/RNSScreen.mm
kkafar: I'll be frank, I do not know why the render is triggered, I haven't debugged that. Will try to do so on final review & let you know.

ios/RNSScreen.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/2952/commits/0a3b922ac304c71f8d527845a5e0dbdc3214c8ac

lodev09: @kkafar Currently, we're using `sheetInitialDetentIndex` as a controlled prop to resize the sheet. Do you have any suggestion on how we could resize the sheet programmatically? This flag will prevent us from doing that.

PR #2953, opened 2025-05-29
docs: add missing information about statusBarStyle

kligarski:
## Description

Add information about supported values for `statusBarStyle` on Android.

## Changes

- add information about `auto` and `inverted` not being supported on Android
- add information about `light` being default value for `statusBarStyle` on Android
- fix missing parenthesis

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 

### DIFF
--- a/src/native-stack/types.tsx
+++ b/src/native-stack/types.tsx
@@ -499,7 +499,9 @@ export type NativeStackNavigationOptions = {
    */
   statusBarHidden?: boolean;
   /**
-   * Sets the status bar color (similar to the `StatusBar` component). Requires enabling (or deleting) `View controller-based status bar appearance` in your Info.plist file on iOS. Defaults to `auto`.
+   * Sets the status bar color (similar to the `StatusBar` component). Requires enabling (or deleting) `View controller-based status bar appearance` in your Info.plist file on iOS.
+   * `auto` and `inverted` are supported only on iOS. On Android, they will fallback to `light`.
+   * Defaults to `auto` on iOS and `light` on Android.
    */
   statusBarStyle?: ScreenProps['statusBarStyle'];
   /**
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -449,7 +449,9 @@ export interface ScreenProps extends ViewProps {
    */
   statusBarHidden?: boolean;
   /**
-   * Sets the status bar color (similar to the `StatusBar` component). Requires enabling (or deleting) `View controller-based status bar appearance` in your Info.plist file on iOS. Defaults to `auto`.
+   * Sets the status bar color (similar to the `StatusBar` component). Requires enabling (or deleting) `View controller-based status bar appearance` in your Info.plist file on iOS.
+   * `auto` and `inverted` are supported only on iOS. On Android, they will fallback to `light`.
+   * Defaults to `auto` on iOS and `light` on Android.
    */
   statusBarStyle?: 'inverted' | 'auto' | 'light' | 'dark';
   /**
--- a/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
+++ b/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
@@ -289,7 +289,7 @@ Defaults to `false`.
 
 ### `statusBarStyle`
 
-Sets the status bar color (similar to the `StatusBar` component). Requires enabling (or deleting) `View controller-based status bar appearance` in your Info.plist file. On iOS, the possible values are: `auto` (based on [user interface style](https://developer.apple.com/documentation/uikit/uiuserinterfacestyle?language=objc), `inverted` (colors opposite to `auto`), `light`, `dark`. On Android, the status bar will be dark if set to `dark` and `light` otherwise.
+Sets the status bar color (similar to the `StatusBar` component). Requires enabling (or deleting) `View controller-based status bar appearance` in your Info.plist file. On iOS, the possible values are: `auto` (based on [user interface style](https://developer.apple.com/documentation/uikit/uiuserinterfacestyle?language=objc)), `inverted` (colors opposite to `auto`), `light`, `dark`. On Android, the status bar will be dark if set to `dark` and `light` otherwise.
 
 Defaults to `auto`.
 
--- a/native-stack/README.md
+++ b/native-stack/README.md
@@ -503,7 +503,7 @@ Defaults to `false`.
 
 #### `statusBarStyle`
 
-Sets the status bar color (similar to the `StatusBar` component). On iOS, the possible values are: `auto` (based on [user interface style](https://developer.apple.com/documentation/uikit/uiuserinterfacestyle?language=objc), `inverted` (colors opposite to `auto`), `light`, `dark`. On Android, the status bar will be dark if set to `dark` and `light` otherwise.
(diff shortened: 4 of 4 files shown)


kkafar (commented):
Is that change coming from edge-to-edge mode being enabled or did we just miss this information in our documentation. 

If it is edge-to-edge related change we should state it explicitly, because not all apps use edge-to-edge.

Otherwise this seems fine. I haven't tested the runtime behaviour, assuming you decided to open this PR for a reason.

src/native-stack/types.tsx
```diff
@@ -485,7 +485,11 @@ export type NativeStackNavigationOptions = {
    */
   stackPresentation?: ScreenProps['stackPresentation'];
   /**
-   * Sets the status bar animation (similar to the `StatusBar` component). Requires enabling (or deleting) `View controller-based status bar appearance` in your Info.plist file on iOS.
+   * Sets the status bar animation (similar to the `StatusBar` component).
+   * On Android, setting either `fade` or `slide` will set the transition of status bar color. On iOS, this option applies to appereance animation of the status bar.
+   * Requires enabling (or deleting) `View controller-based status bar appearance` in your Info.plist file on iOS.
```
kkafar: I do not recall the exact API behaviour here, therefore I need your explicit confirmation, that on Android the transition when changing status bar colour from light to dark is not impacted by this option.

src/native-stack/types.tsx
```diff
@@ -485,7 +485,11 @@ export type NativeStackNavigationOptions = {
...
+   *
+   * Defaults to `fade` on iOS and `none` on Android.
```
kkafar: Is that so? Doesn't it just default to platform default? 

Moreover these are docs of native stack v5, we do not need to keep them updated in any way.

kkafar: I don't see the code of us enforcing the default value anywhere. I think this information is not accurate.

src/native-stack/types.tsx
```diff
@@ -499,7 +503,9 @@ export type NativeStackNavigationOptions = {
    */
   statusBarHidden?: boolean;
   /**
-   * Sets the status bar color (similar to the `StatusBar` component). Requires enabling (or deleting) `View controller-based status bar appearance` in your Info.plist file on iOS. Defaults to `auto`.
+   * Sets the status bar color (similar to the `StatusBar` component). Requires enabling (or deleting) `View controller-based status bar appearance` in your Info.plist file on iOS.
+   * `auto` and `inverted` are supported only on iOS. On Android, they will fallback to `light`.
+   * Defaults to `auto` on iOS and `light` on Android.
```
kkafar: Similar things applies here. 

I would just dump updating these types at all.

src/types.tsx
```diff
@@ -430,7 +430,11 @@ export interface ScreenProps extends ViewProps {
    */
   stackPresentation?: StackPresentationTypes;
   /**
-   * Sets the status bar animation (similar to the `StatusBar` component). Requires enabling (or deleting) `View controller-based status bar appearance` in your Info.plist file on iOS.
+   * Sets the status bar animation (similar to the `StatusBar` component).
+   * On Android, setting either `fade` or `slide` will set the transition of status bar color. On iOS, this option applies to appereance animation of the status bar.
+   * Requires enabling (or deleting) `View controller-based status bar appearance` in your Info.plist file on iOS.
+   *
+   * Defaults to `fade` on iOS and `none` on Android.
```
kkafar, line 437: The remark on this "default" value statement correctness also applies here. I did quick search I haven't found us defaulting to these values. Could you provide code places where this is set?

src/types.tsx
```diff
@@ -449,7 +453,9 @@ export interface ScreenProps extends ViewProps {
    */
   statusBarHidden?: boolean;
   /**
-   * Sets the status bar color (similar to the `StatusBar` component). Requires enabling (or deleting) `View controller-based status bar appearance` in your Info.plist file on iOS. Defaults to `auto`.
+   * Sets the status bar color (similar to the `StatusBar` component). Requires enabling (or deleting) `View controller-based status bar appearance` in your Info.plist file on iOS.
+   * `auto` and `inverted` are supported only on iOS. On Android, they will fallback to `light`.
+   * Defaults to `auto` on iOS and `light` on Android.
```
kkafar, line 458: And similar thing here 🙏🏻

kkafar (changes requested):
I have doubts, whether the reported default values are accurate & also I feel like there is no reason to update `/native-stack/` * `/src/native-stack`. As the whole thing is deprecated anyway.

src/native-stack/types.tsx
kligarski: It does not, the only usage of `isStatusBarAnimated` is `setColor`.

https://github.com/user-attachments/assets/f5664c4a-b12d-4ed1-998f-94dc6310f85e

src/native-stack/types.tsx
kligarski: On Android, we use `false` as the default.
```kotlin
    @ReactProp(name = "statusBarAnimation")
    override fun setStatusBarAnimation(
        view: Screen,
        statusBarAnimation: String?,
    ) {
        val animated = statusBarAnimation != null && "none" != statusBarAnimation
        view.isStatusBarAnimated = animated
    }
```

```kotlin
    internal fun setColor(
        screen: Screen,
        activity: Activity?,
        context: ReactContext?,
    ) {
        // ...
        val animated = screenForAnimated?.isStatusBarAnimated ?: false
        // ...
}
```

kligarski: On iOS:
```objective-c

- (UIStatusBarAnimation)preferredStatusBarUpdateAnimation
{
  UIViewController *vc = [self findChildVCForConfigAndTrait:RNSWindowTraitAnimation includingModals:NO];

  if ([vc isKindOfClass:[RNSScreen class]]) {
    return ((RNSScreen *)vc).screenView.statusBarAnimation;
  }
  return UIStatusBarAnimationFade;
}
```

src/types.tsx
kligarski: https://github.com/software-mansion/react-native-screens/pull/2953#discussion_r2120434773
https://github.com/software-mansion/react-native-screens/pull/2953#discussion_r2120441393

src/types.tsx
kligarski: On Android:
```kotlin
    internal fun setStyle(
        screen: Screen,
        activity: Activity?,
        context: ReactContext?,
    ) {
        // ...
        val style = screenForStyle?.statusBarStyle ?: "light"
        // ...
    }
```

kligarski: On iOS:
```objective-c
- (UIStatusBarStyle)preferredStatusBarStyle
{
  return [RNSScreenWindowTraits statusBarStyleForRNSStatusBarStyle:self.screenView.statusBarStyle];
}
```
```objective-c
@property (nonatomic) RNSStatusBarStyle statusBarStyle;
```
```objective-c
typedef NS_ENUM(NSInteger, RNSStatusBarStyle) {
  RNSStatusBarStyleAuto,
  RNSStatusBarStyleInverted,
  RNSStatusBarStyleLight,
  RNSStatusBarStyleDark,
};
```

`statusBarStyleForRNSStatusBarStyle` is a simple switch mapping `RNSStatusBarStyle` to `UIStatusBarStyle`.

kkafar (commented):
Okay, thanks for the links. I'm not approving yet, because I wonder whether we need to introduce this as stable behaviour or not :D I'll  approve / reject when I make my mind

PR #2957, opened 2025-06-02
test(e2e, Fabric): add e2e tests for issue/PR examples 761..791

kligarski:
## Description

Check which example screens from issues/PRs can be used in e2e testing for tests `Test761, ..., Test791` and implement them if possible for Fabric. 

### Test761
Skipped because we can't check animation.

### Test779
Skipped because we can't check animation.

### Test780
Skipped because we can't use native swipe back gesture with Detox.

### Test791
Test created, only on iOS (original issue was about crashes on iOS). It tests if the app crashes when you push multiple screens quickly behind modal.

## Changes

- add `Test791` e2e test
- modify`Test791` test screen (add explicit `presentation: 'card'` for push screen, add testIDs)

CI

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/FabricExample/e2e/issuesTests/Test791.e2e.ts
+++ b/FabricExample/e2e/issuesTests/Test791.e2e.ts
@@ -1,0 +1,29 @@
+import { device, expect, element, by } from 'detox';
+import { describeIfiOS, selectTestScreen } from '../e2e-utils';
+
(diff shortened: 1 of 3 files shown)


kkafar (approved):
Looks solid, thanks!

PR #2958, opened 2025-06-02
fix(iOS): lack of animation on modal replacement

kkafar:
## Description

The fix boils down to removing check whether `changeRootIndex ==
controllers.count`. It was introduced here:
https://github.com/software-mansion/react-native-screens/pull/502.
It can be represented differently: `changeRootIndex ==
controllers.lastIndex + 1` & if I get this correctly it checks whether
the modal change in question takes place at the top of view hierarchy,
with purpose of not animating non visible changes.

It does not take account the replace operation though. In case of
replace, the `changeRootIndex == controllers.lastIndex`.

Another aspect is that I believe this part of the check is not
necessary. If some modal view controller is being dismissed, **all**
view controllers above it will also endup dismissed.
Therefore, there is no scenario where this check would be useful.

Hence, this commit ends up removing the check.

## Before & After

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/d3a718e2-aef3-49bc-abdd-fa172c2097a3" /> | <video src="https://github.com/user-attachments/assets/e3b7c05c-fd9e-48b7-96e7-c59e75cd4c3f" /> | 

I've enhanced `TestModalNavigation` with additional button that allows for modal replacement.

I've also tested `TestModalNavigation` to make sure other flows keep working. 
I've also tested #502 - setting `animation: 'none'` keeps working for dismissed modals.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/apps/src/tests/TestModalNavigation.tsx
+++ b/apps/src/tests/TestModalNavigation.tsx
@@ -1,7 +1,10 @@
 import * as React from 'react';
 import { View, Text, Button } from 'react-native';
 import { NavigationContainer } from '@react-navigation/native';
-import { NativeStackScreenProps, createNativeStackNavigator } from '@react-navigation/native-stack';
+import {
+  type NativeStackScreenProps,
+  createNativeStackNavigator,
+} from '@react-navigation/native-stack';
 
 type StackParamList = {
   Home: undefined;
@@ -12,7 +15,6 @@ type StackParamList = {
   Screen2: undefined;
   Screen3: undefined;
 };
-
 
 function HomeScreen({
   navigation,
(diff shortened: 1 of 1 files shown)


kkafar:
@hirbod you might wanna check this one out in context of the issue we've talked about. 
I don't think this is it, but there is a slight change it is.

hirbod:
Hi @kkafar, thanks for the tag. I’ll pull the PR and test it. I do think this issue is unrelated though and probably a different one. That said, I love this PR anyway, it makes replacing a modal screen way easier and removes the need for setTimeout and such.

PR #2964, opened 2025-06-05
fix(Android): crash API 25 when going back from screen with flat list

### LINKED ISSUE #2810: Going back in native stack cause crash on Android 7 if screen contains FlatList with refresh props (React Navigation related)
### Description

Original issue: 
https://github.com/react-navigation/react-navigation/issues/12510

I have such react navigation structure:
Home -> Details

When Details screen contains List components (FlatList, VirtualizedList) with refresh props such as onRefresh, refreshing or refreshControl = <RefreshController...> (React Native component) going back causes app crash and it happens only on Android 7. Other Android versions work as expected.

Can be reproduced on emulators and real devices

### Steps to reproduce

1. Open example app
2. Navigate to Details Screen
3. Go back with ui or with hardware button

https://github.com/Zao721/CrashOnGoingBackAndroid7

### Screens version

^4.9.2

### React Native version

78.1

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Release mode

### Device

Real device

### Device model

Samsung Galaxy J2 Pro (Android 7.1.1), Android Emulator 7.1.1

### Acknowledgements

Yes

maciekstosio:
## Description

Fixes #2810

This PR should fix a crash when going back from the screen that has flat list.

## Changes
 There are two exceptions thrown:

**First:**
```
Error: Exception in HostFunction: android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views.
```

**Second:**
```
java.lang.IndexOutOfBoundsException: getChildDrawingOrder() returned invalid index 1 (child count is 1)
```

And it fails around this place in `Screen.kt`:

```
(...)
                if (parent is SwipeRefreshLayout && child is ImageView) {
                    // SwipeRefreshLayout class which has CircleImageView as a child,
                    // does not handle `startViewTransition` properly.
                    // It has a custom `getChildDrawingOrder` method which returns
                    // wrong index if we called `startViewTransition` on the views on new arch.
                    // We add a simple View to bump the number of children to make it work.
                    // TODO: find a better way to handle this scenario
                        it.addView(View(context), i)
                } else {
                    child?.let { view -> it.startViewTransition(view) }
                }
(...)
```

Based on the comment, I assume `it.addView(View(context), i)` is to prevent the **second** crash. Because the **first** exception, we do not add "workaround" view, thus we run into **second**. We get the **first** exception, as the operation is performed from different thread, thus I ensure the adding view is performed from UIThread.

<details>
<summary>Source code</summary>

```
import { View, Text, FlatList } from 'react-native';
import { NavigationContainer, useNavigation } from '@react-navigation/native';
import { createNativeStackNavigator, NativeStackScreenProps } from '@react-navigation/native-stack';
import { Button } from '@react-navigation/elements';
import { SafeAreaProvider } from 'react-native-safe-area-context';

type StackParamList = {
  Home: undefined;
  Details: undefined;
};

function HomeScreen() {
  const navigation = useNavigation<NativeStackScreenProps<StackParamList, 'Home'>['navigation']>();

  return (
      <Text>Home Screen</Text>
      <Button onPress={() => navigation.navigate('Details')}>
        Go to Details
      </Button>
    </View>
  );
}

function DetailsScreen() {
  const [isRefreshing, setIsRefreshing] = React.useState(false);

  return (
      <FlatList
        data={[1, 2, 3, 4, 5]}
        renderItem={({ item }) => <Text>{item}</Text>}
        onRefresh={() => {
          return new Promise<void>((resolve) => {
            setIsRefreshing(true);
            setTimeout(() => {
              resolve();
            }, 3000);
          }).then(() => {
            setIsRefreshing(false);
          });
        }}
        refreshing={isRefreshing}
      />
    </View>
  );
}

function NestedStack() {
  return (
    <Stack.Navigator>
      <Stack.Screen name="Details" component={DetailsScreen} />
    </Stack.Navigator>
  );
}

  return (
    <SafeAreaProvider>
      <NavigationContainer>
        <NestedStack />
      </NavigationContainer>
    </SafeAreaProvider>
  );
};
```

</details>

## Checklist

- [ ] Ensured that CI passes


kkafar:
This is the approach I've mentioned to you: https://github.com/software-mansion/react-native-screens/pull/2532/files, however I'm not happy with this PR neither. Let's have a talk today & I'll explain.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -493,7 +493,7 @@ class Screen(
                     // wrong index if we called `startViewTransition` on the views on new arch.
                     // We add a simple View to bump the number of children to make it work.
                     // TODO: find a better way to handle this scenario
-                    it.addView(View(context), i)
+                        it.addView(View(context), i)
```
maciekstosio:
```suggestion
                    it.addView(View(context), i)
```

kkafar:
@maciekstosio Do you remember, what was the conclusion for this PR? Were there any issues? I have something in the back of my head, that there was an issue, but can not recall exactly.

mensonones:
This problem would not be solved with?

```javascript
fun startRemovalTransition() {
        if (!isBeingRemoved) {
            isBeingRemoved = true
            post { startTransitionRecursive(this) }
        }
    }
```

PR #2967, opened 2025-06-06
feat: add support for `screenId` prop allowing for native screen recognition

kkafar:
## Description

Add support for `screenId` prop on native side. At the moment we use this value only in JS code.
It was brought to my attention, however, that other libs that wanna integrate might want that
to be exposed also on native side. 

## Changes

Added `screenId` prop & exposed it on both platforms.

Any example & add a breakpoint at setter method.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStack.h
+++ b/ios/RNSScreenStack.h
@@ -36,6 +36,18 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+#pragma mark-- Integration
+
+@interface RNSScreenStackView ()
+
+/**
+ * \return Arrray with ids of screens owned by this stack. Ids are returned in no particular order. The list might be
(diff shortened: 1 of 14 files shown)


ios/RNSScreenStack.h
```diff
@@ -36,6 +36,18 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+#pragma mark-- Integration
+
+@interface RNSScreenStackView ()
+
+/**
+ * \return Arrray with ids of screens owned by this stack. Ids are returned in no particular order. The list might be
```
kligarski, line 44:
```suggestion
 * \return Array with ids of screens owned by this stack. Ids are returned in no particular order. The list might be
```

PR #3002, opened 2025-07-16
feat(iOS, SplitView): Add support for showsSecondaryOnlyButton

t0ms0n00:
## Description

Closes https://github.com/software-mansion/react-native-screens-labs/issues/231 

Adding setter for showsSecondaryOnlyButton which adds native button for toggling display mode to `secondaryOnly`.

### Notes:

- In the pure native app, when I wanted to change this value dynamically, update for button visibility was applied after another interaction, like swiping one of columns.

## Changes

- Added native setter
- Updated native component in JS
- Updated app

Added this prop to `SplitViewBaseApp`

## Checklist

- [x] Ensured that CI passes


kkafar (approved):
Let's create ticket on board to research this issue with "dynamic update" - maybe we can come up with some workaround later on. 

Beside that, the code looks good.

t0maboro:
created a ticket for further investigation https://github.com/software-mansion/react-native-screens-labs/issues/285

PR #3009, opened 2025-07-17
feat(Android): Expose configuration for showAsAction with collapse

### LINKED ISSUE #2744: Android: Up/back on search should cancel search, not navigate
### Description

Originally posted here https://github.com/react-navigation/react-navigation/issues/12459#issuecomment-2685489915

### Actual behaviour

Actual behaviour: On Android, tapping the up button or using the back gesture with search open immediately navigates from the screen.

https://github.com/user-attachments/assets/e4b20ff0-6bb5-44e4-968c-e4cc71fb5a5c

This is especially a problem in landscape mode, as the text input covers the X button and tapping the X button once will open the text input again. So to cancel search, you need to awkwardly close the input, press X, close the input, and then press X again

https://github.com/user-attachments/assets/2be274c6-44b1-429f-be31-a5b770f9b48f

### Expected behavior

Expected behaviour: To match native behaviour and platform conventions. On Android, pressing the up or using the back gesture when search is open should first cancel the search. Tapping it again should then return from the screen.

I have tested this on a real device (Pixel 7 Pro) as well as the android emulator

### Steps to reproduce

1. Open search screen
2. Tap search icon
3. Search
4. Press "<-" or do the back gesture
5. Expected: cancel search. Actual: exits search screen

https://github.com/andrewwardmt/mre-search-back-cancel

### Screens version

4.4.0

### React Native version

0.76.7

### Platforms

Android

### JavaScript runtime

None

### Workflow

Expo bare workflow

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

Android emulator

### Device model

_No response_

### Acknowledgements

Yes

kmichalikk:
## Description

Resolves #2744 

We have hardcoded `MenuItem.showAsAction` to `SHOW_AS_ACTION_ALWAYS` to have search icon always visible on toolbar. This setting also indirectly controls the behavior of Up button; OR-ing it with SHOW_AS_ACTION_COLLAPSE_ACTION_VIEW makes the Up button collapse the searchbar first, then pop the screen on the second click.

## Changes

Added `headerSearchBarActions.showAsAction` option for Screen with possible values: `default`, `always`, `collapse`, where `default === always`

Currently, no other SHOW_AS_ACTION_ enum is supported

https://github.com/user-attachments/assets/7442a8e9-8117-45ba-a6c2-8350fc8432be

See `Test2744`


android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
```diff
@@ -412,32 +413,43 @@ class ScreenStackFragment :
...
+            }
+        }
+
+        return MenuItem.SHOW_AS_ACTION_ALWAYS
     }
 
     private fun updateToolbarMenu(menu: Menu) {
         menu.clear()
-        if (shouldShowSearchBar()) {
+        locateSearchBarSubviewInConfig()?.let {
```
kmichalikk, line 442: `shouldShowSearchBar()` was effectively true when config for `type==SEARCH_BAR` existed, so this refactor should not change the behaviour

apps/src/tests/Test2744.tsx
```diff
@@ -0,0 +1,67 @@
...
+};
+
+type NavigationProp<ParamList extends ParamListBase> = {
+  navigation: NativeStackNavigationProp<ParamList>;
+};
+
+type StackNavigationProp = NavigationProp<StackRouteParamList>;
+
+const SearchAlways = ({navigation}: StackNavigationProp) => {
+  return <ScrollView contentInsetAdjustmentBehavior='automatic'>
```
kkafar, line 19: Let's use braces `()` - this will make code more maintainable, if some additional logic will be needed, also it'll stick to the pattern we use everywhere else. 

Another thing - in general I prefer (and try to enforce) usage of regular functions over arrow functions, due to symbol hoisting. Therefore let's rewrite this a bit.

src/types.tsx
```diff
@@ -733,6 +734,16 @@ export interface SearchBarProps {
    * @platform ios
    */
   tintColor?: ColorValue;
+  /**
+   * @platform android
```
kkafar, line 738: This annotation should be placed in the end of the comment block. No technical reason, just to keep the convention.

android/src/main/res/base/drawable-anydpi/ic_action_search.xml
kkafar, line 1: Why do we need this additional resource?

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
```diff
@@ -412,32 +413,43 @@ class ScreenStackFragment :
...
             if (searchView == null && currentContext != null) {
                 val newSearchView = CustomSearchView(currentContext, this)
                 searchView = newSearchView
                 onSearchViewCreate?.invoke(newSearchView)
             }
-            menu.add("").apply {
-                setShowAsAction(MenuItem.SHOW_AS_ACTION_ALWAYS)
+            menu.add("Search").apply {
                 actionView = searchView
+                setIcon(R.drawable.ic_action_search)
```
kkafar, line 451: Why do we need to set this icon manually?

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
```diff
@@ -412,32 +413,43 @@ class ScreenStackFragment :
...
+        locateSearchBarSubviewInConfig()?.let {
             val currentContext = context
             if (searchView == null && currentContext != null) {
                 val newSearchView = CustomSearchView(currentContext, this)
                 searchView = newSearchView
                 onSearchViewCreate?.invoke(newSearchView)
             }
-            menu.add("").apply {
-                setShowAsAction(MenuItem.SHOW_AS_ACTION_ALWAYS)
+            menu.add("Search").apply {
```
kkafar, line 449: Why do we need to change that?

kkafar: Where is this title used?

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
kkafar: This is ok, thanks for the note.

src/fabric/SearchBarNativeComponent.ts
```diff
@@ -23,6 +23,8 @@ type SearchBarPlacement = 'automatic' | 'inline' | 'stacked';
 
 type AutoCapitalizeType = 'none' | 'words' | 'sentences' | 'characters';
 
+type SearchShowAsAction = 'default' | 'always' | 'collapse';
```
kkafar, line 26: Why do we need `default` case? Couldn't we just use `always` if the user does not set the value? Does it give us any advantage over direct approach?

kkafar:
> [!note]
Note for future us. 
>
>On hold due to low prio + issues with existing code.
>
>There already exists code responsible for that interaction (preventing immediate pop when search bar is focused), however it stopped to work at some time. We need to figure out why it has stopped to work & fix it.

PR #3010, opened 2025-07-17
feat(iOS, SplitView) Add support for displayModeButtonVisibility prop

t0ms0n00:
`displayModeButtonVisibility` - A setting that determines whether the display mode button is visible in the interface.

During testing, I found that layout is broken for display mode with 3 columns open from start, therefore I'm fixing it updating the ShadowState with the frame that I received in `viewDidLayoutSubviews`. Additionally, methods using inspector should be hidden behind iOS 26 runtime check to prevent any unexpected results.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/232

## Changes

- Added `displayModeButtonVisibility` prop
- Fixed layout for multiple column display
- Hidden inspector methods behind iOS 26 check

## Screenshots / GIFs

https://github.com/user-attachments/assets/01d311ed-9585-4e82-b09e-5d0dc365e5d6

Updated app to cover this prop

## Checklist

- [ ] Ensured that CI passes


ios/gamma/split-view/RNSSplitViewScreenController.swift
```diff
@@ -54,27 +54,25 @@ public class RNSSplitViewScreenController: UIViewController {
...
-          ofComponent: splitViewScreenComponentView,
-          inContextOfAncestorView: ancestorView!
-        )
-      } else {
-        shadowStateProxy.updateShadowState(
-          ofComponent: splitViewScreenComponentView
-        )
-      }
+    // For modals, which are presented outside the SplitViewHost subtree (and RN hierarchy),
+    // we're attaching our touch handler adn we don't need to apply any offset corrections,
```
kligarski, line 58:
```suggestion
    // we're attaching our touch handler and we don't need to apply any offset corrections,
```

kligarski (approved):

kkafar (approved):
This looks good. Thanks.

PR #3014, opened 2025-07-17
feat(iOS, Tabs, ScreenStack): orientation for Tabs and ScreenStack

kligarski:
## Description

Adds orientation management for `Tabs`.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/97.

Moved from https://github.com/software-mansion/react-native-screens-labs/pull/262.

Overriding `supportedInterfaceOrientaitons` for `TabBarController` and `TabsScreenViewController` is not enough. UIKit calls these methods only once and does not enforce orientation they return. UIKit asks root view controller more often - that's why we already swizzled `supportedInterfaceOrientations` to get access to return value from root `UIViewController` (from react-native) in `ScreenStack` implementation (useful article about [UIKit's behavior](https://gist.github.com/SergLam/802dd8a354ff9b925bb3ede8cae1f644)).

I decided to use our own methods to handle orientation instead of overriding `supportedInterfaceOrientations` for tabs-related view controllers and use swizzled `supportedInterfaceOrientation` in root to return appropriate orientation mask to UIKit.

In `ScreenStack`, `orientation: 'default'` is mapped to `UIInterfaceOrientationMaskAllButUpsideDown` but this is correct only for iPhones (e.g. iPads use `UIInterfaceOrientationMaskAll`). There is no other option that maps to `UIInterfaceOrientationMaskAllButUpsideDown`.

In new implementation, I added options for all `UIInterfaceOrientationMask` enum values and changed `'default'` into `'inherit'` to use as default value. This value means that view controller does not have any preference for orientation. When root is asked for `supportedInterfaceOrientations`, it checks if their last child conforms to `RNSOrientationProviding`. If so, it asks the child view controller for orientation. These calls are propagated down the hierarchy with the last child having priority. If the last child returns `Inherit`, view controller above returns its preference. If root receives `Inherit`, it uses mask defined in `Info.plist` ([docs](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/application(_:supportedinterfaceorientationsfor:)?language=objc)), otherwise it maps received `RNSOrientation` into `UIInterfaceOrientationMask`.

For `ScreenStack`, we use old implementation of handling orientation at the screen level (we map result from its `supportedInterfaceOrientations` to `RNSOrientation`). Old stack does not support new props, including `RNSOrientationInherit`.

### Example
```
UIViewController (React)
       |
       | last child view controller's orientation
       | or orientation from Info.plist
       |
RNSTabBarController
       |
       | selected view controller's orientation 
       | or RNSOrientationInherit
       |
RNSTabsScreenViewController
       |
       | last child view controller's orientation 
       | or orientation provided via prop (defaults to RNSOrientationInherit)
       |
RNSNavigationController
       |
       | top view controller's orientation 
       | or RNSOrientationInherit
       |
RNSScreen
       |
       | RNSOrientation mapped from UIInterfaceOrientationMask 
       | returned from supportedInterfaceOrientations (previous implementation)
       |
      ...
```

In the future, we need to implement orientation management for new `Stack` implementation as well as `SplitView`. This should be straightforward. `Stack` implementation will be a mix of old `ScreenStack` and new `Tabs` implementations. `SplitView` implementation should take into account orientation defined via prop at host controller.

> [!NOTE]
> Changing orientation manually (editing and saving file) via config in `TestBottomTabs.tsx` seems buggy but it is due to "Refreshing..." bar in debug mode. If you change the prop in JS (e.g. with `setInterval`), it works as expected.
> | Manually | In JS | 
> | --- | --- |
> | <video src="https://github.com/user-attachments/assets/eef43c27-e6d0-475f-860e-6854e9b42c4f" /> | <video src="https://github.com/user-attachments/assets/5356a1ee-bb90-4811-a427-12a3e94109e4" /> |

## Changes

- add `RNSOrientationProviding` protocol
- implement it for tabs and screenstack

Run example app and switch between tabs/screens in nested stack.

## Checklist

- [ ] Ensured that CI passes


ios/RNSScreenStack.h
```diff
@@ -8,10 +8,17 @@
...
+
 NS_ASSUME_NONNULL_BEGIN
 
-@interface RNSNavigationController
-    : UINavigationController <RNSViewControllerDelegate, RNSBottomTabsSpecialEffectsSupporting>
+#if !TARGET_OS_TV
+@interface RNSNavigationController : UINavigationController <RNSViewControllerDelegate, RNSBottomTabsSpecialEffectsSupporting, RNSOrientationProviding>
+#else // !TARGET_OS_TV
+@interface RNSNavigationController : UINavigationController <RNSViewControllerDelegate, RNSBottomTabsSpecialEffectsSupporting>
+#endif // !TARGET_OS_TV
```
t0maboro: we don't need to duplicate the whole interface

ios/bottom-tabs/RNSTabBarController.h
```diff
@@ -21,7 +25,11 @@ NS_ASSUME_NONNULL_BEGIN
  * i.e. if you made changes through one of signals method, unless you flush them immediately (not needed atm), they will
  * be executed only after react finishes the transaction (from within transaction execution block).
  */
+#if !TARGET_OS_TV
+@interface RNSTabBarController : UITabBarController <RNSReactTransactionObserving, RNSOrientationProviding>
+#else // !TARGET_OS_TV
 @interface RNSTabBarController : UITabBarController <RNSReactTransactionObserving>
+#endif // !TARGET_OS_TV
```
t0maboro: same as above

ios/bottom-tabs/RNSTabBarController.mm
```diff
@@ -141,6 +152,7 @@ - (void)updateSelectedViewController
 
   [selectedViewController.tabScreenComponentView overrideScrollViewBehaviorInFirstDescendantChainIfNeeded];
   [self setSelectedViewController:selectedViewController];
+  [self updateOrientation];
```
t0maboro: do we need to call it here explicitly? we have: `reactMountingTransactionDidMount -> updateSelectedViewControllerIfNeeded -> updateSelectedViewController` and the orientation update will come after `updateSelectedViewControllerIfNeeded`

ios/bottom-tabs/RNSTabBarController.mm
```diff
@@ -203,4 +215,32 @@ - (void)scheduleControllerUpdateIfNeeded
 
 #endif // !RCT_NEW_ARCH_ENABLED
 
+- (void)updateOrientationIfNeeded
+{
+  if (_needsOrientationUpdate) {
```
t0maboro, line 222: When should it be set back to false after applying the update?

ios/conversion/RNSConversions-BottomTabs.mm
```diff
@@ -287,4 +287,29 @@ RNSBottomTabsIconType RNSBottomTabsIconTypeFromIcon(react::RNSBottomTabsScreenIc
   return iconImageSource;
 }
 
+RNSOrientation RNSOrientationFromRNSBottomTabsScreenOrientation(
+    react::RNSBottomTabsScreenOrientation orientation)
+{
+  using enum facebook::react::RNSBottomTabsScreenOrientation;
+  
+  switch (orientation) {
```
t0maboro, line 294: nit: you may consider adding handling default case

ios/RNSScreenStack.h
kligarski: https://github.com/software-mansion/react-native-screens/pull/3014/commits/fcc1a7cac3313b0be93a9c0fe1dae5aac40f4006

ios/bottom-tabs/RNSTabBarController.h
kligarski: https://github.com/software-mansion/react-native-screens/pull/3014/commits/fcc1a7cac3313b0be93a9c0fe1dae5aac40f4006

ios/bottom-tabs/RNSTabBarController.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3014/commits/9dedb8476f3aeb4f19c9d5e1c1d365f37dd45116

ios/bottom-tabs/RNSTabBarController.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3014/commits/9dedb8476f3aeb4f19c9d5e1c1d365f37dd45116

ios/conversion/RNSConversions-BottomTabs.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3014/commits/9dedb8476f3aeb4f19c9d5e1c1d365f37dd45116

t0maboro (approved):

PR #3015, opened 2025-07-17
feat(Android, Tabs): pass material theme using context wrapper & update the layout flow

maciekstosio:
To have styles and all features described in https://github.com/material-components/material-components-android/blob/master/docs/components/BottomNavigation.md, we should change the theme of the app to inherit from material design theme (the is one of getting started steps https://github.com/material-components/material-components-android/blob/master/docs/getting-started.md#7-material3-theme-inheritance). We don't want force users to do that in each project, thus we wrap `reactContext` using `ContextWrapperThemed` and pass a custom theme there.

Additionally, after adding the items, layout was broken, so I force layout using the pattern from https://github.com/software-mansion/react-native-screens/pull/2332/files.

## Screenshots / GIFs

image

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
@@ -1,5 +1,6 @@
 package com.swmansion.rnscreens.gamma.tabs
 
+import com.swmansion.rnscreens.R
 import android.content.res.ColorStateList
 import android.util.Log
 import android.view.Choreographer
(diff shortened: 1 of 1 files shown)


android/src/main/res/base/values/themes.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <!-- Base application theme. -->
+    <style name="custom" parent="Theme.Material3.DayNight.NoActionBar">
```
kkafar, line 4:
```suggestion
    <style name="material_base" parent="Theme.Material3.DayNight.NoActionBar">
```

or something more informative.

kkafar:
Also if there is a ticket for this on a board, let's link it.

android/src/main/res/base/values/themes.xml
maciekstosio: I removed that in https://github.com/software-mansion/react-native-screens/pull/3026 anyway

PR #3017, opened 2025-07-17
feat(Android, Tabs): support badge in bottom navigation

maciekstosio:
## Description
Resolves: #256

This PRs adds basic support for badge for android. It allows to set color and label. To reuse the iOS logic (which allows string as badge value) I need to parse it to int and I log error otherwise. This isn't ideal and I created low-priority task to change it at some point https://github.com/software-mansion/react-native-screens-labs/issues/249. 

## Screenshots / GIFs

image

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabScreenViewManager.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabScreenViewManager.kt
@@ -86,7 +86,9 @@ class TabScreenViewManager :
     override fun setTabBarItemBadgeBackgroundColor(
         view: TabScreen,
         value: Int?,
-    ) = Unit
+    ) {
+        view.tabBarItemBadgeBackgroundColor = value
+    }
 
     override fun setTabBarItemTitlePositionAdjustment(
         view: TabScreen?,
@@ -142,9 +144,11 @@ class TabScreenViewManager :
     }
 
     override fun setBadgeValue(
-        view: TabScreen?,
+        view: TabScreen,
         value: String?,
-    ) = Unit
+    ) {
+        view.badgeValue = value
+    }
 
     @ReactProp(name = "title")
     override fun setTitle(
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
@@ -304,7 +304,8 @@ class TabsHost(
         bottomNavigationView.itemIconTintList = ColorStateList(states, iconColors)
 
         // ActivityIndicator color
-        val activityIndicatorColor = tabBarItemActivityIndicatorColor ?: com.google.android.material.R.color.m3_sys_color_dynamic_light_on_secondary_container
+        val activityIndicatorColor =
+            tabBarItemActivityIndicatorColor ?: com.google.android.material.R.color.m3_sys_color_dynamic_light_on_secondary_container
         bottomNavigationView.itemActiveIndicatorColor = ColorStateList.valueOf(activityIndicatorColor)
 
         // First clean the menu, then populate it
@@ -376,7 +377,10 @@ class TabsHost(
         }
     }
 
-    private fun updateBadgeAppearance(menuItemIndex: Int, tabScreen: TabScreen) {
+    private fun updateBadgeAppearance(
+        menuItemIndex: Int,
+        tabScreen: TabScreen,
+    ) {
         val badgeValue = tabScreen.badgeValue?.toIntOrNull()
         if (tabScreen.badgeValue != null && badgeValue == null) {
             Log.e(TAG, "[RNScreens] Android supports only numbers as badge value")
@@ -386,7 +390,9 @@ class TabsHost(
             val badge = bottomNavigationView.getOrCreateBadge(menuItemIndex)
             badge.isVisible = true
             badge.number = badgeValue
-            badge.backgroundColor = tabScreen.tabBarItemBadgeBackgroundColor ?: wrappedContext.getColor(com.google.android.material.R.color.error_color_material_light)
+            badge.backgroundColor =
+                tabScreen.tabBarItemBadgeBackgroundColor
+                    ?: wrappedContext.getColor(com.google.android.material.R.color.error_color_material_light)
         } else {
             val badge = bottomNavigationView.getBadge(menuItemIndex)
             badge?.isVisible = false
@@ -473,8 +479,7 @@ class TabsHost(
         menuItem.icon = tabScreen.icon
 
         // Badge
-        updateBadgeAppearance(bottomNavigationView.menu.children.indexOf(menuItem), tabScreen);
-
+        updateBadgeAppearance(bottomNavigationView.menu.children.indexOf(menuItem), tabScreen)
     }
 
     internal fun onViewManagerAddEventEmitters() {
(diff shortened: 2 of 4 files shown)


android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
```diff
@@ -370,6 +377,28 @@ class TabsHost(
...
     }
 
+    private fun updateBadgeAppearance(
+        menuItemIndex: Int,
+        tabScreen: TabScreen,
+    ) {
+        val badgeValue = tabScreen.badgeValue?.toIntOrNull()
+        if (tabScreen.badgeValue != null && badgeValue == null) {
+            Log.e(TAG, "[RNScreens] Android supports only numbers as badge value")
+        }
```
kkafar: I think that sanitization code should be on setter level, or even on view manager level. View should hold only the valid state (of Int type)

apps/src/tests/TestBottomTabs/index.tsx
```diff
@@ -34,7 +34,7 @@ const TAB_CONFIGS: TabConfiguration[] = [
   {
     tabScreenProps: {
       tabKey: 'Tab2',
-      badgeValue: '2',
+      badgeValue: 'SWM',
```
kkafar: Won't this crash?

kkafar (changes requested):
This looks good overall. I have one request regarding value sanitisation.

apps/src/tests/TestBottomTabs/index.tsx
maciekstosio: It won't if it fail to parse as a string we disable the badge, but this behaviour will be changed in: 
https://github.com/software-mansion/react-native-screens/pull/3025
and
https://github.com/software-mansion/react-native-screens/pull/3024

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
maciekstosio: I get rid of it in https://github.com/software-mansion/react-native-screens/pull/3025/files

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabScreenViewManager.kt
```diff
@@ -172,6 +168,31 @@ class TabScreenViewManager :
...
+    // Android specific
+    @ReactProp(name = "tabBarItemBadgeVisible")
+    override fun setTabBarItemBadgeVisible(
+        view: TabScreen,
+        value: Boolean,
+    ) {
+        view.tabBarItemBadgeVisible = value
+    }
+
+    @ReactProp(name = "tabBarItemBadgeTextColor")
```
kkafar: All the color prop need the `customType = "Color" annotation argument, otherwise it wong work on old architecture. 

```suggestion
    @ReactProp(name = "tabBarItemBadgeTextColor", customType = "Color")
```

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
```diff
@@ -370,6 +379,40 @@ class TabsHost(
...
+        badge.isVisible = true
+
+        if (badgeValueNumber != null) {
+            badge.number = badgeValueNumber
+        } else if (badgeValue != null) {
+            badge.text = badgeValue
+        } else {
+            badge.clearNumber()
+            badge.clearText()
+        }
```
kkafar: Okay, so badgeValue is either a string or a number. If it is a number I guess we get this nice shortening mechanism (picture below 👇🏻), is that right? 

We need to document this behaviour. Please let's start creating short documentation around exposed types (similar to what we have in v4 implementation). We need to have such interactions described. It can be done in separate PR, I don't mind here. 

image

kkafar: There is a bug here. When I set a number as a badge value first, then change it to text, I can not go back to a number. See the recording. 

https://github.com/user-attachments/assets/cc19ca82-11ef-44b3-97b0-61939d2880f9

src/components/BottomTabsScreen.tsx
```diff
@@ -75,6 +75,8 @@ export interface BottomTabsScreenProps {
 
   // Android specific
   iconResourceName?: string;
+  tabBarItemBadgeVisible?: boolean;
```
kkafar: Let's document what is the default value here. Also I wonder whether we need this prop? Shouldn't it be enough to determine badge visibility by the value passed as `badgevalue`? e.g. empty string or undefined means its not visible?

kkafar (changes requested):
Good job overall. I've found a bug, see below. Also I have concerns about necessity of `tabbaritembadgevisible` prop.

src/components/BottomTabsScreen.tsx
maciekstosio: We can do:
- "" -> dot icon
- undefined -> disabled
- string -> string
- string parsable to string -> number
But isn't it more tricky?

kkafar: Now to show a dot you pass `badgeValue: undefined & badgeVisible: true` OR `badgeValue: "" & badgeVisible: true`? 

Therefore to get a dot, user still must consult the documentation. Moreover `""` as a value to get a dot only seems natural to me, since it's a badge w/o text 😅 

I think it should work as follows:

1. We should have a single prop, no badgeVisible
2. badge is not visible when `badgeValue == null`
3. `badgeValue === undefined` by default
4. `badgeValue === ""` results in dot only
5. any other badge value gets current treatment.

Additional logic, visible / invisible etc. can be defined by downstream libraries. Let's keep our API minimal here.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabScreenViewManager.kt
```diff
@@ -142,24 +144,18 @@ class TabScreenViewManager :
     }
 
     override fun setBadgeValue(
```
kkafar, line 147: I don't like that every other prop has `tabBarItem...` prefix and this one does not. But this is outside of the scope of this PR. I'll create ticket for this.

src/components/BottomTabsScreen.tsx
maciekstosio: changed in https://github.com/software-mansion/react-native-screens/pull/3017/commits/69df8cfbb14868d15e4969e89f0c4162ef997d95

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
maciekstosio: fixed in https://github.com/software-mansion/react-native-screens/pull/3017/commits/69df8cfbb14868d15e4969e89f0c4162ef997d95

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabScreenViewManager.kt
maciekstosio: changed in https://github.com/software-mansion/react-native-screens/pull/3017/commits/69df8cfbb14868d15e4969e89f0c4162ef997d95

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
maciekstosio: The fix is basically that we clear first - there are separate properties for text and number, text takes precedence over number, thus it didn't change back - it wasn't cleared when setting number.

PR #3022, opened 2025-07-17
feat(iOS, SplitView): Add support for setting column widths

t0maboro:
## Description

For all primary and supplementary columns (but from iOS 26 - also for secondary and inspector), we can manage their widths, passing width in px for following APIs:

- preferred[COLUMN]ColumnWidth
- maximum[COLUMN]ColumnWidth
- minimum[COLUMN]ColumnWidth

Closes https://github.com/software-mansion/react-native-screens-labs/issues/234 

## Changes

- defined JS types
- added native setters
- updated app

## Screenshots / GIFs

Screenshot 2025-07-17 at 16 28 21

Example with primary and supplementary columns with fixed preferred width set to 250

Updated demo app

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/apps/src/tests/TestSplitView/SplitViewBaseApp.tsx
+++ b/apps/src/tests/TestSplitView/SplitViewBaseApp.tsx
@@ -1,29 +1,26 @@
 import React from 'react';
-import { Button, StyleSheet, Text, View } from 'react-native';
+import { StyleSheet, Text, View } from 'react-native';
 import { SplitViewHost, SplitViewScreen } from 'react-native-screens';
 import { Colors } from '../../shared/styling/Colors';
 
 const SplitViewBaseApp = () => {
-  const [buttonVisibility, setButtonVisibility] = React.useState('automatic');
-
   return (
-    <SplitViewHost displayModeButtonVisibility={buttonVisibility} displayMode='twoBesideSecondary' splitBehavior='tile'>
+    <SplitViewHost 
+      columnMetrics={{preferredPrimaryColumnWidth: 250, preferredSupplementaryColumnWidth: 250}} 
+      displayMode='twoBesideSecondary' 
+      primaryEdge='leading' 
+      presentsWithGesture={false} 
+      splitBehavior='tile'
+    >
+      <SplitViewScreen.Column>
+        <View style={[styles.container, { backgroundColor: Colors.RedDark100 }]} />
+      </SplitViewScreen.Column>
+      <SplitViewScreen.Column>
+        <View style={[styles.container, { backgroundColor: Colors.YellowDark100 }]} />
+      </SplitViewScreen.Column>
       <SplitViewScreen.Column>
         <View style={[styles.container, { backgroundColor: Colors.White }]}>
           <Text style={styles.text}>Primary column</Text>
-        </View>
-      </SplitViewScreen.Column>
-      <SplitViewScreen.Column>
-        <View style={[styles.container, { backgroundColor: Colors.YellowDark100 }]}>
-          <Text style={styles.text}>DisplayModeButtonVisibility demo</Text>
-          <Button onPress={() => setButtonVisibility('always')} title='always' />
-          <Button onPress={() => setButtonVisibility('automatic')} title='automatic' />
-          <Button onPress={() => setButtonVisibility('never')} title='never' />
-        </View>
-      </SplitViewScreen.Column>
-      <SplitViewScreen.Column>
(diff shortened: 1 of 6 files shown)


kligarski (approved):

apps/src/tests/TestSplitView/SplitViewBaseApp.tsx
```diff
@@ -1,29 +1,26 @@
...
 import { SplitViewHost, SplitViewScreen } from 'react-native-screens';
 import { Colors } from '../../shared/styling/Colors';
 
 const SplitViewBaseApp = () => {
-  const [buttonVisibility, setButtonVisibility] = React.useState('automatic');
-
   return (
-    <SplitViewHost displayModeButtonVisibility={buttonVisibility} displayMode='twoBesideSecondary' splitBehavior='tile'>
+    <SplitViewHost 
+      columnMetrics={{preferredSupplementaryColumnWidth: 250}}
```
kkafar, line 9: I like that you grouped these props together.

kkafar: In the future we might consider support for percentages.

apps/src/tests/TestSplitView/SplitViewBaseApp.tsx
```diff
@@ -1,29 +1,26 @@
...
-          <Button onPress={() => setButtonVisibility('automatic')} title='automatic' />
-          <Button onPress={() => setButtonVisibility('never')} title='never' />
-        </View>
+        <View style={[styles.container, { backgroundColor: Colors.YellowDark100 }]} />
       </SplitViewScreen.Column>
       <SplitViewScreen.Column>
-        <View style={[styles.container, { backgroundColor: Colors.RedDark100 }]}>
-          <Text style={styles.text}>Secondary column</Text>
+        <View style={[styles.container, { backgroundColor: Colors.White }]}>
+          <Text style={styles.text}>Primary column</Text>
```
kkafar: Why is this column changed from secondary to primary? 🙈

kkafar: Was it wrong before?

kkafar: It would be nice to have text in each column, indicating which is which

ios/gamma/split-view/RNSSplitViewAppearanceCoordinator.mm
```diff
@@ -5,6 +5,11 @@
 
 @implementation RNSSplitViewAppearanceCoordinator
 
+#define ASSIGN_PROP_IF_NONNEGATIVE(target, source, property) \
+  if ((source).property >= 0) {                              \
+    (target).property = (source).property;                   \
+  }
```
kkafar, line 11: It would be nice to #undef this macro after its final usage. It's defined in implementation file, so it won't leak much, however it won't hurt.

kkafar (changes requested):
Okay, few remarks. Beside that we should be good. Well done. 

Additionally I've found a bug: 

image


We need to verify whether it does happen in purely UIKit app or not.

t0maboro:
> Okay, few remarks. Beside that we should be good. Well done.
> 
> Additionally I've found a bug:
> 
>  Screen.Recording.2025-07-18.at.15.00.30.mov 
> We need to verify whether it does happen in purely UIKit app or not.

https://github.com/software-mansion/react-native-screens-labs/issues/294 adding for the further investigation

apps/src/tests/TestSplitView/SplitViewBaseApp.tsx
t0maboro: https://github.com/software-mansion/react-native-screens-labs/issues/296 creating a ticket to do this as a followup

apps/src/tests/TestSplitView/SplitViewBaseApp.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3022/commits/27810825d039b72ce7919ae93180a5af16d74d4c

ios/gamma/split-view/RNSSplitViewAppearanceCoordinator.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3022/commits/627abe584091d1475952b4b4ab354170938c5fd5

kkafar (approved):
Thanks!

PR #3026, opened 2025-07-17
feat(Android, Tabs): Align styles with material design guidelines

maciekstosio:
## Description
Resolves: https://github.com/software-mansion/react-native-screens-labs/issues/225

This PR adjust our default (fallback) values with material values, based on https://m3.material.io/components/navigation-bar/specs and package resources. 

In #3015 I introduced custom theme so we can pass it to context wrapper - because I wanted the theme to inherit from `Theme_Material3_DayNight_NoActionBar`. Now I realized that I can just pass `Theme_Material3_DayNight_NoActionBar` 😅, so I removed unnecessary custom theme.

## Screenshots / GIFs

image

## Checklist

- [ ] Ensured that CI passes


kkafar (commented):
I'll wait with proper review here, for #3017, since I see some duplicated changes & I don't know exactly what is what.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
```diff
@@ -94,7 +95,7 @@ class TabsHost(
 
     private val containerUpdateCoordinator = ContainerUpdateCoordinator()
 
-    private val wrappedContext = ContextThemeWrapper(reactContext, R.style.custom)
+    private val wrappedContext = ContextThemeWrapper(reactContext, com.google.android.material.R.style.Theme_Material3_DayNight_NoActionBar)
```
kkafar, line 99: I'm interested in what happens if we use these symbols inside our custom XML resources. Will they still trigger warnings? Maybe this is the solution. I'm talking about all symbols here: colours, font sizes, themes, etc.

Right now, I think we should proceed with the changes you suggested in this PR. We should experiment with custom xmls, however. Would you mind creating a ticket for this?

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
```diff
@@ -287,27 +288,33 @@ class TabsHost(
 
         bottomNavigationView.isVisible = true
         bottomNavigationView.setBackgroundColor(
-            tabBarBackgroundColor ?: com.google.android.material.R.color.m3_sys_color_light_surface_container,
+            tabBarBackgroundColor ?: wrappedContext.getColor(com.google.android.material.R.color.m3_sys_color_light_surface_container),
```
kkafar, line 292: This also fixes the issue with incorrect colours, right?

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
```diff
@@ -366,15 +373,19 @@ class TabsHost(
...
+            val smallFontSize =
+                tabBarItemTitleFontSize?.takeIf { it > 0 }?.let { it * resources.displayMetrics.density }
+                    ?: wrappedContext.resources.getDimension(com.google.android.material.R.dimen.design_bottom_navigation_text_size)
+            val largeFontSize =
+                tabBarItemTitleFontSizeActive?.takeIf { it > 0 }?.let { it * resources.displayMetrics.density }
+                    ?: wrappedContext.resources.getDimension(com.google.android.material.R.dimen.design_bottom_navigation_text_size)
 
             // Inactive
-            smallLabel.textSize = smallFontSize
+            smallLabel.setTextSize(TypedValue.COMPLEX_UNIT_PX, smallFontSize)
```
kkafar, line 394: Hmm, I'm not so sure about using px here. I have not looked up the spec now, but at a first glance it would make little sense for design system to specify dimensions in `px`. I expect `dip` is what we want & I believe it is usually useD in RN to specify corner radiuses etc. 

Can I ask you to verify this in reference specs & link them if they really specify values in `px`?

apps/src/shared/gamma/containers/bottom-tabs/BottomTabsContainer.tsx
```diff
@@ -85,19 +85,19 @@ export function BottomTabsContainer(props: BottomTabsContainerProps) {
...
+      // tabBarItemBadgeBackgroundColor={Colors.GreenDark100}
+      // tabBarItemIconColor={Colors.BlueLight100}
+      // tabBarItemTitleFontColor={Colors.BlueLight40}
+      // tabBarItemIconColorActive={Colors.GreenLight100}
+      // tabBarItemTitleFontColorActive={Colors.GreenLight40}
+      // tabBarItemTitleFontSize={10}
+      // tabBarItemTitleFontSizeActive={15}
+      // tabBarItemTitleFontFamily='monospace'
+      // tabBarItemTitleFontStyle='italic'
+      // tabBarItemTitleFontWeight="700"
```
kkafar: What's going on here? Why do we comment all these props?

kkafar: I feel like this should not be here, in this PR.

kkafar (commented):
Few remarks & questions

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
maciekstosio: Yes

maciekstosio: image

apps/src/shared/gamma/containers/bottom-tabs/BottomTabsContainer.tsx
maciekstosio: Sure, I'll change it. This was on purpose, so the TestBottomTabs showcases PR changes

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
maciekstosio: https://github.com/software-mansion/react-native-screens-labs/issues/61?issue=software-mansion%7Creact-native-screens-labs%7C301

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
maciekstosio: It's actually in SP as the other retrieved from resources. I calculate it higher in the code

kkafar: You're right! Overlooked that. It does make sense now. One thing though, is that we could use `PixelUtil` utility type.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
kkafar: Thanks!

kkafar (approved):
Okay, I think we're looking good. Please research usage of `PixelUtil`. If it is applicable, please use it. Otherwise let's proceed.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
maciekstosio: Yeah, I used it for conversion. But the rest left the same, as I stated in comment. I assume that size for bottom navigation will be in SP, RN decides between SP and DP using `allowFontScaling`, which is by default `true`.

PR #3027, opened 2025-07-17
feat(Android, Tabs): support ripple color in bottom navigation

maciekstosio:
## Description

Resolves: https://github.com/software-mansion/react-native-screens-labs/issues/136

This PR adds support for ripple color to bottom navigation 

## Screenshots / GIFs

https://github.com/user-attachments/assets/4cc128ab-d92e-4e07-9f46-e6b54a8e9ea0

## Checklist

- [ ] Ensured that CI passes


android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHostViewManager.kt
```diff
@@ -179,6 +179,13 @@ class TabsHostViewManager :
         view?.tabBarItemTitleFontSizeActive = value
     }
 
+    override fun setTabBarItemRippleColor(
```
kkafar, line 183: Please add @ReactProp annotation here with `customType = "Color"`, otherwise it won't work with legacy architecture.

kkafar (changes requested):
One comment regarding legacy architecture. Other than that this looks really solid. Thank you!

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHostViewManager.kt
maciekstosio: changed in https://github.com/software-mansion/react-native-screens/pull/3027/commits/5c64f38b2e23d05e491a122a81c605b07f8dacc6

kkafar (approved):
We good! Thanks!

PR #3028, opened 2025-07-17
feat(Android, Tabs): support visiblity mode in bottom navigation

maciekstosio:
## Description

Resolves: https://github.com/software-mansion/react-native-screens-labs/issues/137

This PR adds support for item visibility mode in bottom navigation. I needed to add refresh layout on requestLayout, otherwise the animation while switching items did not work. 

## Screenshots / GIFs

https://github.com/user-attachments/assets/903eda38-7458-4451-981d-9d1de897f486

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/src/components/BottomTabs.tsx
+++ b/src/components/BottomTabs.tsx
@@ -49,7 +49,7 @@ export interface BottomTabsProps extends ViewProps {
   tabBarItemIconColorActive?: ColorValue;
   tabBarItemActivityIndicatorColor?: ColorValue;
   tabBarItemRippleColor?: ColorValue;
-  tabBarItemVisibilityMode?: TabBarItemLabelVisibilityMode;
+  tabBarItemLabelVisibilityMode?: TabBarItemLabelVisibilityMode;
 
   // Control
 


src/fabric/BottomTabsNativeComponent.ts
```diff
@@ -42,6 +42,12 @@ export type BlurEffect =
   | 'systemThickMaterialDark'
   | 'systemChromeMaterialDark';
 
+export type VisibilityMode =
```
kkafar: If would be nice if the type described visibility mode of what it describes. 
E.g. TabBarItemLabelVisibilityMode.

kkafar: Otherwise its hard to tell from the type name what it describes.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
```diff
@@ -482,6 +498,11 @@ class TabsHost(
         }
     }
 
+    override fun requestLayout() {
+        super.requestLayout()
+        refreshLayout()
+    }
```
kkafar, line 516: ay ay, we're really asking for troubles here 😅 I'll try to debug it in a few a bit, to see whether we can do better.

kkafar (commented):
I really don't like overriding `requestLayout` this way. I'll try to debug it a little now. In meantime I've left few suggestions.

kkafar:
Also I get this error when I open the project on this branch

image

Is this known to you?

kkafar:
Is `auto` the default visibility in material original implementation? I'm asking, because it seems weird to be that by default there are no labels on not focused item.

maciekstosio:
I took the default value from here: https://github.com/material-components/material-components-android/blob/master/docs/components/BottomNavigation.md#navigation-bar-item-attributes

src/fabric/BottomTabsNativeComponent.ts
maciekstosio: changed in https://github.com/software-mansion/react-native-screens/pull/3028/commits/527a5a18ab9261092e83f1d876146d2a6f860c2e

kkafar (approved):
Okay let's proceed here. 

I've did some research, however was not able to get this working smoothly w/o the trick with layout schedule for the current loop. 

| w/o requestlayout | w/ requestlayout |
| -- | -- | 
| <video src="https://github.com/user-attachments/assets/e4b2a9cb-47b5-4c03-ab84-7c47de98ff07" alt="w/o request layout video" /> | <video src="https://github.com/user-attachments/assets/891161f0-686c-4787-80fc-794ac2234385" alt="w/ request layout video" /> |

I'll create ticket to research this further.

PR #3029, opened 2025-07-17
chore(Android, Tabs): sync architectures

maciekstosio:
## Description

Sync archs after adding new props, so the paper builds. 

## Checklist

- [ ] Ensured that CI passes


kkafar (approved):
Yeah, thanks!

maciekstosio:
I think this is no longer needed as CI requires to sync the arch now

PR #3030, opened 2025-07-17
feat(Android, Tabs): Allow to disable active indicator

maciekstosio:
## Changes
I added prop that allows to disable active indicator. At the same time I noticed it's active not activity, so I fixed the typo in both places. 

## Screenshots / GIFs

https://github.com/user-attachments/assets/384cd2fd-8570-4ae4-a616-466373be31fa

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
@@ -129,7 +129,11 @@ class TabsHost(
         updateNavigationMenuIfNeeded(oldValue, newValue)
     }
 
-    var tabBarItemActivityIndicatorColor: Int? by Delegates.observable<Int?>(null) { _, oldValue, newValue ->
+    var tabBarItemActiveIndicatorColor: Int? by Delegates.observable<Int?>(null) { _, oldValue, newValue ->
+        updateNavigationMenuIfNeeded(oldValue, newValue)
+    }
+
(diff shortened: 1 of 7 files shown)


android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
```diff
@@ -129,7 +129,11 @@ class TabsHost(
         updateNavigationMenuIfNeeded(oldValue, newValue)
     }
 
-    var tabBarItemActivityIndicatorColor: Int? by Delegates.observable<Int?>(null) { _, oldValue, newValue ->
+    var tabBarItemActiveIndicatorColor: Int? by Delegates.observable<Int?>(null) { _, oldValue, newValue ->
+        updateNavigationMenuIfNeeded(oldValue, newValue)
+    }
+
+    var tabBarItemActiveIndicatorEnabled: Boolean by Delegates.observable<Boolean>(false) { _, oldValue, newValue ->
```
kkafar, line 136: let's keep the kotlin naming convention here. Name the variable with `is` prefix. Only the variable, prop can stay for now.

kkafar (commented):
This looks good. Only a single remark

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
maciekstosio: fixed in https://github.com/software-mansion/react-native-screens/pull/3053

maciekstosio:
replaced with https://github.com/software-mansion/react-native-screens/pull/3053

PR #3034, opened 2025-07-18
chore(iOS, SplitView): Export types and add documentation coverage

t0maboro:
## Description

Adding proper type definitions for SplitView APIs to external types.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/290

## Changes

- Added new files with type definitions and documentation for external props
- Hidden `NativeProps` definitions inside native components

N/A - no changes affecting current SplitView behavior

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/components/gamma/SplitViewHost.types.ts
+++ b/src/components/gamma/SplitViewHost.types.ts
@@ -41,9 +41,9 @@ export interface SplitViewColumnMetrics {
   /**
    * @summary Preferred width for the primary sidebar.
    *
-   * Specifies the preferred width (in points) for the primary column in the SplitView layout, typically representing the leftmost sidebar.
+   * Specifies the preferred width (in points or as a fraction for percentage width support) for the primary column in the SplitView layout, typically representing the leftmost sidebar.
    */
-  preferredPrimaryColumnWidth?: number;
+  preferredPrimaryColumnWidthOrFraction?: number;
   /**
    * @summary Minimum width for the intermediate sidebar.
    *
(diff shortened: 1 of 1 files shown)


kligarski (commented):
1. I think that we need to use `SplitViewHostProps` interface in the `SplitViewHost` component to make IDE type hints work properly, same with Screen.
2. Remember to include changed props from e.g. https://github.com/software-mansion/react-native-screens/pull/3043 in the type docs as well, not sure in what order to do so.

src/types.tsx
```diff
@@ -913,3 +913,244 @@ export interface GestureProviderProps extends GestureProps {
...
+   * @platform ios
+   */
+  preferredSupplementaryColumnWidth?: number;
+
+  // iOS 26 only
+  /**
+   * Specifies the minimum width for the secondary column in the SplitView layout, typically for the view with the main content.
+   *
+   * @platform iOS 26+
+   * @supported iOS 26 or higher
```
kligarski:
```suggestion
   * @platform ios
   * @supported iOS 26 or higher
```
If we're using `@supported` to note supported versions, maybe we should leave `@platform` without version information?

src/types.tsx
```diff
@@ -913,3 +913,244 @@ export interface GestureProviderProps extends GestureProps {
...
+
+export interface SplitViewHostProps extends ViewProps {
+  children?: React.ReactNode;
+
+  /**
+   * An object describing bounds for column widths.
+   *
+   * It supports definitions for the following columns:
+   *
+   * - "primary" - the leftmost sidebar
```
kligarski: Just want unify if we want to use ", ' or ` for prop values inside docs?

src/types.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3034/commits/9f56c648979add47c959586c391e8864a719a1d9

src/types.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3034/commits/89927b57f4389b2479dc825a8a4377cac5e0de82

kligarski (approved):

kkafar (approved):
Looks good.

PR #3036, opened 2025-07-18
feat(iOS, Tabs): add support for `default` blur effect value

kligarski:
## Description

Adds support for `default` blur effect value.

Previously, `tabBarBlurEffect: 'none'` would enable default UIKit blur effect but this made disabling blur effect impossible. That's why we decided to add separate option for default value. Now, `none` disables the blur completely.

I decided to create new enum for blur styles instead of using previous one in order not to introduce any breaking changes for old stack implementation. When it gets removed, new enum can become the only one.

> [!IMPORTANT]
> UIKit's API allows setting appearance for host and tab bar items. This however seems problematic - deciding which and how properties are inherited. We think that this should be handled by the users of the library and `react-native-screens` should expose common appearance props only via screen component. This will be handled in another PR. Changes from this PR will work as expected after changing props inheritance behavior.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/108.

## Changes

- add `RNSExtendedBlurEffectStyle`
- update prop for Host and Screen component
- update appearance coordinator's logic to use new enum
- update conversion funcitons

Run TestBottomTabs, add `tabBarBackgroundColor: 'transparent'`. Set `tabBarBlurEffect` to `none` or `default` (without any prop set on screen).

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSEnums.h
+++ b/ios/RNSEnums.h
@@ -78,7 +78,7 @@ typedef NS_ENUM(NSInteger, RNSSplitViewScreenColumnType) {
   RNSSplitViewScreenColumnTypeInspector,
 };
 
-// Redefinition of UIBlurEffectStyle. We need to represent additional cases of `None` and `Default`.
+// Redefinition of UIBlurEffectStyle. We need to represent additional cases of `None` and `SystemDefault`.
 typedef NS_ENUM(NSInteger, RNSBlurEffectStyle) {
   /// Default blur effect should be used
   RNSBlurEffectStyleSystemDefault = -2,
--- a/ios/RNSEnums.h
+++ b/ios/RNSEnums.h
@@ -78,8 +78,10 @@ typedef NS_ENUM(NSInteger, RNSSplitViewScreenColumnType) {
   RNSSplitViewScreenColumnTypeInspector,
 };
 
-// Redefinition of UIBlurEffectStyle. We need to represent additional case of `None`.
+// Redefinition of UIBlurEffectStyle. We need to represent additional cases of `None` and `Default`.
 typedef NS_ENUM(NSInteger, RNSBlurEffectStyle) {
+  /// Default blur effect should be used
+  RNSBlurEffectStyleDefault = -2,
   /// No blur effect should be visible
   RNSBlurEffectStyleNone = -1,
   RNSBlurEffectStyleExtraLight = UIBlurEffectStyleExtraLight,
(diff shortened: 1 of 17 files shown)


ios/bottom-tabs/RNSTabBarAppearanceCoordinator.mm
```diff
@@ -72,8 +73,23 @@ - (void)configureTabBarAppearance:(nonnull UITabBarAppearance *)appearance
...
+
+      // This won't work as expected with current inheriting appearance logic:
+      // screen will not revert to default but inherit blur from host
+      // TODO: remove appearance inheritance
+      break;
+
+    default:
+      appearance.backgroundEffect =
+          rnscreens::conversion::RNSUIBlurEffectFromRNSExtendedBlurEffectStyle(appearanceProvider.tabBarBlurEffect);
+      break;
```
t0maboro: not needed

kligarski: https://github.com/software-mansion/react-native-screens/pull/3036/commits/f359f0c84ca4ac93d36a4c05827b8320e116da46

kkafar (commented):
Verbal review 😄 

too long; didn't hear:

Would be nice to have only a single enum, shared between legacy & modern implementation. We can achieve that simply by using different default values for the enum in different components.

ios/RNSEnums.h
```diff
@@ -78,8 +78,10 @@ typedef NS_ENUM(NSInteger, RNSSplitViewScreenColumnType) {
   RNSSplitViewScreenColumnTypeInspector,
 };
 
-// Redefinition of UIBlurEffectStyle. We need to represent additional case of `None`.
+// Redefinition of UIBlurEffectStyle. We need to represent additional cases of `None` and `Default`.
```
kkafar:
```suggestion
// Redefinition of UIBlurEffectStyle. We need to represent additional cases of `None` and `SystemDefault`.
```

ios/bottom-tabs/RNSTabBarAppearanceCoordinator.mm
```diff
@@ -72,8 +73,22 @@ - (void)configureTabBarAppearance:(nonnull UITabBarAppearance *)appearance
...
+    case RNSBlurEffectStyleNone:
+      appearance.backgroundEffect = nil;
+      break;
+
+    case RNSBlurEffectStyleSystemDefault:
+      // Initialized appearance already has default blur effect.
+
+      // This won't work as expected with current inheriting appearance logic:
+      // screen will not revert to default but inherit blur from host
+      // TODO: remove appearance inheritance
```
kkafar, line 86: Thanks for this comment!

ios/RNSEnums.h
kligarski: https://github.com/software-mansion/react-native-screens/pull/3036/commits/d2cb613792047bd11079e452c4dff75dfaa1c34c

PR #3037, opened 2025-07-18
feat!(iOS, Tabs): remove appearance configuration from Host component

kligarski:
## Description

Removes tab appearance configuration from Host component.

Reasoning behind this change can be found here: https://github.com/software-mansion/react-native-screens/pull/3036#issue-3242781419.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/291.

## Changes

- remove appearance configuration from Host component
- remove redundant conversion code
- clean up TS component + codegen specs to better differentiate between Android and iOS-specific props.

Run TestBottomTabs.

## Checklist

- [x] Updated TS types
- [ ] Ensured that CI passes


src/fabric/BottomTabsNativeComponent.ts
```diff
@@ -49,28 +49,23 @@ export interface NativeProps extends ViewProps {
...
-
-  tabBarItemBadgeBackgroundColor?: ColorValue;
-
-  // Android
-  tabBarItemTitleFontColorActive?: ColorValue;
   tabBarItemIconColorActive?: ColorValue;
-  tabBarItemTitleFontSizeActive?: Float;
+
+  // iOS-only appearance
+  tabBarTintColor?: ColorValue;
```
t0maboro, line 86: Why it's marked as iOS-only? I'm seeing some code around it for both platforms?

t0maboro (approved):
one minor comment, but overall it looks good

src/fabric/BottomTabsNativeComponent.ts
kligarski: I think there is only an empty implementation of the setter. Thank you for noticing though!

kkafar:
I gave this another moment of thinking yesterday & still haven't settled on a single idea here. 
I'll let you know as soon as I've clear plan for what should happen here.

kligarski:
Superseded by https://github.com/software-mansion/react-native-screens/pull/3078.

PR #3038, opened 2025-07-18
fix(iOS, Paper, Tabs): fix badge value on iOS Paper

kligarski:
## Description

Fixes missing badge values on iOS Paper.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/279.

## Changes

- update badge value in tab bar item when setting prop

Run TestBottomTabs and verify if badges are visible.

## Checklist

- [x] Ensured that CI passes


kkafar (approved):
Thanks for catching this.

PR #3042, opened 2025-07-21
chore(iOS, SplitView): Add internal documentation for SplitView

t0maboro:
## Description

This PR covers current SplitView APIs with a documentation in Doxygen format.

## Changes

- Covered all native SplitView classes with documentation

No functional changes

## Checklist

- [ ] Ensured that CI passes


ios/gamma/split-view/RNSSplitViewHostController.swift
```diff
@@ -122,6 +174,12 @@ public class RNSSplitViewHostController: UISplitViewController, ReactMountingTra
 
   // MARK: Public setters
 
+  /**
+   * @brief Shows or hides the inspector screen.
+   * @remarks Available only on iOS26 or higher.
```
kligarski:
```suggestion
   * @remarks Available only on iOS 26 or higher.
```
Applies to other files as well.

kligarski (approved):

ios/gamma/split-view/RNSSplitViewHostController.swift
```diff
@@ -20,6 +25,12 @@ public class RNSSplitViewHostController: UISplitViewController, ReactMountingTra
   private let maxNumberOfColumns: Int = 3
   private let maxNumberOfInspectors: Int = 1
 
+  /**
+   * @brief Initializes the SplitView host controller.
+   *
+   * @param splitViewHostComponentView The view managed by this controller.
+   * @param numberOfColumns Expected number of visible columns.
+   */
```
kmichalikk: `yarn lint-swift` wants `///` instead of block comments

for every comment added in .swift unfortunately 😞

ios/gamma/split-view/RNSSplitViewHostController.swift
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3042/commits/997e1e2917a8678751c9c9b1aac834e12785e167

ios/gamma/split-view/RNSSplitViewHostController.swift
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3042/commits/76a574ac1872515f147e0346d69be3abb6545dc8

ios/gamma/split-view/RNSSplitViewAppearanceCoordinator.h
```diff
@@ -2,8 +2,23 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/**
+ * @class RNSSplitViewAppearanceCoordinator
+ * @brief Responsible for batching updates for appearance.
```
kkafar:
```suggestion
 * @brief Responsible for batching updates for appearance & applying them in appropriate order.
```

ios/gamma/split-view/RNSSplitViewHostComponentEventEmitter.h
```diff
@@ -10,14 +10,37 @@ namespace react = facebook::react;
...
 @interface RNSSplitViewHostComponentEventEmitter : NSObject
 
+/**
+ * @brief Emits the onCollapse event to notify React Native that the SplitView has collapsed.
+ *
+ * This event is triggered when the SplitView collapses to a single column due to
+ * size constraints or user interaction.
+ *
+ * @return true if the event was successfully emitted, false otherwise.
+ */
```
kkafar: I think that these comments are a bit off. What I mean is that the doc comment should describe *what this method does* and (optionally) *when / in what context it should/could be used* plus (optionally) some additional important considerations. While the comment in current form describes *how the method is *currently* used*.

I think that "*how the method is currently used*" might be an additional & optional detail, but only that. 

Therefore, my suggestion would be to phrase it as follows: 

```suggestion
 * @brief Emits the onCollapse event to notify associated component instance that the SplitView has collapsed.
 *
 * Call this method when the SplitView collapses to a single column e.g. due to
 * size constraints or user interaction to notify React realm of this event.
 *
 * @return true if the event was successfully emitted, false otherwise.
 */
```

ios/gamma/split-view/RNSSplitViewHostComponentView.h
```diff
@@ -7,50 +7,144 @@ NS_ASSUME_NONNULL_BEGIN
...
 @property (nonatomic, readonly) UISplitViewControllerSplitBehavior splitBehavior;
+
+/**
+ * @brief Determines on which side the primary sidebar is placed relative to the content.
+ */
 @property (nonatomic, readonly) UISplitViewControllerPrimaryEdge primaryEdge;
+
+/**
+ * @brief Determines the display mode for the split view.
+ */
```
kkafar: These comments are ok, and let's let them stay. However in the future such comments are not necessary, as they really do not bring any added value. They basically repeat type & prop name ;D

ios/gamma/split-view/RNSSplitViewHostComponentView.h
```diff
@@ -7,50 +7,144 @@ NS_ASSUME_NONNULL_BEGIN
...
 @property (nonatomic, readonly) UISplitViewControllerDisplayMode displayMode;
+
+/**
+ * @brief Defines whether the system's display mode button is shown automatically.
+ */
 @property (nonatomic, readonly) UISplitViewControllerDisplayModeButtonVisibility displayModeButtonVisibility;
+
+/**
+ * @brief Enables presentation/dismissal of hidden column using swipe gesture.
+ */
```
kkafar: E.g. this doc comment **is** useful & nice that we have it as it brings some additional context not directly inferable from the prop type & name.

ios/gamma/split-view/RNSSplitViewHostController.swift
```diff
@@ -34,41 +45,61 @@ public class RNSSplitViewHostController: UISplitViewController, ReactMountingTra
...
+  ///
+  /// @brief Marks the secondary component's UINavigationBar as needing update.
+  ///
   @objc
   public func setNeedsSecondaryScreenNavBarUpdate() {
+    /// We noticed a bug on the pure-native component, which is blocking dynamic updates for showsSecondaryOnlyButton.
+    /// Toggling this flag doesn't refresh the component and is updated after triggerig some other interaction, like changing layout.
+    /// We noticed that we can forcefully refresh navigation bar from UINavigationController level by toggling setNavigationBarHidden.
+    /// After some testing, it looks well and I haven't noticed any flicker - missing button is appearing naturally.
+    /// Please note that this is a hack rather than a solution so feel free to remove this code in case of any problems and treat the bug with toggling button as a platform's issue.
```
kkafar:
```suggestion
    // We noticed a bug on the pure-native component, which is blocking dynamic updates for showsSecondaryOnlyButton.
    // Toggling this flag doesn't refresh the component and is updated after triggerig some other interaction, like changing layout.
    // We noticed that we can forcefully refresh navigation bar from UINavigationController level by toggling setNavigationBarHidden.
    // After some testing, it looks well and I haven't noticed any flicker - missing button is appearing naturally.
    // Please note that this is a hack rather than a solution so feel free to remove this code in case of any problems and treat the bug with toggling button as a platform's issue.
```

ios/gamma/split-view/RNSSplitViewHostController.swift
```diff
@@ -34,41 +45,61 @@ public class RNSSplitViewHostController: UISplitViewController, ReactMountingTra
...
   public func updateChildViewControllersIfNeeded() {
     if needsChildViewControllersUpdate {
       updateChildViewControllers()
     }
   }
 
+  ///
+  /// @brief Creates and attaches the SplitView child controllers based on the current React subviews.
+  ///
+  /// It validates constraints for SplitView hierarchy, then attaches SplitViewScreen representatives to SplitViewHost component.
```
kkafar: It would be nice to mention here that this method will crash if the appropriate flags are not invalidated beforehand.

kkafar (approved):
I haven't read every single comment, just took a quick look & added some remarks. 

General remark is that it is not necessary to have a doc comment for every single method / property, especially in cases where the comment does not give any additional context about method semantics / behaviour above repeating return value type & prop name. 

My suggestion would be to comment only where there are some internal invariants or behaviours / interactions that are not contained directly in the prop name. 

This is in large part a matter of taste, therefore I won't contest this that hard.

ios/gamma/split-view/RNSSplitViewAppearanceCoordinator.h
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3042/commits/37497f3c8ff8348e4fc6fe8e0bdcd7c66cead660

ios/gamma/split-view/RNSSplitViewHostComponentEventEmitter.h
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3042/commits/5dd5769a64a07a2ec41cfaf533f6b4330c11068e

ios/gamma/split-view/RNSSplitViewHostController.swift
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3042/commits/0ac4e5db2d9f49d6e6a78ca8e24f2ed1327df140

ios/gamma/split-view/RNSSplitViewHostComponentView.h
t0maboro: removed these comments that were not adding any value

ios/gamma/split-view/RNSSplitViewHostController.swift
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3042/commits/dab0a82d6996f6dcdebf9bc199f140f3dbf5c646

PR #3043, opened 2025-07-21
feat(iOS, SplitView): Add support for fraction for columns widths

t0maboro:
## Description

In this PR I'm covering UISplitViewController APIs `preferred[ColumnType]ColumnWidthFraction`. As we discussed internally, we should try to cover it as the same prop for both fraction and points. This is slightly different from the Apple approach which has both versions separated, maintaining two props: `preferred[ColumnType]ColumnWidth` and `preferred[ColumnType]ColumnWidthFraction`.

In our case, if the value passed is between 0 and 1 - we're interpreting it as a fraction and for any value greater than 1, we're using points.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/296 

## Changes

- Updated `preferredColumnWidth` definitions to cover floats.
- Updated appearance coordinator to cover fractions.
- Updated prop definition in the host component.

## Screenshots / GIFs

https://github.com/user-attachments/assets/07ef660c-f74e-4bc8-b855-8d197e4db840

Updated example for SplitView for testing this API

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/gamma/split-view/RNSSplitViewAppearanceCoordinator.mm
+++ b/ios/gamma/split-view/RNSSplitViewAppearanceCoordinator.mm
@@ -5,11 +5,6 @@
 
 @implementation RNSSplitViewAppearanceCoordinator
 
-#define ASSIGN_PROP_IF_NONNEGATIVE(target, source, property) \
-  if ((source).property >= 0) {                              \
-    (target).property = (source).property;                   \
-  }
-
 - (void)updateAppearanceOfSplitView:(RNSSplitViewHostComponentView *_Nonnull)splitView
                      withController:(RNSSplitViewHostController *_Nonnull)controller
 {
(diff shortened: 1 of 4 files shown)


ios/gamma/split-view/RNSSplitViewAppearanceCoordinator.mm
```diff
@@ -26,27 +21,77 @@ - (void)updateAppearanceOfSplitView:(RNSSplitViewHostComponentView *)splitView
...
+
+  if (splitView.preferredInspectorColumnWidthOrFraction >= 0 && splitView.preferredInspectorColumnWidthOrFraction < 1) {
+    if (@available(iOS 26.0, *)) {
+      controller.preferredInspectorColumnWidthFraction = splitView.preferredInspectorColumnWidthOrFraction;
+    }
+  } else if (splitView.preferredInspectorColumnWidthOrFraction >= 1) {
+    if (@available(iOS 26.0, *)) {
+      controller.preferredInspectorColumnWidth = splitView.preferredInspectorColumnWidthOrFraction;
+    }
+  }
```
kligarski: Can we wrap this in one `if (@available(iOS 26.0, *))` instead of repeating it for every block?

ios/gamma/split-view/RNSSplitViewAppearanceCoordinator.mm
```diff
@@ -26,27 +21,77 @@ - (void)updateAppearanceOfSplitView:(RNSSplitViewHostComponentView *)splitView
...
 
   // Step 2 - manipulating columns
-  [controller toggleSplitViewInspector:splitView.showInspector];
+  if (splitView.minimumPrimaryColumnWidth >= 0) {
+    controller.minimumPrimaryColumnWidth = splitView.minimumPrimaryColumnWidth;
+  }
+
+  if (splitView.maximumPrimaryColumnWidth >= 0) {
+    controller.maximumPrimaryColumnWidth = splitView.maximumPrimaryColumnWidth;
+  }
```
kmichalikk: Should we account for situations where user passes minimum > maximum ?

ios/gamma/split-view/RNSSplitViewAppearanceCoordinator.mm
```diff
@@ -26,27 +21,77 @@ - (void)updateAppearanceOfSplitView:(RNSSplitViewHostComponentView *)splitView
...
 
-  ASSIGN_PROP_IF_NONNEGATIVE(controller, splitView, minimumPrimaryColumnWidth);
-  ASSIGN_PROP_IF_NONNEGATIVE(controller, splitView, maximumPrimaryColumnWidth);
-  ASSIGN_PROP_IF_NONNEGATIVE(controller, splitView, preferredPrimaryColumnWidth);
+  if (splitView.preferredPrimaryColumnWidthOrFraction >= 0 && splitView.preferredPrimaryColumnWidthOrFraction < 1) {
+    controller.preferredPrimaryColumnWidthFraction = splitView.preferredPrimaryColumnWidthOrFraction;
+  } else if (splitView.preferredSupplementaryColumnWidthOrFraction >= 1) {
+    controller.preferredPrimaryColumnWidth = splitView.preferredPrimaryColumnWidthOrFraction;
+  }
```
kmichalikk, line 51: What if user passes 1.5 ? Should we round it up to next integer? or show a warning?

ios/gamma/split-view/RNSSplitViewAppearanceCoordinator.mm
```diff
@@ -26,27 +21,77 @@ - (void)updateAppearanceOfSplitView:(RNSSplitViewHostComponentView *)splitView
...
 
-  ASSIGN_PROP_IF_NONNEGATIVE(controller, splitView, minimumSupplementaryColumnWidth);
-  ASSIGN_PROP_IF_NONNEGATIVE(controller, splitView, maximumSupplementaryColumnWidth);
-  ASSIGN_PROP_IF_NONNEGATIVE(controller, splitView, preferredSupplementaryColumnWidth);
+  if (splitView.preferredSupplementaryColumnWidthOrFraction >= 0 &&
+      splitView.preferredSupplementaryColumnWidthOrFraction < 1) {
+    controller.preferredSupplementaryColumnWidthFraction = splitView.preferredSupplementaryColumnWidthOrFraction;
+  } else if (splitView.preferredSupplementaryColumnWidthOrFraction >= 1) {
+    controller.preferredSupplementaryColumnWidth = splitView.preferredSupplementaryColumnWidthOrFraction;
+  }
```
kmichalikk: same as above ... ?

ios/gamma/split-view/RNSSplitViewAppearanceCoordinator.mm
kmichalikk: Or maybe even extract a method with `if !available return` ?

kmichalikk (changes requested):
Please rebase to have Podfile.lock updated with RN 80.1

ios/gamma/split-view/RNSSplitViewAppearanceCoordinator.mm
t0maboro: https://developer.apple.com/documentation/uikit/uisplitviewcontroller/preferredsupplementarycolumnwidth according to the docs, this prop accepts floats so I believe that we should be good

ios/gamma/split-view/RNSSplitViewAppearanceCoordinator.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3043/commits/8dc48c8217c2c062b10e4088441e6e1c4c76b521

ios/gamma/split-view/RNSSplitViewAppearanceCoordinator.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3043/commits/8b9546122bf0cd8f1321794135f9fdb982945515

ios/gamma/split-view/RNSSplitViewAppearanceCoordinator.mm
t0maboro: https://developer.apple.com/documentation/uikit/uisplitviewcontroller/preferredsupplementarycolumnwidth according to the docs, this prop accepts floats so I believe that we should be good

PR #3045, opened 2025-07-21
fix(iOS, Gestures): Ignore inactive preloaded screens in gestureRecognizer

kmichalikk:
## Description

Fixes https://github.com/react-navigation/react-navigation/issues/12683

Currently, `gestureRecognizer:shouldReceivePressOrTouchEvent` decides whether it wants to handle touch events by checking properties of the last screen in subviews list. It doesn't check if the last screens is actually the one that is presented. The bug in https://github.com/react-navigation/react-navigation/issues/12683 came from the fact, that preloaded screen was added to the end of the list and was taken as the `topScreen`.

## Changes

To find the `topScreen`, the subview list is iterated backwards to find the first screen that has state=OnTop. When new screens are preloaded to the end of the list, they have different state, and are ignored.

Use the provided Test.

### DIFF
--- a/apps/src/tests/Test3045.tsx
+++ b/apps/src/tests/Test3045.tsx
@@ -1,8 +1,7 @@
 import React from 'react';
 import { NavigationContainer, ParamListBase } from '@react-navigation/native';
 import { NativeStackNavigationProp, createNativeStackNavigator } from '@react-navigation/native-stack';
-import { Button, ScrollView } from 'react-native';
-import { Text } from 'react-native-gesture-handler';
+import { Button, ScrollView, Text } from 'react-native';
 
 type StackRouteParamList = {
   First: undefined;
@@ -16,7 +15,7 @@ type NavigationProp<ParamList extends ParamListBase> = {
 
 type StackNavigationProp = NavigationProp<StackRouteParamList>;
 
-const First = ({navigation}: StackNavigationProp) => {
+function First({navigation}: StackNavigationProp) {
   return <ScrollView contentInsetAdjustmentBehavior='automatic'>
     <Button onPress={() => navigation.navigate("Second")} title='Second'/>
     <Button onPress={() => navigation.preload("Second")} title='Preload Second'/>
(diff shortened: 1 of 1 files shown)


apps/src/tests/index.ts
```diff
@@ -141,6 +141,7 @@ export { default as Test2899 } from './Test2899';
 export { default as Test2926 } from './Test2926'; // [E2E created](iOS): PR related to iOS search bar
 export { default as Test2949 } from './Test2949'; // [E2E skipped]: can't check system bars styles
 export { default as Test2963 } from './Test2963'; // [E2E created](iOS): issue related to iOS
+export { default as TestPR3045 } from './TestPR3045';
```
kligarski:
```suggestion
export { default as Test3045 } from './Test3045';
```
To follow the convention, I'd just use the number here + in filename.

kmichalikk: I wanted to differentiate issue number vs PR number; we are 4 issues away from 3045 where another Test could be created, and this is PR3045

We don't have issue for this case, the issue is in react-navigation repo

apps/src/tests/TestPR3045.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import { NavigationContainer, ParamListBase } from '@react-navigation/native';
+import { NativeStackNavigationProp, createNativeStackNavigator } from '@react-navigation/native-stack';
+import { Button, ScrollView } from 'react-native';
+import { Text } from 'react-native-gesture-handler';
```
kkafar: I don't think this should be imported from `gesture-handler`. I think you want to import this from `react-native`.

apps/src/tests/TestPR3045.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+type StackNavigationProp = NavigationProp<StackRouteParamList>;
+
+const First = ({navigation}: StackNavigationProp) => {
+  return <ScrollView contentInsetAdjustmentBehavior='automatic'>
+    <Button onPress={() => navigation.navigate("Second")} title='Second'/>
+    <Button onPress={() => navigation.preload("Second")} title='Preload Second'/>
+    <Button onPress={() => navigation.preload("Third")} title='Preload Third'/>
+  </ScrollView>
+}
```
kkafar: Please use a (regular) function here. This is my preference, but I try to enforce it across the repo, since regular functions take part in symbol name hoisting, while arrow functions do not.

kkafar (approved):
Looks good, thanks.

PR #3046, opened 2025-07-21
chore(CI): Remove clang-format

kmichalikk:
## Description

This PR removes outdated clang-format npm package to make git hooks and formatting commands use native clang-format. The npm package is outdated and doesn't work on Apple Silicon Macs.

Should we update the pipeline, and/or automate installation of some alternatives?

Any comments are welcome.


kkafar (approved):
Great you've noticed this. 

Let's go.

PR #3047, opened 2025-07-21
fix(iOS, SplitView): Update showSecondaryToggleButton dynamically

t0maboro:
## Description

This PR adds an option to update `showSecondaryToggleButton` prop dynamically. As for now, it requires some layout update to trigger refresh of the navbar to which this button is attached. After some investigation, we found that we can forcefully update that navbar by hiding it and opening again. With an assumption that it's done in a single transaction, no flicker/lag is observed.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/285 .

## Changes

- Added native flag for signaling that secondary screen navbar update is needed.
- Added logic for triggering updates for secondary navbar.

## Screenshots / GIFs

https://github.com/user-attachments/assets/998a1189-368b-4e42-8dae-7930b02e7874

Updated SplitViewBaseApp

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/apps/src/tests/TestSplitView/SplitViewBaseApp.tsx
+++ b/apps/src/tests/TestSplitView/SplitViewBaseApp.tsx
@@ -1,7 +1,18 @@
 import React from 'react';
-import { Button, StyleSheet, Text, View } from 'react-native';
+import { StyleSheet, Text, View } from 'react-native';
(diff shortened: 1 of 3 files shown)


kkafar (approved):
Looks good. Thank you!

PR #3048, opened 2025-07-21
chore(iOS, Tabs): add prop docs

kligarski:
## Description

Adds docs for props in `BottomTabs` and `BottomTabsScreen` for iOS.

Docs for Android tabs need to be handled in separate PR.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/303.

## Changes

- prop docs for `BottomTabs` and `BottomTabsScreen`

N/A


src/components/BottomTabs.tsx
```diff
@@ -39,12 +45,42 @@ export interface BottomTabsProps extends ViewProps {
   tabBarItemIconColorActive?: ColorValue;
 
   // iOS-only appearance
+
```
kmichalikk: We could remove this comment because the doc includes @platform

src/components/BottomTabs.tsx
```diff
@@ -39,12 +45,42 @@ export interface BottomTabsProps extends ViewProps {
...
+   *
+   * @platform ios
+   */
   tabBarTintColor?: ColorValue;
 
   // Control
 
   // Experimental support
-  experimentalControlNavigationStateInJS?: boolean; // defaults to `false`
+
```
kmichalikk: Same as above, the doc already includes 'Experimental'

src/components/BottomTabsScreen.tsx
```diff
@@ -44,16 +44,52 @@ export type Icon = SFIcon | ImageIcon | TemplateIcon;
...
+   * Title of the tab screen, displayed in the tab bar.
+   *
+   * @platform android, ios
+   */
   title?: string;
+
+  /**
+   * Specifies content of tab bar item badge.
+   *
+   * TODO: Android
```
kmichalikk: Is there maybe a @todo keyword? and should this be included in the doc anyway?

src/components/BottomTabsScreen.tsx
```diff
@@ -63,39 +99,217 @@ export interface BottomTabsScreenProps {
...
 
-  overrideScrollViewContentInsetAdjustmentBehavior?: boolean; // defaults to true
+  /**
+   * Specifies if `contentInsetAdjustmentBehavior` of ScrollViews in first
+   * descendant chain from tab screen should be overriden back from `never`
+   * to `automatic`.
+   *
+   * By default, `react-native`'s ScrollView has `contentInsetAdjustmentBehavior`
+   * set to `never` instead of UIKit-default (which is `automatic`). This
+   * prevents ScrollViews from respecting navigation bar instets.
```
kmichalikk: typo: insets

src/components/BottomTabs.tsx
kligarski: I'd leave it for now because we are still adding new props for both platforms and it helps in deciding where to put them.

src/components/BottomTabs.tsx
kligarski: I'd leave it as it marks start of experimental section that includes experimental props.

src/components/BottomTabsScreen.tsx
kligarski: Good idea with `@todo`.

> and should this be included in the doc anyway?

This is just a placeholder, docs for Android-specific props should be created in separate PR. This prop is shared but has different behavior depending on the platform.

src/components/BottomTabsScreen.tsx
kligarski: https://github.com/software-mansion/react-native-screens/pull/3048/commits/01435dcde8b6bdc5f0780e192b73b4e8cd1ff747

kligarski:
> Also, I've seen @brief being used in our docstrings for the first sentence summarizing the subject and here you already have it structured that way so maybe it's worth adding it?

I've seen it only in internal native code docs, not on props. We can consider it though.

src/components/BottomTabs.tsx
```diff
@@ -44,13 +50,53 @@ export interface BottomTabsProps extends ViewProps {
...
   tabBarTintColor?: ColorValue;
+
+  /**
+   * Specifies minimize behavior for the tab bar. Available since
+   * iOS 26.
+   *
+   * @default Defaults to `automatic`.
+   *
+   * @platform ios
+   * @supported iOS 26 or higher
```
t0maboro: maybe we should consider using `@since`? https://jsdoc.app/tags-since

src/components/BottomTabs.tsx
```diff
@@ -44,13 +50,53 @@ export interface BottomTabsProps extends ViewProps {
...
 
   // iOS-only appearance
+
+  /**
+   * The color used for selected tab's text and icon color.
+   * Starting from iOS 26, it also impacts glow of Liquid Glass tab
+   * selection view.
+   *
+   * tabBarItemTitleFontColor and tabBarItemIconColor defined on
+   * BottomTabsScreen component override this color.
```
t0maboro: you can wrap `tabBarItemTitleFontColor` and `tabBarItemIconColor`

src/components/BottomTabsScreen.tsx
```diff
@@ -44,16 +44,52 @@ export type Icon = SFIcon | ImageIcon | TemplateIcon;
...
+   * Title of the tab screen, displayed in the tab bar.
+   *
+   * @platform android, ios
+   */
   title?: string;
+
+  /**
+   * Specifies content of tab bar item badge.
+   *
+   * @todo Describe prop behavior on Android.
```
t0maboro: idk if this was left intentionally, pointing out to ensure

src/components/BottomTabsScreen.tsx
```diff
@@ -64,39 +100,217 @@ export interface BottomTabsScreenProps {
...
+   */
   tabBarBackgroundColor?: ColorValue;
-  tabBarBlurEffect?: BlurEffect; // defaults to 'none'
 
+  /**
+   * Specifies blur effect applied to tab bar when tab screen is selected.
+   * Works with backgroundColor's alpha < 1.
+   *
+   * Can be:
+   * - one of styles mapped from UIKit's UIBlurEffectStyle, e.g. `systemUltraThinMaterial`,
```
t0maboro: nit: listing all might be problematic, but if the intention is to put this in the final doc, we should consider to be as transparent as possible here

src/components/BottomTabsScreen.tsx
```diff
@@ -64,39 +100,217 @@ export interface BottomTabsScreenProps {
...
+   * @platform ios
+   */
   tabBarItemTitleFontColor?: TextStyle['color'];
+
+  /**
+   * Specifies title offset of every tab item in the tab bar
+   * when tab screen is selected.
+   *
+   * Depending on iOS version, interface orientation,
+   * this setting might impact text, badge and icon.
```
t0maboro:
```suggestion
   * Depending on the iOS version and the device's interface orientation,
   * this setting may affect the appearance of the text, badge and icon.
```

src/components/BottomTabsScreen.tsx
```diff
@@ -64,39 +100,217 @@ export interface BottomTabsScreenProps {
...
+   * @platform ios
+   */
   selectedIcon?: Icon;
 
+  /**
+   * Specifies color of the icons for every tab item in the tab bar
+   * when tab screen is selected. Impacts also title text color.
+   *
+   * On iOS 26, it only applies to selected tab bar item. Other items
+   * are dark or light depending on the theme of the tab bar.
```
t0maboro:
```suggestion
   * adopt a dark or light appearance depending on the theme of the tab bar.
```

t0maboro (commented):
Nice work overall! I think we should sync to discuss and align the documentation between our PRs and maybe standardize that process.

src/components/BottomTabs.tsx
kligarski: Discussed internally - decided to continue using `@supported`.

src/components/BottomTabs.tsx
kligarski: https://github.com/software-mansion/react-native-screens/pull/3048/commits/a942e78719fbe06c5a32be0f05513bea6a155381

src/components/bottom-tabs/BottomTabs.types.ts
```diff
@@ -0,0 +1,134 @@
...
+  TextStyle,
+  NativeSyntheticEvent,
+  ViewProps,
+} from 'react-native';
+
+export type NativeFocusChangeEvent = {
+  tabKey: string;
+};
+
+export type BlurEffect =
```
maciekstosio: Are we sure we want those duplicated? Why can't we import it?

src/components/BottomTabsScreen.tsx
kligarski: yup, I left it for @maciekstosio to describe in future PR.

src/components/BottomTabsScreen.tsx
kligarski: https://github.com/software-mansion/react-native-screens/pull/3048/commits/a942e78719fbe06c5a32be0f05513bea6a155381

src/components/BottomTabsScreen.tsx
kligarski: https://github.com/software-mansion/react-native-screens/pull/3048/commits/a942e78719fbe06c5a32be0f05513bea6a155381

src/components/BottomTabsScreen.tsx
kligarski: https://github.com/software-mansion/react-native-screens/pull/3048/commits/a942e78719fbe06c5a32be0f05513bea6a155381

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -0,0 +1,318 @@
...
+
+// iOS-specific: image as a template usage
+export interface TemplateIcon {
+  templateSource: ImageSourcePropType;
+}
+
+// iOS-specific: SFSymbol, image as a template usage
+export type Icon = SFIcon | ImageIcon | TemplateIcon;
+
+export type BlurEffect =
```
maciekstosio: Are we sure we want those duplicated? Why can't we import it?

src/components/bottom-tabs/BottomTabs.types.ts
kligarski: There are 4 reasons I can think of:
1. Codegen requires type definition in spec file, you can't import it from elsewhere.
2. It follows current convention (e.g. `StackAnimationTypes` in `src/types.tsx`, `StackAnimation` in `src/fabric/ScreenNativeComponent.ts`),
3. AFAIK, spec file is used by Fabric, so using types from Fabric spec also on Paper seems odd (as long as we support both architectures),
4. Separation of TS types from spec files seems reasonable - there might be some cases where we want to use different interface for the JS and native side. There are also codegen-related types (e.g. in `LifecycleStateChangeEvent` we use `Int32`) which we might want to avoid in JS interface (even though in this example `Int32` is an alias for `number`). We might duplicate some types and re-use others but this might be more error-prone than having all of them duplicated.

Maybe we can add a ticket to consider this and change it in all of our components if we think it's better.

src/components/bottom-tabs/BottomTabs.types.ts
```diff
@@ -0,0 +1,229 @@
...
+   * @default Defaults to `automatic`.
+   *
+   * @platform ios
+   * @supported iOS 26 or higher
+   */
+  tabBarMinimizeBehavior?: TabBarMinimizeBehavior;
+  // #endregion
+
+  // #region Control
+  // #endregion
```
t0maboro: I'd skip that until we need to put sth in that region

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -0,0 +1,321 @@
+import type {
+  ColorValue,
+  ImageSourcePropType,
+  NativeSyntheticEvent,
+  TextStyle,
+  ViewProps,
+} from 'react-native';
+
+export type EmptyObject = Record<string, never>;
```
t0maboro, line 9: do we need to export this type?

t0maboro (approved):
👍

src/components/bottom-tabs/BottomTabs.types.ts
maciekstosio: It would be great, if you could create a ticket for that

src/components/bottom-tabs/BottomTabs.types.ts
```diff
@@ -0,0 +1,229 @@
...
+   * @summary Specifies background color of the entire tab bar.
+   *
+   * Since iOS 26, it does not affect the tab bar.
+   *
+   * @platform android, ios
+   * @supported iOS 18 or lower
+   */
+  tabBarBackgroundColor?: ColorValue;
+  /**
+   * @summary Specifies title font family of every tab item in the tab bar.
```
maciekstosio: I wonder if title doesn't sound a bit misleading. What do you think about label?

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -0,0 +1,321 @@
...
+export type BottomTabsScreenEventHandler<T> = (
+  event: NativeSyntheticEvent<T>,
+) => void;
+
+export type LifecycleStateChangeEvent = Readonly<{
+  previousState: number;
+  newState: number;
+}>;
+
+// iOS-specific: SFSymbol usage
```
maciekstosio, line 20: Maybe put it in region instead?

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -0,0 +1,321 @@
...
+   * @platform android, ios
+   */
+  isFocused?: boolean;
+  /**
+   * @summary Identifies screen, e.g. when receiving onNativeFocusChange event.
+   *
+   * @platform android, ios
+   */
+  tabKey: string;
+  // #endregion
```
maciekstosio, line 94: I wonder if it wouldn't be more readable if you write name of the section that ends i.e. #endregion Control

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -0,0 +1,321 @@
...
+   * @supported iOS 18 or lower
+   */
+  tabBarBlurEffect?: BottomTabsScreenBlurEffect;
+  /**
+   * @summary Specifies title font family of every tab item in the tab bar
+   * when tab screen is selected.
+   *
+   * @platform ios
+   */
+  tabBarItemTitleFontFamily?: TextStyle['fontFamily'];
```
maciekstosio, line 157: Actually, those are the same as `BottomTabs` maybe we could share them. I can see that you specified that it happens when it's selected, but maybe This is something that suppose to be in docs or guide not the types?

src/components/bottom-tabs/BottomTabs.types.ts
kligarski: https://github.com/software-mansion/react-native-screens-labs/issues/322

src/components/bottom-tabs/BottomTabsScreen.types.ts
kligarski: it is used in `BottomTabsScreen.tsx` component

src/components/bottom-tabs/BottomTabs.types.ts
kligarski: This is how it's called in [UIKit docs](https://developer.apple.com/documentation/uikit/uitabbaritemstateappearance/titletextattributes?language=objc) as well. Maybe we can specify that this applies to the text of the title, like in the linked docs?

src/components/bottom-tabs/BottomTabsScreen.types.ts
kligarski: These labels are there just to help us with the planned split to separate components for Android and iOS. I think it would be odd to add a region there just to mark iOS-specific types. If we want to use the regions here, then we should probably mark all types with appropriate region, let me know if you think we should do that.

src/components/bottom-tabs/BottomTabsScreen.types.ts
kligarski: I'm not sure how to proceed here because we are planning to remove tab bar appearance props from host component in the iOS API. I don't know if we should extract them to some shared interface just to delete it shortly after.

I think that short clarification how iOS appearance works in this case is useful, especially if we don't know when full docs are going to be created.

src/components/bottom-tabs/BottomTabsScreen.types.ts
kligarski: https://github.com/software-mansion/react-native-screens/pull/3048/commits/fbad3ef29aa84d612109daff043fc49da5ea9726

src/components/bottom-tabs/BottomTabs.types.ts
kligarski: https://github.com/software-mansion/react-native-screens/pull/3048/commits/fbad3ef29aa84d612109daff043fc49da5ea9726

src/components/bottom-tabs/BottomTabsScreen.types.ts
maciekstosio: Let's leave it like that then.

src/components/bottom-tabs/BottomTabsScreen.types.ts
maciekstosio: Let's get back to this once we decide if we're removing it

src/components/bottom-tabs/BottomTabs.types.ts
maciekstosio: Yeah I know, but on first read it confused me, it sounded like there is some top bar that will have a title. It's small thing, leaving it up to you.

maciekstosio (approved):

src/components/bottom-tabs/BottomTabsScreen.types.ts
t0maboro: after verbal discussion, we decided to keep it here at the moment, but if it will be necessary, we're planning to extract all 'repeatable' types to some shared source file and then import

t0maboro (approved):

PR #3051, opened 2025-07-21
fix(iOS, SplitView): Add flag for invalidating preferredDisplayMode

t0maboro:
## Description

Closes https://github.com/software-mansion/react-native-screens-labs/issues/295

displayMode chosen by OS might differ from the `preferredDisplayMode` value that we keep on the JS side. This may cause a flicker, when user is updating dynamically any other prop on the Host component, because, as for now, `preferredDisplayMode` would be set once again and will be synchronized with the JS value, forgetting about all native interactions that resulted in changing state.

In this PR, I'm adding some extra logic to prevent that case:
- update `preferredDisplayMode` only when JS prop was changed

## Changes

- Update appearance coordinator and SplitViewHost logic for updating props

### Before

https://github.com/user-attachments/assets/ca5436f6-7742-41c8-9826-26c0984ff026

### After

https://github.com/user-attachments/assets/fd31d5f0-204e-4594-bfd8-c39dddf7bce1

Updated demo app

## Checklist

- [x] Ensured that CI passes


ios/gamma/split-view/RNSSplitViewHostController.swift
```diff
@@ -7,6 +7,14 @@ public class RNSSplitViewHostController: UISplitViewController, ReactMountingTra
   private var needsAppearanceUpdate = false
   private var needsSecondaryScreenNavBarAppearanceUpdate = false
 
+  @objc
+  public var needsDisplayModeUpdate: Bool {
+    get { _needsDisplayModeUpdate }
+    set { _needsDisplayModeUpdate = newValue }
+  }
```
kkafar: Why do you need a custom getter & setter here?

ios/gamma/split-view/RNSSplitViewHostController.swift
```diff
@@ -7,6 +7,14 @@ public class RNSSplitViewHostController: UISplitViewController, ReactMountingTra
...
+
+  private var _needsDisplayModeUpdate = false
```
kkafar, line 15: Why do we need a backing property?

ios/gamma/split-view/RNSSplitViewHostComponentEventEmitter.mm
```diff
@@ -17,6 +18,20 @@ - (BOOL)emitOnCollapse
...
+- (BOOL)emitOnDisplayModeWillChange:(UISplitViewControllerDisplayMode)
+                 currentDisplayMode:(UISplitViewControllerDisplayMode)nextDisplayMode
+{
+  if (_reactEventEmitter != nullptr) {
+    _reactEventEmitter->onDisplayModeWillChange(
+        {.currentDisplayMode = rnscreens::conversion::UISplitViewControllerDisplayModeToString(currentDisplayMode),
+         .nextDisplayMode = rnscreens::conversion::UISplitViewControllerDisplayModeToString(nextDisplayMode)});
+    return YES;
+  } else {
+    RCTLogWarn(@"[RNScreens] Skipped OnExpand event emission due to nullish emitter");
```
kkafar:
```suggestion
    RCTLogWarn(@"[RNScreens] Skipped <event-name> event emission due to nullish emitter");
```

ios/gamma/split-view/RNSSplitViewHostComponentEventEmitter.h
```diff
@@ -17,6 +17,8 @@ NS_ASSUME_NONNULL_BEGIN
 @interface RNSSplitViewHostComponentEventEmitter : NSObject
 
 - (BOOL)emitOnCollapse;
+- (BOOL)emitOnDisplayModeWillChange:(UISplitViewControllerDisplayMode)
+                 currentDisplayMode:(UISplitViewControllerDisplayMode)nextDisplayMode;
```
kkafar: I think there is a missing parameter name here? 

```suggestion
- (BOOL)emitOnDisplayModeWillChangeFrom:(UISplitViewControllerDisplayMode)currentDisplayMode
                 to:(UISplitViewControllerDisplayMode)nextDisplayMode;
```

ios/gamma/split-view/RNSSplitViewAppearanceCoordinator.mm
```diff
@@ -19,12 +19,16 @@ - (void)updateAppearanceOfSplitView:(RNSSplitViewHostComponentView *_Nonnull)spl
...
-  controller.preferredDisplayMode = splitView.preferredDisplayMode;
   controller.presentsWithGesture = splitView.presentsWithGesture;
   controller.primaryEdge = splitView.primaryEdge;
   controller.showsSecondaryOnlyButton = splitView.showSecondaryToggleButton;
 
-  // Step 2 - manipulating columns
+  // Step 2 - props with dedicated flags for updates
+  if ([controller consumeDisplayModeUpdateIfNeeded]) {
+    controller.preferredDisplayMode = splitView.preferredDisplayMode;
+  }
```
kkafar: I dislike the fact, that appearance coordinator needs to call a special method here to manage controller internal state. Can we move this this logic inside controller? Maybe override setter for `preferredDisplayMode` on controller?

ios/gamma/split-view/RNSSplitViewHostController.swift
```diff
@@ -60,6 +68,18 @@ public class RNSSplitViewHostController: UISplitViewController, ReactMountingTra
...
+  @objc
+  public func setNeedsDisplayModeUpdate() {
+    needsDisplayModeUpdate = true
+  }
+
+  @objc
+  public func consumeDisplayModeUpdateIfNeeded() -> Bool {
+    guard needsDisplayModeUpdate else { return false }
+    needsDisplayModeUpdate = false
+    return true
```
kkafar: It's functionally the same, but IMO much more streamlined to read:

```suggestion
    if needsDisplayMode {
        needsDisplayMode = false
        return true
    }
    return false
```

IIRC `guard` should serve to ensure that some runtime preconditions are satisfied to run the business logic that follows, while this is a very simple branching logic.

ios/gamma/split-view/RNSSplitViewAppearanceCoordinator.mm
kkafar: Or maybe custom setter would be wrong, as it would not update the value always. We could simply have `controller.updatePreferredDisplayModeIfNeeded()`?

kkafar (changes requested):
I dislike the fact that the fix (modifications to prop update logic) and feature (adding a new event emitter) are mixed up in this PR. Additionally this is not reflected in any way in PR title.

Ideally it would be recommended to have these separated. If you think this is too much hustle to split it, I can let it slip, but please do not repeat this pattern. Obviously update the PR title to reflect that new events are emitted. 


By the way, I've just noticed - why do we have two methods:

`- [RNSSplitViewHostComponentView requestSplitViewSecondaryScreenNavBarForAppearanceUpdate]` and `[RNSSplitViewHostComponentView requestSplitViewHostControllerForAppearanceUpdate]`? 

Why isn't it a single single method?

t0maboro:
> I dislike the fact that the fix (modifications to prop update logic) and feature (adding a new event emitter) are mixed up in this PR. Additionally this is not reflected in any way in PR title.
> 
> Ideally it would be recommended to have these separated. If you think this is too much hustle to split it, I can let it slip, but please do not repeat this pattern. Obviously update the PR title to reflect that new events are emitted.
> 

Not a problem to separate, will do that.

> By the way, I've just noticed - why do we have two methods:
> 
> `- [RNSSplitViewHostComponentView requestSplitViewSecondaryScreenNavBarForAppearanceUpdate]` and `[RNSSplitViewHostComponentView requestSplitViewHostControllerForAppearanceUpdate]`?
> 
> Why isn't it a single single method?

the one with `SecondaryScreenNavBar` is related to the issue with non-updating button here: https://github.com/software-mansion/react-native-screens/pull/3047

ios/gamma/split-view/RNSSplitViewHostComponentEventEmitter.h
t0maboro: event-related changes will be moved to a separate PR; reverting them from here 
https://github.com/software-mansion/react-native-screens/pull/3051/commits/05516add734dac175d26d82f7e5a7edf477910fe

ios/gamma/split-view/RNSSplitViewHostComponentEventEmitter.mm
t0maboro: event-related changes will be moved to a separate PR; reverting them from here
https://github.com/software-mansion/react-native-screens/commit/05516add734dac175d26d82f7e5a7edf477910fe

ios/gamma/split-view/RNSSplitViewHostController.swift
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3051/commits/98b62f00c88368317a05faefb04b9ea4813447c9

ios/gamma/split-view/RNSSplitViewHostController.swift
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3051/commits/98b62f00c88368317a05faefb04b9ea4813447c9

ios/gamma/split-view/RNSSplitViewAppearanceCoordinator.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3051/commits/71a29f3afa60a87b0cd399845a6e317e56b7cb1b

ios/gamma/split-view/RNSSplitViewHostController.swift
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3051/commits/71a29f3afa60a87b0cd399845a6e317e56b7cb1b

ios/gamma/split-view/RNSSplitViewHostComponentEventEmitter.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3064/commits/1ef79eaa20495f0a258590489e80d7bce7abc581

ios/gamma/split-view/RNSSplitViewHostComponentEventEmitter.h
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3064/commits/778ed01a4fa91b80e49ca30153b879d955e61490

kkafar:
> the one with SecondaryScreenNavBar is related to the issue with non-updating button here: https://github.com/software-mansion/react-native-screens/pull/3047

Yeah @t0maboro, but could we merge these methods? I mean is there use case when we do not call them both together? Or is it a problem to call these both together?

t0maboro:
> > the one with SecondaryScreenNavBar is related to the issue with non-updating button here: #3047
> 
> Yeah @t0maboro, but could we merge these methods? I mean is there use case when we do not call them both together? Or is it a problem to call these both together?

SecondaryScreenNavBar is only related to one prop, whereas displayMode shouldn't change the value due to *any other prop change*. Therefore, I think that it needs it's own flag for invalidation.

EDIT: 
After the internal discussion with @kkafar we decided that's okay to merge functions related to the appearance updates to a single method: https://github.com/software-mansion/react-native-screens/pull/3065

PR #3054, opened 2025-07-22
fix(iOS, SplitView): Fix SplitViewHostController initialization

t0maboro:
## Description

Initializing Host controller on `didMoveToWindow` is causing in losing some updates when the appearance coordinator was called earlier from `finalizeUpdates`. Props on the native side were already set, the native state was invalidated, but the appearance coordinator was skipped, because controller wasn't initialized at this point.

## Changes

- Updated a logic to consume appearance updates only after the controller was initialized.

Verified that the current app sets all props properly.

## Checklist

- [x] Ensured that CI passes


kkafar (approved):
Thanks for catching this. Looks good ✅

PR #2945, opened 2025-05-27
fix(Android): closing animation on preloaded screen

maciekstosio:
Fixes #2940

## Description

When user goes to preloaded screen closing animation can be observed. It's because we're changing current top, but going to the screen that's already in the stack. Unfortunately the change of `activityState` is independent as it's a prop. So, when we determine that the screen was already on stack, we don't know that it just changed to active from inactive.

## Changes

This PR filters out inactive screens in stack, so it doesn't keep track of preloaded screens. 

## Screenshots / GIFs

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/90355e43-7bdc-4738-80af-cb336d10a34b" />|<video src="https://github.com/user-attachments/assets/5f5816b2-ad88-4d36-ba2a-01e834ebdee3" /> | 

## Test code and steps to reproduce

See `Test2940.tsx`  

## 
- [x] Included code example that can be used to test this change
- [ ] Ensured that CI passes

PR #3062, opened 2025-07-23
fix(Android): Add `isBeingActivated` flag to differentiate between new and preloaded screens

### LINKED ISSUE #3004: [Android] When presenting preloaded screen in stack, wrong animation is used
### Description

When screen with `activityState=0` is presented, then the animation that is used comes from bottom screen, rather then the new one.

This is most likely caused by this check - https://github.com/software-mansion/react-native-screens/blob/main/android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt#L151

The variable `newTopAlreadyInStack` checks if screen exists in the stack and does not take into consideration the `activityState` that the screen had before the transition.

### Steps to reproduce

1. Preload a screen with animation (e.g. slide_from_bottom) (add to stack with `activityState=0`)
2. Change `activityState` to `2`
3. The animation used to present the screen is incorrect

https://snack.expo.dev/@jtkacz/juicy-red-salsa

### Screens version

4.11.1

### React Native version

0.79.4

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

kmichalikk:
## Description

Closes #3004.

When `onUpdate` is issued for ScreenStack, we have no way of knowing whether the screen that is on top with state=2 was already on top or was preloaded with state=0, so we need to add a flag that gives us back this information. This way, we can get the correct animation from preloaded screen that is being moved on top.

## Changes

New `isBeingActivated` flag added to `Screen` class and set to true when state changes from INACTIVE=0 to ON_TOP=2.

Original repro provided in #3004 in `Test3004a` and a more complicated example using react-navigation in `Test3004b`.


maciekstosio:
I wonder how this solution relates to https://github.com/software-mansion/react-native-screens/pull/2945

kmichalikk:
Looks like both solve the same problem and work with the examples. The flag is less elegant but more local, and the second one might break something else.

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
```diff
@@ -148,7 +149,7 @@ class ScreenStack(
         var shouldUseOpenAnimation = true
         var stackAnimation: StackAnimation? = null
 
-        val newTopAlreadyInStack = stack.contains(newTop)
+        val newTopAlreadyInStack = stack.contains(newTop) && !preloadedWrappers.contains(newTop)
```
kligarski, line 154: Maybe we should add a comment explaining new semantics of `newTopAlreadyInStack`.

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
```diff
@@ -262,6 +263,10 @@ class ScreenStack(
             topScreenWrapper = newTop as? ScreenStackFragmentWrapper
             stack.clear()
             stack.addAll(screenWrappers.asSequence().map { it as ScreenStackFragmentWrapper })
+            preloadedWrappers = screenWrappers
+                .asSequence()
+                .filter { it.screen.activityState == Screen.ActivityState.INACTIVE }
+                .toList()
```
kligarski: I'm not sure here if all screens with inactive state are preloaded. Please add a comment abot this assumption here.

apps/src/tests/Test3004b.tsx
```diff
@@ -0,0 +1,117 @@
...
+}
+
+function Modal({ navigation }: StackNavigationProp) {
+  return (<View style={{
+    width: '100%',
+    height: '100%',
+    flexDirection: 'column',
+    alignItems: 'center',
+    justifyContent: 'center',
+    backgroundColor: 'rgba(0.2, 0.2, 0.4, 0.3)'
```
kligarski: Where possible, you can use colors from `apps/src/shared/styling/Colors.ts`

apps/src/tests/index.ts
```diff
@@ -141,6 +141,8 @@ export { default as Test2899 } from './Test2899';
 export { default as Test2926 } from './Test2926'; // [E2E created](iOS): PR related to iOS search bar
 export { default as Test2949 } from './Test2949'; // [E2E skipped]: can't check system bars styles
 export { default as Test2963 } from './Test2963'; // [E2E created](iOS): issue related to iOS
+export { default as Test3004a } from './Test3004a';
+export { default as Test3004b } from './Test3004b';
```
kligarski: These test names won't work with current sorting logic in example app. You can try using `3004.1` & `3004.2` or use issue number and PR number.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -303,6 +304,7 @@ class Screen(
     }
 
     fun setActivityState(activityState: ActivityState) {
+        Log.d("RNScreens", "Activity state updated to %s".format(activityState.toString()))
```
kligarski: Remember to remove logs and redundant imports. Applies also to `ScreenStack.kt`

apps/App.tsx
```diff
@@ -1,11 +1,11 @@
...
+import Example from './Example';
 import * as Test from './src/tests';
 
 enableFreeze(true);
 
 export default function App() {
-  // return <Example />;
-  return <Test.TestSplitView />;
+  return <Example />;
+  // return <Test.Test3004_1 />;
```
kligarski: Better to leave this file as it was if Example app does not work properly without Reanimated/GestureHandler.

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
```diff
@@ -148,7 +150,8 @@ class ScreenStack(
         var shouldUseOpenAnimation = true
         var stackAnimation: StackAnimation? = null
 
-        val newTopAlreadyInStack = stack.contains(newTop)
+        // We don't count preloaded screen as "already in stack"
```
kligarski: I'd consider adding more details why this is the case. You can also link to this PR if you think it would help.

apps/src/tests/Test3004_2.tsx
```diff
@@ -0,0 +1,118 @@
...
+      <Button onPress={() => navigation.push('Second')} title='Go to Second'/>
+    </View>
+    <Button onPress={() => navigation.replace('Replacement')} title='Replace'/>
+    <Button onPress={() => navigation.pop()} title='Pop'/>
+    <Button onPress={() => navigation.popToTop()} title='Pop all'/>
+  </ScrollView>)
+}
+
+function Modal({ navigation }: StackNavigationProp) {
+  return (<View style={{
```
maciekstosio: Small nitpick, wouldn't it be more readable to move parentheses to separate like in `*_1` example?

t0maboro (approved):

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
```diff
@@ -150,7 +149,8 @@ class ScreenStack(
         var shouldUseOpenAnimation = true
         var stackAnimation: StackAnimation? = null
 
-        // We don't count preloaded screen as "already in stack"
+        // We don't count preloaded screen as "already in stack" up until it appears with state == ON_TOP
```
kligarski:
```suggestion
        // We don't count preloaded screen as "already in stack" until it appears with state == ON_TOP
```

PR #3067, opened 2025-07-24
feat(Example): Fix sorting for multiple tests with same issue number

kmichalikk:
## Description

Currently, we write tests as issue reproductions which we name Test<issue number> - one for each issue, and more general tests/examples which we name as Test<descriptive name>. In order to have the former sorted correctly, we need to stick to the scheme and not have anything non-digit after issue number.

We can change the sorting logic to make it possible to define Test123_1 and Test123_2.

## Changes

Implementation of sort function that understands names like Test123_1

### DIFF
--- a/apps/Example.tsx
+++ b/apps/Example.tsx
@@ -149,12 +149,16 @@ if (isTestSectionEnabled()) {
 const screens = Object.keys(SCREENS);
 const examples = screens.filter(name => SCREENS[name].type === 'example');
 const playgrounds = screens.filter(name => SCREENS[name].type === 'playground');
+const issueRegex = /Test(?<issue>\d+)(?<case>.*)/
 const tests = isTestSectionEnabled()
   ? screens
     .filter(name => SCREENS[name].type === 'test')
     .sort((name1, name2) => {
-      const testNumber1 = Number(name1.substring(4));
-      const testNumber2 = Number(name2.substring(4));
+      const spec1 = issueRegex.exec(name1)?.groups
+      const spec2 = issueRegex.exec(name2)?.groups
+
+      const testNumber1 = parseInt(spec1?.issue as string)
+      const testNumber2 = parseInt(spec2?.issue as string)
 
       if (Number.isNaN(testNumber1) && Number.isNaN(testNumber2)) {
         return 0;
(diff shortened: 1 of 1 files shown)


maciekstosio (approved):

apps/Example.tsx
```diff
@@ -149,21 +149,27 @@ if (isTestSectionEnabled()) {
...
         return 0;
       } else if (Number.isNaN(testNumber1)) {
         return 1;
       } else if (Number.isNaN(testNumber2)) {
         return -1;
-      } else {
+      } else if (testNumber1 !== testNumber2) {
         return testNumber1 - testNumber2;
+      } else {
+        return (spec1?.case as string) < (spec2?.case as string) ? 0 : 1
```
t0maboro: 0 keeps original order indicating that compared values are equal, should be -1 here + additional check for equality

t0maboro (changes requested):

t0maboro (approved):
it looks good, I'm wondering whether `localeCompare` would work here, but even if so, this one is very nitpicky and you can merge in the current form https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

apps/Example.tsx
```diff
@@ -149,21 +149,29 @@ if (isTestSectionEnabled()) {
...
         return 1;
       } else if (Number.isNaN(testNumber2)) {
         return -1;
-      } else {
+      } else if (testNumber1 !== testNumber2) {
         return testNumber1 - testNumber2;
+      } else if ((spec1?.case as string) < (spec2?.case as string)) {
+        return -1;
+      } else {
+        return (spec1?.case as string) > (spec2?.case as string) ? 1 : 0;
```
kligarski, line 174: Did you check what happens if you use e.g. `Test1234_1`, `Test1234_11`, `Test1234_111`, `Test1234_2`? I'm thinking whether we should sort by strings or stick just to our convention of `Test<issue_no>_<case_no>`

kligarski (commented):
Not sure if we really need it right now.

kkafar (commented):
Asking for clarification: why would we ever need two tests for the same issue number? 

If doubling/tripling test cases to cover a PR is necessary, isn't it a good indicator, that the PR should be split into multiple smaller ones?

kmichalikk:
One such case was testing on bare screens API vs react-navigation which I wanted to do but ultimately didn't. Other than that, probably not often. But previous implementation won't work if we put something after PR number so I figured it's worth improving.

PR #3068, opened 2025-07-25
fix: Remove newly added logs from user land

### LINKED ISSUE #3066: Excessive "ScreenSN" logs
### Description

Introduced in [this commit](https://github.com/software-mansion/react-native-screens/commit/896c2c6ef83941afb9d1146d5c7e510623e5fc64) it seems that [logs are printed](https://github.com/software-mansion/react-native-screens/blob/94005e32ba4b96d4650fe536895e64a91a7f2bf2/common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.cpp#L121-L128) on layout. This can happen quite a lot which makes Xcode output pretty noisy:

Image

Current workaround is to downgrade from `4.13.1` to `4.12.0` which does not have the logs.

### Steps to reproduce

Seems to happen whenever I scroll a scroll view

n/a

### Screens version

4.13.1

### React Native version

0.80.1

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

theisgroenbech:
## Description
When upgrading `react-native-screens` to `4.13.x` the app starts to output debug logs from `react-native-screens`:

```
[App.debug.dylib] ScreenView [0] setFrame: {{0, 0}, {0, 0}}
[App.debug.dylib] ScreenView [102] setBounds: {{0, 0}, {440, 956}}
[App.debug.dylib] ScreenView [102] setFrame: {{0, 0}, {440, 956}}
```

They seem to have been introduces in this change: https://github.com/software-mansion/react-native-screens/commit/896c2c6ef83941afb9d1146d5c7e510623e5fc64

This PR fixes #3066 🥳 

## Changes
 - Removed NSLog statements from `iOS/RNSScreen.mm`
 - Removed NSLog statements from `ios/bottom-tabs/*`

It seems to just be NSlog-statements left in the code by mistake. Reply if you need a reproduction app.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -1166,18 +1166,6 @@ RNS_IGNORE_SUPER_CALL_END
 
 #endif // !TARGET_OS_TV && !TARGET_OS_VISION
 
-- (void)setFrame:(CGRect)frame
-{
-  NSLog(@"ScreenView [%ld] setFrame: %@", self.tag, NSStringFromCGRect(frame));
-  [super setFrame:frame];
-}
-
-- (void)setBounds:(CGRect)bounds
-{
-  NSLog(@"ScreenView [%ld] setBounds: %@", self.tag, NSStringFromCGRect(bounds));
-  [super setBounds:bounds];
-}
-
 #pragma mark - RNSScrollViewBehaviorOverriding
 
 - (BOOL)shouldOverrideScrollViewContentInsetAdjustmentBehavior
--- a/common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.cpp
+++ b/common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.cpp
@@ -118,14 +118,6 @@ void RNSScreenShadowNode::appendChild(const ShadowNode::Shared &child) {
 void RNSScreenShadowNode::layout(facebook::react::LayoutContext layoutContext) {
   YogaLayoutableShadowNode::layout(layoutContext);
 
-  std::printf(
-      "ScreenSN [%d] layout {{%.2lf, %.2lf}, {%.2lf, %.2lf}}\n",
-      getTag(),
-      layoutMetrics_.frame.origin.x,
-      layoutMetrics_.frame.origin.y,
-      layoutMetrics_.frame.size.width,
-      layoutMetrics_.frame.size.height);
-
 #ifdef ANDROID
   applyFrameCorrections();
 #endif // ANDROID


ios/RNSScreen.mm
```diff
@@ -1168,13 +1168,11 @@ - (void)updateFormSheetPresentationStyle
...
 {
-  NSLog(@"ScreenView [%ld] setFrame: %@", self.tag, NSStringFromCGRect(frame));
   [super setFrame:frame];
 }
 
 - (void)setBounds:(CGRect)bounds
 {
-  NSLog(@"ScreenView [%ld] setBounds: %@", self.tag, NSStringFromCGRect(bounds));
   [super setBounds:bounds];
 }
```
t0maboro: I'm okay with removing these logs, however, this function doesn't do anything more than calling a superclass method. Can you remove them completely, as now we don't need that overrides?

ios/bottom-tabs/RNSBottomTabsHostComponentView.mm
```diff
@@ -138,8 +138,6 @@ - (void)updateContainer
     [tabControllers addObject:childView.controller];
   }
 
-  NSLog(@"updateContainer: tabControllers: %@", tabControllers);
-
```
t0maboro: Tabs are under active development and as for now we don't want to remove these logs. Additionally, the linked issue is mentioning only `ScreenSN` logs - for which we're okay with removing them: https://github.com/software-mansion/react-native-screens/issues/3066 . Applies to all below regarding BottomTabs component

t0maboro (changes requested):
Thanks for your PR. 
As for now, we're okay with removing logs from Stack, but we'd like to keep logs in BottomTabs for a bit longer. 
Additionally https://github.com/software-mansion/react-native-screens/issues/3066 mentions also this log: https://github.com/software-mansion/react-native-screens/blob/main/common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.cpp#L121-L127
Could you remove it either?

ios/RNSScreen.mm
theisgroenbech: Fixed!

ios/bottom-tabs/RNSBottomTabsHostComponentView.mm
theisgroenbech: That makes sense! I have removed the commit

theisgroenbech:
> Thanks for your PR. As for now, we're okay with removing logs from Stack, but we'd like to keep logs in BottomTabs for a bit longer. Additionally #3066 mentions also this log: https://github.com/software-mansion/react-native-screens/blob/main/common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.cpp#L121-L127 Could you remove it either?

Hi t0maboro

Thank you for your work!

Makes sense keeping the bottom tabs logs

I have removed the `RNSScreenShadowNode.cpp` log as well 🥳

PR #3071, opened 2025-07-25
chore(Android, Tabs): extract appearance to separate classes

maciekstosio:
## Changes

I extracted code related to actual styles update to two new classes `TabHostAppearanceCoordinator` and `TabHostAppearanceApplicator`. The first one is responsible for making updates to separate values in specific order, and the other for just applying styles. 

See `TestBottomTabs/index.tsx` styles should be still applied properly. 

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHostAppearanceApplicator.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHostAppearanceApplicator.kt
@@ -26,10 +26,11 @@ class TabsHostAppearanceApplicator(
                 ?: context.getColor(com.google.android.material.R.color.m3_sys_color_light_surface_container),
         )
 
-        val states = arrayOf(
-            intArrayOf(-android.R.attr.state_checked),
-            intArrayOf(android.R.attr.state_checked)
-        )
+        val states =
+            arrayOf(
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHostAppearanceApplicator.kt
```diff
@@ -0,0 +1,184 @@
...
+            // It can be any other int -> https://reactnative.dev/docs/text-style-props#fontweight
+            // Default is 400 -> https://github.com/facebook/react-native/blob/e0efd3eb5b637bd00fb7528ab4d129f6b3e13d03/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/common/assets/ReactFontManager.kt#L117
+            val fontWeight =
+                if (tabsHost.tabBarItemTitleFontWeight ==
+                    "bold"
+                ) {
+                    700
+                } else {
+                    tabsHost.tabBarItemTitleFontWeight?.toIntOrNull() ?: 400
+                }
```
kmichalikk, line 105: This formatting looks weird but if this is what the formatter wants then who am I to judge

t0maboro (approved):

PR #3073, opened 2025-07-25
fix(iOS, SplitView): Update the logic for syncing ShadowNode state on transition

t0maboro:
## Description

This was found during investigating the bug with native stack header. SplitView sets only 2 frames when the transition like pressing the button triggers the animation: current frame and right after the animation **begins** the frame with the final value of the animation. Updating the state to the final size so early is causing a flicker. In this PR, I'm trying to mitigate this issue as much as possible, by tracking with DisplayLink the presentation frame and blocking `updateLayoutMetrics` until the animation ends.

## Changes

- Added some optimizations to the layout synchronization between native and shadow tree.

### Before

https://github.com/user-attachments/assets/153f0ebb-3adf-4ae1-b6a5-8e78b64096bf

### After

https://github.com/user-attachments/assets/57628740-f22e-4214-bf78-b358f0b1a0f6

set in `apps/src/tests/TestSplitView/index.tsx`
```js
  const splitViewBaseConfig: SplitViewBaseConfig = {
    preferredDisplayMode: 'twoBesideSecondary',
    preferredSplitBehavior: 'tile',
  }
```

override example with

```js
import React from 'react';
import { SplitViewHost, SplitViewScreen } from 'react-native-screens';
import { NativeStackNavigatorComponent } from '../helpers';
import { SplitViewBaseConfig } from '../helpers/types';
import { Text, View } from 'react-native';

export const SplitViewWithNativeStackBase = ({ splitViewBaseConfig }: { splitViewBaseConfig: SplitViewBaseConfig }) => {
  return (
    <SplitViewHost {...splitViewBaseConfig}>
      <SplitViewScreen.Column>
        <View style={{flex: 1, backgroundColor: 'red'}}>
          <Text style={{fontSize: 40, alignSelf: 'center', justifyContent: 'center'}}>Primary</Text>
        </View>
      </SplitViewScreen.Column>
      <SplitViewScreen.Column>
        <View style={{flex: 1, backgroundColor: 'green'}}>
          <Text style={{fontSize: 40, alignSelf: 'center', justifyContent: 'center'}}>Supplementary</Text>
        </View>
      </SplitViewScreen.Column>
      <SplitViewScreen.Column>
        <View style={{flex: 1, backgroundColor: 'blue'}}>
          <Text style={{fontSize: 40, alignSelf: 'center', justifyContent: 'center'}}>Secondary</Text>
        </View>
      </SplitViewScreen.Column>
    </SplitViewHost>
  );
}

```

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/gamma/split-view/RNSSplitViewScreenController.swift
+++ b/ios/gamma/split-view/RNSSplitViewScreenController.swift
@@ -169,7 +169,7 @@ public class RNSSplitViewScreenController: UIViewController {
       "[RNScreens] Expected to find RNSSplitViewHost component for RNSSplitViewScreen component"
     )
 
-    // If the animation is currently running, we prefer to apply dynamic updates,
+    // If the resize animation is currently running, we prefer to apply dynamic updates,
     // based on the results from the presentation layer
     // which is read from `trackTransitionProgress` method.
     if let currentSize = lastAnimationFrame?.size {


ios/gamma/split-view/RNSSplitViewScreenController.swift
```diff
@@ -69,16 +76,62 @@ public class RNSSplitViewScreenController: UIViewController {
 
   // MARK: Layout
 
+  ///
+  /// @brief This method is overridden to extract the value to which we're transitioning and attach the DisplayLink to track frame updates on the presentation layer.
+  ///
+  public override func viewWillTransition(
+    to size: CGSize, with coordinator: any UIViewControllerTransitionCoordinator
+  ) {
```
kmichalikk: nitpick: I'd split the comment into multiple lines so we don't need to scroll horizontally (same below)
Also, shouldn't `with coordinator` be in the separate line? Maybe You didn't run the formatter or is this the proper way in swift?

ios/gamma/split-view/RNSSplitViewScreenController.swift
```diff
@@ -69,16 +76,62 @@ public class RNSSplitViewScreenController: UIViewController {
...
+  ///
+  /// @brief This method is responsible for tracking animation frames and requests layout which will synchronize ShadowNode size with the animation frame size.
+  ///
+  @objc
+  private func checkTransitionProgress() {
+    guard let targetSize = transitioningToSize else { return }
+
+    if let currentFrame = view.layer.presentation()?.frame {
+      let currentSize = currentFrame.size
+      let epsilon: CGFloat = 0.5
```
kmichalikk: nitpick: Consider moving the constant somewhere else, maybe to the top of the class near `private var animationFrame` and name it `viewWidthEpsilon`? Just a thought, feel free to ignore

ios/gamma/split-view/RNSSplitViewScreenController.swift
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3073/commits/a144c2982bef0c6f39ccf371c45a6f4c2b823438

ios/gamma/split-view/RNSSplitViewScreenController.swift
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3073/commits/fe9bfc58f81ca6e65e0ccb0f31e2cba9699b88d7

kmichalikk:
There are still some visual glitches on "blue" view -- please verify

https://github.com/user-attachments/assets/7dbff3e8-daff-42e8-9660-918b003026d7

ios/gamma/split-view/RNSSplitViewHostController.swift
```diff
@@ -82,10 +82,10 @@ public class RNSSplitViewHostController: UISplitViewController, ReactMountingTra
 
   @objc
   public func updatePreferredDisplayModeIfNeeded() {
-      if(_needsDisplayModeUpdate) {
-          _needsDisplayModeUpdate = false
-          preferredDisplayMode = self.splitViewHostComponentView.preferredDisplayMode
-      }
+    if _needsDisplayModeUpdate {
```
WoLewicki: Does it have anything to do with this PR?

ios/gamma/split-view/RNSSplitViewScreenController.swift
```diff
@@ -69,16 +77,64 @@ public class RNSSplitViewScreenController: UIViewController {
...
+      displayLink?.add(to: .main, forMode: .common)
+    }
+  }
+
+  ///
+  /// @brief This method is responsible for tracking animation frames and requests layout
+  /// which will synchronize ShadowNode size with the animation frame size.
+  ///
+  @objc
+  private func checkTransitionProgress() {
```
WoLewicki: So maybe `trackTransitionProgress` is a better name for it?

ios/gamma/split-view/RNSSplitViewScreenController.swift
```diff
@@ -69,16 +77,64 @@ public class RNSSplitViewScreenController: UIViewController {
...
+    guard let targetSize = transitioningToSize else { return }
+
+    if let currentFrame = view.layer.presentation()?.frame {
+      let currentSize = currentFrame.size
+
+      if abs(currentSize.width - targetSize.width) < viewWidthEpsilon
+        && abs(currentSize.height - targetSize.height) < viewWidthEpsilon
```
WoLewicki: Should we have another value for `viewHeightEpsilon` or maybe better change the name to fit both?

ios/gamma/split-view/RNSSplitViewScreenController.swift
```diff
@@ -69,16 +77,64 @@ public class RNSSplitViewScreenController: UIViewController {
...
+      {
+        displayLink?.invalidate()
+        displayLink = nil
+        transitioningToSize = nil
+      }
+      animationFrame = currentFrame
+      view.setNeedsLayout()
```
WoLewicki: Do we really need to go through such pipeline? Maybe we just should trigger the shadow state update directly here? Not sure if messing with native layout system during native transition is a good idea but I may be missing something here.

ios/gamma/split-view/RNSSplitViewScreenController.swift
```diff
@@ -90,10 +146,19 @@ public class RNSSplitViewScreenController: UIViewController {
...
-        ofComponent: splitViewScreenComponentView,
-        inContextOfAncestorView: ancestorView!
-      )
+      if let currentAnimationFrame = animationFrame {
+        let localOrigin = splitViewScreenComponentView.convert(
+          splitViewScreenComponentView.frame.origin, to: ancestorView)
+        let convertedFrame = CGRect(origin: localOrigin, size: currentAnimationFrame.size)
+
+        shadowStateProxy.updateShadowState(withFrame: convertedFrame)
+        animationFrame = nil
```
WoLewicki: Why are you setting it to `nil` here?

WoLewicki (commented):
Left some comments!

ios/gamma/split-view/RNSSplitViewHostController.swift
t0maboro: nope, restored, https://github.com/software-mansion/react-native-screens/pull/3073/commits/b709eb9592d5c01e57f648a9765f9598ae106fea

ios/gamma/split-view/RNSSplitViewScreenController.swift
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3073/commits/460ec541286bb2610a13dc0b3c28e6a27f0aff7b

ios/gamma/split-view/RNSSplitViewScreenController.swift
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3073/commits/25a6b37533b60c66a05c7fb24773d43a6ca7cf94

ios/gamma/split-view/RNSSplitViewScreenController.swift
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3073/commits/3cd1cd1f565c570b568044bcebd69dfa0168f462

ios/gamma/split-view/RNSSplitViewScreenController.swift
t0maboro: Note: I needed to revisit the exit criteria for stopping DisplayLink so this was fixed as a part of:

https://github.com/software-mansion/react-native-screens/pull/3073/commits/bdbd501fd2ab312203f96d6dba41ef5afe0f4f7f

ios/gamma/split-view/RNSSplitViewScreenController.swift
```diff
@@ -69,16 +77,77 @@ public class RNSSplitViewScreenController: UIViewController {
 
   // MARK: Layout
 
+  ///
+  /// @brief This method is overridden to extract the value to which we're transitioning
+  /// and attach the DisplayLink to track frame updates on the presentation layer.
+  ///
+  public override func viewWillTransition(
+    to size: CGSize,
+    with coordinator: any UIViewControllerTransitionCoordinator
```
WoLewicki, line 96: Btw, what about this coordinator? Doesn't he have the same information as I mentioned when we talked about screen transitions? Seems like it to me

t0maboro:
```swift
coordinator.animate(alongsideTransition: { _ in
          DispatchQueue.main.async {
              print("TEST")
          }
 })
```

for this snippet, I was receiving 1 log always (I'm not very familiar with that coordinator, but I assumed that `alongsideTransition` should give me more logs without any extra actions on my end)

WoLewicki: It doesn't work that way here, I think you should rather mimic something close to it: https://github.com/software-mansion/react-native-screens/blob/d3aef2558d19987dedbac70e77a6bbeed91139fa/ios/RNSScreen.mm#L1732-L1741. So still have the displayLink but with correct completion event.

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3073/commits/1cd9f34753845c541b85865db922f196eab68009

t0maboro:
> Looks good but I don't feel competent enough to decide if we should merge this fix, especially if this does not fix the bug completely.

This one should be treated as an improvement, rather than a bugfix for header.

ios/gamma/split-view/RNSSplitViewScreenController.swift
```diff
@@ -69,46 +87,144 @@ public class RNSSplitViewScreenController: UIViewController {
...
-      assert(
-        ancestorView != nil,
-        "[RNScreens] Expected to find RNSSplitViewHost component for RNSSplitViewScreen component"
+    let ancestorView = findSplitViewHostController()?.view
+    assert(
+      ancestorView != nil,
+      "[RNScreens] Expected to find RNSSplitViewHost component for RNSSplitViewScreen component"
+    )
+
+    // If the animation is currently running, we prefer to apply dynamic updates,
```
WoLewicki:
```suggestion
    // If the resize animation is currently running, we prefer to apply dynamic updates,
```

WoLewicki (approved):
LGTM

ios/gamma/split-view/RNSSplitViewScreenComponentView.mm
```diff
@@ -112,6 +112,24 @@ - (nonnull RNSSplitViewScreenComponentEventEmitter *)reactEventEmitter
...
+  // 1. size 'A' is set on ShadowNode
+  // 2. animation for the transition starts
+  // 3. `setFrame` is called with the width 'B'
+  // 4. in the same time, we want to track updates and treat intermediate value A' indicated from the presentation layer
+  // as our source of truth
+  if (![_controller isTransitionInProgress]) {
+    [super updateLayoutMetrics:layoutMetrics oldLayoutMetrics:oldLayoutMetrics];
+  }
+}
+
```
kkafar, line 132: A bit out of topic but is it intentional to accept the frames when the transition is not in progress? Do you recall why do we need these frames?

ios/gamma/split-view/RNSSplitViewScreenController.swift
```diff
@@ -18,8 +18,14 @@ public class RNSSplitViewScreenController: UIViewController {
     return splitViewScreenComponentView.reactEventEmitter()
   }
 
+  private var displayLink: CADisplayLink?
+  private var lastAnimationFrame: CGRect?
+  private var transitionInProgress: Bool
```
kkafar, line 23: It would be nice to declare some inline struct / class and group this state together. I'm unsure whether this is possible in Swift. If not, just create a file-private type to group this.

kkafar: e.g. `SizeTransitionTrackingState` or something.

kkafar, line 23: I'd be great for these names to be more informative, as they are used in very specific context: for size transition of the managed view. If you group these in a dedicated struct, then these names are fine.

ios/gamma/split-view/RNSSplitViewScreenController.swift
```diff
@@ -60,6 +66,18 @@ public class RNSSplitViewScreenController: UIViewController {
...
+  /// @brief Determines whether an SplitView animated transition is currently running
+  ///
+  /// Used to differentiate favor frames from the presentation layer over view's frame .
+  ///
+  /// @return true if the transition is running, false otherwise.
+  ///
+  @objc
+  public func isTransitionInProgress() -> Bool {
+    return transitionInProgress
+  }
```
kkafar, line 79: `isViewSizeTransitionInProgress` & if you have a dedicated `ViewSizeTransitionTrackingState` struct then just `return viewSizeTransitionTrackingState != nil`.

ios/gamma/split-view/RNSSplitViewScreenController.swift
```diff
@@ -69,46 +87,144 @@ public class RNSSplitViewScreenController: UIViewController {
 
   // MARK: Layout
 
+  ///
+  /// @brief This method is overridden to extract the value to which we're transitioning
+  /// and attach the DisplayLink to track frame updates on the presentation layer.
+  ///
+  public override func viewWillTransition(
```
kkafar, line 94: One thing about this method is that, I believe, UIKit also fires it on device / interface orientation changes. Have you tested that we work fine with orientation changes here?

ios/gamma/split-view/RNSSplitViewScreenController.swift
```diff
@@ -69,46 +87,144 @@ public class RNSSplitViewScreenController: UIViewController {
...
+        guard let self = self else { return }
+        self.stopAnimation()
+        // After the animation completion, ensure that ShadowTree state
+        // is calculated relatively to the ancestor's frame by requesting
+        // the state update.
+        self.updateShadowTreeState()
+      })
+  }
+
+  private func stopAnimation() {
```
kkafar, line 121: This method is misnamed. It does not stop any animation. It only cleanups some temporary state & invalidates the display link. The name should reflect that.

kkafar (commented):
I've got few remarks. I'll open a PR with the requested changes in few moments.

ios/gamma/split-view/RNSSplitViewScreenController.swift
```diff
@@ -69,46 +87,144 @@ public class RNSSplitViewScreenController: UIViewController {
...
     }
   }
 
+  private func applyTransitioningShadowState(size: CGSize, ancestorView: UIView) {
+    let localOrigin = splitViewScreenComponentView.convert(
+      splitViewScreenComponentView.frame.origin,
+      to: ancestorView
+    )
+    let convertedFrame = CGRect(origin: localOrigin, size: size)
+    shadowStateProxy.updateShadowState(withFrame: convertedFrame)
```
kkafar, line 200: why not use `- [RNSSplitViewScreenShadowStateProxy updateShadowState:inContextOfAncestorView:]`? Why reimplement the method here?

kkafar (commented):
Another remark.

ios/gamma/split-view/RNSSplitViewScreenController.swift
kkafar: Oh, I see now. It would use the model layer frame, not the presentation layer frame. Makes sense. I'll refactor it a bit though.

ios/gamma/split-view/RNSSplitViewScreenController.swift
kkafar: I've checked & seems that orientation change works nicely.

ios/gamma/split-view/RNSSplitViewScreenController.swift
kkafar: https://github.com/software-mansion/react-native-screens/pull/3176

kkafar: https://github.com/software-mansion/react-native-screens/pull/3176

ios/gamma/split-view/RNSSplitViewScreenController.swift
kkafar: https://github.com/software-mansion/react-native-screens/pull/3176

ios/gamma/split-view/RNSSplitViewScreenController.swift
kkafar: https://github.com/software-mansion/react-native-screens/pull/3176

PR #3076, opened 2025-07-28
feat(iOS, Tabs): Expose RNSTabBarController

t0maboro:
## Description

Exposing `UITabBarController`.

Note: this change means that we need further changes described here: https://github.com/software-mansion/react-native-screens-labs/issues/156

## Changes

- Exposing `UITabBarController`.

Tested that controller is publicly accessible.

## Checklist

- [ ] Ensured that CI passes


t0maboro:
linter is passing locally, merging

t0maboro:
> Let's add a link to reference cycle issue somewhere in this PR and I think that we can proceed.

@kligarski I added it in the description

PR #3077, opened 2025-07-28
fix (android, tabs): Use childFragmentManager if tabs are used inside Stack

Ubax:
## Description

When tabs were used inside the Stack, the blank screen was presented. This was because in this case, tabs and stack shared FragmentManager and tabs removed ScreenStackFragment

https://github.com/software-mansion/react-native-screens/blob/main/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt#L466-L477

There is an issue when using Tabs in Stack with header:

```tsx
<ScreenStack style={StyleSheet.absoluteFill}>
      <ScreenStackItem
        // headerConfig={{ hidden: true }}
        screenId="12345"
        activityState={2}
        style={{ flex: 1 }}>
        <BottomTabs
```

moves tabs below the screen

image

## Changes

1. Make Screen extend `FragmentProviding` interface
2. Use `FragmentProviding` as the base to decide wether there is nested Fragment structure

Create Tabs inside Stack, Stack inside Tabs and Stack inside Stack and test if the app work

## Checklist

- [x] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -33,13 +33,15 @@ import com.swmansion.rnscreens.events.HeaderHeightChangeEvent
 import com.swmansion.rnscreens.events.SheetDetentChangedEvent
 import com.swmansion.rnscreens.ext.asScreenStackFragment
 import com.swmansion.rnscreens.ext.parentAsViewGroup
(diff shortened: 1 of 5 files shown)


kkafar:
Thank you! Superseded by #3084

PR #3078, opened 2025-07-28
feat!(iOS, Tabs): support for appearance types, layouts and states

kligarski:
## Description

Adds support for appearance types, layouts and states in iOS BottomTabs.

While working on this PR, I noticed a bug on Paper https://github.com/software-mansion/react-native-screens-labs/issues/326. It is related to this change (now we allow to set scrollEdgeAppearance to `nil`) but I think it should be handled separately.

## Changes

- full support for `standard, scrollEdge` appearance, `stacked, inline, compactInline` layouts, `normal, selected, focused, disabled` states 
- send apperance props via `UnsafeMixed` to avoid codegen issues (nested objects are set to null, no optionals) or having 192 props inside component (if we wanted to flatten the hierarchy)
- make all appearance configurations independent of each other (map 1:1 to native UIKit behavior, without inheritance) - this was done mostly due to some bugs in UIKit/SwiftUI with handling appearance from many appearance configs

https://github.com/user-attachments/assets/26d53af0-341f-4c96-b2cd-1cf622404b02

- remove appearance config from host component

## Test code and steps to reproduce

Run `TestBottomTabs`.

## Checklist

- [x] Included code example that can be used to test this change
- [ ] Ensured that CI passes

PR #3082, opened 2025-07-29
fix(iOS, Tabs): restore default font size after making size prop undefined again

kligarski:
## Description

When receiving prop value from codegen which is undefined on JS side, we receive `CGFloat` of `0.0`, however this is the same as default value in `oldComponentProps.tabBarItemTitleFontSize` therefore the prop does not get updated and remains `nil` (as we're using `NSNumber` to store the value in the component).

When you change the value and then make it undefined again, we also receive `0.0` but this time this is different than value in `oldComponentProps`. Value of `0.0` is stored inside the view property and it is used in appearance coordinator.

I assume that `RCTFont updateFont` method uses some default font size when it receives `0.0` but this value is different than native iOS bottom tabs title font size which results in the bug.

As I don't think that setting `fontSize` prop to `0.0` is a typical use case, I decided to map it to `nil`.

## Changes

- map `0.0` float value to `nil` NSNumber when receiving font size prop

## Screenshots / GIFs

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/5fff6547-116c-4b81-8eb9-f42bbc263138" /> | <video src="https://github.com/user-attachments/assets/432ccfae-7a1e-48da-a509-b54eb26c7895" /> |

Run `TestBottomTabs`, change `tabBarItemTitleFontSize` and then remove the prop. Font size should return to default value.

## Checklist

- [ ] Ensured that CI passes


ios/bottom-tabs/RNSBottomTabsHostComponentView.mm
```diff
@@ -230,7 +230,8 @@ - (void)updateProps:(const facebook::react::Props::Shared &)props
   }
 
   if (newComponentProps.tabBarItemTitleFontSize != oldComponentProps.tabBarItemTitleFontSize) {
-    _tabBarItemTitleFontSize = [NSNumber numberWithFloat:newComponentProps.tabBarItemTitleFontSize];
+    CGFloat fontSize = newComponentProps.tabBarItemTitleFontSize;
+    _tabBarItemTitleFontSize = fontSize != 0.0 ? [NSNumber numberWithFloat:fontSize] : nil;
```
kmichalikk, line 234: Should we be afraid of testing equality on float here? probably not ... ?

kligarski: There are no calculations involved so I don't think so.

t0maboro (approved):

kligarski:
Superseded by https://github.com/software-mansion/react-native-screens/pull/3078.

PR #3083, opened 2025-07-30
fix(iOS, SplitView): Break the retain cycle between component and controller for SplitViewScreen

t0maboro:
## Description

Closes:
https://github.com/software-mansion/react-native-screens-labs/issues/156

Problem:
- Controller has a strong reference to an associated component via `.view` property.
- In our case, Component keeps a strong reference to the controller.
- This creates a retain cycle, which we're not breaking anywhere at the moment.

Solution:
SplitViewHost keeps references to all subviews of type SplitViewScreen all the time (columns might be hidden, but they won't disappear). It means that we can consider Host as an owner of the SplitViewScreens' lifecycles. If the host is about to be unmounted from the view hierarchy, we can perform some additional cleanup in child components to detach the reference to theirs controllers.

## Changes

- Added a method for invalidating a screen component that removes a reference to the controller.

Create an example that includes a SplitView, which may be removed from the hierarchy, add log and observe.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/gamma/split-view/RNSSplitViewHostComponentView.mm
+++ b/ios/gamma/split-view/RNSSplitViewHostComponentView.mm
@@ -112,13 +112,7 @@ static const CGFloat epsilon = 1e-6;
 - (void)willMoveToWindow:(UIWindow *)newWindow
 {
   if (newWindow == nil) {
-    // We assume that split host is removed from view hierarchy **only** when 
-    // whole component is destroyed & therefore we do the necessary cleanup here. 
-    // If at some point that statement does not hold anymore, this cleanup 
-    // should be moved to a different place.
-    for (RNSSplitViewScreenComponentView *subview in _reactSubviews) {
-      [subview invalidate];
-    }
+    [self invalidate];
   }
 }
 
@@ -144,6 +138,17 @@ static const CGFloat epsilon = 1e-6;
       parentView = (UIView *)parentView.reactSuperview;
     }
     return;
+  }
+}
+
+- (void)invalidate
+{
+  // We assume that split host is removed from view hierarchy **only** when
+  // whole component is destroyed & therefore we do the necessary cleanup here.
+  // If at some point that statement does not hold anymore, this cleanup
+  // should be moved to a different place.
+  for (RNSSplitViewScreenComponentView *subview in _reactSubviews) {
+    [subview invalidate];
   }
 }
 


ios/gamma/split-view/RNSSplitViewScreenComponentView.h
```diff
@@ -20,6 +20,13 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong, readonly, nonnull) RNSSplitViewScreenController *controller;
 @property (nonatomic, weak, readwrite, nullable) RNSSplitViewHostComponentView *splitViewHost;
 
+/**
+ * @brief A function responsible for requesting a cleanup in the SplitViewScreen component.
+ *
+ * Should be called before the component when the component is about to be deleted.
```
kkafar:
```suggestion
 * Should be called when the component is about to be deleted.
```

ios/gamma/split-view/RNSSplitViewHostComponentView.mm
```diff
@@ -109,6 +109,15 @@ - (void)setupController
...
 }
 
+- (void)willMoveToWindow:(UIWindow *)newWindow
+{
+  if (newWindow == nil) {
+    for (RNSSplitViewScreenComponentView *subview in _reactSubviews) {
+      [subview invalidate];
+    }
+  }
+}
```
kkafar:
```suggestion
- (void)willMoveToWindow:(UIWindow *)newWindow
{
  if (newWindow == nil) {
    // We assume that split host is removed from view hierarchy **only** when 
    // whole component is destroyed & therefore we do the necessary cleanup here. 
    // If at some point that statement does not hold anymore, this cleanup 
    // should be moved to a different place.
    for (RNSSplitViewScreenComponentView *subview in _reactSubviews) {
      [subview invalidate];
    }
  }
}
```

kkafar (commented):
Added clarification comments. I'd also move the for loop that invalidates the children into separate method, e.g. `invalidate` :D

kkafar (approved):
Thanks!

Let's do something similar for bottom tabs implementation (in separate PR obviously)

t0maboro:
> Thanks!
> 
> Let's do something similar for bottom tabs implementation (in separate PR obviously)

https://github.com/software-mansion/react-native-screens/pull/3085 - created a followup for bottom tabs

PR #3088, opened 2025-07-31
chore(Lint): Run swift lint on macos runner, split jobs between runners, remove formatting step

kmichalikk:
## Description

Currently, GH action for installing swift on ubuntu crashes randomly. In this PR, swift-format is moved to run on macos vm, and other linters are moved to separate jobs on ubuntu runner.

Additionally, `yarn format` is removed because it doesn't return errors, just formats files in runner's filesystem, and it also runs swift-format which won't be available on ubuntu runners.

## Changes

- Lint github workflow
- yarn lint-swift uses xcrun instead of swift-format directly
- `yarn format` is removed


kmichalikk:
There are currently 18 warnings from `eslint`

Should they be converted to errors (by setting --strict flag) and make the pipeline fail?

.github/workflows/lint.yml
```diff
@@ -10,35 +10,56 @@ on:
...
-      
-      - name: Setup Swift
-        uses: swift-actions/setup-swift@v2
-
       - name: Install node dependencies
         run: yarn
-
-      - name: Format code
-        run: |
-          yarn format
```
kligarski, line 36: We don't need format now?

kligarski (approved):
Looks good, I'm just not sure about one thing.

.github/workflows/lint.yml
kmichalikk: `yarn format` runs formatters on currently checked out code. It doesn't really give feedback on what is wrong or what to change, it just makes changes on the runner's filesystem.

kkafar: Okay, that makes sense, however this should be reflected in PR title & description and  ideally be done in separate PR. In few weeks no one will know otherwise why this has been removed & whether it should be restored or not.

kmichalikk: I updated the description. Format should be removed here, because it runs swift-format which won't be available on ubuntu

PR #3091, opened 2025-07-31
fix(Android, Tabs): JS focus change fix

Ubax:
## Description

When JS focus prop changed only the content view updated and the tab bar item stayed the same.

## Before

https://github.com/user-attachments/assets/1712d7e4-cde7-42fc-a428-51cc9db47473

## After

https://github.com/user-attachments/assets/0bd42229-0d8b-4d23-9b6c-7e6cd3ca45f2

## Changes

Changed `invalidateSelectedTab` to `invalidateAll`

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
@@ -281,7 +281,7 @@ class TabsHost(
         isFocused: Boolean,
     ) {
         containerUpdateCoordinator.let {
-            it.invalidateSelectedTab()
+            it.invalidateAll()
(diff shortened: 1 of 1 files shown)


Ubax:
Now I am not sure if this is correct fix. 

Maybe better one will be to use

```kotlin
bottomNavigationView.selectedItemId =
            checkNotNull(getSelectedTabScreenFragmentId()) { "[RNScreens] A single selected tab must be present" }
``` 

inside `updateSelectedTab`

@kkafar what do you think?

kkafar (approved):
This approach is ok, because we'll want to adjust the appearance of the whole navigation bar on tab change anyway, to align behaviour with iOS here.

PR #3093, opened 2025-08-01
fix(iOS): Glitching back navigations on iOS 26

kmichalikk:
## Description, Changes

This PR handles the change in behavior of screen navigation using back button, which, on iOS 26, doesn't wait for one transition to finish and can pop two screens at once - something we are not able to handle correctly. In order to do that, the back button is set to be disabled while transition is happening.

The `rnscreens_findBackButtonWrapperView` method from UINavigationBar extension, which was needed for finding the back button in navigation bar, is fixed for iOS 26. The method has been used to calculate insets for shadow tree and must have been failing for iOS 26 because the button has changed its location and class. This likely fixes another to-be-found bug.

iOS 26 also introduces new `interactiveContentPopGestureRecognizer` which can fire mid transition. Handling the new recognizer is beyond the scope of this PR so it is set to be disabled permanently in a followup PR.

### Before

https://github.com/user-attachments/assets/489b0fd6-2e9a-41cb-8e05-2053de179106

### After

https://github.com/user-attachments/assets/0ba706de-58b4-4373-91fc-7c3db9e4d651

See the provided test case.


apps/src/tests/Test3093.tsx
```diff
@@ -0,0 +1,54 @@
...
+    )
+  }
+}
+
+export default function App() {
+  return (
+    <NavigationContainer>
+      <Stack.Navigator screenOptions={{
+        headerStyle: {
+          backgroundColor: 'purple'
```
kligarski: Please use our color palette from `apps/src/shared/styling/Colors.ts` :D

apps/src/tests/Test3093.tsx
```diff
@@ -0,0 +1,54 @@
...
+type StackNavigationProp = NavigationProp<StackRouteParamList>;
+
+const Stack = createNativeStackNavigator<StackRouteParamList>();
+
+function makeScreen(route: keyof StackRouteParamList, navigateTo?: keyof StackRouteParamList) {
+  return function Screen({ navigation }: StackNavigationProp) {
+    return (
+      <ScrollView contentInsetAdjustmentBehavior='automatic'>
+        <Text>{route}</Text>
+        { navigateTo != null && <Button title={"Go to " + navigateTo} onPress={() => navigation.navigate(navigateTo)} /> }
```
kligarski: extreme nitpick but shouldn't this be `navigateTo !== undefined` (maybe even `!navigateTo`)? I see that this works currently so it's probably not *really* necessary to change it.

ios/RNSScreenStack.mm
```diff
@@ -56,6 +56,39 @@ @interface RNSScreenStackView () <
...
 @implementation RNSNavigationController
 
+#if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && defined(__IPHONE_26_0) && \
+    __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_26_0
+- (void)viewDidLoad
+{
+  [super viewDidLoad];
+  if (@available(iOS 26, *)) {
+    self.interactiveContentPopGestureRecognizer.enabled = false;
+  }
```
kligarski: I'm not sure if we should disable this gesture recognizer completely if we don't know its purpose (unless we do?). Maybe in this PR we should just handle spamming the back button case and in separate PR we can disable this gesture recognizer for the time of transition?

ios/RNSScreenStack.mm
```diff
@@ -56,6 +56,39 @@ @interface RNSScreenStackView () <
...
+}
+#endif
```
kligarski:
```suggestion
#endif // Check for iOS >= 26
```

ios/RNSScreenStack.mm
```diff
@@ -56,6 +56,39 @@ @interface RNSScreenStackView () <
...
+}
+
+- (void)navigationBar:(UINavigationBar *)navigationBar didPopItem:(UINavigationItem *)item
+{
+  UIView *button = [navigationBar rnscreens_findBackButtonWrapperView];
+  if (button != nil) {
+    button.userInteractionEnabled = true;
+  }
+}
+
```
kligarski: I had an idea to use only the setter for the transition coordinator to enable/disable user interaction but after some thought I think that current solution is better - we use navigation bar delegate methods, which is clearer than doing this in seemingly random place like transitionCoordinator setter.

Please test what happens if you use prevent remove from `react-navigation` - we have some quirky logic that handles that and I think we should verify that its behavior does not break by this change. There is a test screen created for this, you can look for `usePreventRemove` in the test screens.

ios/utils/UINavigationBar+RNSUtility.mm
```diff
@@ -4,8 +4,12 @@ @implementation UINavigationBar (RNSUtility)
...
 - (nullable UIView *)rnscreens_findContentView
 {
-  // Class names taken from iOS 17.5, tested on iPhone 15 Pro
-  static Class ContentViewClass = NSClassFromString(@"_UINavigationBarContentView");
+  static Class ContentViewClass;
+  if (@available(iOS 26.0, *)) {
+    ContentViewClass = NSClassFromString(@"UIKit.NavigationBarContentView");
+  } else {
+    ContentViewClass = NSClassFromString(@"_UINavigationBarContentView");
+  }
```
kligarski: We probably should have `#if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) ...` macro here.

ios/utils/UINavigationBar+RNSUtility.mm
```diff
@@ -21,24 +25,35 @@ - (nullable UIView *)rnscreens_findContentView
...
+      return foundView;
+    }
   }
 
   return nil;
 }
 
+- (nullable UIView *)rnscreens_findBackButtonWrapperView
+{
+  UIView *contentView = self.rnscreens_findContentView;
```
kligarski:
```suggestion
  UIView *contentView = [self rnscreens_findContentView]
```
I think that this syntax is clearer, I'm even suprised that dot operator works here, but it's up to you.

ios/utils/UINavigationBar+RNSUtility.mm
```diff
@@ -21,24 +25,35 @@ - (nullable UIView *)rnscreens_findContentView
...
+
+  return [UINavigationBar rnscreens_findNestedBackButtonWrapperInView:contentView];
```
kligarski:
```suggestion
  return [self rnscreens_findNestedBackButtonWrapperInView:contentView];
```

ios/utils/UINavigationBar+RNSUtility.mm
```diff
@@ -21,24 +25,35 @@ - (nullable UIView *)rnscreens_findContentView
...
+  static Class BarButtonViewClass = NSClassFromString(@"_UIButtonBarButton");
+
+  UIView *foundView;
+
+  for (UIView *subview in view.subviews) {
     if ([subview isKindOfClass:BarButtonViewClass]) {
       return subview;
     }
+
+    foundView = [UINavigationBar rnscreens_findNestedBackButtonWrapperInView:subview];
```
kligarski:
```suggestion
    foundView = [self rnscreens_findNestedBackButtonWrapperInView:subview];
```
also, not sure if we want to use recursion for this method

ios/RNSScreenStack.mm
kmichalikk: 1. Looking at it now, I started to dislike it, because shouldXYZ indicate that it's a flag that can be read anytime, but it now has side effects.

2. will look into that

ios/utils/UINavigationBar+RNSUtility.mm
kmichalikk: It worked without it and it's way more readable that way

ios/utils/UINavigationBar+RNSUtility.mm
kmichalikk: I made it a class static method so it needs to be like this

ios/utils/UINavigationBar+RNSUtility.mm
kmichalikk: I think that there is not that much components to go through so it should be okay?

apps/src/tests/Test3093.tsx
```diff
@@ -0,0 +1,54 @@
...
+        <Stack.Screen name="Screen2" component={makeScreen("Screen2", "Screen3")}/>
+      </Stack.Group>
+      <Stack.Group>
+        <Stack.Screen name="Screen3" component={makeScreen("Screen3", "Screen4")}/>
+        <Stack.Screen name="Screen4" component={makeScreen("Screen4")}/>
+      </Stack.Group>
+      </Stack.Navigator>
+    </NavigationContainer>
+  );
+}
```
kkafar:
```suggestion
}

```

ios/RNSScreenStack.mm
kkafar: This should be done definitively in separate PR, with videos before /after just as here, proper issue description with emphasis on interaction with `fullScreenSwipeEnabled` (or whatever that prop is called).

ios/RNSScreenStack.mm
```diff
@@ -56,6 +56,39 @@ @interface RNSScreenStackView () <
...
+  return self.transitionCoordinator == nil;
+}
+
+- (void)navigationBar:(UINavigationBar *)navigationBar didPopItem:(UINavigationItem *)item
+{
+  UIView *button = [navigationBar rnscreens_findBackButtonWrapperView];
+  if (button != nil) {
+    button.userInteractionEnabled = true;
+  }
+}
```
kkafar: Shouldn't this be scoped to iOS 26+? Why do we run this code on earlier versions?

ios/utils/UINavigationBar+RNSUtility.mm
```diff
@@ -4,8 +4,12 @@ @implementation UINavigationBar (RNSUtility)
 
 - (nullable UIView *)rnscreens_findContentView
 {
-  // Class names taken from iOS 17.5, tested on iPhone 15 Pro
-  static Class ContentViewClass = NSClassFromString(@"_UINavigationBarContentView");
+  static Class ContentViewClass;
+  if (@available(iOS 26.0, *)) {
+    ContentViewClass = NSClassFromString(@"UIKit.NavigationBarContentView");
```
kkafar: It seems sus that this is the name of the class, unless it was converted to swift. I'll want to look into that. 

Also initialising a static field this way seems sus. Static field supposed to be initialized once & keep the value between function invocations, this looks like the code will run on every function invocation. Let's verify that please.

ios/utils/UINavigationBar+RNSUtility.mm
kkafar: Since we're not referring to the symbol directly, but rather checking it's existence at runtime, the macro should not be needed here.

ios/RNSScreenStack.mm
kkafar: Ad. 1:

I agree here, however there isn't much better place to do this. One thing: You must make sure that you disable the userInteraction ONLY when you return true. Right now I think the logic might be corrupted.

kkafar: Ad. 2:

yeah, definitively test that.

ios/RNSScreenStack.mm
```diff
@@ -56,6 +56,43 @@ @interface RNSScreenStackView () <
...
+        button.userInteractionEnabled = false;
+      }
+
+      return true;
+    }
+
+    return false;
+  }
+
+  return true;
```
kmichalikk: Without `if (@available)` this doesn't work in iOS 18.5 for `usePreventRemove`, even though there is no breaking change here. This might be due to how we handle preventRemove. In iOS 26 this _does_ work with `preventRemove` though it should break the same way. I suspect this is a timing issue.

With that in mind, the code works on both 18.5 and 26. Should we worry? Should I put some more comments here? If yes, what should I write?

kligarski: I tried to investigate why does this work on iOS 26 but not on iOS 18. The should/didPopItem callbacks are called in the same way but I noticed that on iOS 26 the button instance changes but in iOS 18 it does not.

iOS 18:
18_logs

iOS 26:
26_logs

We really rely on this assumption here but I'm not sure if we can do something about it other than check every new beta release whether it still works.

ios/RNSScreenStack.mm
```diff
@@ -56,6 +56,43 @@ @interface RNSScreenStackView () <
...
+}
+
+- (void)navigationBar:(UINavigationBar *)navigationBar didPopItem:(UINavigationItem *)item
+{
+  if (@available(iOS 26, *)) {
+    UIView *button = [navigationBar rnscreens_findBackButtonWrapperView];
```
kligarski: I'm thinking if there is a case where we get different instance of back button here and leave previous one disabled where we shouldn't. I thought about storing weak reference to the button but I'm not sure.

ios/utils/UINavigationBar+RNSUtility.mm
```diff
@@ -2,10 +2,18 @@
 
 @implementation UINavigationBar (RNSUtility)
 
++ (Class)rnscreens_getBackButtonRuntimeClass
+{
+  if (@available(iOS 26.0, *)) {
+    return NSClassFromString(@"UIKit.NavigationBarContentView");
```
kkafar: Let's add comments at what version of iOS were these names sampled

ios/utils/UINavigationBar+RNSUtility.mm
```diff
@@ -21,24 +29,35 @@ - (nullable UIView *)rnscreens_findContentView
...
-
-  if (contentView == nil) {
+  if (view == nil) {
     return nil;
   }
 
-  for (UIView *subview in contentView.subviews) {
+  static Class BarButtonViewClass = NSClassFromString(@"_UIButtonBarButton");
+
+  UIView *foundView;
```
kkafar:
```suggestion
  UIView *foundView = nil;
```

ios/utils/UINavigationBar+RNSUtility.mm
```diff
@@ -21,24 +29,35 @@ - (nullable UIView *)rnscreens_findContentView
   return nil;
 }
 
-- (nullable UIView *)rnscreens_findBackButtonWrapperView
++ (nullable UIView *)rnscreens_findNestedBackButtonWrapperInView:(nullable UIView *)view
```
kkafar: I prefer instance method. Nit pick.

kkafar (commented):
Few last changes & we're good to go.

kkafar (commented):
I've added review in form of PR: https://github.com/software-mansion/react-native-screens/pull/3124

PR #3097, opened 2025-08-04
fix(iOS, SplitView): Align width of Stack component with SplitView column before transition

t0maboro:
## Description

This PR should be considered rather as a workaround; the proper fix requires synchronous updates. For more context, please see the doc with analysis: https://github.com/software-mansion/react-native-screens-labs/issues/311#issuecomment-3150609759 . 

It introduces a logic that makes an immediate layout update for stack components, for which the delay was resulting in glitchy header animation.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/311 (let me know if you feel that we should create a separate followup ticket for synchronous updates, I have already this one: https://github.com/software-mansion/react-native-screens-labs/issues/310 so I'll leave some info there for now)

## Changes

- Added a helper method for gathering `UINavigationControllers` from the subtree.
- Overriden `layoutSubviews` method for SplitView screen.

## Screenshots / GIFs

### Before

https://github.com/user-attachments/assets/3277eef3-e103-47ce-928a-a4fcedaf0380

### After

https://github.com/user-attachments/assets/910598ea-dcf2-4ccb-a90f-01783a9d54e2

Test `SplitViewWithNestedStack` and `SplitViewWithNativeStackBase` on iOS 26. 

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/gamma/split-view/compat/RNSFrameCorrector.mm
+++ b/ios/gamma/split-view/compat/RNSFrameCorrector.mm
@@ -2,18 +2,20 @@
 
 @implementation RNSFrameCorrector
 
+/**
+ This correction is performed synchronously to resolve the following issue:
+ 1. On the transition begin, the animation may expect the target frame to be set immediately, e.g. in SplitView
+ for triggering the transition animation, we need to know the final frame on animation start.
+ 2. When a layout recalculation is dependent on Yoga - it's performed asynchronously, so the frame for the nested
+ component might be in 'stale' state, e.g. when the transition begins.
+ 3. Until Yoga will compute the new layout, the reference value for the frame passed to the animation will be
+ wrong.
+ 4. As we don't have a control, when the animation will start, we need to apply frame correction synchronously to
+ guarantee that it takes the proper frame width as a reference.
+ */
 + (void)applyFrameCorrectionFor:(UIView *)view
      inContextOfSplitViewColumn:(RNSSplitViewScreenComponentView *)splitViewScreen
 {
-  //    This correction is performed synchronously to resolve the following issue:
-  //    1. On the transition begin, the animation may expect the target frame to be set immediately, e.g. in SplitView
-  //    for triggering the transition animation, we need to know the final frame on animation start.
-  //    2. When a layout recalculation is dependent on Yoga - it's performed asynchronously, so the frame for the nested
-  //    component might be in 'stale' state, e.g. when the transition begins.
-  //    3. Until Yoga will compute the new layout, the reference value for the frame passed to the animation will be
-  //    wrong.
-  //    4. As we don't have a control, when the animation will start, we need to apply frame correction synchronously to
-  //    guarantee that it takes the proper frame width as a reference.
   CGRect originalFrame = view.frame;
   view.frame = CGRectMake(
       originalFrame.origin.x, originalFrame.origin.y, splitViewScreen.frame.size.width, originalFrame.size.height);


ios/RNSStackNavigatorFinder.mm
```diff
@@ -0,0 +1,29 @@
+#import "RNSStackNavigatorFinder.h"
+
+static void SearchNavigationControllers(UIViewController *vc, NSMutableArray<UINavigationController *> *result)
+{
+  for (UIViewController *child in vc.childViewControllers) {
+    if ([child isKindOfClass:[UINavigationController class]]) {
+      UINavigationController *navCtrl = static_cast<UINavigationController *>(child);
+      [result addObject:navCtrl];
+
+      for (UIViewController *subCtrl in navCtrl.viewControllers) {
```
kligarski: I was thinking if we need to traverse every view controller from UINavigationController as maybe those deeper in the stack do not impact the issue but handling this would change semantics of this method, introduce more confusion for a really small benefit in a solution that is meant to be a workaround so I guess that it is okay.

kligarski (approved):
Looks good, great analysis in the doc!

ios/RNSStackNavigatorFinder.mm
t0maboro: The issue was identified for nested stacks either. That's why I'm traversing the entire hierarchy within the subtree.

kkafar: I don't like this traversal approach either. Maybe we could just make nested stacks register with the split view & unregister at the right moments (didMoveToParentViewController). Just create a separate protocol so that the SplitView related symbols are not leaked to the stack implementation. 

I would imagine this solution as follows: 

1. create protocol RNSFrameCorrectionProvider3097 (or something like that, so we know it only servers as a workaround), that has method to register & unregister
2. create RNSFrameCorrector3097 class & implement behaviour that looks for the parent split view and registers appropriate view with it. 
3. use RNSFrameCorrector3097 in did/willMoveToParentViewController. 


What do you think?

ios/RNSStackNavigatorFinder.h
```diff
@@ -0,0 +1,20 @@
+#import <UIKit/UIKit.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface RNSStackNavigatorFinder : NSObject
```
kkafar: We don't have a notion of `navigator` in the screens. I would name it differently, but won't give you suggestions as I suggest other approach in other comments.

kkafar (commented):
I dislike the full DFS scan on every layout. I get it works, but we can progress to the "get it right" stage immediately :)

ios/RNSStackNavigatorFinder.mm
t0maboro: okay, let's do that

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3097/commits/45bd36f44db21001a636dd77b27452cafbe70140

ios/RNSStackNavigatorFinder.h
t0maboro: resolving as no longer valid because of the refactor proposed above
https://github.com/software-mansion/react-native-screens/pull/3097/commits/45bd36f44db21001a636dd77b27452cafbe70140

ios/RNSScreenStack.mm
```diff
@@ -217,6 +230,48 @@ - (void)navigationBar:(UINavigationBar *)navigationBar didPopItem:(UINavigationI
...
 #endif // Check for iOS >= 26
 
+#pragma mark - RNSFrameCorrectionProvider3097
+
+#ifdef RNS_GAMMA_ENABLED
+- (void)registerForSplitView
+{
+  if (auto splitViewScreenController = [self findClosestSplitViewScreenController]) {
+    auto splitViewControllerView = [splitViewScreenController view];
+    if ([splitViewControllerView isKindOfClass:[RNSSplitViewScreenComponentView class]]) {
```
WoLewicki: Is there a possibility that it is not of this class? If we are solely in control of it, I'd either not check it or throw an error in `else`.

ios/RNSScreenStack.mm
```diff
@@ -217,6 +230,48 @@ - (void)navigationBar:(UINavigationBar *)navigationBar didPopItem:(UINavigationI
...
+      auto splitViewScreen = (RNSSplitViewScreenComponentView *)splitViewControllerView;
+      // We need to apply an udpate for the parent of the view which `RNSNavigationController` is describing
```
WoLewicki:
```suggestion
      // We need to apply an update for the parent of the view which `RNSNavigationController` is describing
```

ios/RNSScreenStack.mm
```diff
@@ -217,6 +230,48 @@ - (void)navigationBar:(UINavigationBar *)navigationBar didPopItem:(UINavigationI
...
+  }
+}
+
+- (void)unregisterFromSplitView
+{
+  if (auto splitViewScreenController = [self findClosestSplitViewScreenController]) {
+    auto splitViewControllerView = [splitViewScreenController view];
+    if ([splitViewControllerView isKindOfClass:[RNSSplitViewScreenComponentView class]]) {
+      auto splitViewScreen = (RNSSplitViewScreenComponentView *)splitViewControllerView;
+      // We need to apply an udpate for the parent of the view which `RNSNavigationController` is describing
```
WoLewicki:
```suggestion
      // We need to apply an update for the parent of the view which `RNSNavigationController` is describing
```

Do we want the same comment as above here?

ios/RNSScreenStack.h
```diff
@@ -8,10 +8,20 @@
 #import "RNSBottomTabsSpecialEffectsSupporting.h"
 #import "RNSScreenContainer.h"
 
+#ifdef RNS_GAMMA_ENABLED
+#import "RNSFrameCorrectionProvider3097.h"
```
WoLewicki: I'm really against adding PR number to header name. What if we need a follow-up? xd

ios/gamma/split-view/compat/RNSFrameCorrector3097.mm
```diff
@@ -0,0 +1,13 @@
+#import "RNSFrameCorrector3097.h"
+
+@implementation RNSFrameCorrector3097
+
++ (void)applyFrameCorrectionFor:(UIView *)view
```
WoLewicki: Please add some comment why platform handles setting it as an animation.

WoLewicki (commented):
Left some comments, looks good overall!

ios/RNSScreenStack.mm
t0maboro: very unlikely that the type is wrong, but throwing error sounds safer

ios/gamma/split-view/compat/RNSFrameCorrector3097.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3097/commits/53c8458b169a23bde129da18a4a1d329f7fa5134

ios/RNSScreenStack.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3097/commits/92372253c6cde046f3dd5c9073d6cfe6551e9837

ios/RNSScreenStack.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3097/commits/bfc0e4962b4c4717f08a454308fce7ca2952d41f

ios/RNSScreenStack.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3097/commits/bfc0e4962b4c4717f08a454308fce7ca2952d41f + for consistency, I'd prefer to keep both

ios/RNSScreenStack.h
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3097/commits/c5d000a5109c6f1f3254567aea35da34a668b226

ios/RNSScreenStack.mm
```diff
@@ -34,6 +34,11 @@
 #import "integrations/RNSDismissibleModalProtocol.h"
 #import "utils/UINavigationBar+RNSUtility.h"
 
+#ifdef RNS_GAMMA_ENABLED
+#import "RNSSplitViewScreenComponentView.h"
+#import "Swift-Bridging.h"
```
WoLewicki, line 39: Just one more thing, why do we need this import?

WoLewicki (approved):
LGTM

ios/RNSScreenStack.mm
t0maboro: `RNSSplitViewScreenComponentView.h` has predefined `RNSSplitViewScreenController` symbol which is not visible without `Swift-Bridging` header

ios/RNSScreenStack.mm
```diff
@@ -156,6 +166,9 @@ - (void)didMoveToParentViewController:(UIViewController *)parent
     RNSTabsScreenViewController *parentTabsScreenVC = static_cast<RNSTabsScreenViewController *>(parent);
     [parentTabsScreenVC setTabsSpecialEffectsDelegate:self];
   }
+#ifdef RNS_GAMMA_ENABLED
+  [self registerForSplitView];
+#endif // RNS_GAMMA_ENABLED
```
kkafar, line 171: Why is here no check for `parent` nullability? Why do we try to register even when parent is `nil`?

ios/RNSScreenStack.mm
```diff
@@ -217,6 +230,50 @@ - (void)navigationBar:(UINavigationBar *)navigationBar didPopItem:(UINavigationI
...
+  while (parent) {
+    if ([parent isKindOfClass:[RNSSplitViewScreenController class]]) {
+      auto splitViewScreenController = (RNSSplitViewScreenController *)parent;
+      return splitViewScreenController;
+    }
+    parent = [parent parentViewController];
+  }
+
+  return nil;
+}
```
kkafar, line 274: there is a method for that in UIKit -> https://developer.apple.com/documentation/uikit/uiviewcontroller/splitviewcontroller?language=objc

You don't need to implement this manually, beside doing type checking & casting.

ios/gamma/split-view/compat/RNSFrameCorrectionProvider.h
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * @protocol RNSFrameCorrectionProvider
+ * @brief Temporal solution for registering view for being notified when the split view layout is changed
+ *
+ * @remarks - It's a workaround until we have synchronous updates:
+ * https://github.com/software-mansion/react-native-screens-labs/issues/374
+ *
+ * It resolves problems with jumping content in the SplitView, until we'd have synchronous updates.
+ */
+@protocol RNSFrameCorrectionProvider <NSObject>
```
kkafar, line 14: This ended up being a bit different from what I intended. Currently we are in situation where `RNSFrameCorrectionProvider` protocol is used for `RNSScreenStack` and defines methods that can be called to register the `RNSScreenStack` with the split view. So, well... it provides nothing and therefore is either misnamed or misused 😅 

It should be structured as follows:

1. the split view screen implements `RNSFrameCorrectionProvider` protocol,
2. the `RNSFrameCorrectionProtocol` exposes methods for registering and unregistering views for frame correction. 
3. the view that implements `RNSFrameCorrectionProvider` uses `RNSFrameCorrector` instance to correct the frames of registered views appropriately. 
4. in our case the `RNSScreenStack` should register with frame correction provider, maybe using methods from `RNSFrameCorrector` to find the appropriate split screen.


I'll refactor this in follow-up PR.

ios/RNSScreenStack.mm
kkafar: My bad. You're looking for screen view controller, not split view controller.

ios/RNSScreenStack.mm
kkafar: https://github.com/software-mansion/react-native-screens/pull/3179

ios/RNSScreenStack.mm
kkafar: https://github.com/software-mansion/react-native-screens/pull/3179

ios/gamma/split-view/compat/RNSFrameCorrectionProvider.h
kkafar: https://github.com/software-mansion/react-native-screens/pull/3179

PR #3099, opened 2025-08-06
fix(iOS, Tabs): Updated logic for breaking retain cycles

t0maboro:
## Description

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/340

In this PR, I'm fixing the issue with invalidating the controller on the view component when the view was detached from the window. It wasn't the correct approach, as the view might be reattached with the associated RN component.

In the current approach, we're observing incoming mutations on mounting transactions. If the Host component is about to be removed, it asks all its children to invalidate (remove the reference to the controller), and then it invalidates itself. In this approach, I assume that all Host components will be listening to the mounting transactions and will be responsible for invalidating their children.

Note: There will still be a lot of work to be done. We need to revisit the assumed conditions here, for other components (e.g. for stack) we'd need to detect not only the Host component's `Delete` mutation, but also any `Screen` component's mutation and detach the single screen. What I'd consider is adding `shouldInvalidateOnMutation` to the protocol and treating it as a determinant of what and when should be detached; the logic for detaching might slightly differ between components. I think that it'd be better to prepare a skeleton with a POC for a single component here and then extend other components with this protocol in separate PRs.

## Changes

- Added a protocol and registry responsible for view invalidation
- Updated logic for detecting which components should be invalidated - now we're relying on `RCTMountingTransactionObserving`

Tested with the example from: https://github.com/software-mansion/react-native-screens-labs/issues/340#issuecomment-3158525815

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSViewControllerInvalidator.mm
+++ b/ios/RNSViewControllerInvalidator.mm
@@ -1,6 +1,7 @@
 #ifdef RCT_NEW_ARCH_ENABLED
 
 #import "RNSViewControllerInvalidator.h"
+
 #import <React/RCTAssert.h>
 #import "RNSInvalidatedComponentsRegistry.h"
 #import "RNSViewControllerInvalidating.h"
--- a/ios/bottom-tabs/RNSBottomTabsHostComponentView.mm
+++ b/ios/bottom-tabs/RNSBottomTabsHostComponentView.mm
@@ -10,6 +10,7 @@
 #import <react/renderer/components/rnscreens/RCTComponentViewHelpers.h>
 #import <rnscreens/RNSBottomTabsComponentDescriptor.h>
 #import "RNSBottomTabsHostComponentView+RNSImageLoader.h"
+#import "RNSInvalidatedComponentsRegistry.h"
 #import "RNSViewControllerInvalidator.h"
 #endif // RCT_NEW_ARCH_ENABLED
 
@@ -17,7 +18,6 @@
 #import "RNSConversions.h"
 #import "RNSConvert.h"
 #import "RNSDefines.h"
-#import "RNSInvalidatedComponentsRegistry.h"
 #import "RNSTabBarController.h"
 #import "RNSTabBarControllerDelegate.h"
 
@@ -190,7 +190,7 @@ namespace react = facebook::react;
 }
 #else
 
-#pragma mark - Paper specific
+#pragma mark - RCTInvalidating
 
 - (void)invalidate
 {
--- a/ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
+++ b/ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
@@ -110,9 +110,10 @@ RNS_IGNORE_SUPER_CALL_BEGIN
 }
 RNS_IGNORE_SUPER_CALL_END
 
+#ifdef RCT_NEW_ARCH_ENABLED
+
 #pragma mark - RNSViewControllerInvalidating
 
-#ifdef RCT_NEW_ARCH_ENABLED
 - (void)invalidateController
 {
   _controller = nil;
@@ -120,7 +121,7 @@ RNS_IGNORE_SUPER_CALL_END
 
 #else
 
-#pragma mark - Paper specific
+#pragma mark - RCTInvalidating
 
 - (void)invalidate
 {
--- a/ios/RNSViewControllerInvalidating.h
+++ b/ios/RNSViewControllerInvalidating.h
@@ -5,6 +5,7 @@
 @protocol RNSViewControllerInvalidating
 
 - (void)invalidateController;
+
 - (BOOL)shouldInvalidateOnMutation:(const facebook::react::ShadowViewMutation &)mutation;
 
 @end


ios/RNSInvalidateControllerProtocol.h
```diff
@@ -0,0 +1,5 @@
+#import <Foundation/Foundation.h>
```
kligarski:
```suggestion
```

Not sure if this is necessary? Applies also to RNSInvalidatedComponentsRegistry.

ios/RNSInvalidateControllerProtocol.h
```diff
@@ -0,0 +1,5 @@
...
+
+@protocol RNSInvalidateControllerProtocol <NSObject>
```
kligarski:
```suggestion
@protocol RNSInvalidateControllerProtocol
```
Do we need NSObject here?

kligarski:
```suggestion
@protocol RNSViewControllerInvalidating
```

I think that we're trying to follow Swift naming convention (https://www.swift.org/documentation/api-design-guidelines/#naming).

> Protocols that describe a capability should be named using the suffixes able, ible, or ing (e.g. Equatable, ProgressReporting).

ios/RNSInvalidatedComponentsRegistry.mm
```diff
@@ -0,0 +1,40 @@
...
+
++ (instancetype)invalidatedComponentsRegistry
+{
+  static RNSInvalidatedComponentsRegistry *sharedInstance = nil;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    sharedInstance = [[RNSInvalidatedComponentsRegistry alloc] init];
+  });
+  return sharedInstance;
+}
```
kligarski: Not sure if we want a singleton for this registry - e.g. should invalidation of stack also invalidate tab screens added to registry? This could potentially happen before tab host requests flush, right?

ios/RNSInvalidatedComponentsRegistry.mm
```diff
@@ -0,0 +1,40 @@
+#import "RNSInvalidatedComponentsRegistry.h"
+
+@interface RNSInvalidatedComponentsRegistry ()
+@property (nonatomic, strong) NSMutableSet<UIView<RNSInvalidateControllerProtocol> *> *invalidViews;
```
kligarski: Maybe we can store weak references to `UIView<RNSInvalidateControllerProtocol>`. I guess that somebody will call flush eventually but maybe we should prevent memory leak either way?

ios/bottom-tabs/RNSBottomTabsHostComponentView.mm
```diff
@@ -363,6 +369,29 @@ - (void)mountingTransactionDidMount:(const facebook::react::MountingTransaction
...
+
+  UIView<RNSInvalidateControllerProtocol> *invalidationTarget = (UIView<RNSInvalidateControllerProtocol> *)view;
+
+  if (invalidationTarget.window == nil) {
+    [invalidationTarget invalidateController];
+  } else {
+    [[RNSInvalidatedComponentsRegistry invalidatedComponentsRegistry] pushForInvalidation:invalidationTarget];
+  }
+}
+
```
kligarski: I'm wondering if this logic can be shared between view components?

ios/RNSInvalidatedComponentsRegistry.mm
t0maboro: I wanted to have a single instance that behaves like some kind of "garbage collector" for the whole hierarchy we have. I mean that every Host is responsible for invalidating itself and its children, so when the stack invalidates and bottom tabs are already in the registry waiting for an invalidation, I don't see any problem with invalidating all components at once, as we know that all of them are about being destroyed.

ios/RNSInvalidatedComponentsRegistry.mm
t0maboro: NSMutableSet requires strong refs, we may consider this: https://gist.github.com/chrishulbert/1037232 , let me know if you think that's the better approach

ios/RNSInvalidateControllerProtocol.h
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3099/commits/c0b1bcbbe1d86ecfcdaea9060373045636889d74

ios/RNSInvalidateControllerProtocol.h
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3099/commits/27372c35fbf4c2e41c0c8c3eede41f6ad24171af

ios/bottom-tabs/RNSBottomTabsHostComponentView.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3099/commits/ceb56b1684a28f4d0703976fb489aae653d00f3d

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
```diff
@@ -100,21 +101,20 @@ - (void)resetProps
   _selectedIconSfSymbolName = nil;
 }
 
-- (void)invalidate
```
WoLewicki, line 106: Won't it destroy old arch? Also, if we don't want it, should we also remove conforming to the protocol connected to this method?

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
```diff
@@ -2,6 +2,7 @@
 #import "NSString+RNSUtility.h"
 #import "RNSConversions.h"
 #import "RNSDefines.h"
+#import "RNSInvalidatedComponentsRegistry.h"
```
WoLewicki: Why do we need this import here?

ios/bottom-tabs/RNSBottomTabsHostComponentView.mm
```diff
@@ -110,7 +112,7 @@ - (void)resetProps
 - (void)willMoveToWindow:(UIWindow *)newWindow
 {
   if (newWindow == nil) {
-    [self invalidate];
+    [[RNSInvalidatedComponentsRegistry invalidatedComponentsRegistry] flushInvalidViews];
```
WoLewicki: shouldn't `RNSInvalidatedComponentsRegistry` be a proxy for calling `flushInvalidViews`?

ios/RNSViewControllerInvalidator.mm
```diff
@@ -0,0 +1,24 @@
...
++ (void)invalidateViewIfDetached:(UIView *)view
+{
+  RCTAssert(
+      [view conformsToProtocol:@protocol(RNSViewControllerInvalidating)],
+      @"[RNScreens] View of type: %@ doesn't conform to RNSViewControllerInvalidating",
+      view.class);
+
+  UIView<RNSViewControllerInvalidating> *invalidationTarget = (UIView<RNSViewControllerInvalidating> *)view;
+
+  if (invalidationTarget.window == nil) {
```
WoLewicki: Were there cases when the window was still there? E.g. the mutation removing the view was dispatched and immediately after something else triggered cleanup?

ios/RNSViewControllerInvalidator.mm
```diff
@@ -0,0 +1,24 @@
+#import "RNSViewControllerInvalidator.h"
+#import <React/RCTAssert.h>
+#import "RNSInvalidatedComponentsRegistry.h"
+#import "RNSViewControllerInvalidating.h"
+
+@implementation RNSViewControllerInvalidator
+
++ (void)invalidateViewIfDetached:(UIView *)view
```
WoLewicki: Why don't we just push only views conforming to protocol so we don't have to care about it later?

ios/RNSViewControllerInvalidator.mm
t0maboro: - if you're dismissing the screen using the native back button, the screen might be already dismissed and react doesn't have any information about it, then some event is sent to React and the component is unmounted - the `Delete` mutation is coming
- if you're dismissing the screen from JS, `Delete` comes earlier

we need to have a control over both

ios/bottom-tabs/RNSBottomTabsHostComponentView.mm
t0maboro: I prefer the current version, as the registry has control over all invalid views. Views have `RNSViewControllerInvalidating::invalidateController`, which delegates the work to components conforming to our protocol. I don't see any benefit in having a proxy here, but let me know if you still prefer your version.

ios/RNSViewControllerInvalidator.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3099/commits/4f95dc0b484f35b6f89f608634ae0d2620ce4eaa

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3099/commits/5bf45f4c054bb2f44ace22e3fee792e876d66ae4

ios/bottom-tabs/RNSBottomTabsHostComponentView.mm
t0maboro: (slightly related to this - I resigned from singleton https://github.com/software-mansion/react-native-screens/pull/3099/commits/7a2da4a9c512b9963d76b247d007916168da60d6)

ios/RNSInvalidatedComponentsRegistry.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3099/commits/7a2da4a9c512b9963d76b247d007916168da60d6

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3099/commits/601ee8cb35134d6f89fccc5ae280c81a3a90e427

WoLewicki (approved):
LGTM, just check that we do not have any memleaks with some more advanced navigation actions, as popping the whole navigator etc.

t0maboro:
> LGTM, just check that we do not have any memleaks with some more advanced navigation actions, as popping the whole navigator etc.

I checked the case when we have bottom tabs nested in a stack on the 2nd screen - after dismissing the screen, all tabs and the host received `invalidate` request. - Is this the case that you meant?

WoLewicki:
Yep

hidang:
Hi @t0maboro, 
Sorry for bothering you, but I have tested version **react-native-screens 4.15.4** and the retain cycle dependence issue still persists—RNScreen is still leaking.
Could you please double-check this again in the current version 4.15.4? 
7642567c-6778-4327-93e2-054aec30a523

Thank you.

t0maboro:
Hi @hidang thank you for noticing it. Please keep in mind that this PR applies only to BottomTabs, and this may mean that we'd need to revisit the logic for the stack, either. Do you have any specific repro for which RNScreen is leaking?

PR #3102, opened 2025-08-07
fix(Example): Make Example app build again

kmichalikk:
## Description

For some time, the example app has been crashing due to old and misconfigured versions of reanimated. This PR updates gesture-handle, reanimated and adds worklets which are now required by reanimated.


apps/App.tsx
```diff
@@ -1,11 +1,9 @@
...
-// import Example from './Example';
-import * as Test from './src/tests';
+import Example from './Example';
 
 enableFreeze(true);
 
 export default function App() {
-  // return <Example />;
-  return <Test.TestBottomTabs />;
+  return <Example />;
```
kligarski: We usually leave commented out `import * as Test from './src/tests';` and `return <Test.TestBottomTabs />;` lines for convenience.

kligarski (commented):
We also need to do the same with `Example` app (Paper) but AFAIK Reanimated 4 does not support old architecture, so we should try with an older version I guess?

t0maboro (approved):
👍 remember to restore `apps/App.tsx`

kligarski (approved):
Looks good! Just rebase onto main and regenerate lock files, including both example apps' Podfile.locks.

FabricExample/ios/Podfile.lock
```diff
@@ -1739,8 +1739,8 @@ PODS:
     - React-graphics
     - React-ImageManager
     - React-jsi
-    - react-native-safe-area-context/common (= 5.6.0)
-    - react-native-safe-area-context/fabric (= 5.6.0)
+    - react-native-safe-area-context/common (= 5.4.0)
+    - react-native-safe-area-context/fabric (= 5.4.0)
```
kkafar, line 1743: It is interesting that the version got downgraded here 🤔

PR #3103, opened 2025-08-07
chore(iOS): Add compilation flag for gamma sources

t0maboro:
## Description

I need to use sources that are hidden inside the gamma directory in the other components, like the old native stack. In this PR I'm adding a compilation flag which may help in hiding some symbols when gamma is disabled.

## Changes

- Added `RNS_GAMMA_ENABLED` compilation flag
- Updated Podfiles

Compiled FabricExample and Example with changes from: https://github.com/software-mansion/react-native-screens/pull/3097

## Checklist

- [x] Ensured that CI passes


kkafar (approved):
✅

PR #3106, opened 2025-08-07
feat(iOS, SplitView): Add support for orientation on SplitViewHost

t0maboro:
## Description

Work in progress - refactor is needed before review

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


kmichalikk (approved):

ios/conversion/RNSConversions-SplitView.mm
```diff
@@ -97,6 +97,31 @@ UISplitViewControllerDisplayModeButtonVisibility SplitViewDisplayModeButtonVisib
...
+    case PortraitUpsideDown:
+      return RNSOrientationPortraitUpsideDown;
+    case Landscape:
+      return RNSOrientationLandscape;
+    case LandscapeLeft:
+      return RNSOrientationLandscapeLeft;
+    case LandscapeRight:
+      return RNSOrientationLandscapeRight;
+    case Inherit:
+    default:
```
kligarski, line 123: After your review, I added default case for bottom tabs as well but I also used `RCTLogError` - we should probably do the same for both conversion functions, not sure which one is better.

ios/gamma/orientation/RNSOrientationProviding.swift
```diff
@@ -0,0 +1,6 @@
+// TODO: investigate objc - swift interop and deduplicate this code
+// This class needs to follow the RNSOrientationProviding from objc.
```
kligarski, line 2: Should we also add a comment "on the other side" (obj-c implementation) to remember to update Swift one as well?

kligarski (approved):

ios/conversion/RNSConversions-SplitView.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3106/commits/cf20f72376b80a032197e8b0a8d7f23c152f0bb3

ios/gamma/orientation/RNSOrientationProviding.swift
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3106/commits/c7e0b3fef1153109c44ca33e9dcc9538a6fcd1cf

PR #3111, opened 2025-08-08
fix(iOS 26, Stack v4): fix RNSScreen being partially obstructed by UINavigationBar on iOS 26

kligarski:
## Description

Adds a workaround for UIKit `edgesForExtendedLayout` bug on iOS 26 (beta 5).

On iOS 26 (beta 5), content is rendered partially behind the header even with top edge being excluded from `edgesForExtendedLayout`. This happens because header's size has changed and now UINavigationBar is offset from the parent but it seems like `edgesForExtendedLayout` does not account for this. Apple advises to use `safeAreaLayoutGuide` instead of `edgesForExtendedLayout` but `edgesForExtendedLayout` is not deprecated yet therefore we hope that Apple fixes this in the near future. If this happens, workaround should be removed - I created a ticket for this: https://github.com/software-mansion/react-native-screens-labs/issues/358.

Fixes https://github.com/software-mansion/react-native-screens-labs/issues/332 (also contains more information about the bug).
Fixes https://github.com/software-mansion/react-native-screens-labs/issues/363.
Fixes https://github.com/software-mansion/react-native-screens/issues/3113.
Fixes one of the bugs mentioned here: https://github.com/software-mansion/react-native-screens-labs/issues/364.
## Changes

- use `safeAreaLayoutGuide` when header is shown (and it's not transparent)

## Screenshots / GIFs
| before | after |
| --- | --- |
| Simulator Screenshot - iPhone 16 Pro - 2025-08-12 at 11 47 21 | Simulator Screenshot - iPhone 16 Pro - 2025-08-12 at 11 44 31 |

Run:
- `Test3111`
- `TestModalNavigation`, open Screen1, change orientation to landscape and verify that headers do not have a gap between them
- `BottomTabsAndStack`, change orientation to landscape and verify that top padding is the same size as in portrait (content is not obstructed by header).

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -999,7 +999,7 @@ static RCTResizeMode resizeModeFromCppEquiv(react::ImageResizeMode resizeMode)
   [super prepareForRecycle];
   _initialPropsSet = NO;
 
-  // Part of 3111 workaround
+  // See comment above _safeAreaConstraints declaration for reason why this is necessary.
   [NSLayoutConstraint deactivateConstraints:_safeAreaConstraints];
   _safeAreaConstraints = nil;
 


WoLewicki (approved):

ios/RNSScreenContentWrapper.mm
```diff
@@ -189,6 +189,11 @@ + (void)load
 {
   [super load];
 }
+
++ (BOOL)shouldBeRecycled
```
kkafar, line 193: Can I get a word of explanation why is this disabled? Don't get me wrong - it's fine to do this, however I would like to understand why this was a problem.

kligarski: I can't remember the exact reason why I disabled it, but after reviewing it once again I think it might be unnecessary and clean-up code in HeaderConfig's `prepareForRecycle` is enough in the current solution.

I was worried about creating memory leaks with constraints but I verified now that `prepareForRecycle` correctly handles the cases I was worried about.

I had 2 problems with one of the previous solutions:

1. opening one modal after another caused crash (BAD_ACCESS, but this was due to attempting to re-use constraints before I added cleanup code in `prepareForRecycle`)
2. setting `headerShown: false` and `headerShown: true` once again did not work correctly (can't remember exact problem here, I think the layout was off)

but I can't reproduce them with or without disabling recycling for `ContentWrapper` now.

Sorry for the oversight and thanks for catching this. I'll create a ticket to bring recycling back for `ContentWrapper`.

kligarski: https://github.com/software-mansion/react-native-screens-labs/issues/391

kligarski: I managed to find the reason for disabling recycling but we should be able to bring it back without much trouble.

https://github.com/user-attachments/assets/91388cf1-e070-44fa-a0ff-4522673a0407

Open `Test3111`, fist open `ModalWithScrollView`, then `ModalWithScrollViewAndNoHeader`. Scroll will not work in `ModalWithScrollViewAndNoHeader`. If you first open `ModalWithScrollViewAndTransparentHeader` instead of `ModalWithScrollView`, the scroll will work.

You can also open `ModalWithScrollView`, add `headerShown: false,` and save - scroll will stop working.

This is because in order to use safe area, we need to do 2 things:
1. add constraints
2. enable auto layout (`contentWrapper.translatesAutoresizingMaskIntoConstraints = NO;`)

Currently, we don't disable auto layout if we set `headerShown: false,`, therefore removing `ScreenStackHeaderConfig`.

To fix this, we need to disable auto layout in `prepareForRecycle`.

```obj-c
  // See comment above _safeAreaConstraints declaration for reason why this is necessary.
  ((RNSScreenContentWrapper*)_safeAreaConstraints[0].firstItem).translatesAutoresizingMaskIntoConstraints = YES;
  [NSLayoutConstraint deactivateConstraints:_safeAreaConstraints];
  _safeAreaConstraints = nil;
```

This PoC fixes the issue.

We need to make sure that we will always get valid `RNSScreenContentWrapper` (there shouln't be an edge case where we somehow skip the reset) - now I remember that this was my original concern but maybe this is not a problem at all.

PR #3114, opened 2025-08-11
feat: add support for react-native 0.81

kligarski:
## Description

Adds support for `react-native@0.81`.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/357.

## Changes

- apply updates from upgrade-helper
- set `edgeToEdgeEnabled=true` (might require changes to how we detect edge-to-edge)
- changed imports from `metro-config` to use `private` alias (https://github.com/facebook/metro/pull/1530)
- changed deprecated `ShadowNode::Shared` to `std::shared_ptr<const ShadowNode>` (https://github.com/facebook/react-native/pull/52393)
- bump `react-native-gesture-handler`, `react-native-safe-area-context`
- https://github.com/software-mansion/react-native-screens/pull/3117
- https://github.com/software-mansion/react-native-screens/pull/3122

Run example apps.

## Checklist

- [ ] Ensured that CI passes


kligarski:
Not sure if we can satifly linter without using new imports for `codegenNativeComponent` (0.80.0) and `ImageSource` (0.81.0) which would break compatibility with older react-native versions.

Example/android/gradle.properties
```diff
@@ -43,6 +43,11 @@ newArchEnabled=false
 # If set to false, you will be using JSC instead.
 hermesEnabled=true
 
+# Use this property to enable edge-to-edge display support.
+# This allows your app to draw behind system bars for an immersive UI.
+# Note: Only works with ReactActivity and should not be used with custom Activity.
+edgeToEdgeEnabled=true
```
kkafar: We need to check what this is doing. 

Currently we detect whether the `edgeToEdge` is enabled or not by using `react-native-is-edge-to-edge` package. Will this conflict? What if we set `edgeToEdgeEnabled=false` here & we have `react-native-edge-to-edge` installed? What happens? What is the interaction here?

kkafar:
We did a quick research together with @kligarski about current situation with edge to edge support. 

https://github.com/facebook/react-native/commit/09ef774ff6dac10a00a8b35670f9b3941d810dfb introduced support for detecting whether app is in edge to edge in RN core. This API is available since 0.81. Therefore, while we support older versions we should keep support for `react-native-edge-to-edge` package, then we can migrate to this new API.

Example/android/gradle.properties
kkafar: Let's add comment here, that we do not support this yet.

kkafar (commented):
Nice! I've added single request to add comment & we will want to land the solution for https://github.com/software-mansion/react-native-screens-labs/issues/357

Example/android/gradle.properties
kligarski: created a separate PR with changes related to new edge to edge API: https://github.com/software-mansion/react-native-screens/pull/3121

kkafar (approved):
I think we're looking good.

PR #3115, opened 2025-08-11
feat(BottomTabs): Add `freezeContents` prop

kmichalikk:
## Description

This PR adds a prop to control freezing per screen, in addition to global `enableFreeze(true)` call.

## Testing

Freezing in BottomTabs doesn't work at the moment, but you can see in client logs that `shouldFreeze = true`/`false` behaves correctly.

### DIFF
--- a/src/components/bottom-tabs/BottomTabsScreen.tsx
+++ b/src/components/bottom-tabs/BottomTabsScreen.tsx
@@ -129,11 +129,9 @@ function BottomTabsScreen(props: BottomTabsScreenProps) {
 function shouldFreezeScreen(
   nativeViewVisible: boolean,
   screenFocused: boolean,
-  freezeOverride?: boolean,
+  freezeOverride: boolean | undefined,
 ) {
-  const shouldFreeze = freezeEnabled();
-
-  if (!shouldFreeze) {
+  if (!freezeEnabled()) {
     return false;
   }
 
@@ -143,10 +141,10 @@ function shouldFreezeScreen(
 
   if (featureFlags.experiment.controlledBottomTabs) {
     // If the tabs are JS controlled, we want to freeze only when given view is not focused && it is not currently visible
-    return shouldFreeze && !nativeViewVisible && !screenFocused;
+    return !nativeViewVisible && !screenFocused;
   }
 
-  return shouldFreeze && !nativeViewVisible;
+  return !nativeViewVisible;
 }
 
 function parseIconToNativeProps(icon: Icon | undefined): {


src/components/bottom-tabs/BottomTabsScreen.tsx
```diff
@@ -44,12 +44,14 @@ function BottomTabsScreen(props: BottomTabsScreenProps) {
...
     isFocused = false,
+    freezeContents,
     icon,
     selectedIcon,
     ...rest
   } = props;
 
-  let shouldFreeze = freezeEnabled();
+  let shouldFreeze =
+    freezeEnabled() && (freezeContents === undefined ? true : freezeContents);
```
kkafar: I'd like it to work as follows:

1. `freezeEnabled() == false` ==> we never freeze, no matter other options
2. `freezeContents == false` ==> we do not freeze, 
3. `freezeContents == true` ==> we freeze, unless `freezeEnabled() == false`,
4. `freezeContents == undefined` ==> old behaviour,

The code you wrote achieves this effect, but it does not convey the logic explicitly & it makes it harder to follow & later refactor. 

My suggestion would be to structure it as follows:

```typescript
  let shouldFreeze = false;

  if (freezeEnabled()) {
    if (freezeContents !== undefined) {
      shouldFreeze = freezeContents;
    } else if (featureFlags.experiment.controlledBottomTabs) {
      // If the tabs are JS controlled, we want to freeze only when given view is not focused && it is not currently visible
      shouldFreeze = shouldFreeze && !nativeViewIsVisible && !isFocused;
    } else {
      shouldFreeze = shouldFreeze && !nativeViewIsVisible;
    }
  } else {
    shouldFreeze = false;
  }

```

Additionally, I would consider extracting this whole logic to separate function (outside of the component).

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -297,6 +297,12 @@ export interface BottomTabsScreenProps {
       scrollToTop?: boolean;
     };
   };
+  /**
+   * @summary Enables per-screen control over React Suspense with react-freeze. Pass `false` to never freeze the screen.
```
kkafar:
```suggestion
   * @summary Allows to control whether a contents of a tab screen should be frozen or not. This overrides any default behaviour. 
```

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -297,6 +297,12 @@ export interface BottomTabsScreenProps {
...
+   *
+   * @default true
```
kkafar:
```suggestion
   * @default `undefined`
```

Undefined should be the default.

kkafar:
@Ubax 👀

PR #3119, opened 2025-08-11
chore(iOS): Rename RNSUIBarButtonItem to RNSBackBarButtonItem

kmichalikk:
## Description

From discussion in https://github.com/software-mansion/react-native-screens/pull/2987#discussion_r2253525009. We should rename the class into something more specific because it really does one specific thing.

## Changes

title

No changes, disabling back button menu should still work. You can modify BottomTabsTest / Tab4 with `screenOption={{ headerBackButtonMenuEnabled: false }}`


kkafar (commented):
It should rather be `RNSBackBarButtonItem` to keep the naming convention consistent, right?

PR #3121, opened 2025-08-11
feat: support for react-native's edgeToEdgeEnabled property

kligarski:
## Description

Adds support for `react-native`'s `edgeToEdgeEnabled` property.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/357.

## Changes

- rename `EdgeToEdgePackageDetector` to `EdgeToEdgeDetector`
- make `EdgeToEdgeDetector` use different implementation for RN < 0.81 and RN >= 0.81
- use new edgeToEdgeEnabled and isEdgeToEdgeFeatureFlagOn
- add clarifying comments

N/A

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/Example/android/gradle.properties
+++ b/Example/android/gradle.properties
@@ -46,6 +46,8 @@ hermesEnabled=true
 # Use this property to enable edge-to-edge display support.
 # This allows your app to draw behind system bars for an immersive UI.
 # Note: Only works with ReactActivity and should not be used with custom Activity.
+# This property is supported for RN >= 0.81.
+# For RN < 0.81, react-native-edge-to-edge package is necessary.
 edgeToEdgeEnabled=true
 
 # Disabling multiple instances check for Gesture Handler & Reanimted.
--- a/FabricExample/android/gradle.properties
+++ b/FabricExample/android/gradle.properties
@@ -41,4 +41,6 @@ hermesEnabled=true
 # Use this property to enable edge-to-edge display support.
 # This allows your app to draw behind system bars for an immersive UI.
 # Note: Only works with ReactActivity and should not be used with custom Activity.
+# This property is supported for RN >= 0.81.
+# For RN < 0.81, react-native-edge-to-edge package is necessary.
 edgeToEdgeEnabled=true


android/src/versioned/edge-to-edge/latest/com/swmansion/rnscreens/utils/EdgeToEdgeDetector.kt
```diff
@@ -0,0 +1,17 @@
+package com.swmansion.rnscreens.utils
+
+import com.facebook.react.views.view.isEdgeToEdgeFeatureFlagOn
+
+object EdgeToEdgeDetector {
+    // For RN >= 0.81, we use both isEdgeToEdgeFeatureFlagOn (from react-native) and the presence of
+    // react-native-edge-to-edge package to determine if app is in edge-to-edge.
+    val ENABLED: Boolean =
+        isEdgeToEdgeFeatureFlagOn ||
```
kkafar, line 9: I have doubts when & how this is initialised. Can we verify this? The `setEdgeToEdgeFeatureFlagOn` is called in `ReactActivityDelegate.onCreate` & that is "very late" in the runtime. When would be this value initialized? Is there a possibility that the flag will be read before its proper initialization?

kkafar (commented):
I have one question

android/src/versioned/edge-to-edge/latest/com/swmansion/rnscreens/utils/EdgeToEdgeDetector.kt
kligarski: It is lazily evaluated - currently when updateWindowTraits runs. You need to comment out logic in `edge-to-edge.tsx` to observe it though as now with react-native-edge-to-edge we don't send deprecated props.

obraz

kkafar: Can we get this confirmed in language spec or some trustworthy source? Empirical testing in debug configuration might be misleading.

kligarski: https://kotlinlang.org/docs/object-declarations.html#object-declarations-overview

> The initialization of an object declaration is thread-safe and **done on first access**.

obraz

kkafar: That sounds convincing. Thank you!

PR #3122, opened 2025-08-12
fix(Android): prevent view tag of `TabsHost` colliding with `ReactSurfaceView`

kkafar:
## Description

`TabsHost.id` collied with `ReactSurfaceView.id`, because RN 0.81
started to number the root views from 1, instead of 11 as it was before.
This caused issues, becuase `FragmentTransaction.add(containerViewId:
Int, ...)` requires us to pass the `containerViewId`, which it then uses
to find a view in hierarchy with that id & it mounts added fragment's
view there. If collision occurs, Android Framework mounts the fragment
directly under `ReactSurfaceView` instead of our `TabsHost`.

Proposed workaround does the following:

1. ~it passes int value of react native minor version down from gradle
   script to build config, so that it is accessible in platform code.
   AFAIK there is no dedicated API as of 0.81 for this, therefore we
   need a workaround.~ We start counting from 3, it should work no matter RN version.
2. It adds new `ViewIdGenerator` that implements new `ViewIdProviding`.
   It has capability of overriding the view generator, so that the lib
   can be integrated in brownfield scenarios. We should use this object
   to generate view id for any of our views that need it.
3. The `ViewIdGenerator` relies on internals of RN, that are referenced
   in the code comments, to provide non-colliding view ids (tags).

`TestBottomTabs` should work on 0.81. W/o this patch you won't see the bottom tab bar (it will be rendered, but under the top screen. The view would be mounted directly under `ReactSurfaceView`. 

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/helpers/ViewIdHelpers.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/helpers/ViewIdHelpers.kt
@@ -1,0 +1,58 @@
+package com.swmansion.rnscreens.gamma.helpers
+
+import androidx.annotation.UiThread
+import com.swmansion.rnscreens.BuildConfig
+
+interface ViewIdProviding {
+    fun generateViewId(): Int
+}
+
+/**
+ * This class generates view id that should not collide with ids (tags) assigned by renderer to React
+ * components.
+ *
+ * This class relies on internal renderer behaviours, that happened to change over the time, therefore
+ * it must be revisited & kept up to date with current RN behaviour until better solution is found.
+ *
+ * Until RN 0.81 ReactSurfaceView had tag `11`. In later versions it was changed to `1` (new arch only change).
+ *
+ * Up to the time of writing this (RN 0.81.0-rc.5) every version incremented tag version by 2,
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/swmansion/rnscreens/gamma/helpers/ViewIdHelpers.kt
```diff
@@ -0,0 +1,58 @@
...
+@UiThread
+private class NewArchAwareViewIdGenerator : ViewIdProviding {
+    private var nextId: Int = if (BuildConfig.REACT_NATIVE_VERSION_MINOR >= 81 && BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) 3 else 13
+
+    override fun generateViewId(): Int = nextId.also { progressViewId() }
+
+    private fun progressViewId() {
+        nextId += 2
+        if (isValidReactRootTag(nextId)) {
+            nextId += 2
```
kligarski, line 40: nitpick: we can consider creating and using a constant here with the increment value, something like:
```suggestion
        nextId += ID_INCREMENT
        if (isValidReactRootTag(nextId)) {
            nextId += ID_INCREMENT
```

kligarski (approved):
Looks good!

PR #3123, opened 2025-08-12
chore: add bottomTabsDebugLog

Ubax:
## Description

Hides JS logs for bottom tabs

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


kkafar (commented):
Yeah, we need debug logging system. This looks good, but I don't want to expose it from public API, as we'll refactor it surely. 

I'll apply the fixes, so that this can land swiftly.

kligarski (approved):

PR #3142, opened 2025-08-18
fix(iOS): Disable second transition coming from gestures when one is in progress

kmichalikk:
## Description

This is a followup to #3093 and #3141. We now handle repeatedly clicking back button on iOS 26, and disable interactiveContentPopGestureRecognizer which triggers without fullScreenGesture prop enabled, but it still is possible to click back button and swipe when fullScreenGesture is enabled. One possible solution is to check if transitionCoordinator is set and abort handling the gesture if it is.

## Changes

Added check for transitionCoordinator == nil before allowing for gesture to be received.

Use Test3093.

### DIFF
--- a/ios/RNSScreenStack.mm
+++ b/ios/RNSScreenStack.mm
@@ -1227,8 +1227,12 @@ RNS_IGNORE_SUPER_CALL_END
 // Be careful when adding another type of gesture recognizer.
 - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldReceivePressOrTouchEvent:(NSObject *)event
 {
-  if ([self isTransitionInProgress]) {
-    return NO;
+  if (@available(iOS 26, *)) {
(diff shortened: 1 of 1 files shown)


ios/RNSScreenStack.mm
```diff
@@ -1170,6 +1174,17 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldReceive
   return [self gestureRecognizer:gestureRecognizer shouldReceivePressOrTouchEvent:touch];
 }
 
+- (BOOL)isMidTransition
```
t0maboro: nit: `isTransitionInProgress` sounds better imo

WoLewicki (approved):
Let us keep in mind that being able to swipe fast and swipe from anywhere on screen to dismiss is something users will want with newest `iOS` if it becomes the default behavior.

PR #3143, opened 2025-08-18
feat(android, Tabs): Loading bottom tab icons from external source

t0maboro:
## Description

Adding a basic support for loading bottom-tabs icon from assets. This provides an alternative way to `iconResourceName`  for setting icon for the specific tab.

## Changes

- Adding `coil`: https://coil-kt.github.io/coil/ - a library that handles image loading.
- Adding basic ImageRequest - ImageBuilder integration
- Exposing a prop for setting icon from JS sources.

## Screenshots / GIFs

Screenshot 2025-08-18 at 13 30 04

Adding `iconResource` prop to `TestBottomTabs`.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabScreenViewManager.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabScreenViewManager.kt
@@ -1,24 +1,12 @@
 package com.swmansion.rnscreens.gamma.tabs
 
 import android.content.Context
-import android.graphics.drawable.BitmapDrawable
 import android.graphics.drawable.Drawable
 import android.util.Log
-import androidx.core.net.toUri
 import coil3.ImageLoader
 import coil3.asDrawable
 import coil3.request.ImageRequest
 import coil3.svg.SvgDecoder
-import com.bumptech.glide.Glide
-import com.facebook.common.executors.CallerThreadExecutor
-import com.facebook.common.references.CloseableReference
-import com.facebook.datasource.BaseDataSubscriber
-import com.facebook.datasource.DataSource
-import com.facebook.datasource.DataSubscriber
-import com.facebook.drawee.backends.pipeline.Fresco
-import com.facebook.imagepipeline.image.CloseableBitmap
-import com.facebook.imagepipeline.image.CloseableImage
-import com.facebook.imagepipeline.request.ImageRequestBuilder
 import com.facebook.react.bridge.ReadableMap
 import com.facebook.react.module.annotations.ReactModule
 import com.facebook.react.uimanager.ThemedReactContext
@@ -231,60 +219,12 @@ class TabScreenViewManager :
     ) {
         val uri = value?.getString("uri")
 
-        /*
-            This works, but the first icon is loaded to early, and it don't respect icon color
-            We should be able to handle the first problem using controller from fresco or something
-            similar to listen for finish event. The seccond thing is probably about the difference
-            BitmapDrawable vs RootDrawable, not sure how to resolve that. Tried converting to bitmap and
-            back to BitmapDrawable without luck.
-         */
-
-//        if (uri != null) {
-//            val draweeView = SimpleDraweeView(context)
-//            draweeView.setImageURI(uri)
-//            view.icon = draweeView.drawable
-//        }
-
-        // Works fine lib (coil3) around 160kb in final bundle
         if (uri != null) {
             loadUsingCoil(view.context, uri) {
                 view.icon = it
             }
         }
-
-        // Works fine lib (glide) around 270kb in final bundle
-//        if (uri != null) {
-//            runBlocking {
-//                val drawable = loadGlideImage(view.context, uri)
-//                view.icon = drawable
-//            }
-//        }
-
-    /*
-        When you debug you can see that the drawable is properly downloaded, but it's not visible
-        My guess is that fresco may clear it automatically
-        Additionaly active indicator seems to be broken I have no idea why
-     */
-
-//        if (uri != null) {
-//            loadUsingFresco(uri) {
-//                view.icon = it
-//            }
-//        }
     }
-
-    suspend fun loadGlideImage(
-        context: Context,
-        uri: String,
-    ): Drawable =
-        withContext(Dispatchers.IO) {
-            Glide
-                .with(context)
-                .asDrawable()
-                .load(uri.toUri())
-                .submit()
-                .get()
-        }
 
     fun loadUsingCoil(
         context: Context,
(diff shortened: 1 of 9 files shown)


android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabScreenViewManager.kt
```diff
@@ -192,6 +209,44 @@ class TabScreenViewManager :
...
+                .data(uri)
+                .target { drawable ->
+                    val stateDrawable = drawable.asDrawable(context.resources)
+                    onLoad(stateDrawable)
+                }.listener(
+                    onError = { _, result ->
+                        Log.e("RCTTabView", "Error loading image: $uri", result.throwable)
+                    },
+                    onCancel = {
+                        Log.w("RCTTabView", "Image loading request: $uri has been cancelled")
```
WoLewicki: `RCTTabView` is the name of this class?

src/components/bottom-tabs/BottomTabsScreen.tsx
```diff
@@ -116,6 +123,8 @@ function BottomTabsScreen(props: BottomTabsScreenProps) {
       onWillDisappear={onWillDisappearCallback}
       onDidDisappear={onDidDisappearCallback}
       isFocused={isFocused}
+      // I'm keeping iconResource as a fallback if `Image.resolveAssetSource` has failed for some reason.
+      iconResource={parsedIconResource || iconResource}
```
WoLewicki: Will the native side be able to parse the raw `iconResource` ?

t0maboro: it was failing in `android/src/paper/java/com/facebook/react/viewmanagers/RNSBottomTabsScreenManagerDelegate.java` - a generated code for paper

precisely,
```
      case "iconResource":
        mViewManager.setIconResource(view, (ReadableMap) value);
        break;
```

However, from the JS `require` API will return a number that represents the asset id. If I'm not missing anything, we cannot do that in the native code and we need this conversion in JS to pass our data as an object which will be converted to `ReadableMap`

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabScreenViewManager.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3143/commits/8cf42620f2fc642dd772167144d05a654421215b

changed to `[RNScreens]` for consistency with other files

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabScreenViewManager.kt
```diff
@@ -192,6 +209,44 @@ class TabScreenViewManager :
...
+        val uri = value?.getString("uri")
+
+        if (uri != null) {
+            loadUsingCoil(view.context, uri) {
+                view.icon = it
+            }
+        }
+    }
+
+    fun loadUsingCoil(
```
kmichalikk: we could use `private` here

kmichalikk (approved):
Looks good and loads images correctly when I tested.

I see we added support for .svg -- maybe we should add one to the test to see that it works (it does, but let's maybe have it in repo).

I also noticed that require() allows for .js .json etc and app doesn't crash, just doesn't render anything for the icon -- is this okay?

t0maboro:
> I also noticed that require() allows for .js .json etc and app doesn't crash, just doesn't render anything for the icon -- is this okay?

I believe that this behavior is ok, because from these mentioned formats we cannot have an asset explicitly, so we're ignoring this data.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabScreenViewManager.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3143/commits/a64ae03d06db0a9633e8502a865a50797e3253b4

t0maboro:
> I see we added support for .svg -- maybe we should add one to the test to see that it works (it does, but let's maybe have it in repo).

https://github.com/software-mansion/react-native-screens/pull/3143/commits/34cc64dfa79182d77ac014071304cab2a9422c60

t0maboro:
> Looks good overall, I'd ask Expo team if coil is something they thought of for e.g. expo-image.

Conclusions after internal discussion with Lukasz from Expo:
- `expo-image` is using `glide`
- at the time when the `expo-image` was implemented, `glide` was better in terms of performance
- `glide` may introduce issues with compatibility if we have 2 libraries using different versions
- for our case - loading a few icons, which for most of the cases won't change dynamically, the performance isn't that important to use `glide` whose size is larger than `coil`

src/components/bottom-tabs/BottomTabsScreen.tsx
WoLewicki: Hmm ok but I am not sure what is the conclusion from this? Can we safely use `iconResource` as a fallback?

t0maboro: after internal discussion, I changed that fallback to non-crashing version https://github.com/software-mansion/react-native-screens/pull/3143/commits/2aa158e31d8b2592974c407c06c3b68fa85f86a2 + https://github.com/software-mansion/react-native-screens/pull/3143/commits/96011ea917fc5c8afc8994dc4a2005f18090f204

NiuGuohui:
I don't think directly relying on libraries like Coil in the screen project is a good solution. This will bring a lot of burden to users who don't use bottom-tabs. Additionally, the androidsvg library that Coil depends on will conflict with other versions of androidsvg integrated by the project itself.

WoLewicki:
What alternative would you suggest then @NiuGuohui ?

NiuGuohui:
Hi @WoLewicki ,

The RNScreen project, as a project that is becoming increasingly fundamental, should, in my opinion, maintain its flexibility and lightness to minimize potential future issues. However, I am unsure whether this aligns with the direction of the RNScreen project team.

Regarding iconResource specifically, I think using ReactNative's Fresco directly can reduce library dependencies. Reusing it might reduce the overhead of adding Coil, avoiding potential conflicts (like the androidsvg issue) and keeping the dependency tree slimmer.

For SVG support and other advanced icon needs, I’d suggest guiding developers toward custom tab bar renderers.It lets devs implement exactly what they need without bloating the library with edge-case logic. Both the library and developers would benefit from lower complexity—maintainers avoid supporting endless configuration permutations, and devs get full control over their UI.

What do you think about this direction?

WoLewicki:
@NiuGuohui thanks for your input! We were exploring using `Fresco` and can definitely reiterate on that idea. As for the custom renderers, I'm not sure if it fits the idea of giving users a simple API with options to use different image formats that we wanted to have. We will discuss this internally and come back with something soon. 

Of course PRs are also welcome so if you have an idea how to approach it, please submit one and we can review and hopefully merge it too.

NiuGuohui:
Additionally, the current version will also conflict with react-native-turbo-image because this library also uses coil but with a different version.
I believe that many projects relying on coil will encounter this issue as long as their dependency version is different from that of the screens project.

PR #3144, opened 2025-08-18
chore: Hide native logs

Ubax:
## Description

When using bottom tabs, many logs are printed from the native components. These logs are useful for development and debugging in screens library, but are mostly noise when using in external app

## Changes

Adds `RNSLog` function on iOS and `RNSLog` object on Android. Both can be conditionally enabled for debugging.

Install iOS pods with `RNS_DEBUG_LOGGING=1 pod install` to see the logs on iOS.

For Android `rnsDebugLogsEnabled` needs to be set to `true` in `gradle.properties`

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


RNScreens.podspec
```diff
@@ -16,6 +17,14 @@ min_supported_ios_version = new_arch_enabled ? "15.1" : "15.1"
 min_supported_tvos_version = "15.1"
 min_supported_visionos_version = "1.0"
 
+cpp_flags = ""
+if debug_logging
+  cpp_flags << " -DRNS_DEBUG_LOGGING=1"
+end
+if gamma_project_enabled
+  cpp_flags << " -DRNS_GAMMA_ENABLED=1"
+end
```
t0maboro:
```suggestion
rnscreens_cpp_flags = []

rnscreens_cpp_flags << "-DRNS_DEBUG_LOGGING=1" if debug_logging
rnscreens_cpp_flags << "-DRNS_GAMMA_ENABLED=1" if gamma_project_enabled

other_cpp_flags = rnscreens_cpp_flags.join(" ")
```

I'd suggest a more compact form, but this is a nitpick

RNScreens.podspec
```diff
@@ -48,7 +57,11 @@ Pod::Spec.new do |s|
     # we can not have Swift code in stable package. 
     s.pod_target_xcconfig = {
       'DEFINES_MODULE' => 'YES',
-      'OTHER_CPLUSPLUSFLAGS' => '-DRNS_GAMMA_ENABLED=1'
+      'OTHER_CPLUSPLUSFLAGS' => cpp_flags
+    }
+  else
+    s.pod_target_xcconfig = {
+      'OTHER_CPLUSPLUSFLAGS' => cpp_flags
```
t0maboro: To avoid duplication, can we create a `rnscreens_config` object that will have common flags assigned in the top level, gamma-related flags inside the if statement and then be assigned:
```
s.pod_target_xcconfig = rnscreens_config
```

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
```diff
@@ -286,7 +286,7 @@ class TabsHost(
     override fun getFragmentForTabScreen(tabScreen: TabScreen): TabScreenFragment? = tabScreenFragments.find { it.tabScreen === tabScreen }
 
     private fun updateBottomNavigationViewAppearance() {
-        Log.w(TAG, "updateBottomNavigationViewAppearance")
+        RNSLog.d(TAG, "updateBottomNavigationViewAppearance")
```
t0maboro, line 289: nit: I don't have a full context why it was warning earlier, but I'm wondering if we should keep the verbosity level as it was (at least in the scope of this PR)

Ubax: I think that this was meant to be debug. This is standard operation without any condition, so no need to have it as warning. 

Tagging @kkafar to check.

RNScreens.podspec
Ubax: Thanks

PR #3149, opened 2025-08-20
feat(iOS, Tabs): support system items on iOS

Ubax:
## Description

This PR implements support for different system items on iOS

Currently the most useful one is the search system item, which creates separate tab bar item on iOS 26.

Simulator Screenshot - iPhone 16 Pro - 2025-08-20 at 09 57 13

## Changes

1. Added enums for system items
2. When system item is used, then the base tab bar item from controller is saved. It is restored when the system item is set to none

Uncomment system items in index.tsx

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabScreenViewManager.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabScreenViewManager.kt
@@ -247,7 +247,7 @@ class TabScreenViewManager :
 
         imageLoader?.enqueue(request)
     }
-    
+
     override fun setSystemItem(
         view: TabScreen,
         value: String?,


ios/bottom-tabs/RNSBottomTabsScreenComponentView.h
```diff
@@ -79,6 +79,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, readonly) BOOL overrideScrollViewContentInsetAdjustmentBehavior;
 
+@property (nonatomic) RNSBottomTabsScreenSystemItem role;
```
kmichalikk: It seems that you were renaming role to systemItem, but missed some places, please adjust them too

Ubax: Thanks for finding this

Issue #3152, opened 2025-08-20
[4.15.0] `TypeError: (0 , _reactNative.codegenNativeComponent) is not a function`


huextrat:
### Description

From v4.15.0, we have this error during jest test:

TypeError: (0 , _reactNative.codegenNativeComponent) is not a function
`import { createNativeStackNavigator } from "@react-navigation/native-stack";`

### Steps to reproduce

render a screen with `createNativeStackNavigator` in a jest test

### Snack or a link to a repository

https://github.com/huextrat

### Screens version

4.15.0

### React Native version

0.79.6

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

Paper (Old Architecture)

### Build type

None

### Device

None

### Acknowledgements

Yes

Izocel:
+1

t0maboro:
Hi, thank you for reporting this issue and providing all information. I was able to reproduce it on my end and I'm submitting a PR with fix. It will land in 4.15.1. If you need to unblock, I'd recommend to create a patch with the changes from: https://github.com/software-mansion/react-native-screens/pull/3153

PR #3153, opened 2025-08-20
fix(JS): Keep relative imports for backward compatibility

t0maboro:
## Description

In this commit: https://github.com/software-mansion/react-native-screens/commit/3a3a6c5449796aeef229d40b5805721b808bf3ab `eslint` was complaining about the import. Moreover, running `lint-js` changes this line itself after upgrading to 0.81. As for now, we need to keep relative import for a backward compatibility with versions that we support and for which `codegenNativeComponent` wasn't exported.

Fixes https://github.com/software-mansion/react-native-screens/issues/3152 .

## Changes

- Skipping linter check for the line with the relative import.

- Create a project from template, using for example 0.79.6
- Add `react-native-screens@^4.15.0` and `@react-navigation` 
- Add native stack to `App.tsx`
- Run `yarn test`

For testing my changes I packed RNScreens from this PR to tarball and sideloaded it to the application with `file://`

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/src/fabric/FullWindowOverlayNativeComponent.ts
+++ b/src/fabric/FullWindowOverlayNativeComponent.ts
@@ -1,6 +1,7 @@
 'use client';
 
-import { codegenNativeComponent } from 'react-native';
+// eslint-disable-next-line -- required for backward compatibility
+import codegenNativeComponent from 'react-native/Libraries/Utilities/codegenNativeComponent';
 import type { ViewProps } from 'react-native';
 import { WithDefault } from 'react-native/Libraries/Types/CodegenTypes';
 
--- a/src/fabric/ModalScreenNativeComponent.ts
+++ b/src/fabric/ModalScreenNativeComponent.ts
@@ -1,6 +1,7 @@
 'use client';
 
-import { codegenNativeComponent } from 'react-native';
+// eslint-disable-next-line -- required for backward compatibility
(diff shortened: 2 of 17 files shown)


src/fabric/FullWindowOverlayNativeComponent.ts
```diff
@@ -1,6 +1,7 @@
 'use client';
 
-import { codegenNativeComponent } from 'react-native';
+// eslint-disable-next-line -- required for backward compatibility
```
kmichalikk: I'd consider narrowing it down to the specific rule. I believe
`// eslint-disable-next-line @react-native/no-deep-imports` should work

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3153/commits/3ed60a844fa28aa4315371350ee72d4d85b43ef7

WoLewicki (commented):
Keep in mind that deep imports will probably be removed in of next version so please add it to list of things to change when we drop support for <0.80.

t0maboro:
> Keep in mind that deep imports will probably be removed in of next version so please add it to list of things to change when we drop support for <0.80.

https://github.com/software-mansion/react-native-screens-labs/issues/378

PR #3164, opened 2025-08-25
fix(iOS, Tabs): refactor setting systemItem

kligarski:
## Description

This PR addresses some problems related to `systemItem`:
1. on startup, tab bar item with `systemItem` but without custom icon would not show the default icon (it was overriden by nil SFSymbol),
2. when using custom icon with `systemItem` and then removing the custom icon, the default icon for `systemItem` would not be restored,
3. if badge value was updated as the same time as `systemItem` was set, the value of the badge would not be updated (due to badge value being updated _before_ `systemItem`).

Fixes https://github.com/software-mansion/react-native-screens-labs/issues/380.

## Changes

- move updates to `tabBarItem` to separate method after setting props

### Before 

https://github.com/user-attachments/assets/d5b197d0-d29c-4587-8cfa-5c9ed96734e6

### After

https://github.com/user-attachments/assets/96c71b0b-6c89-4677-a726-a17588dca4a7

1. Run `TestBottomTabs` with and without custom icon for Tab4 (with systemItem). Correct icon should be visible.
5. Run `TestBottomTabs` without systemItem for Tab4. Change both badgeValue and systemItem at the same time and save file. Badge value should be updated.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [x] Ensured that CI passes

### DIFF
--- a/ios/bottom-tabs/RNSTabBarAppearanceCoordinator.mm
+++ b/ios/bottom-tabs/RNSTabBarAppearanceCoordinator.mm
@@ -57,8 +57,6 @@
       tabBarItem.image = [UIImage systemImageNamed:screenView.iconSfSymbolName];
     } else if (screenView.systemItem == RNSBottomTabsScreenSystemItemNone) {
       // We don't want to override systemItem's icon if no custom icon is provided.
-      // There is unhandled edge case: if you use systemItem, set custom icon and then remove custom icon -> we won't
-      // restore systemItem's icon.
       tabBarItem.image = nil;
     }
 
--- a/ios/bottom-tabs/RNSTabBarAppearanceCoordinator.mm
+++ b/ios/bottom-tabs/RNSTabBarAppearanceCoordinator.mm
@@ -53,8 +53,21 @@
               withImageLoader:(RCTImageLoader *_Nullable)imageLoader
 {
   if (screenView.iconType == RNSBottomTabsIconTypeSfSymbol) {
-    tabBarItem.image = [UIImage systemImageNamed:screenView.iconSfSymbolName];
-    tabBarItem.selectedImage = [UIImage systemImageNamed:screenView.selectedIconSfSymbolName];
+    if (screenView.iconSfSymbolName != nil) {
+      tabBarItem.image = [UIImage systemImageNamed:screenView.iconSfSymbolName];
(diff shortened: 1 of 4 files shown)


ios/bottom-tabs/RNSTabBarAppearanceCoordinator.mm
```diff
@@ -53,8 +53,19 @@ - (void)setIconsForTabBarItem:(UITabBarItem *)tabBarItem
...
 {
   if (screenView.iconType == RNSBottomTabsIconTypeSfSymbol) {
-    tabBarItem.image = [UIImage systemImageNamed:screenView.iconSfSymbolName];
-    tabBarItem.selectedImage = [UIImage systemImageNamed:screenView.selectedIconSfSymbolName];
+    if (screenView.iconSfSymbolName != nil) {
+      tabBarItem.image = [UIImage systemImageNamed:screenView.iconSfSymbolName];
+    } else if (screenView.systemItem == RNSBottomTabsScreenSystemItemNone) {
+      // We don't want to override systemItem's icon if no custom icon is provided.
+      tabBarItem.image = nil;
+    }
```
kmichalikk: Could you confirm that we need this else case or is the image already nil because the tab bar item used default init: anyways?

Or is it there to explicitly state that we fallback to default?

ios/bottom-tabs/RNSTabBarAppearanceCoordinator.mm
```diff
@@ -53,8 +53,19 @@ - (void)setIconsForTabBarItem:(UITabBarItem *)tabBarItem
...
+
+    if (screenView.selectedIconSfSymbolName != nil) {
+      tabBarItem.selectedImage = [UIImage systemImageNamed:screenView.selectedIconSfSymbolName];
+    } else if (screenView.systemItem == RNSBottomTabsScreenSystemItemNone) {
+      // We don't want to override systemItem's icon if no custom icon is provided.
+      tabBarItem.selectedImage = nil;
+    }
```
kmichalikk: same as above

ios/bottom-tabs/RNSTabBarAppearanceCoordinator.mm
kligarski: You're right, we don't need this now because we're creating new tab bar item every time we modify the icon.

PR #3167, opened 2025-08-27
feat(Android): Support dark mode in android BottomTabs

kmichalikk:
## Description

This PR adds support for resolving configuration attributes in TabsHostAppearanceCoordinator to fetch correct colors for light and dark mode.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/267.

@kkafar
[Add support for reacting to Appearance.setColorScheme calls](https://github.com/software-mansion/react-native-screens/pull/3167/commits/11b71b12ec7c0ffd0e0245b60b206b324e9e9129)

This was a bit tricky. Our views are not notified in any way that the
configuration got changed, but fortunately our fragments do. Therefore,
I've forwarded configuration changes from `TabScreenFragment` to
`TabsHost`. This was done indirectly, through `TabScreen` &
`TabScreenDelegate` to avoid coupling the components.

When doing so I encountered another problem: the resources of our theme
applied to `ContextThemeWrapper` were not updated. This was done
because, the API that RN uses - `AppCompatDelegate.setDefaultNightMode`
applies theme changes only to top level application context (and
configuration changes are then dispatched). The solution would be to
recreate the `ContextThemeWrapper`, however lifecycle of our fragments &
views is bound to the `ContextThemeWrapper`, therefore we would prefer
to avoid recreation of the whole managed hierarchy.

I've settled with manually applying appropriate theme depending on
uimode provided by the configuration. This solves the problem.

## Changes

Replaced context.getColor call for light mode colors only with context.resolveAttribute to query values set for current mode.

https://github.com/user-attachments/assets/8d6dd1a9-7fa1-4078-8829-60b98090b958

Use TestBottomTabs, make sure to comment out default prop values in `apps/src/shared/gamma/containers/bottom-tabs/BottomTabsContainer.tsx`

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabScreenDelegate.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabScreenDelegate.kt
@@ -15,7 +15,10 @@ internal interface TabScreenDelegate {
      * **If a fragment is associated with the tab screen**, notify the delegate that the fragment
      * got configuration update.
      */
-    fun onFragmentConfigurationChange(tabScreen: TabScreen, config: Configuration)
+    fun onFragmentConfigurationChange(
+        tabScreen: TabScreen,
+        config: Configuration,
+    )
 
     /**
      * This returns fragment **if one is associated with given tab screen**.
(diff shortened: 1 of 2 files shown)


kkafar (approved):
I've added support for handling `Appearance.setColorScheme` RN API. 
I've updated the PR description with appropriate information. 

See here: https://github.com/software-mansion/react-native-screens/pull/3167/commits/11b71b12ec7c0ffd0e0245b60b206b324e9e9129

PR #3168, opened 2025-08-27
feat(iOS 26, Stack v4): support for new search bar placements, integration with toolbar

kligarski:
## Description

WIP

## Changes

WIP

## Screenshots / GIFs

WIP

WIP

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
+++ b/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
@@ -253,7 +253,7 @@ For iOS:
 - `modal` will use:
    *  on iOS 18 and later [`UIModalPresentationAutomatic`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationautomatic?language=objc). However, since the iOS 18 changed the default behaviour from `UIModalPresentationPageSheet` to `UIModalPresentationFormSheet` (it looks vastly different on regular size classes devices, e.g. iPad), for the sake of backward compatibility, we keep the default behaviour from before iOS 18. *This might change in future major release of `react-native-screens`.
    *  on iOS 13 and later [`UIModalPresentationAutomatic`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationautomatic?language=objc)
-   * on iOS 12 and earlier will use [`UIModalPresentationFullScreen`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationfullscreen?language=objc). 
+   * on iOS 12 and earlier will use [`UIModalPresentationFullScreen`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationfullscreen?language=objc).
 - `fullScreenModal` will use [`UIModalPresentationFullScreen`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationfullscreen?language=objc)
 - `formSheet` will use [`UIModalPresentationFormSheet`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/uimodalpresentationformsheet?language=objc)
 - `pageSheet` will use [`UIModalPresentationPageSheet`](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/pagesheet?language=objc)
@@ -471,6 +471,7 @@ To render a search bar use `ScreenStackHeaderSearchBarView` with `<SearchBar>` c
 - `onSearchButtonPress` - A callback that gets called when the search button is pressed. It receives the current text value of the search bar.
 - `placeholder` - Text displayed when search field is empty. Defaults to an empty string.
 - `placement` - Placement of the search bar in the navigation bar. (iOS only)
+- `allowToolbarIntegration` - Indicates whether the system can place the search bar among other toolbar items on iPhone. (iOS only)
 - `textColor` - The search field text color.
 - `hintTextColor` - The search hint text color. (Android only)
 - `headerIconColor` - The search and close icon color shown in the header. (Android only)
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -716,7 +716,11 @@ RNS_IGNORE_SUPER_CALL_END
           if (@available(iOS 26.0, *)) {
             // Workaround for missing search bar on root stack screen.
             // See: https://github.com/software-mansion/react-native-screens/pull/3098
-            navitem.searchBarPlacementAllowsToolbarIntegration = NO;
+            if (navitem.preferredSearchBarPlacement == UINavigationItemSearchBarPlacementStacked) {
+              navitem.searchBarPlacementAllowsToolbarIntegration = NO;
+            } else {
+              navitem.searchBarPlacementAllowsToolbarIntegration = searchBar.allowToolbarIntegration;
+            }
           }
 #endif /* Check for iOS 26.0 */
 #endif /* !TARGET_OS_TV */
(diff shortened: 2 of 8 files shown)


guides/GUIDE_FOR_LIBRARY_AUTHORS.md
```diff
@@ -456,7 +456,7 @@ To render a search bar use `ScreenStackHeaderSearchBarView` with `<SearchBar>` c
 - `autoFocus` - If `true` automatically focuses search bar when screen is appearing. (Android only)
 - `barTintColor` - The search field background color. By default bar tint color is translucent.
 - `tintColor` - The color for the cursor caret and cancel button text. (iOS only)
-- `cancelButtonText` - The text to be used instead of default `Cancel` button text. (iOS only)
+- `cancelButtonText` - The text to be used instead of default `Cancel` button text. (iOS only, **deprecated**)
```
kmichalikk: I believe we shouldn't tag it as deprecated if it is still perfectly valid for 18.5 ? Maybe a note, or **deprecated since iOS 26** would be better?

ios/RNSSearchBar.mm
```diff
@@ -377,6 +411,11 @@ - (void)handleCommand:(const NSString *)commandName args:(const NSArray *)args
   RCTRNSSearchBarHandleCommand(self, commandName, args);
 }
 
++ (BOOL)shouldBeRecycled
+{
+  return NO;
```
kmichalikk, line 418: Let's maybe add why we shouldn't recycle it?

src/types.tsx
```diff
@@ -98,7 +98,13 @@ export type GestureResponseDistanceType = {
   bottom?: number;
 };
 
-export type SearchBarPlacement = 'automatic' | 'inline' | 'stacked';
+export type SearchBarPlacement =
+  | 'automatic'
+  | 'inline' // deprecated
```
kmichalikk: since iOS 26 \*

src/types.tsx
```diff
@@ -816,14 +824,44 @@ export interface SearchBarProps {
...
    *
    * * `automatic` - the search bar is placed according to current layout
-   * * `inline` - the search bar is placed on the trailing edge of navigation bar
    * * `stacked` - the search bar is placed below the other content in navigation bar
+   * * `integrated` - (>= iOS 26) the search bar is placed on the trailing edge of navigation bar. On iPhone,
+   *   it may be integrated into the toolbar
+   * * `integratedButton` - (>= iOS 26) the search bar has the same placement as `integrated`, except that
+   *   the inactive search bar is always shown as a button even when space permits a search field
+   * * `integratedCentered` - (>= iOS 26) the search bar has the same placement as `integrated`, except that
+   *   in the regular width on iPad, the search bar is ceneterd in the navigation bar. Only
```
kmichalikk: centered \*

kmichalikk (commented):
Looks good overall.
There are glitches with some combinations, but there are present in native apps, too.
I'm worried that we didn't check out enough combinations, though. We should also see what happens when we set different styling options for header.
When I disabled headerTransparent, it broke 'automatic' too:

https://github.com/user-attachments/assets/bf034ca0-9e8c-430e-b33b-c16745f6f9a8

src/types.tsx
kligarski: https://github.com/software-mansion/react-native-screens/pull/3168/commits/666afe244385cdf6e7e98c783f87ab7b4b05e466

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
kligarski: https://github.com/software-mansion/react-native-screens/pull/3168/commits/666afe244385cdf6e7e98c783f87ab7b4b05e466

src/types.tsx
kligarski: https://github.com/software-mansion/react-native-screens/pull/3168/commits/666afe244385cdf6e7e98c783f87ab7b4b05e466

ios/RNSSearchBar.mm
kligarski: You're right. I added a link to this PR.

Some details:
recycling was causing a bug when we had this kind of setup
```
A
|- B (allowToolbarIntegration: true)
|- C (allowToolbarIntegration: false)
```
If I recall correctly, if you went A -> B -> A -> C -> A -> B, the search bar would not appear. I suspect that using the same search controller for different values of `allowToolbarIntegration` caused the bug.

kligarski:
> When I disabled headerTransparent, it broke 'automatic' too:

I managed to reproduce this as well in bare UIKit app. By default, UIKit uses `hideWhenScrolling: true` but we use `hideWhenScrolling: false`, that's why it was hard to reproduce. I described the details in PR description and corresponding issue.

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
```diff
@@ -456,7 +456,7 @@ To render a search bar use `ScreenStackHeaderSearchBarView` with `<SearchBar>` c
 - `autoFocus` - If `true` automatically focuses search bar when screen is appearing. (Android only)
 - `barTintColor` - The search field background color. By default bar tint color is translucent.
 - `tintColor` - The color for the cursor caret and cancel button text. (iOS only)
-- `cancelButtonText` - The text to be used instead of default `Cancel` button text. (iOS only)
+- `cancelButtonText` - The text to be used instead of default `Cancel` button text. (iOS only, **deprecated** starting from iOS 26)
```
kkafar:
```suggestion
- `cancelButtonText` - The text to be used instead of default `Cancel` button text. *Has not effect as of iOS 26*. **Deprecated**. (iOS only)
```

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -714,9 +714,13 @@ + (void)updateViewController:(UIViewController *)vc
 #if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && defined(__IPHONE_26_0) && \
     __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_26_0
           if (@available(iOS 26.0, *)) {
-            // Workaround for missing search bar on root stack screen.
-            // See: https://github.com/software-mansion/react-native-screens/pull/3098
-            navitem.searchBarPlacementAllowsToolbarIntegration = NO;
+            if (navitem.preferredSearchBarPlacement != UINavigationItemSearchBarPlacementStacked) {
```
kkafar, line 725: We need some commments why this is implemented like that. Exactly what you told me on call 🙏🏻

src/types.tsx
```diff
@@ -736,6 +742,8 @@ export interface SearchBarProps {
   /**
    * The text to be used instead of default `Cancel` button text
    *
+   * @deprecated Starting from iOS 26, cancel button does not have any text.
```
kkafar:
```suggestion
   * @deprecated Starting from iOS 26, cancel button does not have any text, therefore this prop has no longer any effect.
```

kkafar (commented):
Few adjustments request & we're good.

src/types.tsx
kligarski: https://github.com/software-mansion/react-native-screens/pull/3168/commits/b43898fad323e8f97b516cbe26c42c49eb510763

ios/RNSScreenStackHeaderConfig.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3168/commits/b43898fad323e8f97b516cbe26c42c49eb510763

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
kligarski: https://github.com/software-mansion/react-native-screens/pull/3168/commits/b43898fad323e8f97b516cbe26c42c49eb510763

Issue #3161, opened 2025-08-23
[Xcode 26] `fullScreenGestureEnabled` cause bug on scroll


huextrat:
### Description

From Xcode 26 and iOS 26 if we have `fullScreenGestureEnabled` on a screen with a list, the scroll is buggy.

`fullScreenGestureEnabled: true` -> https://github.com/user-attachments/assets/8cfeaaa0-0981-4765-a026-8c7ba7a30099

`fullScreenGestureEnabled: false` -> https://github.com/user-attachments/assets/ffaf8a43-c27c-470e-b1c6-85eb2be6b2a0

### Steps to reproduce

Scroll one time, release and do a new scroll during the scroll inertia. The latest scroll intention is not working

Check the repo, open profile and do multiple scroll while the list is scrolling, you will notice that the scroll stop itself when it should continue.

Compare it by disabling / enabling `fullScreenGestureEnabled` and you will notice faster the wrong behavior

### Snack or a link to a repository

https://github.com/huextrat/repro-screens-ios

### Screens version

4.15.2

### React Native version

0.79.6

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

PR #3173, opened 2025-08-29
feat(iOS): Handle interactiveContentPopGesture for iOS 26

kmichalikk:
## Description

Resolves https://github.com/software-mansion/react-native-screens-labs/issues/369, might resolve https://github.com/software-mansion/react-native-screens/issues/3161, reverts https://github.com/software-mansion/react-native-screens/pull/3141, https://github.com/software-mansion/react-native-screens/pull/3142

This PR attempts to enable interactiveContentPopGestureRecognizer for iOS 26 to achieve native screen popping behavior. Until 26, the default was to swipe from the edge of the screen. We had the option to do fullscreen switch, which was controlled by a `fullScreenSwipeEnabled` prop. Since the default behavior has changed, this prop needs rethinking, along with `gestureResponseDistance`. The new swipe is strictly for navigation purposes, and the old one, I believe, was more general-purpose.

New iOS allows for popping multiple screens almost at once, which we still cannot support due to asynchronous nature of stack updates coming from host to JS that would create a "feedback loop" in situation like the following: host pops 1. screen + sends update, pops  2. screen + sends update -> JS acknowledges 1. update + sends updated state -> host gets 2. screen from JS and pushes it again. This PR attempts to block more than 1 pop at once by removing interactions from the whole screen. As a (desired) side effect, this also disables interactions on the screen below the one that is popped until the transition finishes.

## Changes

- removed `RNSPanGestureRecognizer` from iOS 26 build and replace it with native `interactiveContentPopGestureRecognizer`
- disabled all interactions when screens are in transition

TBD

### DIFF
--- a/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
+++ b/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
@@ -55,6 +55,7 @@ Defaults to `false`. When `enableFreeze()` is run at the top of the application 
 ### `fullScreenSwipeEnabled` (iOS only)
 
 Boolean indicating whether the swipe gesture should work on whole screen. Swiping with this option results in the same transition animation as `simple_push` by default. It can be changed to other custom animations with `customAnimationOnSwipe` prop, but default iOS swipe animation is not achievable due to usage of custom recognizer. Defaults to `false`.
+IMPORTANT: Starting from iOS 26, full screen swipe to dismiss is the native behavior, and this prop is ignored.
 
 ### `fullScreenSwipeShadowEnabled` (iOS only)
 
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -1466,6 +1466,12 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
 // TODO: Find out why this is executed when screen is going out
 - (void)viewWillAppear:(BOOL)animated
 {
+  if (@available(iOS 26, *)) {
+    // In iOS 26, as soon as another screen appears in transition, it is interactable
+    // To avoid glitches resulting from clicking buttons mid transition, we temporarily disable all interactions
+    // Because nested stack exist, we need to find a common top view for the whole app
+    [self findReactRootViewController].view.userInteractionEnabled = false;
+  }
   [super viewWillAppear:animated];
   if (!_isSwiping) {
     [self.screenView notifyWillAppear];
@@ -1530,6 +1536,10 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
 
 - (void)viewDidAppear:(BOOL)animated
 {
+  if (@available(iOS 26, *)) {
+    // Reenable interactions, see viewWillAppear
+    [self findReactRootViewController].view.userInteractionEnabled = true;
+  }
   [super viewDidAppear:animated];
   if (!_isSwiping || _shouldNotify) {
     // we are going forward or dismissing without swipe
@@ -1599,6 +1609,18 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
     }
 #endif
   }
+}
+
+- (UIViewController *)findReactRootViewController
+{
+  UIView *parent = self.view;
+  while (parent) {
+    parent = parent.reactSuperview;
+    if (parent.isReactRootView) {
+      return parent.reactViewController;
+    }
+  }
+  return nil;
 }
 
 - (BOOL)isModalWithHeader
--- a/ios/RNSScreenStack.mm
+++ b/ios/RNSScreenStack.mm
@@ -62,27 +62,6 @@ namespace react = facebook::react;
 
 @implementation RNSNavigationController
 
-#if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && defined(__IPHONE_26_0) && \
-    __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_26_0
-- (void)viewDidLoad
-{
-  // iOS 26 introduces new gesture recognizer which replaces our RNSPanGestureRecognizer.
-  // The problem is that we are not able to handle it here for various reasons:
-  // - the new recognizer comes with its own delegate and our current approach is to wire
-  //   all recognizers to RNSScreenStackView; to be 100% sure we don't break the logic,
-  //   we would have to decorate its delegate and call it after our code, which would
-  //   break other recognizers that the stack view is the delegate for
-  // - when RNSScreenStackView.setupGestureHandler method is called, the recognizer hasn't been
-  //   loaded yet and there is no other place to configure in a not "hacky" way
-  // - the official docs warn us to not use it for anything other than "setting up failure requirements with it"
-  // - we expose fullScreenGestureEnabled prop to enable/disable the feature,
-  //   so we need control over the delegate
-  if (@available(iOS 26.0, *)) {
-    self.interactiveContentPopGestureRecognizer.enabled = NO;
-  }
-}
-#endif // iOS 26
-
 #if !TARGET_OS_TV
 - (UIViewController *)childViewControllerForStatusBarStyle
 {
@@ -220,51 +199,6 @@ namespace react = facebook::react;
   return false;
 }
 
-#pragma mark - UINavigationBarDelegate
-
-#if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && defined(__IPHONE_26_0) && \
-    __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_26_0
-- (BOOL)navigationBar:(UINavigationBar *)navigationBar shouldPopItem:(UINavigationItem *)item
-{
-  if (@available(iOS 26, *)) {
-    // To prevent popping multiple screens when back button is pressed repeatedly,
-    // We allow for pop operation to proceed only if no transition is in progress,
-    // which we check indirectly by checking if transitionCoordinator is set.
-    // If it's not, we are safe to proceed.
-    if (self.transitionCoordinator == nil) {
-      // We still need to disable interactions for back button so click effects are not applied,
-      // and there is unfortunately no better place for it currently
-      UIView *button = [navigationBar rnscreens_findBackButtonWrapperView];
-      if (button != nil) {
-        button.userInteractionEnabled = false;
-      }
-
-      return true;
-    }
-
-    return false;
-  }
-
-  return true;
-}
-
-- (void)navigationBar:(UINavigationBar *)navigationBar didPopItem:(UINavigationItem *)item
-{
-  if (@available(iOS 26, *)) {
-    // Reset interactions on back button -> see navigationBar:shouldPopItem
-    // IMPORTANT: This reset won't execute when preventNativeDismiss is on.
-    // However, on iOS 26, unlike in previous versions, the back button instance changes
-    // when handling preventNativeDismiss and userIteractionEnabled is reset.
-    // The instance also changes when regular screen pop happens, but in that case
-    // the value of userInteractionEnabled is carried on, and we reset it here.
-    UIView *button = [navigationBar rnscreens_findBackButtonWrapperView];
-    if (button != nil) {
-      button.userInteractionEnabled = true;
-    }
-  }
-}
-#endif // Check for iOS >= 26
-
 #pragma mark - RNSFrameCorrectionProvider
 
 #ifdef RNS_GAMMA_ENABLED
@@ -329,7 +263,7 @@ namespace react = facebook::react;
   UINavigationController *_controller;
   NSMutableArray<RNSScreenView *> *_reactSubviews;
   BOOL _invalidated;
-  BOOL _isFullWidthSwiping;
+  BOOL _isFullWidthSwipingWithPanGesture;
   RNSPercentDrivenInteractiveTransition *_interactionController;
   __weak RNSScreenStackManager *_manager;
   BOOL _updateScheduled;
@@ -524,6 +458,9 @@ RNS_IGNORE_SUPER_CALL_END
         [self addSubview:controller.view];
 #if !TARGET_OS_TV
         _controller.interactivePopGestureRecognizer.delegate = self;
+        if (@available(iOS 26, *)) {
+          _controller.interactiveContentPopGestureRecognizer.delegate = self;
+        }
 #endif
         [controller didMoveToParentViewController:parentView.reactViewController];
         // On iOS pre 12 we observed that `willShowViewController` delegate method does not always
@@ -945,7 +882,7 @@ RNS_IGNORE_SUPER_CALL_END
       // when preventing the native dismiss with back button, we have to return the animator.
       // Also, we need to return the animator when full width swiping even if the animation is not custom,
       // otherwise the screen will be just popped immediately due to no animation
-      ((operation == UINavigationControllerOperationPop && shouldCancelDismiss) || _isFullWidthSwiping ||
+      ((operation == UINavigationControllerOperationPop && shouldCancelDismiss) || _isFullWidthSwipingWithPanGesture ||
        [RNSScreenStackAnimator isCustomAnimation:screen.stackAnimation] || _customAnimation)) {
     return [[RNSScreenStackAnimator alloc] initWithOperation:operation];
   }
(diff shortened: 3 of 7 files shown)


guides/GUIDE_FOR_LIBRARY_AUTHORS.md
```diff
@@ -55,6 +55,7 @@ Defaults to `false`. When `enableFreeze()` is run at the top of the application
 ### `fullScreenSwipeEnabled` (iOS only)
 
 Boolean indicating whether the swipe gesture should work on whole screen. Swiping with this option results in the same transition animation as `simple_push` by default. It can be changed to other custom animations with `customAnimationOnSwipe` prop, but default iOS swipe animation is not achievable due to usage of custom recognizer. Defaults to `false`.
+IMPORTANT: Starting from iOS 26, full screen swipe to dismiss is the native behavior, and this prop is ignored.
```
kligarski: Let's create an issue on our board to update `react-navigation` docs for `fullScreenSwipeEnabled` (I don't think that `gestureResponseDistance` is available via `react-navigation`?)

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
```diff
@@ -55,6 +55,7 @@ Defaults to `false`. When `enableFreeze()` is run at the top of the application
...
 
 ### `fullScreenSwipeShadowEnabled` (iOS only)
```
kligarski, line 60: What about this prop? Is it impacted by the changes in this PR?

ios/RNSScreen.mm
```diff
@@ -1466,6 +1466,12 @@ - (instancetype)initWithView:(UIView *)view
 // TODO: Find out why this is executed when screen is going out
 - (void)viewWillAppear:(BOOL)animated
 {
+  if (@available(iOS 26, *)) {
+    // In iOS 26, as soon as another screen appears in transition, it is interactable
+    // To avoid glitches resulting from clicking buttons mid transition, we temporarily disable all interactions
+    // Because nested stack exist, we need to find a common top view for the whole app
```
kligarski:
```suggestion
    // Disabling interactions for parent navigation controller won't be enough in case of nested stack,
    // that's why we need to find a common top view for the whole app
```
I think `Because nested stack exist`, `Because nested stacks exist`, `Because nested stack exists` all sound weird, I suggest that we rephrase it entirely.

ios/RNSScreen.mm
```diff
@@ -1530,6 +1536,10 @@ - (void)viewWillDisappear:(BOOL)animated
 
 - (void)viewDidAppear:(BOOL)animated
 {
+  if (@available(iOS 26, *)) {
+    // Reenable interactions, see viewWillAppear
+    [self findReactRootViewController].view.userInteractionEnabled = true;
+  }
```
kligarski: I'm thinking, is it possible that `viewWillAppear` will run but `viewDidAppear` will not? Did you check how `prevent remove` works now?

ios/RNSScreenStack.mm
```diff
@@ -329,7 +263,7 @@ @implementation RNSScreenStackView {
   UINavigationController *_controller;
   NSMutableArray<RNSScreenView *> *_reactSubviews;
   BOOL _invalidated;
-  BOOL _isFullWidthSwiping;
+  BOOL _isFullWidthSwipingWithPanGesture; // used only for content swipe with RNSPanGestureRecognizer
```
kligarski, line 266:
```suggestion
  BOOL _isFullWidthSwipingWithCustomRecognizer; // used only for content swipe with RNSPanGestureRecognizer
```
not sure here but maybe the name should highlight that a custom gesture recognizer is used.

src/native-stack/types.tsx
```diff
@@ -160,6 +160,8 @@ export type NativeStackNavigationOptions = {
    * It can be changed to other custom animations with `customAnimationOnSwipe` prop, but default iOS swipe animation is not achievable due to usage of custom recognizer.
    * Defaults to `false`.
    *
+   * IMPORTANT: Starting from iOS 26, full screen swipe to dismiss is the native behavior, and this prop is ignored.
+   *
```
kligarski: We should update docs in `src/types.tsx` instead of old stack (`src/native-stack/types.tsx`).

ios/RNSScreenStack.mm
```diff
@@ -969,23 +909,44 @@ - (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
   }
   RNSScreenView *topScreen = _reactSubviews.lastObject;
 
+  BOOL isCustomAnimation =
+      topScreen.customAnimationOnSwipe && [RNSScreenStackAnimator isCustomAnimation:topScreen.stackAnimation];
+
+  BOOL usePanGestureRecognizerAsContentPopGestureRecognizer = true;
+  if (@available(iOS 26, *)) {
+    // case
```
kligarski: not sure what this comment mean?

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
kmichalikk: Done

ios/RNSScreenStack.mm
kmichalikk: I'd say this depends on from which angle you look at it, the current name makes sense because this recognizer is referenced everywhere as PanGesture.

src/native-stack/types.tsx
kmichalikk: right, my bad

ios/RNSScreenStack.mm
kmichalikk: this code block is a leftover from prev attempt, removing it

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
kmichalikk: I've confirmed that it should be, so this + parts of code needs fixing

ios/RNSScreen.mm
kmichalikk: I tried this on Test2125 which demos preventRemove

When using back button with prevent remove, the calls look like this:
image
We are covered by SecondScreen going back on top

When using swipe and not dismissing fully:
image
It's the same

When using swipe and actually dismissing SecondScreen (preventRemove still on):
image
Which is different ( 🤔 ) but works for our case

and without preventRemove it's like this:
image

so in every case at least one of the screens goes through the whole process

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
```diff
@@ -55,12 +55,14 @@ Defaults to `false`. When `enableFreeze()` is run at the top of the application
...
 
 Boolean indicating whether the swipe gesture should work on whole screen. Swiping with this option results in the same transition animation as `simple_push` by default. It can be changed to other custom animations with `customAnimationOnSwipe` prop, but default iOS swipe animation is not achievable due to usage of custom recognizer. Defaults to `false`.
+IMPORTANT: Starting from iOS 26, full screen swipe is handled by native recognizer, and this prop is ignored.
 
 ### `fullScreenSwipeShadowEnabled` (iOS only)
 
 Boolean indicating whether the full screen dismiss gesture has shadow under view during transition. The gesture uses custom transition and thus
 doesn't have a shadow by default. When enabled, a custom shadow view is added during the transition which tries to mimic the
 default iOS shadow. Defaults to `true`.
+IMPORTANT: Starting from iOS 26, full screen swipe is handled by native recognizer, except for custom animations, where the value is always `true` to best fit native behavior.
```
kligarski: I don't think this is clear enough - we should just state that the value of this prop is ignored starting from iOS 26 and `true` is used.

kligarski: Same applies to `src/types.tsx`

ios/RNSScreen.mm
```diff
@@ -1469,9 +1469,11 @@ - (void)viewWillAppear:(BOOL)animated
...
     // In iOS 26, as soon as another screen appears in transition, it is interactable
     // To avoid glitches resulting from clicking buttons mid transition, we temporarily disable all interactions
-    // Disabling interactions for parent navigation controller won't be enough in case of nested stack;
-    // we need to find a common top view for the whole app
-    [self findReactRootViewController].view.userInteractionEnabled = false;
+    // Disabling interactions for parent navigation controller won't be enough in case of nested stack
+    // Furthermore, a stack put inside a modal will exist in an entirely different hierarchy
+    // To be sure, we block interactions on the whole window
+    // We need to get the window instance from parent view because it is not set until the view has appeared
+    self.parentViewController.view.window.userInteractionEnabled = false;
```
kligarski: Please check one more thing - when you have some push screens and opened modal, when you dismiss the modal and quickly click the back button, will it work correctly (I'm not sure if `viewWillAppear` will be called)?

kligarski (approved):

ios/RNSScreen.mm
kligarski: And another thing - when I tested main (previous solution), this happened:

https://github.com/user-attachments/assets/a15a229c-07df-4311-8cb0-5f423e109e0f

Please check if this solution prevents this but I managed to also reproduce the bug on iOS 18. If this isn't fixed by this PR, let's create a ticket to investigate this.

kmichalikk: > when you dismiss the modal and quickly click the back button

It actually broke again in situation: stack1_screenA / stack2_screen A / stack2_screen B / stack2_modal C; when quickly dismissing modal and screen, the whole stack was dismissed. I managed to fix that, but it's not pretty. The main challenge was to find a reference to correct window without using sharedApplication.

As for dismissing multiple modals, native UIKit prevents dismissing more than 2 at once, blocking the third one mid swipe. I ended up blocking every consecutive dismissal similar to regular screens.

> And another thing - when I tested main (previous solution), this happened:

I didn't notice anything on my PR

kmichalikk:
Paper appears to work, too.

One other thing I found is that when UIWindow has interactions disabled and we still try to input gestures, this log is shown

> Unexpected nil window in latent system gesture client update: windowServerHitTestWindow: <UIWindow: 0x101b05640; frame = (0 0; 402 874); autoresize = W+H; gestureRecognizers = <NSArray: 0x600001829800>; layer = <UIWindowLayer: 0x60000171f100>>, touch: <UITouch: 0x10e788d40> type: Direct; phase: Stationary; is pointer: NO; tap count: 1; force: 0.000; window: (null); responder: (null);

I suspect that's because there is nothing above the window to handle the gesture, but I don't know really. There is not that much info online, other that people having this error when app breaks with black screen, which doesn't happen here.

kligarski (approved):

ios/RNSScreenStack.mm
```diff
@@ -522,6 +458,11 @@ - (void)reactAddControllerToClosestParent:(UIViewController *)controller
         [self addSubview:controller.view];
 #if !TARGET_OS_TV
         _controller.interactivePopGestureRecognizer.delegate = self;
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
+        if (@available(iOS 26, *)) {
+          _controller.interactiveContentPopGestureRecognizer.delegate = self;
```
kkafar, line 463: I'm kinda unsure about this line. Is the delegate `nil` before we assign `self`? Or do we override some behaviour? 

[The docs](https://developer.apple.com/documentation/uikit/uinavigationcontroller/interactivecontentpopgesturerecognizer?language=objc) explicitly mention to not do anything but set up failure requirements with this little one.

ios/RNSScreenStack.mm
```diff
@@ -967,23 +908,38 @@ - (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
   }
   RNSScreenView *topScreen = _reactSubviews.lastObject;
 
+  BOOL isCustomAnimation =
```
kkafar:
```suggestion
  BOOL usesCustomAnimationAndInteractiveShouldMatch =
```

Don't like even my suggestion. It would be nice to come up with something sensible. `isCustomAnimation` is misleading, as this value conveys a bit different information.

ios/RNSScreenStack.mm
```diff
@@ -967,23 +908,38 @@ - (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
...
-      return YES;
+    // On iOS < 26, we have a custom full screen swipe recognizer that functions similarily
+    // to interactiveContentPopGestureRecognizer introduced in iOS 26.
+    // On iOS >= 26, we want to use the native one, but we are unable to handle custom animations
+    // with native interactiveContentPopGestureRecognizer, so we have to fallback to the old implementation in this one
+    if (@available(iOS 26, *)) {
+      if (isCustomAnimation) {
+        _isFullWidthSwipingWithPanGesture = YES;
+        [self cancelTouchesInParent];
+        return YES;
```
kkafar, line 930: We fallback into legacy logic here, but why don't we use `isInGestureResponseDistance`? Do we just ignore it? Is it intended?

ios/RNSScreenStack.mm
```diff
@@ -1249,10 +1199,29 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldReceive
     return NO;
   }
 
+  BOOL isCustomAnimation =
```
kkafar: same here. You might wanna extract it to dedicated method.

kkafar:
## Header back button is exchanged "too quickly"

I do not know yet whether this is a native behaviour (UIKit bug) or maybe intended behaviour, however it looks ugly. 

| Long text | Regular text |
| ---- | ---- |
| <video src="https://github.com/user-attachments/assets/3792f5e6-12c6-464f-8a5d-ff53a77d65da" alt="" /> | <video src="https://github.com/user-attachments/assets/9d9c876c-94ec-4981-8cb3-f8012d8a72e4" alt="" /> |

kkafar:
## White background is visible on fast swipe


https://github.com/user-attachments/assets/e7f3e43c-fe26-4614-a2fa-fb4c1a875c7b

This is most likely because third & further screens on the stack are frozen or detached.

If they're frozen - we might be able to fix it. If they're detached - then it is UIKit who should attach them :/

ios/RNSScreenStack.mm
kmichalikk: Yes they do, but sadly they don't explain anything and give no alternatives. On native UIKit test app, the gesture doesn't seem to work without having the delegate set to our controller, but that could be my lack of knowledge about the interactions in native UIKit.

Setting the delegate is necessary to hook the gesture into our logic, have it stop on `gestureEnabled: false`, etc., and it doesn't work here without setting the delegate either.

For regular edge swipe gesture, we have the same call just above this one, and it, too, doesn't work without it.

I also didn't notice any side effects

ios/RNSScreenStack.mm
kmichalikk: AFAIK it's intended. On iOS 26, we want the new gesture to work on the whole screen. The old one is kept _only_ for when user sets custom animations, because we can't animate the native one easily. And in this scenario, the old one should behave as close as the new native one, having only the difference in animation, and without any customization that is exclusive for it.

kmichalikk:
Re: "White background is visible on fast swipe"

This is a native "rubberband" effect, the white color is actually the background color of `RCTSurfaceHostingProxyRootView`. We can follow up with a way to change it, or - as pointed out - there could be another view inserted which maybe? (needs confirmation) will be probed for the color.

ios/RNSScreenStack.mm
kmichalikk: 35f55953a9f40b77cf8c20d90d1686ae1d0618cf

ios/RNSScreenStack.mm
kmichalikk: 35f55953a9f40b77cf8c20d90d1686ae1d0618cf

kmichalikk: It's so little code in two places that it's not worth it to extract method IMO

ios/RNSScreenStack.mm
kkafar: Okay. That makes sense & complies with weekly conclusions. My suggestion would be to put this reasoning as a comment inside the code. When reading the code I found it "weird" to ignore the `gestureResponseDistance` in that case.

ios/RNSScreenStack.mm
kkafar: thats fine

ios/RNSScreenStack.mm
kkafar: Okay, I see. Thank you for explanation. 

I'll want to revisit this logic one more time before we land it (I'll do it today).

ios/RNSScreenStack.mm
kmichalikk: 84303f82e82dff66c760149fd1def9ddc2d11519

kkafar:
Okay, I've found out what makes it work w/o setting the delegate explicitly. This method:

```objective-c

- (id<UIViewControllerAnimatedTransitioning>)navigationController:(UINavigationController *)navigationController
                                  animationControllerForOperation:(UINavigationControllerOperation)operation
                                               fromViewController:(UIViewController *)fromVC
                                                 toViewController:(UIViewController *)toVC
{ ... }
```

In case where it is **NOT** overwritten, the gestures do work out of the box. If the method is overwritten & we return `nil` (the default case, haven't tried returning an animator yet; that's something to try tomorrow), the gesture recognizers won't work w/o setting the delegates. 

We can't really resign from overwriting this method, as it's only entry point for animation customisation ==> therefore we're forced to overwrite the delegates. 


Let me just note, that this is obviously undocumented UIKit behaviour.

JustJoostNL:
It seems like this change breaks horizontal scrollviews. Is there anything we can do to avoid this?

kmichalikk:
Thanks for info, I'll look into it ASAP.

huextrat:
`fullScreenSwipeEnabled` should works on iOS 26 / Xcode 26 thanks to this? 🤔 

As on my side, full screen swipe do nothing with iOS 26 / Xcode 26 / v4.17.1

kkafar:
Will check this in 1h & get back to you @huextrat

huextrat:
> Will check this in 1h & get back to you @huextrat

I have a repro here : https://github.com/huextrat/repro-screens-fullscreen-swipe

When there is a `ScrollView`, the fullscreen swipe do not work anymore, it was before v4.17.x
This is not related to Xcode 26 and I can reproduce on Xcode 16 also

terrysahaidak:
I can confirm that full screen swipe does not work when you try to swipe inside a vertical scroll view.

PR #3179, opened 2025-08-29
refactor: review 3097 - make split screen controller & component the frame correction provider

kkafar:
## Description

Retrospective review of #3097

## Changes

- **Give register & unregister functions more descriptive names**
  * -[RNSSplitViewScreenComponentView registerForFrameUpdates:] -> -[RNSSplitViewScreenComponentView registerForFrameCorrection:]
  * -[RNSSplitViewScreenComponentView unregisterFromFrameUpdates:] -> -[RNSSplitViewScreenComponentView unregisterFromFrameCorrection:]

  I think this has more sense, as this isn't a regular frame update but
  rather a correction related to `FrameCORRECTOR` &
  `FrameCORRECTIONProvider`.

- **Add missing stars in multiline comment**

- **Refactor the logic according to review**

  * Make `RNSSplitViewScreenComponentView` implement
    `RNSFrameCorrectionProvider`, as it does provide the correct frame for
    the navigation controller,
  * update method signatures in `RNSFrameCorrectionProvider` protocol, so
    that they accept appropriate view parameter,
  * rename `_viewsForFrameUpdate` -> `_viewsForFrameCorrection`,
  * rename registering methods.

- **Do not attempt to register when parent controller is nil**

  Doing so does not make sense, as we will silently fail when looking for
  parent. I prefer to handle this case explicitly.
  
- **Rename findAncestorSplitViewScreenController -> findAncestorFrameCorrectionProvider**

See #3097

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStack.h
+++ b/ios/RNSScreenStack.h
@@ -12,10 +12,6 @@
 #import "RNSOrientationProviding.h"
 #endif // !TARGET_OS_TV
 
-#ifdef RNS_GAMMA_ENABLED
-#import "RNSFrameCorrectionProvider.h"
-#endif // RNS_GAMMA_ENABLED
-
 NS_ASSUME_NONNULL_BEGIN
 
 @interface RNSNavigationController : UINavigationController <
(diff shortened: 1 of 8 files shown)


kkafar:
Converted it to draft, due to following error: 

image

t0maboro (approved):

PR #3186, opened 2025-09-01
feat!(iOS, Stack v4): change search bar related prop defaults

kligarski:
## Description

Changes default values for search bar related props.

### `hideWhenScrolling`
On Fabric, for some reason, default value for `hideWhenScrolling` is `false`, even though:
- on Paper, it is `true`, because we initialize `_hideWhenScrolling` to `YES`,
- in `GUIDE_FOR_LIBRARY_AUTHORS`, react-navigation's docs, it is stated that the default value for this prop is `true`,
- UIKit default is `true`.

This PR fixes this.

### `placement`
To preserve backward compatibility when adding `placement` as a prop, the default value for `placement` became `stacked`. Starting from iOS 26, new search bar placements has been added (`integrated`, ...). In order to make it even easier for users to get *native feel* in their apps, we decided to change the default value of this prop to `automatic`.

> [!WARNING]
>
> This seems like a **BREAKING** change from users perspective, but we consider it just as a alignment with native behaviour - iOS 26 just changes things.

## Changes

- change default value of `hideWhenScrolling` to `true` on Fabric
- change default value of `placement` to `'automatic'`
- adjust test screens to use previous default values (I tried to use `hideWhenScrolling` where necessary)

Run tests related to search bar.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/apps/src/tests/Test1683.tsx
+++ b/apps/src/tests/Test1683.tsx
@@ -47,6 +47,8 @@ const Navigator = () => {
           component={HomeScreen}
           options={{
             headerSearchBarOptions: {
+              // Added in https://github.com/software-mansion/react-native-screens/pull/3186
+              // to preserve test's original search bar configuration.
               placement: 'stacked',
             },
             // statusBarStyle: 'light',
--- a/apps/src/tests/Test1683.tsx
+++ b/apps/src/tests/Test1683.tsx
@@ -46,7 +46,9 @@ const Navigator = () => {
           name="Home"
           component={HomeScreen}
           options={{
-            headerSearchBarOptions: {},
+            headerSearchBarOptions: {
+              placement: 'stacked',
+            },
             // statusBarStyle: 'light',
           }}
         />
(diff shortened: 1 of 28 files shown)


apps/src/tests/Test1683.tsx
```diff
@@ -46,7 +46,9 @@ const Navigator = () => {
           name="Home"
           component={HomeScreen}
           options={{
-            headerSearchBarOptions: {},
+            headerSearchBarOptions: {
+              placement: 'stacked',
```
kkafar, line 52: Add comment that this is added for compat reasons & link to this PR.

kkafar (approved):
One note. 

Let's hightlight this PR in release notes, so that the people are aware of the changes.

apps/src/tests/Test1683.tsx
kligarski: https://github.com/software-mansion/react-native-screens/pull/3186/commits/f5136796a7b78d199513c900a26420b423c1a959

kkafar (approved):
Looks good!

PR #3189, opened 2025-09-02
fix(iOS): Prevent delay when dismissing transparentModals

### LINKED ISSUE #3074: [createNativeStackNavigator] animationDuration not work in transparentmodal and it's slow on closing
### Description

I am using **createNativeStackNavigator** for my expo project navigation. I have Two stack group for navigation, one for regular screen and other for transparentmodal. i opened a transparent modal using **none animation type**, it opened fast but when **closing it get slower**. seems it need time to close.

### Steps to reproduce

I have tryed changing animation type to fade, slide from bottom. when using slide from bottom animation type it working perfectly on closing. But for none, fade it's slow.

I also change the animationDuration time. seems it not work for transparentmodal. 

https://github.com/aargon007/transparent-modal-demo

### Screens version

4.11.1

### React Native version

0.79.5

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

None

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

Real device

### Device model

iPad Air 2

### Acknowledgements

Yes

kmichalikk:
## Description

Closes #3074.

This PR removes the delay when dismissing fullscreen modal-like screens. This behavior can be observed on native apps, but makes things look as if they are broken. We also observe that modal screens don't take transitionDuration into account. We don't plan on supporting it now, but might look into it in the future. Docs are updated to reflect the current state.

## Testing

Use Test3074. It has 4 screens in stack - 1. and 2. are a regular screen, 3. & 4. are transparentModals. The last one should dismiss immediately without animation. Revert changes in RNSScreen to see the bug. Checking Debug/Slow animations in simulator makes it even more apparent.

### DIFF
--- a/apps/src/tests/Test3074.tsx
+++ b/apps/src/tests/Test3074.tsx
@@ -1,0 +1,92 @@
+import React, { useEffect, useRef } from 'react';
+import { NavigationContainer, ParamListBase } from '@react-navigation/native';
+import {
+  NativeStackNavigationProp,
+  createNativeStackNavigator,
+} from '@react-navigation/native-stack';
+import { Button, Text, View, Animated, useAnimatedValue } from 'react-native';
+import Colors from '../shared/styling/Colors';
+import { useTransitionProgress } from 'react-native-screens';
+
+type StackRouteParamList = {
+  Screen1: undefined;
+  Screen2: undefined;
+  TModal1: undefined;
+  TModal2: undefined;
+};
+
+type NavigationProp<ParamList extends ParamListBase> = {
+  navigation: NativeStackNavigationProp<ParamList>;
+};
+
+type StackNavigationProp = NavigationProp<StackRouteParamList>;
+
+const Stack = createNativeStackNavigator<StackRouteParamList>();
+
+function makeScreen(route: keyof StackRouteParamList, navigateTo: keyof StackRouteParamList | undefined, color: string, uiOffset: number) {
+  return function Screen({ navigation }: StackNavigationProp) {
+    const { progress: animatedProgress, closing: animatedClosing, goingForward: animatedGoingForward } = useTransitionProgress();
+
+    const progress = animatedProgress.interpolate({ inputRange: [0, 1], outputRange: [1, 200] });
+    const closing = animatedClosing.interpolate({ inputRange: [0, 1], outputRange: [1, 200] });
+    const goingForward = animatedGoingForward.interpolate({ inputRange: [0, 1], outputRange: [1, 200] });
+
+    animatedProgress.addListener(v => console.log(`progress ${route} = ${Math.round(v.value * 100) / 100}`));
+    animatedClosing.addListener(v => console.log(`closing ${route} = ${v.value}`));
+    animatedGoingForward.addListener(v => console.log(`goingForward ${route} = ${v.value}`));
+
+    const spinValue = useAnimatedValue(0);
+    useEffect(() => {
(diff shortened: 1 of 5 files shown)


ios/RNSScreen.mm
```diff
@@ -1716,18 +1716,25 @@ - (void)setupProgressNotification
...
-          [self->_animationTimer setPaused:YES];
-          [self->_animationTimer invalidate];
-          [self->_fakeView removeFromSuperview];
-        }];
+        animateAlongsideTransition:animation
+                        completion:^(id<UIViewControllerTransitionCoordinatorContext> _Nonnull context) {
+                          [self->_animationTimer setPaused:YES];
+                          [self->_animationTimer invalidate];
+                          [self->_fakeView removeFromSuperview];
+                        }];
```
kkafar: If the `self.transitionCoordinator.isAnimated == NO` we do not need to schedule anything  right? Why don't we just do it? Return early when this condition holds. Do not setup animation & do not schedule completion callback with transition coordinator.

src/types.tsx
```diff
@@ -495,7 +495,7 @@ export interface ScreenProps extends ViewProps {
   swipeDirection?: SwipeDirectionTypes;
   /**
    * Changes the duration (in milliseconds) of `slide_from_bottom`, `fade_from_bottom`, `fade` and `simple_push` transitions on iOS. Defaults to `500`.
-   * The duration of `default` and `flip` transitions isn't customizable.
+   * For screens with `default` and `flip` transitions, and, as of now, for screens with `presentation` set to `modal`, `formSheet`, `pageSheet` (regardless of transition), the duration isn't customizable.
```
kkafar, line 498: It's fine this time, but I would prefer it being done in separate PR.

apps/src/tests/Test3074.tsx
kkafar, line 1: I find this test vastly too complicated. I wonder why do we have "freeze testing component" there & all this progress stuff & no word about it in PR description. 

Can we either justify it's presence in the PR description or simplify the test?

What do you think?

apps/src/tests/Test3074.tsx
```diff
@@ -0,0 +1,92 @@
...
+        </Stack.Navigator>
+      </NavigationContainer>
+      <Text style={{ position: 'absolute', marginTop: 450 }}>
+        The stack above consists of 2 screens and 2 transparent fullscreen modals.
+        When all of them are pushed on stack, Screen1 covers Screen2, and Screen2 is visible below both modals.
+        Notice the difference in transitions when clicking "Go Back", using header back button, and swiping.
+        {"\n\n"}
+        When using "Go back", screens will not be responsive, because React removes it before host is finished. Use back button to see the progress. Modals can only be dismissed using "Go Back",
+        so use logs to observe the changes.
+      </Text>
```
kkafar, line 89: I like this description though.

kkafar:
Okay, I've found, that we emit 0.0 & 1.0 transition progress in `viewWillAppear` & `viewDidAppear`. That's fine then. We do not have to do anything more here.

kkafar:
I've opened diff against this PR. https://github.com/software-mansion/react-native-screens/pull/3205. If you want, land it or apply these manually / start discussion about these changes.

ios/RNSScreen.mm
kmichalikk: Applied your changes & it seems to be working, thanks

apps/src/tests/Test3074.tsx
kmichalikk: tl;dr I needed it to make sure I don't break anything with animation progress

PR #3198, opened 2025-09-05
fix(iOS 26, Stack v4): fix workaround for RNSScreen being partially obstructed by UINavigationBar on iOS 26 Paper

kligarski:
## Description

Fixes the workaround from https://github.com/software-mansion/react-native-screens/pull/3111 on Paper.

On Paper, in debug mode, sometimes there are some views between `RNSScreenView` and `RNSScreenContentWrapper`. This has been an issue before: https://github.com/software-mansion/react-native-screens/pull/2683.

To make the workaround work correctly, we look for `RNSScreenContentWrapper` in first descendant chain, when in debug mode on Paper.

Fixes https://github.com/software-mansion/react-native-screens-labs/issues/412.

## Changes

- look for `RNSScreenContentWrapper` in first descendant chain, when in debug mode on Paper

Run `Test3111` on Paper in debug mode.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -571,24 +571,14 @@ RNS_IGNORE_SUPER_CALL_END
   // See comment above _safeAreaConstraints declaration for reason why this is necessary.
   RNSScreenContentWrapper *contentWrapper = nil;
   if (@available(iOS 26, *)) {
-#if !RCT_NEW_ARCH_ENABLED && !defined(NDEBUG)
-    // On Paper in debug mode, there are some views between RNSScreenView and RNSScreenContentWrapper.
     UIView *currentView = vc.view;
-    while (currentView != nil) {
+    while ([currentView.subviews count] > 0) {
+      currentView = currentView.subviews[0];
       if ([currentView isKindOfClass:RNSScreenContentWrapper.class]) {
         contentWrapper = static_cast<RNSScreenContentWrapper *>(currentView);
         break;
-      } else if ([currentView.subviews count] > 0) {
-        currentView = currentView.subviews[0];
-      } else {
-        break;
       }
     }
-#else // !RCT_NEW_ARCH_ENABLED && !defined(NDEBUG)
-    if (vc.view.subviews.count > 0 && [vc.view.subviews[0] isKindOfClass:[RNSScreenContentWrapper class]]) {
-      contentWrapper = static_cast<RNSScreenContentWrapper *>(vc.view.subviews[0]);
-    }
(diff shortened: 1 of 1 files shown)


ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -571,9 +571,24 @@ + (void)updateViewController:(UIViewController *)vc
...
+    while (currentView != nil) {
+      if ([currentView isKindOfClass:RNSScreenContentWrapper.class]) {
+        contentWrapper = static_cast<RNSScreenContentWrapper *>(currentView);
+        break;
+      } else if ([currentView.subviews count] > 0) {
+        currentView = currentView.subviews[0];
+      } else {
+        break;
+      }
+    }
```
kligarski: This logic is similar to `RNSScrollViewFinder`, we could extract it to separate class but we would probably need to cast the result in places where we would use this function or provide wrappers. I also think that RNSScrollViewFinder might soon need to use more complicated logic. Let me know what you think.

kligarski: We can also use this logic for all cases, not only for Paper in debug mode, but it seemed like an overkill for getting first child.

kmichalikk: We could also just do:
```c
RNSScreenContentWrapper *contentWrapper = nil;
UIView *currentView = vc.view;
while ([currentView.subviews count] > 0) {
    currentView = currentView.subviews[0];
    if ([currentView isKindOfClass:RNSScreenContentWrapper.class]) {
        contentWrapper = static_cast<RNSScreenContentWrapper *>(currentView);
        break;
    }
}
```
performance difference will be negligible (especially if we *expect* the class to be correct) and it improves readability.

kmichalikk (commented):
One issue I found

image

on Paper, ScrollView overflows below the screen

ios/RNSScreenStackHeaderConfig.mm
kligarski: Nice, thank you!

https://github.com/software-mansion/react-native-screens/pull/3198/commits/55ac950f9e3496f98356d31259af0887cb2eada6

kligarski:
> One issue I found
> image
> 
> on Paper, ScrollView overflows below the screen

It also happens on Fabric. I created a separate issue for this: https://github.com/software-mansion/react-native-screens-labs/issues/413.

kligarski:
We decided to remove the workaround as it introduces new bugs (https://github.com/software-mansion/react-native-screens/pull/3206).

PR #3207, opened 2025-09-10
fix(iOS): Reorder back button setup logic so iOS renders it correctly when displayMode=minimal

### LINKED ISSUE #2991: [iOS] Header Title Misaligned When First Stack Screen Has No Native Header
### Description

When using a native stack navigator in React Native, if the **first screen in the stack has the native header hidden** (e.g., `headerShown: false`), the **header title on subsequent screens appears misaligned or not centered**. This only happens when navigating from a screen without a native header to one that uses the default header.

https://github.com/user-attachments/assets/812931b1-2bdb-4074-b9e3-8e6f2142dcd8

I think this might be related to view recycling. It was also happening on old architecture a couple of months ago while I was testing.

### Steps to reproduce

1. Create a stack navigator with two screens:

   * Screen A: `headerShown: false`
   * Screen B: default header shown
2. Navigate from Screen A to Screen B.
3. Observe that the header title on Screen B is not centered.

Make sure you cold start the app with header shown set to false for Screen A.

https://snack.expo.dev/M-RIitAjPdl7xsK7AKYbQ

### Screens version

4.11.1

### React Native version

0.79.5

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

iPhone 16

### Acknowledgements

Yes

kmichalikk:
## Description

Partially fixes #2991. Unfortunately, not all cases are covered; see below.

When navigating from screen with headerShown: false to a screen that has a back button with displayMode=minimal, the width of the button appears to be the same as if it was meant for the first screen, with space allocated for a label that is not visible.

## Changes

Simply moving the line that updates the displayMode after the rest of back button's setup code, solves the problem for apps which use displayMode=minimal.

### Before

https://github.com/user-attachments/assets/d9abc075-ba47-44d9-a350-78f7199b64cc

### After

https://github.com/user-attachments/assets/149044b3-572b-451c-bdbb-a5a5331e9190

## What still needs fixing

When mixing headerShown=false, displayMode=default and displayMode=minimal, the screen with displayMode=minimal still doesn't work. For now, please use workarounds suggested by the community.

https://github.com/user-attachments/assets/68a864f7-a8f1-4567-8df7-d2de27e5d282

### DIFF
--- a/apps/App.tsx
+++ b/apps/App.tsx
@@ -1,11 +1,11 @@
 import React from 'react';
 import { enableFreeze } from 'react-native-screens';
-import Example from './Example';
-// import * as Test from './src/tests';
+// import Example from './Example';
+import * as Test from './src/tests';
 
 enableFreeze(true);
 
 export default function App() {
-  return <Example />;
-  // return <Test.TestBottomTabs />;
+  // return <Example />;
+  return <Test.Test2991 />;
 }
--- a/apps/src/tests/Test2991.tsx
+++ b/apps/src/tests/Test2991.tsx
@@ -1,0 +1,74 @@
+import React from 'react';
+import { NavigationContainer, ParamListBase } from '@react-navigation/native';
+import {
+  NativeStackNavigationProp,
+  createNativeStackNavigator,
+} from '@react-navigation/native-stack';
+import { Button, ScrollView } from 'react-native';
+
+type RouteParamList = {
+  Home: undefined;
+  Second: undefined;
+  Third: undefined;
+};
+
+type NavigationProp<ParamList extends ParamListBase> = {
+  navigation: NativeStackNavigationProp<ParamList>;
+};
+
+type StackNavigationProp = NavigationProp<RouteParamList>;
+
+const Stack = createNativeStackNavigator<RouteParamList>();
+
+const Home = ({ navigation }: StackNavigationProp) => <>
+  <ScrollView contentInsetAdjustmentBehavior='automatic'>
+    <Button title="Open second" onPress={() => navigation.navigate('Second')} />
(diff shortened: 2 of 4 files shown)


kmichalikk:
It looked fine-ish until I tried the other way around, from first to third, to first, to second. Now second is always minimal. I'd say it's not an improvement then. Converting it to Draft.

kmichalikk:
This is superseded by https://github.com/software-mansion/react-native-screens/pull/3422

The PR above implements roughly the same idea.

Unfortunately, it also has the same issues that I described here.

system: iOS 18
using Test2991 from this PR
origin/main as of 19.02

### Issue no. 1

When mixing headerShown=false, displayMode=default and displayMode=minimal, the screen with displayMode=minimal still doesn't work. For now, please use workarounds suggested by the community in #2991.

https://github.com/user-attachments/assets/68a864f7-a8f1-4567-8df7-d2de27e5d282

### Issue no. 2

If I set back button label for Second so it fits and is fully rendered, and I go to Third first, only then to Second, the label is missing (after 0:20, observe that the display is different depending on whether I go third -> second or immediately to second).

https://github.com/user-attachments/assets/f69e5b8c-20c2-4c45-ab39-2916b56e4f4b

kmichalikk:
cc @kkafar @kligarski I need your attention here, please see the previous comment

either way, I'm closing this PR because at this point it's not adding anything

kkafar:
These errors seem a bit exotic & less sever than the original one. The case where a title for a particular screen changes between navigations to **seems** uncommon. 
Let's stay with current state until we get reports that these "new" issues do have a negative impact for someone.

PR #3209, opened 2025-09-10
feat(iOS): safe area component for iOS

kligarski:
## Description

Adds implementation for handling safe area on iOS.

Implementation has been adapted from [`react-native-safe-area-context`](https://github.com/AppAndFlow/react-native-safe-area-context).

https://github.com/user-attachments/assets/38b6afbd-95ee-48c5-83a3-0157c1fbc9c8

> [!IMPORTANT]
>
> Due to native iOS 26 bug, top safeAreaInset is incorrect when changing orientation from portrait to landscape. More details: https://github.com/software-mansion/react-native-screens-labs/issues/424.
>
> https://github.com/user-attachments/assets/93a7aba9-d15f-465b-acdf-8bfe05cd2d2a
>

Support for `SplitView` should be researched and UIKit native behavior should be verified (ticket for this: https://github.com/software-mansion/react-native-screens-labs/issues/426). Now, due to safe area left inset  "jumping" from 0 to full width, using safe area for screen content does not look smooth, even with synchronous updates.

https://github.com/user-attachments/assets/4b00528b-9ca8-49d2-9068-7b62a6f7efed

Related to https://github.com/software-mansion/react-native-screens-labs/issues/167 (but we should close it when we introduce API to use `SafeAreaView` in tabs).

## Changes

- add implementation for `SafeAreaView`
- add `RNSSafeAreaProviding` protocol
- implement `RNSSafeAreaProviding` for `RNSScreenView`, `RNSBottomTabsScreenComponentView` and `RNSSplitViewScreenComponentView`

Run `TestSafeAreaViewIOS`, or:

1. Run `Test3111`. Content shouldn't go under the modal header. You can also change orientation on the home screen (bug would occur there too).
2. Run `TestBottomTabs` with `SafeAreaView` in `Tab1.tsx, ..., Tab4.tsx`.
3. You can play around with `TestSplitView`, adding `SafeAreaView` with left edge (primary column limits safe area of the second column).

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [x] Ensured that CI passes


ios/RNSScreen.mm
```diff
@@ -1171,6 +1171,35 @@ - (void)overrideScrollViewBehaviorInFirstDescendantChainIfNeeded
...
+#pragma mark - RNSSafeAreaProviding
+
+- (UIEdgeInsets)RNS_safeAreaInsets
+{
+  return self.safeAreaInsets;
+}
+
+- (void)invalidateSafeAreaInsets
+{
+  // TODO: limit number of notifications if nothing has changed
```
t0maboro: non-blocking, but please link the ticket or ensure that we have one prepared for this

ios/safe-area/RNSSafeAreaProviding.h
```diff
@@ -0,0 +1,10 @@
+#import <UIKit/UIKit.h>
+
+#define RNSSafeAreaDidChange @"RNSSafeAreaDidChange"
```
t0maboro: I think that notifications-related definitions shouldn't be mixed with the protocols, I'd rather put it in a separate file - maybe one general header file for all definitions related to notifications would be sufficient - the same manner as we have for `RNSEnums`

ios/safe-area/RNSSafeAreaViewComponentView.h
```diff
@@ -1,13 +1,25 @@
 // Implementation adapted from `react-native-safe-area-context`:
 // https://github.com/AppAndFlow/react-native-safe-area-context/blob/v5.6.1/ios/Fabric/RNCSafeAreaViewComponentView.h
+// https://github.com/AppAndFlow/react-native-safe-area-context/blob/v5.6.1/ios/RNCSafeAreaView.h
```
t0maboro: maybe just `https://github.com/AppAndFlow/react-native-safe-area-context/blob/v5.6.1` ?

t0maboro (approved):
left some minor comments, compared the implementation with `react-native-safe-area-insets` - overall looks good

apps/src/tests/Test3111.tsx
kkafar, line 1: I think this new component deserves its own test case & example

apps/src/tests/TestBottomTabs/tabs/Tab1.tsx
```diff
@@ -2,11 +2,14 @@ import React from 'react';
 import { CenteredLayoutView } from '../../../shared/CenteredLayoutView';
 import { TabContentView } from '../components/TabContentView';
 import Colors from '../../../shared/styling/Colors';
+// import { SafeAreaView } from 'react-native-screens/private';
```
kkafar: Why is this commented out here?

ios/safe-area/RNSSafeAreaProviding.h
```diff
@@ -0,0 +1,10 @@
...
+
+@protocol RNSSafeAreaProviding
+
+- (UIEdgeInsets)RNS_safeAreaInsets;
```
kkafar: Why is this method defined in this "peculiar" way? If you're afraid of name clash with `safeAreaInsets` from `UIView` - I wouldn't be (unless it does not compile). Otherwise name it as we name category & extension methods - `rnscreens_XXX`.

ios/safe-area/RNSSafeAreaViewComponentView.mm
```diff
@@ -23,44 +48,162 @@ - (instancetype)initWithFrame:(CGRect)frame
...
+
+- (void)prepareForRecycle
+{
+  [super prepareForRecycle];
+
+  [NSNotificationCenter.defaultCenter removeObserver:self];
+  _state.reset();
+  _providerView = nil;
+  _currentSafeAreaInsets = UIEdgeInsetsZero;
+}
```
kkafar, line 193: What if this component is not recycled? Even if the view recycling is enabled, the component is not guaranteed to be recycled, as the pool of recycled components might be full. Will we leak resources in such scenario?

ios/safe-area/RNSSafeAreaViewComponentView.mm
```diff
@@ -23,44 +48,162 @@ - (instancetype)initWithFrame:(CGRect)frame
...
+    return;
+  }
+
+  _state->updateState(
+      [=](RNSSafeAreaViewShadowNode::ConcreteState::Data const &oldData)
+          -> RNSSafeAreaViewShadowNode::ConcreteState::SharedData {
+        auto newData = oldData;
+        newData.insets = RCTEdgeInsetsFromUIEdgeInsets(_currentSafeAreaInsets);
+        return std::make_shared<RNSSafeAreaViewShadowNode::ConcreteState::Data const>(newData);
+      });
```
kkafar: Why do we use this more verbose form of updateState here, instead of just passing the payload?

kkafar (commented):
Couple of questions, please answer them. 

I've just done code review. Have not tested the runtime yet.

ios/RNSScreen.mm
kligarski: Thank you for the reminder. 

https://github.com/software-mansion/react-native-screens-labs/issues/438

ios/safe-area/RNSSafeAreaViewComponentView.h
kligarski: https://github.com/software-mansion/react-native-screens/pull/3209/commits/22abaae8bc59c681437834203ab1fdab989eb8b1

ios/safe-area/RNSSafeAreaProviding.h
kligarski: https://github.com/software-mansion/react-native-screens/pull/3209/commits/84dc95f25d43c579c8b254bd45fd678e399dc20f

ios/safe-area/RNSSafeAreaProviding.h
kligarski: That's correct, I wanted to avoid name clash. We want this method to be separate from normal `safeAreaInsets` so that in the future, we can add/modify insets passed from the container (provider). I renamed it to `providerSafeAreaInsets`, as we discussed.

https://github.com/software-mansion/react-native-screens/pull/3209/commits/c26c7c3a9ee9dde0d9490105d8e3d414aa99b66b

ios/safe-area/RNSSafeAreaViewComponentView.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3209/commits/93fda618dff9c39426a084fd6e88f29a2e3961b9

apps/src/tests/TestBottomTabs/tabs/Tab1.tsx
kligarski: I did not want to use safe area in the test case by default but now I decided to remove `SAV`-related code from `Test3111`, `TestBottomTabs`.

In one of the follow-up PRs, I'll add `SafeAreaView` to our components behind a feature flag, this should make testing easier - I created a ticket for this: https://github.com/software-mansion/react-native-screens-labs/issues/439.

ios/safe-area/RNSSafeAreaViewComponentView.mm
kligarski: There shouldn't be any memory leak:
- `_providerView` is a weak reference,
- `_currentSafeAreaInsets` is just floats,
- you don't need to remove the observer according to [`NSNotificationCenter`'s docs](https://developer.apple.com/documentation/foundation/notificationcenter/addobserver(_:selector:name:object:)?language=objc#Discussion),
- as discussed internally, there shouldn't be any problem with `_state` shared pointer.

apps/src/tests/Test3111.tsx
kligarski: I added `TestSafeAreaViewIOS`.

apps/src/tests/TestSafeAreaViewIOS/bottom-tabs/BottomTabsComponent.tsx
kkafar, line 1: Just starting to read & wondernig why the test is dedicated to iOS? Won't it be shared between the platforms ultimately?

apps/src/tests/TestSafeAreaViewIOS/bottom-tabs/ConfigTab.tsx
```diff
@@ -0,0 +1,112 @@
...
+} from './BottomTabsSAVExampleContext';
+import { SettingsPicker, SettingsSwitch } from '../../../shared';
+import Info from '../shared/Info';
+
+export default function ConfigTab() {
+  const { config, setConfig } = useBottomTabsSAVExampleContext();
+
+  return (
+    <ScrollView
+      contentInsetAdjustmentBehavior="automatic"
```
kkafar: Do you need to configure this option? Should it not work out-of-the-box?

apps/src/tests/TestSafeAreaViewIOS/bottom-tabs/ConfigTab.tsx
```diff
@@ -0,0 +1,112 @@
...
+        label="tabBarMinimizeBehavior"
+        value={config.tabBarMinimizeBehavior}
+        onValueChange={value =>
+          setConfig({
+            ...config,
+            tabBarMinimizeBehavior: value,
+          })
+        }
+        items={['automatic', 'onScrollDown', 'onScrollUp', 'never']}
+      />*/}
```
kkafar, line 56: Could I get a little bit of explanation why is it the case?

apps/src/tests/TestSafeAreaViewIOS/bottom-tabs/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+      safeAreaLeftEdge: true,
+      safeAreaRightEdge: true,
+    });
+
+  return (
+    <BottomTabsSAVExampleContext.Provider
+      value={{
+        config: exampleConfig,
+        setConfig: setExampleConfig,
+      }}>
```
kkafar: Is this deliberately rendered above whole navigation container?

apps/src/tests/TestSafeAreaViewIOS/bottom-tabs/TestTab.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { useBottomTabsSAVExampleContext } from './BottomTabsSAVExampleContext';
+import { SafeAreaView } from 'react-native-screens/private';
+import { mapContentStringToComponent } from '../shared';
+
+export default function TestTab() {
+  const { config } = useBottomTabsSAVExampleContext();
+
+  let content = mapContentStringToComponent(config.content);
```
kkafar: Notice, that this will return new element each time `TestTab` is rendered, and therefore render recursively returned component. 

Might be sensible to memo it.

apps/src/tests/TestSafeAreaViewIOS/shared/ScrollViewNever.tsx
```diff
@@ -0,0 +1,14 @@
+import { ScrollView } from 'react-native';
+import LongText from './LongText';
+import React from 'react';
+import Colors from '../../../shared/styling/Colors';
+
+export default function ScrollViewNever() {
+  return (
+    <ScrollView
+      contentInsetAdjustmentBehavior="never"
```
kkafar, line 9: Won't this end up being overwritten in a tab?

ios/safe-area/RNSSafeAreaProviding.h
```diff
@@ -0,0 +1,8 @@
+#import <UIKit/UIKit.h>
+
+@protocol RNSSafeAreaProviding
+
+- (UIEdgeInsets)providerSafeAreaInsets;
+- (void)invalidateSafeAreaInsets;
```
kkafar: Let's add appropriate doc comments. What these methods are for. What should they return. What is the expected behvaiour. 

Also adding a doc comment for protocol would be great.

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
```diff
@@ -172,6 +173,35 @@ - (void)updateTabBarItem
...
+#pragma mark - RNSSafeAreaProviding
+
+- (UIEdgeInsets)providerSafeAreaInsets
+{
+  return self.safeAreaInsets;
+}
+
+- (void)invalidateSafeAreaInsets
+{
+  // TODO: limit number of notifications if nothing has changed
```
kkafar: I think there is already ticket for this, right?

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
```diff
@@ -172,6 +173,35 @@ - (void)updateTabBarItem
...
+  [NSNotificationCenter.defaultCenter postNotificationName:RNSSafeAreaDidChange object:self userInfo:nil];
+}
+
+#pragma mark - RNSSafeAreaProviding related methods
+
+// TODO: register for UIKeyboard notifications
```
kkafar, line 190: Let's create ticket for this.

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
```diff
@@ -172,6 +173,35 @@ - (void)updateTabBarItem
...
+
+- (void)safeAreaInsetsDidChange
+{
+  [super safeAreaInsetsDidChange];
+  [self invalidateSafeAreaInsets];
```
kkafar: Maybe we should keep the method naming consistent here? `invalidateProviderSafeAreaInsets`?

kkafar: To (partially) pay respects to grammar - `invalidateSafeAreaInsetsOfProvider`?

kkafar: Not gonna insist on this. Loose suggestion. What do you think?

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
```diff
@@ -172,6 +173,35 @@ - (void)updateTabBarItem
...
+}
+
+- (void)layoutSubviews
+{
+  [super layoutSubviews];
+  [self invalidateSafeAreaInsets];
```
kkafar: Invalidate on every single layout? Why is that necessary? Shouldn't we just observe frames of views we're interested in?

ios/gamma/split-view/RNSSplitViewScreenComponentView.mm
```diff
@@ -137,6 +140,29 @@ - (nonnull RNSSplitViewScreenComponentEventEmitter *)reactEventEmitter
...
+
+#pragma mark - RNSSafeAreaProviding related methods
+
+// TODO: register for UIKeyboard notifications
+
+- (void)safeAreaInsetsDidChange
+{
+  [super safeAreaInsetsDidChange];
+  [self invalidateSafeAreaInsets];
+}
```
kkafar: Obviously similar remarks to this section as above ☝🏻

ios/safe-area/RNSSafeAreaViewComponentView.h
```diff
@@ -1,13 +1,24 @@
...
 
 NS_ASSUME_NONNULL_BEGIN
 
 @interface RNSSafeAreaViewComponentView : RNSReactBaseView
 
+#if !RCT_NEW_ARCH_ENABLED
+- (instancetype)initWithBridge:(RCTBridge *)bridge;
+
+@property (nonatomic, assign) RNSSafeAreaViewEdges edges;
+#endif // !RCT_NEW_ARCH_ENABLED
```
kkafar: Let's move these legacy architecture specific methods / properties to separate extension.

ios/safe-area/RNSSafeAreaViewComponentView.mm
```diff
@@ -24,15 +49,118 @@ - (instancetype)initWithFrame:(CGRect)frame
...
+  _edges = RNSSafeAreaViewEdgesMake(false, false, false, false);
+#endif // RCT_NEW_ARCH_ENABLED
+}
+
+- (void)didMoveToWindow
+{
+  UIView *previousProviderView = _providerView;
+  _providerView = [self findNearestProvider];
+
+  [self invalidateSafeAreaInsets];
```
kkafar: Do we wanna to send notifications if `self.window == nil`?

ios/safe-area/RNSSafeAreaViewComponentView.mm
```diff
@@ -24,15 +49,118 @@ - (instancetype)initWithFrame:(CGRect)frame
...
+    if (_providerView != nil) {
+      [NSNotificationCenter.defaultCenter addObserver:self
+                                             selector:@selector(safeAreaProviderInsetsDidChange:)
+                                                 name:RNSSafeAreaDidChange
+                                               object:_providerView];
+    }
+  }
+}
+
+- (UIView<RNSSafeAreaProviding> *_Nullable)findNearestProvider
```
kkafar:
```suggestion
- (UIView<RNSSafeAreaProviding> *_Nullable)findNearestAncestorProvider
```

ios/safe-area/RNSSafeAreaViewComponentView.mm
```diff
@@ -24,15 +49,118 @@ - (instancetype)initWithFrame:(CGRect)frame
...
+                                                                                   edges:_edges];
+  [_bridge.uiManager setLocalData:localData forView:self];
+}
+#endif // RCT_NEW_ARCH_ENABLED
+
+BOOL UIEdgeInsetsEqualToEdgeInsetsWithThreshold(UIEdgeInsets insets1, UIEdgeInsets insets2, CGFloat threshold)
+{
+  return ABS(insets1.left - insets2.left) <= threshold && ABS(insets1.right - insets2.right) <= threshold &&
+      ABS(insets1.top - insets2.top) <= threshold && ABS(insets1.bottom - insets2.bottom) <= threshold;
+}
```
kkafar: Move this function outside the class & make it static (in C++ sense).

ios/safe-area/RNSSafeAreaViewComponentView.mm
```diff
@@ -41,9 +169,37 @@ - (void)initState
...
+           oldState:(facebook::react::State::Shared const &)oldState
+{
+  using facebook::react::RNSSafeAreaViewShadowNode;
+  _state = std::static_pointer_cast<RNSSafeAreaViewShadowNode::ConcreteState const>(state);
+}
+
+- (void)finalizeUpdates:(RNComponentViewUpdateMask)updateMask
+{
+  [super finalizeUpdates:updateMask];
+  [self invalidateSafeAreaInsets];
```
kkafar: Don't we wanna filter the calls to the `invalidateSafeAreaInsets` function here a bit? I know that there are early returns inside, but still.

ios/RNSNotifications.h
```diff
@@ -0,0 +1,5 @@
+#pragma once
+
+#pragma mark - NSNotificationCenter notification names
+
+#define RNSSafeAreaDidChange @"RNSSafeAreaDidChange"
```
kkafar: I think this header should live inside `safe-area` subdirectory. Maybe name it in a bit less general way, e.g. `RNSSafeAreaViewDefinitions` or `RNSSafeAreaViewNotifications`

ios/RNSScreen.mm
kkafar, line 1: Similar remarks regarding naming in this file & calls on each subview layout in this file.

apps/src/tests/TestSafeAreaViewIOS/bottom-tabs/BottomTabsComponent.tsx
kligarski: I'm 50/50 on this one - there are some differences how `SafeAreaView` will work on both platforms (on Android, it'll be possible to specify which type of insets should be used, `system`, `interface` or both). There are some iOS-related options (ScrollViews with `contentInsetAdjustmentBehavior`, `searchBarPlacement`, `headerLargeTitle`, entire SplitView example). We're also currently not planning to implement `SafeAreaView` for Stack v4 on Android.

On the other hand, maybe it wouldn't be a problem to mark which options are available on which platform?

apps/src/tests/TestSafeAreaViewIOS/bottom-tabs/ConfigTab.tsx
kligarski: Currently in `BottomTabsContainer.tsx`, props are hardcoded to `BottomTabs` component. I created a ticket to handle this: https://github.com/software-mansion/react-native-screens-labs/issues/450.

apps/src/tests/TestSafeAreaViewIOS/shared/ScrollViewNever.tsx
kligarski: It should be but as this component is shared by all test cases, it can be used in stack-only example as well. I will remove an option to use it from content options in bottom tabs example.

apps/src/tests/TestSafeAreaViewIOS/bottom-tabs/ConfigTab.tsx
kligarski: It should. I removed it.

https://github.com/software-mansion/react-native-screens/pull/3209/commits/4ee6ba47e6e7927091f98f3f6084781358e5db35

apps/src/tests/TestSafeAreaViewIOS/bottom-tabs/index.tsx
kligarski: No, I moved it lower now.

https://github.com/software-mansion/react-native-screens/pull/3209/commits/4ee6ba47e6e7927091f98f3f6084781358e5db35

apps/src/tests/TestSafeAreaViewIOS/bottom-tabs/TestTab.tsx
kligarski: I used memo in tabs and stack examples. I'm not sure if I can use it in SplitView without overcomplicating the example.

https://github.com/software-mansion/react-native-screens/pull/3209/commits/4ee6ba47e6e7927091f98f3f6084781358e5db35

apps/src/tests/TestSafeAreaViewIOS/shared/ScrollViewNever.tsx
kligarski: https://github.com/software-mansion/react-native-screens/pull/3209/commits/4ee6ba47e6e7927091f98f3f6084781358e5db35

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
kligarski: Right: https://github.com/software-mansion/react-native-screens-labs/issues/438.

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
kligarski: https://github.com/software-mansion/react-native-screens-labs/issues/451

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
kligarski: I think that `invalidateSafeAreaInsetsOfProvider` sounds a little bit odd when we call the method in the component itself, which *is* the provider. I'd go for `invalidateProviderSafeAreaInsets`.

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
kligarski: I followed `react-native-safe-area-context` implementation here but didn't implement the logic to limit number of notifications yet (checking if safe area actually changed).

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
kligarski: Another thing I thought about is to rename it to `update...` instead of `invalidate...`. When we introduce logic to limit number of notifications, it can even be `update...IfNeeded`.

ios/safe-area/RNSSafeAreaProviding.h
kligarski: https://github.com/software-mansion/react-native-screens/pull/3209/commits/0289b6b088262380b8dc37fcc559386e05dbcb4f

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
kligarski: I removed calls to this method other than from `safeAreaInsetsDidChange` - each call should be valid now.

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
kligarski: I removed calls from `layoutSubviews`.

https://github.com/software-mansion/react-native-screens/pull/3209/commits/4deb88f049e6e9a746c6271911475dd80690102a

ios/safe-area/RNSSafeAreaViewComponentView.h
kligarski: https://github.com/software-mansion/react-native-screens/pull/3209/commits/4deb88f049e6e9a746c6271911475dd80690102a

ios/safe-area/RNSSafeAreaViewComponentView.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3209/commits/4deb88f049e6e9a746c6271911475dd80690102a

ios/safe-area/RNSSafeAreaViewComponentView.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3209/commits/0289b6b088262380b8dc37fcc559386e05dbcb4f

ios/safe-area/RNSSafeAreaViewComponentView.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3209/commits/0289b6b088262380b8dc37fcc559386e05dbcb4f

ios/safe-area/RNSSafeAreaViewComponentView.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3209/commits/0289b6b088262380b8dc37fcc559386e05dbcb4f

ios/RNSNotifications.h
kligarski: https://github.com/software-mansion/react-native-screens/pull/3209/commits/0289b6b088262380b8dc37fcc559386e05dbcb4f

ios/RNSScreen.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3209/commits/4deb88f049e6e9a746c6271911475dd80690102a, https://github.com/software-mansion/react-native-screens/pull/3209/commits/0289b6b088262380b8dc37fcc559386e05dbcb4f

ios/gamma/split-view/RNSSplitViewScreenComponentView.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3209/commits/4deb88f049e6e9a746c6271911475dd80690102a, https://github.com/software-mansion/react-native-screens/pull/3209/commits/0289b6b088262380b8dc37fcc559386e05dbcb4f

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3209/commits/823bc56e73d6080aa011d47ed9e6e3db110c5b48

apps/src/tests/TestSafeAreaViewIOS/bottom-tabs/ConfigTab.tsx
kkafar: 👍🏻

kkafar (commented):
Looks really good. Great job!

I really like the text example you created. It really simplified the review & will make it easier to test for regressions in the future. 

Few questions from playing around:

---

No. 1

I disabled SAV top edge in stack test. Expected behaviour would be to have the content be laid out underneath the navigation bar. But this is not the case. This is because we still use `edgesForExtendedLayout` and this PR does not change that, right? 

https://github.com/user-attachments/assets/5c3c26b4-a7aa-4ee1-be5c-01435a99eb6b

Note: I notice that with header transparent it works fine. Good job!

---

No. 2

I've spotted, that there are some layout constraint conflicts. Have you noticed them? 

<details>

<summary>Logs</summary>

```
Unable to simultaneously satisfy constraints.
	Probably at least one of the constraints in the following list is one you don't want. 
	Try this: 
		(1) look at each constraint and try to figure out which you don't expect; 
		(2) find the code that added the unwanted constraint or constraints and fix it. 
(
    "<NSLayoutConstraint:0x12672f890 _TtCC5UIKit29ButtonBarButtonVisualProviderP33_A98CD29F4F6ECA17AFECE41BBB264E596Button:0x1265ed500.width <= 106.667   (active)>",
    "<NSLayoutConstraint:0x12672e300 'fittingSizeHTarget' _TtCC5UIKit29ButtonBarButtonVisualProviderP33_A98CD29F4F6ECA17AFECE41BBB264E596Button:0x1265ed500.width == 145   (active)>"
)

Will attempt to recover by breaking constraint 
<NSLayoutConstraint:0x12672f890 _TtCC5UIKit29ButtonBarButtonVisualProviderP33_A98CD29F4F6ECA17AFECE41BBB264E596Button:0x1265ed500.width <= 106.667   (active)>

Make a symbolic breakpoint at UIViewAlertForUnsatisfiableConstraints to catch this in the debugger.
The methods in the UIConstraintBasedLayoutDebugging category on UIView listed in <UIKitCore/UIView.h> may also be helpful.
Unable to simultaneously satisfy constraints.
	Probably at least one of the constraints in the following list is one you don't want. 
	Try this: 
		(1) look at each constraint and try to figure out which you don't expect; 
		(2) find the code that added the unwanted constraint or constraints and fix it. 
(
    "<NSLayoutConstraint:0x12672f890 _TtCC5UIKit29ButtonBarButtonVisualProviderP33_A98CD29F4F6ECA17AFECE41BBB264E596Button:0x1265ed500.width <= 106.667   (active)>",
    "<NSLayoutConstraint:0x12672c3c0 'fittingSizeHTarget' _TtCC5UIKit29ButtonBarButtonVisualProviderP33_A98CD29F4F6ECA17AFECE41BBB264E596Button:0x1265ed500.width == 145   (active)>"
)

Will attempt to recover by breaking constraint 
<NSLayoutConstraint:0x12672f890 _TtCC5UIKit29ButtonBarButtonVisualProviderP33_A98CD29F4F6ECA17AFECE41BBB264E596Button:0x1265ed500.width <= 106.667   (active)>

Make a symbolic breakpoint at UIViewAlertForUnsatisfiableConstraints to catch this in the debugger.
The methods in the UIConstraintBasedLayoutDebugging category on UIView listed in <UIKitCore/UIView.h> may also be helpful.
Unable to simultaneously satisfy constraints.
	Probably at least one of the constraints in the following list is one you don't want. 
	Try this: 
		(1) look at each constraint and try to figure out which you don't expect; 
		(2) find the code that added the unwanted constraint or constraints and fix it. 
(
    "<NSLayoutConstraint:0x12672f890 _TtCC5UIKit29ButtonBarButtonVisualProviderP33_A98CD29F4F6ECA17AFECE41BBB264E596Button:0x1265ed500.width <= 106.667   (active)>",
    "<NSLayoutConstraint:0x12672e8a0 'fittingSizeHTarget' _TtCC5UIKit29ButtonBarButtonVisualProviderP33_A98CD29F4F6ECA17AFECE41BBB264E596Button:0x1265ed500.width == 145   (active)>"
)

Will attempt to recover by breaking constraint 
<NSLayoutConstraint:0x12672f890 _TtCC5UIKit29ButtonBarButtonVisualProviderP33_A98CD29F4F6ECA17AFECE41BBB264E596Button:0x1265ed500.width <= 106.667   (active)>

Make a symbolic breakpoint at UIViewAlertForUnsatisfiableConstraints to catch this in the debugger.
The methods in the UIConstraintBasedLayoutDebugging category on UIView listed in <UIKitCore/UIView.h> may also be helpful.
```

</details>

This happens in `BottomTabsSAVExample` when there is stack in the other tab. 

This is the setup when it happens: 

https://github.com/user-attachments/assets/9f0268a1-492a-4660-b7d2-c1644dce98df

just after I navigate to the `Test` tab.

kligarski:
> No. 1
> 
> I disabled SAV top edge in stack test. Expected behaviour would be to have the content be laid out underneath the navigation bar. But this is not the case. This is because we still use edgesForExtendedLayout and this PR does not change that, right?

That's correct. As I understand, for now we don't want to introduce any potentially breaking changes with handling safe area insets in Stack v4, that's why we want to keep `edgesForExtendedLayout` for now. `SafeAreaView` will be used as workaround for native UIKit layout bug on iOS 26 with content being rendered behind UINavigationBar (PR for this: https://github.com/software-mansion/react-native-screens/pull/3231).

I guess that in the future, if we make sure that everything works as expected, we can switch fully to `SafeAreaView`, without `edgesForExtendedLayout`.

For new Stack implementation, we should use `SafeAreaView` only.

> No. 2
> 
> I've spotted, that there are some layout constraint conflicts. Have you noticed them?

I think they are the same as https://github.com/software-mansion/react-native-screens-labs/issues/320 (connected to `backButtonTitle`). There is an issue to handle native back button on iOS 26: https://github.com/software-mansion/react-native-screens-labs/issues/449.

matinzd:
What do you think about exposing the inset values just like what's being done in the other lib?

PR #3212, opened 2025-09-12
feat(iOS): Add scroll edge effects

kmichalikk:
## Description

Closes https://github.com/software-mansion/react-native-screens-labs/issues/419.

This PR adds support for setting [UIScrollEdgeEffect](https://developer.apple.com/documentation/uikit/uiscrolledgeeffect) on ScrollViews that are first descendants of screens.

For each edge, UIKit supports setting number of `style`s and a `hidden` toggle. Our API will merge those two into one enum: `automatic`, `hard`, `soft` and `hidden`. This still leaves us with 4 separate Screen props, for each edge, and another 4 for BottomTabsScreen.

## Changes

Added {bottom,left,right,top}ScrollEdgeEffect prop for Screen and BottomTabsScreen.

https://github.com/user-attachments/assets/d3efc202-a502-46e4-a3a9-06e553401319

Use BottomTabsTest, add `scrollEdgeEffect: { bottom: 'hard' }` to Tab3, add `scrollEdgeEffect: { bottom: 'hard', top: 'hard' }` to a Screen in Tab4.

## Important
Setting the effect on Screen inside BottomTabs won't work on Paper, because it is immediately overwritten by BottomTabs' values, will be fixed in the future.

### DIFF
--- a/src/components/bottom-tabs/BottomTabsScreen.types.ts
+++ b/src/components/bottom-tabs/BottomTabsScreen.types.ts
@@ -5,6 +5,7 @@ import type {
   TextStyle,
   ViewProps,
 } from 'react-native';
+import { ScrollEdgeEffect } from '../shared/types';
 
 export type EmptyObject = Record<string, never>;
 
@@ -85,8 +86,6 @@ export type BottomTabsScreenOrientation =
   | 'landscape'
   | 'landscapeLeft'
   | 'landscapeRight';
-
-export type ScrollEdgeEffect = 'automatic' | 'hard' | 'soft' | 'hidden';
 
 // iOS-specific
 export interface BottomTabsScreenAppearance {
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -9,7 +9,7 @@ import {
   ColorValue,
 } from 'react-native';
 import { NativeStackNavigatorProps } from './native-stack/types';
-import { ScrollEdgeEffect } from './components/bottom-tabs/BottomTabsScreen.types';
+import { ScrollEdgeEffect } from './components/shared/types';
 
 export type SearchBarCommands = {
   focus: () => void;
@@ -989,3 +989,4 @@ export interface GestureProviderProps extends GestureProps {
 
 export * from './components/bottom-tabs/BottomTabs.types';
 export * from './components/bottom-tabs/BottomTabsScreen.types';
+export * from './components/shared/types';
--- a/src/components/shared/types.ts
+++ b/src/components/shared/types.ts
@@ -1,0 +1,1 @@
+export type ScrollEdgeEffect = 'automatic' | 'hard' | 'soft' | 'hidden';


kmichalikk:
There is still a problem that occasionally the size of the effect doubles on reload. Don't know why it happens, can't compare with native app because it shows on reload.

apps/src/tests/TestSplitView/SplitViewBaseApp.tsx
```diff
@@ -2,22 +2,23 @@ import React from 'react';
...
-        <View style={[styles.container, { backgroundColor: Colors.White }]}>
+        {/* <View style={[styles.container, { backgroundColor: Colors.White }]}>
           <Text style={styles.text}>Primary column</Text>
-        </View>
+        </View> */}
+        <TestFoo />
       </SplitViewScreen.Column>
       <SplitViewScreen.Column>
-        <TestBottomTabs />
+        <TestFoo />
```
kligarski: Not sure if `TestFoo` is the most descriptive name d:

kligarski: We should probably restore this file.

ios/RNSConvert.mm
```diff
@@ -133,6 +133,45 @@ + (RNSScreenSwipeDirection)RNSScreenSwipeDirectionFromCppEquivalent:(react::RNSS
...
+    case Automatic:                        \
+      return RNSScrollEdgeEffectAutomatic; \
+    case Hard:                             \
+      return RNSScrollEdgeEffectHard;      \
+    case Soft:                             \
+      return RNSScrollEdgeEffectSoft;      \
+    case Hidden:                           \
+      return RNSScrollEdgeEffectHidden;    \
+    default:                               \
+      return RNSScrollEdgeEffectAutomatic; \
```
kligarski: We can add a warning log if we don't match any option (in case in the future there are more options for the effect).

ios/RNSScreen.mm
```diff
@@ -1369,6 +1392,17 @@ - (void)setScreenId:(NSString *)screenId
 - (void)didSetProps:(NSArray<NSString *> *)changedProps
 {
   [super didSetProps:changedProps];
+
+  BOOL shouldUpdateScrollEffects = NO;
+  for (NSString *propName : changedProps) {
+    shouldUpdateScrollEffects = shouldUpdateScrollEffects || [propName hasSuffix:@"ScrollEdgeEffect"];
+  }
```
kligarski: Maybe we can override the setters on Paper to set `_shouldUpdateScrollEffects` flag to `YES` whenever we change the prop value.

ios/RNSScreen.mm
```diff
@@ -1369,6 +1392,17 @@ - (void)setScreenId:(NSString *)screenId
...
+
+  if (shouldUpdateScrollEffects) {
+    // see finalizeUpdates() for Fabric
+    [RNSScrollEdgeEffectApplicator applyToScrollView:[RNSScrollViewFinder findScrollViewInFirstDescendantChainFrom:self]
+                                        fromProvider:self];
+  }
```
kligarski: Why do we limit the number of calls to `[RNSScrollEdgeEffectApplicator applyToScrollView:...]` only on Paper?

ios/bottom-tabs/RCTConvert+RNSBottomTabs.mm
```diff
@@ -66,6 +66,17 @@ + (UIOffset)UIOffset:(id)json;
...
+    RNSScrollEdgeEffect,
+    (@{
+      @"automatic" : @(RNSScrollEdgeEffectAutomatic),
+      @"hard" : @(RNSScrollEdgeEffectHard),
+      @"soft" : @(RNSScrollEdgeEffectSoft),
+      @"hidden" : @(RNSScrollEdgeEffectHidden),
+    }),
+    RNSScrollEdgeEffectAutomatic,
+    integerValue)
+
```
kligarski: I 90% sure that we don't need to repeat `RCT_ENUM_CONVERTER(...)` in both `ios/RNSScreen.mm` and `ios/bottom-tabs/RCTConvert+RNSBottomTabs.mm` and we can do it in one place only (e.g. `RCTConvert+RNScreens.mm`)

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
```diff
@@ -153,6 +154,36 @@ - (void)overrideScrollViewBehaviorInFirstDescendantChainIfNeeded
   }
 }
 
+- (void)setBottomScrollEdgeEffect:(RNSScrollEdgeEffect)bottomScrollEdgeEffect
+{
+  _bottomScrollEdgeEffect = bottomScrollEdgeEffect;
+  [self findTabBarController].needsUpdateOfSelectedTab = YES;
```
kligarski: I'm not sure if updating selected tab only to apply scroll edge effect is correct. Maybe we should extract the logic for updating scroll edge effects and run it after changing any prop related to scroll edge effects AND in `updateSelectedViewController`.

Also, note that the screen should report the change via its own screen controller (`RNSTabsScreenViewController`) instead of calling tab bar controller (`RNSTabBarController`) directly (for reference, check the end of `updateProps` method: how appearance/orientation updates are handled).

ios/RNSConvert.mm
kligarski: Applies also to `ios/conversion/RNSConversions-BottomTabs.mm`.

ios/services/RNSScrollEdgeEffectApplicator.mm
```diff
@@ -0,0 +1,50 @@
...
+    [RNSScrollEdgeEffectApplicator configureSingleEffect:scrollView.rightEdgeEffect
+                                                withEnum:[provider rightScrollEdgeEffect]];
+    [RNSScrollEdgeEffectApplicator configureSingleEffect:scrollView.topEdgeEffect
+                                                withEnum:[provider topScrollEdgeEffect]];
+  }
+#endif
+}
+
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
++ (void)configureSingleEffect:(UIScrollEdgeEffect *)edgeEffect
```
kligarski: nitpick: not sure if we need to highlight that this is a `single` effect but I don't have a strong opinion here

ios/services/RNSScrollEdgeEffectApplicator.mm
```diff
@@ -0,0 +1,50 @@
...
+        break;
+      case RNSScrollEdgeEffectSoft:
+        edgeEffect.hidden = false;
+        edgeEffect.style = UIScrollEdgeEffectStyle.softStyle;
+        break;
+      case RNSScrollEdgeEffectHidden:
+        edgeEffect.hidden = true;
+        edgeEffect.style = UIScrollEdgeEffectStyle.automaticStyle;
+        break;
+    }
```
kligarski, line 47: We might want to add default case with warning.

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -514,6 +516,22 @@ export interface BottomTabsScreenProps {
    * @platform ios
    */
   overrideScrollViewContentInsetAdjustmentBehavior?: boolean;
+  /**
+   * Configures the scroll edge effect for the ScrollView that is the first descendant of the BottomTabScreen.
```
kligarski: It does not need to be the first child, it can also be e.g. the first child of the first child of the `BottomTabsScreen` component (that's why in the method name there is `firstDescendantChain`)

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -514,6 +516,22 @@ export interface BottomTabsScreenProps {
...
+   * Depending on values set, it will blur the scrolling content below certain UI elements (Header Items, SearchBar)
+   * for the specifed edge of the ScrollView.
+   *
+   * When set in nested containers, i.e. ScreenStack inside BottomTabs, or the other way around,
+   * it will use the closest one (*without* merging the other).
```
kligarski: I'd consider being more specific here (closest to what, what does it mean to be closest, merged with what)?

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -514,6 +516,22 @@ export interface BottomTabsScreenProps {
...
+   *
+   * @platform ios
```
kligarski, line 526: We should add information that this is supported starting from iOS 26. We can also consider adding a link to UIKit documentation.

src/components/bottom-tabs/BottomTabsScreen.types.ts
kligarski: Also applies to `Screen`'s docs.

src/components/bottom-tabs/BottomTabsScreen.types.ts
kligarski: Also applies to `Screen`'s docs.

src/components/bottom-tabs/BottomTabsScreen.types.ts
kligarski: Also applies to `Screen`'s docs.

apps/src/tests/TestSplitView/SplitViewBaseApp.tsx
kmichalikk: Yup, I wasn't meant to push this

ios/RNSConvert.mm
kmichalikk: 58d3304c06229b675411e65de50598ce2c6193d8

ios/RNSScreen.mm
kmichalikk: done

ios/RNSScreen.mm
kmichalikk: added it for Fabric also

ios/bottom-tabs/RCTConvert+RNSBottomTabs.mm
kmichalikk: You're right, 0e60b4a8842e9225a595a3646cc2ce0c95d29b87

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
kmichalikk: Moved the updates to view class, the call in tab is left only for correct initialization

ios/services/RNSScrollEdgeEffectApplicator.mm
kmichalikk: d351ee577589c8c36fafd45b845db10e81118364

ios/services/RNSScrollEdgeEffectApplicator.mm
kmichalikk: 411872dbcdfa1904a0576af53c20483bdc3d269d

src/components/bottom-tabs/BottomTabsScreen.types.ts
kmichalikk: 4b18e812badad1443fa27a6a3ce9eb09d65b9e65

src/components/bottom-tabs/BottomTabsScreen.types.ts
kmichalikk: 4b18e812badad1443fa27a6a3ce9eb09d65b9e65

src/components/bottom-tabs/BottomTabsScreen.types.ts
kmichalikk: 4b18e812badad1443fa27a6a3ce9eb09d65b9e65

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
```diff
@@ -134,11 +137,6 @@ - (nonnull RNSBottomTabsScreenEventEmitter *)reactEventEmitter
   return _reactEventEmitter;
 }
 
-- (nullable RNSTabBarController *)findTabBarController
-{
-  return static_cast<RNSTabBarController *>(_controller.tabBarController);
-}
```
kkafar, line 147: Why is this method removed here?

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
```diff
@@ -312,6 +340,37 @@ - (void)updateProps:(const facebook::react::Props::Shared &)props
...
+        setRightScrollEdgeEffect:
+            rnscreens::conversion::RNSBottomTabsScrollEdgeEffectFromBottomTabsScreenRightScrollEdgeEffectCppEquivalent(
+                newComponentProps.rightScrollEdgeEffect)];
+  }
+
+  if (newComponentProps.topScrollEdgeEffect != oldComponentProps.topScrollEdgeEffect) {
+    [self setTopScrollEdgeEffect:rnscreens::conversion::
+                                     RNSBottomTabsScrollEdgeEffectFromBottomTabsScreenTopScrollEdgeEffectCppEquivalent(
+                                         newComponentProps.topScrollEdgeEffect)];
+  }
```
kkafar, line 367: These prop setters call should be move a bit up & grouped with other setters. Notice that previous setter block is now separated by update calls.

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
```diff
@@ -312,6 +340,37 @@ - (void)updateProps:(const facebook::react::Props::Shared &)props
...
+
+  if (_scrollEdgeEffectsNeedUpdate) {
```
kkafar: Let's align it here with other callbacks. `_scrollEdgeEffectsNeedUpdate` should be function local state on new architecture & instance state in old architecture. Take a look at other props implementation. 

I'm not sure whether its best approach or not, but I prefer it to be aligned with existing one, rather than mixing things up. 

You might wanna take a look at this ticket: https://github.com/software-mansion/react-native-screens-labs/issues/280 after & simplify it later (please do).

ios/conversion/RNSConversions-BottomTabs.mm
```diff
@@ -319,4 +319,44 @@ UITabBarSystemItem RNSBottomTabsScreenSystemItemToUITabBarSystemItem(RNSBottomTa
...
+{
+  SWITCH_EDGE_EFFECT(RNSBottomTabsScreenRightScrollEdgeEffect);
+}
+
+RNSScrollEdgeEffect RNSBottomTabsScrollEdgeEffectFromBottomTabsScreenTopScrollEdgeEffectCppEquivalent(
+    react::RNSBottomTabsScreenTopScrollEdgeEffect edgeEffect)
+{
+  SWITCH_EDGE_EFFECT(RNSBottomTabsScreenTopScrollEdgeEffect);
+}
+
```
kkafar, line 361: `#undef` this macro after its no longer needed.

ios/RNSScreen.mm
```diff
@@ -1345,6 +1398,11 @@ - (void)updateLayoutMetrics:(const react::LayoutMetrics &)layoutMetrics
 - (void)finalizeUpdates:(RNComponentViewUpdateMask)updateMask
 {
   [super finalizeUpdates:updateMask];
+  if (_shouldUpdateScrollEdgeEffects) {
+    [self updateScrollEdgeEffects];
+    _shouldUpdateScrollEdgeEffects = NO;
```
kkafar: This flag could be cleared up in `updateScrollEdgeEffects` method.

kkafar: Would be easier to keep it consistent.

ios/RNSConvert.mm
```diff
@@ -133,6 +133,46 @@ + (RNSScreenSwipeDirection)RNSScreenSwipeDirectionFromCppEquivalent:(react::RNSS
   }
 }
 
+#define SWITCH_EDGE_EFFECT(X)                              \
```
kkafar, line 136: `#undef` it after usage

ios/services/RNSScrollEdgeEffectApplicator.mm
```diff
@@ -0,0 +1,52 @@
+#import "RNSScrollEdgeEffectApplicator.h"
+#import <React/RCTLog.h>
+#import "RNSDefines.h"
+#import "RNSEnums.h"
+
+@implementation RNSScrollEdgeEffectApplicator
+
++ (void)applyToScrollView:(UIScrollView *)scrollView fromProvider:(id<RNSScrollEdgeEffectProviding>)provider
```
kkafar:
```suggestion
+ (void)applyToScrollView:(UIScrollView *)scrollView withProvider:(id<RNSScrollEdgeEffectProviding>)provider
```

just sounds better, big nitpick

ios/services/RNSScrollEdgeEffectApplicator.h
kkafar, line 1: Th code looks good for the applicator, however the `ios/services/` directory naming is questionable. We have utils or helpers (in case we don't, that would be more appropriate name here).

ios/bottom-tabs/RNSBottomTabsScreenComponentView.h
```diff
@@ -37,14 +38,16 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, weak, nullable) RNSBottomTabsHostComponentView *reactSuperview;
 
+- (void)updateScrollEdgeEffects;
```
kkafar: Let's document this method nicely - what it does & all necessary caveats. When does it fail? Paper? Fabric?

kkafar:
```suggestion
- (void)maybeUpdateContentScrollViewEdgeEffects;
```

Current name is misleading, as it is not `self` who is the subject of the update. 

`maybe` prefix indicates that this method might not apply desired effect.

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
kmichalikk: Partially because I thought I left it there after my previous attempts (it was grayed as not used), and partially because of me instinctively following _the boy scout rule_.

For the second one, guess it should have been in another PR, do you want me to make one?

ios/RNSConvert.mm
kkafar: https://github.com/software-mansion/react-native-screens/pull/3212/commits/a82c83aacb9190a81c59fd1b9c36fdb3433dc10f

src/types.tsx
```diff
@@ -106,6 +106,8 @@ export type SearchBarPlacement =
   | 'integratedButton'
   | 'integratedCentered';
 
+export type ScrollEdgeEffect = 'automatic' | 'hard' | 'soft' | 'hidden';
```
kkafar: I dislike the fact, that this public type is defined both here and in `BottomTabsScreen.types.ts`.

There should be only a single definition of it. My suggestion would be to either:

1. import it directly from `src/types.tsx` in bottom tabs component,
2. or create `src/components/shared.types.ts` or something like this. 

If we go with option 2, we should identify all shared (possibly duplicated types) now and move them to shared types file in follow up PR. 

I think I prefer option 1 for start & option 2 in follow-up PR together with the whole refactor.

kmichalikk: f251127fb57283f21e990da858f89f8c80b643ad

ios/bottom-tabs/RNSBottomTabsScreenComponentView.h
kmichalikk: changed to `updateContentScrollViewEdgeEffectsIfExists` -> 9324d95dbd0225695e937546c08f4c437352a098 43505f8cfa74e786b337822836d17fe46b74d4ea 1f55df1dfe899eec8df4a5e1f4b61d4c0e3f096f

kmichalikk: 9324d95dbd0225695e937546c08f4c437352a098

ios/services/RNSScrollEdgeEffectApplicator.mm
kmichalikk: 6822e9b8a9b242540160d34315b7d2e44cf357b1

ios/RNSScreen.mm
kmichalikk: I left it as it to not blur the responsibilities of the function

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
kmichalikk: 6822e9b8a9b242540160d34315b7d2e44cf357b1

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
kmichalikk: 6822e9b8a9b242540160d34315b7d2e44cf357b1

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
kmichalikk: reverted in 6822e9b8a9b242540160d34315b7d2e44cf357b1

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -6,6 +6,8 @@ import type {
   ViewProps,
 } from 'react-native';
 
+import { ScrollEdgeEffect } from 'react-native-screens/types';
```
kkafar: 🤔 There are two things wrong with this line:

1. we should use relative imports here. AFAIK this style of imports can lead to issues with certain bundlers. Can't send you resources right now (do not recall all the details), but while ago we had to change such package-imports to relative ones; that's why I know. Can dig in blame / commit history to find it. 
2. Haven't thought of that, when writing previous review, but this way we create import cycle, right? This is also a case where bundlers might decide to throw errors / warnings. This **might be temporarily fine** if we have such situation somewhere else. If not, we need to move these definitions to separate file, I'm afraid.

kmichalikk: [7d44d6e](https://github.com/software-mansion/react-native-screens/pull/3212/commits/7d44d6e36140a3646e02d306f81de3e594d4de95)

went the other way around & imported from BottomTabsScreen into types, there no imports of types in BottomTabsScreen

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
kkafar: No, I think it can stay in the class. That's fine.

ios/services/RNSScrollEdgeEffectApplicator.h
kkafar: Please apply this in follow up PR.

PR #3214, opened 2025-09-15
refactor(Tabs): Refactor APIs for icons

t0maboro:
## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/apps/src/tests/TestBottomTabs/index.tsx
+++ b/apps/src/tests/TestBottomTabs/index.tsx
@@ -35,13 +35,19 @@ const TAB_CONFIGS: TabConfiguration[] = [
       title: 'Tab1',
       isFocused: true,
       icon: {
-        sfSymbolName: 'house',
+        ios: {
+          type: 'sfSymbolIOS',
+          name: 'house.fill',
+        }, 
+        android: {
+          type: 'imageSource',
+          imageSource: require('../../../assets/variableIcons/icon_fill.png'),
+        }
       },
       selectedIcon: {
-        sfSymbolName: 'house.fill',
+        type: 'sfSymbolIOS',
+        name: 'house.fill',
       },
-      // iconResourceName: 'sym_call_incoming', // Android specific
-      iconResource: require('../../../assets/variableIcons/icon_fill.png'),
     },
     component: Tab1,
   },
@@ -96,13 +102,19 @@ const TAB_CONFIGS: TabConfiguration[] = [
       },
       tabBarItemBadgeBackgroundColor: Colors.GreenDark100,
       icon: {
-        templateSource: require('../../../assets/variableIcons/icon.png'),
+        ios: {
+          type: 'templateSourceIOS',
+          templateSource: require('../../../assets/variableIcons/icon.png'),
+        }, 
+        android: {
+          type: 'drawableResourceAndroid',
+          name: 'sym_call_missed', // Android specific
+        }
       },
       selectedIcon: {
+        type: 'templateSourceIOS',
         templateSource: require('../../../assets/variableIcons/icon_fill.png'),
       },
-      iconResourceName: 'sym_call_missed', // Android specific
-      iconResource: require('../../../assets/variableIcons/icon.png'),
       title: 'Tab2',
       orientation: 'landscape',
     },
@@ -127,13 +139,15 @@ const TAB_CONFIGS: TabConfiguration[] = [
         tabBarBlurEffect: 'none',
       },
       icon: {
-        imageSource: require('../../../assets/variableIcons/icon.png'),
+        shared: {
+          type: 'imageSource',
+          imageSource: require('../../../assets/variableIcons/icon.png'),
+        }
       },
       selectedIcon: {
+        type: 'imageSource',
         imageSource: require('../../../assets/variableIcons/icon_fill.png'),
       },
-      // iconResourceName: 'sym_action_email', // Android specific
-      iconResource: require('../../../assets/variableIcons/icon_fill.png'),
       title: 'Tab3',
       // systemItem: 'search', // iOS specific
       // systemItem: 'contacts', // iOS specific
@@ -146,13 +160,19 @@ const TAB_CONFIGS: TabConfiguration[] = [
     tabScreenProps: {
       tabKey: 'Tab4',
       icon: {
-        sfSymbolName: 'rectangle.stack',
+        ios: {
+          type: 'sfSymbolIOS',
+          name: 'rectangle.stack',
+        },
+        android: {
+          type: 'imageSource',
+          imageSource: require('../../../assets/svg/cart.svg'),
+        }
       },
       selectedIcon: {
-        sfSymbolName: 'rectangle.stack.fill',
+        type: 'sfSymbolIOS',
+        name: 'rectangle.stack.fill',
       },
-      // iconResourceName: 'sym_action_chat', // Android specific
-      iconResource: require('../../../assets/svg/cart.svg'),
       title: 'Tab4',
       systemItem: 'search', // iOS specific
       badgeValue: '123',
--- a/src/components/bottom-tabs/BottomTabsScreen.tsx
+++ b/src/components/bottom-tabs/BottomTabsScreen.tsx
@@ -4,6 +4,7 @@ import React from 'react';
 import { Freeze } from 'react-freeze';
 import {
   Image,
+  ImageResolvedAssetSource,
   StyleSheet,
   findNodeHandle,
   processColor,
@@ -12,7 +13,7 @@ import {
 } from 'react-native';
 import { freezeEnabled } from '../../core';
 import BottomTabsScreenNativeComponent, {
-  type IconType,
+  type NativeIconType,
   type NativeProps,
   type Appearance,
   type ItemAppearance,
@@ -25,7 +26,9 @@ import type {
   BottomTabsScreenItemStateAppearance,
   BottomTabsScreenProps,
   EmptyObject,
-  Icon,
+  PlatformIcon,
+  PlatformIconAndroid,
+  PlatformIconIOS,
 } from './BottomTabsScreen.types';
 import { bottomTabsDebugLog } from '../../private/logging';
 
@@ -55,7 +58,6 @@ function BottomTabsScreen(props: BottomTabsScreenProps) {
     isFocused = false,
     freezeContents,
     icon,
-    iconResource,
     selectedIcon,
     standardAppearance,
     scrollEdgeAppearance,
@@ -118,16 +120,6 @@ function BottomTabsScreen(props: BottomTabsScreenProps) {
 
   const iconProps = parseIconsToNativeProps(icon, selectedIcon);
 
-  let parsedIconResource;
-  if (iconResource) {
-    parsedIconResource = Image.resolveAssetSource(iconResource);
-    if (!parsedIconResource) {
-      console.error(
-        '[RNScreens] failed to resolve an asset for bottom tab icon',
-      );
-    }
-  }
-
   return (
     <BottomTabsScreenNativeComponent
       collapsable={false}
@@ -141,7 +133,8 @@ function BottomTabsScreen(props: BottomTabsScreenProps) {
       // It won't render any icon, but it will prevent from crashing on the native side which is expecting
       // ReadableMap. Passing `iconResource` directly will result in crash, because `require` API is returning
       // double as a value.
-      iconResource={parsedIconResource || undefined}
+      iconResource={iconProps.parsedIconResource || undefined}
+      iconResourceName={iconProps.drawableName}
       {...iconProps}
       standardAppearance={mapAppearanceToNativeProp(standardAppearance)}
       scrollEdgeAppearance={mapAppearanceToNativeProp(scrollEdgeAppearance)}
@@ -241,8 +234,35 @@ function shouldFreezeScreen(
   return !nativeViewVisible;
 }
 
-function parseIconToNativeProps(icon: Icon | undefined): {
-  iconType?: IconType;
+function parseAndroidIconToNativeProps(icon: PlatformIconAndroid | undefined): {
+  parsedIconResource?: ImageResolvedAssetSource;
+  drawableName?: string;
+} {
+  if (!icon) {
+    return {};
+  }
+
+  let parsedIconResource;
+  if (icon.type === 'imageSource') {
+    parsedIconResource = Image.resolveAssetSource(icon.imageSource);
+    if (!parsedIconResource) {
+      console.error(
+        '[RNScreens] failed to resolve an asset for bottom tab icon',
+      );
+    }
+  }
+
+  const drawableName =
+    icon.type === 'drawableResourceAndroid' ? icon.name : undefined;
+
+  return {
+    parsedIconResource,
+    drawableName,
+  };
+}
+
+function parseIOSIconToNativeProps(icon: PlatformIconIOS | undefined): {
+  iconType?: NativeIconType;
   iconImageSource?: ImageSourcePropType;
   iconSfSymbolName?: string;
 } {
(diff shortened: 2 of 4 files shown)


FabricExample/android/app/src/main/res/drawable/custom_home_icon.xml
kligarski, line 1: Don't we need this icon also in Paper example?

apps/src/tests/TestBottomTabs/index.tsx
```diff
@@ -96,13 +102,19 @@ const TAB_CONFIGS: TabConfiguration[] = [
...
       tabBarItemBadgeBackgroundColor: Colors.GreenDark100,
       icon: {
-        templateSource: require('../../../assets/variableIcons/icon.png'),
+        ios: {
+          type: 'templateSourceIOS',
+          templateSource: require('../../../assets/variableIcons/icon.png'),
+        }, 
+        android: {
+          type: 'drawableResourceAndroid',
+          name: 'sym_call_missed', // Android specific
```
kligarski:
```suggestion
          name: 'sym_call_missed',
```
I think those comments were meant to describe which props are related to which platform but now it's redundant.

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -321,6 +332,41 @@ export interface BottomTabsScreenProps {
    * @platform android, ios
    */
   badgeValue?: string;
+  /**
+   * @summary Specifies the icon for the tab bar item.
+   *
+   * You can define an icon separately for each platform (iOS and Android)
+   * or define a default icon in shared. The shared icon will be used on both
```
kligarski:
```suggestion
   * You can define an icon separately for each platform (in `ios` and `android`)
   * or define a default icon in `shared`. The shared icon will be used on both
```

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -321,6 +332,41 @@ export interface BottomTabsScreenProps {
...
+   * platforms unless it is overridden by a platform-specific definition
+   * in ios or android.
```
kligarski:
```suggestion
   * in `ios` or `android`.
```

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -321,6 +332,41 @@ export interface BottomTabsScreenProps {
...
+   * - `{ type: 'templateSourceIOS', templateSource }`
+   *   Uses the provided image as a template image.
+   *   The icon color will depend on the current state
+   *   of the tab bar item and icon color-related props.
+   *
+   * Android-only:
+   * - `{ type: 'drawableResourceAndroid', name }`
+   *   Uses a drawable resource with the given name.
+   *
+   *   Remarks: Requires passing a drawable to resources via AndroidStudio.
```
kligarski:
```suggestion
   *   Remarks: Requires passing a drawable to resources via Android Studio.
```

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -321,6 +332,41 @@ export interface BottomTabsScreenProps {
...
+   *
+   * If no `selectedIcon` is provided, this icon will also be used
+   * as the selected state icon.
```
kligarski:
```suggestion
   * On iOS, if no `selectedIcon` is provided, this icon will also
   * be used as the selected state icon.
```

FabricExample/android/app/src/main/res/drawable/custom_home_icon.xml
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3214/commits/6f1c731d628418859e195879688506f8e256f80f

src/fabric/bottom-tabs/BottomTabsScreenNativeComponent.ts
```diff
@@ -18,7 +18,7 @@ import {
 import { UnsafeMixed } from './codegenUtils';
 
 // iOS-specific: SFSymbol, image as a template usage
-export type IconType = 'image' | 'template' | 'sfSymbol';
+export type NativeIconType = 'image' | 'template' | 'sfSymbol';
```
kkafar: `IconType` was fine here. There is nothing "native" about this type.

If you need to rename it to avoid name conflict somewhere else use [alias](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) when importing

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -17,23 +17,34 @@ export type LifecycleStateChangeEvent = Readonly<{
...
-}
+};
 
-// iOS-specific: image as a template usage
-export interface TemplateIcon {
-  templateSource: ImageSourcePropType;
-}
+export type PlatformIconIOS =
+  | {
+      type: 'sfSymbolIOS';
```
kkafar:
```suggestion
      type: 'sfSymbol';
```

`sfSymbols` are already iOS specific

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -17,23 +17,34 @@ export type LifecycleStateChangeEvent = Readonly<{
...
+      name: string;
+    }
+  | {
+      type: 'templateSourceIOS';
```
kkafar:
```suggestion
      type: 'templateSource';
```

same here

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -17,23 +17,34 @@ export type LifecycleStateChangeEvent = Readonly<{
...
+      templateSource: ImageSourcePropType;
+    }
+  | PlatformIconShared;
 
-// iOS-specific: SFSymbol, image as a template usage
-export type Icon = SFIcon | ImageIcon | TemplateIcon;
+export type PlatformIconAndroid =
+  | {
+      type: 'drawableResourceAndroid';
```
kkafar:
```suggestion
      type: 'drawableResource';
```

src/components/bottom-tabs/BottomTabsScreen.tsx
```diff
@@ -241,57 +228,95 @@ function shouldFreezeScreen(
...
+
   const { iconImageSource, iconSfSymbolName, iconType } =
-    parseIconToNativeProps(icon);
+    parseIOSIconToNativeProps(icon?.ios || icon?.shared);
   const {
     iconImageSource: selectedIconImageSource,
     iconSfSymbolName: selectedIconSfSymbolName,
     iconType: selectedIconType,
-  } = parseIconToNativeProps(selectedIcon);
+  } = parseIOSIconToNativeProps(selectedIcon);
```
kkafar: Shouldn't we just parse here props appropriate to platform we're running on?

src/fabric/bottom-tabs/BottomTabsScreenNativeComponent.ts
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3214/commits/3c38b2a904c8d474b1d2e7273169fc151bfeabaa

src/components/bottom-tabs/BottomTabsScreen.types.ts
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3214/commits/fd9cbb3d156eaac5c3056abdee92182706ab15ee

src/components/bottom-tabs/BottomTabsScreen.types.ts
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3214/commits/7b64747b9b4ae7d7554f442640428d491a3ceef6

src/components/bottom-tabs/BottomTabsScreen.types.ts
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3214/commits/a8488d96f6fa672218b64b917d61d77167813b7f

src/components/bottom-tabs/BottomTabsScreen.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3214/commits/b9238511e37ffda2ba5bdd7c91d1ae3c3290f1ba

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -321,6 +332,41 @@ export interface BottomTabsScreenProps {
...
+   * in `ios` or `android`.
+   *
+   * Supported values:
+   *
+   * Shared (both iOS and Android):
+   * - `{ type: 'imageSource', imageSource }`
+   *   Uses an image from the provided resource.
+   *
+   *   Remarks: `imageSource` type doesn't support SVGs on Android.
+   *   For loading SVGs use `drawableResource` type.
```
MrFunctor, line 322: Does this mean that passing an SVG image as `imageSource` will not work anymore?

Currently, passing an SVG icon like this works on Android, which is very handy:

```tsx
<Icon
	src={{
		default: require('./icon.svgx'),
	}}
/>
```

If only drawables are supported on Android, how are they supposed to be added when using a managed Expo workflow with CNG? Will there be an official config plugin to manage drawables? Either way, it seems like adding unnecessary complexity for the common use case of using SVG images as tab bar icons on Android.

kkafar: Hey! This PR does not change that, it only better documents current state of the lib. 

The changes have already landed in https://github.com/software-mansion/react-native-screens/pull/3216.

With latest stable 4.16 version the svgs are supported as we used `coil` lib for loading the resources, however we run into some issues, which led us to back out from that usage, migrating to `fresco` which is managed by `react-native` installation directly (from our PoV it comes bundled with the app), which does not have svg support.

kkafar: We're in touch with expo-router team, so that they're aware of this current limitation & can take appropriate steps on their end.

kkafar:
Can we update this PR after #2987 landed? To reuse existing types & unify the approach.

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -17,23 +17,34 @@ export type LifecycleStateChangeEvent = Readonly<{
...
+  | PlatformIconShared;
 
-// iOS-specific: SFSymbol, image as a template usage
-export type Icon = SFIcon | ImageIcon | TemplateIcon;
+export type PlatformIconAndroid =
+  | {
+      type: 'drawableResource';
+      name: string;
+    }
+  | PlatformIconShared;
```
satya164: I'd propose merging `imageSource` and `templateSource` for better ergonomics:

```ts
type ImageSourceIcon = {
  type: 'imageSource',
  source: ImageSourcePropType,
  /**
   * Whether to use the icon as a template
   * The image will respect tint color.
   * Defaults to `true`.
   *
   * @platform ios
   */
  template?: boolean
}
```

Why:
- We pass a `ImageSourcePropType` value, so calling it `type: 'imageSource'` feels more natural
- The `sfSymbol` type doesn't include a `sfSymbol` property, but a name property. Similarly `source` property is shorter and nicer than duplicating the type for the property
- On Android, images always respect the tint color. So having the same type that works on both in the same way is nicer. Otherwise, we'd need to do `Platform.select({ android: { type: 'imageSource' ... }, ios: { type: 'templateSource' ... })` most of the time for no benefit
- The additional iOS-specific `template` property still respects iOS nomenclature of calling it "template"

kkafar: What do you mean here by *"On Android, images always respect the tint color."*?

kkafar: Besides that I do not get fully that one part ☝🏻 I see your argumentation here & I think you make a good point, however I'm still sceptical, because on iOS, while you can specify any image for the template, it won't be handled nicely. If you just specify "regular" image it will get turned into colourful (tinted) rectangle & I like the separation from "regular image" here.

kkafar: I'll try to come up with something nice in few minutes

kkafar: Also you don't need to use `Platform.select` right now. It is enough to just specify icon for Android & icon for iOS. I'm gonna split the component in two tomorrow & approach with `template?: boolean` won't have any sense. I think we should stay with what we have now.

src/components/bottom-tabs/BottomTabsScreen.tsx
```diff
@@ -25,9 +27,13 @@ import type {
...
   BottomTabsScreenProps,
   EmptyObject,
-  Icon,
 } from './BottomTabsScreen.types';
 import { bottomTabsDebugLog } from '../../private/logging';
+import type {
+  PlatformIcon,
+  PlatformIconAndroid,
+  PlatformIconIOS,
+} from 'react-native-screens/types';
```
kkafar: This should use relative import.

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -1,11 +1,11 @@
 import type {
   ColorValue,
-  ImageSourcePropType,
   NativeSyntheticEvent,
   TextStyle,
   ViewProps,
 } from 'react-native';
 import { ScrollEdgeEffect } from '../shared/types';
+import type { PlatformIcon, PlatformIconIOS } from 'react-native-screens/types';
```
kkafar: Also relative import here

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3214/commits/10032eb3eedf3baa66c89c3f938ffd4a3468517f

src/components/bottom-tabs/BottomTabsScreen.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3214/commits/b2a8a6702aa5b90f7b3eba4b4c4f070dcfaf79c1

src/components/bottom-tabs/BottomTabsScreen.types.ts
satya164: > What do you mean here by "On Android, images always respect the tint color."

I don't have full details on how the API works, for this was my experience (only tried with the old API in bottom tabs):

- iOS: providing an image in `imageSource` didn't respect `tabBarTintColor`, `templateSource` did
- Android: providing an image in `iconResource` respected `tabBarItemIconColorActive`, similar to `templateSource` on iOS

> It is enough to just specify icon for Android & icon for iOS

Gotcha, but I was trying to demonstrate that in most cases when you pass an icon for tab bar, buttons etc., you're likely passing a monochrome png with transparency that should take on the tint color, and passing the same format one time is simpler instead of passing separate objects for Android & iOS that do the same thing in the end.

i.e. instead of `icon: { shared: { type: 'imageSource', source: require('./my-icon.png') } }`, I'll need `icon: { android: { type: 'imageSource', source: require('./my-icon.png') }, ios: { type: 'templateSource', source: require('./my-icon.png') } }`

kkafar: Yes, but we then include `template?: boolean`, which is iOS specific, to `shared` icon object, which I do not like. I'd like to leave it in current form on react-native-screens side. We can surely wrap this and do otherwise on RN side.

kkafar (approved):
lgtm

src/components/bottom-tabs/BottomTabsScreen.types.ts
oblador: @MrFunctor You can use my lib for using svgs as drawables: https://github.com/oblador/react-native-vector-image

MrFunctor: Hey @oblador, I learned about your lib back then, but it hasn't been updated in a while at the time, so I went with a custom solution, which involves custom config plugins that copy the native assets. Anyway, it's great to see that your lib has been updated now!

PR #3215, opened 2025-09-15
feat(Android, Tabs): safe area component for Android

kligarski:
## Description

Adds implementation for handling safe area on Android for bottom tabs.

Implementation has been adapted from [`react-native-safe-area-context`](https://github.com/AppAndFlow/react-native-safe-area-context).

TODO:
- [x] change native layout so that TabScreen renders under the tab bar (to allow using transparent tab bar)
- [x] check implementation for older Android APIs (I checked APIs: 25, 28, 29, 30, 36 and 24 on Paper)
- [x] add simple API (for all edges) to control which types of insets are used by `SafeAreaView` component (system and/or interface bars)
- [ ] [Separate PR] add per-edge API to control which types of insets are used by `SafeAreaView` component (system and/or interface bars) (https://github.com/software-mansion/react-native-screens-labs/issues/434)
- [ ] [[Separate PR](https://github.com/software-mansion/react-native-screens/pull/3240)] fix interaction with Stack v4 (`CustomToolbar`) (https://github.com/software-mansion/react-native-screens-labs/issues/435)
- [ ] [Separate PR] There were no problems with using margins for now so I left it as it was. We can change this later to padding or add a prop to switch between padding and margins on both Android and iOS. (https://github.com/software-mansion/react-native-screens-labs/issues/436)

https://github.com/user-attachments/assets/c59af116-a653-40e3-9345-fe8d3ba170ee

### Transparent tab bar
| `top: false, bottom: false` | `top: true, bottom: false` | `top: true, bottom: true` |
| --- | --- | --- |
| Screenshot_20250916_091258 | Screenshot_20250916_091303 | Screenshot_20250916_091311 |

#### SafeAreaView
After internal discussion about approach to `SafeAreaView`, we had following conclusions:
- as edge-to-edge becomes desirable (and is the default for apps targeting Android SDK 35 or above), and to simplify layout handling, we want the `Screen`s of our navigation containers (e.g. `StackScreen`, `BottomTabsScreen`) to have **full dimensions of their parents, even if it means that they will be laid out behind navigation bars** (header in Stack, tab bar),
- `SafeAreaView` will provide unified way to handle the safe area,
- on Android, we want to control which insets we want to handle:
  - **system insets** (received from `onApplyWindowInsets`), e.g. `systemBars`, `displayCutout`
  - **interface insets** - custom insets from navigation bars, e.g. `bottomNavigationView`

#### Before this PR
Prior to this PR, we were using `LinearLayout` for `TabsHost`:

```kotlin
class TabsHost(
    val reactContext: ThemedReactContext,
) : LinearLayout(reactContext),
    TabScreenDelegate {
    // ...
    private val bottomNavigationView: BottomNavigationView =
        BottomNavigationView(wrappedContext).apply {
            layoutParams = LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.WRAP_CONTENT)
        }

    private val contentView: FrameLayout =
        FrameLayout(reactContext).apply {
            layoutParams =
                LinearLayout
                    .LayoutParams(
                        LayoutParams.MATCH_PARENT,
                        LayoutParams.WRAP_CONTENT,
                    ).apply {
                        weight = 1f
                    }
            id = ViewIdGenerator.generateViewId()
        }
    // ...
}
```

Screenshot 2025-09-24 at 17 23 20

This approach had following problems:
- `contentView` did not have dimensions of its parent, which is not what we wanted,
- Yoga wasn't aware of `contentView`'s height - all content inside `TabScreen` was laid out as if the screen had full height of its parent -> this meant that `contentView`'s dimensions and actual content dimensions were not in sync,
- it did not support using translucent tab bar - screen's content was cut off outside of `contentView`'s bounds.

3215_before_transparent

#### Approach in this PR
In this PR, we change `TabsHost`'s layout to `FrameLayout` which allows multiple views placed on top of each other - this is what we want to achieve (tab bar floating over content, attached to the bottom).

To attach `bottomNavigationView` to *the bottom*, we use `Gravity.BOTTOM`.

```kotlin
class TabsHost(
    val reactContext: ThemedReactContext,
) : FrameLayout(reactContext),
    TabScreenDelegate,
    SafeAreaProvider,
    View.OnLayoutChangeListener {
    // ...
    private val bottomNavigationView: BottomNavigationView =
        BottomNavigationView(wrappedContext).apply {
            layoutParams =
                LayoutParams(
                    LayoutParams.MATCH_PARENT,
                    LayoutParams.WRAP_CONTENT,
                    Gravity.BOTTOM,
                )
        }

    private val contentView: FrameLayout =
        FrameLayout(reactContext).apply {
            layoutParams =
                LayoutParams(
                    LayoutParams.MATCH_PARENT,
                    LayoutParams.MATCH_PARENT,
                )
            id = ViewIdGenerator.generateViewId()
        }
    // ...
}
```

Screenshot 2025-09-24 at 17 23 29

Now, we can:
1. use opaque or translucent `bottomNavigationView`,
2. use `SafeAreaView` to control how much space can the actual content take (do we allow it to render under `bottomNavigationView`).

3215_after_transparent

### Support for older Android versions
On Android versions prior to R, insets dispatch is broken (children of ViewGroup receive insets from previous child; they should all receive the same insets). That's why we need to override `dispatchApplyWindowInsets` implementation in `TabsHost`. In `ViewGroup`'s implementation of this method, `View`'s implementation is used (via `super`) - we can't access this directly. Unfortunately, `View`'s implementation sets some private flags which are used by default `onApplyWindowInsets` implementation. If we try to use `onApplyWindowInsets` on API 28 without setting the private flag, application goes into infinite loop (`fitSystemWindows` calls `dispatchApplyWindowInsets` -> we might want to investigate this in more detail). As we don't use insets in `TabsHost`, I decided not to call `onApplyWindowInsets` in `TabsHost` at all. I haven't found any problems with it yet.

## Changes

- add implementation for `SafeAreaView` and related classes for both architectures
- add `SafeAreaProvider` interface and implement it for `TabsHost`
- change `TabsHost` to use `FrameLayout`:
  - make `TabsScreen` layout behind tab bar (take full available height to match JS screen)
- override `dispatchApplyWindowInsets` in `TabsHost` in order to fix insets for older Android versions
- add `insetType` prop to control what kind of insets should SafeAreaView respect (all, only system, only interface)

Run `TestBottomTabs` with uncommented SafeAreaView. You can change which edges are enabled and add `insetType` prop.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/safearea/SafeAreaProvider.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/safearea/SafeAreaProvider.kt
@@ -1,9 +1,36 @@
 package com.swmansion.rnscreens.safearea
 
+/**
+ * Allows containers that obscure some part of its children views to provide safe area.
+ *
+ * This protocol only handles **interface insets** (e.g. toolbar, bottomNavigationView).
+ * System insets (e.g. `systemBars`, `displayCutout`) are handled through Android inset
+ * dispatch mechanism (via `onApplyWindowInsets`).
+ *
+ * Classes implementing this protocol are responsible for notifying `SafeAreaView`, which
+ * registers as a listener, about changes to **interface** safe area insets.
+ */
 interface SafeAreaProvider {
+    /**
+     * Responsible for registering **interface** safe area insets listener.
+     *
+     * @param listener `SafeAreaView` instance that wants to receive notifications
+     * about changes to **interface** safe area insets via `onInterfaceInsetsChange`.
+     */
     fun setOnInterfaceInsetsChangeListener(listener: SafeAreaView)
 
+    /**
+     * Responsible for unregistering **interface** safe area insets listener.
+     *
+     * @param listener `SafeAreaView` instance that wants to stop receiving notifications
+     * about changes to **interface** safe area insets.
+     */
     fun removeOnInterfaceInsetsChangeListener(listener: SafeAreaView)
 
+    /**
+     * Responsible for providing current **interface** safe area insets.
+     *
+     * @returns `EdgeInsets` describing the current **interface** safe area insets.
+     */
     fun getInterfaceInsets(): EdgeInsets
 }


android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
```diff
@@ -418,12 +430,72 @@ class TabsHost(
...
+        left: Int,
+        top: Int,
+        right: Int,
+        bottom: Int,
+        oldLeft: Int,
+        oldTop: Int,
+        oldRight: Int,
+        oldBottom: Int,
+    ) {
+        require(view is BottomNavigationView)
```
t0maboro: I'd prefer using `fun require(value: Boolean, lazyMessage: () -> Any)` with some additional info about the class we got

apps/src/tests/TestBottomTabs/tabs/Tab4.tsx
```diff
@@ -108,13 +109,10 @@ function Screen3({ navigation }: StackNavigationProp) {
...
   return (
+    // <SafeAreaView edges={{ top: true, bottom: true }}>
     <NavigationContainer>
       <Stack.Navigator>
-        <Stack.Screen
-          name="Screen1"
-          component={Screen1}
-          options={{ headerTransparent: true, orientation: 'landscape' }}
-        />
+        <Stack.Screen name="Screen1" component={Screen1} />
```
t0maboro: nit: this doesn't seem to be changed as a part of this PR

kligarski: https://github.com/software-mansion/react-native-screens/pull/3215/commits/5015eb0e5d9be2fdf1e371bd52f61c9c4e11874c

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3215/commits/57afd2414463d7b610f2afaf0153fab39cfc793b

kligarski:
Currently, `SafeAreaView` on Android is implemented only for bottom tabs. We're planning to add a flag to enable `SafeAreaView` globally in our components very soon so I'm not sure if I should create any test screen for this PR? (It would be very similar to current TestBottomTabs).

Let me know what you think.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
```diff
@@ -1,25 +1,34 @@
...
+import com.swmansion.rnscreens.safearea.SafeAreaProvider
+import com.swmansion.rnscreens.safearea.SafeAreaView
 import com.swmansion.rnscreens.utils.RNSLog
 import kotlin.properties.Delegates
 
 class TabsHost(
     val reactContext: ThemedReactContext,
-) : LinearLayout(reactContext),
-    TabScreenDelegate {
+) : FrameLayout(reactContext),
```
kkafar, line 28: The PR lacks description of this change of inheritance from `LinearLayout` to `FrameLayout`. I'd like it to be mentioned & described. Why has it been done, how did the layout look before & how it looks now.

android/src/main/java/com/swmansion/rnscreens/safearea/EdgeInsets.kt
```diff
@@ -0,0 +1,35 @@
...
+package com.swmansion.rnscreens.safearea
+
+import androidx.core.graphics.Insets
+import kotlin.math.max
+
+data class EdgeInsets(
+    val top: Float,
+    val right: Float,
+    val bottom: Float,
+    val left: Float,
```
kkafar: These properties are defined against Android convention of `left, top, right, bottom`. 

See `onLayout` methods, or `setPadding`. I'd like to align that.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
```diff
@@ -418,12 +430,74 @@ class TabsHost(
...
+
+    override fun removeOnInterfaceInsetsChangeListener(listener: SafeAreaView) {
+        if (interfaceInsetsChangeListener == listener) {
+            interfaceInsetsChangeListener = null
+            bottomNavigationView.removeOnLayoutChangeListener(this)
+        }
+    }
+
+    override fun getInterfaceInsets(): EdgeInsets =
+        EdgeInsets(0.0f, 0.0f, (bottomNavigationView.bottom - bottomNavigationView.top).toFloat(), 0.0f)
```
kkafar: There is `height` property I think 😄

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
```diff
@@ -418,12 +430,74 @@ class TabsHost(
...
+        top: Int,
+        right: Int,
+        bottom: Int,
+        oldLeft: Int,
+        oldTop: Int,
+        oldRight: Int,
+        oldBottom: Int,
+    ) {
+        require(view is BottomNavigationView) {
+            "[RNScreens] TabsHost's onLayoutChange expects BottomNavigationView, received ${if (view != null) view::class.java else "null"} instead"
```
kkafar: This is fine, no need to change it, just a tip: if you pass "${view}" it will format it nicely anyway. Potentially with more information.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
```diff
@@ -418,12 +430,74 @@ class TabsHost(
...
+        }
+
+        val oldHeight = oldBottom - oldTop
+        val newHeight = bottom - top
+
+        if (newHeight != oldHeight) {
+            interfaceInsetsChangeListener?.apply {
+                this.onInterfaceInsetsChange(EdgeInsets(0.0f, 0.0f, newHeight.toFloat(), 0.0f))
+            }
+        }
```
kkafar, line 497: Yeah, we're interested only in height changes. It's ok.

android/src/main/java/com/swmansion/rnscreens/safearea/paper/SafeAreaViewEdges.kt
```diff
@@ -0,0 +1,32 @@
...
+// https://github.com/AppAndFlow/react-native-safe-area-context/tree/v5.6.1
+package com.swmansion.rnscreens.safearea.paper
+
+import com.facebook.react.bridge.ReadableMap
+
+data class SafeAreaViewEdges(
+    val top: Boolean,
+    val right: Boolean,
+    val bottom: Boolean,
+    val left: Boolean,
```
kkafar: Similar remark to the property order here.

android/src/main/java/com/swmansion/rnscreens/safearea/EdgeInsets.kt
```diff
@@ -0,0 +1,35 @@
...
+) {
+    companion object {
+        val NONE: EdgeInsets = EdgeInsets(0.0f, 0.0f, 0.0f, 0.0f)
```
kkafar:
```suggestion
        val ZERO: EdgeInsets = EdgeInsets(0.0f, 0.0f, 0.0f, 0.0f)
```

android/src/main/java/com/swmansion/rnscreens/safearea/SafeAreaProvider.kt
```diff
@@ -0,0 +1,9 @@
+package com.swmansion.rnscreens.safearea
+
+interface SafeAreaProvider {
+    fun setOnInterfaceInsetsChangeListener(listener: SafeAreaView)
+
+    fun removeOnInterfaceInsetsChangeListener(listener: SafeAreaView)
+
+    fun getInterfaceInsets(): EdgeInsets
```
kkafar, line 35: Okay, I think that this interface & this method deserve a nice doc comment, describing that it is responsible solely for "interface insets" (right?) and "system" insets are handled through Android dispatch mechanism. Right?

android/src/main/java/com/swmansion/rnscreens/safearea/SafeAreaView.kt
```diff
@@ -1,13 +1,264 @@
...
+    private var insetType: InsetType = InsetType.ALL
+
+    fun getStateWrapper(): StateWrapper? = stateWrapper
+
+    fun setStateWrapper(stateWrapper: StateWrapper?) {
+        this.stateWrapper = stateWrapper
+    }
+
+    init {
+        ViewCompat.setOnApplyWindowInsetsListener(this, this)
```
kkafar, line 50: Why are we using insets listener here, instead of regular onApplyWindowInsets method?

android/src/main/java/com/swmansion/rnscreens/safearea/SafeAreaView.kt
```diff
@@ -1,13 +1,264 @@
...
+    }
+
+    override fun onAttachedToWindow() {
+        viewTreeObserver.addOnPreDrawListener(this)
+
+        val newProvider = findProvider()
```
kkafar:
```suggestion
        val newProvider = findAncestorProvider()
```

android/src/main/java/com/swmansion/rnscreens/safearea/SafeAreaView.kt
```diff
@@ -1,13 +1,264 @@
...
+        val startTime = System.nanoTime()
+        var waitTime = 0L
+        getReactContext(this).runOnNativeModulesQueueThread {
+            lock.withLock {
+                if (!done) {
+                    done = true
+                    condition.signal()
+                }
+            }
+        }
```
kkafar, line 208: hate this. Let's keep our fingers crossed we won't run into problems with this.

android/src/main/java/com/swmansion/rnscreens/safearea/SafeAreaView.kt
```diff
@@ -1,13 +1,264 @@
...
+        // does not block waiting on the main thread.
+        var done = false
+        val lock = ReentrantLock()
+        val condition = lock.newCondition()
+        val startTime = System.nanoTime()
+        var waitTime = 0L
+        getReactContext(this).runOnNativeModulesQueueThread {
+            lock.withLock {
+                if (!done) {
+                    done = true
```
kkafar, line 204: We're modifying here non-volatile value possible across thread boundaries 🙈 

I won't demand changes here, since this code is already a bit battle tested, but man...

kkafar (changes requested):
Okay. The PR looks really nice. I have few requests, please answer them.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
kligarski: I updated the PR description.

android/src/main/java/com/swmansion/rnscreens/safearea/SafeAreaView.kt
kligarski: I used it to get `WindowInsetsCompat` instance instead of getting `WindowInsets` and converting it manually.

android/src/main/java/com/swmansion/rnscreens/safearea/EdgeInsets.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3215/commits/d6c47df4ca245b3260655ebdd4f2b04e95553e90

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3215/commits/d6c47df4ca245b3260655ebdd4f2b04e95553e90

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3215/commits/d6c47df4ca245b3260655ebdd4f2b04e95553e90

android/src/main/java/com/swmansion/rnscreens/safearea/paper/SafeAreaViewEdges.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3215/commits/d6c47df4ca245b3260655ebdd4f2b04e95553e90

android/src/main/java/com/swmansion/rnscreens/safearea/EdgeInsets.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3215/commits/d6c47df4ca245b3260655ebdd4f2b04e95553e90

android/src/main/java/com/swmansion/rnscreens/safearea/SafeAreaView.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3215/commits/d6c47df4ca245b3260655ebdd4f2b04e95553e90

android/src/main/java/com/swmansion/rnscreens/safearea/SafeAreaProvider.kt
kligarski: That's correct.

https://github.com/software-mansion/react-native-screens/pull/3215/commits/6942ab4407aee10c92592ca30ae68a28d3830b16

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
kkafar: Really good job there. Description is looking good right now.

android/src/main/java/com/swmansion/rnscreens/safearea/SafeAreaView.kt
```diff
@@ -1,13 +1,264 @@
+// Implementation adapted from `react-native-safe-area-context`:
+// https://github.com/AppAndFlow/react-native-safe-area-context/tree/v5.6.1
 package com.swmansion.rnscreens.safearea
 
 import android.annotation.SuppressLint
+import android.util.Log
```
kkafar, line 6: Seeing the "android.util.log" import here I think it might be worth running `Cmd+shift+A > optimize imports`

kligarski: It is used here: https://github.com/software-mansion/react-native-screens/pull/3215/files/6942ab4407aee10c92592ca30ae68a28d3830b16#diff-00e591b82e104e7285a4b72555170495318548057292e0ead09ed0a1d8cced47R222.

PR #3219, opened 2025-09-17
feat(iOS, Stack): Add controller invalidating to new stack impl

t0maboro:
## Description

TODO:
- [ ] verify if `RCT_NEW_ARCH_ENABLED` flag is really needed - I'm having issues with symbols, without it
- [ ] verify if the following transactions in the screen component is correct, or rather, the host should take care of the children

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/127

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


kligarski (approved):

kkafar (changes requested):
no, wait. I've hurried with the approval. 

Give me few minutes, I'll write another review

ios/gamma/stack/RNSStackScreenComponentView.mm
```diff
@@ -116,6 +138,34 @@ + (BOOL)shouldBeRecycled
...
+
+- (void)mountingTransactionWillMount:(const facebook::react::MountingTransaction &)transaction
+                withSurfaceTelemetry:(const facebook::react::SurfaceTelemetry &)surfaceTelemetry
+{
+  for (const auto &mutation : transaction.getMutations()) {
+    if ([self shouldInvalidateOnMutation:mutation]) {
+      [RNSViewControllerInvalidator invalidateViewIfDetached:self forRegistry:_invalidatedComponentsRegistry];
+    }
+  }
+}
```
kkafar: First thing is: why do we have to scan whole list of mutations *in every screen component*? Why can't we just do it once per transaction in host component?

ios/gamma/stack/RNSStackScreenComponentView.mm
```diff
@@ -18,6 +23,9 @@ @interface RNSStackScreenComponentView () <RCTMountingTransactionObserving>
 @implementation RNSStackScreenComponentView {
   RNSStackScreenController *_Nonnull _controller;
   RNSStackScreenComponentEventEmitter *_Nonnull _reactEventEmitter;
+#ifdef RCT_NEW_ARCH_ENABLED
+  RNSInvalidatedComponentsRegistry *_Nonnull _invalidatedComponentsRegistry;
```
kkafar: Why do we keep "whole registry" here, when only this one single instance can be pushed there?

kkafar (changes requested):

ios/gamma/stack/RNSStackScreenComponentView.mm
t0maboro: `flushInvalidViews` is triggered when the component is moved out of the window. In such a scenario, the Host will keep all Screens in its registry - we don't know for how long (if the Screen cannot invalidate itself immediately on the mutation, it needs to mark that the mutation came and invalidate on `willMoveToWindow:nil`)

t0maboro:
> Hey, but does it work reliably? What if you pop a screen through JS? Won't it be prematurely invalidated?

I was testing it on `TestScreenStack` for:
- native back button
- swipe to dismiss
- 'pop' action

scenarios and all of them are determined by incoming `Delete` mutation so imo it works well as we have a guarantee that the component won't be mounted again.

ios/gamma/stack/RNSStackScreenComponentView.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3219/commits/c8c74ed9b8399abec27f9df992463f7262273a89 - the registry was used for reusability of `RNSViewControllerInvalidator`, but in this case, keeping just private flag is fine

ios/gamma/stack/RNSStackScreenComponentView.mm
kkafar: 🤔 I don't have good suggestion here unfortunately, but I dislike this solution for it's excessive (hopefully) transaction scanning. I'll try to come up with something better soon. We can discuss this on Monday.

t0maboro:
Switching to draft as we can follow the new approach with https://github.com/facebook/react-native/commit/b1994dd547395dbeec868e56762b5de2202cb1a9

ios/gamma/stack/RNSStackScreenComponentView.mm
t0maboro: @kkafar - I'm updating this PR with a new approach based on https://github.com/facebook/react-native/commit/b1994dd547395dbeec868e56762b5de2202cb1a9

kkafar (approved):
Looks good. Thank you.

PR #3220, opened 2025-09-17
chore: Configure OIDC for nightly builds

kmagiera:
## Description

Recently npm enabled an option to use OIDC for publishing package artifacts from Github Actions: https://github.blog/changelog/2025-07-31-npm-trusted-publishing-with-oidc-is-generally-available/

This PR follows the configuration instructions, namely:
1) We add "permissions" for "id-token" which permits the action to retrieve and temporarily store the OIDC token
2) We update npm version to latest in order to use a version that supports OIDC
3) We delete the env variable for the NPM access token we used before. Now the token is provided by OIDC mechanism
4) We're keeping provenance flag for NPM publish despite the fact it is forced when OIDC is enabled

The remaining part of the configuration is already setup on the NPM site and we should expect this workflow to "just work" after the changes.

### DIFF
--- a/.github/workflows/npm-screens-publish-nightly.yml
+++ b/.github/workflows/npm-screens-publish-nightly.yml
@@ -5,13 +5,16 @@ on:
   schedule:
     - cron: '27 23 * * *' # at 23:27 every day
   workflow_dispatch:
+permissions:
+  id-token: write
+  contents: read
 jobs:
   npm-build:
     if: github.repository == 'software-mansion/react-native-screens'
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/npm-screens-publish-nightly.yml
```diff
@@ -5,13 +5,16 @@ on:
   schedule:
     - cron: '27 23 * * *' # at 23:27 every day
   workflow_dispatch:
+permissions:
+  id-token: write
+  contents: read
```
kmagiera:
```suggestion
```

PR #3222, opened 2025-09-17
chore: create `react-native-screens/experimental` submodule

kkafar:
## Description

Issue: https://github.com/software-mansion/react-native-screens-labs/issues/443

Currently, we "publish" our experimental APIs through main `index` file.
We just annotate them with [appropriate comments], counting that people will notice. 

Example: [ `BottomTabs` component definition ](https://github.com/software-mansion/react-native-screens/blob/86b8fefc23637493ae76becea480c5f2eed81de2/src/components/bottom-tabs/BottomTabs.tsx#L19-L22)
Example: [ `BottomTabs` component reexport in `index` ](https://github.com/software-mansion/react-native-screens/blob/86b8fefc23637493ae76becea480c5f2eed81de2/src/index.tsx#L62-L65)

This is a poor solution & we should do at least a bit better. 

## Changes

Create `react-native-screens/experimental` package, just as there is `react-native-screens/private`,
and export new, experimental symbols only from there. 

When stabilized, symbols should be moved from `react-native-screens/experimental` to `react-native-screens`.

There should be no change in behaviour. New files should be published with the package. Run `npm pack --dry-run` to verify this.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/experimental/package.json
+++ b/experimental/package.json
@@ -1,0 +1,7 @@
+{
+  "main": "../lib/commonjs/experimental/index",
+  "module": "../lib/module/experimental/index",
+  "react-native": "../src/experimental/index",
+  "types": "../lib/typescript/experimental/index"
+}
+
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "gesture-handler/",
     "reanimated/",
     "private/",
+    "experimental/",
     "android/src/main/AndroidManifest.xml",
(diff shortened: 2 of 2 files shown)


t0maboro (approved):
tested together with https://github.com/software-mansion/react-native-screens/pull/3223

kmichalikk (approved):

PR #3223, opened 2025-09-17
chore: move gamma symbols into experimental package

kkafar:
## Description

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/443

Follow up to https://github.com/software-mansion/react-native-screens/pull/3222

## Changes

Moved symbols (including types) related to `StackScreen`, `ScreenStackHost`, `SplitViewHost`, `SplitViewScreen` to
`react-native-screens/experimental` submodule.

Updated necessary imports to keep example apps working.

Should not modify runtime behaviour. Example apps should run.

## Checklist

N/A


src/experimental/index.ts
```diff
@@ -0,0 +1,17 @@
...
+
+// Types
+export * from './types';
+
+// Components
+
+export { default as ScreenStackHost } from '../components/gamma/ScreenStackHost';
+export {
+  default as StackScreen,
+  StackScreenLifecycleState,
```
t0maboro, line 14: (non-blocking) it should be moved to types, but keeping in mind that we should extract types for the new stack like we did for other components, it's fine to just create a ticket for covering `StackScreen.types`

t0maboro (approved):
I like this idea, I sideloaded a tarball to the app from the template and I can import symbols properly

src/experimental/index.ts
kkafar: I'm already working on it. There will be a follow-up PR.

t0maboro: ack

src/experimental/index.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * ALL SYMBOLS EXPOSED FROM THIS MODULE ARE EXPERIMENTAL AND MIGHT
+ * BE SUBJECT OF BREAKING CHANGES WITHOUT NOTICE OR LIBRARY MAJOR VERSION CHANGE.
```
kmichalikk: nit: I believe this should be SUBJECT TO BREAKING CHANGES

same for types.ts

kkafar: Thanks. https://github.com/software-mansion/react-native-screens/pull/3223/commits/12faa9f9e36e83ba9b2c58a8ed4c76bb970c1f0a

PR #3228, opened 2025-09-18
fix(iOS, Stack): Deduplicate contentStyle on Screen and ScreenContentWrapper

t0maboro:
## Description

This PR addresses an issue with the support for LiquidGlass FormSheets on iOS, as detailed in this blog post https://nilcoalescing.com/blog/LiquidGlassSheetsWithNavigationStackAndForm/.

Previously, the implementation of the `contentStyle` option led to the background color overlapping. Specifically, with the low opacity styles were duplicated for `Screen` and `ScreenContentWrapper`, which caused visual issues in the content area.

To resolve this, I’ve separated the handling of `contentStyle` properties on iOS, which requires applying background color at the level of `Screen` component due to the safe area insets:

- The `backgroundColor` is now applied **only** at the RNSScreen level to achieve the desired LiquidGlass translucent effect without interfering with content rendering.
- The remaining `contentStyle` properties are applied under the `ScreenContentWrapper`, which is semantically more appropriate.

This results in a more accurate and visually correct rendering of translucent backgrounds on iOS FormSheets using LiquidGlass.

## Changes

- Separated `contentStyle` to props that should be applied on the `Screen` and `ScreenContentWrapper` level.

## Screenshots / GIFs

### Before

https://github.com/user-attachments/assets/7b65fa0f-e478-44d0-a6b6-eb8f60f8461a

### After

https://github.com/user-attachments/assets/c6ec1d34-1459-435b-ac16-886ef9d4a6c8

Added a new example to `TestFormSheet`

## Checklist

- [x] Ensured that CI passes

I'm leaving some follow-up issues and ideas that came during the implementation of this PR.

More important:
- https://github.com/software-mansion/react-native-screens-labs/issues/452 - it makes more sense to apply some other styles on the `Screen` level due to the safe area on iOS - like borders or filters - we should revisit it and define how the props should be split between `Screen` and `ScreenContentWrapper`
- https://github.com/software-mansion/react-native-screens-labs/issues/441 - setting fixed size based on the maximum detent would have 2 significant benefits: we'd have the same model for both Android and iOS and we'd resolve the problem of flickering content on resize without synchronous updates - it has a potential, tested along with this PR, but we're suffering from the lack of API on iOS 15 or lower - APIs for resolving detens is available from iOS 16, what is currently a blocker for us, because we cannot predict the target height for `large` detent, using system detents

Less important:
-  https://github.com/software-mansion/react-native-screens-labs/issues/457 - bug related to the safe area combined with maximum detent on android
- https://github.com/software-mansion/react-native-screens-labs/issues/458 - bug related to positioning the content with `flex-end` on android

### DIFF
--- a/src/components/ScreenStackItem.tsx
+++ b/src/components/ScreenStackItem.tsx
@@ -29,56 +29,6 @@ type Props = Omit<
   contentStyle?: StyleProp<ViewStyle>;
 };
 
-function getPositioningStyle(
-  allowedDetents: ScreenProps['sheetAllowedDetents'],
-  presentation?: StackPresentationTypes,
-) {
-  const isIOS = Platform.OS === 'ios';
-
-  if (presentation !== 'formSheet') {
-    return styles.container;
-  }
-
-  if (isIOS) {
-    if (allowedDetents === 'fitToContents') {
-      return styles.absolute;
-    } else {
-      return styles.container;
-    }
-  }
-
-  // Other platforms, tested reliably only on Android
-  if (allowedDetents === 'fitToContents') {
-    return {};
-  }
-
-  return styles.container;
-}
-
-type SplitStyleResult = {
-  screenStyles: {
-    backgroundColor?: ViewStyle['backgroundColor'];
-  };
-  contentWrapperStyles: StyleProp<ViewStyle>;
-};
-
-// TODO: figure out whether other styles, like borders, filters, etc.
-// shouldn't be applied on the Screen level on iOS due to the inset.
-function extractScreenStyles(style: StyleProp<ViewStyle>): SplitStyleResult {
-  const flatStyle = StyleSheet.flatten(style);
-
-  const { backgroundColor, ...contentWrapperStyles } = flatStyle as ViewStyle;
-
-  const screenStyles = {
-    backgroundColor,
-  };
-
-  return {
-    screenStyles,
-    contentWrapperStyles,
-  };
-}
-
 function ScreenStackItem(
   {
     children,
@@ -217,6 +167,56 @@ function ScreenStackItem(
 }
 
 export default React.forwardRef(ScreenStackItem);
+
+function getPositioningStyle(
+  allowedDetents: ScreenProps['sheetAllowedDetents'],
+  presentation?: StackPresentationTypes,
+) {
+  const isIOS = Platform.OS === 'ios';
+
+  if (presentation !== 'formSheet') {
+    return styles.container;
+  }
+
+  if (isIOS) {
+    if (allowedDetents === 'fitToContents') {
+      return styles.absolute;
+    } else {
+      return styles.container;
+    }
+  }
+
+  // Other platforms, tested reliably only on Android
+  if (allowedDetents === 'fitToContents') {
+    return {};
+  }
+
+  return styles.container;
+}
+
+type SplitStyleResult = {
+  screenStyles: {
+    backgroundColor?: ViewStyle['backgroundColor'];
+  };
+  contentWrapperStyles: StyleProp<ViewStyle>;
+};
+
+// TODO: figure out whether other styles, like borders, filters, etc.
+// shouldn't be applied on the Screen level on iOS due to the inset.
+function extractScreenStyles(style: StyleProp<ViewStyle>): SplitStyleResult {
+  const flatStyle = StyleSheet.flatten(style);
+
+  const { backgroundColor, ...contentWrapperStyles } = flatStyle as ViewStyle;
+
+  const screenStyles = {
+    backgroundColor,
+  };
+
+  return {
+    screenStyles,
+    contentWrapperStyles,
+  };
+}
 
 const styles = StyleSheet.create({
   container: {


apps/src/tests/TestFormSheet.tsx
```diff
@@ -151,6 +154,28 @@ function FormSheetWithScrollView() {
...
+              alignItems: 'center',
+              height: 40,
+              justifyContent: 'center',
+            }}>
+            <Text>Pressable</Text>
+          </View>
+        </PressableWithFeedback>
+        <Spacer space={50} />
+        <Button title="Go back" onPress={() => navigation.goBack()} />
+      </View>
```
kligarski:
```suggestion
    <View>
      <Spacer space={50} />
      <PressableWithFeedback>
        <View
          style={{
            alignItems: 'center',
            height: 40,
            justifyContent: 'center',
          }}>
          <Text>Pressable</Text>
        </View>
      </PressableWithFeedback>
      <Spacer space={50} />
      <Button title="Go back" onPress={() => navigation.goBack()} />
    </View>
```

src/components/DebugContainer.tsx
```diff
@@ -16,28 +17,38 @@ type ContainerProps = ViewProps & {
  * See https://github.com/software-mansion/react-native-screens/pull/1825
  * for detailed explanation.
  */
-let DebugContainer: React.ComponentType<ContainerProps> = props => {
-  return <ScreenContentWrapper {...props} />;
+let DebugContainer: React.FC<ContainerProps> = ({
+  contentStyle,
+  style,
+  ...rest
+}) => {
```
kligarski, line 24: Is it necessary to extract `style, contentStyle` from props here?

src/components/ScreenContentWrapper.tsx
```diff
@@ -1,9 +1,39 @@
...
+  ViewProps,
+  StyleSheet,
+} from 'react-native';
 import ScreenContentWrapperNativeComponent from '../fabric/ScreenContentWrapperNativeComponent';
 
-function ScreenContentWrapper(props: ViewProps) {
-  return <ScreenContentWrapperNativeComponent collapsable={false} {...props} />;
+type Props = ViewProps & {
+  contentStyle?: StyleProp<ViewStyle>;
+  children?: React.ReactNode;
```
kligarski: Isn't `children` already in `ViewProps`?

src/components/ScreenContentWrapper.tsx
```diff
@@ -1,9 +1,39 @@
...
+};
+
+function ScreenContentWrapper({
+  children,
+  contentStyle,
+  style,
```
kligarski: Do we need to extract `style` here as well?

src/components/ScreenContentWrapper.tsx
```diff
@@ -1,9 +1,39 @@
...
+      <View style={[styles.container, contentStyle]}>{children}</View>
+    </ScreenContentWrapperNativeComponent>
+  );
 }
 
+const styles = StyleSheet.create({
+  container: {
+    height: '100%',
+    width: '100%',
+  },
```
kligarski: I'm wondering how does it work with `fitToContents`, if `ScreenContentWrapper` serves as the source of content dimensions.

src/components/ScreenStackItem.tsx
```diff
@@ -25,6 +29,60 @@ type Props = Omit<
...
+};
+
+// TODO: figure out whether other styles, like borders, filters, etc.
+// shouldn't be applied on the Screen level on iOS due to the inset.
+export function extractScreenStyles(
+  style: StyleProp<ViewStyle>,
+): SplitStyleResult {
+  const flatStyle = Array.isArray(style)
+    ? Object.assign({}, ...style)
+    : style ?? {};
```
kligarski: Isn't `StyleSheet.flatten(style)` the same?

apps/src/tests/TestFormSheet.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3228/commits/57482acce2f9352f055be288d4e8141bc8625a29

src/components/DebugContainer.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3228/commits/1902ae584f077e5bb5567680ffcdfe5d701733ea

src/components/ScreenContentWrapper.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3228/commits/5194b5c9dddeefbd5d56abcb03ccdf50630adfc3

src/components/ScreenContentWrapper.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3228/commits/9b62dca2421edbe2c27e74e65e16ad4e402e594f

src/components/ScreenContentWrapper.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3228/commits/8a5dfa38f1d279ba3a3eb8114be46dd966f11133

src/components/ScreenStackItem.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3228/commits/64de4a95d9aac11c9ccb062343e61007debcc3ee

src/components/ScreenContentWrapper.tsx
t0maboro: explanation: tbh I don't need to have that additional intermediate view, as the ScreenContentWrapper seems to be semantically correct to put `contentStyle` directly to it

src/components/DebugContainer.tsx
t0maboro: EDIT: I prefer to keep it for readability, but that's not the strong preference

kligarski: I don't have a strong preference either. If you think it's more readable this way, we can leave it as is.

kligarski (approved):

src/components/ScreenStackItem.tsx
```diff
@@ -25,6 +29,58 @@ type Props = Omit<
...
+type SplitStyleResult = {
+  screenStyles: {
+    backgroundColor?: ViewStyle['backgroundColor'];
+  };
+  contentWrapperStyles: StyleProp<ViewStyle>;
+};
+
+// TODO: figure out whether other styles, like borders, filters, etc.
+// shouldn't be applied on the Screen level on iOS due to the inset.
+export function extractScreenStyles(
```
kkafar: Why do you export this function?

src/components/ScreenStackItem.tsx
```diff
@@ -25,6 +29,58 @@ type Props = Omit<
...
+
+  const screenStyles = {
+    backgroundColor,
+  };
+
+  return {
+    screenStyles,
+    contentWrapperStyles,
+  };
+}
```
kkafar: I'm all for it, but let's move definitions of these functions to the bottom of the file, so that they do not obscure the `ScreenStackItem`.

src/components/ScreenStackItem.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3228/commits/8926fbc506f62e2e2817fe8a9def9a5e5d5e6146

src/components/ScreenStackItem.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3228/commits/0565c31eec58c24cb59141f5641d14a5dd5b9907

kkafar (approved):
Thanks!

PR #3230, opened 2025-09-19
feat(iOS, Tabs): use `SafeAreaView` in Bottom Tabs

kligarski:
## Description

Integrates `SafeAreaView` component into `BottomTabsScreen`.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/167.

This PR adds partial support for Android but full support (with `insetType`) needs to be handled in separate PR, after this is merged to `main`. 

## Changes

- add `respectedSafeAreaEdges` prop to `BottomTabsScreen`
  - on iOS, all edges are disabled by default,
- add `contentStyle` prop to `BottomTabsScreen` to allow adding style to view behind `SafeAreaView`

Run `TestBottomTabs`, verify that default behavior did not change, use new props e.g. on `Tab2`.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/src/components/bottom-tabs/BottomTabsScreen.tsx
+++ b/src/components/bottom-tabs/BottomTabsScreen.tsx
@@ -4,7 +4,9 @@ import React from 'react';
 import { Freeze } from 'react-freeze';
 import {
   Image,
+  Platform,
   StyleSheet,
+  View,
   findNodeHandle,
   processColor,
   type ImageSourcePropType,
@@ -24,10 +26,13 @@ import type {
   BottomTabsScreenItemAppearance,
   BottomTabsScreenItemStateAppearance,
   BottomTabsScreenProps,
+  BottomTabsScreenSafeAreaEdges,
   EmptyObject,
   Icon,
 } from './BottomTabsScreen.types';
 import { bottomTabsDebugLog } from '../../private/logging';
+import { SafeAreaView } from 'react-native-screens/private';
+import { SafeAreaViewProps } from 'react-native-screens/private/types';
 
 /**
  * EXPERIMENTAL API, MIGHT CHANGE W/O ANY NOTICE
@@ -59,6 +64,8 @@ function BottomTabsScreen(props: BottomTabsScreenProps) {
     selectedIcon,
     standardAppearance,
     scrollEdgeAppearance,
+    respectedSafeAreaEdges,
+    contentStyle,
     ...rest
   } = props;
 
@@ -149,7 +156,11 @@ function BottomTabsScreen(props: BottomTabsScreenProps) {
       ref={componentNodeRef}
       {...rest}>
       <Freeze freeze={shouldFreeze} placeholder={rest.placeholder}>
-        {rest.children}
+        <View style={[contentStyle, styles.flex]}>
+          <SafeAreaView edges={getSafeAreaViewEdges(respectedSafeAreaEdges)}>
+            {rest.children}
+          </SafeAreaView>
+        </View>
       </Freeze>
     </BottomTabsScreenNativeComponent>
   );
@@ -241,6 +252,23 @@ function shouldFreezeScreen(
   return !nativeViewVisible;
 }
 
+function getSafeAreaViewEdges(
+  edges?: BottomTabsScreenSafeAreaEdges,
+): SafeAreaViewProps['edges'] {
+  let defaultEdges: SafeAreaViewProps['edges'];
+
+  switch (Platform.OS) {
+    case 'android':
+      defaultEdges = { bottom: true };
+      break;
+    default:
+      defaultEdges = {};
+      break;
+  }
+
+  return { ...defaultEdges, ...edges };
+}
+
 function parseIconToNativeProps(icon: Icon | undefined): {
   iconType?: IconType;
   iconImageSource?: ImageSourcePropType;
@@ -323,5 +351,8 @@ const styles = StyleSheet.create({
     flex: 1,
     width: '100%',
     height: '100%',
+  },
+  flex: {
+    flex: 1,
   },
 });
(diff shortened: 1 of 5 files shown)


src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -370,6 +377,23 @@ export interface BottomTabsScreenProps {
...
   orientation?: BottomTabsScreenOrientation;
+
+  /**
+   * @summary Specifies which edges should respect safe area.
+   *
+   * @default
+   * On Android, bottom edge is enabled.
+   * On iOS, all edges are disabled.
+   */
+  respectedSafeAreaEdges?: BottomTabsScreenSafeAreaEdges;
```
kligarski: I'm wondering if we want to keep this API for users or maybe use an array of edges, e.g. `['top', 'bottom']` (or maybe both?).

src/components/bottom-tabs/BottomTabsScreen.tsx
```diff
@@ -149,7 +156,11 @@ function BottomTabsScreen(props: BottomTabsScreenProps) {
       ref={componentNodeRef}
       {...rest}>
       <Freeze freeze={shouldFreeze} placeholder={rest.placeholder}>
-        {rest.children}
+        <View style={[contentStyle, styles.flex]}>
```
t0maboro, line 165: I'm wondering whether we shouldn't reverse the order, to have an ability to override `flex: 1`

src/components/bottom-tabs/BottomTabsScreen.tsx
```diff
@@ -149,7 +156,11 @@ function BottomTabsScreen(props: BottomTabsScreenProps) {
...
+          <SafeAreaView edges={getSafeAreaViewEdges(respectedSafeAreaEdges)}>
+            {rest.children}
```
t0maboro: nit: I'd extract children from props above

src/components/bottom-tabs/BottomTabsScreen.tsx
kligarski: @kkafar initially suggested to force `flex: 1` here; I don't have a strong opinion on this though.

src/components/bottom-tabs/BottomTabsScreen.tsx
kligarski: https://github.com/software-mansion/react-native-screens/pull/3230/commits/499e537b0036736123002063064dec2382ae3365

src/components/bottom-tabs/BottomTabsScreen.tsx
```diff
@@ -24,10 +26,13 @@ import type {
   BottomTabsScreenItemAppearance,
   BottomTabsScreenItemStateAppearance,
   BottomTabsScreenProps,
+  BottomTabsScreenSafeAreaEdges,
   EmptyObject,
   Icon,
 } from './BottomTabsScreen.types';
 import { bottomTabsDebugLog } from '../../private/logging';
+import { SafeAreaView } from 'react-native-screens/private';
+import { SafeAreaViewProps } from 'react-native-screens/private/types';
```
kkafar: Let's also reexport all types from `react-native-screens/private/types` directly from `react-native-screens/private` as it is done in regular package.

src/components/bottom-tabs/BottomTabsScreen.tsx
kkafar: Yep. It would be ideal to forbid user from using any layout-related props, but it would be tedious. We can go with this solution rn.

kkafar: I must say that I dislike this kind of API, where the style of the content is not determined inside particular screen content component, but needs to be set ahead of time when declaring tabs. This is not really good from UX PoV. 

I'm aware that it is hard to do it in any other way if we wanna have control over SAV here & not only expose insets to the user... I think we might wanna discuss this on nearest weekly.

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -18,6 +18,13 @@ export type LifecycleStateChangeEvent = Readonly<{
   newState: number;
 }>;
 
+export interface BottomTabsScreenSafeAreaEdges {
+  top?: boolean;
+  bottom?: boolean;
+  left?: boolean;
+  right?: boolean;
+}
```
kkafar: Is is specific to bottom tabs? Maybe it should end up in some shared types file. I think @kmichalikk has created one recently. Please communicate here (if my suggestion makes sense).

src/components/bottom-tabs/BottomTabsScreen.types.ts
kkafar: I think this is fine.

kkafar (commented):
Looks good in general. Left suggestion & filled discussions.

src/components/bottom-tabs/BottomTabsScreen.types.ts
kligarski: In the future, it won't be.

https://github.com/software-mansion/react-native-screens/pull/3230/commits/98236a340ce7c468ece16bcf4785cdaf24a45015

src/components/bottom-tabs/BottomTabsScreen.tsx
kligarski: https://github.com/software-mansion/react-native-screens/pull/3230/commits/b783f6f1d8821de1a43633f1adab900e3b328f30

kligarski:
Superseded by https://github.com/software-mansion/react-native-screens/pull/3256.

PR #3231, opened 2025-09-19
fix(iOS, Stack v4): use `SafeAreaView` in Stack v4 to fix content rendering under UINavigationBar on iOS 26

kligarski:
## Description

Integrates `SafeAreaView` component into `ScreenStackItem` to mitigate native UIKit bug with `edgesForExtendedLayout` on iOS 26 (more details in PR with previous workaround attempt: https://github.com/software-mansion/react-native-screens/pull/3111)

## Changes

- wrap content in `SafeAreaView` on iOS
- enable top edge inset if header `(!translucent && !hidden)`

Run `Test3111`.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/src/components/ScreenStackItem.tsx
+++ b/src/components/ScreenStackItem.tsx
@@ -15,6 +15,8 @@ import Screen from './Screen';
 import ScreenStack from './ScreenStack';
 import { RNSScreensRefContext } from '../contexts';
 import { FooterComponent } from './ScreenFooter';
+import { SafeAreaView } from 'react-native-screens/private';
+import { SafeAreaViewProps } from './safe-area/SafeAreaView.types';
 
 type Props = Omit<
   ScreenProps,
@@ -65,6 +67,8 @@ function ScreenStackItem(
     headerHiddenPreviousRef.current = headerConfig?.hidden;
   }, [headerConfig?.hidden, stackPresentation]);
 
+  const shouldUseSafeAreaView = Platform.OS === 'ios';
+
   const content = (
     <>
       <DebugContainer
@@ -79,7 +83,13 @@ function ScreenStackItem(
           contentStyle,
         ]}
         stackPresentation={stackPresentation ?? 'push'}>
-        {children}
+        {shouldUseSafeAreaView ? (
+          <SafeAreaView edges={getSafeAreaEdges(headerConfig)}>
+            {children}
+          </SafeAreaView>
+        ) : (
+          children
+        )}
       </DebugContainer>
       {/**
        * `HeaderConfig` needs to be the direct child of `Screen` without any intermediate `View`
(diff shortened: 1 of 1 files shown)


kmichalikk (approved):

src/components/ScreenStackItem.tsx
```diff
@@ -65,6 +67,8 @@ function ScreenStackItem(
     headerHiddenPreviousRef.current = headerConfig?.hidden;
   }, [headerConfig?.hidden, stackPresentation]);
 
+  const shouldUseSafeAreaView = Platform.OS === 'ios';
```
t0maboro: I'm wondering whether we should have an option to opt-out from the app's side

kligarski: That's a good question but also keep in mind that we're still using `edgesForExtendedLayout` in Stack v4. We're using `SafeAreaView` here because of native UIKit bug - disabling top edge in `edgesForExtendedLayout` does not offset the content enough but `safeAreaInsets` are correct.

If we had a prop to control `safeAreaView`, this would lead to inconsistencies - setting `top: false` wouldn't disable safe area in all cases (becase `edgesForExtendedLayout` would still work, e.g. with opaque header).

In the future, we might want to switch fully to `SafeAreaView` -> then configuration via prop, similar to `BottomTabs`, should be added.

t0maboro: got it, it's fine; can we create a ticket to revisit it?

kligarski: Sure.

https://github.com/software-mansion/react-native-screens-labs/issues/472

src/components/ScreenStackItem.tsx
```diff
@@ -15,6 +15,8 @@ import Screen from './Screen';
 import ScreenStack from './ScreenStack';
 import { RNSScreensRefContext } from '../contexts';
 import { FooterComponent } from './ScreenFooter';
+import { SafeAreaView } from 'react-native-screens/private';
```
kkafar: This should be relative import

src/components/ScreenStackItem.tsx
```diff
@@ -79,7 +83,13 @@ function ScreenStackItem(
           contentStyle,
         ]}
         stackPresentation={stackPresentation ?? 'push'}>
-        {children}
+        {shouldUseSafeAreaView ? (
```
kkafar, line 103: Do we wanna apply it only on iOS 26, or all versions?

kkafar (commented):
This looks good. Only a single remark.

src/components/ScreenStackItem.tsx
kligarski: https://github.com/software-mansion/react-native-screens/pull/3231/commits/1e3d4a85d3c7dc0d67e2f0288d31de92f729a291

src/components/ScreenStackItem.tsx
kligarski: https://github.com/software-mansion/react-native-screens/pull/3231/commits/1e3d4a85d3c7dc0d67e2f0288d31de92f729a291

Issue #2219, opened 2024-06-28
Pressable elements not working in native-stack on Android devices with new architecture


DrZoidberg09:
### Description

When using Android devices (iOS and Android emulator is fine) you cannot use onPress in the header bar. onPressIn and onPress out does work. This was tested mostly with Samsung devices.
This only happens with native-stack / Expo Router.

Please have a look here: 
https://github.com/expo/expo/issues/30032
https://github.com/react-navigation/react-navigation/issues/12039
https://github.com/facebook/react-native/issues/44643

### Steps to reproduce

1. Use react navigation native-stack with new arch enabled on RN 0.74
2. Go to a stack screen and try to press elements with onPress
3. onPress elements wont work. onPressIn and onPressOut does

### Snack or a link to a repository

https://github.com/DrZoidberg09/RN-Android-Touch-Issue/

For comparison a repo with bare RN, one with Expo and react navigation stack and one with Expo router / native-stack

### Screens version

3.32.0

### React Native version

0.74

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

None

### Architecture

Fabric (New Architecture)

### Build type

None

### Device

None

### Device model

Samsung S23, Samsung S8+

### Acknowledgements

Yes

kkafar:
🫨

kkafar:
Is that the thing only on real devices?

DrZoidberg09:
Yes. So far, only tested with Samsung devices, incl. S24, S8+ and a Samsung tablet (dont remember the name)

DrZoidberg09:
Plus everything pressable goes completely crazy in the screen (not only the header bar), if it is an actually build apk instead of an expo development build. However, not sure if this related.

AnzeBlaBla:
Also happening here, can confirm we've only seen it happen on-device, specifically the S24.

martinezguillaume:
I have the same issue here but with iOS (not tested on Android)
It is only present in production build
It seems that when `Pressable` is pressed, it is disappearing for 1 frame and reappear right after causing important flicker.
It is not happening all the time, for me, it is happening when I push a new screen and the previous screens has elements.

Screen <- Pressable working

EmptyScreen 
PushScreen <- Pressable is working

ScreenWithStuff
PushScreen <- Pressable is crazy

I hope it's understandable

Here is the behavior that I have
[...]

PR #3239, opened 2025-09-22
fix(Android): header config shadow node has wrong origin when `translucent == true`

maxencehenneron:
## Description

https://github.com/software-mansion/react-native-screens/pull/2466 fixes Pressable in screen header by shifting the HeaderConfigShadowNode origin.y by its frame height. (more info in the pull request).
However, setting isTranslucent to true on android displays the header over the content so we must not apply this correction.

This fixes Pressable for translucent headers on Android.

## Changes

Skip the frame corrections on Android when translucent is true. I'm not sure that's the cleanest way to fix this issue, but i'm not familiar enough with react-native-screens to find a better solution. Because this is an issue I have in my production app and this is critical, I'm pushing this fix for now.

https://snack.expo.dev/@dernise/header-navigation-broken

(Run this on a real device by scanning the QR code in expo go, do not use the emulator provided by appetize)


maxencehenneron:
related I believe: https://github.com/software-mansion/react-native-screens/issues/2219

kkafar (changes requested):
Hey! Thanks for identifying the problem. I don't think this is a proper solution, though. 

I'd go first with sending appropriate frame correction from native side, which should be (0, 0), I guess. Can we try that out?

maxencehenneron:
I added a new commit that updates the ShadowNode frame with a width/height to 0 on android side when the navbar is translucent. Is this correct though?
In theory we could consider the view to be without a size because it's just floating there.

Are you suggesting adding something like correctionY to the StateWrapper?  That would likely require an iOS refactoring too.

Also I added a test page. Always test on a real device as this issue does not not happen on simulators somehow.

kkafar:
I'll wait for CI to pass & merge it later or tomorrow morning.

fkreinh:
@kkafar I can confirm this issue is still present with TrueSheet bottom sheets on Android devices. I've created a minimal reproduction repository that demonstrates the bug:

Repository: https://github.com/fkreinh/android-stack-pressable

Video Demo: https://www.loom.com/share/f8b7177fd2d7436c9097ae003923601b

APK for testing: https://expo.dev/accounts/felipe-krein/projects/android-stack-pressable/builds/50ed961d-959e-4bbd-a5cd-39e7fc1578a5

Setup:
React Native: 0.81.4
Expo: ~54.0.11
@lodev09/react-native-true-sheet: ^2.0.6
react-native-screens: nightly (4.17.0-nightly-20250930-6525eabd8)
Behavior:
✅ Pressable outside the TrueSheet bottom sheet works correctly
❌ Pressable inside the TrueSheet bottom sheet does not respond to touches on Android physical devices
✅ Both work fine on iOS and Android emulators
The issue appears to be related to how native view containers (like TrueSheet's native bottom sheet implementation) handle touch events on Android devices with the new architecture.

kkafar:
@fkreinh we do not maintain `react-native-true-sheet` here. This is `react-native-screens`. You could checkout our formsheets exposed via `presentation: 'formSheet'` on screens. 

If you found a bug in `react-native-true-sheet` you should report it in appropriate repository. 
Cheers!

fkreinh:
@kkafar thanks for the response. The problem seems to be with react-native-screens, but it might also just be in the react-native-true-sheet library. Either way, I’ll try to reproduce it without that dependency.

Cheers!

Issue #3234, opened 2025-09-20
[android][newArchEnabled] Error navigating back when re-entering the app after opening another app


mgvictor7:
### Description

* I added this issue because it may be related to this issue (https://github.com/react-navigation/react-navigation/issues/12765) in react-native-navigation

If you open a picker to select a file using a library such as @react-native-documents/picker, if you exit the app and re-enter, when you navigate back, the styles of the screen header are lost, as well as the wallpaper.

I've tested with several versions of react-native: 0.79.6, 0.80.2, and 0.81.1.
I've tested with @react-navigation/native versions 7.x and 6.x.
It has been tested on a physical device with Android 16 and on emulators with Android 15 and Android 16.

IMPORTANT -> This error occurs if newArchEnabled=true

### Steps to reproduce

Clone this repository https://github.com/mgvictor7/AwesomeProject_79_6.git

1. Navigate to last screen and press to "Open document picker"
2. Go back until you exit the app
3. Access the app again and navigate back and forth.
4. When you go back, the screen’s content disappears and the header style reverts to default.

### Snack or a link to a repository

https://github.com/mgvictor7/AwesomeProject_79_6.git

### Screens version

4.16.0

### React Native version

0.79.6

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

None

### Device

Android emulator

### Acknowledgements

Yes

t0maboro:
Hey, @mgvictor7 

Thank you for reporting this issue and providing all the information.

When exiting and re-entering the app, a new instance of `MountingCoordinator` is created. As a result, `mountingOverrideDelegates_` is wiped, and our `notifyScreenRemoved` callback is not triggered. This is the RC of inconsistencies in the screen lifecycle.

I'm investigating possible paths for reinitializing our callback for `MountingCoordinator` from `ScreensModule`, and I'm working on the proper fix.

PR #3241, opened 2025-09-23
fix(Android, Stack): Add logic for reattaching mounting override delegate

t0maboro:
## Description

Currently, when the app is exited, a new `Surface`, `ShadowTree`, and `MountingCoordinator` are created due to the destruction and recreation of the associated Activity (triggered with `onDestroy`). As a result, the code that invokes `setMountingOverrideDelegate` is never executed again, which prevents `notifyScreenRemoved` from being called. This leads to an inconsistent state during screen transitions.

In this PR, I added a listener for the `onHostResume` lifecycle event. However, since the order of lifecycle events is not guaranteed, we cannot assume that `onHostDestroy` occurs before `onHostResume`. Therefore, we cannot mark which resume callbacks should trigger setupFabric.

> [!CAUTION]
> As a workaround, I'm storing a weak_ptr to the `MountingCoordinator` and, on every `onHostResume`, I check whether the coordinator has already registered our callback. We initiated an internal discussion about the long-term solution, which may require additional effort on the RN side.

Fixes: https://github.com/software-mansion/react-native-screens/issues/3234

## Changes

- extended ScreensModule with `LifecycleEventListener` to track `onHostResume` callback
- added set for tracking which coordinators has registered our callback

## Screenshots / GIFs

### Before

https://github.com/user-attachments/assets/4dbc6e52-7c8e-44a4-946e-05b645ad6cea

### After

https://github.com/user-attachments/assets/de132990-fb15-48dc-9bff-cc33f71cfbde

Tested with the application from https://github.com/software-mansion/react-native-screens/issues/3234

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreensModule.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreensModule.kt
@@ -153,9 +153,7 @@ class ScreensModule(
         }
     }
 
-    override fun onHostPause() {
-        // NO-OP
-    }
+    override fun onHostPause() = Unit
 
     override fun onHostDestroy() {
         if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {


kligarski (approved):

android/src/main/java/com/swmansion/rnscreens/ScreensModule.kt
```diff
@@ -28,6 +30,10 @@ class ScreensModule(
             } else {
                 Log.e("[RNScreens]", "Could not install JSI bindings.")
             }
+
+            proxy = NativeProxy()
```
kkafar: proxy creation should be behind check for new architecture

android/src/main/java/com/swmansion/rnscreens/ScreensModule.kt
```diff
@@ -28,6 +30,10 @@ class ScreensModule(
...
+
+            reactContext.addLifecycleEventListener(this)
```
kkafar: Same here I think. Same here I think. We don't use it on old arch.

android/src/main/java/com/swmansion/rnscreens/ScreensModule.kt
```diff
@@ -40,18 +46,22 @@ class ScreensModule(
...
         super.invalidate()
         proxy?.invalidateNative()
+        proxy = null
         nativeUninstall()
     }
 
     override fun initialize() {
         super.initialize()
+        setupFabric()
+    }
```
kkafar, line 62: I think we should move proxy instantiation & destruction to initialize / invalidate callbacks. These should be paired up.

The same with registering & unregistering from being context listener. 

> [!warning]
Please be careful & test whether we can safely unregister the lifecycle callback from `reactContext` as it might lead to us missing `onHostDestroy` last callback. If we want to do something there, it might be important to watch out here.

android/src/main/cpp/NativeProxy.h
```diff
@@ -10,9 +10,34 @@ namespace rnscreens {
...
+};
+
 class NativeProxy : public jni::HybridClass<NativeProxy> {
  public:
   std::shared_ptr<RNSScreenRemovalListener> screenRemovalListener_;
+  std::unordered_set<
+      std::weak_ptr<const facebook::react::MountingCoordinator>,
+      WeakMountingCoordinatorPtrHash,
+      WeakMountingCoordinatorPtrEqual>
+      coordinatorsWithMountingOverrides_;
```
kkafar: We're over complicating here a bit. Let's use plain vector here. The performance over couple of elements might even be superior ;D. Just check whether the vector already contains value before inserting & we're good. I'd abstract adding / removing observer to dedicated function, so that this logic is abstracted. 

The same for cleaning up expired coordinator weak pointers.

android/src/main/cpp/NativeProxy.cpp
```diff
@@ -28,17 +28,41 @@ void NativeProxy::nativeAddMutationsListener(
...
+        });
+  }
+
+  for (auto it = coordinatorsWithMountingOverrides_.begin();
+       it != coordinatorsWithMountingOverrides_.end();) {
+    if (it->expired()) {
+      it = coordinatorsWithMountingOverrides_.erase(it);
+    } else {
+      ++it;
+    }
```
kkafar: Iterating over a collection & removing items from it at the same time might be problematic. I don't know whether `std::unordered_set` supports this or not, but we should simply use vector as described in some other comment & `std::remove_if` to achieve the desired effect.

kkafar: We should abstract this to separate function & also call it `onHostDestroy`. As you mentioned in PR description the callback order between `onHostResume` & `onHostDestroy` is not guaranteed, therefore, if we want to play nice & keep this structure clean, we could expose additional native method to the Java side that allows to call this cleanup logic from `onHostDestroy`. This way, no matter the callback order we end up with nicely maintained structure.

android/src/main/java/com/swmansion/rnscreens/ScreensModule.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3241/commits/96dd4037cb0370dbb88eadcdba85b6f6db3911ab

android/src/main/java/com/swmansion/rnscreens/ScreensModule.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3241/commits/59b58c4532d184924779aa2ffbf4cb998009dd69

android/src/main/java/com/swmansion/rnscreens/ScreensModule.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3241/commits/d7d4583cf25b6ebb166f213bda05a2ed53e381dd

android/src/main/cpp/NativeProxy.h
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3241/commits/8e93858d00a99819a80644ddfd5139d39f1c6d0e

android/src/main/cpp/NativeProxy.cpp
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3241/commits/8e93858d00a99819a80644ddfd5139d39f1c6d0e

t0maboro: partially in: https://github.com/software-mansion/react-native-screens/commit/8e93858d00a99819a80644ddfd5139d39f1c6d0e + finished in: https://github.com/software-mansion/react-native-screens/pull/3241/commits/c7b50b37720aeba92b4ae198b24fbc5cd6e64a84

Guuz:
@kkafar / @t0maboro Thanks for the work on this!
When can we expect this to land in a release? this bug affects our app in production so i would like to update to roll out this fix :)

PR #3242, opened 2025-09-24
feat(iOS): Bring back fullScreenSwipeEnabled prop

kmichalikk:
## Description

Resolves [[iOS] Bring back fullScreenSwipeEnabled on iOS 26](https://github.com/software-mansion/react-native-screens-labs/issues/466).

This PR brings back `fullScreenSwipeEnabled` prop on iOS 26. The type of the prop is changed internally to OptionalBoolean with default value differing between versions.

Use `Test3093`

### DIFF
--- a/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
+++ b/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
@@ -54,8 +54,16 @@ Defaults to `false`. When `enableFreeze()` is run at the top of the application 
 
 ### `fullScreenSwipeEnabled` (iOS only)
 
-Boolean indicating whether the swipe gesture should work on whole screen. Swiping with this option results in the same transition animation as `simple_push` by default. It can be changed to other custom animations with `customAnimationOnSwipe` prop, but default iOS swipe animation is not achievable due to usage of custom recognizer. Defaults to `false`.
-IMPORTANT: Starting from iOS 26, full screen swipe is handled by native recognizer, and this prop is ignored.
+Boolean indicating whether the swipe gesture should work on whole screen. The behavior depends on iOS version.
+
+For iOS prior to 26, swiping with this option results in the same transition animation as `simple_push` by default.
+It can be changed to other custom animations with `customAnimationOnSwipe` prop, but default iOS swipe animation
+is not achievable due to usage of custom recognizer.
+
+For iOS 26 and up, native `interactiveContentPopGestureRecognizer` is used, and this prop controls whether it should 
+be enabled or not.
+
+When not set, it defaults to `false` on iOS < 26 and `true` for iOS >= 26.
 
 ### `fullScreenSwipeShadowEnabled` (iOS only)
 
--- a/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
+++ b/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
@@ -60,8 +60,8 @@ For iOS prior to 26, swiping with this option results in the same transition ani
 It can be changed to other custom animations with `customAnimationOnSwipe` prop, but default iOS swipe animation
 is not achievable due to usage of custom recognizer.
 
-For iOS 26 and up, native `interactiveContentPopGestureRecognizer` is used, and this prop controls whether it should 
-be enabled or not.
+For iOS 26 and up, native `interactiveContentPopGestureRecognizer` is used for all cases except for `customAnimationOnSwipe`,
+however the prop still can be used to controls whether it should be enabled or not.
 
 When not set, it defaults to `false` on iOS < 26 and `true` for iOS >= 26.
 
--- a/ios/RNSScreen.h
+++ b/ios/RNSScreen.h
@@ -72,7 +72,7 @@ namespace react = facebook::react;
      RNSSafeAreaProviding,
      RNSScrollEdgeEffectProviding>
 
-@property (nonatomic) BOOL fullScreenSwipeEnabled;
+@property (nonatomic) RNSOptionalBoolean fullScreenSwipeEnabled;
 @property (nonatomic) BOOL fullScreenSwipeShadowEnabled;
 @property (nonatomic) BOOL gestureEnabled;
 @property (nonatomic) BOOL hasStatusBarHiddenSet;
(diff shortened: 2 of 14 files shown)


t0maboro (approved):

kligarski (changes requested):
Please update docs in `src/types.tsx`.

ios/RNSScreen.h
```diff
@@ -164,6 +164,12 @@ namespace react = facebook::react;
 - (BOOL)isModal;
 - (BOOL)isPresentedAsNativeModal;
 
+/**
+ * Returns a boolean equivalen of fullScreenSwipeEnabled OptionalBoolean, resolves Undefined as `false` for iOS < 26,
```
kligarski:
```suggestion
 * Returns a boolean equivalent of fullScreenSwipeEnabled OptionalBoolean, resolves Undefined as `false` for iOS < 26,
```

ios/RNSScreen.mm
kligarski, line 1: Let's add initialization of `_fullScreenSwipeEnabled = RNSOptionalBooleanUndefined` in `initCommonProps`.

kligarski (approved):

ios/RNSScreen.mm
kmichalikk: 952fbb14f017252ffe2528c64fbb826590b8fb1e

ios/RNSScreen.h
kmichalikk: 952fbb14f017252ffe2528c64fbb826590b8fb1e

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
```diff
@@ -54,8 +54,16 @@ Defaults to `false`. When `enableFreeze()` is run at the top of the application
...
-Boolean indicating whether the swipe gesture should work on whole screen. Swiping with this option results in the same transition animation as `simple_push` by default. It can be changed to other custom animations with `customAnimationOnSwipe` prop, but default iOS swipe animation is not achievable due to usage of custom recognizer. Defaults to `false`.
-IMPORTANT: Starting from iOS 26, full screen swipe is handled by native recognizer, and this prop is ignored.
+Boolean indicating whether the swipe gesture should work on whole screen. The behavior depends on iOS version.
+
+For iOS prior to 26, swiping with this option results in the same transition animation as `simple_push` by default.
+It can be changed to other custom animations with `customAnimationOnSwipe` prop, but default iOS swipe animation
+is not achievable due to usage of custom recognizer.
+
+For iOS 26 and up, native `interactiveContentPopGestureRecognizer` is used, and this prop controls whether it should 
+be enabled or not.
```
kkafar: I get impression from this description that `customAnimationOnSwipe` does not work anymore on iOS 26 & I remember from testing that was not the case, was it? 

If it still works, we should rephrase the description so that it indicates that `customAnimationOnSwipe` is possible in both cases but it results in new native gesture recognizer not being used.

src/components/Screen.tsx
```diff
@@ -220,6 +221,13 @@ export const InnerScreen = React.forwardRef<View, ScreenProps>(
             sheetCornerRadius={sheetCornerRadius}
             sheetExpandsWhenScrolledToEdge={sheetExpandsWhenScrolledToEdge}
             sheetInitialDetent={resolvedSheetInitialDetentIndex}
+            fullScreenSwipeEnabled={
+              fullScreenSwipeEnabled === undefined
+                ? 'undefined'
+                : fullScreenSwipeEnabled
+                ? 'true'
+                : 'false'
```
kkafar: I think we should have dedicated function for this. Some kind of `convertBoolToOptionalBool` or something. If possible let's track down different code places where it could be applied.

ios/RNSScreen.h
```diff
@@ -164,6 +164,12 @@ namespace react = facebook::react;
 - (BOOL)isModal;
 - (BOOL)isPresentedAsNativeModal;
 
+/**
+ * Returns a boolean equivalent of fullScreenSwipeEnabled OptionalBoolean, resolves Undefined as `false` for iOS < 26,
+ * `true` otherwise.
+ */
+- (BOOL)fullScreenSwipeEnabledBoolean;
```
kkafar: I don't like this approach here. Please take a look how `obscureBackground`, which also uses RNSOptionalBoolean is implemented & follow that pattern.

kkafar (changes requested):
The PR looks good. I have some requests though. Please answer them.

Good job!

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
kmichalikk: [125c413](https://github.com/software-mansion/react-native-screens/pull/3242/commits/125c413c9541f96171b68d1a0ab537d002c36867)

src/components/Screen.tsx
kmichalikk: [6ea670c](https://github.com/software-mansion/react-native-screens/pull/3242/commits/6ea670cd52b0c443a67aeb3b61032317b6644a46)

kmichalikk: I searched `'undefined'` and found nothing, so maybe this was the only one

ios/RNSScreen.h
kmichalikk: This is somewhat different, because the value is used in RNSScreenStack, and for obscureBackground there is no usages outside of the context it's declared in. Could you point me in some direction here?

kkafar: Yep, I could see `fullScreenSwipeEnabled` property being `readonly`, overwrite the getter to return appropriate `BOOL` value & we do not need any additional method here.

kkafar (commented):
Responded to your question

ios/RNSScreen.h
kkafar: The backing field can be of `RNSOptionalBoolean` type obviously.

kmichalikk: [62e1a05](https://github.com/software-mansion/react-native-screens/pull/3242/commits/62e1a05aa6d86f53e9aadc7318581b18ee7f9189)
I'm still not sure we understand each other, please take a look

kkafar (commented):
Upon request, revisiting my approach here

terrysahaidak:
@kkafar @kmichalikk is it possible to use this prop with a vertical scroll view? somehow if i have scroll view from the gesture handler in the screen it won't swipe on iOS 26. only on the very edge.

PR #3244, opened 2025-09-24
chore(iOS): Add test case for scrollEdgeEffects

kmichalikk:
Resolves https://github.com/software-mansion/react-native-screens-labs/issues/459

## Description

This PR adds a test for `scrollEdgeEffects`. On launch, it presents a set of scenarios that the ScrollView may appear in: Stack, BottomTabs, Stack inside Stack, Stack inside BottomTabs, and BottomTabs inside Stack. In order to keep the code simple, changing the scenario requires reload.

https://github.com/user-attachments/assets/dc1c09ca-830e-42aa-96b0-3475846c2532


apps/src/tests/Test3212.tsx
```diff
@@ -0,0 +1,215 @@
...
+  bottom: ScrollEdgeEffect,
+  top: ScrollEdgeEffect,
+  left: ScrollEdgeEffect,
+  right: ScrollEdgeEffect,
+}
+
+function ScrollViewTemplate() {
+  const emoji = ['😎', '🍏', '👀', '🤖', '👾', '👨‍💻'];
+
+  return (<>
```
t0maboro: using fragment is not necessary here, applies to all below

apps/src/tests/Test3212.tsx
```diff
@@ -0,0 +1,215 @@
...
+  if (navigating) {
+    navigation = useNavigation();
+  }
+
+  return <>
+    <Text style={{ margin: 8, fontSize: 24 }}>scrollEdgeEffects:</Text>
+    <SettingsPicker
+      label="bottom"
+      value={scrollEdgeEffects.bottom}
+      items={['automatic', 'hard', 'soft', 'hidden']}
```
t0maboro: it can be moved to a single array and passed to all pickers, or settings picker can have this prop with default value to avoid duplication

apps/src/tests/Test3212.tsx
```diff
@@ -0,0 +1,215 @@
...
+function StackTemplate() {
+  const Stack = createNativeStackNavigator();
+  const [scrollEdgeEffects, setScrollEdgeEffects] = useState<ScrollEdgeEffects>({
+    bottom: 'automatic',
+    top: 'automatic',
+    left: 'automatic',
+    right: 'automatic',
+  });
+
+  const ConfigComponent = () => <Config scrollEdgeEffects={scrollEdgeEffects} setScrollEdgeEffects={setScrollEdgeEffects} navigating />
```
t0maboro: it will recreate ConfigComponent on every rerender, I'd consider moving it to the upper scope

apps/src/tests/Test3212.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+function BottomTabsTemplate() {
+  const [scrollEdgeEffects, setScrollEdgeEffects] = useState<ScrollEdgeEffects>({
+    bottom: 'automatic',
+    top: 'automatic',
+    left: 'automatic',
+    right: 'automatic',
+  });
+
+  const ConfigComponent = () => <ScrollView contentInsetAdjustmentBehavior='automatic'>
```
kligarski: `contentInsetAdjustmentBehavior` shouldn't be necessary in `BottomTabs` (we override it by default).

apps/src/tests/Test3212.tsx
kmichalikk: 43b49128447b7df1884e1281fd9d1c4d35f6f11a

apps/src/tests/Test3212.tsx
kmichalikk: 43b49128447b7df1884e1281fd9d1c4d35f6f11a; moved to a constant

apps/src/tests/Test3212.tsx
kmichalikk: 43b49128447b7df1884e1281fd9d1c4d35f6f11a; this needs to be in this scope because it requires the setters, changed to useMemo

apps/src/tests/Test3212.tsx
kmichalikk: 43b49128447b7df1884e1281fd9d1c4d35f6f11a

apps/src/tests/Test3212.tsx
t0maboro: nice

t0maboro (approved):

apps/src/tests/Test3212.tsx
```diff
@@ -0,0 +1,235 @@
...
+}
+
+const SCROLL_EDGE_EFFECT_OPTIONS: ScrollEdgeEffect[] = ['automatic', 'hard', 'soft', 'hidden'];
+
+function Config(props: ConfigProps) {
+  const { scrollEdgeEffects, setScrollEdgeEffects, navigating = false } = props;
+  let navigation: NavigationProp<{ Test: undefined }>;
+  if (navigating) {
+    navigation = useNavigation();
+  }
```
kkafar: It is forbidden to call hooks conditionally. https://react.dev/reference/rules/rules-of-hooks#only-call-hooks-at-the-top-level

This must be refactored.

apps/src/tests/Test3212.tsx
```diff
@@ -0,0 +1,235 @@
...
+    bottom: 'automatic',
+    top: 'automatic',
+    left: 'automatic',
+    right: 'automatic',
+  });
+
+  const ConfigComponent = useMemo(
+    () => () => <Config scrollEdgeEffects={scrollEdgeEffects} setScrollEdgeEffects={setScrollEdgeEffects} navigating />,
+    [scrollEdgeEffects],
+  );
```
kkafar: Using a context just would make much more sense here.

apps/src/tests/Test3212.tsx
```diff
@@ -0,0 +1,235 @@
...
+      return <BottomTabsTemplate />;
+    case 'StackInStack':
+      return <StackInStackTemplate />;
+    case 'StackInBottomTabs':
+      return <StackInBottomTabsTemplate />;
+    case 'BottomTabsInStack':
+      return <BottomTabsInStackTemplate />;
+  }
+
+}
```
kkafar:
```suggestion
}

```

apps/src/tests/Test3212.tsx
```diff
@@ -0,0 +1,235 @@
...
+  const [scrollEdgeEffects, setScrollEdgeEffects] = useState<ScrollEdgeEffects>({
+    bottom: 'automatic',
+    top: 'automatic',
+    left: 'automatic',
+    right: 'automatic',
+  });
+
+  const ConfigComponent = useMemo(
+    () => () => <Config scrollEdgeEffects={scrollEdgeEffects} setScrollEdgeEffects={setScrollEdgeEffects} navigating />,
+    [scrollEdgeEffects],
```
kkafar: Btw. this should be a `useCallback` not `useMemo`. These hooks have [different purposes](https://reliasoftware.com/blog/react-usememo-vs-usecallback-hooks).

apps/src/tests/Test3212.tsx
kmichalikk: b9ba16afe01c396d1fd03a0e7ff2929ff9586857

apps/src/tests/Test3212.tsx
kmichalikk: b9ba16afe01c396d1fd03a0e7ff2929ff9586857

apps/src/tests/Test3212.tsx
kmichalikk: b9ba16afe01c396d1fd03a0e7ff2929ff9586857, there are 2 instances where it's still needed after refactoring, switched to `useCallback`

kmichalikk:
Test should be reachable from the `<Example >`.

Currently there is no way to see whether the outer config affects the inner config (in context of `Stack in Stack`, `BottomTabs in Stack`, `Stack in BottomTabs`), but if you attach the debugger inside `RNSScrollViewApplicator`, you can see that sometimes it's called with the outer config, but with ScrollView == nil. It is expected at this point. On Paper + `Stack in BottomTabs`, the outer config should be able to "leak" and be visible. There are PRs that change the behavior of searching for ScrollView, and the test will be updates as needed.

apps/src/tests/Test3212/index.tsx
```diff
@@ -5,17 +5,40 @@ import { BottomTabsScenario } from "./BottomTabsScenario";
...
 import { BottomTabsInStackScenario } from "./BottomTabsInStackScenario";
+import { enableFreeze } from "react-native-screens";
 
 export default function App() {
   const [environment, setEnvironment] = useState('select');
 
+  // With freeze enabled, the contents of some views are not present,
+  // and the ScrollView can't be found. To make changes to props handling
+  // more visible, we disable freezing.
+  enableFreeze(false);
```
kligarski: Are we sure we want to do this? If someone (or test runner) uses example app, will this be restored after they leave the test example? But I guess that when E2E was working, we'd usually do an app refresh for each test case so it'll be okay.

Some context: https://github.com/software-mansion/react-native-screens/pull/2779

I'm not sure here what the right answer is.

kmichalikk: I need this for the test to be able to show the changes I'm after. I quickly checked and `enableFreeze()` call in App.tsx is run first, then the call inside the test, doesn't matter if it's run separately or from <Example />.

Also, E2E should run every test on separate environment IMO.

kkafar: I don't think this is the right approach. If you really need this then you should just pass `freezeContents={false}` to the tab screens, not disable the freeze globally.

kkafar (changes requested):
The test looks good overall ✅ 

I've got only a single change request regarding `enableFreeze` usage.

apps/src/tests/Test3212/index.tsx
kmichalikk: [d723c42](https://github.com/software-mansion/react-native-screens/pull/3244/commits/d723c42eed265542fa4c50a8c2953932a4705f0a)

apps/src/tests/Test3212/BottomTabsScenario.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { BottomTabsContainer } from "../../shared/gamma/containers/bottom-tabs/BottomTabsContainer"
+import { Config } from "./Config";
+import { ScrollViewTemplate } from "./ScrollViewTemplate";
+import { ScrollView } from "react-native";
+
+export function BottomTabsScenario() {
+  const [config, setConfig] = useState<ScrollEdgeEffects>({ ...SCROLL_EDGE_EFFECT_DEFAULTS });
+
+  // Add ScrollView for automatic insets which are missing in BottomTabsScreen
+  const ConfigComponent = useCallback(() => <ScrollView><Config title='BottomTabs / scrollEdgeEffects:' /></ScrollView>, []);
```
kkafar, line 13: This is fine here, I do not require you to change it, but please notice, that this inline component does not depend on any state nor is directly tied to the `BotomTabsScenario` component ==> it could be just a standalone helper component defined  as standalone function in the bottom of the file making things a little bit cleaner.

kkafar (approved):
Looks ok now. Thanks.

Issue #3188, opened 2025-09-02
[Android] Form sheet does not keep the keyboard focus within


timoisalive:
### Description

When opening a form sheet with Android (via keyboard), the keyboard focus should stay within the form sheet. Currently the focus can be moved to the accessible elements on the screen below as well as the bottom tabs, which should not be possible.

On iOS the keyboard focus stays within the modal and also, when using `presentation: 'modal'`, the focus behaves as expected on both platforms.

https://github.com/user-attachments/assets/e99130ad-b78b-47e2-9afc-eb5825e6c639

### Steps to reproduce

Close the linked repo and run.

To test in emulator/simulator, use your keyboard to control the app (move with arrows, press buttons with spacebar).

On iOS, make sure you have Full Keyboard Access switched on in Accessibility Settings.

### Snack or a link to a repository

https://github.com/timoisalive/form-sheet-keyboard-focus

### Screens version

~4.15.4

### React Native version

0.81.1

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

timoisalive:
Related discussion: https://github.com/ArturKalach/react-native-external-keyboard/discussions/92

PR #3245, opened 2025-09-24
fix(Android, Stack v4): fix keyboard navigation focus for form sheet

kligarski:
## Description

Fixes keyboard navigation focus handling for form sheets on Android.

Previously, it was possible to focus elements in the view displayed below the form sheet.

Closes https://github.com/software-mansion/react-native-screens/issues/3188.

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/481435f7-5304-4b74-9d2c-65693a33ffed" /> | <video src="https://github.com/user-attachments/assets/68faf6cb-1ccb-4bdb-aab2-3aa9d20a936a" /> |

### Context

In `ScreenStack`, there is already logic to handle accessibility focus via `importantForAccessibility` prop. I checked that the prop is changed correctly - there were no regressions here.

I stumbled upon an issue in `react-native` repo [where one of the users explained](https://github.com/facebook/react-native/issues/30850#issuecomment-1184756886) that Android handles screen readers focus and keyboard navigation focus differently. In order to block keyboard navigation focus, you need to set [`focusable` and `descendantFocusability`](https://github.com/facebook/react-native/issues/30850#issuecomment-1184958396) properties - when implemented in `ScreenStack`, this started working correctly (after clicking `arrow-down`, focus did not leave the formsheet).

Another problem that appeared is that the button used to open the form sheet remained focused after form sheet had been already opened. I attempted to use `requestFocus()` on the form sheet's screen but if there is a text input in the form sheet, it started working as if we set `autoFocus` on the input field.

https://github.com/user-attachments/assets/40b0d027-f982-494c-9b0a-ed9ae6f52670

This also impacted regular touch navigation. I decided to use `clearFocus` on previous screen instead - this works as expected: first button in the screen is focused after opening the form sheet.

One thing I noticed is that when you go back from a screen, nothing is focused (I checked with layout inspector) - this is not a regression but we should have a look at it in the future. Native app (settings) on API 36 keeps focus when screen is popped.

### Changing focusability on Android

Starting from API 26, `focusable` can be set to [`NOT_FOCUSABLE/FOCUSABLE/FOCUSABLE_AUTO`](https://developer.android.com/reference/android/view/View#setFocusable(int)). Prior to API 26, this was a boolean prop - that's why there is some extra code to handle both cases.

## Changes

- add methods to set focusability for `Screen` for API <= 25 and API >= 26
- change focusability in sync with `importantForAccessibility`
- set `focusable` to `false/NOT_FOCUSABLE` for `DimmingView` (this caused missing focus after popping a `push` screen that was pushed over form sheet)
- clear focus from previous top screen in ScreenStack's `onUpdate`

Run `TestFormSheet` and use the keyboard to navigate.

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -3,6 +3,7 @@ package com.swmansion.rnscreens
 import android.annotation.SuppressLint
 import android.content.pm.ActivityInfo
 import android.graphics.Paint
+import android.os.Build
 import android.os.Parcelable
 import android.util.SparseArray
 import android.view.MotionEvent
@@ -11,6 +12,7 @@ import android.view.ViewGroup
 import android.view.WindowManager
 import android.webkit.WebView
 import android.widget.ImageView
+import androidx.annotation.RequiresApi
 import androidx.coordinatorlayout.widget.CoordinatorLayout
 import androidx.core.view.children
 import androidx.fragment.app.Fragment
@@ -343,6 +345,35 @@ class Screen(
     fun changeAccessibilityMode(mode: Int) {
         this.importantForAccessibility = mode
         this.headerConfig?.toolbar?.importantForAccessibility = mode
+    }
+
+    // Accepts one of 3 focusable flags and one of 3 descendantFocusability flags.
+    // https://developer.android.com/reference/android/view/View#setFocusable(int)
+    // https://developer.android.com/reference/android/view/ViewGroup#setDescendantFocusability(int)
+    @RequiresApi(Build.VERSION_CODES.O)
+    fun changeFocusability(
+        focusable: Int,
+        descendantFocusability: Int,
+    ) {
+        this.focusable = focusable
+        this.headerConfig?.toolbar?.focusable = focusable
+
+        this.descendantFocusability = descendantFocusability
+        this.headerConfig?.toolbar?.descendantFocusability = descendantFocusability
+    }
+
+    // Accepts boolean for focusability and one of 3 descendantFocusability flags.
+    // This method is provided for compatibility reasons only (SDK_API <= 25).
+    // https://developer.android.com/reference/android/view/ViewGroup#setDescendantFocusability(int)
+    fun changeFocusabilityCompat(
+        focusable: Boolean,
+        descendantFocusability: Int,
+    ) {
+        this.setFocusable(focusable)
+        this.headerConfig?.toolbar?.setFocusable(focusable)
+
+        this.descendantFocusability = descendantFocusability
+        this.headerConfig?.toolbar?.descendantFocusability = descendantFocusability
     }
 
     var statusBarStyle: String? = null
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
@@ -4,6 +4,7 @@ import android.content.Context
 import android.graphics.Canvas
 import android.os.Build
 import android.view.View
+import androidx.core.view.ViewCompat
 import com.facebook.react.bridge.ReactContext
 import com.facebook.react.uimanager.UIManagerHelper
 import com.swmansion.rnscreens.Screen.StackAnimation
@@ -262,6 +263,7 @@ class ScreenStack(
                 transaction.add(id, newTop.fragment)
             }
 
+            val previousTopScreen: Screen? = topScreenWrapper?.screen
             topScreenWrapper = newTop as? ScreenStackFragmentWrapper
             stack.clear()
             stack.addAll(screenWrappers.asSequence().map { it as ScreenStackFragmentWrapper })
@@ -279,6 +281,13 @@ class ScreenStack(
                     .toList()
 
             turnOffA11yUnderTransparentScreen(visibleBottom)
+
+            // When form sheet is pushed to the top of the stack, we need to make sure that keyboard
+            // focus doesn't stay on the previous screen. We don't use requestFocus on newTop's
+            // screen because if there is an input field in the form sheet, it works as if autoFocus
+            // was enabled on the input field instead of focusing e.g. button above the input.
+            previousTopScreen?.clearFocus()
+
             transaction.commitNowAllowingStateLoss()
         }
     }
@@ -296,6 +305,20 @@ class ScreenStack(
                             IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS,
                         )
 
+                        // Keyboard navigation focus is separate from screen reader focus, that's
+                        // why we need to use focusable and descendantFocusability.
+                        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                            fragmentWrapper.screen.changeFocusability(
+                                NOT_FOCUSABLE,
+                                FOCUS_BLOCK_DESCENDANTS
+                            )
+                        } else {
+                            fragmentWrapper.screen.changeFocusabilityCompat(
+                                false,
+                                FOCUS_BLOCK_DESCENDANTS
+                            )
+                        }
+
                         // don't change a11y below non-transparent screens
                         if (fragmentWrapper == visibleBottom) {
                             break
@@ -306,6 +329,11 @@ class ScreenStack(
         }
 
         topScreen?.changeAccessibilityMode(IMPORTANT_FOR_ACCESSIBILITY_AUTO)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            topScreen?.changeFocusability(FOCUSABLE_AUTO, FOCUS_BEFORE_DESCENDANTS)
+        } else {
+            topScreen?.changeFocusabilityCompat(true, FOCUS_BEFORE_DESCENDANTS)
+        }
     }
 
     override fun notifyContainerUpdate() {
--- a/android/src/main/java/com/swmansion/rnscreens/bottomsheet/DimmingView.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/bottomsheet/DimmingView.kt
@@ -3,6 +3,7 @@ package com.swmansion.rnscreens.bottomsheet
 import android.annotation.SuppressLint
 import android.content.Context
 import android.graphics.Color
+import android.os.Build
 import android.view.MotionEvent
 import android.view.ViewGroup
 import com.facebook.react.uimanager.ReactCompoundViewGroup
@@ -31,6 +32,7 @@ internal class DimmingView(
 
     init {
         pointerEventsProxy.pointerEventsImpl = DimmingViewPointerEventsImpl(this)
+        setFocusable(false)
     }
 
     internal val blockGestures


android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -345,6 +347,35 @@ class Screen(
...
+    // https://developer.android.com/reference/android/view/ViewGroup#setDescendantFocusability(int)
+    fun changeFocusabilityCompat(
+        focusable: Boolean,
+        descendantFocusability: Int,
+    ) {
+        this.setFocusable(focusable)
+        this.headerConfig?.toolbar?.setFocusable(focusable)
+
+        this.descendantFocusability = descendantFocusability
+        this.headerConfig?.toolbar?.descendantFocusability = descendantFocusability
```
t0maboro, line 574: nit: it repeats with the prev method, so you may consider extracting the logic for descendant focusability

t0maboro (approved):
👍

android/src/main/java/com/swmansion/rnscreens/Screen.kt
t0maboro: nevermind, it's the same for a11y, so I'm ok with that
```
    fun changeAccessibilityMode(mode: Int) {
        this.importantForAccessibility = mode
        this.headerConfig?.toolbar?.importantForAccessibility = mode
    }
```

kkafar:
@kkafar

kkafar (commented):
Okay, so important thing here is that the form sheet on Android is not a modal view byt it's own. 

We use something called "standard bottom sheet" & it is a set of views attached directly to the hierarchy in place where the `Screen` component is mounted. It does not behave like modal: it does not create separate `DecorView` & does not mount above all contents (that's something we'd like to support in next major). 

Therefore I'm not all for emulating all modal features here. 

What we could do is to expose a prop Android only prop to control this. OR alternatively block the focusability ONLY when dimming view is present. The presence of dimming view indicates to the user that he should focus on the sheets contents utterly, not on the background. 

So yeah, what do you think @kligarski? Does that sound sensible?

timoisalive:
Are there any news on this matter...? We'd still like to use the form sheets, but this focus issue is a blocker for us. Thanks! 🙏

kligarski:
> Are there any news on this matter...? We'd still like to use the form sheets, but this focus issue is a blocker for us. Thanks! 🙏

Hi, sorry for the delay. I've updated the PR and hopefully we'll proceed soon.

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
```diff
@@ -141,9 +140,6 @@ class SheetDelegate(
                 }
             }
 
-            // Note: There's no good reason that Screen should be direct target for focus, we're rather
-            // prefer its children to gain it.
-            screen.descendantFocusability = ViewGroup.FOCUS_AFTER_DESCENDANTS
```
kligarski, line 146: This should be handled now with a11y so I removed it here.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -677,6 +708,11 @@ class Screen(
         if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED && isStable) {
             onSheetYTranslationChanged()
         }
+
+        // When detent changes, dimming might also change. If so, we need to update a11y.
+        if (isNativeStackScreen) {
+            (container as ScreenStack).updateA11yForVisibleScreens()
+        }
```
kligarski: This is ugly & I'm open for suggestions here how to handle this case.

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
```diff
@@ -161,6 +164,7 @@ class ScreenStack(
                 .dropWhile { it.isTranslucent() }
                 .firstOrNull()
                 ?.takeUnless { it === newTop }
+        currentVisibleBottom = visibleBottom
```
kligarski, line 167: I need visibleBottom in `updateA11yForVisibleScreens` and now it might be also called outside of `onUpdate`. When I tried to use only the property, I faced errors related to concurrent mutability so I kept local variable in the method. If you have any suggestions here, let me know.

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
```diff
@@ -294,34 +298,71 @@ class ScreenStack(
...
+                    screenWrappers.slice(0 until screenWrappers.size - 1).asReversed()
+                // go from the top of the stack excluding the top screen
+                for (fragmentWrapper in screenFragmentsBeneathTop) {
+                    val accessibilityMode =
+                        if (shouldDisableFocusability) IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS else IMPORTANT_FOR_ACCESSIBILITY_AUTO
+                    fragmentWrapper.screen.changeAccessibilityMode(accessibilityMode)
+
+                    // Keyboard navigation focus is separate from screen reader focus, that's
+                    // why we need to use focusable and descendantFocusability.
+                    changeScreenFocusability(fragmentWrapper.screen, !shouldDisableFocusability)
```
kligarski, line 322: As the dimming view might appear and disappear with detent changes, we need to set and reset the focusability.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
kmichalikk: If we stick to that, we need a more descriptive comment

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
```diff
@@ -294,34 +298,71 @@ class ScreenStack(
...
+            if (it.screen.usesFormSheetPresentation()) {
+                val screenStackFragment = it.fragment.asScreenStackFragment()
+                screenStackFragment.sheetDelegate?.let { delegate ->
+                    return delegate.lastStableDetentIndex > it.screen.sheetLargestUndimmedDetentIndex
+                }
+            } else {
+                return it.isTranslucent()
+            }
+        }
+        return false
```
kmichalikk:
```suggestion
            if (!it.screen.usesFormSheetPresentation()) {
                return it.isTranslucent()
            }
            
            val screenStackFragment = it.fragment.asScreenStackFragment()
            screenStackFragment.sheetDelegate?.let { delegate ->
                return delegate.lastStableDetentIndex > it.screen.sheetLargestUndimmedDetentIndex
            }
        }
        return false
```

nit: I would like to reduce nesting whenever possible, but do as you see fit.

kmichalikk (changes requested):
I don't see how to improve the code in any major way. Left some nitpicks. But I also found some bugs.

1. You can still move focus under dimming view. When you go to second sheet & back, the focus is correct.

https://github.com/user-attachments/assets/1cb38d5e-e7e0-4b42-923b-75f5226b5cc1

2. Here, you can move everywhere but to the "go back" button.

https://github.com/user-attachments/assets/0ebb33e5-9a8a-4281-af5b-c62f3495bef8

kligarski:
> I don't see how to improve the code in any major way. Left some nitpicks. But I also found some bugs.
> 
>     1. You can still move focus under dimming view. When you go to second sheet & back, the focus is correct.
> 
> 
> Nagranie.z.ekranu.2026-02-16.o.10.17.15.mov
> 
>     2. Here, you can move everywhere but to the "go back" button.
> 
> 
> Nagranie.z.ekranu.2026-02-16.o.10.31.45.mov

Thank you for checking this. I have no idea how I missed this as 1. is the basic test case I used when adding the change, maybe I missed this in one of the refactors.

It turns out that `sheetDelegate` might not be initialized yet when I want to check the whether dimming is used, that's why I added a fallback to initial detent value:

https://github.com/software-mansion/react-native-screens/pull/3245/commits/13206a16fbbab476b259a0c733bfdac5a7484568

https://github.com/user-attachments/assets/3e9ea175-8a35-432f-a841-dad39fd49a47

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
kligarski: While fixing the bug you mentioned in review description, I managed to refactor this function:

https://github.com/software-mansion/react-native-screens/pull/3245/commits/13206a16fbbab476b259a0c733bfdac5a7484568

android/src/main/java/com/swmansion/rnscreens/Screen.kt
kligarski: Let me know if this is what you meant:

https://github.com/software-mansion/react-native-screens/pull/3245/commits/9fdf686c9950f83e3bfb997f9b913fafd3da36f4

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
```diff
@@ -294,34 +298,68 @@ class ScreenStack(
...
+
+                    // Keyboard navigation focus is separate from screen reader focus, that's
+                    // why we need to use focusable and descendantFocusability.
+                    changeScreenFocusability(fragmentWrapper.screen, !shouldDisableFocusability)
+
+                    // don't change a11y below non-transparent screens
+                    if (fragmentWrapper == currentVisibleBottom) {
+                        break
                     }
                 }
```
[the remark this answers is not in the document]
kligarski: We have an early return if we reach `currentVisibleBottom`.

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
```diff
@@ -294,34 +298,68 @@ class ScreenStack(
...
             }
         }
 
         topScreen?.changeAccessibilityMode(IMPORTANT_FOR_ACCESSIBILITY_AUTO)
+        topScreen?.let { changeScreenFocusability(it, true) }
+    }
+
+    private fun shouldDisableFocusabilityForVisibleScreens(): Boolean {
```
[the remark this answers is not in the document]
kligarski: https://github.com/software-mansion/react-native-screens/pull/3245/commits/d4254c9767b97f7210bed2c2d8d9c308501afe54

I changed it to `shouldDisableFocusabilityBeneathTopScreen`, in theory it's still not precise enough as we don't disable all screens beneath top but only screens that are visible but new name is probably better than the previous one. Open for suggestions as well.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -677,6 +708,13 @@ class Screen(
         if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED && isStable) {
             onSheetYTranslationChanged()
         }
+
+        // a11y props such as `importantForAccessibility`, `focusable` and `descendantFocusability`
+        // are handled for form sheet depending on dimming view (if dimming view is present, the
+        // background screen is not focusable). When detent changes, dimming view might also change.
+        // That's why we're requesting update of a11y from ScreenStack (which handles those props).
+        check(isNativeStackScreen) { "[RNScreens] Expected native stack screen." }
```
t0maboro: leaving it as a note that this will crash for some reason if we push another screen from FormSheet, go back, dismiss FormSheet and open TextInput - kinda complex setup, but we should at least create a ticket to investigate why FormSheet detents may change after dismissing it

t0maboro (approved):
I have one request for a follow-up ticket that shouldn't block this PR. Overall, it looks good

android/src/main/java/com/swmansion/rnscreens/Screen.kt
kligarski: This should fix the issue for now:
https://github.com/software-mansion/react-native-screens/pull/3245/commits/22a46dd8e865ba7f6ed3031238e12cd54515cb8f

I also created a ticket:
https://github.com/software-mansion/react-native-screens-labs/issues/992

PR #3246, opened 2025-09-25
chore: Move BottomTabs props configuration outside of BottomTabsContainer

kmichalikk:
Resolves https://github.com/software-mansion/react-native-screens-labs/issues/398.

## Description

This PR removes hardcoded prop values from BottomTabsContainer and moves them to main BottomTabs so the test is left as is.

Compare BottomTabsTest before & after, should have no difference.

### DIFF
--- a/apps/src/shared/gamma/containers/bottom-tabs/BottomTabsContainer.tsx
+++ b/apps/src/shared/gamma/containers/bottom-tabs/BottomTabsContainer.tsx
@@ -2,11 +2,11 @@ import React from 'react';
 import type { NativeSyntheticEvent } from 'react-native';
 import {
   BottomTabs,
+  BottomTabsProps,
   BottomTabsScreen,
   BottomTabsScreenProps,
   NativeFocusChangeEvent,
 } from 'react-native-screens';
-import { Colors } from '../../../styling/Colors';
 import ConfigWrapperContext from './ConfigWrapperContext';
 
 export interface TabConfiguration {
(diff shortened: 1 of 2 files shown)


apps/src/shared/gamma/containers/bottom-tabs/BottomTabsContainer.tsx
```diff
@@ -2,34 +2,34 @@ import React from 'react';
...
 export interface TabConfiguration {
   tabScreenProps: BottomTabsScreenProps;
   component: React.ComponentType;
 }
 
-export interface BottomTabsContainerProps {
-  tabConfigs: TabConfiguration[];
-}
+// Currently assumes controlled bottom tabs
+export type BottomTabsContainerProps = Omit<BottomTabsProps, 'experimentalControlNavigationStateInJS' | 'onNativeFocusChange'> & { tabConfigs: TabConfiguration[]; }
```
kkafar: I don't think removing these types is necessary. It might be always nice to pass them during testing. This unnecessarily restricts this possibility.

apps/src/shared/gamma/containers/bottom-tabs/BottomTabsContainer.tsx
```diff
@@ -2,34 +2,34 @@ import React from 'react';
...
 
 export function BottomTabsContainer(props: BottomTabsContainerProps) {
-  // Currently assumes controlled bottom tabs
   console.info('BottomTabsContainer render');
 
+  const { tabConfigs, ...restProps } = props;
```
kkafar, line 25: It's ok, don't change it. Just taking opportunity to make sure you're aware here what this syntax does: `...restProps` - AFAIK you create a copy of whole object here.

kkafar (changes requested):
Looks reasonable. Thank you!

I have one single remark with change requests. Other than that we should be good. Good job!

apps/src/shared/gamma/containers/bottom-tabs/BottomTabsContainer.tsx
kmichalikk: [da96db2](https://github.com/software-mansion/react-native-screens/pull/3246/commits/da96db28bc8dcd10eb323f2ebbf0e8f7c8f709dd)

kkafar (approved):
We're looking good here now. Thanks!

Issue #3181, opened 2025-08-29
[Android] Bottom sheets do not resize when keyboard is open when using `fitToContents`


Armster15:
### Description

Whereas on iOS, bottom sheets with `sheetAllowedDetents: "fitToContents"` resize/adjust automatically when the keyboard is open, this does not occur on Android.

https://github.com/user-attachments/assets/565973ec-6119-4f57-8897-fe4d58c34d3c

Expected behavior: bottom sheet on Android automatically adjusts for keyboard like it does on iOS.

I am using bottom sheets as follows. Note I am using Expo Router for routing:
```ts
import { Stack } from "expo-router";

export default function TabTwoScreen() {
  return (
    <Stack>
      <Stack.Screen name="index" options={{ title: "Explore" }} />
      <Stack.Screen
        name="nested"
        options={{
          title: "Nested",

          presentation: "formSheet",
          gestureDirection: "vertical",
          animation: "slide_from_bottom",
          sheetGrabberVisible: true,
          sheetInitialDetentIndex: 0,

          sheetAllowedDetents: "fitToContents",
        }}
      />
    </Stack>
  );
}
```

### Steps to reproduce

1. `create-expo-app`
2. Create a screen as a bottom sheet using React Navigation with the following options + add a TextInput to the screen
     ```ts
     options={{ 
       presentation: 'formSheet',
       gestureDirection: 'vertical',
       animation: 'slide_from_bottom',
       sheetGrabberVisible: true,
       sheetInitialDetentIndex: 0,
       sheetAllowedDetents: "fitToContents",
     }}
     ````
3. Try opening text input with keyboard on Android. See bug.

My repro is this but using the default `create-expo-router` template.

### Snack or a link to a repository

https://github.com/Armster15/rn-screens-bottom-sheet-bug

### Screens version

4.11.1

### React Native version

0.79.6

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Architecture

Fabric (New Architecture)

### Build type

None

### Device

None

### Acknowledgements

Yes

t0maboro:
Hey, thank you for reporting this issue and providing all the necessary information with repro. I was able to reproduce it on my end and I'm investigating possible approaches to handle this.

PR #3248, opened 2025-09-25
WIP - fix(Android, Stack): Moving formsheet above keyboard

t0maboro:
## Description

WIP - still a lot of work needs to be done:
1. slide-in transition on opening form sheet isn't working atm
2. the current implementation with translationY on `onProgress` doesn't look well - on progress is triggered just a few times and causes ugly form sheet jumps occasionally
3. https://developer.android.com/guide/topics/manifest/activity-element#wsoft - needs to be tested
4. going through all supported API levels

Fixes: https://github.com/software-mansion/react-native-screens/issues/3181

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
@@ -252,7 +252,11 @@ class ScreenStackFragment :
                     override fun onProgress(
                         insets: WindowInsetsCompat,
                         runningAnimations: MutableList<WindowInsetsAnimationCompat>,
-                    ): WindowInsetsCompat = insets
+                    ): WindowInsetsCompat {
+                        val keyboardHeight = insets.getInsets(WindowInsetsCompat.Type.ime()).bottom
+                        screen.translationY = -sheetDelegate.calculateSheetOffsetY(keyboardHeight).toFloat()
+                        return insets
+                    }
                 },
             )
         }
--- a/android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
@@ -237,6 +237,32 @@ class SheetDelegate(
         }
     }
 
+    internal fun calculateSheetOffsetY(keyboardHeight: Int): Int {
+        val containerHeight = tryResolveContainerHeight()
+        check(containerHeight != null) {
+            "[RNScreens] Failed to find window height during bottom sheet behaviour configuration"
+        }
+
+        val isFitToContents = screen.isSheetFitToContents()
+
+        if (isFitToContents) {
+            val contentHeight = screen.contentWrapper?.height ?: 0
+            val offsetFromTop = containerHeight - contentHeight
+            return minOf(offsetFromTop, keyboardHeight)
+        }
+
+        val detents = screen.sheetDetents
+        if (detents.isEmpty()) {
+            throw IllegalStateException("[RNScreens] Cannot determine sheet detent - detents list is empty")
+        }
+
+        val detentValue = detents[detents.size - 1].coerceIn(0.0, 1.0)
+        val sheetHeight = (detentValue * containerHeight).toInt()
+        val offsetFromTop = containerHeight - sheetHeight
+
+        return minOf(offsetFromTop, keyboardHeight)
+    }
+
     // This is listener function, not the view's.
     override fun onApplyWindowInsets(
         v: View,


t0maboro:
Moving back to draft for a while, because I noticed some issues with autofocus and textinput below API 30

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
```diff
@@ -241,18 +246,16 @@ class ScreenStackFragment :
             )
             coordinatorLayout.layout(0, 0, container.width, container.height)
 
-            // Replace InsetsAnimationCallback created by BottomSheetBehavior with empty
-            // implementation so it does not interfere with our custom formSheet entering animation
-            // More details: https://github.com/software-mansion/react-native-screens/pull/2909
```
kligarski, line 247: I'd consider leaving some information that we're overriding inset-handling logic provided by default by `BottomSheetBehavior` with our custom one to handle interaction between keyboard insets and enter/exit animations.

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
```diff
@@ -237,6 +250,32 @@ class SheetDelegate(
...
 
+    internal fun calculateSheetOffsetY(keyboardHeight: Int): Int {
+        val containerHeight = tryResolveContainerHeight()
+        check(containerHeight != null) {
+            "[RNScreens] Failed to find window height during bottom sheet behaviour configuration"
+        }
+
+        val isFitToContents = screen.isSheetFitToContents()
+
+        if (isFitToContents) {
```
kligarski:
```suggestion
        if (screen.isSheetFitToContents()) {
```

nitpick: I think that `screen.isSheetFitToContents` is descriptive and we don't need another variable here but leaving this up to you

kligarski (commented):
I had already started the review before you moved the PR to draft so I'll finish it.

This looks great! I have some remarks/questions:

1. Just to be sure, this is not a regression, right? (not respecting navigation bar insets)

https://github.com/user-attachments/assets/c3d5e75c-5c4f-47e8-bff5-1cb370d76e39

2. Do you think we can do something about the gap between keyboard and formsheet when they're both hiding at the same time? Should we create a ticket to investigate it in the future?

https://github.com/user-attachments/assets/70e08e5e-304e-4b18-935b-8d6804c67438

3. We should probably revisit https://github.com/software-mansion/react-native-screens/pull/2925 to fix this, right?

https://github.com/user-attachments/assets/ad29ee40-15e7-46b1-8f9b-b345001f2abb

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3248/commits/2450ec83c769eda4fd0e042ff89f71fe4261744f

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3248/commits/cba51767d1bd8ae038ff443d1758b4583563ad97 I brought back the comment in this commit - let me know if it's okay

t0maboro:
@kligarski answering your questions:
1. Imo not a regression, because as a top edge I was considering top edge of the component, the same way as we'd put `detent=1.0` previously. However, we should definitely consider whether this is still valid, as now we're avoiding the keyboard. cc @kkafar if you have any thoughts on that

Screenshot 2025-09-29 at 16 08 55

2. I created a ticket on labs repo https://github.com/software-mansion/react-native-screens-labs/issues/484, but the problem here is that I'm dependent on `onProgress` callback, which will cause a slight delay.

3. For status bar, I'd prefer to continue the work in https://github.com/software-mansion/react-native-screens/pull/2925 - I can take it over and adapt to this PR (until anyone has any objections of at.1 whether the status bar is a regression or not)

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
kligarski: I'd just remove the part about `empty implementation` - it used to be empty but now you added some logic inside. Other that that, it's good.

kligarski:
> @kligarski answering your questions:
> 
> Imo not a regression, because as a top edge I was considering top edge of the component, the same way as we'd put detent=1.0 previously. However, we should definitely consider whether this is still valid, as now we're avoiding the keyboard. cc @kkafar if you have any thoughts on that

I'm not sure if we're talking about the same thing: I meant the text input being placed behind navigation bar (3-button navigation at the bottom) in first `fitToContents` example. I checked the main and confirmed that this is not a regression so we're good.

image

t0maboro:
> > @kligarski answering your questions:
> > Imo not a regression, because as a top edge I was considering top edge of the component, the same way as we'd put detent=1.0 previously. However, we should definitely consider whether this is still valid, as now we're avoiding the keyboard. cc @kkafar if you have any thoughts on that
> 
> I'm not sure if we're talking about the same thing: I meant the text input being placed behind navigation bar (3-button navigation at the bottom) in first `fitToContents` example. I checked the main and confirmed that this is not a regression so we're good.
> 
> image

Sorry, I read all together before answering and idk why I started thinking that 1 and 3 are connected. Good that we have an alignment on this one, but we should rethink whether this is expected now

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3248/commits/78f5636f11c14650f24a395bb320195ccabc9462

kligarski (approved):

badalsaibo:
would appreciate this getting merged as it will help with migration from current buggy bottom sheet libs we have to formSheet completely for the new architecture

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
```diff
@@ -64,6 +66,10 @@ class ScreenStackFragment :
     var searchView: CustomSearchView? = null
     var onSearchViewCreate: ((searchView: CustomSearchView) -> Unit)? = null
 
+    private var fadeAnimationRunning = false
```
kkafar:
```suggestion
    private var isSheetAnimatinInProgress = false
```

Let's make it specific.

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
```diff
@@ -287,64 +307,124 @@ class ScreenStackFragment :
...
-            val startValueCallback = { initialStartValue: Number? -> screen.height.toFloat() }
-            val evaluator = ExternalBoundaryValuesEvaluator(startValueCallback, { 0f })
-            val slideAnimator =
-                ValueAnimator.ofObject(evaluator, screen.height.toFloat(), 0f).apply {
-                    addUpdateListener { anim ->
-                        val animatedValue = anim.animatedValue as? Float
-                        animatedValue?.let { screen.translationY = it }
-                    }
-                }
+        val alphaAnimator = createAlphaAnimator(0f, dimmingDelegate.maxAlpha)
```
kkafar: This refactor hides the details of subjects of the animation. 

```suggestion
        val alphaAnimator = createDimmingViewAlphaAnimator(0f, dimmingDelegate.maxAlpha)
```

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
```diff
@@ -287,64 +307,124 @@ class ScreenStackFragment :
...
+        val slideAnimator = createSlideInAnimator()
```
kkafar:
```suggestion
        val slideAnimator = createSheetSlideInAnimator()
```

same here

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
```diff
@@ -287,64 +307,124 @@ class ScreenStackFragment :
...
+    }
+
+    private fun createSlideOutAnimator(): ValueAnimator {
+        val endValue = (coordinatorLayout.bottom - screen.top - screen.translationY)
+        return ValueAnimator.ofFloat(0f, endValue).apply {
+            addUpdateListener {
+                updateScreenTranslation(it.animatedValue as Float)
+            }
+        }
+    }
```
kkafar: I'd be awesome to move all this code to SheetDelegate, so that screen stack fragment is less cluttered. 

I know that previous implementation was also here, but it was my bad. 
If you want, we can do this in separate PR.

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
```diff
@@ -287,64 +307,124 @@ class ScreenStackFragment :
             return null
         }
 
+        return if (enter) createEnterAnimator() else createExitAnimator()
+    }
+
+    private fun createEnterAnimator(): Animator {
```
kkafar:
```suggestion
    private fun createSheetEnterAnimator(): Animator {
```

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
```diff
@@ -64,6 +66,10 @@ class ScreenStackFragment :
     var searchView: CustomSearchView? = null
     var onSearchViewCreate: ((searchView: CustomSearchView) -> Unit)? = null
 
+    private var fadeAnimationRunning = false
+
+    private var lastKeyboardBottomOffset: Int? = null
```
kkafar: does it need to be nullable or will `0` suffice. From initial code reading it seems that 0 would be ok

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
```diff
@@ -287,64 +307,124 @@ class ScreenStackFragment :
...
                 }
-            animatorSet.play(alphaAnimator).with(slideAnimator)
+            }
         }
+
+    private fun createSlideInAnimator(): ValueAnimator {
+        val startValueCallback = { _: Number? -> screen.height.toFloat() }
+        val evaluator = ExternalBoundaryValuesEvaluator(startValueCallback, { 0f })
+
+        return ValueAnimator.ofObject(evaluator, screen.height.toFloat(), 0f).apply {
```
kkafar: Reading this for the first time & I don't get why on slide in animator the start value is `height` and target value is `0`, at least passed here. I also see that there is an evaluator. 

I'll update the comment if I get this

kkafar: I wrote this code intitially, but I do not remember that. 😄

kkafar: Oh, is it because the sheet has already target position & we translate it back to the animation start position? That might be it.

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
```diff
@@ -237,6 +250,30 @@ class SheetDelegate(
...
+        if (screen.isSheetFitToContents()) {
+            val contentHeight = screen.contentWrapper?.height ?: 0
+            val offsetFromTop = containerHeight - contentHeight
+            return minOf(offsetFromTop, keyboardHeight)
+        }
+
+        val detents = screen.sheetDetents
+        if (detents.isEmpty()) {
+            throw IllegalStateException("[RNScreens] Cannot determine sheet detent - detents list is empty")
+        }
```
kkafar, line 296: We should have separate data structure for detents, making sure of this invariant. Let's create ticket for this.

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
```diff
@@ -237,6 +250,30 @@ class SheetDelegate(
...
+
+        val detentValue = detents[detents.size - 1].coerceIn(0.0, 1.0)
```
kkafar, line 298: I don't get this part. Why do you take value of the largest detent here?

kkafar: This method is used only when there is keyboard present, is that right? If so, let's name it appropriately, cause right now it's not obvious at all. 

e.g. `computeSheetOffsetYWithIMEPresent` or something.

kkafar: also this place deserves its own comment, because it really defines the behaviour of the sheet -> that it expands to max detent when the keyboard shows, right?

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
t0maboro: tbh we should consider extracting it to a separate class, moving it to another existing class will introduce the same mess somewhere else

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
t0maboro: yes, the sheet expands to its max detent; the purpose of this code is to determine whether we're able to show the full sheet or if we need to cover it  partially with the keyboard, definitely deserves some description, giving the information about the final offset from bottom, to which we should animate

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
t0maboro: The initial sheet position is when the sheet bottom edge is equal to device's bottom edge, we're translating Y with a positive value to hide the sheet under the bottom edge of the device

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3248/commits/a52b93aa3b9bd21d71cf5e92a052f6dde3ec823e

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3248/commits/c935b5d14b5cc36d8ca803a33d88509a3ede910a

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3248/commits/541be9dadc257bff60ec165b44794793ae9f3ba7 + https://github.com/software-mansion/react-native-screens/pull/3248/commits/f1e63f22de3a82fddab4ef9de027108bf72cb4e4

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3248/commits/6a5476d5fe152fb6a206831d1193b16e9990824a + https://github.com/software-mansion/react-native-screens/pull/3248/commits/58d2d1fd4930fd6d96db09d7c2bba15b1776912b

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3248/commits/08f272e5a0f56b8a01a99f6941336e70388a614d

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
t0maboro: https://github.com/software-mansion/react-native-screens-labs/issues/536

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3248/commits/bcfbf5513b977198809388ef060ab6fdaa3f3fe7

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
kkafar: Discussed in PM. Conclusion: let's move it to SheetDelegate, similarly as it is done with `UINavigationControllerDelegate` in UIKit or alternatively create some kind of `SheetAnimationDelegate` with animation methods separated. (I prefer option 1, but will accept both).

t0maboro: https://github.com/software-mansion/react-native-screens-labs/issues/537 - I'm adding a task to my queue and I'll work on that in a separate PR if that's okay for you

apps/src/tests/Test3248.tsx
```diff
@@ -0,0 +1,169 @@
...
+  FormSheetWithAutoFocusAndDetent: undefined;
+};
+
+type MainProps = {
+  navigation: NativeStackNavigationProp<StackParamList, 'Main'>;
+};
+
+const Stack = createNativeStackNavigator();
+
+const Main = ({ navigation }: MainProps) => {
```
kkafar: One small thing here. I encourage everywhere to use `function` instead of arrow functions. The argument here is mostly about being consistent with the most of the codebase.

kkafar: That obviously applies to whole file.

kkafar:
What about this behaviour? 

It goes up, even if there is no IME present? Can we detect such case? 

https://github.com/user-attachments/assets/f6e435c9-b93d-4e02-8e37-461a369f06c3

kkafar:
Also this is messed up a bit. Is this particular behaviour handled in the safe area view PR? 
The text input lands under the status bar

https://github.com/user-attachments/assets/9344b3ff-257e-4910-b0a2-e621322ab447

t0maboro:
> What about this behaviour?
> 
> It goes up, even if there is no IME present? Can we detect such case?

https://github.com/software-mansion/react-native-screens/pull/3248/commits/718758692544a582f77f7222035781cbdc455d6c + https://github.com/software-mansion/react-native-screens/pull/3248/commits/d354569726179ce791fb49bc486be53675e04a0e (cleanup)

> Also this is messed up a bit. Is this particular behaviour handled in the safe area view PR?
The text input lands under the status bar

As discussed internally, let's proceed with that in the PR with FormSheet - SAV fixes

apps/src/tests/Test3248.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3248/commits/743d8d72ce673a6a52fe4c9ed2cad0291be78c8b

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3378

PR #3250, opened 2025-09-25
fix(android): correctly end view transitions 

### LINKED ISSUE #3249: View recycling breaks due to fragment view transitions
### Description

When a native view manager reuses native views, such as when turning on View Recycling, there can be crashes caused by the react-native mounting layer trying to add a recycled view which still has a parent attached.

The red view you can see in this video is a recycled view, so each screen will try to use the same native instance of the view:

https://github.com/user-attachments/assets/616e9db4-d51f-4250-8937-1d09d3289f64

```
addViewAt: failed to insert view [1560] into parent [1564] at index 1
The specified child already has a parent. You must call removeView) on the child's parent first.
```

To fix this I believe we need to:

- fix RNS `endRemovalTransition` in `Screen.kt` as its currently broken and not correctly ending view transitions
- make changes in react-native core for handling "in-transition" views

I have a PR that tries to address that issue here:

- https://github.com/software-mansion/react-native-screens/pull/3250

### Steps to reproduce

The reproduction can be seen here:

- https://github.com/software-mansion/react-native-screens/compare/main...hannojg:react-native-screens:reproduction/view-recycling-crashing-on-screen-transitions

To test it out simply clone / checkout & build from this branch:

- https://github.com/hannojg/react-native-screens/tree/reproduction/view-recycling-crashing-on-screen-transitions

Note: In this reproduction i didn't turn on view reycling, but just always return the same native view instance. This is reproducing how view recycling behaves. 

I have an additional reproduction with view recycling in the PoC PR for fixing this issue:
- https://github.com/software-mansion/react-native-screens/pull/3250

https://github.com/hannojg/react-native-screens/tree/reproduction/view-recycling-crashing-on-screen-transitions

### Screens version

latest main

### React Native version

0.81.1

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

Real device

### Device model

_No response_

### Acknowledgements

Yes

hannojg:
## Description

Fixes https://github.com/software-mansion/react-native-screens/issues/3249

> [!NOTE]
> This PR still contains the reproduction code necessary + it contains changes related to RN core. This is meant so we can have a conversation about this issue and see how to apply the fixes best.

There are two problems which this PR tries to face to fix usage with recycling views:

### 1. `Screen#endRemovalTransition` is broken:

https://github.com/software-mansion/react-native-screens/blob/4982d4c44bcc788448337ce55e3329c57b87e37a/android/src/main/java/com/swmansion/rnscreens/Screen.kt#L472-L477

Here you can see that we try to iterate over the children of the screen to end the view transition. The problem is that the react-native mounting layer already has all the children removed from the parent. The flow of events is:

- The user requests to remove a screen, we call through `NativeProxy` `startRemovalTransition`. All children will be marked as "in-transition"
- In JS react our screen component gets removed (i believe thats how its working, I might be wrong) and the react-native mounting layer will call `removeView` for all the children of our screen
- The views will remain visible as we marked them as "in transition". HOWEVER, the children will already get removed from the parents `mChildren` list. Only the children will keep their `mParent` field set
- We call `endRemovalTransition` once the transition is done, trying to iterate over the children, but they are already unset. You can verify this by logging the children count when this method gets called
- 🐛 Bug: Our views, that may get recycled, are never called with `endViewTransition(child)` and their `mParent` field never gets reset. Thus the react-native mounting layer will always crash when trying to reuse this view.

### 2. Interrupting a screen transition can cause a wrong order of events

We rely on `Screen#endRemovalTransition` to be called before `ScreenStack#endRemovalTransition`. `ScreenStack` is in some way the parent of `Screen`. So the parent of the screen first calls `endViewTransition`. This is problematic because:

1. When a parent gets called with `endViewTransition` it will loop over all children and call `dispatchDetachedFromWindow()`:
   - [ViewGroup#endViewTransiton](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/view/ViewGroup.java;l=7240;bpv=0;bpt=1)
   - [ViewGroup#dispatchDetachedFromWindow](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/view/ViewGroup.java;l=3957-3990)
2. The problem with that is that the children will be simply detached, and cleared from the `mDisappearingChildren` 
3. Later, when we call `Screen#endRemovalTransition` and loop over the views calling `endViewTransition` will have no effect, since the views are no longer marked as disappearing:
   - https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/view/ViewGroup.java;l=7229-7233
4. 🐛 Bug: Due to this the children's `mParent` field never gets reset. Thus the react-native mounting layer will always crash when trying to reuse this view.

### 1. Fixing `Screen#endRemovalTransition` is broken:

I simply added a list of parent-child pairs that we add to in `startRemovalTransition`. Then when `endRemovalTransition` gets called we loop over that list. 

### 2. Fixing interrupting a screen transition can cause a wrong order of events:

Two fixes are necessary for this:

1. In `ScreenStack` we call `Screen#endViewTransition` first, before calling into `super.endViewTransition`. I think this is safe since `Screen#endRemovalTransition` has a flag checking if we are currently in a removal transition or not
2. We need to call `endViewTransition` from the most inner children up to the top parent so we avoid the problem explained above, where calling `endViewTransition` on a parent will remove all children from `mDisappearingChildren` list

### Additional changes in react-native core

As you can see I patched RN + override the default `ReactViewManager`. I think we also need to address this in react-native somehow. There are two things that need to be addressed I believe:

- When recycling views (`onDropViewInstance`) we could check if a view still has a parent. If so we could assume its in transition (or maybe have it marked as such earlier), and only add the view to the recyling pool on detach (ie. when the transition has ended)
- This crash seems not just reproducible with recycling views. It was just one way for me to get it to reproduce, but it seems that there might be other code paths where the same native view is trying to be reused (at least thats what the crash report's telemetry suggest me) [also in the app where this is happening we don't have view recycling enabled]. So additionally we might want to add something to `addView` in the `ReactViewManager` / `BaseViewManager` that will check if a view is "in-transition" / has a parent, and only attach the view once the transition is done (ie. on detach)

➤ I think that we can still apply the changes in RNS without needing to wait for react-native core though.

## Screenshots / GIFs

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/9ba4ee77-faf4-4e48-b213-78a43ea172da" /> | <video src="https://github.com/user-attachments/assets/fce2d17f-e5f7-4937-92ab-1641578b6c11" /> | 

You can simply run this PR and go to the `TestRecylingViews` tab in the example app to play with it. 

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [x] Ensured that CI passes


kkafar (commented):
Hey, this looks really sensible.

> The views will remain visible as we marked them as "in transition". HOWEVER, the children will already get removed from the parents mChildren list. Only the children will keep their mParent field set
> We call endRemovalTransition once the transition is done, trying to iterate over the children, but they are already unset. You can verify this by logging the children count when this method gets called

You're exactly right here. A while ago I [fixed](https://github.com/facebook/react-native/pull/47634) similar problem in core, yet here I've failed to notice this 😅 

Therefore I'm all for landing part of this patch that applies to screens incorrectly ending view transition. 

As for the view recycling part, I'll admit that we have not considered up to this moment much, as, correct me if I'm wrong, it is disabled by default on Android. 
I would like to avoid (if possible) code for special handling of view recycling in screens. I'd just prefer to fix it in core.
So thanks to my [PR](https://github.com/facebook/react-native/pull/47634) to core, the `ReactViewGroup` tracks which children have been removed during transition and which were not -> this is information we could inspect from child view to make sure, we've been removed from parent while being in transition, or we can add similar bit of logic to `ReactViewGroupManager` (as also you suggest) - check whether the view has parent or not. This is the best method I've found back then. 

> [...] but it seems that there might be other code paths where the same native view is trying to be reused 

Yeah, that can happen. In screens we've gone a while ago with  ["recycle method"](https://github.com/software-mansion/react-native-screens/blob/12704e59c2c5c3b37eb1f74e6b783b695297c759/android/src/main/java/com/swmansion/rnscreens/ext/ViewExt.kt#L11-L29) that solved this for screen instances. Might be reasonable to also upstream it to core. But I'd be more in favour of identifying particular code paths & culprits that lead to those kind of crashes instead of "recycling" every view in core, as it would only encourage incorrect code from libraries side.

hannojg:
> As for the view recycling part, I'll admit that we have not considered up to this moment much, as, correct me if I'm wrong, it is disabled by default on Android.

yes thats correct, also not sure if this is ever gonna land in stable 😅  It was just one way for me to reproduce this crash.

> ReactViewGroup tracks which children have been removed during transition and which were not -> this is information we could inspect from child view to make sure, we've been removed from parent while being in transition

Ah yeah, thats nice, i will experiment with that!

---

Okay cool, so then i'd remove everything else which i added for reproduction from this PR and we can move forward with the main change? 😊

kkafar:
Yes, we can. Let's get this PR up to date & I think we can proceed.

hannojg:
Cool, there is just one thing that might needs to be changed. I had this change in production and saw today the following reports in sentry:

```
java.util.ConcurrentModificationException: null
    at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:1111)
    at java.util.ArrayList$ListItr.previous(ArrayList.java:1138)
    at kotlin.collections.ReversedList$listIterator$1.next(ReversedViews.kt:48)
    at com.discord.view.ScreenOverride.endRemovalTransition(ScreenViewManagerOverride.kt:96)
    at com.swmansion.rnscreens.ScreenStackFragment.onViewAnimationEnd(ScreenStackFragment.kt:141)
    at com.swmansion.rnscreens.stack.views.ScreensCoordinatorLayout$animationListener$1.onAnimationEnd(ScreensCoordinatorLayout.kt:38)
    at android.view.animation.Animation.dispatchAnimationEnd(Animation.java:1171)
    at android.view.animation.AnimationSet.getTransformation(AnimationSet.java:418)
    at android.view.animation.Animation.getTransformation(Animation.java:1190)
    at android.view.View.applyLegacyAnimation(View.java:25871)
    at android.view.View.draw(View.java:26010)
    at android.view.ViewGroup.drawChild(ViewGroup.java:4810)
    at com.swmansion.rnscreens.ScreenStack.performDraw(ScreenStack.kt:335)
    at com.swmansion.rnscreens.ScreenStack.access$performDraw(ScreenStack.kt:20)
    at com.swmansion.rnscreens.ScreenStack$DrawingOp.draw(ScreenStack.kt:348)
    at com.swmansion.rnscreens.ScreenStack.drawAndRelease(ScreenStack.kt:304)
    at com.swmansion.rnscreens.ScreenStack.dispatchDraw(ScreenStack.kt:314)
    at android.view.View.updateDisplayListIfDirty(View.java:25161)
    at android.view.ViewGroup.recreateChildDisplayList(ViewGroup.java:4794)
    ...
```

I can see from the breadcrumbs, that this crash happens, while we are still executing `startRemovalTransition` (there is only a 9ms difference between the two calls).
I think the problem here is that `startRemovalTransition` can be called from the JS thread, setting `isBeingRemoved = true`. Then on the UI thread the animation is very quickly done , and already calling `endRemovalTransition`, concurrently modifying the `intransitionViews` list.

My fix for this right now would be to always run `startRemovalTransition` from the UI thread. What do you think about that change?
Would look something like:
```kt
    // Note: this can be called from the JS thread 👀
   fun startRemovalTransition() {
        // Dispatch on the main thread if not already on it
        val mainLooper = Looper.getMainLooper()
        if (Looper.myLooper() != mainLooper) {
            Handler(mainLooper).post {
                startRemovalTransition()
            }
            return
        }

        if (isBeingRemoved) return
        isBeingRemoved = true

        startTransitionRecursive(this)
    }
```

kkafar:
I see.
The problem here is that we had at least two attempts to schedule the call to `startRemovalTransition` on UI & we backed out. I'll try to find the old PRs

kkafar:
We have this one here: https://github.com/software-mansion/react-native-screens/pull/2964
& it seems I remembered things wrong. We just didn't finalise the PR. 

I'll see to that one first & see whether we can land it before we land this one.

kkafar:
I've landed https://github.com/software-mansion/react-native-screens/pull/2964, we can update this PR now

michael-pingo-ai:
I believe this PR might fix the issue I’ve been seeing with @react-navigation/stack: [GSTJ/react-native-magic-modal#155](https://github.com/GSTJ/react-native-magic-modal/issues/155).
We’re encountering it for roughly 10% of Android users, particularly on devices with stricter view teardown behavior — it seems to be timing-related.

kkafar (commented):
Playing around here a bit. There is one part I do not like - in nested stack scenario we duplicate much of the view hierarchy, which  can be potentially quite big. I'll want us to stop collecting the views during recursion descent, when we hit another stack.

NathanAB:
FWIW I've been experiencing this crash as well, and Claude found this PR and identified this as the issue + fix.

fadookie:
@kkafar Hello, we've started seeing this crash in production after switching to Fabric architecture, any news on getting an official fix in for this merged? I think for now we're gonna try this PR out as a patch. Thanks!

kkafar:
Hey @fadookie. I haven't peeked here for a while, but what I see from my previous comments it still requires some attention.

I don't really have capacity in incoming days to push this, but next week is likely. 

Absolutely give me feedback on how the patch is doing in production, please! This will be really helpful to gain confidence to land these changes.

Rob-latticetraining:
Hey @kkafar. Just as a heads up I pulled this fix in as a patch-package for @4.24.0 for our app on the new arch / fabric.

Since the patched build went out our crash rate has dropped drastically and were no longer seeing reoccurrence of this issue.

shahidrogers:
+1 in support of this fix.

I'm hitting the same `IllegalStateException: The specified child already has a parent` (`ReactClippingViewManager.addView`, during Fabric mount) after adopting the New Architecture. 

Environment: **RN 0.86.0, react-native-screens 4.25.2, Fabric enabled, Android**.

I applied the Kotlin from this PR via `patch-package`. For this crash signature, volume went from **several hundred events on the last unpatched release to a single straggler on the first patched build**, effectively eliminated. I've since rebased the patch onto 4.25.2 and can confirm it's **still required** on the latest stable: `endTransitionRecursive` in `Screen.kt` is unchanged, so the bug is still present.

This lines up with what @fadookie and @Rob-latticetraining reported above.

SirJalias:
Why is not this being merged?

jsnavarroc:
Another production data point, from a platform that hasn't come up in this thread yet: **Android TV / Fire TV** (D-pad navigation, no touch).

We hit the same `startViewTransition` leak and independently arrived at essentially the same fix as this PR before finding it, which I take as a good sign for the approach.

**Environment:** RN 0.83, react-native-screens 4.25.2, Fabric (New Architecture), Hermes. Also reproduced on 4.15.4 — `endTransitionRecursive` is byte-identical in both, so the bug spans a wide version range.

### Measurements

Device: a 32-bit (`armeabi-v7a`) Android 14 TV, release build. One cycle = enter a content screen from the home screen, press back, then force `am send-trim-memory RUNNING_CRITICAL`. `View` counts from `dumpsys meminfo`, measured on the home screen both times so the baseline is identical.

| Cycle | Views without fix | Views with fix |
|---|---|---|
| base | 412 | 412 |
| 1 | 596 | 420 |
| 4 | 1,148 | 444 |
| 8 | ~1,884 | 476 |

**184 retained views per navigation → 8. A 96% reduction.**

Two things worth noting about the unpatched numbers:

- The leak is **perfectly linear**: exactly +184 per cycle, zero deviation across 7 cycles. Each screen instance mounts exactly 256 views and releases exactly 72, in cycle 1 and in cycle 7 alike. So the mount path is healthy and this is purely a release problem.
- The retained views **survive forced GC**. After `RUNNING_CRITICAL` the system freed 20 MB of PSS and 20 MB of swap, but the view count did not move by one (3,704 before, 3,704 after). They are held by a live reference, not awaiting collection.

### One note on the reverse-order detail

Our first version of the patch walked `transitioningViews` in insertion order and still measured the reduction above. Reading the `asReversed()` comment in this PR made it clear that was accidental, not correct: parents processed before their children make the children's `endViewTransition` a no-op via `dispatchDetachedFromWindow`. We're switching to the reverse walk. Worth keeping that comment in whatever lands, it's not obvious from the call site.

### On the recycling question

@kkafar, regarding view recycling being off by default: as @alsajoo notes above, in RN 0.83 the per-type flags (`enableViewRecyclingForView`, `...ForImage`, `...ForText`) default to **`true`** even though the master `enableViewRecycling()` is `false`, and `ReactViewManager` gates on the per-type flag alone. So on 0.83 plain `ReactViewGroup` recycling is effectively on out of the box. That may change the calculus on whether the screens-side fix alone is sufficient there.

### What would help downstream

This PR is currently in conflict with `main`, so it isn't mergeable as-is. For those of us carrying this via `patch-package`, the patch stays necessary until three things happen: the fix lands, a release ships with it, and we upgrade to that release. A rebase and a release including it would let several teams drop their patches entirely.

Either way, the `endRemovalTransition` half of this stands on its own: it's a real leak independent of recycling, and it's been reproduced now by at least four separate teams across phones and TV. Happy to run any further measurement that would help land it.

tibbe:
Some supporting data from a production app (react-native-screens 4.24.0/4.25.2, React Native 0.81.6/0.82.1/0.83.10, new architecture, Android).

We have a recurring production crash, `IllegalStateException: addViewAt: failed to insert view [N] into parent [M] at index K`, caused by `The specified child already has a parent`, thrown from `SurfaceMountingManager.addViewAt` within a couple of seconds of a stack navigation (14 events / 5 users over 30 days in our fleet).

To understand it we instrumented `Screen.startRemovalTransition` / `endRemovalTransition` on 4.25.2 and drove an automated push/pop stress harness on an emulator (Pixel 9, API 35):

- Stock code: every pop marked ~86 subtree views via `startViewTransition`, and `endTransitionRecursive` ended 0 of them, on all 102 pops observed. At animation end `parent.children` is already empty: Fabric removes the subtree during the exit animation, and a child removed while in transition moves to `mDisappearingChildren` (with `mParent` still set), which the children walk never visits. So every pop strands its entire subtree in the transitioning state.
- With a fix equivalent to this PR (record the (parent, child) pairs in `startTransitionRecursive`, end them in reverse order in `endRemovalTransition`): 310 of 324 pops stranded 0 views. The remaining 14 pops each stranded 21 views; in those cases the popped subtree had react-native-reanimated exit animations (`ZoomOut`) in flight, and ending the transition on both the recorded parent and the view's current parent still left them in the disappearing state.

On the crash link: when Fabric later re-inserts a stranded view, the guard in `SurfaceMountingManager.addViewAt` does fire, but its evasive `viewParent.removeView(view)` is also deferred for an in-transition child, so `ViewGroup.addViewInner` throws. So the stranded views are not only a memory leak — any re-insert of one is a hard crash.

PR #3251, opened 2025-09-25
fix(iOS, Stack v4): add `systemDefault` option to `autoCapitalize` prop in `SearchBar`

kligarski:
## Description

Ensures consistent default behavior for `autoCapitalize` prop in `SearchBar`.

This will require changes in `react-navigation`'s docs (https://github.com/software-mansion/react-native-screens-labs/issues/470).

### Before

Prior to this PR, changing `autoCapitalize` to `none` on Fabric required changing the value to something else and then changing it back to `none` as `none` was the default value in Fabric spec file for `SearchBar` (controller would not get the updated value and used UIKit default - `sentences`).

https://github.com/user-attachments/assets/71da7edc-4b99-4d9c-8a2c-b92da6ca0bfd

### After

In this PR, bug described in section above is fixed. Consistent behavior across platforms and architectures is ensured. 

https://github.com/user-attachments/assets/4bb00c22-0cfb-4511-90f0-c40d69ce75f7

`systemDefault` option allows correct handling of different search input default behaviors on Android and iOS:
- on Android, default behavior is the same as `none`,
- on iOS, default behavior is the same as `sentences`.

## Changes

- add `systemDefault` to `autoCapitalize` in `SearchBar`
- handle `systemDefault` on both platforms and both architectures
- parse JS `autoCapitalize` prop by mapping `undefined` to `systemDefault` to ensure consistent behavior between Paper and Fabric: on Paper, setter does not run if we pass `undefined` as the value of the prop whereas on Fabric, default value of `systemDefault` is used.

Run `Test758` with commented-out search bar options. Add `autoCapitalize: 'none'` and verify correct behavior.

## Checklist

- [x] Updated TS types
- [ ] Updated documentation: 
- [x] Ensured that CI passes

### DIFF
--- a/src/components/SearchBar.tsx
+++ b/src/components/SearchBar.tsx
@@ -88,27 +88,49 @@ function SearchBar(
     return View as unknown as React.ReactNode;
   }
 
+  // This is necessary only for legacy architecture (Paper).
+  const parsedProps = parseUndefinedPropsToSystemDefault(props);
+
+  const {
+    obscureBackground,
+    hideNavigationBar,
+    onFocus,
+    onBlur,
+    onSearchButtonPress,
+    onCancelButtonPress,
+    onChangeText,
+    ...rest
+  } = parsedProps;
+
   return (
     <NativeSearchBar
       ref={searchBarRef}
-      {...props}
+      {...rest}
       obscureBackground={parseBooleanToOptionalBooleanNativeProp(
-        props.obscureBackground,
+        obscureBackground,
       )}
       hideNavigationBar={parseBooleanToOptionalBooleanNativeProp(
-        props.hideNavigationBar,
+        hideNavigationBar,
       )}
-      onSearchFocus={props.onFocus as DirectEventHandler<SearchBarEvent>}
-      onSearchBlur={props.onBlur as DirectEventHandler<SearchBarEvent>}
+      onSearchFocus={onFocus as DirectEventHandler<SearchBarEvent>}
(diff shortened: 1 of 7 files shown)


t0maboro (approved):

src/components/SearchBar.tsx
```diff
@@ -88,27 +88,49 @@ function SearchBar(
...
       }
-      onChangeText={props.onChangeText as DirectEventHandler<ChangeTextEvent>}
+      onChangeText={onChangeText as DirectEventHandler<ChangeTextEvent>}
     />
   );
 }
 
+// This function is necessary for legacy architecture (Paper) to ensure
+// consistent behavior for props with `systemDefault` option.
+function parseUndefinedPropsToSystemDefault(
```
kmichalikk, line 130: nit: I'd say something like `decorateAutocapitalizeWithSystemDefaultIfNeeded` would sound better?

kmichalikk (approved):

src/components/SearchBar.tsx
kligarski: I assumed that there might be more props that will require this kind of handling (but at the moment I'm not sure if there will be more) and wanted to use only one function for all of them so that it's easier to delete it in the future when we drop support for legacy architecture.

Also, isn't `decorate` a little bit too much for simple `if (value === undefined) return "systemDefault"`?

I'm open for suggestions here.

PR #3252, opened 2025-09-25
chore: Move scroll edge effect applicator to helpers/scroll-view

kmichalikk:
## Description

In response to https://github.com/software-mansion/react-native-screens/pull/3212#discussion_r2375933061

This PR moves RNSScrollViewEdgeEffectApplicator to helpers/scroll-view directory which makes more sense for our case.


kkafar (approved):
Thanks. This should work, right? We don't have any relative imports that need updates?

kmichalikk:
It compiled and ran with the test from https://github.com/software-mansion/react-native-screens/pull/3244; I didn't see any relative imports.

PR #3254, opened 2025-09-25
Fix custom screen swipe transitions

piaskowyk:
## Description

This PR fixes custom swipe-back transitions. The issue arose because, in the new version of React Native, they changed the way to obtain the shadow node wrapper. Since the ref to the screen is a native ref and we only need this on Fabric, it was possible to simplify the way we obtain that reference. Additionally, I found a small issue with the propagation of the ref and fixed it in the `Screen` implementation.

| before | after |
| --- | --- |
| image | <video src="https://github.com/user-attachments/assets/1352f57e-c789-4735-9e1d-0905ad396a1d" /> |

Example: https://github.com/software-mansion/react-native-reanimated/blob/main/apps/common-app/src/apps/reanimated/examples/ScreenTransitionExample.tsx

## Checklist

- [x] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/src/gesture-handler/fabricUtils.ts
+++ b/src/gesture-handler/fabricUtils.ts
@@ -1,5 +1,7 @@
 'use strict';
 
+import { View } from "react-native";
+
 /* eslint-disable */
 
 type LocalGlobal = typeof global & Record<string, unknown>;
@@ -14,7 +16,7 @@ export type HostInstance = {
   _viewConfig: Record<string, unknown>;
 };
 
-export function getShadowNodeWrapperAndTagFromRef(ref: HostInstance | null): {
+export function getShadowNodeWrapperAndTagFromRef(ref: View | null): {
   shadowNodeWrapper: any;
   tag: number;
 } {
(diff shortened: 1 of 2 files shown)


src/gesture-handler/fabricUtils.ts
```diff
@@ -10,77 +10,25 @@ export function isFabric() {
   return !!(global as LocalGlobal).RN$Bridgeless;
 }
 
-export type ShadowNodeWrapper = {
-  __hostObjectShadowNodeWrapper: never;
+export type HostInstance = {
+  __internalInstanceHandle: Record<string, any>;
+  __nativeTag: number;
+  _viewConfig: Record<string, unknown>;
```
kkafar, line 16: I guess our definitions were already outdated, but will it work with 80, 81 & 82 now?

kkafar:
I'll merge it once I get to test this (soon)

src/gesture-handler/fabricUtils.ts
piaskowyk: I haven't tested it with all of these versions (only with 0.81), but to my knowledge, it should work with all of them.

PR #3258, opened 2025-09-26
fix(Android, Fabric, Stack v4): add `autoFocus` to SearchBar Fabric spec file

kligarski:
## Description

Adds `autoFocus` prop to `SearchBar`'s Fabric spec file. It was probably missed when migrating `SearchBar` to new architecture.

## Changes

- add `autoFocus` to Fabric spec file
- update `SearchBarManager`
- sync architectures

## Screenshots / GIFs

https://github.com/user-attachments/assets/08ef3aa0-f282-429b-8c54-a4cfe768382f

Run `SearchBar` example screen with `autoFocus: true` prop set.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/SearchBarManager.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/SearchBarManager.kt
@@ -55,11 +55,11 @@ class SearchBarManager :
     }
 
     @ReactProp(name = "autoFocus")
-    fun setAutoFocus(
+    override fun setAutoFocus(
         view: SearchBarView,
-        autoFocus: Boolean?,
+        autoFocus: Boolean,
     ) {
-        view.autoFocus = autoFocus ?: false
+        view.autoFocus = autoFocus
     }
 
     @ReactProp(name = "barTintColor", customType = "Color")
--- a/src/fabric/SearchBarNativeComponent.ts
+++ b/src/fabric/SearchBarNativeComponent.ts
@@ -38,6 +38,7 @@ export interface NativeProps extends ViewProps {
   onCancelButtonPress?: DirectEventHandler<SearchBarEvent> | null;
   onChangeText?: DirectEventHandler<ChangeTextEvent> | null;
   hideWhenScrolling?: WithDefault<boolean, true>;
+  autoFocus?: WithDefault<boolean, false>;
   autoCapitalize?: WithDefault<AutoCapitalizeType, 'none'>;
   placeholder?: string;
   placement?: WithDefault<SearchBarPlacement, 'automatic'>;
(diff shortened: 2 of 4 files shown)


src/fabric/SearchBarNativeComponent.ts
```diff
@@ -38,6 +38,7 @@ export interface NativeProps extends ViewProps {
   onCancelButtonPress?: DirectEventHandler<SearchBarEvent> | null;
   onChangeText?: DirectEventHandler<ChangeTextEvent> | null;
   hideWhenScrolling?: WithDefault<boolean, true>;
+  autoFocus?: WithDefault<boolean, false>;
```
t0maboro: Shouldn't we put this under the Android-only marker?

kligarski: You're right, I did not see the section before. Thanks for noticing.

https://github.com/software-mansion/react-native-screens/pull/3258/commits/93f8ef051a0cafce1668fffeb168de54779290a5

t0maboro (approved):

android/src/main/java/com/swmansion/rnscreens/SearchBarManager.kt
```diff
@@ -55,11 +55,11 @@ class SearchBarManager :
     }
 
     @ReactProp(name = "autoFocus")
-    fun setAutoFocus(
+    override fun setAutoFocus(
         view: SearchBarView,
-        autoFocus: Boolean?,
```
kkafar, line 60: Can't this be null on old architecture?

kkafar: E.g. when you set it to `true` initially, and then switch back to `undefined`

kligarski: No, we always get either `true` or `false`. The setter is called from `RNSSearchBarManagerDelegate.java`:
```java
      case "autoFocus":
        mViewManager.setAutoFocus(view, value == null ? false : (boolean) value);
        break;
```

I guess that the default value comes from syncing with `Fabric` (default is defined in spec file), e.g.
```java
      case "shouldShowHintSearchIcon":
        mViewManager.setShouldShowHintSearchIcon(view, value == null ? true : (boolean) value);
        break;
```

```ts
  shouldShowHintSearchIcon?: WithDefault<boolean, true>;
```

kkafar: Okay then. The explanation makes sense to me.

kkafar (approved):
Looks good then!

PR #3259, opened 2025-09-29
chore: bump FabricExample & lib to `react-native@0.82`

kligarski:
## Description

Bumps `FabricExample` & library to `react-native@0.82`.

Legacy architecture example hasn't been bumped as [starting from `react-native@0.82`, you can't set `newArchEnabled` to `false`](https://github.com/facebook/react-native/pull/53025).

## Changes

- apply changes from [Upgrade Helper](https://react-native-community.github.io/upgrade-helper/?from=0.81.1&to=0.82.0-rc.4)
- bump `react-native-reanimated` and `react-native-worklets`
- bump lock files

## Test code and steps to reproduce

Run example apps.

## Checklist

- [x] Included code example that can be used to test this change
- [x] Ensured that CI passes

PR #3261, opened 2025-09-29
fix(Android): sync architectures after bump to `react-native@0.82`

kligarski:
## Description

Syncs architectures after bump to `react-native@0.82`. I must've missed this in https://github.com/software-mansion/react-native-screens/pull/3259 (interesting that CI passed).

## Changes

- run `yarn sync-architectures`

CI

## Checklist

- [x] Ensured that CI passes


kligarski:
I think it is `LayoutShadowNode`, due to the end of legacy architecture starting from `react-native@0.82`: https://github.com/facebook/react-native/commit/c4715886a917eb3eb63aab366de5225090dff5a1.

This is something we might need to keep in mind - we're syncing architectures via `yarn sync-architectures` (and this is deprecated in 0.82 in FabricExample) but Paper example will remain on `0.81` - I guess that in the future this might cause some problems.

Issue #2632, opened 2025-01-17
Your app uses deprecated APIs or parameters for edge-to-edge


gauravSharma2009:
### Description

Using latest version     "react-native-screens": "^4.5.0",
when submitting app to play store getting this recommendation not able to publish app on play store .

1 action recommended
Your app uses deprecated APIs or parameters for edge-to-edge
One or more of the APIs that you use or parameters that you set for edge-to-edge and window display have been deprecated in Android 15. Your app uses the following deprecated APIs or parameters:

android.view.Window.getStatusBarColor
android.view.Window.setStatusBarColor
android.view.Window.setDecorFitsSystemWindows
androidx.core.view.WindowCompat.setDecorFitsSystemWindows
android.view.Window.setNavigationBarColor
android.view.Window.getNavigationBarColor
LAYOUT_IN_DISPLAY_CUTOUT_MODE_SHORT_EDGES
LAYOUT_IN_DISPLAY_CUTOUT_MODE_DEFAULT

These start in the following places:

com.facebook.react.modules.statusbar.StatusBarModule$setColor$1.runGuarded
com.facebook.react.modules.statusbar.StatusBarModule.getTypedExportedConstants
com.facebook.react.views.view.WindowUtilKt.statusBarHide
com.facebook.react.views.view.WindowUtilKt.statusBarShow
com.google.android.material.bottomsheet.BottomSheetDialog.onCreate
com.google.android.material.internal.EdgeToEdgeUtils.applyEdgeToEdge
com.swmansion.rnscreens.ScreenWindowTraits.setColor$react_native_screens_release
com.swmansion.rnscreens.ScreenWindowTraits.setNavigationBarColor$react_native_screens_release
com.swmansion.rnscreens.ScreenWindowTraits.setNavigationBarTranslucent$react_native_screens_release
androidx.activity.EdgeToEdgeApi28.adjustLayoutInDisplayCutoutMode

Please check this many others are facing this issue and not able to publish new version on playstore.

### Steps to reproduce

need to publish app on play store . 
link that other users are also facing this issue : https://github.com/facebook/react-native/issues/48256

### Snack or a link to a repository

https://github.com/gauravSharma2009/guidezmobile/tree/android

### Screens version

4.5.0

### React Native version

0.76.5

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

kkafar:
[...]

PR #3264, opened 2025-09-30
feat!(Android, Stack v4): remove native code/deps related to props deprecated by edge-to-edge enforcement

kligarski:
## Description

Removes native code related to `statusBarColor`, `statusBarTranslucent`, `navigationBarColor`, `navigationBarTranslucent`, `topInsetEnabled` props.

Retained in JS for backward compatibility.

For reasons why those props are deprecated, please see https://github.com/software-mansion/react-native-screens/pull/2638.

After this PR is merged, we should prepare corresponding PRs to `react-navigation` and `react-navigation`'s docs. I created an issue to handle it: https://github.com/software-mansion/react-native-screens-labs/issues/487.

Closes https://github.com/software-mansion/react-native-screens/issues/2632.
Closes https://github.com/software-mansion/react-native-screens-labs/issues/485.

## Changes

- remove native code related to `statusBarColor`, `statusBarTranslucent`, `navigationBarColor`, `navigationBarTranslucent`, `topInsetEnabled` props
- remove usage of those props in example apps (with appropriate comment to maintain full context of the tests)
- remove `edge-to-edge` detection and `react-native-is-edge-to-edge`

Verify that app builds and runs.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
+++ b/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
@@ -100,7 +100,7 @@ Defaults to `false`.
 
 ### `navigationBarColor` (Android only)
 
-This prop has been **removed** due to [edge-to-edge enforcement starting from Android SDK 35](https://developer.android.com/about/versions/15/behavior-changes-15#ux). Setting it has no effect as native code related to this prop has been removed. Retained only for backward compatibility.
+This prop has been **deprecated** due to [edge-to-edge enforcement starting from Android SDK 35](https://developer.android.com/about/versions/15/behavior-changes-15#ux). Setting it has no effect as native code related to this prop has been removed. Kept only for backward compatibility. Will be removed in next major release.
 
 Sets the navigation bar color. Defaults to initial status bar color.
 
@@ -110,7 +110,7 @@ Sets the visibility of the navigation bar. Defaults to `false`.
 
 ### `navigationBarTranslucent` (Android only)
 
-This prop has been **removed** due to [edge-to-edge enforcement starting from Android SDK 35](https://developer.android.com/about/versions/15/behavior-changes-15#ux). Setting it has no effect as native code related to this prop has been removed. Retained only for backward compatibility.
+This prop has been **deprecated** due to [edge-to-edge enforcement starting from Android SDK 35](https://developer.android.com/about/versions/15/behavior-changes-15#ux). Setting it has no effect as native code related to this prop has been removed. Kept only for backward compatibility. Will be removed in next major release.
 
 Sets the translucency of the navigation bar. Defaults to `false`.
 
@@ -287,7 +287,7 @@ Defaults to `fade` on iOS and `none` on Android.
 
 ### `statusBarColor` (Android only)
 
-This prop has been **removed** due to [edge-to-edge enforcement starting from Android SDK 35](https://developer.android.com/about/versions/15/behavior-changes-15#ux). Setting it has no effect as native code related to this prop has been removed. Retained only for backward compatibility.
+This prop has been **deprecated** due to [edge-to-edge enforcement starting from Android SDK 35](https://developer.android.com/about/versions/15/behavior-changes-15#ux). Setting it has no effect as native code related to this prop has been removed. Kept only for backward compatibility. Will be removed in next major release.
 
 Sets the status bar color (similar to the `StatusBar` component). Defaults to initial status bar color.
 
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
```diff
@@ -26,12 +26,10 @@ open class CustomToolbar(
     val config: ScreenStackHeaderConfig,
 ) : Toolbar(context) {
     // Switch this flag to enable/disable display cutout avoidance.
-    // Currently this is controlled by isTopInsetEnabled prop.
-    private val shouldAvoidDisplayCutout
-        get() = config.isTopInsetEnabled
+    // Currently this is hard-coded.
```
kmichalikk: I think we can reword the whole comment here, or remove it, since there is no "real" flag that changes the value

Maybe we can say
> Since SDK >= 35 has been made mandatory, [*consequences*], and the value is hard-coded (see [*issue no*])

?

android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
```diff
@@ -97,6 +98,10 @@ open class ScreenViewManager :
         view.onFinalizePropsUpdate()
     }
 
+    private fun logNotAvailable(propName: String) {
+        Log.w("[RNScreens]", "$propName prop is not available on Android")
+    }
+
```
kmichalikk, line 104: This is a 3rd definition of such method, and there is //TODO on one, so maybe it's time we make an issue for it?

kmichalikk (commented):
Looks good

The only thing I'm not sure about is that we leave the props commented out in Tests. Do we really need it? Did the test rely on the prop - if so, maybe we should remove the test; if not - they don't add much value IMO. But that's just me, if we agreed to do it that way, then it's ok.

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
```diff
@@ -98,12 +98,6 @@ Unfortunately the same behavior is not available on iOS since the behavior of na
 
 Defaults to `false`.
 
-### `navigationBarColor` (Android only)
-
-This prop is **deprecated**. See [here](https://developer.android.com/about/versions/15/behavior-changes-15#ux).
-
-Sets the navigation bar color. Defaults to initial status bar color.
```
kkafar, line 105: As long as these props stay in API, we should keep record of them here. Just add appropriate comment that they're noops kept for backward compat & native code is removed.

android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
kkafar: +1

kkafar: Some explanation why it is hardcoded could be nice.

android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
kligarski: I created a ticket for this: https://github.com/software-mansion/react-native-screens-labs/issues/491

kligarski:
> The only thing I'm not sure about is that we leave the props commented out in Tests. Do we really need it? Did the test rely on the prop - if so, maybe we should remove the test; if not - they don't add much value IMO. But that's just me, if we agreed to do it that way, then it's ok.

After internal discussion with @kkafar, we decided that:
- we should check and remove all redundant/outdated test cases but this probably includes many more examples that aren't related to this PR - we should do it separately and I created a ticket for this: https://github.com/software-mansion/react-native-screens-labs/issues/492,
- comments are okay in this case.

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
kligarski: https://github.com/software-mansion/react-native-screens/pull/3264/commits/ee53515060dd13f56f16d885473af2733aceb9f9

android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3264/commits/fcc733d003b73c28cb71c7b5457fd0d7670b695b

kmichalikk (approved):

PR #3265, opened 2025-09-30
fix(iOS): Fix content pop with horizontal scroll

kmichalikk:
Fixes https://github.com/software-mansion/react-native-screens/pull/3173#issuecomment-3340331930

## Description
https://github.com/software-mansion/react-native-screens/pull/3173 introduced a bug that makes interactiveContentPopGesture dismiss a Screen with content ScrollView immediately, instead of actually scrolling the contents. ~From my testing, the native behavior is to prevent dismissing the screen with such ScrollView entirely.~ The screen should prevent dismissing the screen, only allowing it by swiping from the edge. This change aims to match this behavior. During testing, I also found out, that the previous recognizer works unreliably (sometimes it allows for scrolling, sometimes it dismisses the screen). I'll look into that in separate issue.

## Changes
Modified `gestureRecognizer:shouldBeRequiredToFailByGestureRecognizer`, added `gestureRecognizer:shouldRequireFailureOfGestureRecognizer` (only in iOS >= 26).

## Before / After

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/00b91d0c-d303-4e88-aef5-ec9659ed9121" /> | <video src="https://github.com/user-attachments/assets/f661d12d-4d8a-48ef-897e-d095b5778531" /> |


JustJoostNL:
Thanks for the quick PR!

kkafar (approved):
Seems reasonable & close to what we discussed yesterday. GJ!

kmichalikk:
@kkafar please take a look at [466908c](https://github.com/software-mansion/react-native-screens/pull/3265/commits/466908cc5bda3b086e4b3c95d9fb42e6d532fff8)

kkafar (approved):
I've checked the runtime. Seems to work nice. Good job.

PR #3271, opened 2025-10-02
Verify that ScrollView gesture hasn't begun when triggering pan gesture

kmichalikk:
Closes https://github.com/software-mansion/react-native-screens-labs/issues/489

## Description

This PR addresses the issue that, when using custom PanGestureRecognizer on iOS 26 (for handling custom animations on swipe) with a screen that has a horizontal ScrollView, and when swiping for the second time before the view stopped scrolling, the screen is popped with pan gesture.

## Changes

Added check for ScrollGestureRecognizer state=Began in `shouldRecognizeSimultaneouslyWithGestureRecognizer` to return NO.

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/b7fd4bcc-1902-44b7-8f85-26dc215619db" /> | <video src="https://github.com/user-attachments/assets/8d76b11e-350f-4866-b751-43b88268795c" /> |

-->

Use Test3265 on iOS 26; make sure that the screen has `animation` set to custom animation and `animationMatchesGesture: true`.

### DIFF
--- a/ios/RNSScreenStack.mm
+++ b/ios/RNSScreenStack.mm
@@ -1242,7 +1242,8 @@ RNS_IGNORE_SUPER_CALL_END
     BOOL isBackGesture = [panGestureRecognizer translationInView:panGestureRecognizer.view].x > 0 &&
         _controller.viewControllers.count > 1;
 
-    if (gestureRecognizer.state == UIGestureRecognizerStateBegan || isBackGesture) {
+    if (otherGestureRecognizer.state == UIGestureRecognizerStateBegan ||
+        gestureRecognizer.state == UIGestureRecognizerStateBegan || isBackGesture) {
       return NO;
     }
 


kkafar (changes requested):
I approve the native code.
To land this PR we need to add dedicated test case. Let's do it, please.

apps/src/tests/Test3271.tsx
```diff
@@ -0,0 +1,49 @@
+import { NavigationContainer, NavigationProp } from '@react-navigation/native';
+import { createNativeStackNavigator } from '@react-navigation/native-stack';
+import React, { Button, ScrollView, Text, View } from 'react-native';
```
t0maboro: `React` should be imported from `react`, not as a default export from RN

kkafar: +1

kkafar (commented):
We need to fix the imports before landing. Other than that 🟢

apps/src/tests/Test3271.tsx
kmichalikk: [9f9dc04](https://github.com/software-mansion/react-native-screens/pull/3271/commits/9f9dc0419b6055aad0232c02283eba550602f994)

PR #3274, opened 2025-10-03
feat(iOS, SplitView): SplitView synchronous updates

t0maboro:
## Description

We have an access to API for sending synchronous state updates, which makes it possible to resolve the issue of jumping content when resizing columns in SplitView.

> [!CAUTION]  
> Our assumption is that SplitView will be supported since RN minor version 0.82, compiling it with older version may fail!

Fixes https://github.com/software-mansion/react-native-screens-labs/issues/493

## Changes

- Added `UpdateMode::unstable_Immediate` to SplitView shadow node state updates

### Before

https://github.com/user-attachments/assets/02d306df-aa1f-4e48-b1b6-25d36814b3aa

### After

https://github.com/user-attachments/assets/76eb6bce-e05f-43fb-b8c3-5185c8c7ad44

Went through SplitView examples. Checked that it still compiles successfully with RN 0.81.

Some follow-up actions are needed to be taken:
- https://github.com/software-mansion/react-native-screens-labs/issues/495
- https://github.com/software-mansion/react-native-screens-labs/issues/496

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/gamma/split-view/RNSSplitViewScreenShadowStateProxy.mm
+++ b/ios/gamma/split-view/RNSSplitViewScreenShadowStateProxy.mm
@@ -3,7 +3,6 @@
 
 #import <React/RCTAssert.h>
 #import <React/RCTConversions.h>
-#import <cxxreact/ReactNativeVersion.h>
 #import <rnscreens/RNSSplitViewScreenShadowNode.h>
 
 namespace react = facebook::react;
@@ -45,13 +44,7 @@ namespace react = facebook::react;
 
   if (!CGRectEqualToRect(frame, _lastScheduledFrame)) {
     auto newState = react::RNSSplitViewScreenState{RCTSizeFromCGSize(frame.size), RCTPointFromCGPoint(frame.origin)};
-    _state->updateState(
-        std::move(newState)
-#if REACT_NATIVE_VERSION_MINOR >= 82
-            ,
-        facebook::react::EventQueue::UpdateMode::unstable_Immediate
-#endif
-    );
+    _state->updateState(std::move(newState), facebook::react::EventQueue::UpdateMode::unstable_Immediate);
 
     _lastScheduledFrame = frame;
   }


kmichalikk (approved):

RNScreens.podspec
```diff
@@ -19,6 +20,7 @@ min_supported_visionos_version = "1.0"
 
 rnscreens_cpp_flags = []
 
+rnscreens_cpp_flags << "-DREACT_NATIVE_MINOR_VERSION=#{rnscreens_find_react_native_minor_version()}"
```
kkafar: You don't need this. There is `ReactNativeVersion.h` header which defines for you. I think we can use it in gamma project.

ios/gamma/split-view/RNSSplitViewScreenShadowStateProxy.mm
```diff
@@ -44,7 +44,13 @@ - (void)updateShadowStateWithFrame:(CGRect)frame
 
   if (!CGRectEqualToRect(frame, _lastScheduledFrame)) {
     auto newState = react::RNSSplitViewScreenState{RCTSizeFromCGSize(frame.size), RCTPointFromCGPoint(frame.origin)};
-    _state->updateState(std::move(newState));
+    _state->updateState(
+        std::move(newState),
+#if REACT_NATIVE_MINOR_VERSION >= 82
+        facebook::react::EventQueue::UpdateMode::unstable_Immediate
+#endif
+    );
```
kkafar: Will this compile on 0.81? There is a trailing comma in line 48. Didn't know C++ allows for these. TIL I guess

kkafar (changes requested):
One important remark here.

ios/gamma/split-view/RNSSplitViewScreenShadowStateProxy.mm
t0maboro: I checked it and you're right here, 
```cpp
void foo(int a, int b=2) {
    std::cout << a << " " << b << std::endl;
}

int main() {
    foo(1,); // fails
    return 0;
}
```
I'm reverting the last commit

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3274/commits/fff434044c6da8d12a53b9bb2f81e8c282b32ead

RNScreens.podspec
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3274/commits/bef37e0a22a3004ced9e8d8843ab3e9f75b26ddd

t0maboro:
> Let's update PR description before we land this. Beside cleanup put there a note that we require 0.82 now with this commit.

@kkafar I updated the description

kkafar (approved):

PR #3276, opened 2025-10-05
fix(Android): Restore TV focus in Jetpack Compose views

msynowski:
## Description

Currently, when navigating back to a screen containing a Compose-based view, lastFocusedChild is assigned to the internal AndroidComposeView leaving the screen without any focused element.

Instead of targeting the internal AndroidComposeView, Screens would restore it now to it's Android layout parent, that all custom RN views have, allowing developers to restore it back to compose via listeners or overriding the requestFocus method.

## Changes
Restore focus from Jetpack Compose views to the parent Android layout, allowing to redirect it back to Compose via focus listeners/overriding requestFocus

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
@@ -442,10 +442,33 @@ class ScreenStackFragment :
         }
     }
 
+    private fun isComposeView(view: View): Boolean {
+        val name = view.javaClass.name
+        return name.startsWith("androidx.compose.ui.platform.")
+    }
+
+    private fun findComposeContainer(view: View?): View? {
+        var current = view
+
+        while (current != null) {
+            if (!isComposeView(current)) return current
+            current = (current.parent as? View)
+        }
+
+        return null
+    }
+
     private fun findLastFocusedChild(): View? {
         var view: View? = screen
         while (view != null) {
-            if (view.isFocused) return view
+            if (view.isFocused) {
+                if (isComposeView(view)) {
+                    return findComposeContainer(view)
+                }
+
+                return view
+            }
+
             view = if (view is ViewGroup) view.focusedChild else null
         }
 


android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
```diff
@@ -442,10 +442,33 @@ class ScreenStackFragment :
         }
     }
 
+    private fun isComposeView(view: View): Boolean {
+        val name = view.javaClass.name
+        return name.startsWith("androidx.compose.ui.platform.")
```
msynowski, line 447: Found out it doesn't work in release mode - R8 strips class names.

Normally I'd just check if view is instance of a compose class, however here adding compose as a dependency doesn't seem an option. 

Looking forward to some advice on how to tackle this

msynowski: Best solution would be adding proguard rules to keep compose classes - lmk if you're cool with it

kkafar: Good that you catched this.  

Isn't proguard configured on the app level? This is not really a proper solution from the library perspective & I'd like to avoid additional installation steps here. 

I'll try to come up with something sensible when doing review & testing this. 

I've withheld myself with review, before I fully understand the code & see the issue firsthand first. I'll be on PTO until end of the week. On Monday I'll try to do a proper testing & review here.

msynowski: > Isn't proguard configured on the app level? 

Yup. There's this thing called consumerProguardFiles, but it only works when the dependency is distributed as AAR and that's unfortunately not the case

msynowski: > On Monday I'll try to do a proper testing & review here.

Thanks!

PR #3277, opened 2025-10-06
fix(Android, SplitView): Add fallback for SplitView import error

t0maboro:
## Description

SplitView is not currently supported on Android. So, I'm adding a fallback to avoid throwing an error on app launch.

## Changes

- Added fallback components for Android.

### Before

Screenshot 2025-10-06 at 09 29 54

### After

Displaying a clear warning in Metro

Screenshot 2025-10-06 at 09 40 32

Open an example with SplitView (avoid the one with the new stack, because it will throw an error about missing components for the new stack on android)

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/components/gamma/split-view/SplitViewHost.android.tsx
+++ b/src/components/gamma/split-view/SplitViewHost.android.tsx
@@ -1,0 +1,9 @@
+import { View } from 'react-native';
+
+const SplitViewHost = View;
+
+console.warn(
+  'SplitView is supported only for iOS. Consider using an alternative layout for Android.',
+);
+
+export default SplitViewHost;
--- a/src/components/gamma/split-view/SplitViewScreen.android.tsx
+++ b/src/components/gamma/split-view/SplitViewScreen.android.tsx
@@ -1,0 +1,10 @@
+import { View } from 'react-native';
+
+const Column = View;
+const Inspector = View;
+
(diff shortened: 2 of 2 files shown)


kmichalikk (approved):

src/components/gamma/split-view/SplitViewHost.android.tsx
```diff
@@ -0,0 +1,9 @@
+import { View } from 'react-native';
+
+const SplitViewHost = View;
+
+console.warn(
+  'SplitView is supported only for iOS. Consider using an alternative layout for Android.',
```
kkafar:
```suggestion
console.warn(
  '[RNScreens] SplitView is supported only for iOS. Consider using an alternative layout for Android.',
```

We don't do it everywhere in our JS-side logs, but I think this is a good practice.

kkafar (approved):
I have one cosmetic suggestion, beside that we are ✅

src/components/gamma/split-view/SplitViewHost.android.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3277/commits/43499290b9201b41e276f13387b328a2c86b624b

PR #3278, opened 2025-10-06
feat(iOS): Implement and use ContentScrollViewProviding for BottomTabsHostComponentView

kmichalikk:
## Description

TBD

TBD


kmichalikk:
There were probably several new approaches since this had been drafted, unsalvageable.

PR #3282, opened 2025-10-06
feat(Stack): Add support for synchronous state updates on fabric for native stack

t0maboro:
## Description

This PR adds support for synchronous updates in the native stack for React Native 0.82+. Currently, enabling synchronous state updates is possible via an experimental feature flag: `unstable_synchronousStateUpdatesEnabled`. 

At this time, I haven't found a way to perform synchronous updates on Android, so this PR only includes mocks for that platform. I'm creating a ticket for covering that: https://github.com/software-mansion/react-native-screens-labs/issues/497

Fixes https://github.com/software-mansion/react-native-screens-labs/issues/495

## Changes

- Added feature flags in JS per native component for testing synchronous updates:
  - synchronousScreenUpdatesEnabled
  - synchronousHeaderConfigUpdatesEnabled
  - synchronousHeaderSubviewUpdatesEnabled
- Added mocks for android implementation
- Added implementation for all native components on iOS that are performing shadow tree state updates

### Before

https://github.com/user-attachments/assets/bc21afa9-1e6f-4dbd-9dcf-0319b5d72ac7

### After

https://github.com/user-attachments/assets/e131260d-6667-4652-909c-13be77e6fdf3

Adding a new example, which is verifying that with synchronous updates enabled, the form sheet content flicker issues are resolved.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/components/ScreenStackHeaderConfig.tsx
+++ b/src/components/ScreenStackHeaderConfig.tsx
@@ -26,11 +26,6 @@ import ScreenStackHeaderSubviewNativeComponent, {
 import { prepareHeaderBarButtonItems } from './helpers/prepareHeaderBarButtonItems';
 import { isHeaderBarButtonsAvailableForCurrentPlatform } from '../utils';
 
-const HEADER_CONFIG_SYNCHRONOUS_UPDATES_ENABLED =
-  featureFlags.experiment.synchronousHeaderConfigUpdatesEnabled;
-const HEADER_SUBVIEW_SYNCHRONOUS_UPDATES_ENABLED =
-  featureFlags.experiment.synchronousHeaderSubviewUpdatesEnabled;
-
 export const ScreenStackHeaderSubview: React.ComponentType<ScreenStackHeaderSubviewNativeProps> =
   ScreenStackHeaderSubviewNativeComponent;
 
@@ -125,7 +120,7 @@ export const ScreenStackHeaderConfig = React.forwardRef<
       style={styles.headerConfig}
       pointerEvents="box-none"
       synchronousShadowStateUpdatesEnabled={
-        HEADER_CONFIG_SYNCHRONOUS_UPDATES_ENABLED
+        featureFlags.experiment.synchronousHeaderConfigUpdatesEnabled
       }
     />
   );
@@ -140,7 +135,7 @@ export const ScreenStackHeaderBackButtonImage = (
     type="back"
     style={styles.headerSubview}
     synchronousShadowStateUpdatesEnabled={
-      HEADER_SUBVIEW_SYNCHRONOUS_UPDATES_ENABLED
+      featureFlags.experiment.synchronousHeaderSubviewUpdatesEnabled
     }>
     <Image resizeMode="center" fadeDuration={0} {...props} />
   </ScreenStackHeaderSubview>
@@ -156,7 +151,7 @@ export const ScreenStackHeaderRightView = (
       {...rest}
       type="right"
       synchronousShadowStateUpdatesEnabled={
-        HEADER_SUBVIEW_SYNCHRONOUS_UPDATES_ENABLED
+        featureFlags.experiment.synchronousHeaderSubviewUpdatesEnabled
       }
       style={[styles.headerSubview, style]}
     />
@@ -173,7 +168,7 @@ export const ScreenStackHeaderLeftView = (
       {...rest}
       type="left"
       synchronousShadowStateUpdatesEnabled={
-        HEADER_SUBVIEW_SYNCHRONOUS_UPDATES_ENABLED
+        featureFlags.experiment.synchronousHeaderSubviewUpdatesEnabled
       }
       style={[styles.headerSubview, style]}
     />
@@ -188,7 +183,7 @@ export const ScreenStackHeaderCenterView = (props: ViewProps): JSX.Element => {
       {...rest}
       type="center"
       synchronousShadowStateUpdatesEnabled={
-        HEADER_SUBVIEW_SYNCHRONOUS_UPDATES_ENABLED
+        featureFlags.experiment.synchronousHeaderSubviewUpdatesEnabled
       }
       style={[styles.headerSubviewCenter, style]}
     />
@@ -202,7 +197,7 @@ export const ScreenStackHeaderSearchBarView = (
     {...props}
     type="searchBar"
     synchronousShadowStateUpdatesEnabled={
-      HEADER_SUBVIEW_SYNCHRONOUS_UPDATES_ENABLED
+      featureFlags.experiment.synchronousHeaderSubviewUpdatesEnabled
     }
     style={styles.headerSubview}
   />


ios/RNSScreen.mm
```diff
@@ -2204,6 +2220,7 @@ @implementation RNSScreenManager
 RCT_EXPORT_VIEW_PROPERTY(leftScrollEdgeEffect, RNSScrollEdgeEffect);
 RCT_EXPORT_VIEW_PROPERTY(rightScrollEdgeEffect, RNSScrollEdgeEffect);
 RCT_EXPORT_VIEW_PROPERTY(topScrollEdgeEffect, RNSScrollEdgeEffect);
+RCT_EXPORT_VIEW_PROPERTY(unstable_synchronousUpdatesEnabled, BOOL);
```
kligarski: I'm not sure if we need this prop on Paper, as it won't work there but we might want to check if there is any warning if you pass a prop that hasn't been exported.

We can also leave the prop exported but for consistency I'd do it in other components as well (e.g. `RNSScreenStackHeaderConfig`).

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3282/commits/e58ba751c472aa898c64ed312ea50d8cefce638f

apps/src/tests/TestSplitView/index.tsx
```diff
@@ -24,7 +27,7 @@ const App = () => {
   }
 
   return (
-    <SplitViewBaseApp splitViewBaseConfig={splitViewBaseConfig} />
+    <SplitViewWithNativeStackBase splitViewBaseConfig={splitViewBaseConfig} />
```
kkafar: Why does it change here?

kkafar: Is it rename or you changed the example for some reason?

ios/RNSScreen.h
```diff
@@ -94,6 +94,7 @@ namespace react = facebook::react;
 @property (nonatomic) RNSScrollEdgeEffect leftScrollEdgeEffect;
 @property (nonatomic) RNSScrollEdgeEffect rightScrollEdgeEffect;
 @property (nonatomic) RNSScrollEdgeEffect topScrollEdgeEffect;
+@property (nonatomic) BOOL unstable_synchronousUpdatesEnabled;
```
kkafar:
```suggestion
@property (nonatomic, readwrite) BOOL synchronousShadowStateUpdatesEnabled
```

we have many different "updates", I'd like to be more precise here, so that it is clear what this field does. 

Also, our native API is not stable and not really part of our library API, which is in JS. Therefore when doing similar things I omit the `ustable` / `experimental` prefix.

ios/RNSScreen.mm
```diff
@@ -468,6 +477,11 @@ - (void)setTopScrollEdgeEffect:(RNSScrollEdgeEffect)topScrollEdgeEffect
   _topScrollEdgeEffect = topScrollEdgeEffect;
 }
 
+- (void)setUnstable_synchronousUpdatesEnabled:(BOOL)unstable_synchronousUpdatesEnabled
+{
+  _unstable_synchronousUpdatesEnabled = unstable_synchronousUpdatesEnabled;
+}
```
kkafar: This won't be needed when we apply naming changes.

ios/RNSScreenStackHeaderConfig.h
```diff
@@ -65,6 +65,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic) RCTDirectEventBlock onPressHeaderBarButtonItem;
 @property (nonatomic) RCTDirectEventBlock onPressHeaderBarButtonMenuItem;
 #endif
+@property (nonatomic) BOOL unstable_synchronousUpdatesEnabled;
```
kkafar: similar naming suggestions everywhere on native side

ios/RNSScreenStackHeaderConfig.mm
kkafar, line 1: Does this fix any particular problem? I guess my real question is whether we have some reasoning for applying the synchronous updates in header config? The same question applies to other components.

src/fabric/ModalScreenNativeComponent.ts
```diff
@@ -113,6 +113,7 @@ export interface NativeProps extends ViewProps {
   navigationBarTranslucent?: boolean;
   navigationBarHidden?: boolean;
   nativeBackButtonDismissalEnabled?: boolean;
+  unstable_synchronousUpdatesEnabled?: WithDefault<boolean, false>;
```
kkafar: This is kind of ok, but I recommend not adding the `unstable` prefix for the props **on codegen spec** layer, since this also isn't a part of public API & not doing so, vastly simplifies method naming on native side.

src/flags.ts
```diff
@@ -33,10 +34,45 @@ export const compatibilityFlags = {
 const _featureFlags = {
   experiment: {
     controlledBottomTabs: RNS_CONTROLLED_BOTTOM_TABS_DEFAULT,
+    unstable_synchronousStateUpdatesEnabled:
```
kkafar:
```suggestion
    synchronousStateUpdatesEnabled:
```

kkafar: This is already a feature flag, no point of making experimental feature flags.

src/flags.ts
```diff
@@ -33,10 +34,45 @@ export const compatibilityFlags = {
...
+};
+
+const controlledBottomTabsAccessor = createExperimentalFeatureFlagAccessor(
+  'controlledBottomTabs',
+  RNS_CONTROLLED_BOTTOM_TABS_DEFAULT,
+);
+const synchronousUpdatesAccessor = createExperimentalFeatureFlagAccessor(
+  'unstable_synchronousStateUpdatesEnabled',
+  RNS_SYNCHRONOUS_STATE_UPDATES_DEFAULT,
+);
```
kkafar, line 80: I like this. Would love it in separate PR, but its ok.

kkafar (commented):
Few questions & suggestions. Please answer them. 

Thank you for the PR.

ios/RNSScreenStackHeaderConfig.mm
t0maboro: afaik no, the only sync updates issues I'm aware of occur on a screen level, I've decided to expose this prop for consistency, but if you prefer, we can skip that until we have any good reason to do that

apps/src/tests/TestSplitView/index.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3282/commits/badbd95435b898898b458373804f782a2934bb66

ios/RNSScreen.h
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3282/commits/0ffe03708d0e1c942d62ce58f6128e6e61730cd9

ios/RNSScreen.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3282/commits/67979fc6620c36e8a520ed30e7793b534ded085b

ios/RNSScreenStackHeaderConfig.h
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3282/commits/c4881f5b03729bfe85dbb77c6528351d47c35093

src/fabric/ModalScreenNativeComponent.ts
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3282/commits/c4881f5b03729bfe85dbb77c6528351d47c35093

src/flags.ts
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3282/commits/b1e3723630abf0dd5da9637ac7a193779c237e86

src/flags.ts
t0maboro: I can extract it to a separate PR if that's strong preference

ios/RNSScreenStackHeaderConfig.mm
kkafar: What do you think about having a granular control? A feature flag per component? My reasoning here is that ideally, we want to use synchronous updates only where necessary to avoid clogging the UI thread with renders & layout

src/flags.ts
kkafar: No, let's leave it be. I'm just hinting for the future.

t0maboro: ack

ios/RNSScreenStackHeaderConfig.mm
t0maboro: It makes sense to separate Screen from Header, but I'm wondering whether having separate flags for `HeaderConfig` and `HeaderSubview` makes sense, or rather, we should have 1 flag for handling the header

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3282/commits/d943ec9bd90525a090101cf2add3e69e960bf223 + https://github.com/software-mansion/react-native-screens/pull/3282/commits/1cb5ca954c528e524586337c86ae2ea90b7d135a

apps/src/tests/Test3282.tsx
```diff
@@ -0,0 +1,100 @@
...
+}
+
+const Stack = createNativeStackNavigator();
+
+const HomeScreen = ({ navigation }: RouteProps) => {
+  useEffect(() => {
+    featureFlags.experiment.synchronousScreenUpdatesEnabled = SYNCHRONOUS_STATE_UPDATES_ENABLED
+    featureFlags.experiment.synchronousHeaderConfigUpdatesEnabled = SYNCHRONOUS_STATE_UPDATES_ENABLED
+    featureFlags.experiment.synchronousHeaderSubviewUpdatesEnabled = SYNCHRONOUS_STATE_UPDATES_ENABLED
+  }, [])
```
kkafar: I haven't notice earlier, but we don't want to modify these during render. This is not a good example. These should be enabled / disabled in top level scope.

ios/RNSScreenStackHeaderConfig.mm
kkafar: Thanks!

apps/src/tests/Test3282.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3282/commits/1c09c040712f8aec28906b674a7790814841b1d9

kkafar (approved):

PR #3283, opened 2025-10-06
fix(iOS): Clear _shouldUpdateScrollEdgeEffects flag in RNSScreen.finalizeUpdates

kmichalikk:
## Description

#3212 introduced a bug when scrolling the view would trigger a scroll edge effect prop change. This happened because there was one place that didn't clear the flag that the prop was already set.

## Changes

The `_shouldUpdateScrollEdgeEffects` flag is now being cleared in RNSScreen.finalizeUpdates.

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/236e0317-b22d-45be-9513-081c57dadff8" /> | <video src="https://github.com/user-attachments/assets/753af5bd-7653-458e-8ca4-880ee728c468" /> |

Use Test3212.


kkafar (approved):
Thanks for catching this

PR #3284, opened 2025-10-06
test: Refactor Test3212, add option to set config from scroll view screen

kmichalikk:
## Description

This PR includes a small refactor for Test3212 that removes useCallback in favor of regular functions, and adds a button to change scrollEdgeEffect config right on the screen it is applied to - to have more paths to test.

For instance, you can set the scrollEdgeEffect config to 'hard' on the first Stack screen, push the second one with ScrollView, see the configuration being applied, then click the button and see the effect changed to 'hidden'.

### DIFF
--- a/apps/src/tests/Test3212/ScrollViewTemplate.tsx
+++ b/apps/src/tests/Test3212/ScrollViewTemplate.tsx
@@ -1,4 +1,5 @@
-import React, { Button, ScrollView, Text } from "react-native";
+import React from "react";
+import { Button, ScrollView, Text } from "react-native";
 import { useScrollEdgeEffectsConfigContext } from "./context";
 
 export function ScrollViewTemplate() {


apps/src/tests/Test3212/ScrollViewTemplate.tsx
```diff
@@ -1,10 +1,15 @@
-import React, { ScrollView, Text } from "react-native";
+import React, { Button, ScrollView, Text } from "react-native";
```
t0maboro: `React` should be imported from 'react'

kkafar: For sure it should

kmichalikk: https://github.com/software-mansion/react-native-screens/pull/3284/commits/eab9fb1810eb9c2bc5ed48fd7d6e3ca732dcc00d

PR #3285, opened 2025-10-06
fix(iOS, SplitView): Remove tracking presentation layer in favor of sync state update

t0maboro:
## Description

This code was originally added to mitigate issues with jumping content during transitions. By observing the presentation layer, we were able to send intermediate column size updates, which helped with making smoother animation.

However, with the current ability to send immediate updates, it seems that it now works well with the native SplitView transition. As a result, we can remove the previously added logic, which was initially introduced in: https://github.com/software-mansion/react-native-screens/pull/3073

Fixes https://github.com/software-mansion/react-native-screens-labs/issues/496

## Changes

- Removed logic related to observing presentation layer for sending asynchronous state updates.

## Screenshots / GIFs

### Before

https://github.com/user-attachments/assets/966d0b7f-665d-4add-947e-ceeddb81d0e2

### After

https://github.com/user-attachments/assets/69955aca-e9ae-48e2-a9fd-7d04ff65835f

I went through SplitView examples, having synchronous updates in SplitView and additionally sync updates for Stack v4 https://github.com/software-mansion/react-native-screens/pull/3282 . I focused on testing that there's no regression with pressables.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/gamma/split-view/RNSSplitViewScreenComponentView.mm
+++ b/ios/gamma/split-view/RNSSplitViewScreenComponentView.mm
@@ -167,24 +167,6 @@ namespace react = facebook::react;
   return react::concreteComponentDescriptorProvider<react::RNSSplitViewScreenComponentDescriptor>();
 }
 
-- (void)updateLayoutMetrics:(const facebook::react::LayoutMetrics &)layoutMetrics
-           oldLayoutMetrics:(const facebook::react::LayoutMetrics &)oldLayoutMetrics
-{
-  // We're tracking presentation layer updates in the RNSSplitViewScreen.
-  // There's a problem with SplitView that it sets the frame to the end value of the animation right after the animation
-  // begins. Because of that, the size of our component is desynchronizing easily and we're blocking a communication
-  // between native and shadow layout for a while until the transition ends. For the following case when we want to make
-  // a transition from width A to B:
-  // 1. size 'A' is set on ShadowNode
-  // 2. animation for the transition starts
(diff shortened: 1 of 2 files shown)


t0maboro:
It looks like the investigation above wasn't correct, because removing displayLink is causing a new issue with layouting texts and components with borders. Moving this PR to draft for further investigation, but it seems that we may need a completely different approach.

t0maboro:
After further investigation, I've concluded that, as for now, I'm not able to propose a better solution for SplitView. Synchronous updates bring our flow closer to the native implementation, significantly improving our main goal, which is achieving a truly native feeling, while `displayLink` is providing some delay in every single scenario (as we're updating the frame which is drawn or is about to be drawn and we have no guarantee that the Yoga layout will be computed fast enough to do not lose frames from the presentation layer).

That said, I'm aware that there are some side effects to the proposed approach, such as text layout behavior and limited support for non-uniform borders or Image components with `width: 100%` style in React Native. These issues seem unavoidable for now, as they stem from deviations from the native flow.

After performing the additional round of testing, I believe this is currently the most appropriate way to handle transitions. Therefore, I'm reopening this PR for further review.

t0maboro:
> Lovely. Awesome to get rid of the workaround code.
> 
> Can we also create a ticket to signalise the "non-uniform border" problem to react-native team? Ideally, we would prepare a repro to showcase this & open a issue on their repo.
> 
> ☝🏻 This obviously does not block the PR.

https://github.com/software-mansion/react-native-screens-labs/issues/515

PR #3286, opened 2025-10-06
feat: move `SafeAreaView` component from `/private` to `/experimental`

kkafar:
## Description

This PR moves `SafeAreaView` from `private` to `experimental` submodule. We decided to publish this component
despite earlier plans, as keeping it internal would lead to not-so-great API for the end user. To give an example, 
we'd have to render it above user content, causing a siutation, where user views are already displayed in safe area 
restricting options for background configuration (we'd have to expose `contentStyle` or `backgroundComponent` props to
mitigate this). We decided against such approach & release the component. The downstream solutions can decide whether
to handle this on their end or let the user use the component directly. 

#### Why this component is needed

TLDR: While `react-native-safe-area-context` provides really nice safe area view solution it is not enough for our use cases. 
Our component additionally treats native header, native tab bar, etc. as kind of insets, which fulfills our need 
of displaying content with respect to "real" safe area. 

## Changes

- **Move `SafeAreaView` & associated types from `/private` to `/experimental` submodule**
- **Import `SafeAreaView` from `/experimental` in apps**

Our example apps should build & run as before. Especially check out a test that uses the safe area view,
e.g. `TestSafeAreaViewIOS`.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/apps/src/tests/TestSafeAreaViewIOS/bottom-tabs/TestTab.tsx
+++ b/apps/src/tests/TestSafeAreaViewIOS/bottom-tabs/TestTab.tsx
@@ -1,6 +1,5 @@
-import React, { useMemo } from 'react';
+import { useMemo } from 'react';
 import { useBottomTabsSAVExampleContext } from './BottomTabsSAVExampleContext';
-import { SafeAreaView } from 'react-native-screens/private';
 import { mapContentStringToComponent } from '../shared';
 
 export default function TestTab() {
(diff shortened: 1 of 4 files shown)


src/private/types.ts
```diff
@@ -1 +1,3 @@
-export * from '../components/safe-area/SafeAreaView.types';
+// Just to keep this as a module. Remove this once
+// any other symbol is exported from this module again.
+export const __MODULE_KEEP__: number = 42;
```
t0maboro, line 3: 42 is a solid choice. I’d rethink suggesting its removal. It might upset the cosmic balance or Douglas Adams fans.

t0maboro (approved):

PR #3287, opened 2025-10-06
test(iOS, SplitView): Add example for testing performance

t0maboro:
## Description

Adding an example with some heavier components for testing whether the transitions are working smoothly.

## Changes

- Added example

## Screenshots / GIFs

https://github.com/user-attachments/assets/3255edc3-d666-4599-82a4-ceab666362d9

Tested changes with the new example and https://github.com/software-mansion/react-native-screens/pull/3285

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/apps/src/tests/TestSplitView/SplitViewPerfApp.tsx
+++ b/apps/src/tests/TestSplitView/SplitViewPerfApp.tsx
@@ -8,14 +8,20 @@ import {
   Image,
   Dimensions,
 } from 'react-native';
-import { SplitViewHost, SplitViewScreen } from 'react-native-screens/experimental';
+import {
+  SplitViewHost,
+  SplitViewScreen,
+} from 'react-native-screens/experimental';
 import { Colors } from '../../shared/styling/Colors';
 import { SplitViewBaseConfig } from './helpers/types';
 
 const { width } = Dimensions.get('window');
 
 const FlatListColumn = () => {
-  const data = useMemo(() => Array.from({ length: 500 }, (_, i) => `Item ${i + 1}`), []);
+  const data = useMemo(
+    () => Array.from({ length: 500 }, (_, i) => `Item ${i + 1}`),
+    [],
+  );
 
   return (
     <FlatList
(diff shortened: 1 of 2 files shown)


apps/src/tests/TestSplitView/SplitViewPerfApp.tsx
```diff
@@ -0,0 +1,145 @@
...
+import {
+  SplitViewHost,
+  SplitViewScreen,
+} from 'react-native-screens/experimental';
+import { Colors } from '../../shared/styling/Colors';
+import { SplitViewBaseConfig } from './helpers/types';
+
+const { width } = Dimensions.get('window');
+
+const FlatListColumn = () => {
```
kmichalikk, line 20: nit: I'd prefer regular functions instead of arrow functions in such top-level declarations

apps/src/tests/TestSplitView/SplitViewPerfApp.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const images = useMemo(
+    () =>
+      Array.from({ length: 30 }).map((_, index) => ({
+        id: `img-${index}`,
+        uri: `https://picsum.photos/seed/${index}/200`,
+      })),
+    [],
+  );
+
+  const numColumns = 2;
```
kmichalikk: This would look better as a prop with default value

kmichalikk (approved):

apps/src/tests/TestSplitView/SplitViewPerfApp.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3287/commits/48bb3ddf95174f66dd75d5cd190e58c0a0bb729b

apps/src/tests/TestSplitView/SplitViewPerfApp.tsx
t0maboro: I'd prefer to keep it as an arrow functions for consistency, because I'm using them everywhere in SplitView examples

PR #3288, opened 2025-10-07
feat(iOS, Tabs): add bottomAccessory support

kligarski:
## Description

Adds support for `botttomAccessory` in Bottom Tabs starting from iOS 26.

## Synchronization with ShadowTree

When bottom accessory transitions between `regular` and `inline` environments (when tab bar is minimized), we need to update the position and size of the bottom accessory. Our approach is different for RN < 0.82 and RN >= 0.82.

### Possible approaches

We considered following approaches:
1. asynchronous state updates & events
  - this would result in frame changing right after animation start to the final size
  - using `double-rendering` approach (explained further in PR description) wouldn't really improve the situation
2. (a)synchronous state updates & asynchronous events + DisplayLink
  - this allows us to track the native animation of the bottom accessory and match frame size
  - unfortunately, as DisplayLink allows us to read presentation layer frame, we will always be 1 frame delayed
  - with asynchronous updates, this delay increases to multiple frames; with synchronous updates this would be exactly 1 frame (but this doesn't mean it looks better!)
3. synchronous state updates & asynchronous events
  - synchronous state updates allow us to rely on `CoreAnimations` framework that animates views natively (details explained further in PR description)
  - unfortunately, this mechanism requires all changes to be performed synchronously - state update (size change) is handled synchronously thanks to `immediate` mode for state update introduced in RN 0.82 but "synchronous" event dispatch in RN 0.82 isn't exactly `synchronous` (update is performed on next loop beat, this is not enough for CA)
  - due to this limitation, any changes in reaction to environment change are buggy (see `Mounting/unmounting views during transition` section)
4. synchronous state updates & `double-rendering`
  - to mitigate problems described above, we render the component twice - first for `regular` environment, second for `inline` environment
  - we change which component is visible on environment change -> we can do this fully synchronously in native code
  - CA animates this change with cross-fade animation
  - this results seems to be a good compromise but requires React state synchronization between components, e.g. via context

For now, we decided to use:
- Paper: approach 2 (asynchronous)
- Fabric RN < 0.82: approach 2 (asynchronous)
- Fabric RN >= 0.82: approach 4

Those solutions are described in more detail below:

### Legacy architecture & New architecture prior to `react-native@0.82.0`

In versions prior to RN 0.82, we need use `DisplayLink` and presentation layer frames to get intermediate frames during the transition. This approach however has a major drawback - we are always at least one frame behind the current state as we're observing what is currently presented. When the difference in size/origin between frames is significant, you can see the content "jumping". In the case of bottom accessory, this is especially visible when using non-translucent background and transitioning from `inline` to `regular` environment (pay attention to the right edge of the accessory).

https://github.com/user-attachments/assets/3931c318-2bb7-4bc2-847f-95168578d7d2

Introduction of synchronous state updates in RN 0.82 (more details below) does not improve the situation when using this approach as we are still going to be at least one frame behind the animation.

### `react-native@0.82.0` and higher

Thanks to introduction of [synchronous state updates in RN 0.82](https://github.com/facebook/react-native/pull/52604), we can rely fully on native mechanisms for handling the transition. Bottom accessory only receives the final frame of the transition and thanks to synchronous state updates, we can immediately recalculate the layout in the Shadow Tree and update the Host Tree. This allows Core Animation framework to make the transition smooth. Details of how we think this works are available [here](https://github.com/software-mansion/react-native-screens/pull/3285#issue-3488360063).

Unfortunately, when using `react-native`, the situation is a little bit more complicated. 

#### Text

Text component behaves differently to the native platform. During the transition, it immediately adapts to the final frame size and then it is stretched. In bare UIKit app, the text adapt to new frame size at the end of the transition.

| `react-native` | `UIKit` |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/ea509fee-53bb-4300-a28d-3404525d47dd" /> | <video src="https://github.com/user-attachments/assets/17b2d3f8-5f04-406a-828c-504d040ad013" /> |

This requires more investigation and potentially changes in `react-native`.

#### Borders

`CoreAnimation` does not support non-uniform borders so `react-native` handles them in a custom way that does not seem to be compatible with the transition mechanism.

| non-uniform borders | uniform borders + **CA enabled** |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/e9ce53a9-bb4b-4465-a986-1a1f08791454" /> | <video src="https://github.com/user-attachments/assets/bcf029a4-764a-446c-82b6-91bfd42494c6" /> |

This requires more investigation and potentially changes in `react-native`.

#### Images

Similar problem (in a way it looks, not the exact mechanism of the bug) happens when using images e.g. with `width: 100%`.

https://github.com/user-attachments/assets/1aa0b77a-7800-471f-bd38-ac3eed506b87

This requires more investigation and potentially changes in `react-native`.

#### Mounting/unmounting views during transition

While state updates are performed synchronously, any changes to React Element Tree in reaction to environment change are handled asynchronously. We think that this is why the transition handled by `CoreAnimation` breaks when trying to mount/unmount components on environment change.

https://github.com/user-attachments/assets/ef0730db-386a-46be-8fcf-e0def3ca4097

Here, we try to remove the note icon. Unfortunately, the rest of the layout does not adapt. You can also observe the text stretching as mentioned 2 sections above.

In order to mitigate this issue, we use `double-rendering` approach, which has been described in `Possible approaches` section.

https://github.com/user-attachments/assets/7d105e62-1f69-4886-845a-464addfa608b

## Changes

- add `BottomTabsAccessory` JS component and use it in `BottomTabs`,
- add `BottomTabsAccessoryComponentView`, `BottomTabsAccessoryEventEmitter`, `BottomTabsAccessoryComponentViewManager`,
- add `BottomAccessoryHelper` to handle size and environment changes,
- add `BottomTabsAccessoryShadowStateProxy` to synchronize state between Host and ShadowTree,
- adapt `BottomTabsHost` to accept 2 types of children (`Screen` and `Accessory`),
- add test screen.

Run `Test3288`.

## Checklist

- [x] Updated TS types
- [ ] Updated documentation: 
- [x] Ensured that CI passes

### DIFF
--- a/ios/bottom-tabs/accessory/RNSBottomTabsAccessoryEventEmitter.mm
+++ b/ios/bottom-tabs/accessory/RNSBottomTabsAccessoryEventEmitter.mm
@@ -40,11 +40,9 @@ namespace react = facebook::react;
 {
 #if RCT_NEW_ARCH_ENABLED
   if (_reactEventEmitter != nullptr) {
-    react::RNSBottomTabsAccessoryEventEmitter::OnEnvironmentChangeEnvironment payloadEnvironment;
-    if (!rnscreens::conversion::RNSBottomTabsAccessoryOnEnvironmentChangePayloadFromUITabAccessoryEnvironment(
-            &payloadEnvironment, environment)) {
-      return NO;
-    }
+    auto payloadEnvironment =
+        rnscreens::conversion::RNSBottomTabsAccessoryOnEnvironmentChangePayloadFromUITabAccessoryEnvironment(
+            environment);
 
     _reactEventEmitter->onEnvironmentChange({.environment = payloadEnvironment});
     return YES;
@@ -57,9 +55,6 @@ namespace react = facebook::react;
     NSString *environmentString =
         rnscreens::conversion::RNSBottomTabsAccessoryOnEnvironmentChangePayloadFromUITabAccessoryEnvironment(
             environment);
-    if (environmentString == nil) {
-      return NO;
-    }
 
     self.onEnvironmentChange(@{@"environment" : environmentString});
     return YES;
--- a/ios/conversion/RNSConversions-BottomTabs.mm
+++ b/ios/conversion/RNSConversions-BottomTabs.mm
@@ -401,26 +401,22 @@ RNSScrollEdgeEffect RNSBottomTabsScrollEdgeEffectFromBottomTabsScreenTopScrollEd
 
 #undef SWITCH_EDGE_EFFECT
 
-#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0) && !TARGET_OS_TV && !TARGET_OS_VISION
+#if RNS_BOTTOM_ACCESSORY_AVAILABLE
 
 #if RCT_NEW_ARCH_ENABLED
 API_AVAILABLE(ios(26.0))
-BOOL RNSBottomTabsAccessoryOnEnvironmentChangePayloadFromUITabAccessoryEnvironment(
-    react::RNSBottomTabsAccessoryEventEmitter::OnEnvironmentChangeEnvironment *payloadEnvironment,
-    UITabAccessoryEnvironment environment)
+react::RNSBottomTabsAccessoryEventEmitter::OnEnvironmentChangeEnvironment
+RNSBottomTabsAccessoryOnEnvironmentChangePayloadFromUITabAccessoryEnvironment(UITabAccessoryEnvironment environment)
 {
   switch (environment) {
     case UITabAccessoryEnvironmentRegular:
-      *payloadEnvironment = react::RNSBottomTabsAccessoryEventEmitter::OnEnvironmentChangeEnvironment::Regular;
-      break;
+      return react::RNSBottomTabsAccessoryEventEmitter::OnEnvironmentChangeEnvironment::Regular;
     case UITabAccessoryEnvironmentInline:
-      *payloadEnvironment = react::RNSBottomTabsAccessoryEventEmitter::OnEnvironmentChangeEnvironment::Inline;
-      break;
+      return react::RNSBottomTabsAccessoryEventEmitter::OnEnvironmentChangeEnvironment::Inline;
     default:
-      return NO;
+      RCTLogError(@"[RNScreens] Unsupported environment for onEnvironmentChange event");
+      return react::RNSBottomTabsAccessoryEventEmitter::OnEnvironmentChangeEnvironment::Regular;
   }
-
-  return YES;
 }
 
 #if REACT_NATIVE_VERSION_MINOR >= 82
@@ -459,6 +455,7 @@ NSString *RNSBottomTabsAccessoryOnEnvironmentChangePayloadFromUITabAccessoryEnvi
       environmentString = BOTTOM_TABS_ACCESSORY_INLINE_ENVIRONMENT;
       break;
     default:
+      RCTLogError(@"[RNScreens] Unsupported environment for onEnvironmentChange event");
       break;
   }
 
@@ -467,7 +464,7 @@ NSString *RNSBottomTabsAccessoryOnEnvironmentChangePayloadFromUITabAccessoryEnvi
 
 #endif // RCT_NEW_ARCH_ENABLED
 
-#endif // RNS_IPHONE_OS_VERSION_AVAILABLE(26_0) && !TARGET_OS_TV && !TARGET_OS_VISION
+#endif // RNS_BOTTOM_ACCESSORY_AVAILABLE
 
 UIUserInterfaceStyle UIUserInterfaceStyleFromBottomTabsScreenCppEquivalent(
     react::RNSBottomTabsScreenUserInterfaceStyle userInterfaceStyle)
--- a/ios/conversion/RNSConversions.h
+++ b/ios/conversion/RNSConversions.h
@@ -73,13 +73,12 @@ RNSScrollEdgeEffect RNSBottomTabsScrollEdgeEffectFromBottomTabsScreenRightScroll
 RNSScrollEdgeEffect RNSBottomTabsScrollEdgeEffectFromBottomTabsScreenTopScrollEdgeEffectCppEquivalent(
     react::RNSBottomTabsScreenTopScrollEdgeEffect edgeEffect);
 
-#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0) && !TARGET_OS_TV && !TARGET_OS_VISION
+#if RNS_BOTTOM_ACCESSORY_AVAILABLE
 
 #if RCT_NEW_ARCH_ENABLED
 API_AVAILABLE(ios(26.0))
-BOOL RNSBottomTabsAccessoryOnEnvironmentChangePayloadFromUITabAccessoryEnvironment(
-    react::RNSBottomTabsAccessoryEventEmitter::OnEnvironmentChangeEnvironment *payloadEnvironment,
-    UITabAccessoryEnvironment environment);
+react::RNSBottomTabsAccessoryEventEmitter::OnEnvironmentChangeEnvironment
+RNSBottomTabsAccessoryOnEnvironmentChangePayloadFromUITabAccessoryEnvironment(UITabAccessoryEnvironment environment);
 
 #if REACT_NATIVE_VERSION_MINOR >= 82
 RNSBottomTabsAccessoryEnvironment RNSBottomTabsAccessoryEnvironmentFromCppEquivalent(
@@ -91,7 +90,7 @@ NSString *_Nullable RNSBottomTabsAccessoryOnEnvironmentChangePayloadFromUITabAcc
     UITabAccessoryEnvironment environment);
 #endif // RCT_NEW_ARCH_ENABLED
 
-#endif // RNS_IPHONE_OS_VERSION_AVAILABLE(26_0) && !TARGET_OS_TV && !TARGET_OS_VISION
+#endif // RNS_BOTTOM_ACCESSORY_AVAILABLE
 
 UIUserInterfaceStyle UIUserInterfaceStyleFromBottomTabsScreenCppEquivalent(
     react::RNSBottomTabsScreenUserInterfaceStyle userInterfaceStyle);


common/cpp/react/renderer/components/rnscreens/RNSBottomTabsAccessoryComponentDescriptor.h
```diff
@@ -0,0 +1,34 @@
...
+
+class RNSBottomTabsAccessoryComponentDescriptor final
+    : public ConcreteComponentDescriptor<RNSBottomTabsAccessoryShadowNode> {
+ public:
+  using ConcreteComponentDescriptor::ConcreteComponentDescriptor;
+
+  void adopt(ShadowNode &shadowNode) const override {
+    react_native_assert(
+        dynamic_cast<RNSBottomTabsAccessoryShadowNode *>(&shadowNode));
+    auto &bottomTabsAccessoryShadowNode =
```
t0maboro, line 18: nit: might be considered as const

common/cpp/react/renderer/components/rnscreens/RNSBottomTabsAccessoryShadowNode.cpp
```diff
@@ -0,0 +1,18 @@
+#include "RNSBottomTabsAccessoryShadowNode.h"
+
+#include <yoga/Yoga.h>
+
+namespace facebook::react {
+
+using namespace yoga;
```
t0maboro: nit: I prefer using `yoga::` instead of via `using namespace` - yoga classes and other definitions have pretty common names, so it may lead to some conflicts

common/cpp/react/renderer/components/rnscreens/RNSBottomTabsAccessoryShadowNode.cpp
```diff
@@ -0,0 +1,18 @@
+#include "RNSBottomTabsAccessoryShadowNode.h"
+
+#include <yoga/Yoga.h>
```
t0maboro: not used

common/cpp/react/renderer/components/rnscreens/RNSBottomTabsAccessoryShadowNode.cpp
t0maboro: btw it's not used

common/cpp/react/renderer/components/rnscreens/RNSBottomTabsAccessoryState.h
```diff
@@ -0,0 +1,17 @@
...
+class JSI_EXPORT RNSBottomTabsAccessoryState final {
+ public:
+  using Shared = std::shared_ptr<const RNSBottomTabsAccessoryState>;
+
+  RNSBottomTabsAccessoryState() {};
+  RNSBottomTabsAccessoryState(Size frameSize_, Point offset_)
+      : frameSize(frameSize_), offset(offset_) {};
+
+  const Size frameSize{};
+  const Point offset{};
```
t0maboro: seems like transitive dependencies, can we import these classes explicitly?

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryComponentView.mm
```diff
@@ -0,0 +1,195 @@
...
+RNS_IGNORE_SUPER_CALL_END
+
+#pragma mark - UIKit callbacks
+
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0) && !TARGET_OS_TV && !TARGET_OS_VISION
+
+- (void)didMoveToWindow
+{
+  if (self.window != nil) {
+    [_helper registerForAccessoryFrameChanges];
```
t0maboro, line 75: Do we need some cleanup counterpart for this method?

ios/bottom-tabs/RNSBottomTabsAccessoryComponentView.h
```diff
@@ -0,0 +1,72 @@
...
+ * If not null, the bottom accesory's helper that handles synchronization with ShadowNode.
+ */
+@property (nonatomic, strong, readonly, nullable) RNSBottomAccessoryHelper *helper;
+
+#endif // RNS_IPHONE_OS_VERSION_AVAILABLE(26_0) && !TARGET_OS_TV && !TARGET_OS_VISION
+
+/**
+ * If not null, the bottom tabs host view that this accessory component view belongs to.
+ */
+@property (nonatomic, weak, nullable) RNSBottomTabsHostComponentView *reactSuperview;
```
t0maboro: maybe reactSuperview -> bottomTabsHostView ?

ios/bottom-tabs/RNSBottomTabsAccessoryComponentView.mm
```diff
@@ -0,0 +1,195 @@
...
+
+#endif // !RCT_NEW_ARCH_ENABLED
+
+- (void)initState
+{
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0) && !TARGET_OS_TV && !TARGET_OS_VISION
+  if (@available(iOS 26, *)) {
+    _helper = [[RNSBottomAccessoryHelper alloc] initWithBottomAccessoryView:self];
+  }
+  _reactEventEmitter = [RNSBottomTabsAccessoryEventEmitter new];
```
t0maboro: why is this under the compile-time check, but not under runtime check?

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryComponentView.mm
```diff
@@ -0,0 +1,195 @@
...
+  // We could consider enabling it someday though.
+  return NO;
+}
+
+#pragma mark - RNSViewControllerInvalidating
+
+- (void)invalidateController
+{
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0) && !TARGET_OS_TV && !TARGET_OS_VISION
+  [_helper invalidate];
```
t0maboro: should we check if helper is not `nil`?

ios/bottom-tabs/RNSBottomTabsAccessoryEventEmitter.mm
```diff
@@ -0,0 +1,84 @@
...
+    switch (environment) {
+      case UITabAccessoryEnvironmentRegular:
+        payloadEnvironment = react::RNSBottomTabsAccessoryEventEmitter::OnEnvironmentChangeEnvironment::Regular;
+        break;
+      case UITabAccessoryEnvironmentInline:
+        payloadEnvironment = react::RNSBottomTabsAccessoryEventEmitter::OnEnvironmentChangeEnvironment::Inline;
+        break;
+      default:
+        return NO;
+    }
```
t0maboro: I'm wondering whether this conversion can be extracted

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryEventEmitter.mm
```diff
@@ -0,0 +1,84 @@
...
+#if RCT_NEW_ARCH_ENABLED
+  std::shared_ptr<const react::RNSBottomTabsAccessoryEventEmitter> _reactEventEmitter;
+#endif // RCT_NEW_ARCH_ENABLED
+}
+
+- (instancetype)init
+{
+  if (self = [super init]) {
+#if RCT_NEW_ARCH_ENABLED
+    _reactEventEmitter = nullptr;
```
t0maboro, line 26: it will be nullptr by default, we don't need to initialize it explicitly

ios/bottom-tabs/RNSBottomTabsAccessoryEventEmitter.mm
```diff
@@ -0,0 +1,84 @@
...
+  } else {
+    RCTLogWarn(@"[RNScreens] Skipped OnEnvironmentChange event emission due to nullish emitter");
+    return NO;
+  }
+#else
+  if (self.onEnvironmentChange) {
+    NSString *environmentString;
+    switch (environment) {
+      case UITabAccessoryEnvironmentRegular:
+        environmentString = @"regular";
```
t0maboro: I'd consider defining these strings as static variables

ios/bottom-tabs/host/RNSBottomTabsHostComponentView.mm
```diff
@@ -149,13 +151,37 @@ - (void)updateContainer
 {
   NSMutableArray<RNSTabsScreenViewController *> *tabControllers =
       [[NSMutableArray alloc] initWithCapacity:_reactSubviews.count];
-  for (RNSBottomTabsScreenComponentView *childView in _reactSubviews) {
-    [tabControllers addObject:childView.controller];
+  RNSBottomTabsAccessoryComponentView *bottomAccessory = nil;
+  for (UIView *childView in _reactSubviews) {
+    if ([childView isKindOfClass:[RNSBottomTabsScreenComponentView class]]) {
+      RNSBottomTabsScreenComponentView *childScreen = static_cast<RNSBottomTabsScreenComponentView *>(childView);
```
t0maboro, line 171: nit: by obj-c convention, should we use static_cast or c-style cast here?

ios/bottom-tabs/RNSBottomTabsHostComponentView.mm
```diff
@@ -356,40 +396,55 @@ - (void)mountingTransactionDidMount:(const facebook::react::MountingTransaction
 
 #pragma mark - LEGACY RCTComponent protocol
 
-RNS_IGNORE_SUPER_CALL_BEGIN
 - (void)insertReactSubview:(UIView *)subview atIndex:(NSInteger)index
 {
+  [super insertReactSubview:subview atIndex:index];
+  BOOL isValidBottomAccessory = [subview isKindOfClass:[RNSBottomTabsAccessoryComponentView class]];
```
t0maboro: isValidBottomAccessory -> isBottomAccessory ?

ios/bottom-tabs/RNSBottomTabsHostComponentView.mm
```diff
@@ -356,40 +396,55 @@ - (void)mountingTransactionDidMount:(const facebook::react::MountingTransaction
...
 
-  [_reactSubviews insertObject:childScreen atIndex:index];
+  [_reactSubviews insertObject:subview atIndex:index];
 }
 
 - (void)removeReactSubview:(UIView *)subview
 {
+  [super removeReactSubview:subview];
+  BOOL isValidBottomAccessory = [subview isKindOfClass:[RNSBottomTabsAccessoryComponentView class]];
+  BOOL isTabsScreen = [subview isKindOfClass:[RNSBottomTabsScreenComponentView class]];
```
t0maboro: this is repeated across some functions, I'd consider extracting it

src/components/bottom-tabs/BottomTabsAccessory.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * EXPERIMENTAL API, MIGHT CHANGE W/O ANY NOTICE
+ */
+export default function BottomTabsAccessory(props: BottomTabsAccessoryProps) {
+  return (
+    <BottomTabsAccessoryNativeComponent
+      {...props}
+      collapsable={false}
+      style={[props.style, styles.container]}>
```
t0maboro: shouldn't this be reversed? container might override some user-defined styles

src/components/bottom-tabs/BottomTabsAccessory.tsx
```diff
@@ -0,0 +1,25 @@
...
+      {props.children}
```
t0maboro: shorter version <BottomTabsAccessoryNativeComponent ... /> should work well

aledebla03:
Is there any ETA? Looks fantastic 🔥

kligarski:
I'm glad you like it. Recently I had to switch my focus to investigate some higher priority issues related to iOS 26 support. Once I finish, I'll come back to bottom accessory.

kkafar (commented):
Thank you for very nice PR description. It's really helpful during review process. Keep it up! 

Here I address the problems mentioned in PR description. I'll do a follow-up review regarding the code. 

---

The "text" behaviour you describe is fine. Having it at the beginning / end is not that much of a difference. Text animations are expensive, probably one of the reasons this is done this way in UIKit + the accessory view *does not seem like a place for pages of text.

---

The "borders problem" is the same we have with sync updates in split view. @t0maboro already created ticket for this. We should start conversation with react-native team how we should approach this problem. 

---

The "images" problem is interesting. It would be worth to investigate it a bit, initially in direction: who is responsible for image frame change. Is that Yoga or platform framework? Who interpolates it? 

---

The "Mounting/unmounting views during transition" problem

Yeah, it looks ugly. But it should not be like that. At least since 0.82. There are mechanisms in RN to emit sync events to JS now. Take a look [here](https://github.com/facebook/react-native/blob/d9262c60f4c02d66417008970dc9c34b742aaa75/packages/react-native/React/Fabric/Mounting/ComponentViews/VirtualView/RCTVirtualViewComponentView.mm#L338-L375) for usage examples. We should try that out. We can schedule a call here. Using this we should be able to trigger sync **react** render & have it hopefully nicely animated.

common/cpp/react/renderer/components/rnscreens/RNSBottomTabsAccessoryComponentDescriptor.h
kkafar: We call `setSize` later on it. Should stay non-const

common/cpp/react/renderer/components/rnscreens/RNSBottomTabsAccessoryShadowNode.cpp
kkafar: +1

common/cpp/react/renderer/components/rnscreens/RNSBottomTabsAccessoryShadowNode.cpp
kkafar: +1

common/cpp/react/renderer/components/rnscreens/RNSBottomTabsAccessoryState.h
kkafar: +1

kkafar: Also use the same naming as in e.g. `RNSScreeState` -> `contentOffset`

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.mm
```diff
@@ -0,0 +1,194 @@
...
+}
+
+- (void)handleDisplayLink:(CADisplayLink *)sender
+{
+  // We use self.nativeWrapperView because it has both the size and the origin
+  // that we want to send to the ShadowNode.
+  CGRect presentationFrame = self.nativeWrapperView.layer.presentationLayer.frame;
+  if (CGRectEqualToRect(presentationFrame, CGRectZero)) {
+    return;
+  }
```
kkafar, line 170: When can it be CGRectZero? Only before the very first layout or do you handle some other cases here?

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryComponentView.h
```diff
@@ -0,0 +1,72 @@
...
+
+#endif // RNS_IPHONE_OS_VERSION_AVAILABLE(26_0) && !TARGET_OS_TV && !TARGET_OS_VISION
+
+@interface RNSBottomTabsAccessoryComponentView : RNSReactBaseView <
+#ifdef RCT_NEW_ARCH_ENABLED
+                                                     RNSViewControllerInvalidating
+#else
+                                                     RCTInvalidating
+#endif
+                                                     >
```
kkafar, line 31: This should be available since 0.82: https://github.com/facebook/react-native/commit/b1994dd547395dbeec868e56762b5de2202cb1a9

Will we still need the `RNSViewControllerInvalidating`?

ios/bottom-tabs/RNSBottomTabsAccessoryComponentView.h
kkafar: +1 here, unless you just want to intentionally override this field

ios/bottom-tabs/RNSBottomTabsAccessoryComponentView.mm
kkafar: +1

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryComponentView.mm
kkafar: I think the invalidation will be done in `[self invalidate]`. Which is a bit hidden, but it's ok.

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryComponentView.mm
```diff
@@ -0,0 +1,195 @@
...
+#pragma mark - RCTInvalidating
+
+- (void)invalidate
+{
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0) && !TARGET_OS_TV && !TARGET_OS_VISION
+  [_helper invalidate];
+  _helper = nil;
+#else // RNS_IPHONE_OS_VERSION_AVAILABLE(26_0) && !TARGET_OS_TV && !TARGET_OS_VISION
+  return;
+#endif // RNS_IPHONE_OS_VERSION_AVAILABLE(26_0) && !TARGET_OS_TV && !TARGET_OS_VISION
```
kkafar: Please note that since https://github.com/facebook/react-native/commit/b1994dd547395dbeec868e56762b5de2202cb1a9 this will be also called on new architecture. You might wanna reuse that instead of RNSViewControllerInvalidating

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryEventEmitter.mm
kkafar: I'll sing in opposition here. I prefer things initialised explicitly, as it lefts hints that it's intentional & not something that just has been overlooked.

ios/bottom-tabs/RNSBottomTabsAccessoryEventEmitter.mm
kkafar: +1

ios/bottom-tabs/RNSBottomTabsAccessoryEventEmitter.h
```diff
@@ -0,0 +1,50 @@
...
+
+#if !RCT_NEW_ARCH_ENABLED
+#import <React/RCTComponent.h>
+#endif // !RCT_NEW_ARCH_ENABLED
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface RNSBottomTabsAccessoryEventEmitter : NSObject
+
+- (BOOL)emitOnEnvironmentChangeIfNecessary:(UITabAccessoryEnvironment)environment API_AVAILABLE(ios(26.0));
```
kkafar: Nit: we usually use `ifNeeded` suffix

ios/bottom-tabs/host/RNSBottomTabsHostComponentView.mm
kkafar: We're in Obj-C++ & I say we prefer C++ type casting. While `static_cast` is not much better / safer than regular C-style cast, besides it being more readable & visible I just want us to stick to a convention. We use `dynamic_cast` / `const_cast` if few places in the codebase.

ios/bottom-tabs/RNSBottomTabsHostComponentView.mm
```diff
@@ -212,29 +238,43 @@ - (BOOL)emitOnNativeFocusChangeRequestSelectedTabScreen:(RNSBottomTabsScreenComp
 
 - (void)mountChildComponentView:(UIView<RCTComponentViewProtocol> *)childComponentView index:(NSInteger)index
 {
+  BOOL isValidBottomAccessory = [childComponentView isKindOfClass:[RNSBottomTabsAccessoryComponentView class]];
```
kkafar:
```suggestion
  BOOL isBottomAccessory = [childComponentView isKindOfClass:[RNSBottomTabsAccessoryComponentView class]];
```

what is the "valid" part here for?

ios/bottom-tabs/RNSTabBarController.mm
```diff
@@ -41,8 +41,23 @@ - (void)tabBar:(UITabBar *)tabBar didSelectItem:(UITabBarItem *)item
...
+      if (_tabScreenControllers[i] != reactChildControllers[i]) {
+        changed = YES;
+        break;
+      }
+    }
+
+    if (!changed) {
+      return;
+    }
+  }
```
kkafar: Why do it in such way? Can't we just have a flag inside mount / unmount methods & call this method only if necessary?

src/components/bottom-tabs/BottomTabs.types.ts
```diff
@@ -1,9 +1,15 @@
+import { ReactNode } from 'react';
 import type {
   ColorValue,
   TextStyle,
   NativeSyntheticEvent,
   ViewProps,
 } from 'react-native';
+import { BottomTabsAccessoryEnvironment } from './BottomTabsAccessory.types';
```
kkafar:
```suggestion
import type { BottomTabsAccessoryEnvironment } from './BottomTabsAccessory.types';
```

ios/bottom-tabs/RNSTabBarController.mm
kkafar: What do you think?

common/cpp/react/renderer/components/rnscreens/RNSBottomTabsAccessoryState.h
kligarski: This is the case also in other classes, e.g. `RNSSplitViewScreenState`. I created a ticket to handle this in other places: https://github.com/software-mansion/react-native-screens-labs/issues/530. For bottom accessory, I'll fix it in this PR.

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryComponentView.mm
kligarski: In obj-c, *sending messages* to `nil` is just a no-op which is my intention here. If you think that wrapping this in an `if` make this more readable, I can do so.

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.mm
kligarski: Only before the first layout.

common/cpp/react/renderer/components/rnscreens/RNSBottomTabsAccessoryShadowNode.cpp
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/44f4c5fb6ae06254ac7f066ba7c624808893269a

common/cpp/react/renderer/components/rnscreens/RNSBottomTabsAccessoryShadowNode.cpp
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/44f4c5fb6ae06254ac7f066ba7c624808893269a

common/cpp/react/renderer/components/rnscreens/RNSBottomTabsAccessoryState.h
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/44f4c5fb6ae06254ac7f066ba7c624808893269a

ios/bottom-tabs/RNSBottomTabsAccessoryComponentView.h
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/44f4c5fb6ae06254ac7f066ba7c624808893269a

ios/bottom-tabs/RNSBottomTabsAccessoryComponentView.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/44f4c5fb6ae06254ac7f066ba7c624808893269a

ios/bottom-tabs/RNSBottomTabsAccessoryEventEmitter.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/2c69f3e61367cde61ddbbb025e8b1edc950a6a43

ios/bottom-tabs/RNSBottomTabsAccessoryEventEmitter.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/2c69f3e61367cde61ddbbb025e8b1edc950a6a43

ios/bottom-tabs/RNSBottomTabsHostComponentView.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/1696be40e78f204ff5110ecb2095d9209b0f2cbc

ios/bottom-tabs/RNSBottomTabsHostComponentView.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/1696be40e78f204ff5110ecb2095d9209b0f2cbc

ios/bottom-tabs/RNSBottomTabsHostComponentView.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/1696be40e78f204ff5110ecb2095d9209b0f2cbc

src/components/bottom-tabs/BottomTabsAccessory.tsx
kligarski: Decided to remove `flex` as it is not needed. We want to enforce `position: absolute` because that's how we're synchronizing the native position of the bottom accessory with Yoga/JS.

https://github.com/software-mansion/react-native-screens/pull/3288/commits/5f83151d62ce3511928f82a715bf6459c4a673e8

src/components/bottom-tabs/BottomTabsAccessory.tsx
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/2f5d7c33620a3c0c2df5444c4577ea2dcbcdbefe

common/cpp/react/renderer/components/rnscreens/RNSBottomTabsAccessoryState.h
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/e98390a3ed8cfefdad0acd08f4f03091a8bd51ff

ios/bottom-tabs/RNSBottomTabsAccessoryEventEmitter.h
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/946b54685a86d7dee5ab0a4308195b9b7e79c928

src/components/bottom-tabs/BottomTabs.types.ts
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/cdb7d840d8af57377174e8b047467189d49e9ead

ios/bottom-tabs/RNSTabBarController.mm
kligarski: Sounds good.

https://github.com/software-mansion/react-native-screens/pull/3288/commits/65524d8b997d7f3f2f1a650938975820b48ea49a

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryComponentView.mm
t0maboro: with the confidence that's intended, I'm okay to leave it as it is

src/components/bottom-tabs/BottomTabsAccessory.tsx
t0maboro: ok, but still I think that user-defined styles should be put last

common/cpp/react/renderer/components/rnscreens/RNSBottomTabsAccessoryState.h
t0maboro: > This is the case also in other classes, e.g. `RNSSplitViewScreenState`. I created a ticket to handle this in other places: [software-mansion/react-native-screens-labs#530](https://github.com/software-mansion/react-native-screens-labs/issues/530). For bottom accessory, I'll fix it in this PR.

nice catch, thanks

kkafar:
Bottom accessory view + view mount / unmount during transition between environments. 

| patched "sync" events | "sync" events (current state) |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/21b8cb26-ceb8-44ee-a350-c0e5f268c5ce" alt="patched sync events" /> | <video src="https://github.com/user-attachments/assets/ef0730db-386a-46be-8fcf-e0def3ca4097" alt="sync events (current state)" />

src/components/bottom-tabs/BottomTabsAccessory.tsx
kligarski: As discussed internally, we'll leave this as is.

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryComponentView.h
kligarski: We won't but as discussed [here](https://github.com/software-mansion/react-native-screens/pull/3367#pullrequestreview-3415227737), we want to wait until 0.82 becomes minimum RN version in order not to introduce multiple solutions for the same problem at the same time.

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryComponentView.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288#discussion_r2460578537

common/cpp/react/renderer/components/rnscreens/RNSBottomTabsAccessoryComponentDescriptor.h
```diff
@@ -0,0 +1,34 @@
...
+        static_cast<RNSBottomTabsAccessoryShadowNode &>(shadowNode);
+
+    auto state = std::static_pointer_cast<
+        const RNSBottomTabsAccessoryShadowNode::ConcreteState>(
+        shadowNode.getState());
+    auto stateData = state->getData();
+
+    if (stateData.frameSize.width != 0 && stateData.frameSize.height != 0) {
```
kkafar, line 26: We need refactor across whole codebase that will extract this particular condition to a separate function. We have too many places where this is stated explicitly & it is hard to change definition of what "empty state" means. Let's create ticket for this, please.

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.mm
```diff
@@ -0,0 +1,254 @@
...
+#if RCT_NEW_ARCH_ENABLED
+#import <rnscreens/RNSBottomTabsAccessoryShadowNode.h>
+#else // RCT_NEW_ARCH_ENABLED
+#import <React/RCTUIManager.h>
+#endif // RCT_NEW_ARCH_ENABLED
+
+namespace react = facebook::react;
+
+@implementation RNSBottomAccessoryHelper {
+  RNSBottomTabsAccessoryComponentView *_bottomAccessoryView;
```
kkafar: Will this create retain cycle? If so, let's document it here & state how to break it.

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.mm
```diff
@@ -0,0 +1,254 @@
...
+#pragma mark - Observing environment changes
+
+- (id<UITraitChangeRegistration>)registerForAccessoryEnvironmentChanges
+{
+  return [_bottomAccessoryView
+      registerForTraitChanges:@[ [UITraitTabAccessoryEnvironment class] ]
+                  withHandler:^(__kindof id<UITraitEnvironment>, UITraitCollection *previousTraitCollection) {
+                    UITabAccessoryEnvironment environment =
+                        self->_bottomAccessoryView.traitCollection.tabAccessoryEnvironment;
+                    [self->_bottomAccessoryView.reactEventEmitter emitOnEnvironmentChangeIfNeeded:environment];
```
kkafar, line 104: Do you recall whether this callback is invoked only when the trait has really changed value? Or is it called also if some one just wrote to the value? My question basically is whether we should diff with previousTraitCollection inside the handler or is this done for us?

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.mm
```diff
@@ -0,0 +1,254 @@
...
+
+- (UIView *)nativeWrapperView
+{
+  RCTAssert(
+      _bottomAccessoryView.superview.superview != nil,
+      @"[RNScreens] RNSBottomTabsAccessoryComponentView must be the set as bottom accessory.");
+  return _bottomAccessoryView.superview.superview;
+}
+
+- (void)notifyFrameUpdate
```
kkafar: I'd like the naming of this method to be more specific. Something like `nativeWrapperViewChangedFrame` or something of this kind.

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.mm
```diff
@@ -0,0 +1,254 @@
...
+{
+#if !RCT_NEW_ARCH_ENABLED || REACT_NATIVE_VERSION_MINOR < 82
+  // Make sure that bottom accessory's size is sent to ShadowNode as soon as possible.
+  // We set origin to (0,0) because initially self.nativeWrapperView's origin is incorrect.
+  // We want the enable the display link as well so that it takes over later with correct origin.
+  if (!_initialStateUpdateSent) {
+    CGRect frame = CGRectMake(0, 0, self.nativeWrapperView.frame.size.width, self.nativeWrapperView.frame.size.height);
+    [self updateShadowStateWithFrame:frame];
```
kkafar: Let's go with SplitView example & create a shadow state proxy for handling all shadow state updates. 

Also the proxy should be retained by the view & helper should access it via the view.

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.mm
```diff
@@ -0,0 +1,254 @@
...
+#endif // RCT_NEW_ARCH_ENABLED
+
+namespace react = facebook::react;
+
+@implementation RNSBottomAccessoryHelper {
+  RNSBottomTabsAccessoryComponentView *_bottomAccessoryView;
+
+#if RCT_NEW_ARCH_ENABLED
+  react::RNSBottomTabsAccessoryShadowNode::ConcreteState::Shared _state;
+#endif // RCT_NEW_ARCH_ENABLED
```
kkafar: Why do we hold pointer to the state of the component here? Can not we just access it through the component itself?

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.mm
```diff
@@ -0,0 +1,254 @@
...
+{
+  RCTAssert(
+      _bottomAccessoryView.superview.superview != nil,
+      @"[RNScreens] RNSBottomTabsAccessoryComponentView must be the set as bottom accessory.");
+  return _bottomAccessoryView.superview.superview;
+}
+
+- (void)notifyFrameUpdate
+{
+#if !RCT_NEW_ARCH_ENABLED || REACT_NATIVE_VERSION_MINOR < 82
```
kkafar: TBH I think that this condition occurs so many times, that you could consider creating *file scoped* macro for this.

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryComponentView.h
```diff
@@ -0,0 +1,72 @@
...
+
+@property (nonatomic, weak, readonly, nullable) RCTBridge *bridge;
+#endif // !RCT_NEW_ARCH_ENABLED
+
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0) && !TARGET_OS_TV && !TARGET_OS_VISION
+
+/**
+ * If not null, the bottom accesory's helper that handles synchronization with ShadowNode.
+ */
+@property (nonatomic, strong, readonly, nullable) RNSBottomAccessoryHelper *helper;
```
kkafar, line 45: Yep, it definitely looks like we have a retain cycle here. Is it necessary?

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryComponentView.mm
```diff
@@ -0,0 +1,188 @@
...
+{
+  return react::concreteComponentDescriptorProvider<react::RNSBottomTabsAccessoryComponentDescriptor>();
+}
+
+#endif // RNS_IPHONE_OS_VERSION_AVAILABLE(26_0) && !TARGET_OS_TV && !TARGET_OS_VISION
+
++ (BOOL)shouldBeRecycled
+{
+  // There won't be tens of instances of this component usually & it's easier for now.
+  // We could consider enabling it someday though.
```
kkafar, line 118: I like this being copied from component to component 😄 Law of precedence :D

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryComponentViewManager.h
```diff
@@ -0,0 +1,11 @@
...
+
+#import <React/RCTViewManager.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface RNSBottomTabsAccessoryComponentViewManager : RCTViewManager
+
+@end
+
+NS_ASSUME_NONNULL_END
```
kkafar, line 13: I believe this can be old arch only. IIRC the view manager is not required at all in new architecture. I know that some of previous component define it, but I think that we just weren't aware of this fact. Can we try wrapping whole file in old arch check?

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryComponentViewManager.mm
kkafar, line 1: Obviously same here. The implementation could be old arch only.

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryContentComponentView.h
```diff
@@ -0,0 +1,24 @@
...
+#endif // defined(__cplusplus)
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface RNSBottomTabsAccessoryContentComponentView : RNSReactBaseView
+
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0) && !TARGET_OS_TV && !TARGET_OS_VISION && RCT_NEW_ARCH_ENABLED && \
+    defined(__cplusplus) && REACT_NATIVE_VERSION_MINOR >= 82
+
+@property (nonatomic) RNSBottomTabsAccessoryEnvironment environment;
```
kkafar: Why is `defined(__cplusplus)` here?

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryContentComponentView.mm
```diff
@@ -0,0 +1,105 @@
...
+        rnscreens::conversion::RNSBottomTabsAccessoryEnvironmentFromCppEquivalent(newComponentProps.environment);
+  }
+
+  [super updateProps:props oldProps:oldProps];
+}
+
+- (void)finalizeUpdates:(RNComponentViewUpdateMask)updateMask
+{
+  [super finalizeUpdates:updateMask];
+  [_accessoryView.helper handleContentViewVisibilityForEnvironmentIfNeeded];
```
kkafar, line 78: Why is this called here? What case does it cover?

ios/bottom-tabs/host/RNSBottomTabsHostComponentView.mm
```diff
@@ -147,15 +151,47 @@ - (void)reactAddControllerToClosestParent:(UIViewController *)controller
...
+      [tabControllers addObject:childScreen.controller];
+    } else if ([childView isKindOfClass:[RNSBottomTabsAccessoryComponentView class]]) {
+      RCTAssert(bottomAccessory == nil, @"[RNScreens] There can only be one child RNSBottomTabsAccessoryComponentView");
+      bottomAccessory = static_cast<RNSBottomTabsAccessoryComponentView *>(childView);
+    }
```
kkafar, line 180: Add an `else` branch here & add assertion there.

ios/bottom-tabs/host/RNSBottomTabsHostComponentView.mm
```diff
@@ -357,7 +378,7 @@ - (void)mountingTransactionWillMount:(const facebook::react::MountingTransaction
 - (void)mountingTransactionDidMount:(const facebook::react::MountingTransaction &)transaction
                withSurfaceTelemetry:(const facebook::react::SurfaceTelemetry &)surfaceTelemetry
 {
-  if (_hasModifiedReactSubviewsInCurrentTransaction) {
+  if (_hasModifiedTabsScreensInCurrentTransaction || _hasModifiedBottomAccessoryInCurrentTransation) {
```
kkafar: Maybe worth extracting this to separate property? You could declare an extension on the `RNSBottomTabsHostComponentView` *in this source* file & add a readonly computed property.

ios/bottom-tabs/host/RNSBottomTabsHostComponentView.mm
```diff
@@ -368,40 +389,26 @@ - (void)mountingTransactionDidMount:(const facebook::react::MountingTransaction
...
-  RCTAssert(
-      [subview isKindOfClass:RNSBottomTabsScreenComponentView.class],
-      @"BottomTabsView only accepts children of type BottomTabScreen. Attempted to mount %@",
-      subview);
-
-  auto *childScreen = static_cast<RNSBottomTabsScreenComponentView *>(subview);
-  childScreen.reactSuperview = self;
-
-  [_reactSubviews insertObject:childScreen atIndex:index];
+  [super insertReactSubview:subview atIndex:index];
```
kkafar, line 404: Why call super here?

ios/bottom-tabs/host/RNSBottomTabsHostComponentView.mm
```diff
@@ -368,40 +389,26 @@ - (void)mountingTransactionDidMount:(const facebook::react::MountingTransaction
...
+  [self validateAndHandleReactSubview:subview didMount:YES];
+  [_reactSubviews insertObject:subview atIndex:index];
```
kkafar: Also why isn't this call to insert / remove object from `_reactSubview` handled inside the callback?

ios/bottom-tabs/host/RNSBottomTabsHostComponentView.mm
```diff
@@ -491,6 +499,32 @@ - (void)setOnNativeFocusChange:(RCTDirectEventBlock)onNativeFocusChange
 
 #endif // RCT_NEW_ARCH_ENABLED
 
+#pragma mark - Common
+
+- (void)validateAndHandleReactSubview:(UIView *)subview didMount:(BOOL)mount
```
kkafar:
```suggestion
- (void)validateAndHandleReactSubview:(UIView *)subview shouldMount:(BOOL)mount
```

src/components/bottom-tabs/BottomTabs.tsx
```diff
@@ -63,6 +71,26 @@ function BottomTabs(props: BottomTabsProps) {
...
       {filteredProps.children}
+      {bottomAccessory &&
+        Platform.OS === 'ios' &&
+        parseInt(Platform.Version, 10) >= 26 &&
+        (Platform.constants.reactNativeVersion.minor >= 82 ? (
+          <BottomTabsAccessory>
+            <BottomTabsAccessoryContent environment="regular">
+              {bottomAccessory('regular')}
+            </BottomTabsAccessoryContent>
+            <BottomTabsAccessoryContent environment="inline">
```
kkafar, line 82: Do we do this double-rendering trick only on >= 82?  whaat

src/components/bottom-tabs/BottomTabsAccessory.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * EXPERIMENTAL API, MIGHT CHANGE W/O ANY NOTICE
+ */
+export default function BottomTabsAccessory(props: BottomTabsAccessoryProps) {
+  return (
+    <BottomTabsAccessoryNativeComponent
+      {...props}
+      collapsable={false}
+      style={[props.style, styles.container]}
```
kkafar:
```suggestion
      style={[props.style, styles.absolutePositioned]}
```

kkafar: or `positionAbsolute` or align with predefined style sets from `StyleSheet.xxx` namespace.

src/components/bottom-tabs/BottomTabsAccessoryContent.tsx
```diff
@@ -0,0 +1,29 @@
...
+  );
+}
+
+const styles = StyleSheet.create({
+  container: {
+    position: 'absolute',
+    top: 0,
+    left: 0,
+    width: '100%',
+    height: '100%',
```
kkafar: Is this intentionally chosen over `right: 0, bottom: 0`?

kkafar (commented):
Really good job. The PR is in really good shape & code is of high quality - kudos. 

I've got few remarks, refactor requests & questions. Please answer them.

common/cpp/react/renderer/components/rnscreens/RNSBottomTabsAccessoryComponentDescriptor.h
kligarski: https://github.com/software-mansion/react-native-screens-labs/issues/542

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.mm
kligarski: I can't find any definitive answer in UIKit's docs but they mention that the handler is "A closure that the system executes when one of the registered traits **changes**". When testing, I did not receive any repeated notifications. Changing other traits (e.g. switching between light and dark appearance) does not invoke any notification for observed `UITraitTabAccessoryEnvironment`.

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryContentComponentView.h
kligarski: We can import `cxxreact/ReactNativeVersion.h` in header file only in C++ guards, otherwise the build fails.

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryContentComponentView.mm
kligarski: `RCTViewComponentView` changes view opacity via `view.layer.opacity`, e.g. in `updateProps` and `invalidateLayer` -> this overrides our setting. In order to bring back correct values, I update the visibility here.

Do you think we should do it in some other way? I also noticed now that when you enable devtools, the component breaks (both views are visible) and the only way to fix it is to remount the accessory. I thought about changing it in `_props` but I'm not sure here.

ios/bottom-tabs/host/RNSBottomTabsHostComponentView.mm
kligarski: Changes to Paper `insert/removeReactSubviews` and `didUpdateReactSubviews` are needed to handle removing bottom accessory (you can test it in test screen by setting "shown: false").

On Paper, when views are removed from view registry ([here](https://github.com/facebook/react-native/blob/v0.81.1/packages/react-native/React/Modules/RCTUIManager.mm#L967) ->  [here](https://github.com/facebook/react-native/blob/v0.81.1/packages/react-native/React/Modules/RCTUIManager.mm#L996) -> [here](https://github.com/facebook/react-native/blob/v0.81.1/packages/react-native/React/Modules/RCTUIManager.mm#L818)), `[container reactSubviews]` are used. That's why we need to keep Paper's `reactSubviews` in sync (not only our `_reactSubviews` that we use on Fabric and adapted it on Paper). If we don't do it, we won't receive `removeReactSubview` at all. Easiest way to handle it seems to be using super implementations of `insert/removeReactSubviews`.

After setting `reactSubviews`, I had to remove super call from `didUpdateReactSubviews` in order to prevent [`react-native` from inserting subviews](https://github.com/facebook/react-native/blob/v0.81.1/packages/react-native/React/Views/UIView%2BReact.m#L172) (we do it manually to handle bottom tab screens/accessory via appropriate UIKit APIs).

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.mm
kligarski: Decided to use weak pointer instead.

https://github.com/software-mansion/react-native-screens/pull/3288/commits/32414292125bc87db3e8a4604c85cc3a0a8a52ec

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/7d8bbafaf327a21020eb822e267e2756c6dc7c9f

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/0bc2cb81ab0bbc9a1895903e1d60ec6dde9dfe4e

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/48360ce44066128c5ca5c8ce83a4fb807908795c

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.mm
kligarski: Decided to just use `REACT_NATIVE_VERSION_MINOR < 82` instead of `!RCT_NEW_ARCH_ENABLED || REACT_NATIVE_VERSION_MINOR < 82` as they are equivalent (Paper is not supported on >=0.82).

https://github.com/software-mansion/react-native-screens/pull/3288/commits/f3660942376aff81e785dc5cac4c8470200a139c

I created file-scoped macros for OS version.

https://github.com/software-mansion/react-native-screens/pull/3288/commits/f13c54bd67c649f16e915b4b9be94cb9634034e5

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryComponentView.h
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288#discussion_r2510352696

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryComponentViewManager.h
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/0a96a35c5500c26f064375a300572310e5b2bf76

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryComponentViewManager.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/0a96a35c5500c26f064375a300572310e5b2bf76

ios/bottom-tabs/host/RNSBottomTabsHostComponentView.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/eb0dcf374fc0cb93be5d15e650207cce33c9ad1c

ios/bottom-tabs/host/RNSBottomTabsHostComponentView.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/c8f8d3d8a7f6302d55ec147cb6e0b96845dbc304

ios/bottom-tabs/host/RNSBottomTabsHostComponentView.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/cf4e24c506de6ec10030873df41e4cbbf08b7d44

ios/bottom-tabs/host/RNSBottomTabsHostComponentView.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/cc84e2a817605b9c06d286f9aa0ce2d5d8f089d3

src/components/bottom-tabs/BottomTabsAccessory.tsx
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/efba719e66f6fb738d618e047a5716c4a0711969

src/components/bottom-tabs/BottomTabsAccessoryContent.tsx
kligarski: No. d:

https://github.com/software-mansion/react-native-screens/pull/3288/commits/efba719e66f6fb738d618e047a5716c4a0711969

src/components/bottom-tabs/BottomTabs.tsx
kligarski: I updated the PR description to explain why this is the case.

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.mm
kkafar: Okay, that's fine then.

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryContentComponentView.h
kkafar: Didn't see that. Makes kina sense now. However this could be problematic. This property won't be visible from Swift.

Please create a ticket to make this work with Swift.

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.h
```diff
@@ -0,0 +1,72 @@
+#import <RNSDefines.h>
+
+#define BOTTOM_ACCESSORY_AVAILABLE RNS_IPHONE_OS_VERSION_AVAILABLE(26_0) && !TARGET_OS_TV && !TARGET_OS_VISION
```
kkafar: This macro is used in many places. You can extract it to separate header file. 

There is `RNSDefines.h` already I believe, you might place it there. If you want, you might be even more specific & create `RNSDefines-tabs.h` or something similar. Up to you, but I'ld like to extract it. 

I guess you did not prefix it with `RNS_` since it was local macro & you undef it later on, but in case of putting it in shared header please remember to apply the prefix.

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.h
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * Invalidates observers, display link (if it is used); resets internal properties.
+ */
+- (void)invalidate;
+
+@end
+
+#pragma mark - Content view switching workaround
+
+#if defined(__cplusplus) && REACT_NATIVE_VERSION_MINOR >= 82
```
kkafar, line 37: Applying `defined(__cplusplus)` generally excludes these symbols from being visible from Swift.

I'm not decided what we should do here. Most likely we can stick to the ObjectiveC for now & this won't be a problem for a long time... I'm willing to 🙈 and approve this, but we need to create ticket to refactor this condition. 

Maybe, in the future, we should use a technique similar to what's already done in [`RNSDefines`](https://github.com/software-mansion/react-native-screens/blob/b2a8e0121dc936d5f6d080348a2fd6009d021ead/ios/utils/RNSDefines.h#L13-L23), instead of using `cxx/ReactNativeVersion`? That's not really a scalable solution, though. Let's create ticket for this & think there.

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.mm
```diff
@@ -0,0 +1,213 @@
+#import "RNSBottomAccessoryHelper.h"
+#import "RNSBottomTabsAccessoryShadowStateProxy.h"
+
+#define BOTTOM_ACCESSORY_AVAILABLE RNS_IPHONE_OS_VERSION_AVAILABLE(26_0) && !TARGET_OS_TV && !TARGET_OS_VISION
+
+#if BOTTOM_ACCESSORY_AVAILABLE
+
+#import <React/RCTAssert.h>
+#include <cxxreact/ReactNativeVersion.h>
```
kkafar: Why all of a sudden `include` here, instead of `import`? Also the check for cplusplus is missing? Or am I missing something? Maybe its included somewhere transitively?

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryContentComponentView.mm
kkafar: > RCTViewComponentView changes view opacity via view.layer.opacity, e.g. in updateProps and invalidateLayer -> this overrides our setting. In order to bring back correct values, I update the visibility here.

Please document this in code. Because this is something you can not come up with by just reading the code. 

Also document / create a ticket for the bug with devtools you described. Let's not let it go under radar. 

---

That said, I can see that the `RCTViewComponentView` diffs the opacity before mutating it & it diffs it with prop values, not the value applied on the layer. 

Image

Does this problem occur only in first render then? Or how does it work otherwise?

ios/bottom-tabs/host/RNSBottomTabsHostComponentView.mm
```diff
@@ -147,15 +159,51 @@ - (void)reactAddControllerToClosestParent:(UIViewController *)controller
 
 - (void)updateContainer
 {
+  if (!_hasModifiedTabsScreensInCurrentTransaction && !_hasModifiedBottomAccessoryInCurrentTransation) {
```
kkafar: Let's use the newly added `hasModifiedReactSubviewsInCurrentTransaction` here & in every appropriate place.

ios/conversion/RNSConversions-BottomTabs.mm
```diff
@@ -400,6 +401,74 @@ RNSScrollEdgeEffect RNSBottomTabsScrollEdgeEffectFromBottomTabsScreenTopScrollEd
...
+      break;
+    case UITabAccessoryEnvironmentInline:
+      *payloadEnvironment = react::RNSBottomTabsAccessoryEventEmitter::OnEnvironmentChangeEnvironment::Inline;
+      break;
+    default:
+      return NO;
+  }
+
+  return YES;
+}
```
kkafar: I just noticed this. Why do you use raw pointers here instead of just returning a value & assigning it at the call site? Otherwise it'd be nice to handle "the null case".

kkafar (commented):
Looks nice. I've added few more suggestions & answered some questions.

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryContentComponentView.h
kligarski: https://github.com/software-mansion/react-native-screens-labs/issues/557

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.h
kligarski: https://github.com/software-mansion/react-native-screens-labs/issues/557

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryContentComponentView.mm
kligarski: `self.layer.opacity` is also set in `invalidateLayer` which runs on every update (e.g. state update).

image

Without calling `handleContentViewVisibilityForEnvironmentIfNeeded` in `finalizeUpdates`, the bug only happens on first render. It does not happen "later" because `invalidateLayer` is called in reaction to `updateState` and trait change comes just **after** it (and it calls `handleContentViewVisibilityForEnvironmentIfNeeded` as intended).

kligarski: Ticket for devtools bug: https://github.com/software-mansion/react-native-screens-labs/issues/558.

kligarski: Added a comment here: https://github.com/software-mansion/react-native-screens/pull/3288/commits/4929f34d372342d68e5533325a7eac548801358d.

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.mm
kligarski: Fixing `include` -> `import` here: https://github.com/software-mansion/react-native-screens/pull/3288/commits/b52165821043f020bb129278a82b388dc276ab24.

> Also the check for cplusplus is missing?

I don't think that they are necessary in `.mm` file, right?

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.h
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/c55db8dd226ee0745db81150c2cec200a0a3e471

ios/bottom-tabs/host/RNSBottomTabsHostComponentView.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/286378184ebe689dffb6df0622c8184e182cc04a

ios/conversion/RNSConversions-BottomTabs.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3288/commits/487ec07a33ac5b31eba4be39491ceb90cfbafc8c

ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.mm
kkafar: You're absolutely right here. +1

ios/bottom-tabs/accessory/RNSBottomTabsAccessoryContentComponentView.mm
kkafar: ok 👍🏻

JustJoostNL:
I am not entirely sure, but I think this PR introduced a build issue:

```
❌  (node_modules/react-native-screens/ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.h:3:9)

  1 | #pragma once
  2 | 
> 3 | #import <RNSDefines.h>
    |         ^ 'RNSDefines.h' file not found with <angled> include; use "quotes" instead
  4 | 
  5 | #if RNS_BOTTOM_ACCESSORY_AVAILABLE
  6 | 
```

kligarski:
> I am not entirely sure, but I think this PR introduced a build issue:
> 
> ```
> ❌  (node_modules/react-native-screens/ios/bottom-tabs/accessory/RNSBottomAccessoryHelper.h:3:9)
> 
>   1 | #pragma once
>   2 | 
> > 3 | #import <RNSDefines.h>
>     |         ^ 'RNSDefines.h' file not found with <angled> include; use "quotes" instead
>   4 | 
>   5 | #if RNS_BOTTOM_ACCESSORY_AVAILABLE
>   6 | 
> ```

Thanks for catching this. I created a PR with the fix: https://github.com/software-mansion/react-native-screens/pull/3439.

Issue #3273, opened 2025-10-02
[Tabs][iOS 26] Setting badge value on active tab, causes a glitch


Ubax:
### Description

When bagde value is updated on the active tab, then there is a glitch. This only happens on iOS 26.

https://github.com/user-attachments/assets/5c08a2ef-ba0e-4a6a-b587-13c40aa6ccdc

The same happens when icon is updated

### Steps to reproduce

Change the value of selected badge dynamically. For example

```
const [i, setI] = useState(0);

  useEffect(() => {
    const interval = setInterval(() => {
      setI(prev => prev + 1);
    }, 2000);
    return () => clearInterval(interval);
  }, []);

  TAB_CONFIGS[1].tabScreenProps.badgeValue = `${i}`;
```

### Snack or a link to a repository

FabricExample

### Screens version

4.16.0

### React Native version

0.82

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

Brma1048:
@Ubax how did you change the background color for the badge?

PR #3290, opened 2025-10-08
fix(iOS, Tabs): update tab bar item only when necessary

kligarski:
## Description

Limits changing tab bar item instances ([PR introducing this](https://github.com/software-mansion/react-native-screens/pull/3164)) to only necessary cases (changing `systemItem`). Unnecessary updates caused a bug when changing e.g. only `badgeValue`. 

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/d99e1120-8311-48d5-8d69-b7079c328ddd" /> | <video src="https://github.com/user-attachments/assets/444a60aa-e576-42e3-ba32-2b4232c0234c" /> |

Fixes https://github.com/software-mansion/react-native-screens/issues/3273.

## Changes

- separate creating new `UITabBarItem` instance and updating tab bar item related props
- create new `UITabBarItem` instance only on `systemItem` change
- restore default system item's icon by extracting it from new `UITabBarSystemItem` instance instead of changing the entire `UITabBarItem`

Run `TestBottomTabs`. Change `badgeValue` on selected tab. You can also check test cases from https://github.com/software-mansion/react-native-screens/pull/3164.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
+++ b/ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
@@ -35,7 +35,7 @@ namespace react = facebook::react;
 #if !RCT_NEW_ARCH_ENABLED
   BOOL _tabItemNeedsAppearanceUpdate;
   BOOL _tabScreenOrientationNeedsUpdate;
-  BOOL _tabBarItemNeedsBaseChange;
+  BOOL _tabBarItemNeedsRecreation;
   BOOL _tabBarItemNeedsUpdate;
   BOOL _scrollEdgeEffectsNeedUpdate;
 #endif // !RCT_NEW_ARCH_ENABLED
@@ -65,7 +65,7 @@ namespace react = facebook::react;
 #if !RCT_NEW_ARCH_ENABLED
   _tabItemNeedsAppearanceUpdate = NO;
   _tabScreenOrientationNeedsUpdate = NO;
-  _tabBarItemNeedsBaseChange = NO;
+  _tabBarItemNeedsRecreation = NO;
   _tabBarItemNeedsUpdate = NO;
   _scrollEdgeEffectsNeedUpdate = NO;
 #endif
@@ -174,7 +174,7 @@ RNS_IGNORE_SUPER_CALL_END
 
 #pragma mark - Prop update utils
 
-- (void)changeBaseTabBarItem
+- (void)createTabBarItem
 {
   UITabBarItem *tabBarItem = nil;
   if (_systemItem != RNSBottomTabsScreenSystemItemNone) {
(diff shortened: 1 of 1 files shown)


ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
```diff
@@ -35,6 +35,7 @@ @implementation RNSBottomTabsScreenComponentView {
 #if !RCT_NEW_ARCH_ENABLED
   BOOL _tabItemNeedsAppearanceUpdate;
   BOOL _tabScreenOrientationNeedsUpdate;
+  BOOL _tabBarItemNeedsBaseChange;
```
kkafar: When reading the code for the first time the `base` part made no sense to me. What we want to signal here is that the tab bar item needs to be recreated & requires allocating completely new instance. What about naming it `_tabBarItemNeedsRecreation` then? What do you think?

kkafar: Also rename the `changeTabBarItem` appropriately. `createTabBarItem`?

ios/bottom-tabs/RNSTabBarAppearanceCoordinator.mm
```diff
@@ -55,10 +55,20 @@ - (void)setIconsForTabBarItem:(UITabBarItem *)tabBarItem
...
+      tabBarItem.image = [[UITabBarItem alloc] initWithTabBarSystemItem:systemItem tag:0].image;
     }
 
     if (screenView.selectedIconSfSymbolName != nil) {
       tabBarItem.selectedImage = [UIImage systemImageNamed:screenView.selectedIconSfSymbolName];
+    } else if (screenView.systemItem != RNSBottomTabsScreenSystemItemNone) {
+      // Restore default system item icon
+      UITabBarSystemItem systemItem =
+          rnscreens::conversion::RNSBottomTabsScreenSystemItemToUITabBarSystemItem(screenView.systemItem);
+      tabBarItem.selectedImage = [[UITabBarItem alloc] initWithTabBarSystemItem:systemItem tag:0].selectedImage;
```
kkafar, line 71: I like this trick.

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
kligarski: I like the names you suggested.

https://github.com/software-mansion/react-native-screens/pull/3290/commits/0e305ab5145a9cd26b99f98b40cf7fe3e934d6cb

ios/bottom-tabs/RNSTabBarAppearanceCoordinator.mm
kligarski: I'm 50/50 whether this is nice or ugly but it prevents the reported bug in more cases so I think it's worth it.

PR #3295, opened 2025-10-14
fix(Android): layout shift on orientation change

### LINKED ISSUE #2933: [android] layout shift bug on new architecture
### Description

When changing orientation we see large layout shifts occurring in the new
archicture. Upon close investigation we isolated the problem to be related to
react-native-screen's Screen component. 

https://github.com/user-attachments/assets/21db04b3-357e-4445-bed8-df640023ea5f

Note, while in this reproducible this is fast. In a large app with busy threads
the layout shift is way more prominent giving users a broken App sensation.

Also note how the padding applied to Screen is showing but then disappears? 

At first we thought this was particularly related to absolute layout positioning
but further investigation proved us wrong. 

We found that this onLayout updateState call [in FabricEnabledViewGroup onLayout](https://github.com/software-mansion/react-native-screens/blob/f79fcae0a35d3f9db23d437c2354ede2ff1906b8/android/src/fabric/java/com/swmansion/rnscreens/FabricEnabledViewGroup.kt#L34-L63) triggered by the [ScreenContainer's onLayout](https://github.com/software-mansion/react-native-screens/blob/f79fcae0a35d3f9db23d437c2354ede2ff1906b8/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt#L46-L56) causes a layout change that is different from the layout represented in the Shadow Tree. In consequence it causes a large layout shift when orientation change happens and overrides the underlying padding.

We don't understand yet why this updateState exists but removing it does seem to
fix things, see video below:

https://github.com/user-attachments/assets/034435a5-bcac-4218-b0f5-61ea490e11bf

Now I'm not sure how to proceed here, I'd like to understand the following:
1. Why does this updateState exist?
2. if we need to keep it how do we ensure possibly schedule the update to occur
   together with its children layout to avoid the layout shift? 
3. is this maybe just a problem of incorrect dimenisions here?
4. does Fabric now make this code unnecessary given I didn't find any bugs in
   our App by testing it without this logic?

Note that I used React Native 0.78.2 and RN Screens 3.37.0 because it's what we have in our app but the code is exactly the same in the latest version which I assume means the same behavior exists. 

### Steps to reproduce

1. Run the [reproducible locally](https://github.com/AndreiCalazans/reproducible-rn-screens-absolute-layout-shift-new-arch/tree/main)
2. Change orientation back and forth notice the blue background flash.

https://github.com/AndreiCalazans/reproducible-rn-screens-absolute-layout-shift-new-arch/tree/main

### Screens version

3.37.0

### React Native version

0.78.2

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

Real device

### Device model

Samsung A22

### Acknowledgements

Yes

kmichalikk:
Fixes #2933

# Description

This PR fixes the bug on android that results in Screen orientation being calculated incorrectly for some time after the physical device orientation has changed. This bug seems to happen due to asynchronous nature of Screen state update emitted to react from native layer. The state contains a size that is set on the Screen frame.

The bug is more visible on slower devices, because with longer processing of consecutive commits, the delay is longer.

It can be noticed on a trivial application (some text + different background color for screen and content view) with some heavy computations artificially added to JS.

### More detailed description:
In `RNSScreenComponentDescriptor`, on Android specifically, we call `setSize()` on the Screen view to hardcode its dimensions. This is needed to account for screen header size, of which React has no idea. It is however the core reason for the bug. The size which we set is outdated. Only after the asynchronous state update from native side is processed, it is corrected. This can take some time. If we comment the code for handling the header size, this problem disappears. But we cannot.

Since the code is staying here, we need to work with it. We agreed on the special value of frameSize = { 0, 0 }, for which we set YogaNode size to Undefined. This forces Yoga to recalculate the layout. It can do so because the parent view is correctly sized at this point (no calls to `setSize()` with outdated values). We cannot do this manually, because the descriptor doesn't have access to parent nodes. Using the commit hook, we detect the orientation change (it can compare old and new RootShadowNode, which have the correct dimensions) and set the frameSize for the ShadowNode which is passed to the descriptor afterwards.

Another problem is that the hook detects the change only once, for one instance of the shadow node. There could be more instances processed before the state is corrected. Luckily, React uses the same state instance for each shadow node instance between the updates. We take advantage of that, once reset to 0, the frameSize stays that way until some other updates changes the instance. Since we only store the screen offset and frameSize in the state, this works good enough. **But it's still pretty hacky**. 

# Changes

Description above. A new feature flag is added. When set, we register a new commit hook that checks if screen orientation changed & sets frameSize to 0. It is detected in Screen descriptor and Yoga is forced to recalculate the layout.

# Before / After

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/8ff1187c-1ac3-4a15-8ccf-f751c692dc44" /> | <video src="https://github.com/user-attachments/assets/d8ebdbb4-4500-4968-98ea-75b475041b76" /> |

# Testing
Use Test2933. You can take a look at the "checkerboard" test from react-native-reanimated.

### DIFF
--- a/android/src/paper/java/com/facebook/react/viewmanagers/RNSScreenManagerDelegate.java
+++ b/android/src/paper/java/com/facebook/react/viewmanagers/RNSScreenManagerDelegate.java
@@ -138,7 +138,7 @@ public class RNSScreenManagerDelegate<T extends View, U extends BaseViewManager<
         mViewManager.setSynchronousShadowStateUpdatesEnabled(view, value == null ? false : (boolean) value);
         break;
       case "androidResetScreenShadowStateOnOrientationChangeEnabled":
-        mViewManager.setAndroidResetScreenShadowStateOnOrientationChangeEnabled(view, value == null ? false : (boolean) value);
+        mViewManager.setAndroidResetScreenShadowStateOnOrientationChangeEnabled(view, value == null ? true : (boolean) value);
         break;
       default:
         super.setProperty(view, propName, value);
--- a/src/fabric/ScreenNativeComponent.ts
+++ b/src/fabric/ScreenNativeComponent.ts
@@ -122,7 +122,7 @@ export interface NativeProps extends ViewProps {
   synchronousShadowStateUpdatesEnabled?: WithDefault<boolean, false>;
   androidResetScreenShadowStateOnOrientationChangeEnabled?: WithDefault<
     boolean,
-    false
+    true
   >;
 }
 


common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
```diff
@@ -2,11 +2,15 @@
...
 #include <fbjni/fbjni.h>
+#include "RNSScreenShadowNodeCommitHook.h"
 #endif // ANDROID
 #include <react/debug/react_native_assert.h>
 #include <react/renderer/components/rnscreens/Props.h>
 #include <react/renderer/components/rnscreens/utils/RectUtil.h>
+#include <react/renderer/components/root/RootShadowNode.h>
 #include <react/renderer/core/ConcreteComponentDescriptor.h>
+#include <react/renderer/uimanager/UIManager.h>
+#include <react/renderer/uimanager/UIManagerCommitHook.h>
```
kkafar: These are most likely not needed on Apple platforms. Move these includes into Android specific section please.

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
```diff
@@ -16,12 +20,22 @@ using namespace rnscreens;
...
+  mutable std::shared_ptr<RNSScreenShadowNodeCommitHook> commitHook_;
+#endif // ANDROID
+
  public:
   using ConcreteComponentDescriptor::ConcreteComponentDescriptor;
 
   void adopt(ShadowNode &shadowNode) const override {
     react_native_assert(dynamic_cast<RNSScreenShadowNode *>(&shadowNode));
-    auto &screenShadowNode = static_cast<RNSScreenShadowNode &>(shadowNode);
+    auto &screenShadowNode = dynamic_cast<RNSScreenShadowNode &>(shadowNode);
```
kkafar: Why change here?

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
```diff
@@ -0,0 +1,111 @@
...
+  auto oldRootProps =
+      std::static_pointer_cast<const RootProps>(oldRootShadowNode->getProps());
+  auto newRootProps =
+      std::static_pointer_cast<const RootProps>(newRootShadowNode->getProps());
+
+  const bool wasHorizontal = isHorizontal_(oldRootProps);
+  const bool willBeHorizontal = isHorizontal_(newRootProps);
+
+  const bool orientationDidChange = (wasHorizontal && !willBeHorizontal) ||
+      (!wasHorizontal && willBeHorizontal);
```
kkafar: wasHorizontal != willBeHorizontal

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
```diff
@@ -0,0 +1,111 @@
...
+          });
+
+      if (rootShadowNodeClone) {
+        finalRootShadowNode = rootShadowNodeClone;
+      }
+    }
+  }
+
+  return std::static_pointer_cast<RootShadowNode>(
+      finalRootShadowNode->ShadowNode::clone(ShadowNodeFragment{}));
```
kkafar: Do we need to make a clone in case we did not modify anything?

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
```diff
@@ -0,0 +1,111 @@
...
+  if (contextContainer_) {
+    auto fabricUIManager =
+        contextContainer_
+            ->at<jni::alias_ref<facebook::react::JFabricUIManager::javaobject>>(
+                "FabricUIManager");
+    fabricUIManager->getBinding()
+        ->getScheduler()
+        ->getUIManager()
+        ->unregisterCommitHook(*this);
+  }
```
kkafar: We never cleanup reference to the `contextContainer`. Shall we do it here?

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
```diff
@@ -0,0 +1,111 @@
...
+        ->getScheduler()
+        ->getUIManager()
+        ->registerCommitHook(*this);
+  }
+}
+
+RNSScreenShadowNodeCommitHook::RNSScreenShadowNodeCommitHook(
+    const RNSScreenShadowNodeCommitHook &other) {
+  contextContainer_ = other.contextContainer_;
+}
```
kkafar: Why is the copy constructor needed?

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
```diff
@@ -0,0 +1,111 @@
...
+        finalRootShadowNode = rootShadowNodeClone;
+      }
+    }
+  }
+
+  return std::static_pointer_cast<RootShadowNode>(
+      finalRootShadowNode->ShadowNode::clone(ShadowNodeFragment{}));
+}
+
+void RNSScreenShadowNodeCommitHook::findScreenNodes(
```
kkafar, line 71: This method might be static

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
```diff
@@ -0,0 +1,111 @@
...
+    const std::shared_ptr<const ShadowNode> &rootShadowNode,
+    std::vector<std::shared_ptr<const RNSScreenShadowNode>> &screenNodes) {
+  std::stack<std::shared_ptr<const ShadowNode>> shadowNodesToVisit;
+  shadowNodesToVisit.emplace(rootShadowNode);
+
+  while (!shadowNodesToVisit.empty()) {
+    auto node = shadowNodesToVisit.top();
+    shadowNodesToVisit.pop();
+
+    for (auto child : node->getChildren()) {
```
kkafar: This copies whole list of shared pointers. Take them by const reference instead
```suggestion
    for (const auto &child : node->getChildren()) {
```

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  return std::static_pointer_cast<RootShadowNode>(
+      finalRootShadowNode->ShadowNode::clone(ShadowNodeFragment{}));
+}
+
+void RNSScreenShadowNodeCommitHook::findScreenNodes(
+    const std::shared_ptr<const ShadowNode> &rootShadowNode,
+    std::vector<std::shared_ptr<const RNSScreenShadowNode>> &screenNodes) {
+  std::stack<std::shared_ptr<const ShadowNode>> shadowNodesToVisit;
```
kkafar: These two lists could also operate on pointer references, instead of copying whole shared pointers.

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
```diff
@@ -0,0 +1,111 @@
...
+  shadowNodesToVisit.emplace(rootShadowNode);
+
+  while (!shadowNodesToVisit.empty()) {
+    auto node = shadowNodesToVisit.top();
+    shadowNodesToVisit.pop();
+
+    for (auto child : node->getChildren()) {
+      if (std::strcmp(node->getComponentName(), "RNSScreen") == 0) {
```
kkafar: What about the component handle approach? The commit hook is initialized by screen component descriptor, which has access to the component handle.

kkafar: You could initialize the commit hook with that information & use it here.

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
```diff
@@ -0,0 +1,111 @@
...
+        screenNodes.push_back(
+            std::dynamic_pointer_cast<const RNSScreenShadowNode>(node));
```
kkafar: why `dynamic_pointer_cast` here?

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.h
```diff
@@ -0,0 +1,49 @@
...
+  virtual void commitHookWasUnregistered(const UIManager &) noexcept override {}
+
+  virtual RootShadowNode::Unshared shadowTreeWillCommit(
+      const ShadowTree &shadowTree,
+      const RootShadowNode::Shared &oldRootShadowNode,
+      const RootShadowNode::Unshared &newRootShadowNode,
+      const ShadowTreeCommitOptions & /*commitOptions*/) noexcept override;
+
+ private:
+  std::shared_ptr<const ContextContainer> contextContainer_;
```
kkafar: Is the decision of retaining the context container here intentional?

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.h
```diff
@@ -0,0 +1,49 @@
...
+
+  void findScreenNodes(
+      const std::shared_ptr<const ShadowNode> &rootShadowNode,
+      std::vector<std::shared_ptr<const RNSScreenShadowNode>> &screenNodes);
+
+  inline bool isHorizontal_(std::weak_ptr<const RootProps> props) {
```
kkafar: this also could be static

kkafar: you should take the props by reference.

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
kmichalikk: My bad, changed it back to static_cast. After consulting, I also changed the cast style in RNSScreenProps

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
kmichalikk: tbh I followed some tutorial and left it there, there's a high chance it's redundant

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
kmichalikk: Even better, there is ::Handle() static method.

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
kmichalikk: dunno, at some point I was trying to be extra safe, but it didn't go anywhere

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.h
kmichalikk: Since the commit hook needs to unregister in the destructor, I need the ref to the container

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.h
kmichalikk: [f6ef130](https://github.com/software-mansion/react-native-screens/pull/3295/commits/f6ef1308d4e5ac40f209953540fcda7c97115978)

kmichalikk: [f6ef130](https://github.com/software-mansion/react-native-screens/pull/3295/commits/f6ef1308d4e5ac40f209953540fcda7c97115978)

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
kmichalikk: [f6ef130](https://github.com/software-mansion/react-native-screens/pull/3295/commits/f6ef1308d4e5ac40f209953540fcda7c97115978)

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
kmichalikk: [f6ef130](https://github.com/software-mansion/react-native-screens/pull/3295/commits/f6ef1308d4e5ac40f209953540fcda7c97115978)

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
kmichalikk: [f6ef130](https://github.com/software-mansion/react-native-screens/pull/3295/commits/f6ef1308d4e5ac40f209953540fcda7c97115978)

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
kmichalikk: [f6ef130](https://github.com/software-mansion/react-native-screens/pull/3295/commits/f6ef1308d4e5ac40f209953540fcda7c97115978)

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
kmichalikk: [f6ef130](https://github.com/software-mansion/react-native-screens/pull/3295/commits/f6ef1308d4e5ac40f209953540fcda7c97115978)

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
kmichalikk: not needed

[f6ef130](https://github.com/software-mansion/react-native-screens/pull/3295/commits/f6ef1308d4e5ac40f209953540fcda7c97115978)

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
kmichalikk: [f6ef130](https://github.com/software-mansion/react-native-screens/pull/3295/commits/f6ef1308d4e5ac40f209953540fcda7c97115978)

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
kmichalikk: [f6ef130](https://github.com/software-mansion/react-native-screens/pull/3295/commits/f6ef1308d4e5ac40f209953540fcda7c97115978)

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
kmichalikk: [f6ef130](https://github.com/software-mansion/react-native-screens/pull/3295/commits/f6ef1308d4e5ac40f209953540fcda7c97115978)

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
kmichalikk: [f6ef130](https://github.com/software-mansion/react-native-screens/pull/3295/commits/f6ef1308d4e5ac40f209953540fcda7c97115978)

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
kmichalikk: should we? doesn't the smart pointer handle this on its own?

kkafar:
@kkafar

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
```diff
@@ -26,14 +39,32 @@ class RNSScreenComponentDescriptor final
...
     auto stateData = state->getData();
-
 #ifdef ANDROID
+    // get featureFlags.experiment.earlyScreenOrientationChangeEnabled from
+    // Screen props enable the commit hook only when the developer asks to
+    react_native_assert(
+        dynamic_cast<const RNSScreenProps *>(
+            screenShadowNode.getProps().get()));
+    auto props =
+        dynamic_cast<const RNSScreenProps *>(screenShadowNode.getProps().get());
```
t0maboro: static_cast

kkafar:
```suggestion
        static_cast<const RNSScreenProps *>(screenShadowNode.getProps().get());
```

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
```diff
@@ -0,0 +1,103 @@
...
+    std::shared_ptr<const ContextContainer> contextContainer)
+    : contextContainer_(contextContainer) {
+  if (contextContainer_) {
+    auto fabricUIManager =
+        contextContainer_
+            ->at<jni::alias_ref<facebook::react::JFabricUIManager::javaobject>>(
+                "FabricUIManager");
+    fabricUIManager->getBinding()
+        ->getScheduler()
+        ->getUIManager()
```
kkafar: Let's extract this part to separate helper method (static method in this file) that extracts the fabricuimanager from context container. This is very specific code & very error prone (we don't check for any null pointers), let's put it in single, self contained place.

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
kkafar: One thing: have you ever observed this destructor run? I'm afraid of retain cycle here.

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
```diff
@@ -0,0 +1,103 @@
...
+            yogaNode->setSize({YGUndefined, YGUndefined});
+
+            return clone;
+          });
+
+      if (rootShadowNodeClone) {
+        finalRootShadowNode = rootShadowNodeClone;
+      }
+    }
+  }
```
kkafar: code style nitpick: to avoid excessive code nesting you can either put the early return condition earlier:

```c++
if (!orientationDidChange) {
    return newRootShadowNode;
}
```

or put the inner code inside separate function.

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.h
kkafar: yeah, but should it be a strong reference?

kkafar: If you observe desctructor being called then it's fine, otherwise we might have retain cycle here.

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
kmichalikk: [250b7cc](https://github.com/software-mansion/react-native-screens/pull/3295/commits/250b7cc1f0b3a383e7c7e290c59555ec6b987d09)

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
kmichalikk: [250b7cc](https://github.com/software-mansion/react-native-screens/pull/3295/commits/250b7cc1f0b3a383e7c7e290c59555ec6b987d09)

common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h
kmichalikk: [250b7cc](https://github.com/software-mansion/react-native-screens/pull/3295/commits/250b7cc1f0b3a383e7c7e290c59555ec6b987d09)

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
kmichalikk: I wasn't able to log anything when application closes, but was able to verify that it runs when metro reloads. I'm changing strong ref to weak ref

[250b7cc](https://github.com/software-mansion/react-native-screens/pull/3295/commits/250b7cc1f0b3a383e7c7e290c59555ec6b987d09)

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.h
kmichalikk: I wasn't able to log anything when application closes, but was able to verify that it runs when metro reloads. I'm changing strong ref to weak ref

[250b7cc](https://github.com/software-mansion/react-native-screens/pull/3295/commits/250b7cc1f0b3a383e7c7e290c59555ec6b987d09)

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.h
```diff
@@ -0,0 +1,53 @@
...
+      const ShadowTree &shadowTree,
+      const RootShadowNode::Shared &oldRootShadowNode,
+      const RootShadowNode::Unshared &newRootShadowNode,
+      const ShadowTreeCommitOptions & /*commitOptions*/) noexcept override;
+
+ private:
+  std::weak_ptr<const ContextContainer> contextContainer_;
+
+  static inline bool isHorizontal_(const RootProps &props) {
+    auto layoutConstraints = props.layoutConstraints;
```
kkafar:
```suggestion
    const auto &layoutConstraints = props.layoutConstraints;
```

No need to copy this.

src/flags.ts
```diff
@@ -42,6 +43,8 @@ const _featureFlags = {
       RNS_SYNCHRONOUS_HEADER_CONFIG_STATE_UPDATES_DEFAULT,
     synchronousHeaderSubviewUpdatesEnabled:
       RNS_SYNCHRONOUS_HEADER_SUBVIEW_STATE_UPDATES_DEFAULT,
+    earlyScreenOrientationChangeEnabled:
```
kkafar: Sorry, didn't notice it earlier, but have a nitpick here. This name is not really informative. I think that we should be either very specific what this flag does or describe issue it solves. 

I'm taking inspiration from [here](https://github.com/facebook/react-native/blob/main/packages/react-native/scripts/featureflags/ReactNativeFeatureFlags.config.js)

What about `androidResetScreenShadowStateOnOrientationChange`?

kmichalikk: [edf8b9f](https://github.com/software-mansion/react-native-screens/pull/3295/commits/edf8b9f2ddd275170019fab0cb1de52791572e0b)

kmichalikk: Added -Enabled to match other flags

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.h
kmichalikk: [edf8b9f](https://github.com/software-mansion/react-native-screens/pull/3295/commits/edf8b9f2ddd275170019fab0cb1de52791572e0b)

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
```diff
@@ -0,0 +1,103 @@
...
+    const facebook::react::ShadowTree &shadowTree,
+    const RootShadowNode::Shared &oldRootShadowNode,
+    const RootShadowNode::Unshared &newRootShadowNode,
+    const ShadowTreeCommitOptions &) noexcept {
+  auto oldRootProps =
+      std::static_pointer_cast<const RootProps>(oldRootShadowNode->getProps());
+  auto newRootProps =
+      std::static_pointer_cast<const RootProps>(newRootShadowNode->getProps());
+
+  const bool wasHorizontal = isHorizontal_(*oldRootProps.get());
```
bartlomiejbloniarz, line 33: Do you want to only check for the explicit orientation change? What if someone enters split screen, the orientation doesn't change, but dimensions do?

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
```diff
@@ -0,0 +1,103 @@
...
+          yogaNode->setSize({YGUndefined, YGUndefined});
+
+          return clone;
+        });
+
+    if (rootShadowNodeClone) {
+      rootShadowNode =
+          std::static_pointer_cast<RootShadowNode>(rootShadowNodeClone);
+    }
+  }
```
bartlomiejbloniarz, line 66: If you have many screens, this approach could have a performance impact (Though if the expected number of screens is small, then it's probably fine (take a look [here](https://github.com/software-mansion/react-native-reanimated/pull/6214))). You can use [this method instead](https://github.com/facebook/react-native/blob/7c603ee261c3eb2d9bf160505427ba06607d3465/packages/react-native/ReactCommon/react/renderer/core/ShadowNode.cpp#L442). Though be careful, it's [internal behaviour will change](https://github.com/facebook/react-native/commit/976c849057ec6056727ffa71cac0d0c0a513a97e) a bit in the upcoming release.

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
kmichalikk: this case should also be covered explicitly, will look into that in another PR, thanks

common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
kmichalikk: will look into that in another PR, thanks

PR #3303, opened 2025-10-16
feat(iOS, Stack): update back button configuration logic to handle iOS 26 and fix other problems

kligarski:
## Description

Fixes back button configuration logic to handle following problems:
- on iOS 26, by default, the back button does not use the title of the previous screen; prior to this PR, `react-native-screens`' behavior was different, 
- `minimal` display mode with `disableBackButtonMenu: true` did not work correctly (back button menu remained enabled) .

The only case that won't be supported after this PR (but it also hasn't been supported prior to this PR) is `generic` display mode on iOS versions prior to 26 when `backTitleFontFamily`, `backTitleFontSize` or `disableBackButtonMenu` is set. I decided that the appropriate approach is to fallback to `default` display mode behavior.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/578.
Closes https://github.com/software-mansion/react-native-screens/issues/3551.

### Reasoning

Prior to this PR, there were many attempts to fix the back button configuration logic (e.g. https://github.com/software-mansion/react-native-screens/pull/2800, https://github.com/software-mansion/react-native-screens/pull/2860, https://github.com/software-mansion/react-native-screens/pull/2867) but after native behavior change in iOS 26, I decided to have a look at this problem once again.

Here are the most important observations:
- In order to use custom fonts or disable back button menu, we need to use a custom back button. If custom back button is used, [native `backButtonDisplayMode` property](https://developer.apple.com/documentation/uikit/uinavigationitem/backbuttondisplaymode-swift.property?language=objc) is ignored (what, to be fair, makes sense).
- Starting from iOS 26, text in the back button is used only if you set `backButtonTitle` property (or use custom back button).
- We can't use `generic` back button text in custom back button (there is no API to get the localized string).

Keeping those observations in mind, this is the configuration flow proposed in this PR:

diagram

Here is the comparison between prior and after this PR based on `Test2809`:

image

### Compatibility with previous versions

In order to maintain compatibility with previous `react-native-screens` and `react-navigation` versions, I added `backButtonUseModernImplementation` prop.

#### New `react-navigation`, new `react-native-screens`

In new `react-navigation` release, we need to set `backButtonUseModernImplementation` to `true` and always pass `backButtonDisplayMode`. Here is the patch to `react-navigation`:

```
diff --git a/packages/native-stack/src/views/useHeaderConfigProps.tsx b/packages/native-stack/src/views/useHeaderConfigProps.tsx
index 74ff388d4..6656946fb 100644
--- a/packages/native-stack/src/views/useHeaderConfigProps.tsx
+++ b/packages/native-stack/src/views/useHeaderConfigProps.tsx
@@ -273,9 +273,8 @@ export function useHeaderConfigProps({
     backTitleVisible: isBackButtonDisplayModeAvailable
       ? undefined
       : headerBackButtonDisplayMode !== 'minimal',
-    backButtonDisplayMode: isBackButtonDisplayModeAvailable
-      ? headerBackButtonDisplayMode
-      : undefined,
+    backButtonDisplayMode: headerBackButtonDisplayMode,
+    backButtonUseModernImplementation: true,
     backTitleFontFamily,
     backTitleFontSize,
     blurEffect: headerBlurEffect,
```

New version of `react-native-screens` will use new implementation only if it receives `backButtonUseModernImplementation=true`. By default, the prop is `false`.

#### Old `react-navigation`, new `react-native-screens`

No `backButtonUseModernImplementation` -> `backButtonUseModernImplementation=false` (default value) -> `react-native-screens` uses legacy implementation, no changes to the back button logic.

#### New `react-navigation`, old `react-native-screens`

`react-native-screens` uses old implementation. Passing `backButtonDisplayMode` always directly, without the logic from `react-navigation` to pass `undefined` in some cases will not change anything - `undefined` was passed only when back button required customization. In such cases, [`backButtonDisplayMode` is ignored by the system](https://developer.apple.com/documentation/uikit/uinavigationitem/backbuttondisplaymode-swift.property?language=objc) already.

## Changes

- refactor back button configuration logic
- update docs, deprecate `backTitleVisible`
- add `backButtonUseNewImplementation` to switch between legacy and modern implementations
- add new cases to `Test2809`

Run `Test2809`. Please note the changes required to `react-navigation` to see new behavior.

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
+++ b/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
@@ -580,11 +580,11 @@ If `backButtonUseModernImplementation` is set to `false`, this prop is used only
 
 ### `backButtonUseModernImplementation` (iOS only)
 
-Used for backward compatibility only. Setting this prop to `false` is recommended.
+Used for backward compatibility only. Setting this prop to `true` is recommended.
 
-Uses previous implementation of native back button configuration.
+Uses new implementation of native back button configuration.
 
-Default is `true`.
+Default is `false`.
 
 ### `headerLeftBarButtonItems` / `headerRightBarButtonsItems` (iOS only)
 


guides/GUIDE_FOR_LIBRARY_AUTHORS.md
```diff
@@ -553,18 +555,39 @@ Controls whether the stack should be in `rtl` or `ltr` form.
...
+On iOS versions prior to 26, using `generic` display mode with `backTitleFontFamily`, `backTitleFontSize`
+or `disableBackButtonMenu` property set is not supported due to limitations in the native platform.
+In such cases, display mode will fallback to `default`.
+
+If `backButtonUseModernImplementation` is set to `false`, this prop is used only when none of: `backTitleFontFamily`,
+`backTitleFontSize`, `disableBackButtonMenu` and `backTitleVisible=false` is set.
+
+### `backButtonUseModernImplementation` (iOS only)
+
+Used for backward compatibility only. Setting this prop to `false` is recommended.
```
t0maboro: That's not recommended in every case - imo we should rather recommend migrating to the new implementation; we should be more specific about the versioning here. If we don't know which version of React Navigation it will be in, we should at least create a ticket to revisit it.

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -747,6 +747,171 @@ + (void)updateViewController:(UIViewController *)vc
...
+                                          withFamily:config.backTitleFontFamily
+                                                size:size
+                                              weight:nil
+                                               style:nil
+                                             variant:nil
+                                     scaleMultiplier:1.0];
+  } else {
+    attrs[NSFontAttributeName] = [UIFont boldSystemFontOfSize:[size floatValue]];
+  }
+  [RNSScreenStackHeaderConfig setTitleAttibutes:attrs forButton:backBarButtonItem];
```
t0maboro: `setTitleAttibutes` => `setTitleAttributes`

kligarski:
I thought about using `compatibilityFlags` but it would solve "new `react-navigation`, old `react-native-screens`" case but not "old `react-navigation`, new `react-native-screens`".

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
kligarski: Sorry, I forgot to change values and description here. Previously the prop was `useLegacyImplementation`. Now it should be correct.

https://github.com/software-mansion/react-native-screens/pull/3303/commits/8009f8c62e5292871b1af8b16252fcfb0ae09766

I created a ticket to add info about versions in the future: https://github.com/software-mansion/react-native-screens-labs/issues/513.

ios/RNSScreenStackHeaderConfig.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3303/commits/83237609da274badce122df336690661df127170

kligarski:
@kkafar

kkafar:
Just a status update from my end: this is a complicated PR that requires a lot of testing and at the same time is not of high priority, therefore I can not really find time to review it properly. I've added it to next iteration with higher priority.

PR #3311, opened 2025-10-20
fix: Disable interactions on minimal required subroot

### LINKED ISSUE #3307: react-native-bottom-tabs got slower from 4.16.x to 4.17.x
### Description

If you install screens 4.16.0, https://github.com/callstackincubator/react-native-bottom-tabs works just fine. If you install 4.17.0 or above, switching tabs is slow (like blocking). See videos:

4.16.0:

https://github.com/user-attachments/assets/e43a5a1f-4cf5-45f3-b2c5-2b4488f93ed8

4.17.0

https://github.com/user-attachments/assets/8cd10b1d-d2d2-431f-8bb5-4a5b85d0b12d

### Steps to reproduce

Install react-native-screens 4.17.x or above and use @bottom-tabs/react-navigation@0.12.x.

https://github.com/software-mansion/react-native-screens/releases/tag/4.17.0

### Screens version

4.17.1

### React Native version

0.81.4

### Platforms

iOS

### JavaScript runtime

None

### Workflow

Expo bare workflow

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

iPhone 17 Pro - iOS 26

### Acknowledgements

Yes

kmichalikk:
## Description

Fixes #3307 

This PR changes the logic for setting `userInteractionsEnabled = false` for the time of screen transitions. Currently, when transition happens, the whole UIWindow gets disabled. This included the time when the whole stack is detached from the window, for instance, when changing the tab in BottomTabs. This unnecessarily removes the ability to click through tabs for the time of the transition. Moreover, in some cases on hot reload, the call to disable the interactions is not matched with re-enabling them. This PR fixes both issues by introducing a helper class with one instance shared across screens, that finds an appropriate view to disable, and stores a reference to it, to make sure the correct one is re-enabled.

The final attempt disables interaction for RNSScreenStack's first subview, which wraps navigation header and screen content, and adds a temporary gesture recognizer for RNSScreenStackView which captures swipe gestures and does nothing so the gestures are not propagated further (for instance, to outer stack when one is nested in another, which would result in the whole stack being dismissed if user happens to swipe again mid transition).

## Changes

Set `userInteractionsEnabled = false` on RNSScreenStack subview. Temporarily add UIPanGestureRecognizer to sink all gesture event.

Use BottomTabsTest to verify that BottomTabs can be clicked through, and Test3093 to verify that the logic for stack doesn't change.


ios/helpers/RNSViewInteractionManager.mm
```diff
@@ -0,0 +1,53 @@
...
+
+@implementation RNSViewInteractionManager {
+  __weak UIView *lastRootWithInteractionsDisabled;
+}
+
+- (instancetype)init
+{
+  lastRootWithInteractionsDisabled = nil;
+  return self;
+}
```
t0maboro, line 15: don't we need to call `[super init]` ?

ios/helpers/RNSViewInteractionManager.mm
```diff
@@ -0,0 +1,53 @@
...
+
+  if (current) {
+    if (lastRootWithInteractionsDisabled && lastRootWithInteractionsDisabled != current) {
+      // When one view already has interactions disabled, and we request a different view,
+      // we need to restore the first one
+      [self enableInteractionsForLastSubtree];
+    }
+
+    if ([current respondsToSelector:@selector(disableInteractions)]) {
+      [static_cast<id<RNSViewInteractionAware>>(current) disableInteractions];
```
t0maboro: I'd check the object type before casting

focux:
Is there any ETA for this to be merged 👀

tarekrached:
Hi @kmichalikk @kmichalikk @kkafar any idea when you’re planning to cut a release with this in it?  Asking bc this will resolve https://github.com/bluesky-social/social-app/issues/9320 which bites me daily 😀

(I really appreciate all you’re doing to give react-native these superpowers)

kmichalikk:
Hi @tarekrached
Once we fix some other regressions we found, we're going to release it asap

matinzd:
What are the current regressions preventing you from the new release? @kmichalikk

kmichalikk:
There are some issues with ScrollView inside FormSheet [draft here](https://github.com/software-mansion/react-native-screens/pull/3479)

matinzd:
Isn't it better to release fixes to the current version that already has some regressions including keyboard handling for bottom sheets on Android? Maybe 4.18.x? 

I am unsure why sending hotfixes to older versions are so slow for react native screens. I want to know how your release schedule works because it's important that everyone get the regression fixes ASAP instead of waiting for a new minor release.

kkafar:
Nightly, use nightly. `react-native-screens@nightly`. We publish a version every single day.

matinzd:
Is it safe to use nightly? What if nightly introduces some other regressions? 

I know about `nightly`. I am using nightly for an app that is not in production yet but not for production apps.

kkafar:
It is as stable as a version released right now would be. 

We do additional testing before releasing stable and that is why 4.19.0 has been postponed. 

But in general we merge only "complete" set of changes to main & main should always be working (ideal scenario). You can pin a nightly version which works for you atm. & in case there is a regression / bug we fix, upgrade to another one before stable version is released.

PR #3312, opened 2025-10-20
chore: Remove orientation settings in BottomTabsTest

kmichalikk:
## Description

The `orientation` options set on BottomTabsTest's TabScreens get annoying when testing other scenarios. Let's maybe remove them, and if they are needed in one specific case, they can be added back for testing purposes without commiting.

## Changes

Removed `orientation` settings from BottomTabsTest.

### DIFF
--- a/apps/src/tests/TestBottomTabs/tabs/Tab4.tsx
+++ b/apps/src/tests/TestBottomTabs/tabs/Tab4.tsx
@@ -118,10 +118,7 @@ export function Tab4() {
         <Stack.Screen
           name="Screen2"
           component={Screen2}
-          options={{
-            headerLargeTitle: true,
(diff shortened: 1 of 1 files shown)


apps/src/tests/TestBottomTabs/tabs/Tab4.tsx
kligarski, line 1: For consistency, I'd also remove `orientation: 'default',` from Screen2.

kmichalikk: [296ae5a](https://github.com/software-mansion/react-native-screens/pull/3312/commits/296ae5a3a5c55998425a2d523af559f4b89b1f36)

PR #3314, opened 2025-10-20
fix(iOS, Tabs): `tabBarTintColor` on iPadOS

kligarski:
## Description

Fixes setting tint color for the tab bar on iPadOS (>= 18).

Simulator Screenshot - iPad Pro 13-inch (M4) - 2025-10-20 at 14 09 05

`tabBar.tintColor` does not work for the floating tab bar. [Apple recommends](https://developer.apple.com/forums/thread/761056?answerId=798245022#798245022) setting the tint color for the `UITabBarController`'s view - that's what we do in this PR.

Unfortunately, when using the sidebar, tint color is set after a delay. This is also reproducible in a native UIKit app and has been mentioned in the same forum thread from the link above to happen [in native iPadOS apps (e.g. Numbers)](https://developer.apple.com/forums/thread/761056?answerId=798343022#798343022).

https://github.com/user-attachments/assets/8e1e0d9f-91be-47ce-9712-a80030bf3b9b

## Changes

- set tint color for `tabBarController.view`

Run `TestBottomTabs` on an iPad.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/bottom-tabs/RNSTabBarAppearanceCoordinator.mm
+++ b/ios/bottom-tabs/RNSTabBarAppearanceCoordinator.mm
@@ -3,6 +3,7 @@
 #import <React/RCTImageLoader.h>
 #import "RCTConvert+RNSBottomTabs.h"
 #import "RNSConversions.h"
+#import "RNSTabBarController.h"
 #import "RNSTabsScreenViewController.h"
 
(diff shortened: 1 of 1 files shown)


kmichalikk (approved):

kkafar (approved):
Thanks!

PR #3320, opened 2025-10-22
feat(iOS, SplitView): Add support for primaryBackgroundStyle prop

t0maboro:
## Description

Added support for the primaryBackgroundStyle prop: https://developer.apple.com/documentation/uikit/uisplitviewcontroller/primarybackgroundstyle?language=objc

This property was previously omitted due to a documentation note stating it was not supported on iOS. However, while working on a separate task, I noticed that different values for this prop do produce varying effects on iOS 26. Therefore, I’ve added support for it accordingly.

## Changes

- Added TS definition
- Added conversions
- Added handling in the appearance coordinator

## Screenshots / GIFs

https://github.com/user-attachments/assets/3b9c81ab-ea04-4306-89cd-9006e054b6f2

Added new example for testing

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/gamma/split-view/RNSSplitViewHostComponentView.mm
+++ b/ios/gamma/split-view/RNSSplitViewHostComponentView.mm
@@ -65,11 +65,8 @@ static const CGFloat epsilon = 1e-6;
   _preferredDisplayMode = UISplitViewControllerDisplayModeAutomatic;
   _displayModeButtonVisibility = UISplitViewControllerDisplayModeButtonVisibilityAutomatic;
 #if !TARGET_OS_TV
-  if (@available(iOS 26.0, *)) {
-    _primaryBackgroundStyle = UISplitViewControllerBackgroundStyleSidebar;
-  } else {
-    _primaryBackgroundStyle = UISplitViewControllerBackgroundStyleNone;
-  }
+  UISplitViewController *tempSplitVC = [[UISplitViewController alloc] init];
+  _primaryBackgroundStyle = tempSplitVC.primaryBackgroundStyle;
 #endif // !TARGET_OS_TV
   _presentsWithGesture = true;
   _showSecondaryToggleButton = false;
--- a/ios/conversion/RNSConversions-SplitView.mm
+++ b/ios/conversion/RNSConversions-SplitView.mm
@@ -73,10 +73,8 @@ UISplitViewControllerBackgroundStyle SplitViewPrimaryBackgroundStyleFromHostProp
       return UISplitViewControllerBackgroundStyleSidebar;
     case Default:
     default:
-      if (@available(iOS 26.0, *)) {
-        return UISplitViewControllerBackgroundStyleSidebar;
-      }
-      return UISplitViewControllerBackgroundStyleNone;
+      UISplitViewController *tempSplitVC = [[UISplitViewController alloc] init];
+      return tempSplitVC.primaryBackgroundStyle;
   }
 }
 #endif // !TARGET_OS_TV


apps/src/tests/TestSplitView/SplitViewPrimaryBackgroundStyle.tsx
```diff
@@ -0,0 +1,91 @@
...
+      <Button title="sidebar" onPress={() => setPrimaryBgStyle('sidebar')} />
+    </View>
+  );
+
+  const renderItems = () =>
+    data.map((item, idx) => (
+      <View key={idx} style={styles.item}>
+        <Text style={styles.itemText}>{item}</Text>
+      </View>
+    ));
```
kmichalikk: nit: this might as well be extracted to separate components

src/components/gamma/split-view/SplitViewHost.types.ts
```diff
@@ -262,6 +267,26 @@ export interface SplitViewHostProps extends ViewProps {
    * @default automatic
    */
   preferredSplitBehavior?: SplitViewSplitBehavior;
+  /**
+   * @summary Specifies the background style of the primary view controller.
+   *
+   * According to the documentation, this property shouldn't have any effect on iOS. However, on iOS 26 the support for this prop was added.
```
kmichalikk: I think it is not that important for library users to put it at the top. I'd consider putting it at the bottom with *Note:* prefix

src/components/gamma/split-view/SplitViewHost.types.ts
```diff
@@ -262,6 +267,26 @@ export interface SplitViewHostProps extends ViewProps {
...
+   * On iOS 18 or lower, we always fall back to `none` which is a system default value.
```
kmichalikk: "system default value" and our "systemDefault" enum is effectively the same for this case, isn't it? Can we just define below what does the systemDefault enum correlate to for each version?

apps/src/tests/TestSplitView/SplitViewPrimaryBackgroundStyle.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3320/commits/3dcc6e29ba6dbc393fe146c5e069349781f177b1

src/components/gamma/split-view/SplitViewHost.types.ts
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3320/commits/d27bf878b4eba0f48497c25c89861ac9aa68444e

src/components/gamma/split-view/SplitViewHost.types.ts
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3320/commits/c609555e721c5dfc50fdc51f59b0c9212efb3254

src/fabric/gamma/SplitViewHostNativeComponent.ts
```diff
@@ -43,6 +43,8 @@ type SplitViewOrientation =
   | 'landscapeLeft'
   | 'landscapeRight';
 
+type SplitViewPrimaryBackgroundStyle = 'systemDefault' | 'none' | 'sidebar';
```
kkafar:
```suggestion
type SplitViewPrimaryBackgroundStyle = 'default' | 'none' | 'sidebar';
```

I'd just name it `default` or `auto`. I'm skeptical sceptical of `systemDefault` because well, it's not "system" default, it's UIKit default. I'm being very nitpicky here. 

I think `default` is the best here, as it regards the styling. `auto` would be better for behaviour related prop. 

We need to look into our naming schemes & create some guidelines. I'll take care of it soon.

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3320/commits/3122c22d017202ff22fed5a85115cc6ebd5feacf

ios/gamma/split-view/RNSSplitViewHostComponentView.mm
```diff
@@ -64,6 +64,13 @@ - (void)resetProps
   _primaryEdge = UISplitViewControllerPrimaryEdgeLeading;
   _preferredDisplayMode = UISplitViewControllerDisplayModeAutomatic;
   _displayModeButtonVisibility = UISplitViewControllerDisplayModeButtonVisibilityAutomatic;
+#if !TARGET_OS_TV
+  if (@available(iOS 26.0, *)) {
+    _primaryBackgroundStyle = UISplitViewControllerBackgroundStyleSidebar;
+  } else {
+    _primaryBackgroundStyle = UISplitViewControllerBackgroundStyleNone;
+  }
```
kligarski: Is there any way to get default value, e.g. by creating a controller and extracting primaryBackgroundStyle from it? Now if something changes, we'll need to update it manually.

src/components/gamma/split-view/SplitViewHost.types.ts
```diff
@@ -262,6 +264,28 @@ export interface SplitViewHostProps extends ViewProps {
...
+   *
+   * On iOS 18 or lower, we always fall back to `none` which is a system default value. Since iOS 26 the system default value is `sidebar`.
+   *
+   * The following values are currently supported:
+   *
+   * - `default` - chosen by the OS, the appropriate background style is based on the device preferences
+   * - `none` - a style that has no visual effect on the background appearance of the primary view controller
+   * - `sidebar` - a style that applies a blurred effect to the background of the primary view controller
+   *
+   * The supported values corresponds to the official UIKit documentation:
```
kligarski:
```suggestion
   * The supported values correspond to the official UIKit documentation:
```

also a nitpick but do those values correspond to UIKit docs if there is no `default` option there?

src/components/gamma/split-view/SplitViewHost.types.ts
```diff
@@ -262,6 +264,28 @@ export interface SplitViewHostProps extends ViewProps {
...
+   * @see {@link https://developer.apple.com/documentation/uikit/uisplitviewcontroller/backgroundstyle|UISplitViewController.BackgroundStyle}
+   *
+   * @default default
+   *
+   * @supported iOS 26 or higher
+   *
+   * @remarks
+   * According to the documentation, this property shouldn't have any effect on iOS. However, on iOS 26 the support for this prop was added.
```
kligarski, line 286: Did you consider the difference between iOS (iPhone) and iPadOS? Maybe what they meant is that on iOS it has no effect but on iPadOS it does?

t0maboro: I was testing it on iPhone in landscape, the prop works either

ios/gamma/split-view/RNSSplitViewHostComponentView.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3320/commits/0275d326ac9e94c9574d5f5ad2e96d2c4a31b8e7

Issue #3321, opened 2025-10-22
[android] Occasional crash when navigating


hannojg:
### Description

In reanimated an issue has been reported of a crash that happens when navigating:

- https://github.com/software-mansion/react-native-reanimated/issues/8422

I debugged this and found that the culprit is that we:

- dispatch transitionProgress events, _while we are drawing_
- reanimated intercepts this event dispatch and starts to perform its operations synchronously
- this causes the native view hierarchy to change eventually, while we are drawing which is causing said crash

```
FATAL EXCEPTION: main
Process: com.rpyoyo.test, PID: 23462
java.lang.NullPointerException: Attempt to read from field 'int android.view.View.mViewFlags' on a null object reference in method 'void android.view.ViewGroup.dispatchGetDisplayList()'
	at android.view.ViewGroup.dispatchGetDisplayList(ViewGroup.java:4584)
	at android.view.View.updateDisplayListIfDirty(View.java:24280)
	at android.view.ViewGroup.recreateChildDisplayList(ViewGroup.java:4612)
	at android.view.ViewGroup.dispatchGetDisplayList(ViewGroup.java:4585)
	at android.view.View.updateDisplayListIfDirty(View.java:24280)
	at android.view.ViewGroup.recreateChildDisplayList(ViewGroup.java:4612)
	at android.view.ViewGroup.dispatchGetDisplayList(ViewGroup.java:4585)
	at android.view.View.updateDisplayListIfDirty(View.java:24280)
	at android.view.ViewGroup.recreateChildDisplayList(ViewGroup.java:4612)
	at android.view.ViewGroup.dispatchGetDisplayList(ViewGroup.java:4585)
	at android.view.View.updateDisplayListIfDirty(View.java:24280)
	at android.view.ViewGroup.recreateChildDisplayList(ViewGroup.java:4612)
	at android.view.ViewGroup.dispatchGetDisplayList(ViewGroup.java:4585)
	at android.view.View.updateDisplayListIfDirty(View.java:24280)
	at android.view.ViewGroup.recreateChildDisplayList(ViewGroup.java:4612)
	at android.view.ViewGroup.dispatchGetDisplayList(ViewGroup.java:4585)
	at android.view.View.updateDisplayListIfDirty(View.java:24280)
	at android.view.ViewGroup.recreateChildDisplayList(ViewGroup.java:4612)
	at android.view.ViewGroup.dispatchGetDisplayList(ViewGroup.java:4585)
[...]

PR #3322, opened 2025-10-22
fix(android): crash while drawing in `dispatchGetDisplayList`

hannojg:
## Description

Fixes https://github.com/software-mansion/react-native-screens/issues/3321

## Changes

The `applyTransformation` callback is called while we are drawing in android, as can be seen in this stack:

Screenshot 2025-10-22 at 11 18 43

The event we dispatch might be intercepted by reanimated which will cause to synchronously execute a mount transaction, which can alter the native view hierarchy, causing the mentioned issue.

The fix here is to move the dispatching off of the drawing stack, and post to the next queue spot.

I think `mFragment.view` is always present? If not we can also use something like `UIThreadUtils` if thats preferred. 

/

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/stack/anim/ScreensAnimation.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/stack/anim/ScreensAnimation.kt
@@ -12,7 +12,12 @@ internal class ScreensAnimation(
         t: Transformation,
     ) {
         super.applyTransformation(interpolatedTime, t)
-        // interpolated time should be the progress of the current transition
-        mFragment.dispatchTransitionProgressEvent(interpolatedTime, !mFragment.isResumed)
+        // This is called while android is drawing. The event might be intercepted by reanimated
+        // which could trigger a sync mount, which can alter the native view hierarchy, which
+        // could lead to a crash, see: https://github.com/software-mansion/react-native-screens/issues/3321
+        mFragment.view?.post {
+            // interpolated time should be the progress of the current transition
+            mFragment.dispatchTransitionProgressEvent(interpolatedTime, !mFragment.isResumed)
+        }
     }
 }


hannojg:
I am wondering if we are worried that animations could be delayed by one frame when using reanimated screens, but i am not sure if thats really a problem here?

hannojg:
An alternative would be to shift the responsibility to either reanimated or rn-core to check to only mount if we are not drawing currently. 
But then i think in android we want to avoid changing anything in the native view hierarchy while we are drawing, so it seems fair that this code here should be explicit about it?

kkafar (commented):
I'm rather reluctant to land this change in screens. 

This is an event sent to JS asynchronously and meant to be consumed in JS. 

The fact that reanimated could hijack it & run sync commit w/ mounting is reanimated implementation detail / operation model & I don't think this lib should rely / adjust to it, especially that likely this can be handled on the other side. 

I'll talk this through with reanimated guys.

RodolfoSilva:
I totally agree with you @kkafar, just wanted to let you know this PR fixes the issue for me. But I'd also appreciate it if this could be solved on the Reanimated side.

kkafar:
Initial conclusions are that we'll want to investigate it further, before we settle on any solution. Will be back once we have some decisions.

RodolfoSilva:
@kkafar this patch fix the issue in the reanimated package:

https://github.com/software-mansion/react-native-reanimated/issues/8422#issuecomment-3437310014

kkafar:
I'm doing something entirely different atm., but it has just crossed my mind that potentially posting the `onTransitionProgress` event might lead to situation, where `onTransitionProgress` is delivered *after* `onTransitionEnd`. Haven't confirmed it by testing yet, just noting a thought experiment. 

This makes me even more reluctant to land it.

tomekzaw:
FYI We've merged https://github.com/software-mansion/react-native-reanimated/pull/8459 to `react-native-reanimated`.

kkafar:
Thanks for the update. I'm closing this PR then. Will reopen on demand if for some reason we change approach to fixing this.

chengit16:
The react native adapted version I am currently using is 4.2.0, and there may still be crashes and crashes, so it should not be an issue with react native adapted, but rather with react native screens. I hope this can be resolved

chengit16:
There is still a crash error on version 4.2.1. My replication scenario is that I have a form, and after adding it, I return to the list and then enter the editing page from the list, and this problem will appear

Bowlerr:
This currently fixes my issue with latest version of reanimated and react-native-screens

Bowlerr:
it fixes most causes but if I spam between two screens, I can still re-create it

PR #3330, opened 2025-10-22
chore: Add a note about the known issue with blur for SplitView

t0maboro:
## Description

Added a note about the known issue with blur for SplitView. Linked to the reported issue and proposed workarounds, as the bug occurs natively and we cannot apply any reasonable fix on RNScreens side.

## Changes

- Updated `README.md` docs


kkafar (approved):
Thanks!

PR #3333, opened 2025-10-22
fix(iOS, Tabs): fix problems with dynamically changing title, icon, systemItem

kligarski:
## Description

Fixes problems related to updating title and icon dynamically, with or without systemItem.

If we use `systemItem`, we have 3 possible cases for `title`:
- we want to use title from `systemItem`, (e.g. "Contacts"),
- we want to use custom title,
- we want to use no title (hiding the label).

On Fabric, there is no difference between `undefined` and `''` title. That's why I added `isTitleUndefined` native prop that will perform this check for us in JS. On Paper, this problem doesn't happen, which simplifies the logic a little bit.

Also, I noticed that icons are not reset if we remove them completely - fixed that as well by setting image to nil if there is no icon name and `systemItem` is `none`.

### Before

https://github.com/user-attachments/assets/78b6a35a-a363-41f1-9e42-f51aa2fe34c0

### After

https://github.com/user-attachments/assets/76ec7f2b-f387-4ea0-ad8b-21d000dfe452

## Changes

- add `isTitleUndefined` prop
- use `systemItem` title if `title == nil`
- match logic on Paper

Run `TestBottomTabs`. Change icon, title (e.g. `undefined`, `''`, `'Title'`) and systemItem (e.g. `'contacts'`, `'search'`) and observe the changes. 

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
+++ b/ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
@@ -201,12 +201,22 @@ RNS_IGNORE_SUPER_CALL_END
     evaluatedTitle = [[UITabBarItem alloc] initWithTabBarSystemItem:systemItem tag:0].title;
   }
 
-  if (![tabBarItem.title isEqualToString:evaluatedTitle]) {
-    tabBarItem.title = evaluatedTitle;
-  }
+  [self updateTabBarItemTitle:evaluatedTitle];
 
   if (![tabBarItem.badgeValue isEqualToString:_badgeValue]) {
     tabBarItem.badgeValue = _badgeValue;
+  }
+}
+
+- (void)updateTabBarItemTitle:(NSString *)newTitle
+{
+  // Setting _controller.title updates also _controller.tabBarItem.title but only if there
+  // is a change to _controller.title. After creating new tabBarItem, _controller.title
+  // remains the same but _controller.tabBarItem.title is nil. For consistency, we always
+  // update both.
+  if (![_controller.tabBarItem.title isEqualToString:newTitle] || ![_controller.title isEqualToString:newTitle]) {
+    _controller.title = newTitle;
+    _controller.tabBarItem.title = newTitle;
   }
 }
 
@@ -257,7 +267,6 @@ RNS_IGNORE_SUPER_CALL_END
       _title = RCTNSStringFromString(newComponentProps.title);
     }
 
-    _controller.title = _title;
     tabBarItemNeedsUpdate = YES;
   }
 
@@ -527,7 +536,6 @@ RNS_IGNORE_SUPER_CALL_END
 {
   _title = title;
   _isTitleUndefined = title == nil;
-  _controller.title = title;
   _tabBarItemNeedsUpdate = YES;
 }
 


ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
```diff
@@ -192,8 +193,16 @@ - (void)updateTabBarItem
...
+  NSString *evaluatedTitle = _title;
+  if (_title == nil && _systemItem != RNSBottomTabsScreenSystemItemNone) {
+    // Restore default system item title
+    UITabBarSystemItem systemItem =
+        rnscreens::conversion::RNSBottomTabsScreenSystemItemToUITabBarSystemItem(_systemItem);
+    evaluatedTitle = [[UITabBarItem alloc] initWithTabBarSystemItem:systemItem tag:0].title;
+  }
+
+  if (![tabBarItem.title isEqualToString:evaluatedTitle]) {
+    tabBarItem.title = evaluatedTitle;
```
kligarski: I wonder if I should use `controller.title` or `tabBarItem.title` here. `tabBarItem.title` seems reasonable in `updateTabBarItem` method but we're initially setting the title via `controller.title` which [handles tabBarItem/navigationItem/etc. at the same time](https://developer.apple.com/documentation/uikit/uiviewcontroller/title). I wonder if desynchronization of `controller.title` and `tabBarItem.title` is problematic or not.

src/fabric/bottom-tabs/BottomTabsScreenNativeComponent.ts
```diff
@@ -125,6 +125,7 @@ export interface NativeProps extends ViewProps {
 
   // General
   title?: string | undefined | null;
+  isTitleUndefined?: WithDefault<boolean, true>;
```
t0maboro, line 128: I'd consider passing just a boolean. For the default value, it's hard for me to determine what it should be.

kligarski: We need to synchronize default values for native side and JS side.

image

1. If we set `isTitleUndefined` to `true`, then we need initial `tabScreenView.title = nil`.
2. If we set `isTitleUndefined` to `false`, then we need initial `tabScreenView.title = ""`.

I used option 1 but 2 is also possible I think. Not sure if it's better though.

t0maboro: okay, thanks for explanation, 1 seems to be better

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
kkafar: Have you checked whether setting the `tabBarItem.title` also updates `controller.title`?

kkafar: Or do they indeed desynchronize?

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
```diff
@@ -238,9 +247,18 @@ - (void)updateProps:(const facebook::react::Props::Shared &)props
...
+  if (newComponentProps.title != oldComponentProps.title ||
+      newComponentProps.isTitleUndefined != oldComponentProps.isTitleUndefined) {
+    _isTitleUndefined = newComponentProps.isTitleUndefined;
+
+    if (_isTitleUndefined) {
+      _title = nil;
+    } else {
+      _title = RCTNSStringFromString(newComponentProps.title);
+    }
+
```
kkafar, line 269: Damn, we need these optionals in codegen

ios/bottom-tabs/RNSBottomTabsScreenComponentView.mm
kligarski: > Have you checked whether setting the tabBarItem.title also updates controller.title?

No. They can desynchronize when using `systemItem` and `title` is `undefined`.

image

kligarski: It turns out that if `controller.title` is set but `tabBarItem.title` is not (e.g. after creating new instance), if you set `controller.title` to the same string, the update is not propagated to `tabBarItem.title`. That's why I decided to just update both: https://github.com/software-mansion/react-native-screens/pull/3333/commits/9daa035dd23e17ebe24e33ad9b218bb990e68287.

kkafar (approved):
Looks good, thanks!

PR #3335, opened 2025-10-23
feat(iOS, SafeAreaView): use synchronous state updates

kligarski:
## Description

Adds support for synchronous state updates for `SafeAreaView` on iOS for RN 0.82+.

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/68cd55ec-9845-40b0-b989-f68b9d840503" /> | <video src="https://github.com/user-attachments/assets/e4ad13d5-076c-40df-bda3-9b74feb831a6" />

## Changes

- add `immediate` flag to `updateState`

Run `TestBottomTabs`. Switch between `Tab1` and `Tab2`.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/safe-area/RNSSafeAreaViewComponentView.mm
+++ b/ios/safe-area/RNSSafeAreaViewComponentView.mm
@@ -8,6 +8,7 @@
 #import "RNSSafeAreaViewNotifications.h"
 
 #if RCT_NEW_ARCH_ENABLED
+#include <cxxreact/ReactNativeVersion.h>
 #import <react/renderer/components/rnscreens/ComponentDescriptors.h>
 #import <react/renderer/components/rnscreens/Props.h>
 #import <rnscreens/RNSSafeAreaViewComponentDescriptor.h>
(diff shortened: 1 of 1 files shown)


kligarski:
> Q: Do we want to have it enabled by default since 82, or follow the same approach with FF per component as I did here: https://github.com/software-mansion/react-native-screens/pull/3282?

cc @kkafar

ios/safe-area/RNSSafeAreaViewComponentView.mm
```diff
@@ -8,6 +8,7 @@
 #import "RNSSafeAreaViewNotifications.h"
 
 #if RCT_NEW_ARCH_ENABLED
+#include <cxxreact/ReactNativeVersion.h>
```
kkafar, line 11: This header exists since 0.79 [(link)](https://github.com/facebook/react-native/commit/f2553d39da18130137945aba8f0d13cdafb76d12), so we should be fine here.

PR #3336, opened 2025-10-23
fix(Android, Stack): Fix FormSheet - SafeAreaView integration

t0maboro:
## Description

Bugfixing the integration between SafeAreaView and FormSheet. I'm preventing the consumption of the bottom inset, allowing it to be passed to SAV component. Additionally, I'm adding some other fixes I caught during testing, for fitToContents and zeroing the bottom inset for keyboard handling.

> [!WARNING]  
> Note:
> We do not manipulate the top inset manually. Therefore, if `SafeAreaView` has top insets enabled, we must retain the top inset even if the formSheet does not currently overflow into the status bar.
>
> This is important because in some specific edge cases - for example, when the keyboard slides in - the formSheet might overlap the status bar. If we ignored the top inset and it suddenly became necessary, it would result in a noticeable visual content jump. To ensure consistency and avoid layout shifts, we always include the top inset upfront, which can be disabled from the application perspective.

Creating a follow-up ticket for handling some cases that I wasn't able to cover with SAV: https://github.com/software-mansion/react-native-screens-labs/issues/565

Fixes: https://github.com/software-mansion/react-native-screens-labs/issues/457 , https://github.com/software-mansion/react-native-screens/issues/2896

## Changes

- Updated the logic for consuming insets in the `SheetDelegate`
- Added a new example for testing.

## Screenshots / GIFs

### Before

https://github.com/user-attachments/assets/c32fc9f4-63e3-41b2-ab31-110a3ec46d68

### After

https://github.com/user-attachments/assets/647c1674-535e-4dd8-8156-9b490588d123

Added `Test3336` for testing.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
@@ -311,7 +311,7 @@ class SheetDelegate(
     ): WindowInsetsCompat {
         val isImeVisible = insets.isVisible(WindowInsetsCompat.Type.ime())
         val imeInset = insets.getInsets(WindowInsetsCompat.Type.ime())
-        val prevInsets = insets.getInsets(WindowInsetsCompat.Type.systemBars())
+        val prevSystemBarsInsets = insets.getInsets(WindowInsetsCompat.Type.systemBars())
 
         if (isImeVisible) {
             isKeyboardVisible = true
@@ -332,7 +332,7 @@ class SheetDelegate(
             isKeyboardVisible = false
         }
 
-        val newBottomInset = if (!isImeVisible) prevInsets.bottom else 0
+        val newBottomInset = if (!isImeVisible) prevSystemBarsInsets.bottom else 0
 
         // Note: We do not manipulate the top inset manually. Therefore, if SafeAreaView has top insets enabled,
         // we must retain the top inset even if the formSheet does not currently overflow into the status bar.
@@ -344,7 +344,7 @@ class SheetDelegate(
             .Builder(insets)
             .setInsets(
                 WindowInsetsCompat.Type.systemBars(),
-                Insets.of(prevInsets.left, prevInsets.top, prevInsets.right, newBottomInset),
+                Insets.of(prevSystemBarsInsets.left, prevSystemBarsInsets.top, prevSystemBarsInsets.right, newBottomInset),
             ).build()
     }
 


android/src/main/java/com/swmansion/rnscreens/bottomsheet/BottomSheetMetrics.kt
```diff
@@ -0,0 +1,29 @@
...
+    val availableHeight: Int,
+    val maxDetent: Double,
+    val maxSheetHeight: Int,
+)
+
+fun getSheetMetrics(
+    screen: Screen,
+    availableHeight: Int,
+    sheetHeight: Int,
+): BottomSheetMetrics {
```
kligarski: Shouldn't we add a guard for `screen.usesFormSheetPresentation()`?

apps/src/tests/Test3336.tsx
```diff
@@ -0,0 +1,207 @@
...
+const withOptionalSafeArea =
+  (
+    Component: React.ComponentType,
+    useSafeArea: boolean,
+    safeAreaStyle?: ViewStyle,
+  ) =>
+  () => {
+    if (Platform.OS === 'android' && useSafeArea) {
+      return (
+        <SafeAreaView edges={{ top: true, bottom: true }} style={safeAreaStyle}>
```
kligarski: In theory, we can separate enabling top edge and bottom edge of safe area view in the test but I'm not sure if it adds any value.

kligarski (approved):
Looks great!

I have 2 questions:
- did you check older Android APIs as well?
- did you check how this behaves with changes from https://github.com/software-mansion/react-native-screens/pull/3248?

android/src/main/java/com/swmansion/rnscreens/bottomsheet/BottomSheetMetrics.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3336/commits/961a119d9b0ce3b27e88d958c798a98270055c1b

apps/src/tests/Test3336.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3336/commits/133378ffd36f573c9de0d15ddff458945db02553

t0maboro:
> * did you check older Android APIs as well?

Yes, I went down to 28

> * did you check how this behaves with changes from #3248?

Yes

android/src/main/java/com/swmansion/rnscreens/bottomsheet/BottomSheetMetrics.kt
```diff
@@ -12,12 +12,16 @@ fun getSheetMetrics(
     screen: Screen,
     availableHeight: Int,
     sheetHeight: Int,
-): BottomSheetMetrics {
+): BottomSheetMetrics? {
+    if (!screen.usesFormSheetPresentation()) {
+        return null
+    }
```
kligarski: Maybe we can use `require(screen.usesFormSheetPresentation()) { "..." }`?

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3336/commits/1f05c0cc788836995f0c0a96c3433c2962d26fc8

t0maboro:
Switching to draft, because the additional effort is needed after merging: #3248

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
```diff
@@ -311,26 +311,14 @@ class SheetDelegate(
     ): WindowInsetsCompat {
         val isImeVisible = insets.isVisible(WindowInsetsCompat.Type.ime())
         val imeInset = insets.getInsets(WindowInsetsCompat.Type.ime())
+        val prevInsets = insets.getInsets(WindowInsetsCompat.Type.systemBars())
```
kkafar: Name it so it's later clear in the code what this variable refers to. When I first read the code I thought those are all the insets. 
```suggestion
        val prevSystemBarsInsets = insets.getInsets(WindowInsetsCompat.Type.systemBars())
```

kligarski (approved):
Looks good!

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3336/commits/ccfc0964c1dd0dc98b3878a7d0d185b0a6474ce5

PR #3337, opened 2025-10-24
feat(Android, Tabs): Add special effects to bottom tabs on Android

Ubax:
## Description

Adds special effects to bottom tabs implementation on Android

https://github.com/user-attachments/assets/bb103090-876c-4058-91d3-f7874c09c4c0

## Changes

1. Added `ViewFinder` helper, which has utils for finding ScrollView and Stack. The implementation is based on iOS one.
2. Added `popToRoot` method to `ScreenStack`. It dispatches dismiss event for every non-root active screen on Stack. It is JS responsibility to remove this screen from stack. This way the dismiss can be prevented.
3. Added handler for repeated tab press event.

Bottom tabs test app

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes


android/src/main/java/com/swmansion/rnscreens/gamma/helpers/ViewFinder.kt
```diff
@@ -0,0 +1,43 @@
+package com.swmansion.rnscreens.gamma.helpers
+
+import android.view.View
+import android.view.ViewGroup
+import android.widget.ScrollView
+import androidx.core.view.isNotEmpty
+import com.swmansion.rnscreens.ScreenStack
+
+class ViewFinder {
```
kkafar: This can just be `object` to make it a singleton. 

```suggestion
object ViewFinder {
```

You might also make these two standalone functions.

kkafar (commented):
Okay, so as we talked, I agree for current dismissal model, but it won't be stable. That is - it might be a subject of change in the future. 

I'll revisit the code now & land it soon

kkafar:
FYI: The root cause here is `SafeAreaListener` recently added to react-navigation. I've started pushing changes to fix this there.

PR #3339, opened 2025-10-24
feat(iOS, SplitView): Add an option for changing number of columns dynamically

t0maboro:
## Description

Previously, the constructor of UISplitViewController relied on a predefined number of columns, and we enforced this constraint using an assert to prevent unexpected runtime behavior when the number of children changed dynamically.

Inspired by the `expo-router` approach, this PR removes the static assert and introduces a more dynamic pattern: the `SplitViewHost` now actively monitors the number of child columns. If the number of children changes, the entire `UISplitViewController` is remounted to reflect the updated configuration.

Fixes https://github.com/software-mansion/react-native-screens-labs/issues/499 .

## Changes

- Added counters to monitor the number of columns and inspectors

## Screenshots / GIFs

### Before

https://github.com/user-attachments/assets/8474db46-9c87-4523-a6fc-effae0ad8555

### After

https://github.com/user-attachments/assets/7a5522c2-c9c8-4afe-8b1f-a3144f97b140

Run any example with SplitView and add/remove columns or inspector.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/gamma/split-view/RNSSplitViewHostController.swift
+++ b/ios/gamma/split-view/RNSSplitViewHostController.swift
@@ -21,6 +21,9 @@ public class RNSSplitViewHostController: UISplitViewController, ReactMountingTra
 
   private let splitViewHostComponentView: RNSSplitViewHostComponentView
 
+  /// This variable is keeping the value of how many columns were set in the initial render. It's used for validation, because SplitView doesn't support changing number of columns dynamically.
+  private let fixedColumnsCount: Int
+
   private let minNumberOfColumns: Int = 2
   private let maxNumberOfColumns: Int = 3
   private let maxNumberOfInspectors: Int = 1
@@ -48,6 +51,7 @@ public class RNSSplitViewHostController: UISplitViewController, ReactMountingTra
     self.splitViewHostComponentView = splitViewHostComponentView
     self.splitViewAppearanceCoordinator = RNSSplitViewAppearanceCoordinator()
     self.splitViewAppearanceApplicator = RNSSplitViewAppearanceApplicator()
+    self.fixedColumnsCount = numberOfColumns
 
     super.init(style: RNSSplitViewHostController.styleByNumberOfColumns(numberOfColumns))
 
@@ -296,6 +300,10 @@ public class RNSSplitViewHostController: UISplitViewController, ReactMountingTra
         && columns.count <= maxNumberOfColumns,
       "[RNScreens] SplitView can only have from \(minNumberOfColumns) to \(maxNumberOfColumns) columns"
     )
+
+    assert(
+      columns.count == fixedColumnsCount,
+      "[RNScreens] SplitView number of columns shouldn't change dynamically")
   }
 
   ///


kligarski (approved):

t0maboro:
> If the entire UISplitNavigationController is remounted (I guess it is recreated, why do we remove the assert for column count, since it is instance-wise?

The instance of SplitView is recreated when the number of columns changes. The assert is no longer necessary because there's no scenario where the number of columns can change dynamically without fully remounting SplitView. Once the `UISplitNavigationController` is created, the column count remains fixed for that instance.

> I think I don't see the benefits this PR introduces. Could you give me some more context here? Maybe a video of what effect we achieve?

The goal of this PR is to improve the development experience. It makes it easier for developers to test different layouts (e.g. 2- or 3-column configurations) without having to rebuild the app each time. With this change, using fast refresh is enough to apply layout updates, which significantly speeds up the development workflow and makes testing layout variations much smoother (as it's presented on the 2nd video in the description).

ios/gamma/split-view/RNSSplitViewHostController.swift
```diff
@@ -300,10 +296,6 @@ public class RNSSplitViewHostController: UISplitViewController, ReactMountingTra
         && columns.count <= maxNumberOfColumns,
       "[RNScreens] SplitView can only have from \(minNumberOfColumns) to \(maxNumberOfColumns) columns"
     )
-
-    assert(
-      columns.count == fixedColumnsCount,
-      "[RNScreens] SplitView number of columns shouldn't change dynamically")
```
kkafar, line 306: These should stay right? As violating them still leads to "bad behaviour".

src/components/gamma/split-view/SplitViewHost.tsx
```diff
@@ -59,8 +60,23 @@ function SplitViewHost(props: SplitViewHostProps) {
...
+    // @ts-ignore - type is valid attribute for child
+    child => child.type === SplitViewScreen.Inspector,
+  );
+
   return (
-    <SplitViewHostNativeComponent {...props} style={styles.container}>
+    <SplitViewHostNativeComponent
+      key={`columns-${columns.length}-inspectors-${inspectors.length}`}
+      {...props}
+      style={styles.container}>
```
kkafar, line 82: This is kinda ok, but let's please add a comment why we do this. I'm thinking of comment for the `key` prop.

ios/gamma/split-view/RNSSplitViewHostController.swift
kkafar: Assertions are not there to assert that things will work in **current setup**. The assertions are there to *assert* inherent invariants of the component and give us warning when we mess something up in the future. So I say, we leave them.

t0maboro: it shouldn't, now we can handle nicely that behavior by recreating the entire SplitView

src/components/gamma/split-view/SplitViewHost.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3339/commits/713764adc0706c319612556976539a08e28489f8

ios/gamma/split-view/RNSSplitViewHostController.swift
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3339/commits/6623fefec8c39bf1cb3f24416e396198d14ec51f

Issue #3162, opened 2025-08-24
[Xcode 26] Theme is not applied everywhere when system is dark mode


huextrat:
### Description

Forcing `DefaultTheme` on react-navigation should applied light theme on native component even if system is in dark mode.

On iOS, if we have `DefaultTheme` we can see that native back button, bottom bar from https://github.com/callstackincubator/react-native-bottom-tabs are not applying the light theme but system theme instead

Home back button style is updated with no reason:
https://github.com/user-attachments/assets/7fa97532-c04e-4a1c-9ca5-e8ffa4df450c

### Steps to reproduce

1. Xcode 26 & iOS 26 (with dark theme on system)
2. Launch app from the repo
3. Open profile 
4. See that the back button does not have the right theme

### Snack or a link to a repository

https://github.com/huextrat/repro-screens-ios

### Screens version

4.15.2

### React Native version

0.79.6

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

kligarski:
Hi, @huextrat!
Thank you for reporting the issue.

Please have a look at [this section ("Keeping the native theme in sync")](https://reactnavigation.org/docs/themes/#keeping-the-native-theme-in-sync) from `react-navigation`'s docs. It fixes the issue in the minimal reproduction example I created.

```tsx
export default function App() {
  useEffect(() => {
    Appearance.setColorScheme('light');
  }, []);

  return (
    <NavigationContainer>
      <Stack.Navigator screenOptions={{ statusBarStyle: 'dark' }}>
        <Stack.Screen name="Screen1" component={Screen1} />
        <Stack.Screen name="Screen2" component={Screen2} />
      </Stack.Navigator>
    </NavigationContainer>
  );
}
```

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/f49851a9-2295-426e-b086-8fb687e03b25" /> | <video src="https://github.com/user-attachments/assets/ca58e4e8-ca8b-4860-923c-4660bb69f7fc" /> |

Please let us know if this fixes the issue for you.

huextrat:
Hi @kligarski,

Yes, defining the `colorScheme` with `Appearance` does work, but this is not necessarily desirable since it will also put the keyboard in light mode on the app. Whereas for the keyboard, for example, for a good user experience, it is preferable to keep the system color preference.

[...]

PR #3342, opened 2025-10-27
feat(iOS, Stack, Tabs, **experimental**): add prop to override user interface style

kligarski:
## Description

Adds `experimental_userInterfaceStyle` prop to `RNSScreenStackHeaderConfig` and `RNSBottomTabsScreenComponentView`. This allows enforcing light theme for navigation bars when device is in dark mode, preventing navigation bar flashes on iOS 26.

> [!IMPORTANT] 
>
> Changes to `react-navigation` are required for this PR to work. To test the chagnes, apply the following patch:
>
> <details>
>   <summary>Patch</summary>
>   
> ```
> diff --git a/packages/native-stack/src/types.tsx b/packages/native-stack/src/types.tsx
> index c2c00d32d..fbdec1817 100644
> --- a/packages/native-stack/src/types.tsx
> +++ b/packages/native-stack/src/types.tsx
> @@ -675,6 +675,15 @@ export type NativeStackNavigationOptions = {
>     * @platform android
>     */
>    unstable_sheetFooter?: () => React.ReactNode;
> +  /**
> +   * Allows to override system appearance for the navigation bar.
> +   *
> +   * Does not support dynamic changes to the prop value for the currently visible screen.
> +   *
> +   * @default unspecified
> +   * @platform ios
> +   */
> +  experimental_userInterfaceStyle?: ScreenStackHeaderConfigProps['experimental_userInterfaceStyle'];
>  };
>  
>  export type NativeStackNavigatorProps = DefaultNavigatorOptions<
> diff --git a/packages/native-stack/src/views/useHeaderConfigProps.tsx b/packages/native-stack/src/views/useHeaderConfigProps.tsx
> index 74ff388d4..5df1ce1e7 100644
> --- a/packages/native-stack/src/views/useHeaderConfigProps.tsx
> +++ b/packages/native-stack/src/views/useHeaderConfigProps.tsx
> @@ -49,6 +49,7 @@ export function useHeaderConfigProps({
>    headerBack,
>    route,
>    title,
> +  experimental_userInterfaceStyle,
>  }: Props) {
>    const { direction } = useLocale();
>    const { colors, fonts } = useTheme();
> @@ -302,6 +303,7 @@ export function useHeaderConfigProps({
>      titleFontWeight: String(titleFontWeight),
>      topInsetEnabled: headerTopInsetEnabled,
>      translucent: translucent === true,
> +    experimental_userInterfaceStyle,
>      children,
>    } as const;
>  }
> ```
>   
> </details>

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/4441139c-894c-4cbe-b9ee-ec83565435f5" /> | <video src="https://github.com/user-attachments/assets/d4e7396d-e918-498d-a262-d806a1384b23" /> |

Fixes https://github.com/software-mansion/react-native-screens/issues/3162.

## Changes

- add `experimental_userInterfaceStyle` to `RNSScreenStackHeaderConfig`
- add `experimental_userInterfaceStyle` to `RNSBottomTabsScreenComponentView`
- use new props when navigating between screens

Run `Test3342`. Please note `IMPORTANT` section above.

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/src/components/bottom-tabs/BottomTabsScreen.types.ts
+++ b/src/components/bottom-tabs/BottomTabsScreen.types.ts
@@ -4,8 +4,12 @@ import type {
   TextStyle,
   ViewProps,
 } from 'react-native';
-import { ScrollEdgeEffect } from '../shared/types';
-import type { PlatformIcon, PlatformIconIOS } from '../../types';
+import type {
+  PlatformIcon,
+  PlatformIconIOS,
+  UserInterfaceStyle,
+  ScrollEdgeEffect,
+} from '../../types';
 
 export type EmptyObject = Record<string, never>;
 
@@ -56,11 +60,6 @@ export type BottomTabsSystemItem =
   | 'recents'
   | 'search'
   | 'topRated';
-
-export type BottomTabsScreenUserInterfaceStyle =
-  | 'unspecified'
-  | 'light'
-  | 'dark';
 
 // Currently iOS-only
 export type BottomTabsScreenOrientation =
@@ -523,10 +522,19 @@ export interface BottomTabsScreenProps {
    *
    * Please note that this prop is marked as **experimental** and might be subject to breaking changes or even removal.
    *
+   * The following values are currently supported:
+   * - `unspecified` - an unspecified interface style,
+   * - `light` - the light interface style,
+   * - `dark` - the dark interface style.
+   *
+   * The supported values correspond to the official UIKit documentation:
+   *
+   * @see {@link https://developer.apple.com/documentation/uikit/uiuserinterfacestyle|UIUserInterfaceStyle}
+   *
    * @default unspecified
    * @platform ios
    */
-  experimental_overrideUserInterfaceStyle?: BottomTabsScreenUserInterfaceStyle;
+  experimental_userInterfaceStyle?: UserInterfaceStyle;
   // #endregion iOS-only appearance
 
   // #region Events
--- a/apps/src/tests/Test3342.tsx
+++ b/apps/src/tests/Test3342.tsx
@@ -77,7 +77,7 @@ function Screen2(stackNavProp: StackNavigationProp) {
             name: 'house',
           },
         },
-        experimental_overrideUserInterfaceStyle: 'light',
+        experimental_userInterfaceStyle: 'light',
       },
       component: () => TabScreen(stackNavProp),
     },
@@ -125,7 +125,7 @@ export default function App() {
       <Stack.Navigator
         screenOptions={{
           statusBarStyle: 'dark',
-          experimental_overrideUserInterfaceStyle: 'light',
+          experimental_userInterfaceStyle: 'light',
         }}>
         <Stack.Screen name="Screen1" component={Screen1} />
         <Stack.Screen
--- a/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
+++ b/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
@@ -770,11 +770,19 @@ A flag to that lets you opt out of insetting the header. You may want to set thi
 
 When set to true, it makes native navigation bar semi transparent. It adds blur effect on iOS. The default value is false.
 
-### `experimental_overrideUserInterfaceStyle` (iOS only)
+### `experimental_userInterfaceStyle` (iOS only)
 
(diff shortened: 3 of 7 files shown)


src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -57,6 +57,11 @@ export type BottomTabsSystemItem =
   | 'search'
   | 'topRated';
 
+export type BottomTabsScreenUserInterfaceStyle =
+  | 'unspecified'
```
kligarski: I thought about using different name than native UIKit (`inherit`) to better describe the default behavior but `RCTConvert` for `UIUserInterfaceStyle` already exists. If we want to use different name, I can do a trick similar to here: https://github.com/software-mansion/react-native-screens/pull/3251/files#diff-27b60a24c9993e708ee299807f00d76eeb2b84b5d7f521101763ad93d0bf513eR495-R505.

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -511,6 +516,15 @@ export interface BottomTabsScreenProps {
     right?: ScrollEdgeEffect;
     top?: ScrollEdgeEffect;
   };
+  /**
+   * @summary Allows to override system appearance for the navigation bar.
```
t0maboro: tab bar? https://github.com/software-mansion/react-native-screens/blob/2e1696f2db1929ad9167e1c802ce5efefbb3e9c2/ios/bottom-tabs/RNSTabBarController.mm#L163

kligarski: Nice catch, thanks.

https://github.com/software-mansion/react-native-screens/pull/3342/commits/ec6dd0a0e624cee7fb5d0a7b28e1be482a19f145

ios/RNSScreenStackHeaderConfig.h
```diff
@@ -58,6 +58,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic) BOOL backButtonInCustomView;
 @property (nonatomic) UISemanticContentAttribute direction;
 @property (nonatomic) UINavigationItemBackButtonDisplayMode backButtonDisplayMode;
+@property (nonatomic) UIUserInterfaceStyle unstable_userInterfaceStyle;
```
kkafar: Let's make this not `unstable` here, as our internal native API is not part of the library public API. 

Dropping this prefix here simplifies method naming & overall usage & we do not lose any flexibility of removing it etc.

What you can do is ot create a separate category (Experimental) to highlight that this is an experimental feature. 

Same for bottom tabs. 

Only the public types should have `experimental` / `unstable` (I prefer `experimental`) prefix.

kligarski: https://github.com/software-mansion/react-native-screens/pull/3342/commits/6e614e3518f72bc38c0f945b196e0ea5b636ecf5, https://github.com/software-mansion/react-native-screens/pull/3342/commits/39aa1acf713799354dc06aba99bddb9aa38c5d14

src/components/bottom-tabs/BottomTabsScreen.types.ts
kkafar: I think that this prop could be in some shared types directory, to not duplicate it between different component as it relates 1:1 with existing native type.

kkafar: I don't mind naming it `unspecified`. Let's link apple docs & go with it.

kkafar (commented):
Hey, this looks good. I'd only ask if we can deduplicate the `UserInterfaceStyle` props. 

Also naming a prop `overrideUserInterfaceStyle` suggests that this is a boolean. Maybe let's name it `userInterfaceStyleOverride` or just `userInterfaceStyle`? I know it's named like this in UIKit but imo id does not make sense. What do you think?

src/components/bottom-tabs/BottomTabsScreen.types.ts
kligarski: https://github.com/software-mansion/react-native-screens/pull/3342/commits/e3f0e3ff11786f19d6db6ae46401b672d774a1f5

PR #3345, opened 2025-10-27
fix(iOS): modal content flicker during enter transition on iPad

### LINKED ISSUE #3569: Incorrect frame height of `formSheet` stack
### Description

Quite often when screens with presentation `formSheet` is opened one over another the height of view is being cropped. Sometimes it will be cropped when expand the sheet to the higher detent. I noticed it happens only with sheets with `ScrollView`.

| Expand issue | Sheet stack issue |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/a74cf10f-a427-4950-8b0b-c6dfbe8942b1"> | <video src="https://github.com/user-attachments/assets/e47c7343-976b-4b0d-9d44-10022105ef42"> |

I can't provide reproducible example, because when I try to create clean example (without app logic) issue reproduces much less. But I did some research and understand that incorrect height is setting here:
https://github.com/software-mansion/react-native-screens/blob/2eed7d391642e180e28922b89dcf4c83d6cf861f/ios/RNSScreenContentWrapper.mm#L161-L166

**If you could tell any workaround it would be nice. Issue is very critical for me**

### Steps to reproduce

---

---

### Screens version

4.19.0

### React Native version

0.82.1

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

None

### Device

iOS simulator

### Device model

_No response_

### Acknowledgements

Yes

kkafar:
## Description

Closes https://github.com/software-mansion/react-native-screens-labs/issues/531
Closes #3569 - also requires enabling synchronous state updates. Must be done with enabling the appropriate feature flag. This PR only improves the situation on RN below 0.82, where synchronous state updates are not available yet, but we won't be able to offer better solution there. 

This PR filters out calls to `updateLayoutMetrics:oldLayoutMetrics:` for modal screen.

This is done to prevent exactly the same issue we have with native
header. Basically e.g. on iPads, the horizontal space available is
constrained (the modal view is narrower than the view underneath) and
the native layout differs from first Yoga layout leading to content
jumps & flickers in the very beginning, before we send first state
updates.

I filter these out now. Relying fully on native layout. Really hard to
state whether it won't break anything. Until now manual testing indicates that the change
is ok. 

### Code remarks

I've put the code into `- [RNSScreenView updateLayoutMetrics:oldLayoutMetrics:]` and not into 
`- [RNSModalScreen updateLayoutMetrics:oldLayoutMetrics:]` deliberately. We're not extending the `RNSScreenView` behaviour. We alter it. We can not omit calling super in `- [RNSModalScreen updateLayoutMetrics:oldLayoutMetrics:]` as we'd fail to update state gathered  there **on every call**, potentially reintroducing https://github.com/software-mansion/react-native-screens/pull/1463. Therefore I decided to handle this special case of subclass related code in superclass. What we could do here is to treat these as separate components & make them separate (do not inherit). This is something that we plan to do in next major, therefore decided against such big change here. 

## Changes

- **Initial reproduction**
- **Filter out setting frame from RN for modals**

## Before

https://github.com/user-attachments/assets/6c6c328f-295c-42e8-9d63-6d1737c15879

Notice this frame and the lack of green bar on the right side: 
image

## After

https://github.com/user-attachments/assets/d63bd902-cd5b-47a7-8cc7-7b5989ad6dd9

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -1511,8 +1511,8 @@ RNS_IGNORE_SUPER_CALL_END
 
   if (parentVC == nil && [self isKindOfClass:RNSModalScreen.class]) {
     // If we're in modal presentation, we don't want to set the frame from RN,
-    // as the available space is most likely restriced & differs from what Yoga
-    // resolves during firt layout. We want to rely on native layout here.
+    // as the available space is most likely restricted & differs from what Yoga
+    // resolves during first layout. We want to rely on native layout here.
     return;
   }
 


apps/src/tests/index.ts
```diff
@@ -167,7 +165,7 @@ export { default as TestFormSheet } from './TestFormSheet';
 export { default as TestAndroidTransitions } from './TestAndroidTransitions';
 export { default as TestAnimation } from './TestAnimation';
 export { default as TestBottomTabs } from './TestBottomTabs';
-export { default as TestBottomTabsOrientation } from './TestBottomTabsOrientation';
 export { default as TestScreenStack } from './TestScreenStack';
 export { default as TestSplitView } from './TestSplitView';
 export { default as TestSafeAreaViewIOS } from './TestSafeAreaViewIOS';
+export { default as Test3345 } from './Test3345';
```
kligarski: nitpick: I'd move this a few lines above, closer to other numbered tests.

ios/RNSScreen.mm
```diff
@@ -1475,6 +1464,14 @@ - (void)updateLayoutMetrics:(const react::LayoutMetrics &)layoutMetrics
   _newLayoutMetrics = layoutMetrics;
   _oldLayoutMetrics = oldLayoutMetrics;
   UIViewController *parentVC = self.reactViewController.parentViewController;
+
+  if (parentVC == nil && [self isKindOfClass:RNSModalScreen.class]) {
+    // If we're in modal presentation, we don't want to set the frame from RN,
+    // as the available space is most likely restriced & differs from what Yoga
+    // resolves during firt layout. We want to rely on native layout here.
```
kligarski:
```suggestion
    // resolves during first layout. We want to rely on native layout here.
```

ios/RNSScreen.mm
```diff
@@ -1464,6 +1464,14 @@ - (void)updateLayoutMetrics:(const react::LayoutMetrics &)layoutMetrics
   _newLayoutMetrics = layoutMetrics;
   _oldLayoutMetrics = oldLayoutMetrics;
   UIViewController *parentVC = self.reactViewController.parentViewController;
+
+  if (parentVC == nil && [self isKindOfClass:RNSModalScreen.class]) {
+    // If we're in modal presentation, we don't want to set the frame from RN,
+    // as the available space is most likely restriced & differs from what Yoga
```
t0maboro:
```suggestion
    // as the available space is most likely restricted & differs from what Yoga
```

kkafar:
The problem is still reproducible when using very specific styles on scrollview. Added reproduction in https://github.com/software-mansion/react-native-screens/pull/3345/commits/ad8eda8462b05dd586f2543323de00303eb76d8f

https://github.com/user-attachments/assets/8dbc00b8-b7f4-4896-aa3a-cc94bd3e2913

t0maboro:
## Notes for Modal Layout Flow - Async vs Sync `updateState`

**Glossary:**

- `W_SCREEN`, `H_SCREEN` – device screen dimensions
- `W_MODAL`, `H_MODAL` – desired modal dimensions (on tablets: `W_MODAL < W_SCREEN`, `H_MODAL < H_SCREEN`)

---

### 1) Baseline flow

We patched `setFrame/Bounds/Center`, and regardless of whether synchronous updates are enabled or not, the flow (based on logs) is identical:

- First/Root Screen [1] is created with frame: `(x, y, w, h) = (0, 0, 0, 0)`
- React Native sends a layout transaction for [1]:
  - `updateLayoutMetrics` is called:
    - `setCenter` => `(x_center, y_center) = (0.5 * W_SCREEN, 0.5 * H_SCREEN)`
    - `setBounds` => `(x, y, w, h) = (0, 0, W_SCREEN, H_SCREEN)`
- Native layout is performed for [1], and `layoutSubviews` adjusts frame:
  - Final frame: `(0, PADDING_TOP, W_SCREEN, H_SCREEN - PADDING_TOP)`
  - `PADDING_TOP` probably accounts for status/header bar - X axis and width are what matters for this example, so we don't care

- Modal screen [2] is created with frame: `(0, 0, 0, 0)`
- `RNSNavigationController` presents [2] via `setModalViewControllers`
- This PR disables `updateLayoutMetrics` for [2]
- Instead, `presentationTransitionWillBegin` triggers `setFrame`:
  - `[2]` gets frame: `(0, 0, W_MODAL, H_MODAL)`
- `didMoveToWindow` is called for [2]

All should be good - **BUT:** `RNSScreenContentWrapper` at this point still has frame `(0, 0, W_SCREEN, H_SCREEN)`, so the content is laid out incorrectly until Yoga recalculates based on [2].

Screenshot 2026-01-27 at 17 28 10

---

### 2a) What happens to `ScreenContentWrapper` [3] in ASYNC state update

- Repeat baseline flow (1)
- `didMoveToWindow` is called on [2]
- `[UISheetPresentationController transitionWillBegin]` runs layout internally, calling `[UISheetLayoutInfo _layout]`
  - This triggers `[RNSScreen viewDidLayoutSubviews]` and then `updateBounds`
- In `updateBounds`, we call `updateState` to tell Yoga about the modal's dimensions
  - Before state is defined as: `state.frameSize = (0, 0)`
- Frame for [2] is already `(0, 0, W_MODAL, H_MODAL)`
- We construct new state: `newState.frameSize = (W_MODAL, H_MODAL)`
- **RIGHT BEFORE updating state**, [3] has frame `(0, 0, W_SCREEN, H_SCREEN)`
  - Presumably, Yoga assumed full screen due to lack of constraints
- State update is performed **asynchronously**
- **RIGHT AFTER updating state**: [3] frame remains `(0, 0, W_SCREEN, H_SCREEN)`
- Layout pass completes and Modal [2] is rendered with incorrect layout as shown in pt. 1

Eventually:
- Yoga updates layout later
- `updateBounds` is triggered again, but nothing is updated (state already matches frame)
- However, at this point [3] has correct frame: `(0, 0, W_MODAL, H_MODAL)`

---

### 2b) What happens to `ScreenContentWrapper` [3] in SYNC state update

- Same exact baseline flow (1) as (2a)
- `didMoveToWindow` is called on [2]
- `[UISheetPresentationController transitionWillBegin]` runs layout internally, calling `[UISheetLayoutInfo _layout]`
  - This triggers `[RNSScreen viewDidLayoutSubviews]` and then `updateBounds`
- In `updateBounds`, we call `updateState` to tell Yoga about the modal's dimensions
  - Before state is defined as: `state.frameSize = (0, 0)`
- Frame for [2] is already `(0, 0, W_MODAL, H_MODAL)`
- We construct new state: `newState.frameSize = (W_MODAL, H_MODAL)`
- **RIGHT BEFORE updating state**, [3] has frame `(0, 0, W_SCREEN, H_SCREEN)`
-------------------------------------------
- State update is performed **synchronously**
- **RIGHT AFTER updating state**, [3] has frame `(0, 0, W_MODAL, H_MODAL)`
- **Yoga immediately recomputes layout in children** [3] has frame `(0, 0, W_MODAL, H_MODAL)` in **the same layout pass**
- Layout pass completes - Drawing pass is started - Modal is rendered with correct layout from the start
- Yoga does not recompute again, `updateBounds` is not hit again - everything is correct already

---

### Potential takeaways:

- Thanks to Yoga adjusting content size in the same layout pass (with sync update), all frames are corrected **before the render phase starts**.
- Better UX: No flickers
- Drawback - Slight delay: The modal appears a bit later because we block to wait on Yoga's layout calculations

https://github.com/user-attachments/assets/17aad7a7-4d80-4bad-b4a2-a1ae34c62296

---

### Conclusion:

Slightly delayed modal is better UX than glitchy one, even if final result is the same.

PR #3348, opened 2025-10-27
feat(Stack): Add background component

t0maboro:
## Description

Add new API to set background when screen or modal doesn't cover the full display

Corresponding PR in react-navigation: https://github.com/react-navigation/react-navigation/pull/12813

Fixes https://github.com/software-mansion/react-native-screens-labs/issues/514

## Changes

- Added a new prop for setting a custom component as a screen background

## Screenshots / GIFs

Screenshot 2025-10-27 at 15 22 19

Screenshot 2025-10-27 at 15 22 25

Adding a new example (requires changes in react navigation, corresponding PR: https://github.com/react-navigation/react-navigation/pull/12813)

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -148,14 +148,6 @@ export interface ScreenProps extends ViewProps {
   shouldFreeze?: boolean;
   children?: React.ReactNode;
   /**
-   * Optional custom React element to be rendered as the background of the screen.
-   *
-   * This component will be rendered behind all foreground screen content.
-   * You can provide any valid React element, such as an ImageBackground,
-   * gradient, or a styled View.
-   */
-  backgroundComponent?: () => React.ReactNode;
-  /**
    * Boolean indicating that swipe dismissal should trigger animation provided by `stackAnimation`. Defaults to `false`.
    *
    * @platform ios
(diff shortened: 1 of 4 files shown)


src/types.tsx
```diff
@@ -147,6 +147,14 @@ export interface ScreenProps extends ViewProps {
...
   shouldFreeze?: boolean;
   children?: React.ReactNode;
+  /**
+   * Optional custom React element to be rendered as the background of the screen.
+   *
+   * This component will be rendered behind all foreground screen content.
+   * You can provide any valid React element, such as an ImageBackground,
+   * gradient, or a styled View.
+   */
+  backgroundComponent?: () => React.ReactNode;
```
satya164: This prop shouldn't be called `*Component`, because it accepts a render function and not a component. e.g., you can't pass a class component here.

In React Navigation, we typically would call it something like `contentBackground` (like we have `tabBarBackground`, `headerBackground`). Alternatively, since it's a render function, usually it's common to use `render*` prefix, e.g., `renderBackground`.

Though this API seems unnecessary in screens as it can be implemented by consumers if needed,

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3348/commits/b8fe5964147fde62a159e6570d61a4c9c575d58e

t0maboro:
As discussed internally, we're considering not using SAV for FormSheet here https://github.com/software-mansion/react-native-screens/pull/3336 , so the background component doesn't have any concrete use case atm. Moving this PR do draft until the situation will clear and the final decision will be made.

t0maboro:
Regarding the latest updates for FormSheets on Android, it seems like it's no longer necessary. I'm closing this PR, but feel free to reopen if we have any use cases.

PR #3350, opened 2025-10-28
feat(Android, Tabs): handle preventing tab change on Android

kligarski:
## Description

Allows preventing tab change on Android. Previously, bottom navigation bar would be desynchronized with shown tab screen (menu item would change even if tab screen would not).

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/ac6da530-29d3-4772-ac34-33dbb68cb951" /> | <video src="https://github.com/user-attachments/assets/dde80ace-aed1-4e2c-b34a-07478adb6e65" /> |

> [!IMPORTANT]
> 
> Due to changes in this PR, tab change behaves differently when tab screens are heavy.
> 
> | before | after |
> | --- | --- |
> | <video src="https://github.com/user-attachments/assets/351bf5cc-6520-4d89-97e1-125d47dafa7b" /> | <video src="https://github.com/user-attachments/assets/fe0d3765-6e39-4da3-b1f8-bfcd7268dc8e" />
>
> This matches behavior on iOS but I'm not sure if this is something we want.
>
> https://github.com/user-attachments/assets/505a5707-e89f-4c86-a4df-7742cd5643c3

Closes https://github.com/software-mansion/react-native-screens-labs/issues/561.

## Changes

- add `menuItemSelectedViaContainerUpdate` to differentiate between user menu item taps and updates from JS
- allow item selection only if `menuItemSelectedViaContainerUpdate`, otherwise just send event to JS

Run `TestBottomTabs`. Apply this patch to `BottomTabsContainer` to simulate preventing tab change to Tab4:
<details>
  <summary>Patch</summary>
  
```
diff --git a/apps/src/shared/gamma/containers/bottom-tabs/BottomTabsContainer.tsx b/apps/src/shared/gamma/containers/bottom-tabs/BottomTabsContainer.tsx
index 7bba66b48..5f535c945 100644
--- a/apps/src/shared/gamma/containers/bottom-tabs/BottomTabsContainer.tsx
+++ b/apps/src/shared/gamma/containers/bottom-tabs/BottomTabsContainer.tsx
@@ -51,6 +51,10 @@ export function BottomTabsContainer(props: BottomTabsContainerProps) {
     (event: NativeSyntheticEvent<NativeFocusChangeEvent>) => {
       const tabKey = event.nativeEvent.tabKey;
 
+      if (tabKey === 'Tab4') {
+        return;
+      }
+
       // Use `startTransition` only if the state is controlled in JS
       // const transitionFn = !configWrapper.config.controlledBottomTabs
       //   ? startTransition

```
  
</details>

Try to select tab4.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
@@ -326,6 +326,10 @@ class TabsHost(
 
         appearanceCoordinator.updateTabAppearance(this)
 
+        menuItemSelectedViaContainerUpdate = true
+        bottomNavigationView.selectedItemId =
+            checkNotNull(getSelectedTabScreenFragmentId()) { "[RNScreens] A single selected tab must be present" }
+
         post {
             refreshLayout()
             RNSLog.d(TAG, "BottomNavigationView request layout")
@@ -342,10 +346,6 @@ class TabsHost(
         if (newFocusedTab === oldFocusedTab) {
             return
         }
-
-        menuItemSelectedViaContainerUpdate = true
-        bottomNavigationView.selectedItemId =
-            checkNotNull(getSelectedTabScreenFragmentId()) { "[RNScreens] A single selected tab must be present" }
 
         requireFragmentManager
             .beginTransaction()


android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
```diff
@@ -317,6 +326,7 @@ class TabsHost(
 
         appearanceCoordinator.updateTabAppearance(this)
 
+        menuItemSelectedViaContainerUpdate = true
```
kmichalikk, line 329: Can we call it in updateSelectedTab instead of updateBottomNavigationViewAppearance? It would be a bit more logical there, or these 2 are not always called together ?

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
```diff
@@ -134,6 +134,10 @@ class TabsHost(
 
     private var interfaceInsetsChangeListener: SafeAreaView? = null
 
+    // We need to differentiate between user tapping the menu item
+    // and update requested from JS.
+    private var menuItemSelectedViaContainerUpdate = false
```
kmichalikk, line 139: Am I missing something, or is this supposed to not be triggered when controlledBottomTabs=false, but it is? Also IMO the name is a bit unclear.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
kligarski: I thought about it as well but I think on Android managing `BottomNavigationView` and shown screens/fragments is separated and those methods reflect that. Maybe we should think about renaming those methods though.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
kligarski: On Android, only controlled bottom tabs are currently supported.

I did use `viaContainerUpdate` name because I guess that there is a possibility to call container update from native code, not only as a result of JS focused tab change. I can maybe reverse the logic and use `menuItemSelectedByUser` or something like this but I'm not sure.

kligarski:
This PR is outdated, this will be handled via future prevent default mechanism (https://github.com/software-mansion/react-native-screens-labs/issues/1078).

PR #3353, opened 2025-10-29
feat: Bring back gestureResponseDistance on iOS 26

kmichalikk:
## Description

Closes https://github.com/software-mansion/react-native-screens-labs/issues/521

We previously removed the support for gestureResponseDistance on iOS 26, because it was seemingly conflicting with the new native behavior - now, each screen can be dismissed by swiping on the content, in addition to swiping from the edge. For the one case of animations, the old gesture recognizer is used, but to match the new one, it also didn't respond the the responseDistance.

After the changes, the recognizer is brought back to work as in ios 18, and the new interactiveContentPop has the check added.

Use Gestures test -- should work the same way as iOS 18. Use Test3093 with modifications in JS to test custom gestures.
```ts
animation: 'slide_from_bottom',
animationMatchesGesture: true,
// gestureResponseDistance: { ... }
```

### DIFF
--- a/ios/RNSScreenStack.mm
+++ b/ios/RNSScreenStack.mm
@@ -902,12 +902,15 @@ RNS_IGNORE_SUPER_CALL_END
       // it should only recognize with `customAnimationOnSwipe` set
       return NO;
     }
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
     if (@available(iOS 26, *)) {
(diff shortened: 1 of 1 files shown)


kkafar (approved):
Looks sensible, thank you!

I think we need to refactor this code with separated approach for each gesture recognizer. Currently these are all entangled & the code is hard to follow. Not in this PR obviously.

PR #3355, opened 2025-10-29
fix(iOS): load header items icons synchronously when feasible

johankasperi:
## Description

When using RCTImageLoader for images in headerRightItems/headerLeftItems the asynchronous `loadImageWithURLRequest` produces som layout bugs as shown under "Screenshots". These are:
1. When setting title and image the title is briefly visible before the image is set.
2. When setting only image the bar button item is briefly shown as a very wide item without any content. 

## Changes

1. If the image is a `ImageRequireSource` load the image with `[RCTConvert UIImage]`. I know its deprecated but thats the only synchronous image loader that I know of.
2. If the image is a `ImageURISource` set the bar button title in the completion block of `loadImageWithURLRequest`
3. Updates `BarButtonItems` example to latest types/api from react-navigation/native-stack.

### Before

**Title briefly shown**

https://github.com/user-attachments/assets/0b964e1a-cfea-4be8-bdf8-a50297ceee9e

**Wide bar button item while `loadImageWithURLRequest` completes**

https://github.com/user-attachments/assets/0c2fcca0-1860-4cea-b4fe-16d07a5e3a95

### After

https://github.com/user-attachments/assets/653f4c16-c1ff-4aea-bd18-211a1e6eb404

Use any of the examples with required images in BarButtonItems

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSBarButtonItem.mm
+++ b/ios/RNSBarButtonItem.mm
@@ -25,25 +25,20 @@
   NSDictionary *templateSourceObj = dict[@"templateSource"];
   NSString *sfSymbolName = dict[@"sfSymbolName"];
 
-  RCTImageSource *imageSource = nil;
-  if (imageSourceObj) {
-    imageSource = [RCTConvert RCTImageSource:imageSourceObj];
-  } else if (templateSourceObj) {
-    imageSource = [RCTConvert RCTImageSource:templateSourceObj];
-  }
-
-  if (imageSource) {
+  if (imageSourceObj != nil) {
     void (^completionAction)(NSError *_Nullable, UIImage *_Nullable) =
         ^(NSError *_Nullable error, UIImage *_Nullable image) {
-          UIImage *imageWithRenderingMode = nil;
-          if (imageSourceObj) {
-            imageWithRenderingMode = [image imageWithRenderingMode:UIImageRenderingModeAlwaysOriginal];
-          } else if (templateSourceObj) {
-            imageWithRenderingMode = [image imageWithRenderingMode:UIImageRenderingModeAlwaysTemplate];
-          }
-          self.image = imageWithRenderingMode;
+          self.image = [image imageWithRenderingMode:UIImageRenderingModeAlwaysOriginal];
         };
     [self loadImageSyncFromImageSourceJson:imageSourceObj withImageLoader:imageLoader completionBlock:completionAction];
+  } else if (templateSourceObj != nil) {
+    void (^completionAction)(NSError *_Nullable, UIImage *_Nullable) =
+        ^(NSError *_Nullable error, UIImage *_Nullable image) {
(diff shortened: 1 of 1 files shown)


kkafar:
Also disadvantage of using RCTConvert is that we don't have any image caching.

ios/RNSBarButtonItem.mm
```diff
@@ -20,25 +20,32 @@ - (instancetype)initWithConfig:(NSDictionary<NSString *, id> *)dict
     return self;
   }
 
-  self.title = dict[@"title"];
```
kkafar, line 23: Why did you resign from setting title here and postponed it for later? Was there a particular reason here?

kkafar (approved):
I've refactored the code a bit. I'll be willing to land this if you confirm that also works for you.

johankasperi:
> I've refactored the code a bit. I'll be willing to land this if you confirm that also works for you.

Thank you! And thanks for explaining how RCTImageLoader works in production/debug mode. Your refactor is working in my testing so feel free to merge

ios/RNSBarButtonItem.mm
johankasperi: I didn't want to set the title before remote images had been downloaded

johankasperi: But shouldn't be needed anymore since from what I understand also remote images will be loaded synchronously in production mode?

johankasperi:
> > I've refactored the code a bit. I'll be willing to land this if you confirm that also works for you.
> 
> Thank you! And thanks for explaining how RCTImageLoader works in production/debug mode. Your refactor is working in my testing so feel free to merge

One thing to be aware of now though is that you cant load images of type `ImageURISource` in debug mode. It works in production mode. But using `ImageURISource` for icons in the header should not be a common use case

johankasperi:
@kkafar I refactored now so we got the changes from https://github.com/software-mansion/react-native-screens/pull/3347 Feel free to have a look at it. Thanks!

ios/RNSBarButtonItem.mm
kkafar: No, only assets stored on the device will be loaded synchronously. If you need to download header button images from some remote source during runtime (is this a valid use case?) it'll be done asynchronously. Packager (metro) assets are treated like device-local assets, so they're exception to that rule & will be loaded synchronously.

johankasperi: Yeah I don't think its a valid use case. Just something to be aware of that the title will be shown until the image is loaded then if the download is done asynchronously.

kkafar (approved):
Thanks for catching this. 

I'll land this PR once CI passes. 

Regarding ImageURISource - thanks for reporting this. We might wanna handle this separately.

PR #3364, opened 2025-10-31
feat(iOS, SplitView): Migrate to new invalidate method

t0maboro:
## Description

SplitViewHost initially was responsible for invalidating children when the whole structure was destroyed. This approach is no longer necessary, as every screen can invalidate itself directly when it's deleted.

## Changes

- Removed invalidate from SplitViewHost

The easiest way to test is to change with fast refresh the number of columns between 2 and 3 and verifying if the breakpoint in RNSSplitViewScreenComponentView was hit.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/gamma/split-view/RNSSplitViewScreenComponentView.mm
+++ b/ios/gamma/split-view/RNSSplitViewScreenComponentView.mm
@@ -85,14 +85,6 @@ namespace react = facebook::react;
   _columnType = RNSSplitViewScreenColumnTypeColumn;
 }
 
-- (void)invalidate
-{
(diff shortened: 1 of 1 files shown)


t0maboro:
@kkafar iirc, the only reason it was implemented this way was because `SplitView` maintains a fixed number of columns that can't be unmounted due to existing assertion. This made it possible for the `Host` to decide to remove all children at once. After the 2nd thought, it seems to me to be more reasonable for every `Screen` to take care of itself. For `SplitView` we don't need to wait for any transitions to be performed.

kkafar (approved):
Let's do it then.

PR #3374, opened 2025-11-06
fix(JS, Stack): consider default value for `stackPresentation` and `headerConfig.hidden` in `ScreenStackItem`

kligarski:
## Description

Fixes logic in `ScreenStackItem` to properly take into account default values of `stackPresentation` and `headerConfig.hidden` props. Previously, if you e.g. specified `headerConfig.hidden = false` but did not `stackPresentation = 'push'`, `isHeaderInModal` would become true, which is incorrect.

In this PR, we consider those values taking into account default value.

## Changes

- consider default value for `stackPresentation` and `headerConfig.hidden` in `ScreenStackItem`

## Screenshots / GIFs

| before | after |
| --- | --- |
| image | image |

```tsx
import React from 'react';
import { Text } from 'react-native';
import { ScreenStack, ScreenStackItem } from 'react-native-screens';

  return (
    <ScreenStack style={{ flex: 1 }}>
      <ScreenStackItem
        key="home"
        screenId="home"
        activityState={2}
        // stackPresentation="push"
        headerConfig={{
          title: 'Home',
          hidden: false,
        }}>
        <Text>Hello, World!</Text>
      </ScreenStackItem>
    </ScreenStack>
  );
}
```

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/components/ScreenStackItem.tsx
+++ b/src/components/ScreenStackItem.tsx
@@ -185,7 +185,7 @@ export default React.forwardRef(ScreenStackItem);
 
 function getPositioningStyle(
   allowedDetents: ScreenProps['sheetAllowedDetents'],
-  presentation?: StackPresentationTypes,
+  presentation: StackPresentationTypes,
 ) {
   const isIOS = Platform.OS === 'ios';
 


kligarski:
Added a ticket on our board to investigate this: https://github.com/software-mansion/react-native-screens-labs/issues/538.

PR #3375, opened 2025-11-06
chore(iOS): Add multi-scene support for example apps

t0maboro:
## Description

Added support for `UIWindowSceneDelegate` to enable multi-window application testing. Introduced a new flag, `RNS_USE_SCENE_DELEGATE`, which is disabled by default. Enabling this flag allows testing of the application in multi-window mode, which is especially useful on iPads. At this point, it's recommended to continue supporting both flows during testing.

## Changes

- Implemented SceneDelegate in our examples
- Separated AppDelegate and SceneDelegate logic
- Added a script for handling both setups
- Updated Info.plist

## Screenshots / GIFs

FabricExample - multiwindow

Screenshot 2025-11-06 at 20 57 29

FabricExample - single app

Screenshot 2025-11-06 at 21 01 01

Example - multiwindow

Screenshot 2025-11-06 at 21 07 28

Example - single app

Screenshot 2025-11-06 at 21 10 04

Launched FabricExample & Example on iPad with both `RNS_USE_SCENE_DELEGATE` enabled & disabled. Verified that `new window` option is present when the flag is enabled.

Screenshot 2025-11-06 at 17 29 15

## Checklist

- [x] Ensured that CI passes


Example/ios/AppDelegate.swift
```diff
@@ -1,48 +1,75 @@
...
-      in: window,
-      launchOptions: launchOptions
-    )
-
-    return true
-  }
-}
+#if RNS_USE_SCENE_DELEGATE
+
+  import ReactAppDependencyProvider
```
t0maboro: I lost the fight with the formatter 😄 , any idea how to bypass it?

Example/Gemfile
```diff
@@ -13,3 +13,5 @@ gem 'bigdecimal'
 gem 'logger'
 gem 'benchmark'
 gem 'mutex_m'
+
+gem 'plist'
```
kkafar, line 18: Example & FabricExample apps are mostly kept as close as possible to the bare RN project template. 

Let's add a comment here, that this is our custom dependency & needs to be preserved, because somebody is gonna remove that when upgrading the app for sure.

scripts/ios/rns_set_swift_compilation_flags.rb
kkafar, line 1: Let's not duplicate these scripts. We have some now in Example & FabricExample. Would there be a way to share these?

FabricExample/ios/ReactNativeDelegate.swift
```diff
@@ -0,0 +1,16 @@
...
+  }
+
+  override func bundleURL() -> URL? {
+    #if DEBUG
+      RCTBundleURLProvider.sharedSettings().jsBundleURL(forBundleRoot: "index")
+    #else
+      Bundle.main.url(forResource: "main", withExtension: "jsbundle")
+    #endif
+  }
+}
```
kkafar, line 16: Is this delegate needed? does it do something more than `RCTDefaultReactNativeFactoryDelegate`?

t0maboro: Default has an empty implementation, forcing us to override it https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/AppDelegate/RCTDefaultReactNativeFactoryDelegate.mm#L22-L27

Example/Gemfile
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3375/commits/87eb61837d46b3c6d533bf10fe48a624bab1f677

scripts/ios/rns_set_swift_compilation_flags.rb
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3375/commits/90eb064c4f2564a02a251588498751fcc8c36484

FabricExample/ios/ReactNativeDelegate.swift
kkafar: Makes sense now. Thanks

PR #3378, opened 2025-11-07
refactor(Android, Stack): Refactor sheet animation code

t0maboro:
## Description

Followup for: https://github.com/software-mansion/react-native-screens/pull/3248#discussion_r2495067335 

Note: @kkafar I'm aware that it differs from what we discussed internally, but imo it may make sense to do it in the way I'm proposing here, let me know what do you think

Fixes https://github.com/software-mansion/react-native-screens-labs/issues/537

## Changes

- Extracted sheet animation code to a separate class

Regression testing on `Test3248`

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
@@ -304,7 +304,7 @@ class ScreenStackFragment :
         val sheetDelegate = requireSheetDelegate()
         val dimmingDelegate = requireDimmingDelegate()
 
-        val sheetAnimationData =
+        val sheetAnimationContext =
             SheetDelegate.SheetAnimationContext(
                 this,
                 this.screen,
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
```diff
@@ -568,4 +462,14 @@ class ScreenStackFragment :
...
+    private fun requireSheetAnimationDelegate(
+        sheetDelegate: SheetDelegate,
+        dimmingDelegate: DimmingViewManager,
+    ): SheetAnimationDelegate {
+        if (sheetAnimationDelegate == null) {
+            sheetAnimationDelegate = SheetAnimationDelegate(this, screen, coordinatorLayout, sheetDelegate, dimmingDelegate)
+        }
+        return sheetAnimationDelegate!!
+    }
 }
```
kkafar: Don't take & store `sheetDelegate` & `dimmingDelegate` inside `SheetAnimationDelegate`. This only potentially creates retain cycles. 
 
Instead pass the information needed as arguments to particular function invocations.

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3378/commits/7f42ff38544f1d3d43851e4b43c776d4a9ab070b

PR #3383, opened 2025-11-10
chore(iOS): bump min deployment target to 15.1 for example apps

kligarski:
## Description

Bumps minimum deployment target to iOS 15.1 for example apps.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/535.

## Changes

- bump minimum deployment target to iOS 15.1 for `Example` and `FabricExample`

Run example apps.

## Checklist

- [x] Ensured that CI passes


kligarski:
> How about this one?
> 
> https://github.com/software-mansion/react-native-screens/blob/47950dc67a4d3f3f9b9522dda57063db0f89db2a/FabricExample/ios/FabricExample.xcodeproj/project.pbxproj#L398

Thank you for catching this. I used Xcode to change the target but it turns out you need to change it for project and target separately.

https://github.com/user-attachments/assets/9515b41e-e3d0-467d-b641-beeceee68a8b

PR #3384, opened 2025-11-10
fix(Android): check if `selectedItemId` has changed before setting

johankasperi:
## Description

`onNativeFocusChange` is called on android after closing a screen "on top of" a screen having bottom tabs. This is because we are setting `bottomNavigationView.selectedItemId` in `updateBottomNavigationViewAppearance` even though it have not changed because [`onNavigationItemSelectedListener` is notified on reselect](https://developer.android.com/reference/com/google/android/material/bottomnavigation/BottomNavigationView#setOnNavigationItemSelectedListener(com.google.android.material.bottomnavigation.BottomNavigationView.OnNavigationItemSelectedListener))

This differs from the implementation on iOS (where `onNativeFocusChange` is only called on tab press) and creates some problems. For example will ["tab press" be emitted in react-navigation](https://github.com/react-navigation/react-navigation/blob/5da0c934f7e8cc557e7d598307dee4ac5a6a2c74/packages/bottom-tabs/src/unstable/NativeBottomTabView.native.tsx#L157) causing the scrollview in the tab to scroll to top after closing a modal. Or if there is a nested stack in the tab, the stack will pop to the route on index 0.

## Changes

Check that `selectedItemId` has changed before setting it to `bottomNavigationView.selectedItemId`

### Before

https://github.com/user-attachments/assets/d6c377b8-62a6-4366-b553-1634623e1fca

### After

https://github.com/user-attachments/assets/7f60d90f-74d0-4e49-9098-0836a8af6085

I've created a new test named "Test3384".

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/apps/src/tests/Test3384.tsx
+++ b/apps/src/tests/Test3384.tsx
@@ -1,0 +1,61 @@
+import { NavigationContainer, useNavigation } from '@react-navigation/native';
+import React from 'react'
+import { BottomTabsContainer } from '../shared/gamma/containers/bottom-tabs/BottomTabsContainer';
+import { createNativeStackNavigator } from '@react-navigation/native-stack';
+import { Alert, Button, Text, View } from 'react-native';
+
+function ModalScreen() {
+  return <Text>Modal</Text>
+}
+
+function Tab1() {
+  const navigation = useNavigation()
+
+  return <View>
+    <Text>Tab 1</Text>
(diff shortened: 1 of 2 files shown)


johankasperi:
Same change was made in https://github.com/software-mansion/react-native-screens/pull/3440 so closing this

PR #3389, opened 2025-11-11
[iOS] Fix user interaction never being reenabled when cancelling a video fullscreen dismiss gesture

### LINKED ISSUE #3388: [iOS] Swiping a video player to dismiss in the wrong direction causes the app to become unresponsive to input
### Description

Hey there!

We're seeing an issue when using `expo-video` with the latest version of React Native screens. Swiping to dismiss the video when in the native fullscreen mode works fine, but when you swipe in the _wrong direction_, instead of just resetting the position it instead becomes completely unresponsive to user interaction. See the video below.

This was introduced in https://github.com/software-mansion/react-native-screens/pull/3173 - the Bluesky app uses v4.16.0 with just that one commit cherry-picked, and we're seeing the issue. I verified that it's still happening in v4.18.0 in the minimal repro

@knasher found a fix (see https://github.com/bluesky-social/social-app/pull/9369), but it's timeout-based - I'd hope there's a better way to fix it? I attempted a different approach in #3389 

https://github.com/user-attachments/assets/515a240e-5e29-4c33-a655-edc6506ddd91

Relevant issue: https://github.com/bluesky-social/social-app/issues/9320

### Steps to reproduce

1. Build the repro app (needs a dev client)
2. Open the video player in fullscreen mode
3. Swipe _down_ to exit fullscreen mode (works correctly)
4. Open again and swipe _up_ - this causes the app to become unresponsive to user input

https://github.com/mozzius/screens-video-repro

### Screens version

3.18.0

### React Native version

0.81.5

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo bare workflow

### Architecture

Fabric (New Architecture)

### Build type

Release mode

### Device

Real device

### Device model

iPhone 14

### Acknowledgements

Yes

mozzius:
## Description

Fixes #3388

In https://github.com/software-mansion/react-native-screens/pull/3173, interaction is disabled during screen transitions. However, when using the native fullscreen video player from `expo-video`, there is an edge case - you can cancel your swipe gesture, which causes user interaction to be disabled and never reenabled. The order of events is this:

- `willMoveToWindow` fires, disabling user interaction
- I cancel the swipe gesture by swiping up instead of down
- `willMoveToWindow` again, as we're moving back to the video player
- `viewDidAppear` does **not** fire, because nothing actually changed

Therefore user interaction is never reenabled, even though the transition is over

## Changes

> [!CAUTION]
> This was vibecoded. I have done my best to verify the code.

My fix is to track the current window, and in `didMoveToWindow` check if the movement is from `nil` to a window, reenabling user interaction if so. This appears to cover the dismiss case.

I am no expert here and I don't trust Claude, so please correct me if I'm wrong, there's a good chance I am. However, this does appear to fix the issue both in my minimal repro and in the main app.

### Before

https://github.com/user-attachments/assets/5f82e6ce-3af1-4e5d-923d-0016dd987572

### After

https://github.com/user-attachments/assets/0866dece-e276-4633-a69d-b34b6b6e5fc7

My repro is here: https://github.com/mozzius/screens-video-repro/tree/samuel/patch-fix

I patched in this change in this branch, so you can see the fix working: https://github.com/mozzius/screens-video-repro/tree/samuel/patch-fix

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -75,6 +75,7 @@ struct ContentWrapperBox {
   ContentWrapperBox _contentWrapperBox;
   bool _sheetHasInitialDetentSet;
   BOOL _shouldUpdateScrollEdgeEffects;
+  __weak UIWindow *_previousWindow;
 #ifdef RCT_NEW_ARCH_ENABLED
   RCTSurfaceTouchHandler *_touchHandler;
   react::RNSScreenShadowNode::ConcreteState::Shared _state;
@@ -739,6 +740,15 @@ RNS_IGNORE_SUPER_CALL_END
 
 - (void)didMoveToWindow
 {
+  if (@available(iOS 26, *)) {
+    // If we moved from nil to a window, that means a dismiss gesture was cancelled
+    // Re-enable interactions that were disabled in willMoveToWindow
+    if (self.window != nil && _previousWindow == nil) {
+      self.window.userInteractionEnabled = true;
+    }
+    _previousWindow = nil;
+  }
+
   // For RN touches to work we need to instantiate and connect RCTTouchHandler. This only applies
   // for screens that aren't mounted under RCTRootView e.g., modals that are mounted directly to
   // root application window.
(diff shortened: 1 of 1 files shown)


xieweizhi:
This will fix https://github.com/software-mansion/react-native-screens/issues/3390, same issue.

kmichalikk:
Superseded by #3311

PR #3393, opened 2025-11-13
fix(iOS): fix header height in modal, add TestHeaderHeight

kligarski:
## Description

Fixes header height in modal with `headerShown: true`. Prior to this PR, header height was sometimes incorrect and did not update dynamically, e.g. when search bar was enabled with `hideWhenScrolling: true`.

Adds `TestHeaderHeight` test screen.

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/2fb62d71-91d8-41d1-9e21-d86e4fa8c19f" /> | <video src="https://github.com/user-attachments/assets/7e2b5431-fbc1-484a-9eb1-8bfcfab65f66" /> |

## Changes

- pass `onHeaderHeightChange` callback to correct `Screen`
- create `TestHeaderHeight`
- extract `LongText` to shared components directory

Run `TestHeaderHeight`. Set `headerTransparent: true`, `presentation: modal`, `content: scrollView`, `searchBarPlacement: stacked`. Scroll up to see the search bar. Note header height values being updated.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/apps/src/tests/TestHeaderHeight.tsx
+++ b/apps/src/tests/TestHeaderHeight.tsx
@@ -366,6 +366,6 @@ function HeaderHeightTabsWrapper() {
 }
 
 export default function App() {
-  // return <HeaderHeightTest />;
-  return <HeaderHeightTabsWrapper />;
+  return <HeaderHeightTest />;
+  // return <HeaderHeightTabsWrapper />;
 }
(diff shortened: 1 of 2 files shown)


kligarski:
> The `ScreenStackItem` changes make sense. What about the incorrect unit being reported from native side you told me on the meeting?
> 
> I recall that we send px from platform, while dp is expected in React code.

Issue with incorrect unit happens on Android and will be handled in separate PR.

Issue #3040, opened 2025-07-18
The useHeaderHeight hook doesnt work on Android?


haltesh:
### Description

Initially posted here: https://github.com/react-navigation/react-navigation/issues/12692.

I also realize the issue was already previously reported in this repo (https://github.com/software-mansion/react-native-screens/issues/2661), but I don't want to just post another reply there since the issue appears to be abandoned.

The `useHeaderHeight` hook is consistently returning a massively oversized value on Android. I have no idea if it behaves differently on iOS. I thought one of my other dependencies may be problematic, but I was easily able to replicate the behaviour in a very minimal App.

App.tsx

```tsx
import { useHeaderHeight } from '@react-navigation/elements';
import { NavigationContainer } from '@react-navigation/native';
import { createNativeStackNavigator } from '@react-navigation/native-stack';
import { Text, View } from 'react-native';

const Stack = createNativeStackNavigator();

export default function App() {
    return (
        <NavigationContainer>
            <Stack.Navigator screenOptions={{
                headerStyle: {
                    backgroundColor: 'green'
                },
            }}>
                <Stack.Screen name="Default" component={DefaultRoute} />
            </Stack.Navigator>
        </NavigationContainer>
    );
}

const DefaultRoute = () => {
    const height = useHeaderHeight();

    return <View style={{ height, backgroundColor: 'red' }}>
        <Text style={{ color: 'white' }}>{height}</Text>
    </View>
}
```

Result:

<img width="583" height="1058" alt="Image" src="https://github.com/user-attachments/assets/90f1d45e-e3ac-4512-b88a-b5c1bca51a85" />

I'd expect the green (Header) and red (using `useHeaderHeight`) to be the same height, or at least close to the same height

### Steps to reproduce

It as very easy to reproduce with a native stack navigator with a single route. Example here: Reproduction repo: https://github.com/haltesh/React-native-navigation-use-header-height.

### Snack or a link to a repository

Reproduction repo: https://github.com/haltesh/React-native-navigation-use-header-height.

### Screens version

3.16.0

### React Native version

0.79.5

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

Android emulator

### Device model

Google Pixel 3a (Android 14, emulator), Samsung A50 (Android 11, physical device)

### Acknowledgements

Yes

JustJoostNL:
[...]

Issue #2661, opened 2025-01-29
useHeaderHeight returns incorrect value for native stack on Android


ykliuiev:
### Description

I noticed that useHeaderHeight returns inconsistent values for native vs. simple stacks on Android. For example, native stack returns 246, whereas simple stack yields 95 as a value. This may be related to [Screen.kt#L164](https://github.com/software-mansion/react-native-screens/blob/main/android/src/main/java/com/swmansion/rnscreens/Screen.kt#L164), where notifyHeaderHeightChange(t) might be passing an unexpectedly large t value.

I verified this with a HeaderHeightView component in the react-navigation/example app, and the native and simple stacks show different values.

| Native Stack | Simple Stack |
|--------|--------|
| ![Image](https://github.com/user-attachments/assets/dd029b01-aba0-4a6c-a9f7-50a0c9c29dde) | ![Image](https://github.com/user-attachments/assets/8e7512c3-d034-4ddb-be5b-3dc0a54669f0) | 

### Steps to reproduce

It's reproducible in https://github.com/react-navigation/react-navigation/tree/main/example as well as in the Example app in this repo.
Calling useHeaderHeight in apps/src/screens/SimpleNativeStack.tsx will log two times:

import { useHeaderHeight } from '@react-navigation/elements';

const MainScreen = ({ navigation }: MainScreenProps): React.JSX.Element => {
  const header = useHeaderHeight();
  console.log('header', header);

logs:
header height: 56
header height: 126.711111

### Snack or a link to a repository

https://github.com/react-navigation/react-navigation/tree/main/example

### Screens version

4.3.0

### React Native version

0.76.2

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

Paper (Old Architecture)

### Build type

None

### Device

Real device

### Device model

Samsung Galaxy A53 Android 14, Google Pixel 6 Pro Android 15

### Acknowledgements

Yes

faljabi:
I am facing a similar issue where a segmented control for example is not clickable if it is right below the header on Android. If I padding top the segmented control say 64px it is clickable. My guess is that something from the header has extra height on Android covering content. This is happening only when I set headerRight with a close button for example.

No issues on iOS.

faljabi:
Hi,

Is this solved?

davidjbng:
This might be caused by #2758.
There is a fix for it in the comments. Haven't verified this myself though.
[...]

PR #3402, opened 2025-11-17
fix(Android, Stack): incorrect header height values

kligarski:
## Description

Fixes header height values reported via `onHeaderHeightChange` event.

Fixes https://github.com/software-mansion/react-native-screens/issues/3040.
Fixes https://github.com/software-mansion/react-native-screens/issues/2661.

> [!NOTE]
>
> Due to incorrect values being reported previously, there are some workarounds in `react-navigation`'s `NativeStackView.native.tsx`. Those should be redundant now and we should clean-up that code when this PR is merged. We intend to handle backward compatibility through a flag. PR in `react-navigation`: https://github.com/react-navigation/react-navigation/pull/12852.

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/b5d2a967-f9d8-480a-9c28-41309f7b8f03" /> | <video src="https://github.com/user-attachments/assets/476a044a-352e-42fa-b67a-8d72a4a6f795" /> |

## Changes

- report height of the header instead of Screen's top offset
- convert px to dp
- update `TestHeaderHeight` to allow testing dynamic changes to the prop and `react-navigation`'s `useHeaderHeight()` hook

Run `TestHeaderHeight`.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderHeightUpdateProxy.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderHeightUpdateProxy.kt
@@ -3,7 +3,10 @@ package com.swmansion.rnscreens
 class ScreenStackHeaderHeightUpdateProxy {
     var previousHeaderHeightInPx: Int? = null
 
-    fun updateHeaderHeightIfNeeded(config: ScreenStackHeaderConfig, screen: Screen?) {
+    fun updateHeaderHeightIfNeeded(
+        config: ScreenStackHeaderConfig,
+        screen: Screen?,
+    ) {
         val currentHeaderHeightInPx = if (config.isHeaderHidden) 0 else config.toolbar.height
 
         if (currentHeaderHeightInPx != previousHeaderHeightInPx) {


android/src/main/java/com/swmansion/rnscreens/utils/DimensionUtils.kt
kkafar, line 1: No, let's not define our custom utility. There is `PixelUtil` from ReactAndroid & we should use that.

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderHeightUpdateProxy.kt
```diff
@@ -0,0 +1,16 @@
+package com.swmansion.rnscreens
+
+class ScreenStackHeaderHeightUpdateProxy(
+    val config: ScreenStackHeaderConfig,
```
kkafar: This introduces retain cycle & memory leak, between the proxy & header config.

Instead of keeping the header config in state here, pass an instance (or just necessary values) as method parameters.

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.kt
```diff
@@ -162,7 +166,7 @@ class ScreenStackHeaderConfig(
             ?.dispatchEvent(HeaderDetachedEvent(surfaceId, id))
     }
 
-    private val screen: Screen?
```
kkafar, line 165: make it at least `internal`

kkafar: also, I'd just prefer it being passed as param where needed.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -456,12 +454,18 @@ class Screen(
             super.onTouchEvent(event)
         }
 
-    private fun notifyHeaderHeightChange(headerHeight: Int) {
+    fun notifyHeaderHeightChange(headerHeight: Int) {
```
kkafar: internal

kkafar (changes requested):
Thanks for the PR. I've left few remarks.

android/src/main/java/com/swmansion/rnscreens/utils/DimensionUtils.kt
kligarski: Thanks, I forgot about that.

https://github.com/software-mansion/react-native-screens/pull/3402/commits/1c7e882f31d08801e83edb256b912821536706c2

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderHeightUpdateProxy.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3402/commits/1c7e882f31d08801e83edb256b912821536706c2

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3402/commits/1c7e882f31d08801e83edb256b912821536706c2

android/src/main/java/com/swmansion/rnscreens/Screen.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3402/commits/1c7e882f31d08801e83edb256b912821536706c2

PR #3404, opened 2025-11-18
fix(Android, Stack): Constraint FormSheet height by top inset

t0maboro:
## Description

Add a new prop to control formSheet overflow behavior relative to the status bar.

This PR introduces a new prop that allows to control whether a formSheet should overflow the status bar and display cutouts or not. The purpose of this addition is to align the formSheet behavior more closely with iOS, where the top safe area inset is respected by default.

> [!CAUTION]
> This is a breaking change that has an impact on the sheet's size (and indirectly may lead to significant visual changes). To bring back the original behavior, you can set `sheetShouldOverflowTopInset` prop to `true`

Fixes https://github.com/software-mansion/react-native-screens-labs/issues/565

## Changes

- Added a class for coordinating the bottom sheet slide animation, to be started after we receive insets.
- Added a class to aggregate listeners for insets on bottom sheet level
- Added sheetShouldOverflowTopInset prop

## Screenshots / GIFs

#### With overflow

https://github.com/user-attachments/assets/4d6933b4-6082-4a83-b33a-033727dbd0ad

#### Without overflow

https://github.com/user-attachments/assets/b278db92-c28a-4485-a0a5-ab4301269ee0

Test with 3336 with API levels over and below 30

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
@@ -144,9 +144,7 @@ class SheetDelegate(
                 else -> null
             }
 
-        behavior.apply {
-            updateMetrics(maxAllowedHeight, expandedOffsetFromTop)
-        }
+        behavior.updateMetrics(maxAllowedHeight, expandedOffsetFromTop)
     }
 
     internal fun configureBottomSheetBehaviour(


apps/src/tests/Test3336.tsx
```diff
@@ -129,6 +129,7 @@ const formSheetBaseOptions: NativeStackNavigationOptions = {
   contentStyle: {
     backgroundColor: Colors.GreenLight100,
   },
+  // TODO(@t0maboro) - add `sheetShouldOverflowStatusBar` prop here when possible
```
kligarski: That'd be nice.

src/types.tsx
```diff
@@ -472,6 +472,17 @@ export interface ScreenProps extends ViewProps {
...
+   * Whether the sheet content should be rendered behind the Status Bar
+   *
+   * When set to `true`, the sheet will extend to the physical edges of the screen, allowing the content to be visible behind the status bar.
+   * When set to `false`, the sheet's layout will be constrained by the system insets from the top.
+   *
+   * Defaults to `true`.
+   *
+   * @platform android
+   */
+  sheetShouldOverflowStatusBar?: boolean;
```
kligarski: I think that we should highlight in the description that with `sheetShouldOverflowStatusBar: false`, ratios from `sheetAllowedDetents` will be relative to safe area, not the container height (e.g. if you set `sheetAllowedDetents: [0.5]`, the height (relative to window) will be different depending on `sheetShouldOverflowStatusBar` prop).

Another question is whether this is the behavior we want.

apps/src/tests/Test3336.tsx
t0maboro: requires changes in react-navigation, or a dedicated test, idk which approach we should follow here, but let me create a ticket as a reminder https://github.com/software-mansion/react-native-screens-labs/issues/568

src/types.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3404/commits/ff4d0db64f9da36323ec49ff8f2914229901f82d

t0maboro: > Another question is whether this is the behavior we want.

Imo it does make sense to calculate detent within the available sheet space. Like, having detent 1.0 and sheetShouldOverflowStatusBar: false would result in content cutoff if we'd calculate maxHeight as detent*containerHeight

android/src/main/java/com/swmansion/rnscreens/bottomsheet/BottomSheetTransitionCoordinator.kt
```diff
@@ -0,0 +1,135 @@
...
+import androidx.core.view.WindowInsetsCompat
+import com.swmansion.rnscreens.Screen
+import com.swmansion.rnscreens.ScreenStackFragment
+
+class BottomSheetTransitionCoordinator {
+    private var isLayoutComplete = false
+    private var areInsetsApplied = false
+
+    private var lastInsets: WindowInsets? = null
+    private var lastInsetsCompat: WindowInsetsCompat? = null
```
kkafar: Why store both? Wouldn't it be easier to maintain if we stored only one of them & converted to second ones when needed?

android/src/main/java/com/swmansion/rnscreens/bottomsheet/BottomSheetTransitionCoordinator.kt
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        // Although the layout of the screen container and CoordinatorLayout hasn't changed,
+        // the BottomSheetBehavior has updated the maximum height.
+        // We manually trigger the callback to notify that the bottom sheet layout has been applied.
+        screen.onBottomSheetBehaviorDidLayout(true)
+
+        areInsetsApplied = true
+        triggerSheetEnterTransitionIfReady(screen)
+    }
```
kkafar: I don't like the fact that transition coordinator applies the insets to the sheet. 

I'll come back with recommendation in followup review.

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
```diff
@@ -233,6 +233,16 @@ corresponding legacy prop values for `sheetAllowedDetents` prop.
 
 Defaults to `none`, indicating that the dimming view should be always present.
 
+### `sheetShouldOverflowStatusBar` (Android only)
+
+Whether the sheet content should be rendered behind the Status Bar.
+
+When set to `true`, the sheet will extend to the physical edges of the screen, allowing content to be visible behind the status bar. Detent ratios in sheetAllowedDetents will be measured relative to the full screen height.
```
kligarski:
```suggestion
When set to `true`, the sheet will extend to the physical edges of the screen, allowing content to be visible behind the status bar. Detent ratios in sheetAllowedDetents will be measured relative to the full stack height.
```

Applies to other places as well.

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3404/commits/8f31dacdedf2e954239f934bcdf099570878a0b6

android/src/main/java/com/swmansion/rnscreens/bottomsheet/BottomSheetTransitionCoordinator.kt
kkafar: IMO it's responsibilities should be restricted to waiting for appropriate moment to start the transition & start it.

android/src/main/java/com/swmansion/rnscreens/bottomsheet/BottomSheetTransitionCoordinator.kt
```diff
@@ -0,0 +1,135 @@
...
+            // Needs to be highlighted that nothing changes at the container level.
+            // However, calling additional measure will trigger BottomSheetBehavior's `onMeasureChild` logic.
+            // This method ensures that the bottom sheet respects the maxHeight we update in `configureBottomSheetBehavior`.
+            coordinatorLayout.forceLayout()
+            coordinatorLayout.measure(
+                View.MeasureSpec.makeMeasureSpec(container.width, View.MeasureSpec.EXACTLY),
+                View.MeasureSpec.makeMeasureSpec(container.height, View.MeasureSpec.EXACTLY),
+            )
+            coordinatorLayout.layout(0, 0, container.width, container.height)
+        }
```
kkafar: We take coordinatorLayout as the parameter. Maybe we could use it directly here? Also if it's supposed to be coordinator layout we should give appropriate type to method parameter.

android/src/main/java/com/swmansion/rnscreens/bottomsheet/BottomSheetWindowInsetListenerChain.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    override fun onApplyWindowInsets(
+        v: View,
+        insets: WindowInsetsCompat,
+    ): WindowInsetsCompat {
+        var currentInsets = insets
+
+        for (listener in listeners) {
+            currentInsets = listener.onApplyWindowInsets(v, currentInsets)
+        }
```
kkafar, line 32: Having it written this way, the order of listeners actually matters, which contradicts the class doc comment. We have here "the rolling insets" version here. Is this intentional?

src/types.tsx
kkafar: Yep. + this way I believe we keep alignment between platforms, which is nice when possible

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
```diff
@@ -233,6 +233,16 @@ corresponding legacy prop values for `sheetAllowedDetents` prop.
...
 
+### `sheetShouldOverflowStatusBar` (Android only)
+
+Whether the sheet content should be rendered behind the Status Bar.
+
+When set to `true`, the sheet will extend to the physical edges of the stack, allowing content to be visible behind the status bar. Detent ratios in sheetAllowedDetents will be measured relative to the stack height.
+
+When set to `false`, the sheet's layout will be constrained by the status bar insets from the top and the detent ratios will then be measured relative to the adjusted height (excluding the top inset). This means that sheetAllowedDetents will result in different sheet heights depending on this prop.
+
+Defaults to `true`.
```
kkafar: I'm in two minds regarding the default value here. 

There are a couple of things to consider here: 

1. what's the previous behaviour, so that we don't introduce breaking changes,
2. alignment between platforms,
3. reasonable default.

I think that the previous state is that the sheet does expand to full screen height, so that it does not respect the top insets. 

As per alignment - iOS does the opposite, it respects the insets (cutout, status bar, etc.).

As per reasonable default - I think it is to respect the insets.

IMO we could treat this as a fix/improvement rather than behaviour change and make it aligned with iOS.

t0maboro: > I think that the previous state is that the sheet does expand to full screen height, so that it does not respect the top insets.

Correct

> As per alignment - iOS does the opposite, it respects the insets (cutout, status bar, etc.).

Correct

> As per reasonable default - I think it is to respect the insets.

50/50 😄 

The only argument I have is to keep the compatibility, because that change will have some impact on the appearance in existing apps, which are having 1.0 detent. Not a strong preference, aligning with iOS seems to be a reasonable approach either.

android/src/main/java/com/swmansion/rnscreens/bottomsheet/BottomSheetTransitionCoordinator.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3404/commits/85fee02f0d1ed20c4e084391be30f43281ee101a

android/src/main/java/com/swmansion/rnscreens/bottomsheet/BottomSheetTransitionCoordinator.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3404/commits/be0568c2b526937c775056f0f77c38b42f76a365

android/src/main/java/com/swmansion/rnscreens/bottomsheet/BottomSheetTransitionCoordinator.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3404/commits/bbf6cc92d21dfeb515e702ca6ddefa426c60c57c

android/src/main/java/com/swmansion/rnscreens/bottomsheet/BottomSheetWindowInsetListenerChain.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3404/commits/467470a316c9dee939d854a5ab6ba258cf450247

guides/GUIDE_FOR_LIBRARY_AUTHORS.md
t0maboro: I'm proposing this https://github.com/software-mansion/react-native-screens/pull/3404/commits/5cc93727e7c4b188a61dbe0156d4148aea7f8b9c

t0maboro:
> Looks good overall. I have few comments / requests.
> 
> Looking at the final comment from the review: https://github.com/software-mansion/react-native-screens/pull/3404/files#r2545394786 (read first please)
> 
> I think that we should treat this as a fix & make the sheets respect insets by default.
> 
> Therefore we should work on naming. First, it can not be `overflowStatusBar`, since it is too specific and not entirely true, right? Doesn't this PR also handle display cutout? Can you verify this? How does it look when there is a big display cutout present (you can turn it on in emulator settings (google for it please))?
> 
> If so, we should rather refer to it as `topInset` IMO.
> 
> Additionally if the default's gonna be `false`, then we should phrase it more like `sheetShouldIgnoreTopInset`. Then `false` as the deafult value & if someone wants to active the prop he sets it to true.
> 
> What do you think?

I wasn't sure about the support for display cutout, but added it here: https://github.com/software-mansion/react-native-screens/pull/3404/commits/25a543bee6bab7474e756450387f983fee391571 . Then here: https://github.com/software-mansion/react-native-screens/pull/3404/commits/5cc93727e7c4b188a61dbe0156d4148aea7f8b9c I'm proposing the new name for the prop.

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
```diff
@@ -312,6 +354,12 @@ class SheetDelegate(
         val isImeVisible = insets.isVisible(WindowInsetsCompat.Type.ime())
         val imeInset = insets.getInsets(WindowInsetsCompat.Type.ime())
         val prevSystemBarsInsets = insets.getInsets(WindowInsetsCompat.Type.systemBars())
+        val prevDisplayCutoutInsets = insets.getInsets(WindowInsetsCompat.Type.displayCutout())
+
+        // We save the top inset (status bar height + display cutout) to later
+        // subtract it from the window height during sheet size calculations.
+        // This ensures the sheet respects the safe area.
+        lastTopInset = prevSystemBarsInsets.top + prevDisplayCutoutInsets.top
```
kkafar: Should this be summed here? Haven't verified it, but my thoughts are that we should take `max` of the two values here. Otherwise we might end up wasting too much space. What's the reasoning behind adding the values here?

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
```diff
@@ -312,6 +354,12 @@ class SheetDelegate(
         val isImeVisible = insets.isVisible(WindowInsetsCompat.Type.ime())
         val imeInset = insets.getInsets(WindowInsetsCompat.Type.ime())
         val prevSystemBarsInsets = insets.getInsets(WindowInsetsCompat.Type.systemBars())
+        val prevDisplayCutoutInsets = insets.getInsets(WindowInsetsCompat.Type.displayCutout())
```
kkafar: Why do these variables have `prev` prefix? How are these previous values? What are current values then?

kkafar (commented):
I have two more remarks. I'll come back to this PR tomorrow with few more questions.

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
t0maboro: okay, I might have misinterpreted how it works, especially for `LAYOUT_IN_DISPLAY_CUTOUT_MODE_ALWAYS` definition, I was thinking that it'd do a whole content cutout, so the sum seemed to be safer, but you're right that max is proper here https://developer.android.com/develop/ui/views/layout/display-cutout#handle

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3404/commits/e40916f086e2bca06cc2b9371202e61877749440

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3404/commits/30baf69829c601539b4dc4005b3ffbb9b7296a2b

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
```diff
@@ -113,12 +114,47 @@ class SheetDelegate(
...
+        // has been reduced by this inset.
+        val expandedOffsetFromTop =
+            when (screen.sheetDetents.count) {
+                3 -> screen.sheetDetents.expandedOffsetFromTop(containerHeight, lastTopInset)
+                else -> null
+            }
+
+        behavior.apply {
+            updateMetrics(maxAllowedHeight, expandedOffsetFromTop)
+        }
```
kkafar:
```suggestion
        behavior.updateMetrics(maxAllowedHeight, expandedOffsetFromTop)
```

PR #3408, opened 2025-11-20
chore: change default bottom tabs implementation to managed

kligarski:
## Description

Changes default bottom tabs implementation to managed.

Bottom tabs offer 2 ways of controlling the container:
- managed/natively-driven,
- controlled/programmatically-driven.

Most of the libraries consuming `react-native-screens` use managed tabs: [expo-router](https://github.com/expo/expo/blob/a5b3082932187d54ef3311dee21462dc07d86402/apps/expo-go/src/navigation/NativeBottomTabsNavigator.tsx#L41) and [react-navigation](https://github.com/react-navigation/react-navigation/blob/4568ec61decebf42f63e3d3b710d21b33d698846/packages/bottom-tabs/src/unstable/NativeBottomTabView.native.tsx#L150). 

## Changes

- change `RNS_CONTROLLED_BOTTOM_TABS_DEFAULT` to `false`

Run `TestBottomTabs`. Verify that managed implementation is used (use heavy render and switch between Tab1 and Tab2, selected tab should change immediately).

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/components/bottom-tabs/BottomTabs.types.ts
+++ b/src/components/bottom-tabs/BottomTabs.types.ts
@@ -264,7 +264,8 @@ export interface BottomTabsProps extends ViewProps {
    * On iOS, some features are not fully implemented for managed tabs
    * (e.g. overrideScrollViewContentInsetAdjustmentBehavior).
    *
-   * On Android, only controlled tabs are currently supported.
+   * On Android, only controlled tabs are currently supported and the
+   * value of this prop is ignored.
    *
    * @default Defaults to `false`.
    *


src/flags.ts
```diff
@@ -1,4 +1,4 @@
-const RNS_CONTROLLED_BOTTOM_TABS_DEFAULT = true;
+const RNS_CONTROLLED_BOTTOM_TABS_DEFAULT = false;
```
kmichalikk, line 1: did you check for places to update the documentation?

kligarski: It was already matching fabric spec which uses `false` as default but I added some more explanation how it works on Android.

https://github.com/software-mansion/react-native-screens/pull/3408/commits/ef0919df141781515a4a58cfdf06276ed578c2f4

kmichalikk (approved):

PR #3411, opened 2025-11-20
fix(iOS, Tabs): add support for bottom accessory environment none

johankasperi:
## Description

When closing a bottomaccessory you get a RCTLogError with "Unsupported environment for onEnvironmentChange event". Thats because the environment is of type ["none"](https://developer.apple.com/documentation/uikit/uitabaccessory/environment/none) when the bottom accessory is closed.

## Changes

Add support for "none".

### Before

https://github.com/user-attachments/assets/92c04ce5-742c-4426-8354-595918e95132

### After

Uploading Simulator Screen Recording - iPhone 17 Pro - 2025-11-20 at 16.10.15.mov…

Run Test3288 and set shown to false using the button on the screen

## Checklist

- [x] Updated TS types
- [ ] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSEnums.h
+++ b/ios/RNSEnums.h
@@ -184,5 +184,6 @@ typedef NS_ENUM(NSInteger, RNSOptionalBoolean) {
 
 typedef NS_ENUM(NSInteger, RNSBottomTabsAccessoryEnvironment) {
   RNSBottomTabsAccessoryEnvironmentRegular,
-  RNSBottomTabsAccessoryEnvironmentInline
+  RNSBottomTabsAccessoryEnvironmentInline,
+  RNSBottomTabsAccessoryEnvironmentNone
 };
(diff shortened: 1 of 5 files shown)


kkafar:
Thanks @johankasperi!

cc @kligarski 

Will we need to update workarounds code?

kligarski:
I'm not sure if we need `none` environment. Currently, the only way to hide the bottom accessory is to set `bottomAccessory` prop in `BottomTabs` to `undefined`, the user doesn't really need an event for their own action, right? I'm not sure if there is any other situation where `none` environment is used.

The warning is just something I missed during one of the refactors in https://github.com/software-mansion/react-native-screens/pull/3288. What I meant to do there is to ignore any events other that `inline` and `regular`, it shouldn't be a problem to fix that.

What do you think, @johankasperi, @kkafar?

johankasperi:
> I'm not sure if we need `none` environment. Currently, the only way to hide the bottom accessory is to set `bottomAccessory` prop in `BottomTabs` to `undefined`, the user doesn't really need an event for their own action, right? I'm not sure if there is any other situation where `none` environment is used.
> 
> The warning is just something I missed during one of the refactors in #3288. What I meant to do there is to ignore any events other that `inline` and `regular`, it shouldn't be a problem to fix that.
> 
> What do you think, @johankasperi, @kkafar?

Yeah I considered that as well but I didn't want to remove the warning because I thought it might was needed. I agree that the `none` event isn't necessary (at least in my use case) so I'm fine by just removing the warning.

kligarski:
Superseded by https://github.com/software-mansion/react-native-screens/pull/3416.

kkafar:
Let's go in that direction than. Thanks!

PR #3412, opened 2025-11-20
chore(e2e): detox config general cleanup and ios management

KrzysztofWojnar:
## Description

This is first of three PRs in the scope of [labs issue 569](https://github.com/software-mansion/react-native-screens-labs/issues/569).

Realizing plans from [labs discussion 572](https://github.com/software-mansion/react-native-screens-labs/discussions/572).

## Changes

e2e tests could be run in more parametrized way - choosing ios version/device before the start.
CI_AVD_NAME and DETOX_AVD_NAME env variables combined into `RNS_E2E_AVD_NAME`
`RNS_ADB_NAME` environment variable introduced for local testing against attached [physical] devices.
Common e2e scripts restructured.

## Screenshots / GIFs

### Before

Choosing apple simulator by changing config file for "ios.simulator" entry

### After

Running test script with optional `RNS_E2E_APPLE_SIM_NAME` and `RNS_E2E_IOS_VERSION` environment variables.

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/scripts/e2e/detox-utils.cjs
+++ b/scripts/e2e/detox-utils.cjs
@@ -6,14 +6,14 @@ const DEFAULT_CI_AVD_NAME = 'e2e_emulator';
 
 const isRunningCI = process.env.CI != null;
 
-// Assumes that local developement is done on arm64-v8a.
+// Assumes that local development is done on arm64-v8a.
 const apkBulidArchitecture = isRunningCI ? 'x86_64' : 'arm64-v8a';
 // test-butler requires AOSP emulator image, which is not available to download for arm64-v8a in Android Studio SDK Manager, therefore
 // it is assumed here that arm64-v8a AOSP emulator is not available in local setup.
 const testButlerApkPath = isRunningCI ? ['../Example/e2e/apps/test-butler-app-2.2.1.apk'] : undefined;
 
 function detectLocalAndroidEmulator() {
-  // "RNS_E2E_AVD_NAME" can be set for local developement
+  // "RNS_E2E_AVD_NAME" can be set for local development
   const avdName = process.env.RNS_E2E_AVD_NAME ?? null;
   if (avdName !== null) {
     return avdName
@@ -45,7 +45,7 @@ function detectLocalAndroidEmulator() {
 }
 
 function detectAndroidEmulatorName() {
-  // "RNS_E2E_AVD_NAME" can be set for local developement
+  // "RNS_E2E_AVD_NAME" can be set for local development
   return isRunningCI ? DEFAULT_CI_AVD_NAME : detectLocalAndroidEmulator();
 }
 
--- a/scripts/e2e/ios-devices.js
+++ b/scripts/e2e/ios-devices.js
@@ -1,5 +1,5 @@
 const DEFAULT_APPLE_DEVICE = 'iPhone 17';
-const DEFEAULT_IOS_VERSION = 'iOS 26.2';
+const DEFAULT_IOS_VERSION = 'iOS 26.2';
 
 /**
  * @return {string}
@@ -29,7 +29,7 @@ function getIOSVersion() {
       throw new Error(`Environment variable ${envVariableKey} should be "iOS xyz".`);
     }
   }
-  return DEFEAULT_IOS_VERSION;
+  return DEFAULT_IOS_VERSION;
 }
 
 /**


scripts/e2e/detox-utils.cjs
```diff
@@ -1,6 +1,7 @@
 const ChildProcess = require('node:child_process');
+const { iosDevice } = require('./ios-devices');
 
-const CI_AVD_NAME = 'e2e_emulator';
+const CI_ADB_NAME = 'e2e_emulator';
```
kkafar: Writing from top of my head but:

`ADB` ~= Android Debug Bridge,
`AVD` ~= Android Virtual Device.

I believe this variable describes the name of the emulator (avd) we want to use -> avd makes more sense here.  What do you think?

scripts/e2e/ios-devices.js
```diff
@@ -0,0 +1,37 @@
+const DEFEAULT_APPLE_MODEL = 'iPhone 17';
+const DEVICE_MODEL = process.env.E2E_APPLE_DEVICE || DEFEAULT_APPLE_MODEL
```
kkafar: `DEVICE_MODEL` is not really a constant here. I'd prefer it being extracted to a function and computed there.

kkafar: Similarly to `detectLocalAndroidEmulator` function defined in the "main" util module.

scripts/e2e/ios-devices.js
```diff
@@ -0,0 +1,37 @@
+const DEFEAULT_APPLE_MODEL = 'iPhone 17';
```
kkafar:
```suggestion
const DEFAULT_APPLE_MODEL = 'iPhone 17';
```

just a typo

kkafar: Also would be nice to have consistent naming convention here. If we change the name of the env var, this should be renamed accordingly.

scripts/e2e/ios-devices.js
```diff
@@ -0,0 +1,37 @@
...
+const DEVICE_MODEL = process.env.E2E_APPLE_DEVICE || DEFEAULT_APPLE_MODEL
+const DEFEAULT_IOS_VERSION = 'iOS 26.2';
+
+/**
+ * @return {`iOS ${string}`} requested version of ios, or default if not specified
+ */
+function getiosVersion() {
```
kkafar:
```suggestion
function getIOSVersion() {
```

kkafar (changes requested):
Hey! Nice job. I have few remarks & questions. Please answer them.

---

According to https://github.com/software-mansion/react-native-screens-labs/discussions/572 should I expect support for `E2E_ANDROID_API_LEVEL` in this PR or what are plans for that? 

--- 

If you want to structure the code this way, i. e. extract the "device" detecting code to separate module, then we should do analogous thing with Android related code. It can be done in separate PR, I don't mind, but let's create ticket for this, so that we do not forget.

scripts/e2e/detox-utils.cjs
KrzysztofWojnar: Yes, you are right
when I was doing the cleanup I was sure that we assign it to "adbName" in detox-utils, but we assign it to "avdName" in this case, I'm going to correct that

KrzysztofWojnar: Oh, no,
This specific variable seems to be used in two places and it looks to me that one case is wrong:
We used it as a `devices.attached.device.adbName` in line 104->102 and it may be assigned to `devices.emulator.device.avdName` because of line 46->47

I suspect one of these entries was a dead code.

scripts/e2e/ios-devices.js
KrzysztofWojnar: For me, it makes no difference (okay, now that you've pointed it out, I'd write it camelCase).

Anyway, turned into a function

scripts/e2e/detox-utils.cjs
kkafar: The name we specify is clearly an AVD name, not ADB name. Therefore let's name it appropriately, please. 

I took a look into [Detox docs](https://wix.github.io/Detox/docs/19.x/config/overview/#device-configurations) and it seems that two fields are allowed:

* `avdName` - for emulator name as listed by `emulators -list-avd`
* `adbName` - for emulator device hash (? not sure what that hash-like id is) that can be used with ADB to identify the device.

Seems that for the wole time we've been just using the wrong field, unless the e2e CI workflow specifies `e2e_emulator` as the adb name of the device it creates. 

PS: I've just checked and it seems that we specify [`avd-name` field there](https://github.com/software-mansion/react-native-screens/blob/172a7bbda9ac4563b8e1cc4ed4507a121081715e/.github/workflows/android-e2e-test-fabric.yml#L89). Not sure of it exact semantics, but it seems that our detox setup worked so far only by accident. **Let's fix that**.

KrzysztofWojnar: ok, I'll replace CI_AVD_NAME and DETOX_AVD_NAME with RNS_E2E_AVD_NAME (from https://github.com/software-mansion/react-native-screens-labs/discussions/572#discussioncomment-15031761)

.github/workflows/android-e2e-test-fabric.yml
```diff
@@ -86,7 +86,7 @@ jobs:
           disable-animations: false
           force-avd-creation: false
           emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
-          avd-name: e2e_emulator
+          rns_e2e_avd_name: e2e_emulator
```
kkafar: `avd-name` is a configuration option from the [github action this job step uses](https://github.com/ReactiveCircus/android-emulator-runner?tab=readme-ov-file#configurations). Renaming here does not make much sense. I think you went too far with find & replace 😅

kkafar (changes requested):
I still have some more remarks. I feel like we're proceeding here in a bit chaotic manner, with some unrequested changes being applied here.

I'll open a diff with suggested changes, so that we can proceed more smoothly. Will link it here once a do

scripts/e2e/detox-utils.cjs
```diff
@@ -32,22 +27,24 @@ function detectLocalAndroidEmulator() {
...
     throw new Error(errorMessage);
   }
 }
 
 function detectAndroidEmulatorName() {
-  return isRunningCI ? CI_AVD_NAME : detectLocalAndroidEmulator();
+  // "RNS_E2E_AVD_NAME" can be set for local developement
+  return process.env.RNS_E2E_AVD_NAME || isRunningCI
+    ? DEFAULT_CI_AVD_NAME
+    : detectLocalAndroidEmulator();
```
kkafar: I have few remarks to this code:

1. It introduces unnecessary uncertainty how to interpret this. Might be a skill issue on my end, but I prefer to have it written in such way I don't have to wonder whether `||` operator has higher precedence than ternary `?:` operator in JavaScript or not:
```javascript
(process.env.RNS_E2E_AVD_NAME || isRunningCI)
    ? DEFAULT_CI_AVD_NAME
    : detectLocalAndroidEmulator();
```
or 
```javascript
process.env.RNS_E2E_AVD_NAME || (isRunningCI
    ? DEFAULT_CI_AVD_NAME
    : detectLocalAndroidEmulator());
```
?

2. Why do we change this code at all here? From all I see we haven't started passing emulator name from CI through env, it is still hard coded name, therefore there is no need to change this code like that.

KrzysztofWojnar: I confess, I assumed we were going to pass this value all the time, but we used hardcoded CI_AVD_NAME as a fallback due to a misconfig 😅 but that part was fine all along. That's also the reason I touched those github configs.
I'm withdrawing this now

KrzysztofWojnar: fixed with commit 3fbdad36fdf853744340054bd346679be36d1ae3

kkafar:
I've posted suggested review changes here: https://github.com/software-mansion/react-native-screens/pull/3414

scripts/e2e/detox-utils.cjs
KrzysztofWojnar: ... and replaced with solution from e4074bd79023c1d1b3e7eee788db0db0c010776a and 8c2044b3c53e1913dedd76c17ca60d3e0a8f8f66

scripts/e2e/ios-devices.js
```diff
@@ -0,0 +1,41 @@
...
+ */
+function getAppleDevice() {
+  return process.env.RNS_E2E_APPLE_SIM_NAME || DEFAULT_APPLE_DEVICE;
+}
+/**
+ * @return {`iOS ${string}`} requested version of ios, or default if not specified
+ */
+function getIOSVersion() {
+  const passedVersion = process.env.RNS_E2E_IOS_VERSION;
+  if (passedVersion) {
```
t0maboro: I assume that the passed version should have some specific format, so I'd add some documentation and additional validation that it matches the proper regex

scripts/e2e/ios-devices.js
```diff
@@ -0,0 +1,41 @@
...
+    if (passedVersion.startsWith('iOS ')) {
+      return /** @type {`iOS ${string}`} */ (passedVersion);
+    }
+    return `iOS ${passedVersion}`;
+  }
+  return DEFEAULT_IOS_VERSION;
+}
+
+/**
+ * @typedef {Object} AppleDevice - creates a new type named 'SpecialType'
```
t0maboro: `creates a new type named 'AppleDevice'`?

scripts/e2e/ios-devices.js
```diff
@@ -0,0 +1,41 @@
+const DEFAULT_APPLE_DEVICE = 'iPhone 17';
+const DEFEAULT_IOS_VERSION = 'iOS 26.2';
+
+/**
+ * @return {string}
+ */
+function getAppleDevice() {
+  return process.env.RNS_E2E_APPLE_SIM_NAME || DEFAULT_APPLE_DEVICE;
```
t0maboro, line 17: We may consider adding some extra validation and define the types of devices that we're supporting for e2e automation, e.g. for now I'm seeing a use-case for passing `iPhone` or `iPad`, but when someone passes `watchOS`, we should have some error handling prepared

scripts/e2e/detox-utils.cjs
```diff
@@ -1,6 +1,8 @@
 const ChildProcess = require('node:child_process');
+const { iosDevice } = require('./ios-devices');
 
-const CI_AVD_NAME = 'e2e_emulator';
+// Should be kept in sync with the constant defined in e2e workflow file
+const DEFAULT_CI_AVD_NAME = 'e2e_emulator';
```
t0maboro, line 5: nit: would it be possible to get this value automatically? This way, we wouldn't need to remember to synchronize it? I'm seeing how it was working prior to this PR, so I'd consider it only as an improvement, which might be done in a separate task, but it's not a high priority and won't block this PR from merging.

scripts/e2e/ios-devices.js
KrzysztofWojnar: configuration will be passed to Detox and I planned to rely on validation in Detox itself. However, now I think you're right, because for as long as I can remember, iPhones have had consistent naming conventions and I _can expect_ them to maintain it.

scripts/e2e/ios-devices.js
KrzysztofWojnar: I admit, I forgot the "typedef" keyword and asked AI, then copy-pasted the line and adjusted the type name only 😅

scripts/e2e/ios-devices.js
KrzysztofWojnar: okay I can add it
btw I wont be "parsing" iPhone/iPad "numbers" due to exceptions like iPhone X or iPhone SE

scripts/e2e/detox-utils.cjs
KrzysztofWojnar: It's rather static, anyway that's a good idea.
[Created a ticket](https://github.com/software-mansion/react-native-screens-labs/issues/586)

scripts/e2e/ios-devices.js
t0maboro: > due to exceptions like iPhone X or iPhone SE

it's fine, the device type is absolutely sufficient for me, thanks

scripts/e2e/detox-utils.cjs
```diff
@@ -11,15 +13,15 @@ const apkBulidArchitecture = isRunningCI ? 'x86_64' : 'arm64-v8a';
 const testButlerApkPath = isRunningCI ? ['../Example/e2e/apps/test-butler-app-2.2.1.apk'] : undefined;
 
 function detectLocalAndroidEmulator() {
-  // "DETOX_AVD_NAME" can be set for local developement
-  const detoxAvdName = process.env.DETOX_AVD_NAME ?? null;
-  if (detoxAvdName !== null) {
-    return detoxAvdName
+  // "RNS_E2E_AVD_NAME" can be set for local developement
```
t0maboro: developement -> development - unrelated to this PR, but can correct it and all below?

scripts/e2e/ios-devices.js
```diff
@@ -0,0 +1,52 @@
+const DEFAULT_APPLE_DEVICE = 'iPhone 17';
+const DEFEAULT_IOS_VERSION = 'iOS 26.2';
```
t0maboro:
```suggestion
const DEFAULT_IOS_VERSION = 'iOS 26.2';
```
applies to all ocurrencies below

scripts/e2e/ios-devices.js
```diff
@@ -0,0 +1,52 @@
...
+  return process.env.RNS_E2E_APPLE_SIM_NAME || DEFAULT_APPLE_DEVICE;
+}
+/**
+ * @return {`iOS ${string}`} requested version of ios, or default if not specified
+ */
+function getIOSVersion() {
+  const envVariableKey = 'RNS_E2E_IOS_VERSION';
+  const passedVersion = process.env[envVariableKey];
+  if (passedVersion) {
+    if (/^(iOS)\s.+/.test(passedVersion)) {
```
t0maboro, line 26: nit: I haven't made up my mind, but we'll have a rather fixed number of checks; therefore, I'm wondering if `startsWith` won't improve readability compared to a regex. Treat it as a nitpick; for me, it's okay to leave it in its current form.

scripts/e2e/ios-devices.js
```diff
@@ -0,0 +1,52 @@
...
+const DEFEAULT_IOS_VERSION = 'iOS 26.2';
+
+/**
+ * @return {string}
+ */
+function getAppleDevice() {
+  const envVariableKey = 'RNS_E2E_APPLE_SIM_NAME';
+  const passedDevice = process.env[envVariableKey];
+  if (passedDevice) {
+    if (/^(iPhone|iPad)\s.+/.test(passedDevice)) {
```
t0maboro, line 11: nit: I haven't made up my mind, but we'll have a rather fixed number of checks; therefore, I'm wondering if `startsWith` won't improve readability compared to a regex. Treat it as a nitpick; for me, it's okay to leave it in its current form.

t0maboro (approved):

Issue #3399, opened 2025-11-14
[Android] Icons in native bottom tabs not working in release mode


johankasperi:
### Description

Moving https://github.com/react-navigation/react-navigation/issues/12838 to this repo because issue is in react-native-screens

Icons of type `ImageRequireSource` for the native tab items is not visible in release mode on Android. They are working in debug mode.

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/f680da02-49b4-4ff0-b286-c878f7589d4b" />

### Steps to reproduce

Build the react-navigation example app on android with release configuration:
`yarn android --variant release`

### Snack or a link to a repository

https://github.com/react-navigation/react-navigation/tree/main/example

### Screens version

4.18.0

### React Native version

0.81.4

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

pedrommcarrasco:
Can reproduce as well.

PR #3413, opened 2025-11-20
fix(Android, Tabs): Fix tabs icons loading in release mode

t0maboro:
## Description

This PR improves the approach to loading image assets used as tab icons, addressing issues with inconsistent or unreliable URI resolution - particularly around local resources and assets in Android builds (debug vs. release).

Previously, the image loading logic assumed that assets would always be bundled with a `_` prefix in release builds. However, this is not a reliable assumption and can break under certain project setups or custom bundling configurations. Assets are now resolved directly using the appropriate `android.resource` or `android_asset` URI schemes.

Additionally, now I'm resolving both drawable and raw into the `res:/<id>` URI format before being converted to the proper `android.resource://` URI expected by Fresco. This ensures consistent behavior regardless of resource type.

This change ensures that both local resources and bundled assets are correctly rendered as tab icons across different build variants, without relying on unstable heuristics.

Fixes https://github.com/software-mansion/react-native-screens/issues/3399

## Changes

- Updated the approach in TabsImageLoader

## Screenshots / GIFs

### Before

Debug:
debug-before

Release:
release-before

### After

Debug:
debug-after

Release:
release-after

It's working fine it both debug and release in our examples, before and after applying these changes. Therefore, I was testing it on a separate application with this code snippet:

```js
import React from 'react';
import { StyleSheet, Text, View } from 'react-native';
import { createNativeBottomTabNavigator } from '@react-navigation/bottom-tabs/unstable';

const Tab = createNativeBottomTabNavigator();

function App() {
  return (
    <NavigationContainer>
      <Tab.Navigator>
        <Tab.Screen
          name="Home"
          component={HomeScreen}
          options={{
            tabBarIcon: {
              type: "drawableResource",
              name: "custom_home_icon"
            },
          }}
        />
        <Tab.Screen
          name="Profile"
          component={ProfileScreen}
          options={{
            tabBarIcon: {
              type: 'image',
              source: require('./assets/icon_fill.png'),
            },
          }}
        />
        <Tab.Screen
          name="Settings"
          component={SettingsScreen}
          options={{
            tabBarIcon: {
              type: 'image',
              source: require('./assets/icon.png'),
            },
          }}
        />
      </Tab.Navigator>
    </NavigationContainer>
  );
}

const HomeScreen = () => (
    <Text>Home</Text>
  </View>
);

const ProfileScreen = () => (
    <Text>Profile</Text>
  </View>
);

const SettingsScreen = () => (
    <Text>Settings</Text>
  </View>
);

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center'
  },
});

export default App;
```

I'm adding `custom_home_icon` via android resource manager and other icons via `assets/` directory

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/image/TabsImageLoader.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/image/TabsImageLoader.kt
@@ -1,12 +1,13 @@
 package com.swmansion.rnscreens.gamma.tabs.image
 
+import android.annotation.SuppressLint
 import android.content.Context
 import android.graphics.drawable.Drawable
+import android.net.Uri
 import android.os.Handler
 import android.os.Looper
 import android.util.Log
 import androidx.core.graphics.drawable.toDrawable
-import androidx.core.net.toUri
 import com.facebook.common.executors.CallerThreadExecutor
 import com.facebook.common.references.CloseableReference
 import com.facebook.datasource.BaseDataSubscriber
@@ -16,6 +17,7 @@ import com.facebook.imagepipeline.image.CloseableImage
 import com.facebook.imagepipeline.image.CloseableStaticBitmap
 import com.facebook.imagepipeline.request.ImageRequestBuilder
 import com.swmansion.rnscreens.gamma.tabs.TabScreen
+import java.util.Locale
 
 internal fun loadTabImage(
     context: Context,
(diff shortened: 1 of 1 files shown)


dancamdev:
Hey Folks, thanks for the patch, is there an ETA on when this is going to land as a new `react-native-screens` version?

kkafar:
@dancamdev, we still need to land few more things before release. Until then use `react-native-screens@nightly`.

PR #3415, opened 2025-11-23
fix(IOS): Fix non dismissible modal presentation

lodev09:
## Description

This PR fixes an issue with the modal presentation chain when non-dismissible external modals (e.g., third-party modals like TrueSheet) are present in the view controller hierarchy. 

Previously, when a non-dismissible modal was encountered, the stack didn't properly update the root controller for subsequent modal presentations, which could cause modals to be presented from the wrong controller or fail to dismiss modals that were presented by the non-dismissible modal.

## Changes

- **Enhanced non-dismissible modal handling in `RNSScreenStack.mm`**: 
- Refactored the dismissal check into a separate boolean variable for better readability
- Added an `else` branch to handle the case when `firstModalToBeDismissed` is non-dismissible
- When a non-dismissible modal is detected, the `changeRootController` is now updated to point to that modal
- Added logic to check if the non-dismissible modal itself has presented any owned modals, and properly dismisses them before presenting new ones

### Before

**Sheet will not present, will get warning in XCode.**

### After

https://github.com/user-attachments/assets/90d8534b-7115-47f0-88ba-3707737da16c

To test this change with a non-dismissible third-party modal (e.g., TrueSheet):

**1. First, ensure your third-party modal implements the protocol (library author's responsibility):**

```objc
// In your third-party library's view controller (e.g., TrueSheet)
#import "RNSDismissibleModalProtocol.h"

@interface YourModalViewController : UIViewController <RNSDismissibleModalProtocol>
@end

@implementation YourModalViewController

- (BOOL)isDismissible {
  return NO; // Mark as non-dismissible
}

@end
```

**2. Then test with this example:**

```tsx
import React from 'react';
import { View, Button } from 'react-native';
import { TrueSheet } from '@lodev09/react-native-true-sheet'; // Example third-party modal

function HomeScreen({ navigation }) {
  const sheet = React.useRef<TrueSheet>(null);

  return (
    <View style={{ flex: 1, justifyContent: 'center', alignItems: 'center' }}>
      <Button title="Show TrueSheet" onPress={() => sheet.current?.present()} />
      <TrueSheet ref={sheet}>
        <Button 
          title="Open React Navigation Modal" 
          onPress={() => navigation.navigate('Modal')} 
        />
      </TrueSheet>
    </View>
  );
}

function ModalScreen() {
  return (
    <View style={{ flex: 1, justifyContent: 'center', alignItems: 'center' }}>
      <Text>This is a modal</Text>
    </View>
  );
}

function App() {
  return (
    <Stack.Navigator>
      <Stack.Screen 
        name="Modal" 
        component={ModalScreen} 
        options={{ presentation: 'modal' }} 
      />
    </Stack.Navigator>
  );
}
```

1. Open the TrueSheet (non-dismissible modal that implements `RNSDismissibleModalProtocol`)
2. From within the sheet, open a React Navigation modal
3. The modal should present correctly from the non-dismissible modal
4. Subsequent modal presentations and dismissals should work as expected
5. The TrueSheet should remain visible and not be dismissed

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation: 
- [x] Ensured that CI passes


lodev09:
@kkafar an alternative approach would be to create another method to the protocol to return a ViewController that should update the root controller to keep things safe. Let me know if that's better. Thank you!

lodev09:
Can we merge this @kligarski @kkafar? This should be straightforward and backwards compatible. Tested and works in a production app. Thank you!

FYI @erickreutz

kkafar:
Hey, sorry for very late response here & on other PRs. I'll look into this after we release 4.19.0 (planned for upcoming week).

erickreutz:
@kkafar happy new year! do you have time to take a look now please?

kkafar (commented):
Hey! Thanks for the PR. There a couple of different topics I wanna discuss here.

TLDR:

I'm reluctant to land the changes in current shape, but willing to cooperate on a fix / guiding. To do that a full explanation of the issue this PR aims to solve + issue mechanism breakdown is required.

---

First, I do not understand the problem yet. The PR describes it only vaguely:

> Previously, when a non-dismissible modal was encountered, the stack didn't properly update the root controller for subsequent modal presentations, which could cause modals to be presented from the wrong controller or fail to dismiss modals that were presented by the non-dismissible modal.

and therefore I can't tell whether this is the appropriate solution or not. I need to understand what does it mean that "[...] the stack didn't properly update the root controller [...]" and whole error mechanism & only then we'll be able to discuss possible solutions.
I'll try to find time to setup a proper reproducer app. 

Second, please note that you do not propose an fix the the library itself, but rather an additional mechanism that you want to use to integrate a third-party (from the library perspective) component with the `react-native-screens` and **it'll be on `react-native-screens` side to maintain it**. At the same time `react-native-screens` exposes `presentation: 'formSheet'` option, which allows to achieve *similar* (likely not as extensive) functionality.

I acknowledge the need for this kind of integration, and that other components often need to be able to work with `react-native-screens` components, but we need to understand the needs first and weight benefits from the library and users perspective. 

Third - now, talking directly about the code - I do not understand the changes (and their implications) completely. This is a very fragile fragment of code that went under a lot of fixing throughout last few years:

* https://github.com/software-mansion/react-native-screens/commit/c21de50e4c8dfb2495ff2850999807f33ea60e84
* https://github.com/software-mansion/react-native-screens/commit/87f2dbcfb981b39144d5cb9cf2f99cb6169282e1
* https://github.com/software-mansion/react-native-screens/commit/b6c325b3fcf729251afd755824edddc573526154
* https://github.com/software-mansion/react-native-screens/commit/471127e26aa23eed4d14388a9646bf83de5e8699
* https://github.com/software-mansion/react-native-screens/commit/0564d43bb5db9c437af422de76b5724c14488d08

and often when we fixed one thing we broke some other interaction. Therefore I need profound understanding of the changes before I give a green light to merge this. I'll put more time into testing this & considering different scenarios. Comprehensive test plan would be beneficial here. 

Fourth - yet thing is that we are in progress of [new stack implementation](https://github.com/software-mansion/react-native-screens/pull/3518) and we're reluctant on landing any new APIs we'll potentially have to maintain in the soon-to-be-old stack implementation.

lodev09:
Hey @kkafar, thanks for taking the time to review the PR! I understand where you're coming from. For a reproducer, I have an extensive example that integrates react-navigation (react-native-screens) with my library. You can check it out here if you want to try it: https://github.com/lodev09/react-native-true-sheet/tree/main/example

Basically I was re-using the existing non-dismissible protocol (if not mistaken, was from expo's link preview). The goal was to not dismiss my sheet when presenting a modal screen. The issue is that react-native-screens will now use the wrong presenting controller so the additional protocol for my lib to provide the presenting controller was necessary.

Please let me know if there's any other way, I figured a protocol is the safest approach since it doesn't introduce any breaking changes.

EDIT:
Here's the warning I get in xcode when trying to present a modal screen without the additional protocol:
```
Attempt to present <RNSScreen: 0x116b3fe00> on <UIViewController: 0x105c10740> (from <RNSNavigationController: 0x115d7f800>) which is already presenting <TrueSheetViewController: 0x1079cf000>.
```

kkafar:
Thanks for the update here. I'll check out your test application (likely tomorrow) and get back to you.

kkafar:
I remember about this PR, haven't found time yet.

ugoi:
Is this fix related to the issue below? When navigating to the sign-in screen from the true sheet, it won't navigate or dismiss. Anyone else seen this?

https://github.com/user-attachments/assets/968ee248-7ef4-484d-be8e-3bfcc6cab42f

ugoi:
I'm having this issue with TrueSheet:

https://sheet.lodev09.com/guides/navigation

https://github.com/user-attachments/assets/91b8875a-8c5f-4e06-9e94-b40608873947

Tehnix:
Blocked on this PR to be able to use `formSheet` via Expo Router (which uses react-native-screens) together with TrueSheet. Currently having a TrueSheet open on a page means that my routes using a `formSheet` now doesn't work anymore.

Is it possible to get any guidance on how to move this PR forward?

erickreutz:
@kkafar what do you think about merging this?

RodolfoSilva:
@ugoi I have the same issue on Expo SDK 56, and I can confirm that this patch solves my problem! Anyone interested, here's the patch file:
[react-native-screens+4.25.2.patch](https://github.com/user-attachments/files/28671207/react-native-screens%2B4.25.2.patch)

Another alternative, is avoid to use `presentation: 'modal'` on react-router

iverly:
@kkafar up ? 🙏

lodev09:
@kkafar any issue I need to address for this to land? :/

Issue #3302, opened 2025-10-16
[v4.17.x] `fullScreenGestureEnabled` conflict with `ScrollView`


huextrat:
### Description

With v4.17.x enabling `fullScreenGestureEnabled` on native stack is conflicting when there is a `ScrollView` in the screen.

So it has no effect when the screen content is scrollable and only default iOS gesture is working

Before v4.17.x this behavior is working fine

Suspected PR: https://github.com/software-mansion/react-native-screens/pull/3173

### Steps to reproduce

1. Launch repro on iOS
2. Open Profile
3. Try to back by doing a swipe back gesture

### Snack or a link to a repository

https://github.com/huextrat/repro-screens-fullscreen-swipe

### Screens version

4.17.1

### React Native version

0.81.4

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

kkafar:
You're right in terms, that there is difference of behaviour between iOS 18 & 26 as can be seen on recording below: 

https://github.com/user-attachments/assets/904b8ce9-6662-49a3-ad4c-5afc6a0a6d46

However I'm not yet sure that this is a bug. I need to verify what's the native behaviour here. This might just be the way the native UIKit-provided `interactivePopContentGestureRecognizer` works, and we'll like to go with that. 

Will be back soon with more info

kkafar:
Seems that this is indeed a bug. Under appropriate conditions I can activate `interactivePopContentGestureRecognizer` while having scrollview inside on pure UIKit app. 

Thanks for the report! I'll patch that soon & release patch version. 

https://github.com/user-attachments/assets/65c73eeb-395f-4c52-b2c8-1de39b2081a1

mozzius:
Thank you!!

kkafar:
This won't be a quick fix unfortunately. The code looks good and logs indicate that the content gesture recognizer should recognize, but yet it doesnt. I dont have access to its code (UIKit) & it'll require some deeper debugging. I'll get back to this once I handle few other matters

mozzius:
I'm trying to narrow down when exactly this started happening to us, as we patched in https://github.com/software-mansion/react-native-screens/pull/3173 a while back and it solved our issue with the swipe conflicts. Yesterday, I ran a fresh native build and I started seeing this issue - note, we haven't merged v3.17.1 yet, this is just with the cherry-pick patch.

[...]

PR #3420, opened 2025-11-26
fix(iOS): Distinguish horizontal and vertical scroll in gesture failure requirements

kmichalikk:
Fixes https://github.com/software-mansion/react-native-screens/issues/3302

Supersedes https://github.com/software-mansion/react-native-screens/pull/3304

## Description

This PR fixes the failure requirements for content pop gesture. The changes introduced in #3265 fixed the immediate dismissing of screen when scrolling on horizontal ScrollView, but at the same time introduced a regression for vertical scrolling. Here, we apply the failure requirement only if the ScrollView has horizontal scrolling.

## Changes

Updated `RNSScreenStack`'s `gestureRecognizer:shouldRequireFailureOf:otherGestureRecognizer` to check if ScrollView has horizontal scrolling.

### After changes

https://github.com/user-attachments/assets/de8bfdaa-aa9b-4c25-917e-2d631bc590df

## Testing
Use `Test1072`. Make sure to check custom animations with `animation` and `animationMatchesGesture`.

### DIFF
--- a/ios/RNSScreenStack.mm
+++ b/ios/RNSScreenStack.mm
@@ -1152,6 +1152,14 @@ RNS_IGNORE_SUPER_CALL_END
   return scrollView.panGestureRecognizer == gestureRecognizer;
 }
 
+// Check if ScrollView has horizontal scrolling.
+// This is the same logic as in RCTScrollView:
+// https://github.com/facebook/react-native/blob/69b131c2b1627f64f34a534dac4cf48a542e6ea6/packages/react-native/React/Views/ScrollView/RCTScrollView.m#L557
+- (BOOL)scrollViewHasHorizontalPart:(UIScrollView *)scrollView
+{
+  return scrollView.contentSize.width > scrollView.frame.size.width;
+}
+
 // Custom method for compatibility with iOS < 13.4
 // RNSScreenStackView is a UIGestureRecognizerDelegate for three types of gesture recognizers:
 // RNSPanGestureRecognizer, RNSScreenEdgeGestureRecognizer, _UIParallaxTransitionPanGestureRecognizer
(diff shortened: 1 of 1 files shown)


ios/RNSScreenStack.mm
```diff
@@ -1152,6 +1152,14 @@ - (BOOL)isScrollViewPanGestureRecognizer:(UIGestureRecognizer *)gestureRecognize
   return scrollView.panGestureRecognizer == gestureRecognizer;
 }
 
+// Check if ScrollView has horizontal scrolling.
+// This is the same logic as in RCTScrollView:
+// https://github.com/facebook/react-native/blob/69b131c2b1627f64f34a534dac4cf48a542e6ea6/packages/react-native/React/Views/ScrollView/RCTScrollView.m#L557
+- (BOOL)scrollViewHasHorizontalPart:(UIScrollView *)scrollView
```
kligarski, line 1158: Not sure about the naming here, maybe `isScrollViewHorizontal` is okay?

kkafar: +1 here

The name we have currently makes sense only in case scrollview can have both vertical & horizontal behaviour at the same time. Is that the case? Maybe it is, I just dunno.

kkafar: If it has, I'd stick to the current naming, otherwise follow @kligarski suggestion.

kmichalikk: The logic I add for gestures should work for SV with horizontal or vertical+horizontal, hence having "horizontal part". I didn't check if react allows for both at the same time, but UIKit does.

kkafar: If UIKit does, we should stick with the current naming. 👍🏻

PR #3422, opened 2025-11-26
fix(iOS, Stack): delay setting navigation bar visibility to mitigate bar button bug on iOS 26

kligarski:
## Description

Moves setting navigation bar visibility to the end of header configuration (after bar button items are setup) to mitigate bug on iOS 26 with bar button items being set after a delay.

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/96a0341a-de50-4628-9fd1-bb41a1258ed8" /> | <video src="https://github.com/user-attachments/assets/ec28b7c5-7d12-4461-bcb9-3ac58846cbfa" /> |

Closes https://github.com/react-navigation/react-navigation/discussions/12844.
Closes https://github.com/software-mansion/react-native-screens-labs/issues/560.

## Changes

- split setting nav bar visibility and delay it to the end of config (after bar button items)

Run `Test3422`.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -581,7 +581,7 @@ RNS_IGNORE_SUPER_CALL_END
     navitem.title = config.title;
 
     // Setting navigation bar visibility is split to mitigate iOS 26 bug with bar button items.
-    [navctr setNavigationBarHidden:shouldHide animated:animated];
+    [navctr setNavigationBarHidden:YES animated:animated];
     return;
   }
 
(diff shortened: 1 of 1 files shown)


ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -575,12 +575,13 @@ + (void)updateViewController:(UIViewController *)vc
...
 
-  [navctr setNavigationBarHidden:shouldHide animated:animated];
-
   [config applySemanticContentAttributeIfNeededToNavCtrl:navctr];
 
   if (shouldHide) {
     navitem.title = config.title;
+
+    // Setting navigation bar visibility is split to mitigate iOS 26 bug with bar button items.
+    [navctr setNavigationBarHidden:shouldHide animated:animated];
```
kkafar: pass explicitly YES here, and NO in the other palce

kkafar: You might even add assertion in the other place, that the `shouldHide == NO`, as we expect to return here early, when it is true.

kligarski: https://github.com/software-mansion/react-native-screens/pull/3422/commits/53ffd95cbbbb0ef1e60671511b6a255f5329c101

PR #3426, opened 2025-11-26
chore(e2e): detox basic config android

KrzysztofWojnar:
## Description

Unfortunately I had to split android config into 2 pieces because of the volume of the changes. So `RNS_E2E_ANDROID_API_LEVEL` environment variable will be introduced in the next PR. Detecting the highest available API level will be introduced too.

## Changes

1. Adjusted environment variables name to be up-to-date with the [discussion about running automated e2e tests](https://github.com/software-mansion/react-native-screens-labs/discussions/572).
2. Complex support for "RNS_DEVICE_SERIAL" and "RNS_AVD_NAME" environment variables – you could pass emulator's adb serial.
3. Running emulators on initialization config – this will be needed in the future change to query emulators for their API Level
4. Extracting Android logic from detox-utils.cjs to android-devices.js
5. Initialization for unused config skipped (running devices are time-expensive)
6. Handling edge cases (waiting for devices to boot, or handling their unavailability)

#### After
RNS_DEVICE_SERIAL
Fixed apple
iOS passed version
iPhone passed
iOS no vars passed

RNS_AVD_NAME

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [ ] Ensured that CI passes

### DIFF
--- a/scripts/e2e/detox-utils.cjs
+++ b/scripts/e2e/detox-utils.cjs
@@ -35,9 +35,18 @@ const testButlerApkPath = isRunningCI
  *
  * * `RNS_APPLE_SIM_NAME` env var can be set in case of running tests on iOS simulator.
  * The expected value here is exactly as one listed in XCode.
+ * Example: RNS_APPLE_SIM_NAME="iPhone 16 Pro"
  *
  * * `RNS_IOS_VERSION` env var can be specified to request particular iOS version
  * for the given simulator. Note that required SDK & simulators must be installed.
+ * Example: RNS_IOS_VERSION="iOS 26.1"
+ * 
+ * * Remember:
+ * Device versions are assigned to iOS versions.
+ * That means running a version that has never been available
+ * on a given device will result in an error.
+ * Example: `RNS_IOS_VERSION="iOS 18.6" RNS_APPLE_SIM_NAME="iPhone 17 Pro" yarn test-e2e-ios` will fail
+ * as iPhone 17 Pro was released with iOS 26
  *
  * @param {string} applicationName name (FabricExample / ScreensExample)
  * @returns {Detox.DetoxConfig}


scripts/e2e/android-devices.js
```diff
@@ -0,0 +1,93 @@
+const { getCommandLineResponse } = require('./command-line-helpers');
+
+const DEFAULT_CI_AVD_NAME = 'e2e_emulator';
+const isRunningCI = process.env.CI;
```
t0maboro:
```suggestion
const isRunningCI = !!process.env.CI;
```
if we're naming the variable is... , I'd expect the outcome to be boolean

t0maboro: or maybe `process.env.CI !== undefined` ?

scripts/e2e/android-devices.js
```diff
@@ -0,0 +1,93 @@
...
+  adbSerial: 'RNS_DEVICE_SERIAL',
+  avdName: 'RNS_AVD_NAME'
+});
+
+const passedAdbSerial = process.env[envVarKeys.adbSerial];
+
+function resolveAttachedAndroidDeviceSerial() {
+  if (passedAdbSerial) return passedAdbSerial;
+  const connectedNonEmulators = getDeviceIds()
+    .filter(deviceID => !deviceID.startsWith('emulator'));
```
t0maboro: if the device is not in state `device` we should filter it out (it might be e.g. in `offline` state, the list of states is available here https://developer.android.com/tools/adb#devicestatus)

scripts/e2e/android-devices.js
```diff
@@ -0,0 +1,93 @@
...
+  } else if (connectedNonEmulators.length === 1) {
+    return connectedNonEmulators[0];
+  } else throw new Error(
+    `Connected devices: ${connectedNonEmulators.join(', ')}\nUnplug ${connectedNonEmulators.length - 1} device(s) or use ${envVarKeys.adbSerial} environment variable to select a specific one.`
+  );
+}
+
+function detectAndroidEmulatorName() {
+  const isEmulatorConfig = process.argv.some(runtimeArg => runtimeArg.includes('android.emu'));
+  if (!isEmulatorConfig) return 'INACTIVE CONFIG';
```
t0maboro, line 51: Q: should we pass `INACTIVE CONFIG` to `avdName` or throw an error?

scripts/e2e/android-devices.js
```diff
@@ -0,0 +1,93 @@
...
+  if (passedAdbSerial) {
+    return getDeviceName(passedAdbSerial);
+  }
+  return process.env[envVarKeys.avdName] || isRunningCI
+    ? DEFAULT_CI_AVD_NAME
+    : detectLocalAndroidEmulator();
```
t0maboro: Can we refactor it? I'm okay even with extracting it to a separate function, I strongly recommend not using alternative and ternary together as it's confusing what's the intention

scripts/e2e/android-devices.js
```diff
@@ -0,0 +1,93 @@
...
+}
+
+function detectLocalAndroidEmulator() {
+  return getAvailableEmulatorNames()[0];
```
t0maboro: please leave some comment somewhere that `getAvailableEmulatorNames` is checking that the list won't be empty, and that's why we can take the value from array without any check

scripts/e2e/android-devices.js
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/**
+ * @returns {string[]}
+ */
+function getDeviceIds() {
+  const nonEmptyLines = getCommandLineResponse('adb devices')
+    .split('\n')
+    .map(line => line.trim())
+    .filter(Boolean);
```
t0maboro: nice, can it be added to `avdList` in line 44? just in case

scripts/e2e/android-devices.js
```diff
@@ -0,0 +1,93 @@
...
+  nonEmptyLines.shift(); // The first line is always the "List of devices attached" (header) so we ignore it
+  if (nonEmptyLines.length === 0) {
+    throw new Error("Seems that the attached device list is empty");
+  }
+  return nonEmptyLines.map(line => {
+    const [id, state] = line.split('\t');
+    if (state !== 'device') {
+      console.warn(`THE DEVICE (ID ${id}) HAS STATUS "${state}". ITS STATUS SHOULD BE 'device' TO CONTINUE!`);
```
t0maboro: no necessity to caps here, at least we're not doing it in other places, so we should keep the consistency

scripts/e2e/android-devices.js
```diff
@@ -0,0 +1,93 @@
...
+    }
+    return id;
+  });
+}
+
+/**
+ * @param {string} deviceId adb identifier, device serial
```
t0maboro:
```suggestion
 * @param {string} device adb identifier, device serial
```

scripts/e2e/command-line-helpers.js
```diff
@@ -0,0 +1,15 @@
+const ChildProcess = require('node:child_process');
+/**
+ * Runs a command and returns it's response as string
+ * @param {string} command to run
```
t0maboro:
```suggestion
 * @param {string} command - The command to run.
```

scripts/e2e/command-line-helpers.js
```diff
@@ -0,0 +1,15 @@
...
+ * @returns {string} command-line response
+ */
+function getCommandLineResponse (command) {
+    const stdout = ChildProcess.execSync(command);
+    // Possibly convert Buffer to string
+    return typeof stdout === 'string' ? stdout : stdout.toString();
```
t0maboro: you can pass encoding option to `execSync`, then you wouldn't need to check the type of stdout

scripts/e2e/command-line-helpers.js
```diff
@@ -0,0 +1,15 @@
+const ChildProcess = require('node:child_process');
+/**
+ * Runs a command and returns it's response as string
+ * @param {string} command to run
+ * @returns {string} command-line response
+ */
+function getCommandLineResponse (command) {
```
t0maboro: add error handling

scripts/e2e/detox-utils.cjs
```diff
@@ -12,44 +9,34 @@ const apkBulidArchitecture = isRunningCI ? 'x86_64' : 'arm64-v8a';
...
-    throw new Error(errorMessage);
-  }
-}
-
-function detectAndroidEmulatorName() {
-  // "RNS_E2E_AVD_NAME" can be set for local development
-  return isRunningCI ? DEFAULT_CI_AVD_NAME : detectLocalAndroidEmulator();
-}
+/**
+ * @typedef {import('../../FabricExample/node_modules/detox')} Detox
```
t0maboro, line 15: I don't like checking `node_modules` from a specific app in a common script, please remember that we're still having many apps and we shouldn't force installing `node_modules` in a specific (sometimes different, not currently tested app)

scripts/e2e/detox-utils.cjs
```diff
@@ -12,44 +9,34 @@ const apkBulidArchitecture = isRunningCI ? 'x86_64' : 'arm64-v8a';
...
+ * Android:
+ *
+ * * `RNS_DEVICE_SERIAL` env var can be specified in case of running
+ * tests with an attached Android device. It can also be an emulator.
+ * The expected value here is the same as you would pass to `adb -s`.
+ * You can find device serial by running `adb devices` command.
+ * Example: RNS_DEVICE_SERIAL=33221FDH3000VT
+ *
+ * * `RNS_AVD_NAME` env var can be specified in case of running tests on Android emulator.
+ * The exepected value here is the same as displayed in Android Studio or listed by
```
t0maboro:
```suggestion
 * The expected value here is the same as displayed in Android Studio or listed by
```

scripts/e2e/turn-on-android-emulators.js
```diff
@@ -0,0 +1,132 @@
...
+    avdNames.forEach(name => {
+      console.log(`   🚀 Spawning ${name}...`);
+      spawnEmulator(name);
+      execSync('sleep 1'); // prevent choking the ADB server
+    });
+
+    console.log(`⏳ Waiting for devices to appear in ADB...`);
+    let currentDevices = [];
+
+    while (true) {
```
t0maboro: do not use infinite loop, define maxAttempts and fail if it doesn't work for any reason

scripts/e2e/turn-on-android-emulators.js
```diff
@@ -0,0 +1,132 @@
...
+    return [];
+  }
+}
+
+/**
+ * @param {string} serial
+ */
+function waitForBootSync(serial) {
+  let booted = false;
+  while (!booted) {
```
t0maboro: also, I'd add max attempts

scripts/e2e/turn-on-android-emulators.js
```diff
@@ -0,0 +1,132 @@
...
+          stdio: ['ignore', 'pipe', 'ignore'], // offline/unauthorized may cause stderr so we ignore it
+        },
+      ).trim();
+      if (res === '1') {
+        booted = true;
+      } else {
+        execSync('sleep 2');
+      }
+    } catch (_) {
+      execSync('sleep 2');
```
t0maboro: Here we should handle an error properly, I don't see any case for just calling sleep here

scripts/e2e/android-devices.js
KrzysztofWojnar: This script runs expensive logic like `adb -s ${deviceId} emu avd name` or `adb devies`. I just return _a string_ when I don't need **this** config, because I run e.g. config ios.sim.release.
I considered making a proxy (js Proxy) object, so that **function could be called when Detox gets this property**, but I don't really like it (in fact, I hate it).
I can consider this text as a placeholder.

scripts/e2e/detox-utils.cjs
KrzysztofWojnar: Yes, I see your point, In this case missing `node_modules` results in nothing but a red underline in IDE, as long as we don't validate jsdoc types as a quality gate. It could also potentially get outdated (but unlikely), but even if that happens there is no real risk here. I admit, this solution isn't perfect, but surely it's better than before.
I am also not sure about adding Detox to the root package.json just to validate config's type in the common scripts, what do you think?

scripts/e2e/android-devices.js
KrzysztofWojnar: afaik it should never happen for an emulator. But just in case `getDeviceIds()` shows a warning for emulators in different state (line 71).

scripts/e2e/android-devices.js
KrzysztofWojnar: absolutely!

scripts/e2e/android-devices.js
KrzysztofWojnar: sure, will do

scripts/e2e/command-line-helpers.js
KrzysztofWojnar: insightful tip
In that case now I think I don't need a special util for that.

KrzysztofWojnar: I made my mind, I would keep it.

scripts/e2e/turn-on-android-emulators.js
KrzysztofWojnar: I meant: if the device is not ready **_yet_** the execSync will throw so I catch, and sleep (in the loop).

scripts/e2e/android-devices.js
t0maboro: yes, but here we're filtering emulators out, so having only physical devices, it may happen (e.g. with bad network connection)

scripts/e2e/android-devices.js
```diff
@@ -0,0 +1,110 @@
...
+const envVarKeys = /** @type {const} */ ({
+  adbSerial: 'RNS_DEVICE_SERIAL',
+  avdName: 'RNS_AVD_NAME',
+});
+
+const passedAdbSerial = process.env[envVarKeys.adbSerial];
+
+function resolveAttachedAndroidDeviceSerial() {
+  if (passedAdbSerial) return passedAdbSerial;
+  const connectedNonEmulators = getDeviceIds().filter(
```
t0maboro: `connectedNonEmulators` -> `connectedPhysicalDevices` - no necessary to have a negation in the variable name

scripts/e2e/android-devices.js
```diff
@@ -0,0 +1,110 @@
...
+    return connectedNonEmulators[0];
+  } else
+    throw new Error(
+      `Connected devices: ${connectedNonEmulators.join(', ')}\nUnplug ${
+        connectedNonEmulators.length - 1
+      } device(s) or use ${
+        envVarKeys.adbSerial
+      } environment variable to select a specific one.`,
+    );
+}
```
t0maboro: nit: mixing throwing exceptions & returning values doesn't look good, I'd propose to reorganize this code a little bit:
```js
if (connectedNonEmulators.length === 0) {
// throw
}

if (connectedNonEmulators.length > 1) {
// throw
}

return ...
```

scripts/e2e/android-devices.js
```diff
@@ -0,0 +1,110 @@
...
+
+    return avdList;
+  } catch (error) {
+    const errorMessage = `Failed to find any Android emulator. Set "${envVarKeys.avdName}" env variable pointing to one. Cause:\n${error}`;
+    throw new Error(errorMessage);
+  }
+}
+
+/**
+ * @returns {string[]}
```
t0maboro: would be nice to mention what is the expected outcome

scripts/e2e/android-devices.js
```diff
@@ -0,0 +1,110 @@
...
+ */
+function getDeviceIds() {
+  const nonEmptyLines = getCommandLineResponse('adb devices')
```
t0maboro: `nonEmptyLines` -> `adbDeviceLines`

scripts/e2e/android-devices.js
```diff
@@ -0,0 +1,110 @@
...
+    .split('\n')
+    .map(line => line.trim())
+    .filter(Boolean);
+  nonEmptyLines.shift(); // The first line is always the "List of devices attached" (header) so we ignore it
```
t0maboro: 1. (nit): for longer comments, I prefer to place them above the line which it's touching
2. can be rephrased to `// Remove header line: "List of devices attached"
`

scripts/e2e/android-devices.js
```diff
@@ -0,0 +1,110 @@
...
+  if (nonEmptyLines.length === 0) {
+    throw new Error('Seems that the attached device list is empty');
```
t0maboro:
```suggestion
    throw new Error('The attached device list is empty');
```

scripts/e2e/android-devices.js
```diff
@@ -0,0 +1,110 @@
...
+  }
+  return nonEmptyLines.map(line => {
+    const [id, state] = line.split('\t');
+    if (state !== 'device') {
+      console.warn(
+        `The device (id ${id}) has status "${state}". Its status should be "device" to continue!`,
+      );
+    }
```
t0maboro: we're signaling that some action needs to be taken, but we do not take any action, allowing devices with unexpected state to be passed further

scripts/e2e/android-devices.js
```diff
@@ -0,0 +1,110 @@
...
+    return id;
+  });
+}
+
+/**
+ * @param {string} deviceId - Device adb identifier, device serial
+ * @returns {string} device name (avd name)
+ */
+function getDeviceName(deviceId) {
```
t0maboro: `resolveAvdNameFromDeviceId` is more precise

scripts/e2e/android-devices.js
```diff
@@ -0,0 +1,110 @@
...
+  const deviceName = getCommandLineResponse(
+    `adb -s ${deviceId} emu avd name`,
+  ).split('\r\n')[0];
```
t0maboro, line 92: do we need to have some error handling for the physical device?

scripts/e2e/command-line-helpers.js
```diff
@@ -0,0 +1,21 @@
+const ChildProcess = require('node:child_process');
+/**
+ * Runs a command and returns it's response as string
```
t0maboro:
```suggestion
 * Runs a command and returns its response as a string
```

scripts/e2e/command-line-helpers.js
```diff
@@ -0,0 +1,21 @@
...
+ * @param {string} command - The command to run.
+ * @returns {string} command-line response
+ */
+function getCommandLineResponse(command) {
+  try {
+    return ChildProcess.execSync(command, { encoding: 'utf-8' });
+  } catch (e) {
+    throw new Error(
+      `Execution command "${command}" failed! Reason:\n${
+        /** @type {Error} */ (e).message
```
t0maboro, line 13: please verify whether `stderr` won't be more precise here

scripts/e2e/detox-utils.cjs
t0maboro: Okay, thanks for the explanation. Unfortunately, I don't have any good recommendations on how you could bypass it. Keeping it open for a while to revisit, because if we decide to introduce types, I'd prefer to have them well integrated, either. I mean, it's not even touching this PR, but it sounds like a problem that will be returning in the future.
Having detox in the root's package.json just for satisfying types doesn't sound well, so let me rethink what's possible here, I'll be back soon. For now, treat this comment as non-blocking for merging this PR.

scripts/e2e/turn-on-android-emulators.js
t0maboro: ok, so it would be nice to add a comment in this place, why we handle it like that

scripts/e2e/turn-on-android-emulators.js
```diff
@@ -0,0 +1,146 @@
...
+ * @param {string[]} avdNames - Array of AVD names to boot (e.g. ['Nexus_6P', 'Pixel_4'])
+ * @returns {string[]} - The list of new ADB serial IDs (e.g. ['emulator-5554', 'emulator-5556'])
+ */
+
+function bootDevices(avdNames) {
+  if (!avdNames || avdNames.length === 0) return [];
+
+  try {
+    const initialDevices = getConnectedDevices();
+    const expectedTotal = initialDevices.length + avdNames.length;
```
t0maboro: `expectedTotal` -> `targetDeviceCount`

scripts/e2e/turn-on-android-emulators.js
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    console.log(`⏳ Waiting for devices to appear in ADB...`);
+
+    /**
+     * @type {string[]}
+     */
+    let currentDevices = [];
+
+    const giveUpTimestamp = Date.now() + ALL_DEVICES_TIMEOUT;
```
t0maboro: do we need to base on the time? would it be possible to estimate the number of attempts required to cover, let's say 99% runs? it the time estimation more solid in this case?

scripts/e2e/turn-on-android-emulators.js
```diff
@@ -0,0 +1,146 @@
...
+}
+
+/**
+ * @param {string} serial
+ */
+function waitForBootSync(serial) {
+  let booted = false;
+  const giveUpTimestamp = Date.now() + DEVICE_RESPONSE_TIMEOUT;
+  while (!booted) {
+    if (giveUpTimestamp < Date.now()) break;
```
t0maboro: is it correct?

scripts/e2e/turn-on-android-emulators.js
```diff
@@ -0,0 +1,146 @@
...
+
+    console.log('🎉 All requested devices are running!');
+    return newSerials;
+  } catch (error) {
+    console.error(
+      '❌ Error booting devices:',
+      /** @type {Error} */ (error).message,
+    );
+    process.exit(1);
+  }
```
t0maboro: shouldn't we rather throw an error?

scripts/e2e/turn-on-android-emulators.js
```diff
@@ -0,0 +1,146 @@
...
+  });
+  subprocess.unref();
+}
+
+/**
+ * @returns {string[]} array of device ids
+ */
+function getConnectedDevices() {
+  try {
+    return execSync('adb devices', { encoding: 'utf8' })
```
t0maboro: can we reuse `getCommandLineResponse`?

scripts/e2e/android-devices.js
t0maboro: if we run e.g. ios config, I'd expect this function to be never called - is it doable, or we're limited by detox here?

KrzysztofWojnar: the whole config is determined at once, and then "passed" to Detox. But I can **move** this check to [detox config around line 97](https://github.com/software-mansion/react-native-screens/blob/6f576b6570e23e508a1c85c0df7b5d6391887b08/scripts/e2e/detox-utils.cjs#L97C1-L98C1)

scripts/e2e/android-devices.js
KrzysztofWojnar: I don't know if I should care about it (for now it's user's responsibility to have proper devices).
However, the device we are querying **may** not be the device we are looking for, so I don't want to interrupt the entire flow. I will rephrase this warning, but let me know if you have a better idea.

scripts/e2e/android-devices.js
KrzysztofWojnar: I guess line 104 is not enough? I would rephrase it to "device name" to "emulator name"

scripts/e2e/command-line-helpers.js
KrzysztofWojnar: Sorry, I am not 100% what do you mean, should I mention in the error message that error in stderr occurred?

scripts/e2e/detox-utils.cjs
KrzysztofWojnar: Unfortunately this specific type is also a bit too complex to declare some naive copy in this file itself

scripts/e2e/turn-on-android-emulators.js
KrzysztofWojnar: We can switch to attempts, but as we perform one attempt per second (approx), I think it's more human to have seconds. We could interpret the issue "It didn't make it in x seconds" instead of "x attempts". But that's my opinion, not an objective truth.

scripts/e2e/turn-on-android-emulators.js
KrzysztofWojnar: Yeah, let's throw an error

scripts/e2e/turn-on-android-emulators.js
KrzysztofWojnar: how did I missed that??

scripts/e2e/turn-on-android-emulators.js
KrzysztofWojnar: it works, but I break the loop, not the flow... so the warning is missing

scripts/e2e/turn-on-android-emulators.js
KrzysztofWojnar: I moved getDeviceIds here as it's quite similar

scripts/e2e/android-devices.js
t0maboro: From this context, imo it's detox limitation, so I'm okay with the current approach

scripts/e2e/android-devices.js
t0maboro: I mean that `getCommandLineResponse` will throw sth like this:
```
/Users/tomaszboron/react-native-screens/scripts/e2e/test111.js:7
    throw new Error(
    ^

Error: Execution command "adb -s XXX emu avd name" failed! Reason:
Command failed: adb -s XXX emu avd name
    at getCommandLineResponse (/Users/tomaszboron/react-native-screens/scripts/e2e/test111.js:7:11)
    at resolveAvdNameFromDeviceId (/Users/tomaszboron/react-native-screens/scripts/e2e/test111.js:16:22)
    at Object.<anonymous> (/Users/tomaszboron/react-native-screens/scripts/e2e/test111.js:25:13)
    at Module._compile (node:internal/modules/cjs/loader:1692:14)
    at Object..js (node:internal/modules/cjs/loader:1824:10)
    at Module.load (node:internal/modules/cjs/loader:1427:32)
    at Module._load (node:internal/modules/cjs/loader:1250:12)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:235:24)
    at Module.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:152:5)
```

I'm not sure whether the reason is precise enough, or if we rather should intercept the error and here give more precise explanation what has happened, rather than just the stacktrace

scripts/e2e/command-line-helpers.js
t0maboro: I was thinking about sth like this, but it seems that adb command are not redirecting output to stderr, please ignore
```
const stderr = error.stderr?.toString() || '';

throw new Error(`Command "${command}" failed with ${stderr}`);
```

scripts/e2e/android-devices.js
```diff
@@ -45,7 +50,27 @@ function detectAndroidEmulatorName() {
...
+  try {
+    const alreadyRunningDevices = new Set(
+      getDeviceIds().map(resolveAvdNameFromDeviceId)
+    );
+    const inactiveEmulators = allAvailableEmulatorNames.filter(
+      deviceName => !alreadyRunningDevices.has(deviceName),
+    );
+    bootDevices(inactiveEmulators);
+  } catch (_) {
+    bootDevices(allAvailableEmulatorNames);
```
t0maboro: bootDevices may throw an error, so having it in catch block is not safe

scripts/e2e/turn-on-android-emulators.js
```diff
@@ -0,0 +1,173 @@
...
+  try {
+    return getDeviceIds().filter(line => line.trim().startsWith('emulator-'));
+  } catch (_) {
+    return [];
+  }
+}
+
+/**
+ * @returns {string[]} list of device adb serials,
+ * for both physicall and emulated devices, but only
```
t0maboro:
```suggestion
 * for both physical and emulated devices, but only
```

scripts/e2e/turn-on-android-emulators.js
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * @param {string} serial
+ */
+function waitForBootSync(serial) {
+  let booted = false;
+  const giveUpTimestamp = Date.now() + DEVICE_RESPONSE_TIMEOUT;
+  while (!booted) {
+    if (giveUpTimestamp < Date.now()) {
+      console.warn(`The heathcheck for device ${serial} timed out!`);
```
t0maboro:
```suggestion
      console.warn(`The healthcheck for device ${serial} timed out!`);
```

scripts/e2e/android-devices.js
```diff
@@ -0,0 +1,112 @@
...
+  }
+  const passedAvdName = process.env[envVarKeys.avdName];
+  if (passedAvdName) {
+    return passedAvdName;
+  }
+  return isRunningCI ? DEFAULT_CI_AVD_NAME : detectLocalAndroidEmulator();
+}
+
+function detectLocalAndroidEmulator() {
+  bootInactiveEmulators();
```
t0maboro: why do we need to boot all inactive devices and to return only 1 device name?

scripts/e2e/turn-on-android-emulators.js
t0maboro: Note: I recommend installing https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker if you're working with VSCode, it's catching typos well

scripts/e2e/turn-on-android-emulators.js
```diff
@@ -0,0 +1,173 @@
...
+        `\r   Found: ${currentDevices.length} / ${targetDeviceCount}`,
+      );
+
+      if (currentDevices.length >= targetDeviceCount) {
+        console.log('\n✅ All devices registered.');
+        break;
+      }
+
+      // check every second
+      execSync('sleep 1');
```
t0maboro: I believe that we're okay with the fact that execSync is blocking https://www.zignuts.com/blog/nodejs-sleep-function#:~:text=execSync%20can%20create%20a%20sleep%20function%20in%20Node.js%20from%20the%20%27child_process%27%20module%2C%20which%20allows%20you%20to%20execute%20shell%20commands%20synchronously.%20This%20method%20is%20a%20bit%20unconventional%20and%20also%20blocks%20the%20event%20loop%2C%20so%20it%20should%20be%20used%20with%20caution. I believe we are, because we don't need to have super-performant or scalable version for simple scripts.

scripts/e2e/turn-on-android-emulators.js
```diff
@@ -0,0 +1,173 @@
...
+    throw new Error('The attached device list is empty');
+  }
+  return adbDeviceLines.map(line => {
+    const [id, state] = line.split('\t');
+    if (state !== 'device') {
+      console.warn(
+        `The device (id ${id}) has status "${state}". Its status should be "device" to continue!`,
+      );
+    }
+    return id;
```
t0maboro: if it should be in a `device` state to continue, I think we should handle it asap, e.g., filtering out this device and not allowing to proceed forward, because this log suggests that, we shouldn't

scripts/e2e/android-devices.js
KrzysztofWojnar: This is the problem of breaking the changes into smaller PRs. I'll need this to be able to query the device for its API level in upcoming changes. I would need this to run tests on the latest version, or to find a device with a specified API level.

t0maboro: ok, makes sense

scripts/e2e/android-devices.js
KrzysztofWojnar: yes I expect them to panic

scripts/e2e/turn-on-android-emulators.js
KrzysztofWojnar: added in the very last commit

scripts/e2e/turn-on-android-emulators.js
KrzysztofWojnar: I agree

scripts/e2e/android-devices.js
```diff
@@ -73,6 +73,21 @@ function bootInactiveEmulators() {
   }
 }
 
+/**
+ * @param {[string, string]} deviceIdAndState
+ */
+function passOnlyReadyDevices(deviceIdAndState) {
```
t0maboro: maybe `isInDeviceState` ?

scripts/e2e/android-devices.js
t0maboro: If it's expected it'd would be nice to clearly describe why we're expecting such behavior

kkafar:
Could we improve the PR description in such manner, that when coming back to the PR to track the changes it is clear why & for what reason the code was added?

scripts/e2e/android-devices.js
KrzysztofWojnar: fixed in a42d8c77fe3995982d122e396c6d9e4af75c8cbe

scripts/e2e/android-devices.js
KrzysztofWojnar: Oh, yes, now I see. it was clearly a mistake on my side

KrzysztofWojnar: changed in a321f5f01bdced8f627cb3649d9ac0eb4e1a4cfc

t0maboro: thanks!

scripts/e2e/android-devices.js
```diff
@@ -0,0 +1,146 @@
...
+      deviceName => !alreadyRunningDevices.has(deviceName),
+    );
+  } catch (e) {
+    if (e instanceof EmptyAttachedDeviceList) {
+      inactiveEmulators = allAvailableEmulatorNames;
+    } else throw e;
+  }
+
+  bootDevices(inactiveEmulators);
+}
```
kkafar: Booting all available emulators is really problematic. 

First, e. g. in my case I have over 10 emulators set up on machine, I don't want you to boot all of them simultaneously. I can easily image a scenario where my RAM usage explodes, especially having IDE instances opened at the same time. 

Second, it simply takes too long. Booting an emulator takes a minute. It will make the detox tests uncomfortable to use.

Third, API level of the device can be retrieved from its configuration files. 

Image

Not sure how reliable it is, though.

Issue #566, opened 2020-07-15
Redundant space between StatusBar and Stack header on Android


enheit:
Hi, after updating the expo to the version 38, I encountered with a strange issue on Android device. (Before SDK 38, everything worked as expected)

Whenever I use the [StatusBar](https://reactnative.dev/docs/statusbar) component from ReactNative it adds additional space between stack header and status bar.

With `StatusBar` component
![image](https://user-images.githubusercontent.com/8645216/87609003-0f637c80-c70a-11ea-97be-11c01523b985.png)

Without `StatusBar` component
![image](https://user-images.githubusercontent.com/8645216/87609101-4d60a080-c70a-11ea-81d2-900a4530a598.png)

The source code looks like this

```jsx
const Stack = createNativeStackNavigator();

export default App() {
  return (
    <SafeAreaProvider initialMetrics={initialWindowMetrics}>
      <StatusBar />
      <Stack.Navigator>
        <Stack.Screen name='Screen1' component={Screen1} />
      </Stack.Navigator>
    </SafeAreaProvider>
  )
}
```

I'm not sure if this kind of issue is related to the `react-native-screens` or to the `react-navigation` with `react-native-safe-area-context`. 

I was trying to create snack example but faced with issue which complains about react-native-screens imports.

![image](https://user-images.githubusercontent.com/8645216/87609423-2ce51600-c70b-11ea-9821-4ca457e30008.png)

enheit:
I resolved my question by reading the release notes of the `react-native-screens`.

> It also adds #545 by @brentvatne, which introduces additional padding to the header, which is a BREAKING CHANGE and will cause the header to have too much top padding in case of not having a translucent status bar.

See https://github.com/software-mansion/react-native-screens/releases/tag/2.9.0

So, I just set the `translucent` prop to `true` on the `StatusBar` component.

```jsx
const Stack = createNativeStackNavigator();

export default App() {
  return (
    <SafeAreaProvider initialMetrics={initialWindowMetrics}>
      <StatusBar translucent={true} /> {/* <--- Here */}
      <Stack.Navigator>
        <Stack.Screen name='Screen1' component={Screen1} />
      </Stack.Navigator>
    </SafeAreaProvider>
  )
}
```

I close this issue

PR #3430, opened 2025-11-27
chore(e2e): maintenance for automated tests after iOS 26.2

KrzysztofWojnar:
## Description

Closes https://github.com/software-mansion/react-native-screens-labs/issues/540

Broad changes after iOS 26.2 and some by-the-way fixes or improvements.
Not fixed tests:
iOS tests for example Test2809 should be aligned later with an output of PR #3303.
iOS tests for examples Test645 and Test649 are affected by issue [#566](https://github.com/software-mansion/react-native-screens-labs/issues/566)

## Changes

Detox e2e tests. Increased test isolation. Bar back button technical selector depends now on ios version.
Some android emulators my get a pop-up about using stylus which interrupts the test flow – it is disabled in the global config now. 

## Checklist

- [x] Updated TS types
- [x] Updated documentation: 
- [ ] Ensured that CI passes


Example/e2e/examplesTests/bottomTabs.e2e.ts
```diff
@@ -3,6 +3,7 @@ import { device, expect, element, by } from 'detox';
 describe('Bottom tabs and native stack', () => {
   beforeEach(async () => {
     await device.reloadReactNative();
+    await device.launchApp({ newInstance: true });
```
t0maboro: Q: Why do we need both? 1st we're reloading the same app instance, but in the next call we're killing the process and launching another one.

Example/e2e/examplesTests/events.e2e.ts
```diff
@@ -11,43 +11,33 @@ const pressBack = async () => {
...
+  for (const expectedEventNotiication of expectedEvents) {
+    await expect(element(by.text(expectedEventNotiication))).toExist();
   }
 };
 
 describe('Events', () => {
   beforeEach(async () => {
     await device.reloadReactNative();
-    // await device.launchApp({ newInstance: true });
+    await device.launchApp({ newInstance: true });
```
t0maboro: same question as above

FabricExample/e2e/component-objects/back-button.ts
```diff
@@ -0,0 +1,31 @@
...
+  const iosVersion = semverCoerce(getIOSVersion().replace('iOS', ''))!;
+  if (semverSatisfies(iosVersion, '>=26.0')) {
+    const elementsByAttributes =
+      (await backButtonElement.getAttributes()) as unknown as {
+        elements: { className: string }[];
+      };
+    const elements = elementsByAttributes.elements;
+    if (Array.isArray(elements)) {
+      return backButtonElement.atIndex(
+        elements.findIndex(elem => elem.className === '_UIButtonBarButton'),
```
t0maboro: what will happen when it doesn't find index?

t0maboro: _UIButtonBarButton might be extracted as constant

t0maboro (commented):
left some questions

FabricExample/e2e/component-objects/back-button.ts
KrzysztofWojnar: ios development would have to go in an absurd direction, but ok, I can guard it

KrzysztofWojnar: extracted

Example/e2e/examplesTests/events.e2e.ts
KrzysztofWojnar: removed "await device.reloadReactNative();"

Example/e2e/examplesTests/bottomTabs.e2e.ts
KrzysztofWojnar: removed "await device.reloadReactNative();"

FabricExample/e2e/elements/back-button.ts
```diff
@@ -0,0 +1,35 @@
...
+    return backButtonElement.tap();
+  } else throw new Error(`Platform "${platform}" not supported`);
+}
+async function getIOSBackButton() {
+  const iosVersion = semverCoerce(getIOSVersion().replace('iOS', ''))!;
+  if (semverSatisfies(iosVersion, '>=26.0')) {
+    const elementsByAttributes =
+      (await backButtonElement.getAttributes()) as unknown as {
+        elements: { className: string }[];
+      };
```
kligarski, line 24: `await backButtonElement.getAttributes()` returns `Detox.IosElementAttributes | Detox.AndroidElementAttributes | {
    elements: Detox.IosElementAttributes[];
} | {
    elements: Detox.AndroidElementAttributes[];
}`, maybe we can handle this type somehow instead of forcing a type.

FabricExample/e2e/helpers/disableStylus.ts
```diff
@@ -0,0 +1,12 @@
+import { getCommandLineResponse } from "react-native-screens/scripts/e2e/command-line-helpers";
+import { device } from 'detox';
+
+export function disableStylusPopupOnAndroid() {
+  if (device.getPlatform() === 'ios') return;
+  try {
+    getCommandLineResponse(`adb -s ${device.id} shell settings put secure stylus_handwriting_enabled 0`);
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  } catch (_) {
```
kligarski:
```suggestion
  } catch {
```

FabricExample/e2e/component-objects/back-button.ts
```diff
@@ -0,0 +1,35 @@
...
+  const platform = device.getPlatform();
+  if (platform === 'ios') {
+    return (await getIOSBackButton()).tap();
+  } else if (platform === 'android') {
+    return backButtonElement.tap();
+  } else throw new Error(`Platform "${platform}" not supported`);
+}
+async function getIOSBackButton() {
+  const iosVersion = semverCoerce(getIOSVersion().replace('iOS', ''))!;
+  if (semverSatisfies(iosVersion, '>=26.0')) {
```
kligarski: I'm wondering if we need `semver` for this? On the other hand, it's just the example app so I guess that it's okay.

KrzysztofWojnar: I wanted to save some time and several lines of code by adding the dev dependency, but I agree - it is an overkill and we can parse MAJOR.MINOR (or just bare MAJOR) oursevles. I don't have a strong opinion about it.

FabricExample/e2e/elements/back-button.ts
KrzysztofWojnar: That's the tough one.
It is an iOS-version-specific, behavior-based implementation for a technical selector, BUT QUITE READABLE (comparing to my other options). The biggest problem with the 'className' field is that it is **not documented** (I'm using the Hyrum's law).
It may break in future versions, but I can't predict it and finding this header back button in another way (filtering by other properties) is equally uncertain (or may break even easier). I don't have more robust solutions, you can call it a skill issue.
This probably won't make you feel any better, but I don't like it either.

KrzysztofWojnar: Oh, I can use magic numbers alternatively (`backButtonElement.atIndex(0)`).
I bet most people would do this in this case.

FabricExample/e2e/elements/back-button.ts
```diff
@@ -0,0 +1,35 @@
...
+    const elements = elementsByAttributes.elements;
+    if (Array.isArray(elements)) {
+      const uiBarButtonIndex = elements.findIndex(
+        elem => elem.className === IOS_BAR_BUTTON_TYPE,
+      );
+      if (uiBarButtonIndex !== -1) {
+        return backButtonElement.atIndex(uiBarButtonIndex);
+      }
```
KrzysztofWojnar, line 32: Do we want to have an arbitrary fallback?
```suggestion
      } else {
        // Fallback
        return backButtonElement.atIndex(0);
      }
```

FabricExample/e2e/elements/back-button.ts
kligarski: Ok, I see what you mean. I think the current version is all right then.

FabricExample/e2e/elements/back-button.ts
kligarski: I don't think so, it's probably better to get an error so that we know that something has changed and we need to update back button retrieval logic.

FabricExample/e2e/component-objects/back-button.ts
kligarski: cc @kkafar

Example/e2e/examplesTests/bottomTabs.e2e.ts
```diff
@@ -2,7 +2,7 @@ import { device, expect, element, by } from 'detox';
 
 describe('Bottom tabs and native stack', () => {
   beforeEach(async () => {
-    await device.reloadReactNative();
+    await device.launchApp({ newInstance: true });
```
kkafar: I have a question here. Why do we relaunch application instead of reloading RN? This was done mostly to save CI time IIRC, so it is a subject to be changed, but I need reasoning here.

Example/e2e/examplesTests/events.e2e.ts
```diff
@@ -11,43 +11,32 @@ const pressBack = async () => {
...
+  for (const expectedEventNotification of expectedEvents) {
+    await expect(element(by.text(expectedEventNotification))).toExist();
   }
 };
 
 describe('Events', () => {
   beforeEach(async () => {
-    await device.reloadReactNative();
-    // await device.launchApp({ newInstance: true });
+    await device.launchApp({ newInstance: true });
```
kkafar: Same here, and in every other place, where it has been done.

FabricExample/e2e/component-objects/back-button.ts
kkafar: No, we don't need external dependency for this. Simple helper function is enough for that. Given what is going on with NPM & supply chain attacks recently -> lets avoid any unnecessary dependencies.

FabricExample/e2e/elements/back-button.ts
kkafar, line 1: I wonder about naming here. What are `component-objects`? Shouldn't be this kind of util directory with `locators.js` / `actions.js`?

FabricExample/e2e/component-objects/back-button.ts
kkafar: Does this function work only on iOS 26+?

FabricExample/e2e/issue-tests/Test2926.e2e.ts
```diff
@@ -26,7 +27,7 @@ describeIfiOS('Test2926', () => {
     await element(by.type('UISearchBarTextField')).replaceText('Item 2');
     await element(by.id('home-button-open-second')).tap();
 
-    await element(by.id('BackButton')).tap();
+    await tapBarBackButton();
```
kkafar, line 30: This is an iOS only test. We have nice & clean way to detect a back button & tap it & we replace it with the complicated logic in `component-objects/back-button.ts`? Why do we do this? My guess would be that this version does not work, but I need confirmation & explanation.

Example/e2e/examplesTests/bottomTabs.e2e.ts
KrzysztofWojnar: Yes, I discovered that the Example ("Paper") app behaves differently: the app may start ignoring touch events (the problem is most clearly visible on the SettingsPicker from [apps\src\screens\Events.tsx:77](https://github.com/software-mansion/react-native-screens/blob/fb49fca61630a0893d743d19d2dab77ff5d06171/apps/src/screens/Events.tsx#L77), unluckly I _lost_ the video with the proof). I don't know if the issue is in the react-native-screens or just in the test app. The logcat looked fine. I didn't have time to reproduce the exact conditions of the problem. But it [always] appeared only in these test suites (in different test cases) and only with automated executions (I failed with reproducing it manually, even accidentally).

Existing tests passed, when I added `device.launchApp({ newInstance: true });` which increases their isolation.

FabricExample/e2e/component-objects/back-button.ts
KrzysztofWojnar: I am not able to push to this repository anymore, so I applied this suggestion in the fork:
#3451
(If I meesed up something with github like forking and pull requesting, please let me know too)

KrzysztofWojnar: This function _exists_ for iOS26+ but returns just the `element(by.id('BackButton'))` for other versions (line 34)

FabricExample/e2e/issue-tests/Test2926.e2e.ts
KrzysztofWojnar: yes, selector `element(by.id('BackButton'))` resolves to two elements since iOS 26+ so taping on it with Detox causes an error.

FabricExample/e2e/elements/back-button.ts
KrzysztofWojnar: yes, we can discuss that:
One of the top industry standards for e2e and ui automated tests is Page Object Model (a.k.a. Page Object Pattern) which separates behavior from a technical selectors and operations. This pattern is about having separate

test case files (the higher abstraction level, ideally even without touching automation library, like Selenium, Playwright, Detox etc),
"Page Object" files which define functions used be test cases (a middle abstraction layer that "maps" the elements and actions needed to execute the test with the automation library actions, so basically hiding all "technicall" elements(by.id(... and .tap() from test scenario)
and of course autamation library is the lowest abstraction layer in this pattern.

I tried also to re-invent wheel here as these Page Object files group actions and selectors depending on the page on which they are available (e.g. login/email and password inputs would be in the page-object/login.ts file) which is not usefull for us. I came up with an idea to shift focus on reusable componenst rather than pages...
Right now, there's only this header back button.
Should I move it and change the name (like e2e/component-objects/back-button.ts -> e2e/actions.ts)?

KrzysztofWojnar:
@kkafa
> Hey! Thanks for the PR. I've left few questions & remarks. Please answer them (or let us know that the PR needs an assignee).

Adding an assignee could be helpful: I can answer in threads but this PR belongs directly to react-native-screens and I cannot change its code anymore. That means I need someone else to resolve conflicts (it seems we have some already) and merge the PR at the end. I can apply suggestions only by PRs from a fork (like I do in #3451).

The alternative is to close this PR and I could open a new one - entirely as a fork PR

kmichalikk:
Cleaned up the PR a bit
- removed linter fixes as they are a bit distracting (may be done in another PR)
- refactored the helper classes
- we need to decide whether we want `component-object/` or just `helpers/`, and how to name those helpers (cc @kkafar)
- reverted https://github.com/software-mansion/react-native-screens/pull/2999 to see if the test are working

~I keep the simulated scrolling instead of searching on iOS for now - this PR focuses only on making the CI complete successfully. This will be rewritten in a followup after https://github.com/software-mansion/react-native-screens/pull/3532 and I will make the search work there.~ I'm rebasing to the other PR and fixing the search here.

FabricExample/e2e/issuesTests/Test432.e2e.ts
```diff
@@ -26,7 +27,7 @@ describe('Test432', () => {
     await expect(element(by.id('details-headerRight-red'))).toBeVisible(100);
 
     if (device.getPlatform() === 'ios') {
-      await element(by.id('BackButton')).tap();
+      await tapBarBackButton();
     } else {
       await device.pressBack();
     }
```
kmichalikk, line 33: the whole thing should be put inside `tabBarBackButton()`. Will to that tomorrow.

kmichalikk:
Okay, so the CI passes, but there are lots of skipped tests (and I'm adding another one here, Test2809 which should be updated after https://github.com/software-mansion/react-native-screens/pull/3303 anyways). Also there are .js files which are imported in TS and need .d.ts declarations - should we just rewrite them to TS? And there is still the question about file structure (see the comment above).

FabricExample/e2e/elements/back-button.ts
kkafar: Today I learned. Thanks! I did that intuitively & did not know "formal nomenclature" here. 

Having that said, I understand the approach, but at the same time I don't think we need to apply it to directory naming, especially that we don't use it in its exact phrasing, e.g. "pom" or `page-objects` but rather we call it `component-objects` which sounds confusing anyway. 

I think we should either name it directly "page-objects" or use detox nomenclature - "locators" / "selectors" or whatever it is called there.

kkafar: @kmichalikk would you mind following up on that?

FabricExample/e2e/helpers/disableStylusPopupOnAndroid.ts
```diff
@@ -0,0 +1,11 @@
+import { getCommandLineResponse } from "../../../scripts/e2e/command-line-helpers";
+import { device } from 'detox';
+
+export default function disableStylusPopupOnAndroid() {
```
kkafar, line 11: This method lacks a comment why it is required exactly. I can guess what's the reason behind it, but would be nice to have it stated explicitly, since it is not obvious.

My guess is to make sure that a keyboard opens on Android when a e.g. text field is focused, instead of Android thinking that input will come from stylus. This is only a guess, however.

FabricExample/e2e/issuesTests/Test432.e2e.ts
kkafar: I also think this should be refactored out. Not sure whether `tabBarBackButton` is the same as pressing back, but moving it to a separate meaningful function definitely makes sense.

FabricExample/e2e/e2e-utils.ts
```diff
@@ -18,7 +18,10 @@ export async function selectTestScreen(screenName: string) {
     // More details: https://github.com/software-mansion/react-native-screens/pull/2919
     await device.pressBack();
   } else {
-    await element(by.type('UISearchBarTextField')).replaceText(screenName);
+    await waitFor(element(by.id('root-screen-tests-' + screenName)))
+      .toBeVisible()
+      .whileElement(by.id('root-screen-examples-scrollview'))
+      .scroll(600, 'down', Number.NaN, 0.85);
```
kkafar: This logic should be abstracted away to some helper function. Its goal is to find an element with given name. Whether we do it through search bar or by scrolling the scrollview directly is an implementation detail and should be abstracted to separate function, e.g. `findElementForScreenWithName(screenName: string)` or something similar.

scripts/e2e/command-line-helpers.d.ts
```diff
@@ -0,0 +1 @@
+export function getCommandLineResponse(command: string): string;
```
kkafar: Let's avoid using declaration files. When I needed to type something before I've just used JSDoc. This should be enough.

scripts/e2e/ios-devices.d.ts
```diff
@@ -0,0 +1,2 @@
+export function resolveAppleSimulatorName(): string;
+export function getIOSVersion(): string;
```
kkafar: Similarly here - let's use JSDoc

kkafar: I mean, I don't mind which solution we use, but let's just be consistent and either do JSDoc or TS declaration files. 

I guess since we have already multiple scripts that use JSdoc I'd be easier to keep that approach.

FabricExample/e2e/elements/back-button.ts
KrzysztofWojnar: detox uses "matchers" (and I don't like it, because e.g. JUnit "Matcher" is something **completely** different), so I suggest "elements" (still present in Detox naming) or "selectors"/"locators" as you suggested. It's up to you, I don't have a strong opinion about it and I agree, it sounds better than `component-objects`

FabricExample/e2e/helpers/disableStylusPopupOnAndroid.ts
KrzysztofWojnar:
```suggestion
/**
 * Devices that support this feature may display an introductory popup to enable stylus input as the default input method. This popup is not a subject of react-native-screens testing and breaks the react-native-screens e2e test flow.
 * This function disables the display of the pop-up window by modifying a special system flag.
 */
export default function disableStylusPopupOnAndroid() {
```

scripts/e2e/command-line-helpers.d.ts
kmichalikk: [18bc5b2](https://github.com/software-mansion/react-native-screens/pull/3430/commits/18bc5b2f7ef3dc203597340189c610745af8a79b)

scripts/e2e/ios-devices.d.ts
kmichalikk: [18bc5b2](https://github.com/software-mansion/react-native-screens/pull/3430/commits/18bc5b2f7ef3dc203597340189c610745af8a79b)

FabricExample/e2e/e2e-utils.ts
kmichalikk: It actually simplified significantly and it looks better for iOS than for android. The only thing is to scroll to the [Legacy Tests], then click it and use the search - and it works for both platforms (the code diverges but it's 2 lines for each platform).

[6292fb3](https://github.com/software-mansion/react-native-screens/pull/3430/commits/6292fb3c7689f31237e160a196d50a5ed5b4d578)

FabricExample/e2e/issuesTests/Test432.e2e.ts
kmichalikk: I'm backing off from the idea because this applies only to this test and is not a general routine. For most of the cases, tabBarBackButton should be enough.

FabricExample/e2e/elements/back-button.ts
kmichalikk: I'd leave it as "elements" because we're probably only put there functions that perform some actions tied to some elements, i.e. "tap back button". Selectors and locators suggest that only functions that find elements would be placed here, and page-objects is too formal for our needs.

apps/src/tests/issue-tests/Test593.tsx
```diff
@@ -378,115 +361,3 @@ function Another({ navigation }: Props) {
     </View>
   );
 }
-
```
kmichalikk, line 381: Removing this implementation of toast; we have a shared version which works the same, except for longer timeout and smaller padding (and this makes more toasts fit on the screen)

.github/workflows/ios-e2e-test-fabric.yml
```diff
@@ -1,10 +1,24 @@
...
+      - 'src/**'
+  push:
+    branches:
+      - main
   workflow_dispatch:
 jobs:
   test:
-    runs-on: macos-14
-    timeout-minutes: 60
+    runs-on: macos-26
```
kligarski, line 20: I'm wondering if we should test both iOS 18 and 26 with appropriate Xcode versions.

apps/src/tests/IssueTestsScreen.tsx
```diff
@@ -6,13 +6,12 @@ import {
 } from '@react-navigation/native';
 import * as Tests from './issue-tests';
 import { ScrollView, StyleSheet, useColorScheme } from 'react-native';
```
kligarski: nit: unused import (StyleSheet)

apps/src/tests/IssueTestsScreen.tsx
```diff
@@ -99,26 +95,19 @@ function MainScreen(props: {
...
+    <ScrollView
+      testID="issue-tests-scrollview"
+      contentInsetAdjustmentBehavior="automatic">
+      {filteredTests.map(name => (
+        <ListItem
+          key={name}
+          testID={`issue-tests-${name}`}
+          title={SCREENS[name].title}
+          onPress={() => navigation.navigate(name)}
+          disabled={false}
```
kligarski: nit: probably isn't necessary to pass `disabled=false`

Example/e2e/examplesTests/events.e2e.ts
```diff
@@ -11,43 +11,38 @@ const pressBack = async () => {
...
+    .toExist()
+    .withTimeout(5000);
+
+  for (const expectedEventNotification of expectedEvents) {
+    await expect(element(by.text(expectedEventNotification))).toExist();
   }
 };
 
-describe('Events', () => {
+describe.skip('Events', () => {
```
kligarski: Let's just add a comment here why we're skipping it + create an issue to investigate.

FabricExample/e2e/elements/back-button.ts
```diff
@@ -0,0 +1,36 @@
...
+  const platform = device.getPlatform();
+  if (platform === 'ios') {
+    return (await getIOSBackButton()).tap();
+  } else if (platform === 'android') {
+    return backButtonElement.tap();
+  } else throw new Error(`Platform "${platform}" not supported`);
+}
+
+async function getIOSBackButton() {
+  const iosVersion = getIOSVersion().replace('iOS', '').trim();
```
kligarski: Shouldn't we also have a util function for getting only the number that would do this string manipulation? Something like `getIOSVersionNumber()`. I'm not sure about calling it "number" if it's a string but you get the point.

kligarski (approved):
Looks good. Let's verify that CI passes.

.github/workflows/ios-e2e-test-fabric.yml
kmichalikk: We'll discuss this on the next sync meeting, but I don't want to do it in this PR.

Example/e2e/examplesTests/events.e2e.ts
kmichalikk: https://github.com/software-mansion/react-native-screens-labs/issues/953

kkafar (approved):
Let's align directory naming `issueTests` vs `issue-tests` etc. Beside that good job. Thank you!

I approve to remove merging blockade after I've requested changes. I have not done a proper review here.

PR #3432, opened 2025-11-28
fix(examples): Stop triggering CPU intensive tasks in top level scope in example apps

kkafar:
When `Test2933` is imported (loaded) we always trigger CPU intensive
tasks leading to laggy application.

I've added a button to trigger the computation.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/601


t0maboro (approved):
nice catch, thanks

PR #3433, opened 2025-11-28
feat: Remove native-stack v5 from screens repo

kmichalikk:
## Description

This PR removes native-stack directory from screens repo. This dir held the code for native-stack v5, which has been moved to react-navigation and is v7, v8 now. The rest of the deprecated imports are removed and the unused code is deleted.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/589

## Changes

Removed native-stack/*, fixed imports.

Check if the library still builds successfully.

### DIFF
--- a/apps/src/tests/index.ts
+++ b/apps/src/tests/index.ts
@@ -166,7 +166,9 @@ export { default as Test3379 } from './Test3379';
 export { default as Test3422 } from './Test3422';
 export { default as Test3425 } from './Test3425';
 export { default as TestScreenAnimation } from './TestScreenAnimation';
-export { default as TestScreenAnimationV5 } from './TestScreenAnimationV5';
+// The following test was meant to demo the "go back" gesture using Reanimated
+// but the associated PR in react-navigation is currently put on hold
+// export { default as TestScreenAnimationV5 } from './TestScreenAnimationV5';
 export { default as TestHeader } from './TestHeader';
 export { default as TestPreload } from './TestPreload';
 export { default as TestActivityStateProgression } from './TestActivityStateProgression';


kkafar (commented):
Guide: this is a feature related change, as it modifies public interface of the library, therefore use `feat` as the commit type in the PR title.

apps/src/tests/TestScreenAnimationV5.tsx
kkafar, line 1: Does this test case even work? I think that we never downstreamed this feature to react-navigation. There was a PR from @maciekstosio, but it has not been completed. 

Please verify whether we have ticket for downstreaming it (I think I recently saw one on the board, it could have been a draft so it won't be visible in labs issues). If not -> create one. 

Now, regarding this test -> if it does not work, it should be "suspended". I'd recommend commenting it out in index file, so that it is not loaded or marking it somehow as "disabled" (can be done in separate, follow up PR).

src/types.tsx
```diff
@@ -610,8 +609,13 @@ export interface ScreenContainerProps extends ViewProps {
...
 
 export interface GestureDetectorBridge {
+  /**
+   * Callback to attach into ScreenStack's useEffect() from ScreenGestureDetector that wraps the stack.
+   *
+   * @param stackRef holds a reference to an instance of ScreenStackNativeComponent
+   */
   stackUseEffectCallback: (
-    stackRef: React.MutableRefObject<React.Ref<NativeStackNavigatorProps>>,
+    stackRef: React.MutableRefObject<React.Ref<View>>,
```
kkafar, line 618: This is potentially a breaking change. Given that we never really downstreamed the feature it is used for I give a green light and take the responsibility here.

kkafar: Let's make sure that this callback is not used e.g. in reanimated. Just search for the string please and let me know.

apps/src/tests/TestScreenAnimationV5.tsx
kmichalikk: This is the PR:
https://github.com/react-navigation/react-navigation/pull/12225

src/types.tsx
kmichalikk: https://github.com/search?q=org%3Asoftware-mansion%20stackUseEffectCallback&type=code

3 results; only screens

kkafar: looks legit, thanks

apps/src/tests/TestScreenAnimationV5.tsx
kmichalikk: https://github.com/software-mansion/react-native-screens-labs/issues/610

PR #3435, opened 2025-11-28
fix(Android, FormSheet): Fix pressables with TextInput for Fabric

t0maboro:
## Description

This PR addresses an issue where pressables became unresponsive when the keyboard was open and a FormSheet had been adjusted for keyboard height.

Previously, after introducing support for offsetting FormSheets by the keyboard height, I overlooked updating the corresponding ShadowNodes to reflect this layout change.

Note: this only applies on Fabric, where we need to pass offsetY to ShadowNodes.

Fixes: https://github.com/software-mansion/react-native-screens-labs/issues/607

## Changes

- Added a callback to notify and update ShadowNodes when the FormSheet layout shifts due to keyboard visibility.

## Screenshots / GIFs

### Before

https://github.com/user-attachments/assets/f5839b64-6a0c-43a7-ab27-7bb4f32aab05

### After

https://github.com/user-attachments/assets/db7ac121-74d9-4169-bd6c-665852351692

Updated `Test3336`

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -491,7 +491,7 @@ class Screen(
         // There is no need to update shadow state for transient sheet states -
         // we are unsure of the exact sheet position anyway.
         if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED && isStable) {
-            updateScreenSizeFabric(width, height, top + translationY.toInt())
+            onSheetYTranslationChanged()
         }
     }
 


kkafar (changes requested):
I'd like to align the approach here with what's already done in this field for regular sheet dragging. We only send updates with target values, skipping all intermediate states.

t0maboro:
@kkafar I updated the logic - sending an update when the slide-in animation ends & onEnd of WindowInsetsAnimationCallback

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -491,7 +491,14 @@ class Screen(
         // There is no need to update shadow state for transient sheet states -
         // we are unsure of the exact sheet position anyway.
         if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED && isStable) {
-            updateScreenSizeFabric(width, height, top)
+            updateScreenSizeFabric(width, height, top + translationY.toInt())
```
kligarski: For consistency, maybe we should also call `onSheetYTranslationChanged()` here.

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3435/commits/e286be20f5890c99bccf426b0b54f1eeb8873cff

PR #3440, opened 2025-12-01
feat(Tabs): special effects refactor

kligarski:
## Description

Refactors special effects on both platforms:

- fixes disabling `popToRoot` on iOS 26,
- fixes special effects running on regular (not repeated) tab change on Android,
- adds `repeatedSelectionHandledBySpecialEffect` to `NativeFocusChangeEvent` on both platforms (now the event is sent on every tab press with appropriate flag value),
- adds support for `specialEffects` on Paper+iOS.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/579.
Closes https://github.com/software-mansion/react-native-screens-labs/issues/574.

### Recordings

Pay attention to logs.

#### Android

https://github.com/user-attachments/assets/54887d06-f6fa-4bb9-a9bd-2642a8af31b2

#### iOS

https://github.com/user-attachments/assets/c06505dc-e4e6-44b9-90f3-5d18c3345c01

### Native pop to root on iOS 26

UIKit natively has "pop to root" effect in `UITabBarController` but prior to iOS 26, it worked only when `UINavigationController` was direct child of `UITabBarController`. In `screens`, `RNSTabBarController` has `RNSTabsScreenViewController` children view controllers and `RNSScreenStack` can be a child VC of `RNSTabsScreenViewController`. This prevents native effect from working on iOS versions prior to 26.

Starting from iOS 26, `UITabBarController` also detects nested `UINavigationController`s. This interaction happens when we return `true` from `tabBarController:shouldSelectViewController:` delegate method. This would work for natively-driven tabs but for controlled tabs, we need to return `false` to prevent tab change on the native side -> that's why we decided to use only our own implementation and disable native effect by always returning `false` from `tabBarController:shouldSelectViewController:` on repeated tab selection.

## Changes

- handle `specialEffects` prop on iOS, Paper,
- add `repeatedSelectionHandledBySpecialEffect` to `NativeFocusChangeEvent`
- update docs (move special effects and freeze to general section, clean up)
- return `false` from `tabBarController:(UITabBarController *)tabBarController
shouldSelectViewController:(UIViewController *)viewController` on repeated tab selection on iOS (disables native pop to root interaction which is available starting from iOS 26)
- prevent running `onItemSelectedListener` after menu selection on Android (it runs on click for the first time and it would run for the second time after update from JS which would activate special effect erroneously)

Run `TestBottomTabs`. Test repeated selection with different special effects enabled on `Tab4`. Make sure that special effects don't run when changing from other tab to `Tab4`.

## Checklist

- [x] Updated TS types
- [x] Updated documentation
- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/event/TabsHostNativeFocusChangeEvent.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/event/TabsHostNativeFocusChangeEvent.kt
@@ -17,6 +17,10 @@ class TabsHostNativeFocusChangeEvent(
 
     override fun getEventRegistrationName() = EVENT_REGISTRATION_NAME
 
+    // If the user taps currently selected tab 2 times and e.g. scroll to top effect can run,
+    // we should send 2 events [(tabKey, true), (tabKey, false)]. We don't want them to be coalesced
+    // as we would lose information about activation of special effect. That's why we take into
+    // account `repeatedSelectionHandledBySpecialEffect` for coalescingKey.
     override fun getCoalescingKey(): Short = (tabNumber * 10 + if (repeatedSelectionHandledBySpecialEffect) 1 else 0).toShort()
 
     override fun getEventData(): WritableMap? =


android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
```diff
@@ -352,8 +356,11 @@ class TabsHost(
 
         appearanceCoordinator.updateTabAppearance(this)
 
-        bottomNavigationView.selectedItemId =
+        val selectedTabScreenFragmentId =
             checkNotNull(getSelectedTabScreenFragmentId()) { "[RNScreens] A single selected tab must be present" }
+        if (bottomNavigationView.selectedItemId != selectedTabScreenFragmentId) {
+            bottomNavigationView.selectedItemId = selectedTabScreenFragmentId
+        }
```
kkafar, line 363: Does this change anything? Any side effects of avoiding the setter here?

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/event/TabsHostNativeFocusChangeEvent.kt
```diff
@@ -9,25 +9,32 @@ class TabsHostNativeFocusChangeEvent(
...
+    val repeatedSelectionHandledBySpecialEffect: Boolean,
 ) : Event<TabScreenDidAppearEvent>(surfaceId, viewId),
     NamingAwareEventType {
     override fun getEventName() = EVENT_NAME
 
     override fun getEventRegistrationName() = EVENT_REGISTRATION_NAME
 
-    // All events for a given view can be coalesced.
-    override fun getCoalescingKey(): Short = 0
+    override fun getCoalescingKey(): Short = (tabNumber * 10 + if (repeatedSelectionHandledBySpecialEffect) 1 else 0).toShort()
```
kkafar, line 24: What's the reasoning here? Why different key when `repeatedSelectionHandledBySpecialEffect`? 

FYI: I don't think this works anymore, as recently discovered by @t0maboro.

ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
```diff
@@ -649,6 +649,30 @@ - (void)setSystemItem:(RNSBottomTabsScreenSystemItem)systemItem
...
+  } else {
+    _shouldUseRepeatedTabSelectionPopToRootSpecialEffect = YES;
+  }
+
+  if (repeatedTabSelection[@"scrollToTop"] != nil) {
+    _shouldUseRepeatedTabSelectionScrollToTopSpecialEffect = [RCTConvert BOOL:repeatedTabSelection[@"scrollToTop"]];
+  } else {
+    _shouldUseRepeatedTabSelectionScrollToTopSpecialEffect = YES;
+  }
+}
```
kkafar, line 674: Could you explain why is this code added & how has it worked before? Is this Paper only?

kkafar: Oh, it seems that this is paper specific.

kkafar: Ok, it makes sense now.

ios/bottom-tabs/screen/RNSBottomTabsScreenComponentViewManager.mm
```diff
@@ -33,9 +33,6 @@ - (UIView *)view
 RCT_EXPORT_VIEW_PROPERTY(selectedIconImageSource, RCTImageSource);
 RCT_EXPORT_VIEW_PROPERTY(selectedIconSfSymbolName, NSString);
 
-RCT_EXPORT_VIEW_PROPERTY(shouldUseRepeatedTabSelectionPopToRootSpecialEffect, BOOL);
-RCT_EXPORT_VIEW_PROPERTY(shouldUseRepeatedTabSelectionScrollToTopSpecialEffect, BOOL);
```
kkafar, line 37: I don't get what's going on here. I don't see these properties being removed from JS. Were they hollow & not used before? Please elaborate

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
kligarski: `bottomNavigationView` has an `onItemSelectedListener`. It activates:
1. when you tap the tab bar item on the screen,
2. when you set `bottomNavigationView.selectedItemId`.

In current implementation, we allow the instant tab bar item change on tap, then we send an event to JS. After that, we receive focused tab change from JS - if we set `bottomNavigationView.selectedItemId` when it's the same id as before, `onItemSelectedListener` runs again, now detecting repeated selection (`fragment == currentFocusedTab`) -> this activates special effects which is incorrect behavior.

kligarski: I clicked through `TestBottomTabs` and did not find any regressions, I hope that we don't rely on this behavior in any way.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/event/TabsHostNativeFocusChangeEvent.kt
kligarski: If the user taps currently selected tab 2 times and e.g. scroll to top effect can run, we should send [(tabKey, true), (tabKey, false)]. If I understand correctly, coalescing key is used to combine events to prevent spamming them. If those 2 events were combined to only one, we would lose some information that somebody can rely on (e.g. running some action at the same time as special effect or after special effect).

> FYI: I don't think this works anymore, as recently discovered by @t0maboro.

Just wanted to make sure that it works correctly when it gets fixed 😅

ios/bottom-tabs/screen/RNSBottomTabsScreenComponentViewManager.mm
kligarski: We send `specialEffects: { repeatedTabSelection: { popToRoot: boolean; scrollToTop: boolean }}` from JS. In native component, this is flattened to 2 boolean properties (not sure what was the reasoning here, probably it was just easier; also back then we thought that tabs will be Fabric-only).

It seems like we missed that in [Paper support implementation](https://github.com/software-mansion/react-native-screens-labs/pull/273/files#diff-7da69f520c0dd82b41aeea69142bfc51a4487bdc3478c00b80427bc7f3949065R40-R41) and tried to export flattened properties (but JS is unaware of them).

ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
kligarski: Details here: https://github.com/software-mansion/react-native-screens/pull/3440#discussion_r2580674499

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
kkafar: I see now. 

No, this changes makes sense. Thanks.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/event/TabsHostNativeFocusChangeEvent.kt
kkafar: This convinces me. I think this deserves code comment. Let's make one, please.

ios/bottom-tabs/screen/RNSBottomTabsScreenComponentViewManager.mm
kkafar: makes sense now

kkafar (approved):
Okay, looks good! Thank you!

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/event/TabsHostNativeFocusChangeEvent.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3440/commits/efea3cc2de5f7f5de189f1b9b961c75fe43d72e3

PR #3441, opened 2025-12-01
fix(iOS, Tabs): `overrideScrollViewContentInsetAdjustmentBehavior` in managed tabs

kligarski:
## Description

Fixes `overrideScrollViewContentInsetAdjustmentBehavior` in managed tabs on iOS, Fabric.

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/c3ba2519-c710-4348-836c-b06d5ccb2c2e" /> | <video src="https://github.com/user-attachments/assets/d593dfcc-e758-406b-8a1c-d8ab89ac1303" /> |

### Details

Previously, `overrideScrollViewBehaviorInFirstDescendantChainIfNeeded` was called in `updateSelectedViewController` which is called when `reactMountingTransactionDidMount`. This worked without any problems with controlled tabs as tab changes happen only via `updateSelectedViewController` in reaction to update from JS (the native tab change is prevented by returning `NO` from `tabBarController:shouldSelectViewController:` and appropriate event is sent to JS).

In managed tabs, native tab change is not prevented and screen view controller is selected immediately. Tab screen is temporarily empty before content is mounted. When the content is mounted, it immediately becomes visible. After transaction had been mounted, we would run `overrideScrollViewBehaviorInFirstDescendantChainIfNeeded` but this was too late - screen's content was already visible so overriding behavior wouldn't change initial scroll position.

To provide a solution that will work in both control modes, we decided to run `overrideScrollViewBehaviorInFirstDescendantChainIfNeeded` exactly when child is mounted to tab screen. This works on Fabric because children are mounted from bottom to top - when child is mounted to tab screen, we're sure that the entire hierarchy below it already exists.

### Paper

On Paper, due to the order of mounting children, we are unable to consistently find and override scroll view behavior in time (when the entire hierarchy is mounted under screen but before screen is shown). The best solution I found (`dispatch_async` in `didUpdateReactSubviews`) worked 50% of the time.

For now, we decided not to implement support for the prop on Paper. 

Ticket to research Paper support: https://github.com/software-mansion/react-native-screens-labs/issues/283.

## Changes

- move finding ScrollView and applying behavior override + scroll edge effects to `mountChildComponentView`

Run `TestBottomTabs`. Change Tab4 to use transparent header on first screen. Open Tab3, Tab4. ScrollViews should respect safe area insets.

Use controlled and managed bottom tabs, with and without freeze.

## Checklist

- [x] Updated documentation
- [x] Ensured that CI passes

### DIFF
--- a/ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
+++ b/ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
@@ -456,8 +456,13 @@ RNS_IGNORE_SUPER_CALL_END
 {
   RNSLog(@"TabScreen [%ld] mount [%ld] at %ld", self.tag, childComponentView.tag, index);
   [super mountChildComponentView:childComponentView index:index];
-  [self overrideScrollViewBehaviorInFirstDescendantChainIfNeeded];
-  [self updateContentScrollViewEdgeEffectsIfExists];
+
+  // overrideScrollViewBehavior and updateContentScrollViewEdgeEffects use first descendant chain
+  // from screen to find ScrollView, that's why we're only interested in child mounted at index 0.
+  if (index == 0) {
+    [self overrideScrollViewBehaviorInFirstDescendantChainIfNeeded];
+    [self updateContentScrollViewEdgeEffectsIfExists];
+  }
 }
 
 - (void)unmountChildComponentView:(UIView<RCTComponentViewProtocol> *)childComponentView index:(NSInteger)index


kligarski:
I updated the PR description and added an optimization: run override logic only when child is mounted at index 0 (both `overrideScrollViewBehaviorInFirstDescendantChainIfNeeded` and `updateContentScrollViewEdgeEffectsIfExists` use first descendant chain from screen to find `ScrollView`).

PR #3405, opened 2025-11-19
feat(Android, SAV+Stack): add support for SAV in Stack

kligarski:
## Description

Adds support for SAV in Stack on Android. Fixes handling insets by the header (`CustomToolbar`).

https://github.com/user-attachments/assets/bfe6817b-7918-4ffb-95f0-f5718254e31e

Fixes https://github.com/software-mansion/react-native-screens/issues/3398.
Closes https://github.com/software-mansion/react-native-screens-labs/issues/464.

## Changes

- use different insets listener API based on Android SDK (<30/>=30) to correctly handle consuming display cutout inset
- override `ScreensCoordinatorLayout`'s `dispatchApplyWindowInsets` to manage order and flow of inset consumption
- add `screenInsets` in `CustomToolbar` to store insets designated for Screen component
- fix bug with missing `menuView` in `CustomToolbar` when used in nested stack
- add `Test3405`

## Test code and steps to reproduce

Run `Test3405`.

## Checklist

- [x] Included code example that can be used to test this change
- [x] Ensured that CI passes

PR #3442, opened 2025-12-01
fix(Android, Stack, Fabric): Fix jumping content in nested stack for Fabric

t0maboro:
## Description

This PR introduces a set of fixes and manual layout adjustments to address layout/content jumps in Fabric. I listed them under `Changes` section with some motivations.

> [!CAUTION]
> This PR aims to resolve the issue only on Fabric; on Paper some additional effort is needed and (potentially) would require adapting `ScreenDummyLayoutHelper` to old arch. The work will be continued in a separate PR.

## Changes

- Added `CustomAppBarLayout`: This custom implementation triggers a layout refresh on the `AppBarLayout` when we apply padding adjustments to the `CustomToolbar`. `AppBarLayout` just dispatches insets to `Toolbar`; therefore, I'm still handling them at the `Toolbar` level.
- Ensured early initialization of `ActionMenuView` in the `Toolbar`: Initializing the `ActionMenuView` early in the layout process prevents inconsistencies and avoids fallback to the system default toolbar height in `onMeasure`, ensuring a consistent layout.
- Manual shadow offset correction in the `Screen`: I'm applying some temporary padding corrections at the `Screen` level until insets are applied. 
- Layout correction for `ScreenDummyLayoutHelper`: A similar early correction approach is used for `ScreenDummyLayoutHelper` to keep the consistency for the header size before full layout pass (this one with insets applied) is complete.
- Added `DecorViewInsetHelper`: React's layout arrives before insets have been applied, leading to content jumps. To counter this, we rely on insets available from the `DecorView` (child of `RootView`). A dedicated helper class extracts these inset values. Additionally, it ensures that the insets are applied only on the topmost Screen with Toolbar, preventing the unnecessary padding on the nested Toolbars.

## Screenshots / GIFs

### Before
<table>
  <tr>
    <td width="50%">
      <video src="https://github.com/user-attachments/assets/8c16b90c-2809-4ed3-8a8e-f2fc2a4a3827"></video>
    </td>
    <td width="50%">
      <video src="https://github.com/user-attachments/assets/12e82a10-1ace-462a-a1cb-3fed9b17d246"></video>
    </td>
  </tr>
</table>

### After
<table>
  <tr>
    <td width="50%">
      <video src="https://github.com/user-attachments/assets/1706b298-5a93-4175-acec-832341598db8"></video>
    </td>
    <td width="50%">
      <video src="https://github.com/user-attachments/assets/596c6797-94a0-49aa-913e-42f80036a22d"></video>
    </td>
  </tr>
</table>

Added Test3006, tested with API levels above and below 30, with both Status Bar and Display Cutout. As mentioned earlier only testing on FabricExample is in the scope of this PR.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
@@ -79,6 +79,8 @@ open class CustomToolbar(
         val maybeAppBarLayout = parent as? CustomAppBarLayout
         maybeAppBarLayout?.let {
             if (shouldApplyLayoutCorrectionForTopInset && !it.isInLayout) {
+                // In `applyToolbarLayoutCorrection`, we call and immediate layout on AppBarLayout
+                // to update it right away and avoid showing a potentially wrong UI state.
                 it.applyToolbarLayoutCorrection(paddingTop)
                 shouldApplyLayoutCorrectionForTopInset = false
             }


android/src/main/java/com/swmansion/rnscreens/utils/DecorViewInsetsUtils.kt
```diff
@@ -0,0 +1,32 @@
...
+            insets.systemWindowInsetTop
+        }
+    }
+
+    @RequiresApi(Build.VERSION_CODES.R)
+    private fun getTopInsetModern(insets: WindowInsets): Int {
+        val systemBarsTop = insets.getInsets(WindowInsets.Type.systemBars()).top
+        val cutoutTop = insets.displayCutout?.safeInsetTop ?: 0
+        return maxOf(systemBarsTop, cutoutTop)
+    }
```
kligarski: We might try to use `WindowInsetsCompat` here and I think that you can do `insets.getInsets(WindowInsetsCompat.Type.systemBars() | WindowInsetsCompat.Type.displayCutout()).top` to get both at the same time (but I might be wrong).

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
```diff
@@ -179,11 +179,16 @@ internal class ScreenDummyLayoutHelper(
         }
 
         val topLevelDecorView = requireActivity().window.decorView
+        val topInset = DecorViewInsetsUtils.getDecorViewTopInset(topLevelDecorView)
 
         // These dimensions are not accurate, as they do include status bar & navigation bar, however
         // it is ok for our purposes.
```
kligarski: Is this comment still correct?

t0maboro:
> Did you test how this behaves with fixes from #3405 and nested stacks in general?

Good callout about 3405 I tested with and without SAV for nested stacks, results look legit

https://github.com/user-attachments/assets/624fa5dc-efdd-4018-9f02-c9c92bce21a0

https://github.com/user-attachments/assets/61edd49d-4c31-4f32-a9b8-e9e06ada13f5

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3442/commits/59a9a9f9bf8ff7b087942339f9d450315699cd78

android/src/main/java/com/swmansion/rnscreens/utils/DecorViewInsetsUtils.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3442/commits/a4df45a283a6e1859c5981a5cb1681c6e832ccba

JoniVR:
This resolved all my jumping issues on a rather complex split-view tablet Android application. I have applied these changes as a patch for now. Thanks for this!!

android/src/main/java/com/swmansion/rnscreens/utils/DecorViewInsetsUtils.kt
```diff
@@ -0,0 +1,25 @@
...
+
+        return getTopInset(insetsCompat)
+    }
+
+    private fun getTopInset(insetsCompat: WindowInsetsCompat): Int =
+        insetsCompat
+            .getInsets(
+                WindowInsetsCompat.Type.systemBars() or WindowInsetsCompat.Type.displayCutout(),
+            ).top
+}
```
kkafar, line 17: Let's remove the object and make it plain top-level functions.

I know that we've used `object` in few other places, but that was a mistake. Object introduces a singleton instance. We don't keep any shared state, we're interested only in namespace. We can achieve this with `import ... as ...`.

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
```diff
@@ -179,11 +179,16 @@ internal class ScreenDummyLayoutHelper(
...
-        // These dimensions are not accurate, as they do include status bar & navigation bar, however
+        // These dimensions are not accurate, as they do include navigation bar, however
         // it is ok for our purposes.
         val decorViewWidth = topLevelDecorView.width
-        val decorViewHeight = topLevelDecorView.height
+        // Adjust the height of the top-level decor view by subtracting the top inset.
+        // This accounts for edge-to-edge support where we need to manually do frame correction
+        // for AppBarLayout. To ensure proper layout for the Screen, we manually downscale the content
+        // to accommodate the extra padding.
+        val decorViewHeight = topLevelDecorView.height - topInset
```
kkafar: Does this impact the layout result in any way? It seems to me, that the computed header height would be independent of the `decorViewHeight` as long as `decorViewHeight` is high enough and it does not limit the header.

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
```diff
@@ -208,7 +213,11 @@ internal class ScreenDummyLayoutHelper(
         // scenarios when layout violates measured dimensions.
         coordinatorLayout.layout(0, 0, decorViewWidth, decorViewHeight)
 
-        val headerHeight = PixelUtil.toDIPFromPixel(appBarLayout.height.toFloat())
+        // Include the top inset to account for the extra padding manually applied to the CustomToolbar.
+        // This ensures the total height calculation of the AppBarLayout is accurate in edge-to-edge layouts.
+        val totalAppBarLayoutHeight = appBarLayout.height.toFloat() + topInset
+
+        val headerHeight = PixelUtil.toDIPFromPixel(totalAppBarLayoutHeight)
```
kkafar, line 215: Using words like "ensures" in this context is courageous 😄

android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
```diff
@@ -55,6 +75,15 @@ open class CustomToolbar(
 
     override fun requestLayout() {
         super.requestLayout()
+
+        val maybeAppBarLayout = parent as? CustomAppBarLayout
+        maybeAppBarLayout?.let {
+            if (shouldApplyLayoutCorrectionForTopInset && !it.isInLayout) {
+                it.onToolbarLayout(paddingTop)
```
kkafar: Reading this I have a problem with naming here. We call `onToolbarLayout` in `requestLayout`. This clearly contradicts. 

Why do we call this callback in `requestLayout`? Shouldn't it be called in `onLayout`?

kkafar: By doing it this way, we call `onToolbarLayout` *before* the layout triggered by this `requestLayout` call happens. Is this intentional?

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -191,7 +195,20 @@ class Screen(
...
+                val topLevelDecorView =
+                    requireNotNull(
+                        reactContext.currentActivity?.window?.decorView,
+                    ) { "[RNScreens] DecorView is required for applying inset correction, but was null." }
+
+                val topInset = DecorViewInsetsUtils.getDecorViewTopInset(topLevelDecorView)
+                val correctedHeight = height - topInset
+                val correctedOffsetY = t + topInset
+
+                dispatchShadowStateUpdate(width, correctedHeight, correctedOffsetY)
```
kkafar, line 208: This code assumes the top insets won't be consumed along the way right? 

Do we still experience problems in case of `Tabs > Stack > Stack`? I know this is exotic scenario, but just so that we are aware.

kkafar (commented):
Great job here. Looks as hacky as I expected after our initial discussions 😄 

Asked few questions & added remarks. Please answer them.

android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
t0maboro: > By doing it this way, we call onToolbarLayout before the layout triggered by this requestLayout call happens. Is this intentional? 

yes, in CustomAppBarLayout I'm calling `layout` explicitly to make a correction asap, without any overhead regarding calling `requestLayout`

android/src/main/java/com/swmansion/rnscreens/utils/DecorViewInsetsUtils.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3442/commits/851c345d65bda60033ace43eee8dabb44cdd5fd6

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3442/commits/13bded578c6e0d8734120ea982135a4efc3d79f1

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3442/commits/07101c685d6ef2bdad7b5319ec11b03b3f37053c

android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3442/commits/b4247d5e77e502c69890d61172307332c2dd6e57

android/src/main/java/com/swmansion/rnscreens/Screen.kt
t0maboro: > This code assumes the top insets won't be consumed along the way right?

Yes

> Do we still experience problems in case of Tabs > Stack > Stack? I know this is exotic scenario, but just so that we are aware.

I haven't checked that case, testing it right now, will send an update soon here

t0maboro: Okay, the content is not jumping, the inset is not consumed and the header is applying extra padding - the last one shouldn't be needed for the nested stack, we don't need to offset the header. Need to test some more scenarios, because when only the nested header is shown, then we need to apply the update for the nested stack. Investigating the solution how to apply the correction to the top-level header only ..

https://github.com/user-attachments/assets/0590ae95-c2e5-4471-92d9-55cb223651eb

t0maboro: ~https://github.com/software-mansion/react-native-screens/pull/3442/commits/cf775238d2d44375560366c8bf8cb7196a8f870a~
EDIT: sorry, this one doesn't work

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3442/commits/c7895b1ba35d706073b57a017414065a921984c0 this one

t0maboro: as discussed internally, this shouldn't be part of this PR, I'm reverting it and creating myself a ticket for handling double insets

t0maboro: https://github.com/software-mansion/react-native-screens-labs/issues/700

android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
kkafar: Okay. Can we add comment documenting this? That we call layout here, to ensure immediate frame update & prevent potentially invalid state being displayed before requested layout is executed.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
kkafar: Thanks
x

android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3442/commits/955d5884b804b2f05514e0340719ee697bffeebb

JoniVR:
@t0maboro 

One issue I'm facing now every once in a while is that the app crashes with the Error:
`[RNScreens] DecorView is required for applying inset correction, but was null.`,
which was added in this PR.

This crash always happens when my app is in the background. I use a background service for location tracking of assets. My assumption is that this is due to the fact that the decorView can be null when the app is in the background, and if it executes this logic in the background, it should probably handle that gracefully instead of throwing an exception.

t0maboro:
Thanks for reporting it. I'll try to reproduce the crash on my side. The check for the DecorView being non-null was added intentionally, as it plays important role in synchronizing layout across multiple components. That's why we made the check strict - to ensure consistency and prevent layout bugs that can be very difficult to debug later.

That said, I'll investigate whether this scenario can be handled gracefully without breaking the layout completely. It's not yet clear to me if allowing DecorView to be null would be safe, or if we'd need to rethink the layout handling in such edge cases.

t0maboro:
Hi @JoniVR , 

Unfortunately I haven't been able to reproduce it when the app is in the background. I managed to find a "hacky" way during hot reload where the check against DecorView fails - however, testing in that setup isn't reliable, as it causes other issues related to the synchronization in the ShadowTree.

For now, I'm opening a draft: https://github.com/software-mansion/react-native-screens/pull/3554 that introduces potential fallbacks which could replace the strict check. Would it be possible for you to apply this as a patch and see if this approach helps in your case? Thanks in advance!

JoniVR:
Hi @t0maboro thanks for following up. I will apply this patch and keep you posted!

PR #3443, opened 2025-12-01
feat(iOS): add xcassets support

ajanuar:
## Description

This PR adds support for using .xcassets asset catalog icons across key navigation UI elements. Developers can now reference native asset names directly for both bottom-tab icons and header bar button items, removing the need to bundle standalone image files. By integrating with iOS/macOS asset catalogs, the update aligns icon handling with platform best practices, simplifies asset management, and enables automatic support for multi-resolution and variant images (such as dark mode).

Addresses [discussion](https://github.com/software-mansion/react-native-screens/discussions/3395).
Closes https://github.com/software-mansion/react-native-screens-labs/issues/576. 

### Usage

You can add logo to asset catalog via Xcode:

Screenshot 2025-12-03 at 12 11 43

## Changes

- add `xcasset` icon type for iOS
- handle `xcasset` for Tabs
- handle `xcasset` for Bar Button Items in Stack
- update `BarButtonItems` example screen and add `Test3443` that tests xcassets in Bottom Tabs
  - `Test3443` should be converted in the future into single feature test when https://github.com/software-mansion/react-native-screens/pull/3528 is merged

| Bottom Tabs | Header Bar Button Item |
| --- | --- |
| Simulator Screenshot - iPhone 17 Pro - 2025-12-03 at 12 11 45 | Simulator Screenshot - iPhone 17 Pro - 2025-12-03 at 12 49 15 |

| Menu | Submenu |
| --- | --- |
| Simulator Screenshot - iPhone 17 Pro - 2025-12-03 at 12 52 03 | Simulator Screenshot - iPhone 17 Pro - 2025-12-03 at 13 06 08 |

Run `BarButtonItems` example screen. Use `XcassetIconButtonDemo` and `AdvancedMenuButtonDemo`.
Run `Test3443`.

## Checklist

- [ ] Ensured that CI passes


kkafar:
Hey! Thanks for the PR. We'll look into it after 4.19.0 release

ajanuar:
Thank you @kkafar ! I'll add more tests

ajanuar:
Hi @kkafar @kligarski , any chance to review this PR? Thank you

kkafar:
Hey @ajanuar, it's in our task queue for January. We'll look into it.

kligarski:
Hi, @ajanuar. Thanks for the PR and sorry for the delay.

I'll push some changes and modify the description a bit. Hopefully, we'll be ready to land this soon.

src/fabric/bottom-tabs/BottomTabsScreenNativeComponent.ts
```diff
@@ -144,10 +144,10 @@ export interface NativeProps extends ViewProps {
   iconType?: CT.WithDefault<IconType, 'sfSymbol'>;
 
   iconImageSource?: ImageSource;
-  iconSfSymbolName?: string;
+  iconResourceName?: string;
 
   selectedIconImageSource?: ImageSource;
-  selectedIconSfSymbolName?: string;
+  selectedIconResourceName?: string;
```
kligarski, line 150: I'm not sure about the name of this prop (`iconResourceName`). It's used for both sfsymbols and xcassets. We also have `drawableIconResourceName` and `imageIconResource` on Android which have very similar name. 

Maybe those props should be unified (`iconImageSource` (removing `imageIconResource`), `iconResourceName` (removing `drawableIconResourceName`) for both platforms and `selectedIconImageSource`, `selectedIconResourceName` as iOS-only).

src/components/helpers/prepareHeaderBarButtonItems.ts
kligarski, line 1: I don't like this solution (adding new prop to pass xcasset's name), we could use only one prop for both sfsymbolname and xcassetname however we're currently not passing `iconType` so we can't differentiate between the two.

At least we're not passing it intentionally - we're using `...icon` so it actually is present on the native side but I don't want to rely on that. We need to refactor this logic and add proper typing in JS layer because it's very difficult to add any changes now without being afraid of this whole feature breaking apart.

Decided to go with separate prop for now as refactoring this logic would be way out of scope of this PR.

ios/bottom-tabs/RNSTabBarAppearanceCoordinator.mm
```diff
@@ -70,8 +74,12 @@ - (void)setIconsForTabBarItem:(UITabBarItem *)tabBarItem
...
     }
 
-    if (screenView.selectedIconSfSymbolName != nil) {
-      tabBarItem.selectedImage = [UIImage systemImageNamed:screenView.selectedIconSfSymbolName];
+    if (screenView.selectedIconResourceName != nil) {
+      if (screenView.iconType == RNSBottomTabsIconTypeSfSymbol) {
+        tabBarItem.selectedImage = [UIImage systemImageNamed:screenView.selectedIconResourceName];
+      } else {
+        tabBarItem.selectedImage = [UIImage imageNamed:screenView.selectedIconResourceName];
+      }
```
kkafar, line 82: This is out of scope of this PR, however we need to have whole image loading code extracted to a separate set of utility classes, so that this logic is reusable between different components.

src/components/helpers/prepareHeaderBarButtonItems.ts
kkafar: +1 here,

this code needs a solid refactor. Let's open a ticket for that task on our board.

kligarski: https://github.com/software-mansion/react-native-screens-labs/issues/799

PR #3446, opened 2025-12-02
fix(iOS, Stack, Fabric): Pressable `hitSlop` in header subviews on Fabric

kligarski:
## Description

Fixes Pressable `hitSlop` in header subviews on Fabric, iOS.

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/df8823fb-efb5-4424-9cd3-8a6fff0ca088" /> | <video src="https://github.com/user-attachments/assets/e661f723-6129-4c05-8f09-f47e6e9a2f69" />

Closes https://github.com/software-mansion/react-native-screens-labs/issues/552.

## Changes

- convert hit test point from `RNSScreenStack`'s coordinates to that of `RNSScreenStackHeaderConfig` (on Fabric, it has negative origin.y so that ShadowTree knows about navigation bar + subviews positioning)
- fix logic in `RNSScreenStackHeaderConfig`'s hitTest:
  - previous implementation would always return after checking only one subview leading to inconsistent behavior
  - change order of iteration through`HeaderSubview`'s subviews and add early return, matching `RCTViewComponentView` implementation
- add test screen
- make `PressableWithFeedback` pass props to `Pressable` component

Run `Test3466`.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/apps/src/shared/PressableWithFeedback.tsx
+++ b/apps/src/shared/PressableWithFeedback.tsx
@@ -45,6 +45,7 @@ const PressableWithFeedback = React.forwardRef((props: PressableProps, ref: Forw
   return (
     <View ref={ref} style={[contentsStyle]}>
       <Pressable
+        {...props}
         onPressIn={onPressInCallback}
         onPress={onPressCallback}
         onPressOut={onPressOutCallback}
--- a/apps/src/tests/Test3446.tsx
+++ b/apps/src/tests/Test3446.tsx
@@ -1,0 +1,153 @@
+import React, { useState, useEffect } from 'react';
+import { NavigationContainer, ParamListBase } from '@react-navigation/native';
+import {
+  createNativeStackNavigator,
+  NativeStackNavigationProp,
+} from '@react-navigation/native-stack';
+import PressableWithFeedback from '../shared/PressableWithFeedback';
+import { ScrollView, View } from 'react-native';
+import Colors from '../shared/styling/Colors';
+import { SettingsPicker, SettingsSwitch } from '../shared';
+
+type RouteParamList = {
+  Home: undefined;
+};
+
+type NavigationProp<ParamList extends ParamListBase> = {
+  navigation: NativeStackNavigationProp<ParamList>;
+};
+
+type StackNavigationProp = NavigationProp<RouteParamList>;
+
+const Stack = createNativeStackNavigator();
+
+interface Configuration {
+  size: 'sm' | 'md' | 'lg';
+  hidesSharedBackground: boolean;
+  hitSlop: '0' | '10' | '30';
+  pressRetentionOffset: '0' | '20' | '50';
+}
+
+function getPressableFromConfig(config: Configuration) {
+  return (
+    <PressableWithFeedback
+      key={
+        config.size +
+        '_' +
+        config.hidesSharedBackground +
+        '_' +
+        config.hitSlop +
+        '_' +
+        config.pressRetentionOffset
+      }
+      hitSlop={Number(config.hitSlop)}
+      pressRetentionOffset={Number(config.pressRetentionOffset)}>
+      <View
+        style={{
+          width: config.size === 'sm' ? 10 : config.size === 'md' ? 36 : 80,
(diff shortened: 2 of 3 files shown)


ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -169,19 +169,17 @@ - (UIView *)hitTest:(CGPoint)point withEvent:(UIEvent *)event
...
+      // (we match RCTViewComponentView implementation).
+      for (UIView *headerComponentSubview in [subview.subviews reverseObjectEnumerator]) {
         CGPoint convertedPoint = [self convertPoint:point toView:headerComponentSubview];
         UIView *hitTestResult = [headerComponentSubview hitTest:convertedPoint withEvent:event];
 
         if (hitTestResult != nil) {
-          headerComponent = hitTestResult;
+          return hitTestResult;
         }
       }
```
kkafar, line 182: The comment above the for loop explicitly states, that the default behaviour of reverse iterating over the children) has been deliberately changed to left to right.

Can we do some research & see what problem exactly this approach solved & whether we do not introduce a regression with swapping this back?

kligarski: [In the PR that introduced the change](https://github.com/software-mansion/react-native-screens/pull/2466/files#diff-e5ef5b6e29f17bca80b51bc0c5faef1a44bac24e00952b30ac822520213dc6a5L148), the code has changed from

```objc
// we wrap the headerLeft/Right component in a UIBarButtonItem
// so we need to use the only subview of it to retrieve the correct view
UIView *headerComponent = subview.subviews.firstObject;
```

to

```objc
// we wrap the headerLeft/Right component in a UIBarButtonItem
// so we need to hit test subviews from left to right, because of the view flattening
UIView *headerComponent = nil;
for (UIView *headerComponentSubview in subview.subviews) {
  // ...
}
```

so I understood it as "due to view flattening in Fabric, we need to iterate through header subview's subviews instead of testing only the first/last one" with the order of iteration not being relevant.

Version prior to [the exact commit that introduced the change](https://github.com/software-mansion/react-native-screens/commit/61f2f9d36df85e6dcc11d2c688a1415a6ee68ee0) was this:

```objc
// we wrap the headerLeft/Right component in a UIBarButtonItem
// so we need to use the only subview of it to retrieve the correct view
UIView *headerComponent = subview.subviews.count > 1 ?
headerComponent = subview.subviews[subview.subviews.count - 1] :
subview.subviews.firstObject;
```

Also, previous version did not have an early return so both versions should be equivalent as long as `[headerComponentSubview hitTest:convertedPoint withEvent:event];` does not have any unexpected side effects (I don't see any in `RCTViewComponentView`'s implementation).

kligarski: I checked how view flattening works for iOS.

For JS hierarchy:

```
Pink wrapper (collapsable = false)
|
+ Yellow (collapsable = true)
  |
  + Orange (collapsable = true)
  |
  + Purple (collapsable = true)
```

<details>
<summary>JS code</summary>

```tsx
import { StyleSheet, View } from 'react-native';
import React from 'react';

  return (
    <View
      collapsable={false}
      style={[
        styles.container,
        {
          backgroundColor: 'pink',
        },
      ]}>
      <View
        collapsable={true}
        style={[
          styles.container,
          {
            backgroundColor: 'yellow',
          },
        ]}>
        <View
          collapsable={true}
          style={[
            styles.container,
            {
              backgroundColor: 'orange',
            },
          ]}
        />
        <View
          collapsable={true}
          style={[
            styles.container,
            {
              backgroundColor: 'purple',
            },
          ]}
        />
      </View>
    </View>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    padding: 50,
  },
});
```
</details>

We get native hierarchy:

```
Pink wrapper (collapsable = false)
| 
+ Yellow (collapsable = true)
|
+ Orange (collapsable = true)
|
+ Purple (collapsable = true)
```

https://github.com/user-attachments/assets/5902f871-2544-4b97-bdb4-fe6316d424d7

PR #3448, opened 2025-12-02
chore: remove CMake code with support for RN 0.75 or older

kkafar:
Remove build code covering RN 0.75 or older
We do not support such old versions for a long time now.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/624


kkafar:
Yeah, good catch. Added in https://github.com/software-mansion/react-native-screens/pull/3448/commits/0ddb5312fcca2d1044dea39b30bbb737dc09d82e

Issue #2990, opened 2025-07-07
[iOS] Header buttons are not centered in iOS 26 (w/ some investigation done on my end)


Armster15:
### Description

On iOS 26 with the new Liquid Glass, header buttons are not vertically centered.

<img width="1566" height="1686" alt="Image" src="https://github.com/user-attachments/assets/2b07308f-976d-4745-bc0a-e3641a0715ea" />

## 🐞 Minimal reproducible: 
1. Build a RN app with Xcode 26 on iOS 26
2. Have code like so:
    ```ts
    import { Stack } from "expo-router";
    import { Text, TouchableOpacity } from "react-native";
      
    export default function RootLayout() {
      return (
        <Stack
          screenOptions={{
            headerRight: () => (
              <TouchableOpacity style={{ backgroundColor: "blue" }}>
                <Text>Hello</Text>
              </TouchableOpacity>
            ),
          }}
        />
      );
    }
    ```
    I'm using Expo Router, but regardless the main thing is to have some component other than React Native's default Button export (which seems to work for some reason) as the component in `headerRight`.

3. See the following difference on iOS 18 and iOS 26

## 🧐 My Investigations
I inspected the view hierarchies of both iOS 18 and 26 and discovered the following:

iOS 18:
<img width="3790" height="2334" alt="Image" src="https://github.com/user-attachments/assets/f0c1e9be-565c-4645-923b-f7b36c096393" />

iOS 26:

<img width="3790" height="2334" alt="Image" src="https://github.com/user-attachments/assets/6c618114-2798-46f8-8697-b77505ddbae8" />

To me I thought a quick and dirty fix would be to just forcefully center the subviews inside the `RNSScreenStackHeaderSubview` view. So that's what I did: I modified `RNSScreenStackHeaderSubview.mm`, more specifically the `layoutSubviews` function like so:

Before:
```objc
- (void)layoutSubviews
{
  [super layoutSubviews];
  [self updateShadowStateInContextOfAncestorView:[self findNavigationBar]];
}
```

After:
```objc
- (void)layoutSubviews
{
    [super layoutSubviews];

    if (@available(iOS 26.0, *)) {
      if (self.subviews.count > 0) {
        UIView *child = self.subviews[0];
        CGRect frame = child.frame;
        CGFloat x = (self.bounds.size.width - frame.size.width) / 2.0;
        CGFloat y = (self.bounds.size.height - frame.size.height) / 2.0;
[...]

PR #3449, opened 2025-12-04
fix(iOS 26, Stack): center view inside bar button item

kligarski:
## Description

Fixes alignment of custom view inside bar button item on iOS 26.

> [!NOTE]
>
> Please note that this will also change positioning of bar button items with `hidesSharedBackground: true` on iOS 26.

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/d188be5c-48bf-4dee-97e2-599eac487af0" /> | <video src="https://github.com/user-attachments/assets/82d5ca33-8ef2-40bb-8d4e-030b0b53054c" /> |

Closes https://github.com/software-mansion/react-native-screens/issues/2990.

### Explanation

Starting from iOS 26, custom view inside `UIBarButtonItem` is stretched to at least 36 width. This causes our `RNSScreenStackHeaderSubview` to stretch if its width is smaller than 36, leaving subviews in the same origin (for one subview, it's usually `(0,0)`; that's why it is aligned to the left, visible in "before" recording).

In order to mitigate this, we add `wrapperView` which will be stretched by UIKit to required minimum size. `RNSScreenStackHeaderSubview` will be centered inside of it. Additional constraints and properties needed to be added to make sure that `wrapperView` is not stretched to all available space in `UINavigationBar` (this would happen in `headerRight` if we used only `wrapperView.width >= RNSScreenStackHeaderSubview.width` constraint):

- `wrapperView.width == RNSScreenStackHeaderSubview.width` - makes sure that wrapper matches content. By using lower priority than default, we allow UIKit to "take over" with its minimum width, while maintaining "fit-content"/"content hugging" behavior,
- `[self setContentHuggingPriority:UILayoutPriorityRequired forAxis:UILayoutConstraintAxisHorizontal]` - prevents stretching to all available width in `UINavigationBar`.

`RNSScreenStackHeaderSubview` needs to use `intrinsicContentSize` in order to bridge auto-layout centering and frame size from Yoga.

## Changes

- on iOS 26, add wrapper view between `UIBarButtonItem` and `RNSScreenStackHeaderSubview`
- center `RNSScreenStackHeaderSubview` inside `wrapperView` while maintaining correct sizing via auto layout
  - use `intrinsicContentSize` to bridge between auto-layout and frames calculated by Yoga
- fix frame passed to shadow tree (now, `RNSScreenStackHeaderSubview` can have non-zero offset (which was not the case previously); without change to `self.bounds`, this offset would be accounted for 2 times when using `[self convertRect:frame toView:ancestorView]`)

Run `Test3446`.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStackHeaderSubview.mm
+++ b/ios/RNSScreenStackHeaderSubview.mm
@@ -174,8 +174,11 @@ RNS_IGNORE_SUPER_CALL_BEGIN
         self);
   } else {
 #if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
+    BOOL sizeHasChanged = _layoutMetrics.frame.size != layoutMetrics.frame.size;
     _layoutMetrics = layoutMetrics;
-    [self invalidateIntrinsicContentSize];
+    if (sizeHasChanged) {
+      [self invalidateIntrinsicContentSize];
+    }
 #else // RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
     self.bounds = CGRect{CGPointZero, frame.size};
 #endif // RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)


ios/RNSScreenStackHeaderSubview.mm
```diff
@@ -199,23 +207,74 @@ - (void)reactSetFrame:(CGRect)frame
...
+    // 1. Set width of wrapperView to match RNSScreenStackHeaderSubview BUT when
+    //    RNSScreenStackHeaderSubview's width is smaller that minimal required 36 width, it breaks
+    //    UIKit's constraint. That's why we need to lower the priority of the constraint.
+    NSLayoutConstraint *widthEqual = [wrapperView.widthAnchor constraintEqualToAnchor:self.widthAnchor];
+    widthEqual.priority = UILayoutPriorityDefaultHigh;
+    widthEqual.active = YES;
+
+    NSLayoutConstraint *heightEqual = [wrapperView.heightAnchor constraintEqualToAnchor:self.heightAnchor];
+    heightEqual.priority = UILayoutPriorityDefaultHigh;
+    heightEqual.active = YES;
```
kligarski, line 249: Technically, we can remove height constraints and allow wrapperView to grow in height. This would make the entire liquid glass button interactive (currently, area inside liquid glass button but above and below custom view is not interactive as wrapperView is stretched by UIKit only in horizontal axis).

ios/RNSScreenStackHeaderSubview.mm
```diff
@@ -175,7 +178,12 @@ - (void)updateLayoutMetrics:(const react::LayoutMetrics &)layoutMetrics
         NSStringFromCGRect(frame),
         self);
   } else {
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
+    _layoutMetrics = layoutMetrics;
+    [self invalidateIntrinsicContentSize];
```
t0maboro: nit: we may consider calling it only if layoutMetrics have changed

kligarski: I think that it's already done for us:

obraz

I can add something like this:

```objc
    BOOL sizeHasChanged = _layoutMetrics.frame.size != layoutMetrics.frame.size;
    _layoutMetrics = layoutMetrics;
    if (sizeHasChanged) {
      [self invalidateIntrinsicContentSize];
    }
```

but I'm not sure if it's really necessary - let me know what you think.

t0maboro: I don't see any reason for invalidating until the size has changed, so I like the version you shared in the last snippet

kligarski: https://github.com/software-mansion/react-native-screens/pull/3449/commits/c506a3df4d3bc00da1d29724e654cff07ea7f3e2

ios/RNSScreenStackHeaderSubview.mm
kkafar: I see two ways to think about this now. 

First is that it might be weird experience clicking on the liquid glass & have it not respond if you click outside the custom view. However, if it were to respond & not fire an action I think it'll be even worse.

What we'd need to make this right is to be able to specify onPress for UIBarButtonItem when passing custom subview. 

But I think this is not in scope of this PR.

kligarski: > However, if it were to respond & not fire an action I think it'll be even worse.

The only thing I'm worried about is that this is exactly what happens (**liquid glass responds**, no action) when you click to the left/right of header subview if it's too small (because wrapperView must be stretched to 36 to satisfy UIKit) but when you click above/below the subview, we have **no liquid glass response** and no action.

kligarski:
> Please make sure that dynamic change of content size work (e.g. change content on button press) and we're good.

Seems to be working.

https://github.com/user-attachments/assets/b2fc0763-48cb-4719-8862-ad086225df00

PR #3450, opened 2025-12-04
fix(JS, stack): Fix push fail after 3rd screen in JS stack

kmichalikk:
Fixes https://github.com/software-mansion/react-native-screens-labs/issues/583

## Description

This PR changes the logic for creating update transactions in `ScreenContainer.onUpdate()`. In JS stack, the list of screens on stack is repopulated on each update. Some screens can be set to not detach when covered. By some mysterious logic, we've been adding them to the `pendingFront` list and adding them on top. We change this logic so every screen except for the first one are added in the same way.

## Changes

Modified & refactored logic for reattaching screens in `ScreenContainer.onUpdate()`.

Use Test3450.


kmichalikk:
Reported some more issues that are out of scope of this PR: https://github.com/software-mansion/react-native-screens-labs/issues/697

Issue #2522, opened 2024-11-19
StackNavigator as formSheet content has size of zero (iOS)


arnekolja:
### Description

After upgrading to react-navigation 7 and react-native-screens 4.2.0 my modal's content, a native stack navigator, isn't showing up any more. After some debugging I found out that it has a layout of zero width and zero height. Applying a wrapping `<View />` with fixed width and height works, applying `flex: 1` does not. However, a fixed height cannot be the solution for my flexible-height content. Using a normal view works fine and grows with its content, but not with the stack navigator.

This does only apply to `presentation: 'formSheet'` but not `presentation: 'modal'`, despite the fact that at least on iOS (only tested on iOS) the resulting modal looks the same. `presentation: 'modal'` handles sizing of my stack navigator correctly.

1. Is `presentation: 'formSheet'` and `presentation: 'modal'` really the same UI element here? And how can I fix the layout/height of my content? It is definitely a change from 3.x to 4.x, as it worked on 3.x for a long time.

### Steps to reproduce

Place a native stack navigator inside a modal of type 'formSheet'.

### Snack or a link to a repository

https://snack.expo.dev/Pqu3Cikm2Pk84jev-qtF9

### Screens version

4.2.0

### React Native version

0.76.2

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

iPhone 15 (iOS 17.2)

### Acknowledgements

Yes

colin-chadwick:
Having the same issue.

bojandurmic:
Same here `flex: 1` just won't work, I ended up using fixed height, but definitely not happy with that solution.

kkafar:
Are you using `formSheet` for its ability to specify detents or is `modal` presentation good replacement enough? 

The `formSheet` presentation has changed with 4.0 (see [here](https://github.com/software-mansion/react-native-screens/releases/tag/4.0.0)) and it is a bit impaired (but has other possibilities) at the moment. I'll try to handle this asap

colin-chadwick:
@kkafar Yes, the formSheet allows for better customization than the modal (for e.g. specifying detents, showing the grabber, corner radius, etc.).

kkafar:
Then please make sure to read the [release notes](https://github.com/software-mansion/react-native-screens/releases/tag/4.0.0)

[...]

PR #3454, opened 2025-12-05
fix(iOS, FormSheet): Allow handling dynamic content size in FormSheet since 0.82 with synchronous updates enabled

t0maboro:
## Description

Follow-up to https://github.com/software-mansion/react-native-screens/commit/460569215747281440102693f7759795aefd3cdf where I introduced a regression causing content flickering. The regression occurs when the content is allowed to dynamically adjust to the screen size (e.g., using `flex: 1` in styles). The issue is caused by the asynchronous nature of the ShadowNode state update for the screen, which leads to its size being updated with a delay relative to the detent change transition.

It's important to note that this issue does not occur when the update is sent **synchronously**. Therefore, I'd like to propose allowing `flex` in screen styles when a set of conditions are met:

- The formSheet does not have `fitToContents` defined in `sheetDetents`.
- React Native version is at least 0.82 (the one which introduced synchronous state updates).
- The feature flag `featureFlags.experiment.synchronousScreenUpdatesEnabled` in `react-native-screens` is enabled (for now, the default value remains disabled).

> [!WARNING]  
> This change is introducing another inconsistency in how components inside FormSheet are positioned in the flexbox model. We'll share some guidance soon, as for now, you can follow the progress here: https://github.com/software-mansion/react-native-screens/pull/3464

Fixes https://github.com/software-mansion/react-native-screens/issues/2522

## Changes

- Added conditional style with `flex: 1` for FormSheet

## Screenshots / GIFs

### Before

https://github.com/user-attachments/assets/fcfe7e32-5d3c-4536-83ba-01851a555120

### After

https://github.com/user-attachments/assets/79b864e2-f04f-4a37-9bde-693282b6f131

Test2522 - covering 4 cases: {flex, fixed height} x {detents from 0.0 to 1.0, fitToContents}

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/apps/src/tests/Test2522.tsx
+++ b/apps/src/tests/Test2522.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { createNativeStackNavigator } from '@react-navigation/native-stack';
 import type {
   NativeStackNavigationOptions,
@@ -8,6 +8,11 @@ import { NavigationContainer } from '@react-navigation/native';
 import { Button, ScrollView, StyleSheet, Text, View } from 'react-native';
 import { Spacer } from '../shared';
 import Colors from '../shared/styling/Colors';
+import { featureFlags } from 'react-native-screens';
+
+// For keeping the reference to the original value from the global scope,
+// because we need to override it for this specific example.
+let originalSynchronousScreenUpdatesFlagEnabled: boolean;
 
 type StackParamList = {
   Main: undefined;
@@ -94,6 +99,18 @@ function FormSheetWithFixedHeight() {
 }
 
 export default function App() {
+  useEffect(() => {
+    originalSynchronousScreenUpdatesFlagEnabled =
+      featureFlags.experiment.synchronousScreenUpdatesEnabled;
+    featureFlags.experiment.synchronousScreenUpdatesEnabled = true;
+
+    return () => {
+      // Note: It signals an error that the flag value has changed, but this is intentional
+      featureFlags.experiment.synchronousScreenUpdatesEnabled =
+        originalSynchronousScreenUpdatesFlagEnabled;
+    };
+  }, []);
+
   return (
     <NavigationContainer>
       <Stack.Navigator>


src/components/ScreenStackItem.tsx
```diff
@@ -21,6 +21,7 @@ import { RNSScreensRefContext } from '../contexts';
 import { FooterComponent } from './ScreenFooter';
 import { SafeAreaViewProps } from './safe-area/SafeAreaView.types';
 import SafeAreaView from './safe-area/SafeAreaView';
+import featureFlags from '../flags';
```
kkafar: Let's not use default import here. I just noticed that it indeed works, but it should change in the future, as there are two symbols exported from the `flags` module.


```suggestion
import { featureFlags } from '../flags';
```

t0maboro: https://github.com/software-mansion/react-native-screens-labs/issues/699

apps/src/tests/Test2522.tsx
```diff
@@ -0,0 +1,138 @@
...
+              onPress={() =>
+                navigation.navigate(screen as keyof StackParamList)
+              }
+            />
+          </View>
+        ))}
+      </ScrollView>
+    </View>
+  );
+}
```
kligarski:
```suggestion
interface Example {
  title: string;
  screen: keyof StackParamList;
}

const EXAMPLES: Example[] = [
  { title: 'flex:1', screen: 'FormSheetWithFlex' },
  { title: 'height:fixed', screen: 'FormSheetWithFixedHeight' },
  {
    title: 'flex:1 & fitToContents',
    screen: 'FormSheetWithFlexAndFitToContents',
  },
  {
    title: 'height:fixed & fitToContents',
    screen: 'FormSheetWithFixedHeightAndFitToContents',
  },
];

function Main({ navigation }: MainProps) {
  return (
    <View style={{ flex: 1, padding: 16 }}>
      <ScrollView>
        {EXAMPLES.map(({ title, screen }) => (
          <View key={screen} style={{ marginVertical: 4 }}>
            <Button title={title} onPress={() => navigation.navigate(screen)} />
          </View>
        ))}
      </ScrollView>
    </View>
  );
}
```

Maybe something like this to avoid `as`?

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3454/commits/9053b308270e1f757e5b5d08caaabffc74f33bbd

t0maboro:
> Looks good.
> 
> By the way, shouldn't synchronous state updates feature flags be enabled in our example app by default?

Imo we should have in the example app a default value that matches the default value of the flag we export. I'd prefer to have this option disabled in the app to reproduce issues with the same setup that users have. Having the same value for the flag will cover most cases. For specific examples, we'll see whether the flag is set in the code snippet.

kligarski:
> For specific examples, we'll see whether the flag is set in the code snippet.

But in `Test2522` it is not changed - the intention is that we set it manually after seeing the info in the example?

t0maboro:
> But in `Test2522` it is not changed - the intention is that we set it manually after seeing the info in the example?

Yes, that was the intention, but at 2nd thought I don't have any good reason against overriding the flag in this specific example, I'll add a commit for that soon

t0maboro:
cc @kligarski https://github.com/software-mansion/react-native-screens/pull/3454/commits/4620062854b89c125fb821f183d1c04fa3ad4b8a

kligarski:
I'm okay with both solutions - just wanted to clarify it.

src/components/ScreenStackItem.tsx
```diff
@@ -208,13 +209,22 @@ function getPositioningStyle(
...
   if (isIOS) {
-    return styles.absolute;
+    if (
+      allowedDetents !== 'fitToContents' &&
+      rnMinorVersion >= 82 &&
+      featureFlags.experiment.synchronousScreenUpdatesEnabled
+    ) {
+      return styles.container;
+    } else {
+      return styles.absolute;
```
kkafar: Can we also rename this style to `absoluteWithNoBottom` so that the fact that there is `bottom: undefined` specified is not hidden? Can be done in separate PR.

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3461

t0maboro:
> There is a weird effect with `flex: 1 & fitToContents` where the sheet when closed with gesture lags for a moment, before closing. Please see this video:
> 
>  Screen.Recording.2025-12-08.at.12.34.30.mov 
> I have no idea whether it is related to these changes or not. But we need to investigate it.
> 
> The close animation, when the dimming view is clicked, works perfectly fine. Only gesture dismissal has this problem.

I can reproduce the same problem on `main`. Interestingly, it depends on the content size
with 
```js
      <Text style={styles.text}>Start</Text>
      <Spacer space={100} />
      <Text style={styles.text}>End</Text>
```

the issue doesn't occur, video:

https://github.com/user-attachments/assets/fd86d717-34b7-4587-8753-dc151b50808f

with 
```js
      <Text style={styles.text}>Start</Text>
      <Spacer space={10} />
      <Text style={styles.text}>End</Text>
```

the issue occurs, video:

https://github.com/user-attachments/assets/731de87b-80c5-4d5e-87a6-ddb6f1d71d79

kkafar:
This one is currently on hold, as it'd "stabilize" behaviour on iOS that is vastly different from what we observe on Android. Currently we decided to delve into further research on how to mitigate platform discrepancies.

t0maboro:
> This one is currently on hold, as it'd "stabilize" behaviour on iOS that is vastly different from what we observe on Android. Currently we decided to delve into further research on how to mitigate platform discrepancies.

I created a follow-up PR with the guidance that should cover all possible cases https://github.com/software-mansion/react-native-screens/pull/3464 . I added some explanation, why we should consider proceeding with this PR and how we could align the approach between platforms when specific conditions would be met

t0maboro:
> There is a weird effect with `flex: 1 & fitToContents` where the sheet when closed with gesture lags for a moment, before closing. Please see this video:
> 
>  Screen.Recording.2025-12-08.at.12.34.30.mov 
> I have no idea whether it is related to these changes or not. But we need to investigate it.
> 
> The close animation, when the dimming view is clicked, works perfectly fine. Only gesture dismissal has this problem.


native bug, I put repro here: https://github.com/software-mansion/react-native-screens-labs/issues/705#issuecomment-3631346632

https://github.com/user-attachments/assets/20b4bb75-3a8d-4f74-a01c-a66a070efe6e

PR #3455, opened 2025-12-05
chore: Bump material components version

kmichalikk:
## Description

This PR bumps the version of material components (com.google.android.material:material) to version 1.13.

Tested FormShets, BottomTabs, Stack and Search component.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -220,7 +220,7 @@ dependencies {
     implementation 'androidx.fragment:fragment-ktx:1.6.1'
     implementation 'androidx.coordinatorlayout:coordinatorlayout:1.2.0'
     implementation 'androidx.swiperefreshlayout:swiperefreshlayout:1.1.0'
-    implementation 'com.google.android.material:material:1.12.0'
+    implementation 'com.google.android.material:material:1.13.0'
     implementation "androidx.core:core-ktx:1.8.0"
 
     constraints {


kmichalikk:
Points form release notes that might be relevant:
```
A11y
* Move responsibility of disabling hide on scroll to HideViewOnScrollBehavior and BottomAppBar

AppBar
* Fixed the bug that setBackground will reset the internal MaterialShapeDrawable for lift-on-scroll feature.

BottomNavigationView
* Add item gravity attribute and setter/getter
* Add active indicator attributes to change active indicator height, width, and horizontal margins when it is in an expanded state (i.e. when start icon gravity is set to 'start')
* Added new textviews for horizontal textview instead of moving it between layouts.
* Consume all touch events in BottomNavigationView so touch events are not passed to the view underneath 

CollapsingToolbarLayout
* Refactor multiline rotation fix to address related issue caused by keyboard

BottomSheet
* Updated to recycle the TypedArray after usage
* Deprecated enableEdgeToEdge for bottom sheet dialogs on API 35+.
* Prevent ACTION_DOWN events on the BottomSheetHandleDragView from setting touchingScrollChild to true. 
* Fix keyboard animation on Android 14. 
* Fixed crash when protections are set before setContentView is called
* Added half-expanded state description for accessibility
* Added support for androidx.core ProtectionLayout API and update dependency to 1.16

BottomAppBar
* Added guidance to prevent hide on scroll when Talkback is on

Dialog
* Fixed issue with child views overflowing from the alert dialog.

NavigationView
* Fixed section label alignment in RTL.
* Fixed subheader and divider insets and use relative listPreferredItemPaddings.

TextInputLayout
* Added hintMaxLines attribute.
* Start outline cutout bounds at a y location of 0 to encompass any possible vertical shifts due to a size difference between the placeholder text and the edit text.

Transitions / Motion
* Added spring motion subsystem
* Update androidx transition dependency to 1.5.0
* Update MaterialContainerTransform to use getLocationInWindow() instead of getLocationOnScreen() when calculating bounds

Tabs
* Fix selected text appearance not being applied correctly.

Search
* Set alpha for toolbar button directly rather than setting through a FadeThroughDrawable
* Fix centering text when set by the search view
* Fix animation when searchbar doesn't cover whole screen

SideSheet
* Added support for enabling edge to edge via setFitsSystemWindows()
```

PR #3457, opened 2025-12-05
chore: Update docs about dropping the support for RN 0.79 since 4.19.0

t0maboro:
## Description

According to the policy of supporting 3 past versions of React Native, I'm adding a note that starting from `react-native-screens@4.19.0`, we are dropping support for `react-native@0.79.x`. I reviewed the code and did not find any specific version checks related to 0.79 that we need to remove (I only checked for 0.79, not earlier versions).

Fixes https://github.com/software-mansion/react-native-screens-labs/issues/702

## Changes

- Updated `README.md` docs

N/A

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ Here's a table with summary of supported `react-native` versions:
 
 | library version | react-native version |
 | --------------- | -------------------- |
-| 4.19.0+         | 0.80.0+              |
+| 4.19.0+         | 0.81.0+              |
 | 4.14.0+         | 0.79.0+              |
 | 4.5.0+          | 0.77.0+              |
 | 4.0.0+          | 0.76.0+              |


t0maboro:
> According to https://reactnative.dev/docs/next/releases, it seems like 0.83 is going to be released today. Do we want to keep 0.80+ requirement or bump to 0.81+?

cc @kkafar if we're okay with removing 0.80 support either, with 0.83 support ready, I'm seeing the reasoning for dropping both 0.79 and 0.80. Additionally, these two were compatible with Expo SDK 53, so dropping them together would be nice. However, if we need to wait for the next Expo SDK for dropping 0.80 I'd propose to move this PR to 'on hold'

README.md
```diff
@@ -121,6 +121,7 @@ Here's a table with summary of supported `react-native` versions:
 
 | library version | react-native version |
 | --------------- | -------------------- |
+| 4.19.0+         | 0.80.0+              |
```
kkafar: We can drop supported version for the default architecture up to 0.81. 

The 0.83 will be released soon (most likely before we release 4.19.0).

t0maboro:
> The `Paper` version change is a bit controversial here, since I believe 0.81 is the last version, where Paper is supported.
> 
> I'd leave it at 0.79 (but we need to test before release, that it indeed works). We might consider adding a cap in form of `0.79..0.81`

If we still want to support 0.79, then we need to be really cautious with this PR for stable API migration: https://github.com/software-mansion/react-native-screens/pull/3458/files#diff-7ef3055ee848d40f846efdd42b67ecf3f9f5e79958d58c146e284078e9203ae6 , especially that the highlighted file hasn't an impact on Paper (I do not expect that problem for files under `fabric/`, but I need to ensure the backward compatibility with an additional round of testing)

kkafar:
I see. 

The policy for legacy architecture support is not certain yet. 

I'm willing to accept supporting only 0.80 & 0.81. I think that later, for some time we might settle on supporting only the 0.81 to finally sunset the support utterly. 

Okay then. Let's bump it for legacy architecture also to 0.80. ✅

README.md
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3457/commits/8100e1c64b9ffae6474f8d6e969abee3992a0a65

t0maboro:
> I see.
> 
> The policy for legacy architecture support is not certain yet.
> 
> I'm willing to accept supporting only 0.80 & 0.81. I think that later, for some time we might settle on supporting only the 0.81 to finally sunset the support utterly.
> 
> Okay then. Let's bump it for legacy architecture also to 0.80. ✅

Okay, I sent an update, atm Fabric is supporting 0.81+, Paper 0.80+

gabrieldonadel:
How are you guys supporting macOS now?

gabrieldonadel:
Just letting you know that https://github.com/software-mansion/react-native-screens/commit/e172686cd65d22cb67fb60ea11d2f7868e49bf79 breaks support with the latest react-native macOS unfortunately

t0maboro:
Hi @gabrieldonadel ,

Thanks for bringing up this topic - I'd like to share a few thoughts on how I see the situation:

- Our policy for supporting React Native versions is to maintain compatibility with the latest version and 2-3 versions back.
- Currently, as Meta has started stabilizing the API layer (https://reactnative.dev/blog/2025/06/12/moving-towards-a-stable-javascript-api), we're gradually migrating to using the new stable API wherever possible. This change was introduced in RN 0.80 and is aligned with our current policy. We're aware this breaks compatibility with 0.79 and older, as not all symbols were previously exported from react-native `index.js`.
- `react-native-macos` is currently not listed as a supported platform for this package (https://github.com/software-mansion/react-native-screens/tree/4.19.0?tab=readme-ov-file#supported-platforms ). If `react-native-screens` happens to work on macOS - that's great! But it's not something we actively test before releasing.
- At the same time, I noticed that other out-of-tree platforms we list as supported are also not yet on versions beyond 0.80/0.81 - for example, react-native-vision-os (https://www.npmjs.com/package/@callstack/react-native-visionos/v/latest ).

From my perspective, maintaining support for X versions back (where X is the lowest RN version used by any supported platform) creates overhead. I'm concerned about scenarios where upcoming RN versions may remove deep imports entirely, but we're still tied to a supported platform that is stuck on 0.79.

cc-ing @kkafar - we can revisit our strategy around support for older versions - we'll follow up with more feedback soon.

kkafar:
We never supported macOS, nor claimed that we do. Nobody on our side tests that platform. 

@gabrieldonadel If this is something that would be problematic for upcoming SDK 55 lemme know - we can figure something out for that one particular commit, however in long term we don't make any claims that we won't break "compatibility" again.

gabrieldonadel:
No worries guys. Mostly asking this because we run some compilation tests for macOS in the Expo Repo that includes react-native-screens. But for now I'll just lock the macOS tests to use react-native-screens version 4.18.
Once we can get react-native-macos 0.81 released things should be back to normal! Thanks

PR #3458, opened 2025-12-05
refactor: Remove codegen-related deep imports

t0maboro:
## Description

This PR initiates the migration towards the stable JS APIs introduced in React Native 0.80+, as described in the official post https://reactnative.dev/blog/2025/06/12/moving-towards-a-stable-javascript-api.

> [!NOTE]  
> As RN 0.79 does not export the required codegen bindings, this PR should be merged **only after** https://github.com/software-mansion/react-native-screens/pull/3457 is merged.

> [!NOTE]
> There is still one remaining deep import (`react-native/Libraries/ReactNative/AppContainer`) used in `DebugContainer`. This relies on private APIs with a complex structure (e.g., Inspector) that do not have a feasible workaround at this point. The further investigation and potential resolution should be handled in a follow-up PR to avoid mixing it with codegen-related changes made in this one. - https://github.com/software-mansion/react-native-screens-labs/issues/703
>
> EDIT: After checking, we're following the same approach as Modal implementation in Core, therefore I asked Meta team whether it'd be possible to expose `AppContainer` publicly: https://github.com/react-native-community/discussions-and-proposals/discussions/893#discussioncomment-15195267

```js
 WARN  Deep imports from the 'react-native' package are deprecated ('react-native/Libraries/ReactNative/AppContainer'). Source: /Users/tomaszboron/react-native-screens/src/components/DebugContainer.tsx 5:0
```

Fixes https://github.com/software-mansion/react-native-screens-labs/issues/378

## Changes

- Replaced deep imports for codegen with public RN APIs

Launched FabricExample on both android and iOS. Tested backward compatibility with RN 0.81 test app.

Listed all occurencies of `react-native/`

```bash
➜  pwd
/Users/tomaszboron/react-native-screens/src
➜  grep -r "react-native/" .
./components/DebugContainer.tsx:import AppContainer from 'react-native/Libraries/ReactNative/AppContainer';
./components/Screen.tsx:// react-native/Libraries/Components/View/ReactNativeViewViewConfig.js
```

Both are okay:
1. AppContainer, left because lack of the public API
2. Just a comment inside the component, not an import

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/fabric/SearchBarNativeComponent.ts
+++ b/src/fabric/SearchBarNativeComponent.ts
@@ -1,6 +1,5 @@
 'use client';
 
-/* eslint-disable */
 import { codegenNativeCommands, codegenNativeComponent } from 'react-native';
 import type {
   CodegenTypes as CT,
@@ -9,6 +8,7 @@ import type {
   HostComponent,
 } from 'react-native';
 
+// eslint-disable-next-line @typescript-eslint/ban-types
 export type SearchBarEvent = Readonly<{}>;
 
 export type SearchButtonPressedEvent = Readonly<{


src/fabric/SearchBarNativeComponent.ts
```diff
@@ -1,13 +1,13 @@
 'use client';
 
 /* eslint-disable */
```
kligarski: Maybe we can also get rid of this flie-scoped `eslint-disable` and do something like in `ScreenStackNativeComponent.ts`?

```ts
// eslint-disable-next-line @typescript-eslint/ban-types
export type SearchBarEvent = Readonly<{}>;
```

src/components/SearchBar.tsx
```diff
@@ -16,7 +16,7 @@ import SearchBarNativeComponent, {
   SearchButtonPressedEvent,
   ChangeTextEvent,
 } from '../fabric/SearchBarNativeComponent';
-import { DirectEventHandler } from 'react-native/Libraries/Types/CodegenTypes';
+import type { CodegenTypes as CT } from 'react-native';
```
kligarski, line 19: Is `CT` alias some kind of convention across libraries or did you just use it make it more concise? I wonder if using full `CodegenTypes` name would be clearer but I'm okay with both.

t0maboro: just making it more concise , full name looked too long for me for example in `CodegenTypes.DirectEventHandler<TransitionProgressEvent>` 
tbh, idk if this alias adds enough value in terms of clean code to use it, I don't have any strong preference whether we should use it or a full name

src/fabric/SearchBarNativeComponent.ts
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3458/commits/440d0f10a5f1523196778837514af1dcca41fd93

src/components/SearchBar.tsx
kkafar: Why not just import the type directly here?

src/fabric/bottom-tabs/BottomTabsAccessoryContentNativeComponent.ts
```diff
@@ -1,14 +1,12 @@
 'use client';
 
-// eslint-disable-next-line @react-native/no-deep-imports
-import codegenNativeComponent from 'react-native/Libraries/Utilities/codegenNativeComponent';
-import type { ViewProps } from 'react-native';
-import { WithDefault } from 'react-native/Libraries/Types/CodegenTypes';
+import { codegenNativeComponent } from 'react-native';
+import type { CodegenTypes as CT, ViewProps } from 'react-native';
```
kkafar, line 4: Same here. Can't we import `WithDefault` now? Is it private?

src/fabric/bottom-tabs/BottomTabsNativeComponent.ts
```diff
@@ -37,40 +31,40 @@ type TabBarControllerMode = 'automatic' | 'tabBar' | 'tabSidebar';
...
+  tabBarItemTitleFontSizeActive?: CT.Float;
   tabBarItemTitleFontWeight?: string;
   tabBarItemTitleFontStyle?: string;
   tabBarItemTitleFontColor?: ColorValue;
   tabBarItemTitleFontColorActive?: ColorValue;
   tabBarItemIconColor?: ColorValue;
   tabBarItemIconColorActive?: ColorValue;
   tabBarItemActiveIndicatorColor?: ColorValue;
-  tabBarItemActiveIndicatorEnabled?: WithDefault<boolean, true>;
+  tabBarItemActiveIndicatorEnabled?: CT.WithDefault<boolean, true>;
```
kkafar, line 51: Does this work? I mean, codegen does not run these files. It is rather a simple regex based parsing tool & I'm not sure it can interpret such syntax.

Even if it works, why not import the required types directly?

src/components/SearchBar.tsx
t0maboro: https://github.com/facebook/react-native/blob/v0.83.0-rc.5/packages/react-native/types/index.d.ts#L142 it's exported via namespace

src/fabric/bottom-tabs/BottomTabsAccessoryContentNativeComponent.ts
t0maboro: https://github.com/facebook/react-native/blob/v0.83.0-rc.5/packages/react-native/types/index.d.ts#L142 I have them stored in the namespace

src/fabric/bottom-tabs/BottomTabsNativeComponent.ts
t0maboro: > Does this work? I mean, codegen does not run these files. It is rather a simple regex based parsing tool & I'm not sure it can interpret such syntax.

it would fail on the compilation if it doesn't work, because it shouldn't recognize symbols if the codegen won't work properly

> Even if it works, why not import the required types directly?

https://github.com/facebook/react-native/blob/v0.83.0-rc.5/packages/react-native/types/index.d.ts#L142 it's exported in the namespace

src/fabric/bottom-tabs/BottomTabsAccessoryContentNativeComponent.ts
kkafar: Okay, that's ok then

kkafar (approved):

PR #3459, opened 2025-12-06
fix(iOS): Fix memory leaks in RNSScreen for brownfield environments

intmain:
## Description

This PR fixes critical memory leaks in `RNSScreen` that occur when React Native is completely dismissed in brownfield applications. In brownfield environments, when RN views and the Hermes engine are fully torn down, `RNSScreen` and related objects were not being deallocated, causing significant memory leaks.

**Root causes identified:**
1. **Retain cycle between RNSScreen ↔ RNSScreenView**: UIViewController holds a strong reference to its view, and RNSScreenView holds a strong reference back to controller
2. **CADisplayLink retain cycle**: CADisplayLink holds a strong reference to its target (self)
3. **Missing cleanup**: No `dealloc` method and incomplete timer cleanup in `setupProgressNotification`
4. **Lifecycle management**: Controller needs temporary strong reference until parent takes ownership

Fixes memory leaks in brownfield React Native environments.

## Changes

- Changed `RNSScreenView.controller` property from `strong` to `weak` reference
- Added `_retainedController` internal property to temporarily hold strong reference until parentViewController takes ownership
- Implemented `RNSWeakProxy` class using NSProxy pattern to break CADisplayLink retain cycle
- Added cleanup logic at the start of `setupProgressNotification` to handle multiple `viewWillDisappear` calls
- Added `dealloc` method to `RNSScreen` to properly clean up `_animationTimer`
- Added nil assignment after `invalidate` in completion block
- Release `_retainedController` in `willMoveToParentViewController` when parent takes ownership

### Testing for memory leaks in brownfield environment:

**Prerequisites**: Brownfield iOS app with React Native integration

**Steps**:
1. Integrate this version of react-native-screens in your brownfield app
2. Present a React Native screen
3. Completely dismiss the React Native screen (tear down entire RN instance)
4. Use Xcode Instruments with "Leaks" template or check Allocations

**Verification**:
You can add a temporary log to verify dealloc is called:

```objc
// In RNSScreen.mm
- (void)dealloc
{
  NSLog(@"✅ RNSScreen dealloc called - memory leak fixed!");
  [_animationTimer invalidate];
  _animationTimer = nil;
}
```

When you completely dismiss RN, you should see the log message in console.

**Expected behavior**: 
- Before fix: RNSScreen remains in memory, dealloc never called
- After fix: RNSScreen is properly deallocated, log appears in console

### Testing in FabricExample:

```bash
cd FabricExample
yarn install
cd ios && pod install && cd ..
yarn ios
# Navigate through screens and observe memory behavior
```

## Checklist

- [ ] Included code example that can be used to test this change ✅ (Testing instructions above)
- [ ] Updated TS types ❌ (Not applicable - iOS-only memory management fix)
- [ ] Updated documentation: ❌ (Not applicable - internal implementation fix, no API changes)
  - [ ] GUIDE_FOR_LIBRARY_AUTHORS.md
  - [ ] native-stack/README.md
  - [ ] src/types.tsx
  - [ ] src/native-stack/types.tsx
- [ ] Ensured that CI passes ⏳ (Will verify after PR submission)

---

**Note**: This is an iOS-only internal implementation fix with no API changes, so TypeScript types and documentation updates are not required.

### DIFF
--- a/ios/RNSScreen.h
+++ b/ios/RNSScreen.h
@@ -104,7 +104,7 @@ namespace react = facebook::react;
 @property (nonatomic) BOOL hideKeyboardOnSwipe;
 @property (nonatomic) BOOL customAnimationOnSwipe;
 @property (nonatomic) BOOL preventNativeDismiss;
-@property (nonatomic, retain) RNSScreen *controller;
+@property (nonatomic, weak) RNSScreen *controller;
 @property (nonatomic, copy) NSDictionary *gestureResponseDistance;
 @property (nonatomic) int activityState;
 @property (nonatomic, nullable) NSString *screenId;
@@ -212,6 +212,11 @@ namespace react = facebook::react;
  * tree.
  */
 - (BOOL)registerContentWrapper:(nonnull RNSScreenContentWrapper *)contentWrapper contentHeightErrata:(float)errata;
+
+/**
+ * Sets the retained controller to break the retain cycle.
+ */
+- (void)setRetainedController:(RNSScreen *_Nullable)controller;
 
 @end
 
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -53,6 +53,36 @@ struct ContentWrapperBox {
   float contentHeightErrata{0.f};
 };
 
+@interface RNSWeakProxy : NSProxy
+
+@property (nonatomic, weak, readonly) id target;
+
+- (instancetype)initWithTarget:(id)target;
+
+@end
+
+@implementation RNSWeakProxy
+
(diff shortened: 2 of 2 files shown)


maciejczechowski:
We've encountered this issue in our brownfield integration, yet for us it seems that it was RNSScreenStack not invalidating screens. Our current fix is adding ` [screenChildComponent invalidate];` after ` screenChildComponent.reactSuperview = nil;` in `unmountChildComponentView`

The screens seems to be properly discarded after (since invalidate effectively removes cycle) and I feel invalidate should be there - similar to avoiding cycle by using weak references as in this PR

PR #3463, opened 2025-12-08
feat: Add ContentScrollViewDetector component

kmichalikk:
## Description

This PR adds the `ContentScrollViewDetector` native component. This component allows us to locate and interface with [content ScrollView](https://developer.apple.com/documentation/uikit/uiviewcontroller/setcontentscrollview(_:for:)). 

The component holds an instance to the ScrollView and registers it for the closest view that implements ContentScrollViewConsumer. It is also responsible for unregistering the view when the ScrollView is destroyed.

We do not support the case when `ContentScrollViewDetector` does not wrap any ScrollView or the ScrollView is detached after some time. In the second case, the application will likely crash. Similarly, we do not support nested `ContentScrollViewDetector`s.

## Changes

Added `ContentScrollViewDetector` component. Added `ContentScrollViewConsumer` protocol that the `Detector` looks for.

Wrap a ScrollView with `<ContentScrollViewDetector>`. Add the following to `RNSScreen` and/or `RNSBottomTabsScreenComponentView`:

```
#pragma mark - ContentScrollViewConsumer

- (void)registerContentScrollView:(UIScrollView *)scrollView
{
  RCTLogInfo(@"Registering content ScrollView");
}

- (void)unregisterContentScrollView:(UIScrollView *)scrollView
{
  RCTLogInfo(@"Unregistering content ScrollView");
}
```
along with conformance declaration in the header. Observe the logs.

### DIFF
--- a/ios/RNSContentScrollViewDetector.h
+++ b/ios/RNSContentScrollViewDetector.h
@@ -12,7 +12,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface RNSContentScrollViewDetector : RNSReactBaseView
 
-- (nullable UIScrollView *)findContentScrollView;
+- (nullable UIScrollView *)findContentScrollViewWithinDetector;
 
 - (void)registerContentScrollViewInAncestors;
 
--- a/ios/RNSContentScrollViewDetector.mm
+++ b/ios/RNSContentScrollViewDetector.mm
@@ -17,13 +17,13 @@ namespace react = facebook::react;
       [self unregisterContentScrollViewInAncestors];
     }
   } else {
-    if ((_scrollView = [self findContentScrollView])) {
+    if ((_scrollView = [self findContentScrollViewWithinDetector])) {
       [self registerContentScrollViewInAncestors];
     }
   }
 }
 
-- (nullable UIScrollView *)findContentScrollView
+- (nullable UIScrollView *)findContentScrollViewWithinDetector
 {
   auto scrollView = [RNSScrollViewFinder findScrollViewInFirstDescendantChainFrom:self];
   if (scrollView) {


ios/RNSContentScrollViewDetector.h
```diff
@@ -0,0 +1,35 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+@interface RNSContentScrollViewDetector :
+#ifdef RCT_NEW_ARCH_ENABLED
+    RCTViewComponentView
+#else
+    UIView
+#endif
+
+- (UIScrollView *)findContentScrollView;
```
t0maboro: Should we mark the returned type as nullable?

ios/RNSContentScrollViewDetector.h
```diff
@@ -0,0 +1,35 @@
...
+
+#if !RCT_NEW_ARCH_ENABLED
+#import <React/RCTViewManager.h>
+#endif // !RCT_NEW_ARCH_ENABLED
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface RNSContentScrollViewDetector :
+#ifdef RCT_NEW_ARCH_ENABLED
+    RCTViewComponentView
```
t0maboro: 1. RNSReactBaseView is included and it contains include for RCTViewComponentView, so we have a transitive dependency here
2. btw shouldn't we use RNSReactBaseView instead of RCTViewComponentView?

ios/RNSContentScrollViewDetector.h
```diff
@@ -0,0 +1,35 @@
...
+#else
+    UIView
+#endif
+
+- (UIScrollView *)findContentScrollView;
+
+- (void)registerContentScrollViewInHierarchy;
```
t0maboro: The function traverses the hierarchy only through ancestors. I'm wondering whether the name of this method could better emphasize that we're interested in ancestors

ios/RNSContentScrollViewDetector.mm
```diff
@@ -0,0 +1,101 @@
...
+{
+  if (_linkedView != nullptr) {
+    RCTLogWarn(
+        @"Content ScrollView has already been registered. Make sure to only have one ScrollViewWrapper for content ScrollView.");
+    [self unregisterContentScrollViewInHierarchy];
+  }
+
+  UIView *parent = self.superview;
+  while (parent) {
+    if ([parent respondsToSelector:@selector(findContentScrollView)]) {
```
t0maboro, line 47: what if there will be another library that will use this (imo) common name for selector?

ios/RNSContentScrollViewDetector.mm
```diff
@@ -0,0 +1,101 @@
...
+
+@implementation RNSContentScrollViewDetector {
+  UIScrollView *_scrollView;
+  UIView *_linkedView;
+}
+
+- (void)didMoveToWindow
+{
+  if (self.window == nil) {
+    [self unregisterContentScrollViewInHierarchy];
```
t0maboro: do we need to call this if scrollView is nil?

ios/RNSContentScrollViewDetector.mm
```diff
@@ -0,0 +1,101 @@
+#import "RNSContentScrollViewDetector.h"
+#import <React/RCTLog.h>
+#import <react/renderer/components/rnscreens/ComponentDescriptors.h>
+
+namespace react = facebook::react;
+
+@implementation RNSContentScrollViewDetector {
+  UIScrollView *_scrollView;
+  UIView *_linkedView;
```
t0maboro: maybe contentScrollViewConsumer? linkedView is general

ios/RNSContentScrollViewDetector.mm
```diff
@@ -0,0 +1,101 @@
...
+
+  UIView *parent = self.superview;
+  while (parent) {
+    if ([parent respondsToSelector:@selector(findContentScrollView)]) {
+      RCTLogWarn(
+          @"Nested ScrollViewWrapper detected. Make sure to only have one ScrollViewWrapper for content ScrollView.");
+    }
+
+    if ([parent respondsToSelector:@selector(registerContentScrollView:)]) {
+      [(id<ContentScrollViewConsumer>)parent registerContentScrollView:_scrollView];
```
t0maboro, line 53: same consideration as above, for selector naming conventions, because this cast doesn't look very safe

ios/RNSContentScrollViewDetector.mm
```diff
@@ -0,0 +1,101 @@
...
+
+    parent = parent.superview;
+  }
+}
+
+- (void)unregisterContentScrollViewInHierarchy
+{
+  if (_linkedView == nullptr) {
+    return;
+  }
```
t0maboro: nit: it may take no action, so I'd say that it should be rather named `maybeUnregisterContentScrollViewInHierarchy`
tbh I'd consider other approaches here either:
1. do not calling this method from other places at all if `_linkedView` is null
2. having `maybeUnregisterContentScrollViewInHierarchy` which will call `unregisterContentScrollViewInHierarchy` which has the rest of logic

this one is very nitpicky, I'm okay with the current version, but I see potential improvement in terms of semantics, so I'm sharing some ideas

ios/RNSContentScrollViewDetector.h
kmichalikk: [d152c66](https://github.com/software-mansion/react-native-screens/pull/3463/commits/d152c66516d2a66359b6bde800e99bc8088cbf0c)

ios/RNSContentScrollViewDetector.h
kmichalikk: [d152c66](https://github.com/software-mansion/react-native-screens/pull/3463/commits/d152c66516d2a66359b6bde800e99bc8088cbf0c)

ios/RNSContentScrollViewDetector.h
kmichalikk: [d152c66](https://github.com/software-mansion/react-native-screens/pull/3463/commits/d152c66516d2a66359b6bde800e99bc8088cbf0c)

ios/RNSContentScrollViewDetector.mm
kmichalikk: [d152c66](https://github.com/software-mansion/react-native-screens/pull/3463/commits/d152c66516d2a66359b6bde800e99bc8088cbf0c)

ios/RNSContentScrollViewDetector.mm
kmichalikk: Changed to scrollViewConsumer to make it shorter. We're only working with one ScrollView in the class & the class is named accordingly.

[d152c66](https://github.com/software-mansion/react-native-screens/pull/3463/commits/d152c66516d2a66359b6bde800e99bc8088cbf0c)

ios/RNSContentScrollViewDetector.mm
kmichalikk: [8be2ef9](https://github.com/software-mansion/react-native-screens/pull/3463/commits/8be2ef925ad6a4c4264e6f346486f2fe5db4079a)

ios/RNSContentScrollViewDetector.mm
kmichalikk: This we agreed internally to leave as is.

benkingcode:
Hi any movement on this? I have a use case of a scrollview not being a direct child of a screen and it's breaking the native tabs, I think this would fix it

kkafar:
I'll close here, because we've landed https://github.com/software-mansion/react-native-screens/pull/3674, which adds a component that is going to serve very similar function.

Issue #3466, opened 2025-12-09
[iOS] App Freeze regression when using PagerView nested with native-stack


gabemeola:
### Description

When nesting a `Stack.Navigator` (`@react-navigation/native-stack`) all application touches are disable (full app freeze). I confirmed this wasn't a react-navigation or a react-native-pager-view regression.

iOS logs this warning when the issue occurs. Android works fine.

> Unexpected nil window in latent system gesture client update: windowServerHitTestWindow: <UIWindow: 0x10760f620; frame = (0 0; 402 874); userInteractionEnabled = NO; gestureRecognizers = <NSArray: 0x600000d3d1d0>; layer = <UIWindowLayer: 0x600001723e80>>, touch: <UITouch: 0x16d664690> type: Direct; phase: Stationary; is pointer: NO; tap count: 1; force: 0.000; window: (null); responder: (null);

Confirmed breaking starting at version `4.17.0` to `4.18.0`

Downgrading to `4.16.0` fixes the issue.

Related commits:
- https://github.com/software-mansion/react-native-screens/pull/3173
- https://github.com/software-mansion/react-native-screens/pull/3265

### Steps to reproduce

```tsx
import PagerView from 'react-native-pager-view';
import { createNativeStackNavigator } from '@react-navigation/native-stack';

const Stack = createNativeStackNavigator();

function App() {
  <PagerView>
    <View key="0">
      <Text>This Works</Text>
    </View>
    <View key="1">
       <NavigationContainer>
          <Stack.Navigator>
            <Stack.Screen  component={() => <Text>Test</Text>} />
          </Stack.Navigator>
       </NavigationContainer>
    </View>
  </PagerView>
}
```

### Snack or a link to a repository

todo

### Screens version

4.18.0

### React Native version

0.81.5

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

Fabric (New Architecture)

### Build type

None

### Device

Real device

### Device model

iOS 26, including MacOS Catalyst

### Acknowledgements

Yes

kmichalikk:
Hi @gabemeola 
Please check the patch here and let us know if it fixed it for you:

https://github.com/software-mansion/react-native-screens/pull/3469

thanhcuong1990:
I’m experiencing the same issue with react-native-screens `4.19.0`.
In my case, I’m using `@react-navigation/material-top-tabs`.

JMeyer-FSHealth:
[...]

PR #3469, opened 2025-12-10
fix: Handle enabling interactions for RNSScreenStack that is moved to window

kmichalikk:
## Description

This fix handles the case where the whole nested ScreenStack is moved to window with UIPageViewController. The logic worked under the assumption that the nested stack is put inside outer stack's screen, but this is not the case here.

We had a similar case when the hierarchy changed on JS reload. The same fix can be used here by just removing a condition.

## Changes

Modified RNSScreenStack.didMoveToWindow

Install `react-native-pager-view` and paste the following code into `App.tsx`:

```ts
import PagerView from 'react-native-pager-view';
import { Button, ScrollView, Text, View } from 'react-native';
import { NavigationContainer, NavigationIndependentTree, NavigationProp } from '@react-navigation/native';

const OuterStack = createNativeStackNavigator();
const InnerStack = createNativeStackNavigator();

function Screen(props: { navigation: NavigationProp<any> }) {
  const [click, setClick] = useState(false);

  return (<>
    <Button title='click' onPress={() => setClick(c => !c)}/>
    <Button title='push' onPress={() => props.navigation.navigate('inner2')}/>
    { click && <Text>clicked</Text>}
  </>)
}

function Inner() {
    return (<NavigationIndependentTree>
      <PagerView style={{ flex: 1 }} initialPage={0} scrollEnabled={true}>
        <View key="0" style={{ backgroundColor: 'blue' }}>
          <Text>This Works</Text>
        </View>
        <View key="1" style={{ backgroundColor: 'red' }}>
          <NavigationContainer>
              <InnerStack.Navigator>
                <InnerStack.Screen name='inner' component={Screen} options={{ headerBackButtonDisplayMode: 'minimal' }} />
                <InnerStack.Screen name='inner2' component={Screen} options={{ headerBackButtonDisplayMode: 'minimal' }} />
              </InnerStack.Navigator>
          </NavigationContainer>
        </View>
        <View key="2" style={{ backgroundColor: 'red' }}>
          <NavigationContainer>
              <InnerStack.Navigator>
                <InnerStack.Screen name='inner' component={Screen} />
                <InnerStack.Screen name='inner2' component={Screen} />
              </InnerStack.Navigator>
          </NavigationContainer>
        </View>
      </PagerView>
    </NavigationIndependentTree>)
}

  return (
    <NavigationContainer>
      <OuterStack.Navigator>
        <OuterStack.Screen name='outer' component={Inner} />
      </OuterStack.Navigator>
    </NavigationContainer>
  )
}
```

### DIFF
--- a/ios/RNSScreenStack.mm
+++ b/ios/RNSScreenStack.mm
@@ -367,10 +367,16 @@ RNS_IGNORE_SUPER_CALL_END
     [self maybeAddToParentAndUpdateContainer];
   }
 #endif
-  if (self.window == nil) {
-    // When hot reload happens that would remove the whole stack, disabling the interaction on a screen out transition
-    // will not be matched with enabling the interactions on another screen's in transition. We need to make sure
-    // that the subtree is interactive again
+  if (@available(iOS 26, *)) {
+    // Re-enable interactions that have been disabled by screen transition
+    //
+    // There are some cases where stack screens cannot reenable the interactivity on their own.
+    // 1) When hot reload happens that would remove the whole stack, disabling the interaction on a screen out
+    // transition will not be matched with enabling the interactions on another screen's in transition. 2) In
+    // UIPageViewController, you can make a whole ScreenStack appear by switching pages. In this case we found out that
+    // the screen.viewDidAppear is called and enables interactions before willMoveToWindow that disables them.
+    //
+    // For these cases, we need ScreenStack to help making sure that the subtree is interactive again.
     [RNSScreenView.viewInteractionManagerInstance enableInteractionsForLastSubtree];
   }
 }


kkafar:
What is the state of this PR here? Is it ready for review?

huats:
I do hope this patch will be merged. Tested on RN 0.83.1 + New Architecture + iOS 26. Confirmed: react-native-screens@4.23.0 with PR #3469 applied manually fixes the freeze when using Stack.Navigator nested inside Material Top Tabs
  (UIPageViewController). Before that my only solution was to pin to 4.16

kkafar:
@kmichalikk can you update me what's the status of this PR?

kmichalikk:
This is a potential fix for https://github.com/software-mansion/react-native-screens/issues/3466.
Originally, we hesitated to merge this because it touches core functionalities, so I tagged is as a draft and waited for issue author to respond. There's still no response there (some unrelated issue got solved in the thread), so the PR just sits here.

Since this solves @huats problem we can revisit this now.

huats:
I'll be happy to test some elements if that can be helpful !

kmichalikk:
@huats could you check if https://github.com/software-mansion/react-native-screens/pull/3631 fixes the problem instead this patch?

This draft here is an old PR that fixes old logic. We're withdrawing from blocking interactions during transition, because we now have a proper fix for some state-related bugs, and we'll probably make it a default some time in the feature, and, if there are no bugs, remove the interaction disabling code entirely.

huats:
Sorry for the time to get back to you. Yes I can confirm that https://github.com/software-mansion/react-native-screens/pull/3631 also fixes my issue ! So if would be great to have either of one to be merged so that I can avoid to maintain a separe patch in my tree. Thanks for your help !

kmichalikk:
Since #3631 got merged, and it solves the issue, this is no longer necessary

Issue #3473, opened 2025-12-11
[RNScreens] Unsupported orientation mask on iOS


rahimn20:
### Description

After upgrading a project from React Native 0.73.11 → 0.80.0, screen components that previously used the orientation option inside React Navigation now throw the following runtime error:

<img width="1170" height="2532" alt="Image" src="https://github.com/user-attachments/assets/038c0947-3d41-4bd4-8f57-d245e194f7ff" />

### Steps to reproduce

1. Create a React Native project on 0.80.0 using React Navigation with @react-navigation/native-stack.
2. Add a screen configured with the orientation option:

`<Stack.Screen
  name="AnyScreen"
  component={AnyScreen}
  options={{
    gestureEnabled: false,
    orientation: 'landscape_right',
  }}
/>`

3. Launch the app on iOS 
4. Navigate to the screen configured above.
5. Observe that the app logs or throws the following error:

`[RNScreens] Unsupported orientation mask`

6. Notice that the orientation is not applied to the screen.
7. If the issue does not reproduce initially, try navigating back from this screen using:

`navigation.reset({
  index: 0,
  routes: [{ name: 'SomeOtherScreen' }],
});`

instead of navigation.goBack().
The error consistently appears when the screen is mounted again via navigation.reset()

### Snack or a link to a repository

https://github.com/rahimn20/RNScreensOrientationBug

### Screens version

4.18.0

### React Native version

0.80.0

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

Real device

### Device model

iPhone 14 pro

### Acknowledgements

Yes

PR #3474, opened 2025-12-11
fix(iOS, Stack): Fix enum conversion for landscape right

t0maboro:
## Description

There was a typo, mask should be used instead of UIDeviceOrientation value.

Fixes: https://github.com/software-mansion/react-native-screens/issues/3473

`Orientation.tsx`

## Checklist

- [x] Ensured that CI passes


kkafar:
Let's land it quickly. I want to include this in 4.19.0

PR #3478, opened 2025-12-14
fix(iOS, Tabs): use clearColor background to prevent flash during tab transitions

### LINKED ISSUE #3472: iOS 26: Custom Native-Stack Header Gets Rounded/Clipped During Navigation Transition
## Issue:
On iOS 26, custom header components rendered through @react-navigation/native-stack are being visually cut off during the new navigation animation introduced in this iOS version. This only affects screens using a custom React component as the header.

## Details:
When using the native-stack navigator with a custom header, the top portion of the header becomes clipped during the transition animation on iOS 26. This behavior does not occur on earlier iOS versions.

```ts
const DashboardStack = createNativeStackNavigator<DashboardStackParamList>();

return (
   <StoreDashboardStack.Navigator
      screenOptions={{
        header: () => <View style={{ backgroundColor: 'red', height: 100, width: '100%' }} />,
      }}>
)
```

https://github.com/user-attachments/assets/b01506af-0798-4465-bec1-7f1f7932b551

## Expected behavior

The custom header should behave the same as the default native-stack header on iOS 26—that is, no rounded corners should appear during navigation transitions. The header should remain fully rectangular and should not be visually clipped or masked at the top.

The following example shows the default native header, which demonstrates the expected behavior:

```ts
const DashboardStack = createNativeStackNavigator<DashboardStackParamList>();

return (
   <StoreDashboardStack.Navigator
      screenOptions={{
        headerStyle: {
          backgroundColor: 'red',
        },
        title: '',
        headerBackVisible: false,
      }}>
)
```

https://github.com/user-attachments/assets/2938470e-3b01-478d-bd8c-2f88d2700940

Original issue react-navigation: https://github.com/react-navigation/react-navigation/issues/12894 

### Steps to reproduce

1. Have a screen with a custom header header component
2. Navigate to another screen

https://github.com/Jonas-Bostoen/react-native-navigation-native-stack-headers

### Screens version

4.18.0

### React Native version

0.81.5

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo bare workflow

### Architecture

Fabric (New Architecture)

### Build type

None

### Device

None

### Device model

All iPhones running iOS26 in debug and release mode

### Acknowledgements

Yes

jpancotti:
Previously using systemBackgroundColor, which could cause a visible flash of white or black background during tab transitions in apps with custom themes. Using clearColor ensures the background is transparent, preventing any visual artifacts regardless of the app's color scheme.

Fixes #3472 

## Description

When using NativeTabs with custom app themes (especially dark themes), there can be a brief flash of the system background color during tab transitions. Using `clearColor` makes the background fully transparent, preventing any flash regardless of the app's color scheme.

## Changes

- Changed `self.backgroundColor` from `[UIColor systemBackgroundColor]` to `[UIColor clearColor]` in `RNSBottomTabsScreenComponentView.mm`

```tsx
import { Tabs } from 'expo-router/tabs';

export default function TabLayout() {
  return (
    <Tabs
      screenOptions={{
        tabBarStyle: { backgroundColor: '#1a1a2e' },
      }}
    >
      <Tabs.Screen name="index" options={{ title: 'Home' }} />
      <Tabs.Screen name="settings" options={{ title: 'Settings' }} />
    </Tabs>
  );
}
```

1. Create app with dark theme
2. Switch between screens rapidly
3. **Before:** Brief flash of system background color visible around edges
4. **After:** No flash, smooth transitions

## Checklist

- [ ] Updated TS types
- [ ] Updated documentation:
- [x] Ensured that CI passes

### DIFF
--- a/ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
+++ b/ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
@@ -70,10 +70,10 @@ namespace react = facebook::react;
   _scrollEdgeEffectsNeedUpdate = NO;
 #endif
 
-  // Prevents incorrect tab bar appearance after tab change on iOS 26.0
-  // TODO: verify if it's still necessary on iOS 26.1
+  // Use clearColor to prevent any background flash during tab transitions.
+  // This ensures compatibility with both light and dark themes.
 #if !TARGET_OS_TV
-  self.backgroundColor = [UIColor systemBackgroundColor];
+  self.backgroundColor = [UIColor clearColor];
 #endif // !TARGET_OS_TV
 
   [self resetProps];


kligarski:
Hi, @jpancotti. Thank you for the PR. 

Unfortunately, this is not as simple as it seems 😔. When you switch between tabs with similar color after tabs are rendered for the first time, there is another flash even with `clearColor` approach (for details see: https://github.com/software-mansion/react-native-screens/pull/3492). Also, we added background color to tab screen to mitigate problems with liquid glass tab bar appearance adapting to the background (more details here: https://github.com/software-mansion/react-native-screens/pull/3279). I'm not sure how it behaves with `clearColor` - this will require more testing.

We decided to expose in `react-native-screens` full configuration of background color (for host container and every tab screen) in this PR: https://github.com/software-mansion/react-native-screens/pull/3492. This will allow you to modify background color as you like (including approach similar to this PR with tab screen being transparent).

As https://github.com/software-mansion/react-native-screens/pull/3492 allows this configuration and more, I'll close this PR.

PR #3479, opened 2025-12-15
fix(iOS, FormSheet): Add fallback for searching ScrollView which is inside SAV

t0maboro:
## Description

Fixes a regression in ScrollView with non-scrolling content when using absolute styles. Adding a SAV into the hierarchy - as a workaround to content's padding issues under the header on iOS 26 - broke the logic for finding the ScrollView inside the `ScreenContentWrapper`, preventing the frame correction from being applied properly. This PR adds fallbacks to search for the `ScrollView` within the subviews of the `SafeAreaView`.

> [!NOTE]  
> Allowing the `flex:1` style for the `ContentWrapper` also resolves an issue with ScrollView's `contentInsetsAdjustmentBehavior`, which now works correctly with synchronous state updates enabled. It would be good to align this behavior for the `absoluteWithNoBottom` style as well - https://github.com/software-mansion/react-native-screens-labs/issues/718

## Changes

- Added fallback logic to search within `SafeAreaView` subviews for `ScrollView` in `ContentWrapper`.

## Screenshots / GIFs

### Before

<table>
  <tr>
    <td width="50%">
      <video src="https://github.com/user-attachments/assets/013ea98e-c4ec-4c96-8963-4f2c6829da61"></video>
    </td>
    <td width="50%">
      <video src="https://github.com/user-attachments/assets/acef493d-a37e-482f-be67-a5c2a8ea746c"></video>
    </td>
  </tr>
</table>

### After

<table>
  <tr>
    <td width="50%">
      <video src="https://github.com/user-attachments/assets/7d96a236-4ba6-4b6a-8c1f-8f64871b1faa"></video>
    </td>
    <td width="50%">
      <video src="https://github.com/user-attachments/assets/3bbd69e1-8760-4246-9a21-50eb48e3e54a"></video>
    </td>
  </tr>
</table>

Tested on TestFormSheet

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenContentWrapper.h
+++ b/ios/RNSScreenContentWrapper.h
@@ -25,10 +25,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
-struct RNSScrollViewSearchResult {
+typedef struct {
   RNS_REACT_SCROLL_VIEW_COMPONENT *scrollViewComponent;
   UIView *contentContainerView;
-};
+} RNSScrollViewSearchResult;
 
 @interface RNSScreenContentWrapper :
 #ifdef RCT_NEW_ARCH_ENABLED


t0maboro:
Still observing the regression with FlatList, converting to draft

t0maboro:
Discussed internally FlatList issues, seems like a renderer problem, not RNScreens, PR is ready for review

kligarski (approved):

ios/RNSScreenContentWrapper.mm
```diff
@@ -113,12 +114,33 @@ - (nullable RNS_REACT_SCROLL_VIEW_COMPONENT *)childRCTScrollViewComponent
...
+          if ([subview isKindOfClass:RNS_REACT_SCROLL_VIEW_COMPONENT.class]) {
+            scrollViewComponent = static_cast<RNS_REACT_SCROLL_VIEW_COMPONENT *>(subview);
+            containerView = maybeSafeAreaView;
+            break;
+          }
+        }
+      }
+    }
+  }
+#endif // RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
```
kkafar: Can we move this logic inside childRCTScrollViewComponent function.
This method should not be about finding a scrollview but rather acting upon one already found. If we need additional information about which view is the container view, we can return it from the method. You can use standard `std::pair`, that will work.

ios/RNSScreenContentWrapper.mm
kkafar: For this warning to fullfill it's purpose I think it should also take into consideration where the scrollview is found.

ios/RNSScreenContentWrapper.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3479/commits/3b83f9b379a2242c0494e8685c8d1189fcca0195 + I believe that struct is adding better semantic https://github.com/software-mansion/react-native-screens/pull/3479/commits/97987ad75f708ad257e5b8345c56b304218ca9c6

ios/RNSScreenContentWrapper.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3479/commits/2199ea314d94fa006d3ac23f7576676e617fe5b4

Issue #2560, opened 2024-12-09
FormSheet with sheetAllowedDetents: 'fitToContents' doesn't update height dynamically on Android


SickanK:
### Description

The new native stack presentation formSheet with sheetAllowedDetents: 'fitToContents' on Android initially works correctly by adjusting to the content height. However, when the content height changes dynamically, the sheet does not update its height to match. 

I had a short interaction over on [X](https://x.com/kafara_kacper/status/1863560771252195700) and was directed to post an issue here. Love the new formSheets, looking forward to being able to use them with dynamic height adjustment!

### Steps to reproduce

1. Create a new screen with a formSheet presentation mode and sheetAllowedDetents set to fitToContents
2. Add content with a fixed height to the formSheet
3. Open the formSheet from another screen
4. Verify that the initial sheet height matches the fixed content height correctly
5. Change the content's fixed height programmatically
6. Observe that the sheet height remains at its initial size and does not adjust to match the new content height

### Snack or a link to a repository

https://snack.expo.dev/SnPwLOzeISrX17ay1RuQY

### Screens version

4.1.0

### React Native version

0.76.3

### Platforms

Android

### JavaScript runtime

JSC

### Workflow

Expo bare workflow

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

Real device

### Device model

Nothing Phone (2) (Android 14)

### Acknowledgements

Yes

ljukas:
I heavily use https://github.com/lodev09/react-native-true-sheet for my sheets and they have the same issue regarding auto-sized sheets on Android. How to work around it there is that you also give the largest possible detent as an available size. So you would give ['fitToContents', '1.0'] or similar. This might work here to.

kkafar:
My first impression is that it should be straightforward to do. Gimme few minutes

kkafar:
I can make it jump to the right height, but I can't animate it properly (I don't want to animate it manually, would prefer to delegate it to the bottom sheet behaviour). Will look into it some more

kkafar:
Here's PoC of the changes: https://github.com/software-mansion/react-native-screens/pull/2570

SickanK:
Thanks for working on this! 

Wouldn't the expected behavior mirror what happens when you initially open a sheet without content? What cases would the content disappear where you wouldn't want to instead e.g. close the sheet?

RohovDmytro:
[...]

PR #3484, opened 2025-12-16
fix(Android, FormSheet): Fix dynamic height change for fitToContents

t0maboro:
## Description

This PR introduces improved support for `fitToContents` in BottomSheet by modifying its style to `absoluteWithNoBottom`. This style provides a reference to the current top position of the sheet, making it easier to handle dynamic content within FormSheet.

Having improved control over `maxHeight` (added in previous PRs), we are now able to implement a smooth resizing animation for the BottomSheet.

> [!IMPORTANT] 
> Key assumptions (these should be carefully reviewed)
>
> - Unmounting of components remains the responsibility of the application code. We do not include default enter/exit animations at the component level.
> - Visual styles such as `backgroundColor` should be applied via `contentStyle` on the Screen level. In fitToContents mode, the height of the ScreenContentWrapper should always match the height of the content.
> - When a component unmounts, the ScreenContentWrapper's size shrinks before we can start the transition. Since its size is necessary to calculate the animation delta, styles should be applied at the Screen level, where we have better control over when layout occurs.

> [!NOTE]
>  Note: `bottom: 0` is intentionally excluded from these styles for two reasons:
>
> 1. Omitting the bottom constraint ensures the Yoga layout engine does not dynamically recalculate the Screen and content size during animations.
>
> 2. Including `bottom: 0` with 'position: absolute' would force the component to anchor itself to an ancestor's bottom edge. This creates a dependency on the ancestor's size, whereas 'fitToContents' requires the FormSheet's dimensions to be derived strictly from its children.

<table>
  <tr>
    <td width="50%">
      <video src="https://github.com/user-attachments/assets/bfec13fd-08db-42fe-b0ea-67605890dd1f"></video>
    </td>
    <td width="50%">
      <video src="https://github.com/user-attachments/assets/3faa10fa-8bcc-402a-be0c-bd43dd1b18f2"></video>
    </td>
  </tr>
</table>

This PR also introduces a prop that allows overriding the default enter/exit animation, enabling users who use the Animated/Reanimated API to handle animations manually. When the `sheetContentDefaultResizeAnimationEnabled` prop is set, we skip performing our resize + translate animation and only adjust the screen size to match the current content.

Fixes: https://github.com/software-mansion/react-native-screens/issues/2560

## Changes

- Updated style to `absoluteWithNoBottom` for android `fitToContents`
- Added default expand/shrink animations
- Added `sheetContentDefaultResizeAnimationEnabled` property for disabling default animation

## Screenshots / GIFs

### Before

<table>
  <tr>
    <td width="50%">
      <video src="https://github.com/user-attachments/assets/b9fc3461-31c7-4a71-9a00-6badd1ba3633"></video>
    </td>
    <td width="50%">
      <video src="https://github.com/user-attachments/assets/75b19679-a500-4586-9b01-33ddd0674530"></video>
    </td>
  </tr>
</table>

### After

<table>
  <tr>
    <td width="50%">
      <video src="https://github.com/user-attachments/assets/ef2fcdd9-7942-40c3-8528-d4d8fdcf4e0e"></video>
    </td>
    <td width="50%">
      <video src="https://github.com/user-attachments/assets/894e02ec-8580-4cb1-a2bd-209977eec548"></video>
    </td>
  </tr>
</table>

Test2560

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -93,7 +93,7 @@ class Screen(
     var sheetClosesOnTouchOutside = true
     var sheetElevation: Float = 24F
     var sheetShouldOverflowTopInset = false
-    var sheetContentDefaultResizeAnimationEnabled = true
+    var sheetDefaultResizeAnimationEnabled = true
 
     /**
      * On Paper, when using form sheet presentation we want to delay enter transition in order
@@ -162,7 +162,7 @@ class Screen(
 
                 if (isInitial) {
                     setupInitialSheetContentHeight(sheetBehavior, height)
-                } else if (sheetContentDefaultResizeAnimationEnabled) {
+                } else if (sheetDefaultResizeAnimationEnabled) {
                     updateSheetContentHeightWithAnimation(sheetBehavior, oldHeight, height)
                 } else {
                     updateSheetContentHeightWithoutAnimation(sheetBehavior, height)
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -519,7 +519,7 @@ export interface ScreenProps extends ViewProps {
    *
    * @platform android
    */
-  sheetContentDefaultResizeAnimationEnabled?: boolean;
+  sheetDefaultResizeAnimationEnabled?: boolean;
   /**
    * How the screen should appear/disappear when pushed or popped at the top of the stack.
    * The following values are currently supported:
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
@@ -277,12 +277,12 @@ open class ScreenViewManager :
         view?.sheetShouldOverflowTopInset = sheetShouldOverflowTopInset
     }
 
-    @ReactProp(name = "sheetContentDefaultResizeAnimationEnabled")
-    override fun setSheetContentDefaultResizeAnimationEnabled(
+    @ReactProp(name = "sheetDefaultResizeAnimationEnabled")
+    override fun setSheetDefaultResizeAnimationEnabled(
         view: Screen?,
-        sheetContentDefaultResizeAnimationEnabled: Boolean,
+        sheetDefaultResizeAnimationEnabled: Boolean,
     ) {
-        view?.sheetContentDefaultResizeAnimationEnabled = sheetContentDefaultResizeAnimationEnabled
+        view?.sheetDefaultResizeAnimationEnabled = sheetDefaultResizeAnimationEnabled
     }
 
     // mark: iOS-only
--- a/android/src/paper/java/com/facebook/react/viewmanagers/RNSScreenManagerDelegate.java
+++ b/android/src/paper/java/com/facebook/react/viewmanagers/RNSScreenManagerDelegate.java
@@ -53,8 +53,8 @@ public class RNSScreenManagerDelegate<T extends View, U extends BaseViewManager<
       case "sheetShouldOverflowTopInset":
         mViewManager.setSheetShouldOverflowTopInset(view, value == null ? false : (boolean) value);
         break;
-      case "sheetContentDefaultResizeAnimationEnabled":
-        mViewManager.setSheetContentDefaultResizeAnimationEnabled(view, value == null ? true : (boolean) value);
+      case "sheetDefaultResizeAnimationEnabled":
+        mViewManager.setSheetDefaultResizeAnimationEnabled(view, value == null ? true : (boolean) value);
         break;
       case "customAnimationOnSwipe":
         mViewManager.setCustomAnimationOnSwipe(view, value == null ? false : (boolean) value);
--- a/android/src/paper/java/com/facebook/react/viewmanagers/RNSScreenManagerInterface.java
+++ b/android/src/paper/java/com/facebook/react/viewmanagers/RNSScreenManagerInterface.java
@@ -25,7 +25,7 @@ public interface RNSScreenManagerInterface<T extends View> extends ViewManagerWi
   void setSheetInitialDetent(T view, int value);
   void setSheetElevation(T view, int value);
   void setSheetShouldOverflowTopInset(T view, boolean value);
-  void setSheetContentDefaultResizeAnimationEnabled(T view, boolean value);
+  void setSheetDefaultResizeAnimationEnabled(T view, boolean value);
   void setCustomAnimationOnSwipe(T view, boolean value);
   void setFullScreenSwipeEnabled(T view, @Nullable String value);
   void setFullScreenSwipeShadowEnabled(T view, boolean value);
--- a/apps/src/tests/Test2560.tsx
+++ b/apps/src/tests/Test2560.tsx
@@ -124,7 +124,7 @@ export default function App() {
             contentStyle: {
               backgroundColor: Colors.YellowLight40,
             },
-            // TODO(@t0maboro) - add `sheetContentDefaultResizeAnimationEnabled` prop here when possible
+            // TODO(@t0maboro) - add `sheetDefaultResizeAnimationEnabled` prop here when possible
           }}
         />
       </Stack.Navigator>
--- a/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
+++ b/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
@@ -260,7 +260,7 @@ When set to `false`, the sheet's layout will be constrained by the inset from th
 
 Defaults to `false`.
 
-### `sheetContentDefaultResizeAnimationEnabled` (Android only)
+### `sheetDefaultResizeAnimationEnabled` (Android only)
 
 Whether the default native animation should be used when the sheet's with `fitToContents` content size changes.
 
--- a/src/components/Screen.tsx
+++ b/src/components/Screen.tsx
@@ -99,7 +99,7 @@ export const InnerScreen = React.forwardRef<View, ScreenProps>(
       sheetElevation = 24,
       sheetInitialDetentIndex = 0,
       sheetShouldOverflowTopInset = false,
-      sheetContentDefaultResizeAnimationEnabled = true,
+      sheetDefaultResizeAnimationEnabled = true,
       // Other
       screenId,
       stackPresentation,
@@ -232,8 +232,8 @@ export const InnerScreen = React.forwardRef<View, ScreenProps>(
             sheetLargestUndimmedDetent={resolvedSheetLargestUndimmedDetent}
             sheetElevation={sheetElevation}
             sheetShouldOverflowTopInset={sheetShouldOverflowTopInset}
-            sheetContentDefaultResizeAnimationEnabled={
-              sheetContentDefaultResizeAnimationEnabled
+            sheetDefaultResizeAnimationEnabled={
+              sheetDefaultResizeAnimationEnabled
             }
             sheetGrabberVisible={sheetGrabberVisible}
             sheetCornerRadius={sheetCornerRadius}
--- a/src/fabric/ModalScreenNativeComponent.ts
+++ b/src/fabric/ModalScreenNativeComponent.ts
@@ -82,7 +82,7 @@ export interface NativeProps extends ViewProps {
   sheetInitialDetent?: CT.WithDefault<CT.Int32, 0>;
   sheetElevation?: CT.WithDefault<CT.Int32, 24>;
   sheetShouldOverflowTopInset?: CT.WithDefault<boolean, false>;
-  sheetContentDefaultResizeAnimationEnabled?: CT.WithDefault<boolean, true>;
+  sheetDefaultResizeAnimationEnabled?: CT.WithDefault<boolean, true>;
   customAnimationOnSwipe?: boolean;
   fullScreenSwipeEnabled?: CT.WithDefault<OptionalBoolean, 'undefined'>;
   fullScreenSwipeShadowEnabled?: CT.WithDefault<boolean, true>;
--- a/src/fabric/ScreenNativeComponent.ts
+++ b/src/fabric/ScreenNativeComponent.ts
@@ -84,7 +84,7 @@ export interface NativeProps extends ViewProps {
   sheetInitialDetent?: CT.WithDefault<CT.Int32, 0>;
   sheetElevation?: CT.WithDefault<CT.Int32, 24>;
   sheetShouldOverflowTopInset?: CT.WithDefault<boolean, false>;
-  sheetContentDefaultResizeAnimationEnabled?: CT.WithDefault<boolean, true>;
+  sheetDefaultResizeAnimationEnabled?: CT.WithDefault<boolean, true>;
   customAnimationOnSwipe?: boolean;
   fullScreenSwipeEnabled?: CT.WithDefault<OptionalBoolean, 'undefined'>;
   fullScreenSwipeShadowEnabled?: CT.WithDefault<boolean, true>;


RohovDmytro:
NOW it's feels like Christmas. 

🎉

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -117,6 +117,8 @@ class Screen(
     private val isNativeStackScreen: Boolean
         get() = container is ScreenStack
 
+    private val sheetHeightAnimationDuration = 500L
```
kligarski: Isn't this too long? From what I can see, in regular enter/exit formSheet animation we use default duration from `ValueAnimator` which [according to docs](https://developer.android.com/reference/android/animation/ValueAnimator#getDuration()) is 300 ms.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -168,6 +175,80 @@ class Screen(
...
+            this
+                .animate()
+                .translationY(0f)
+                .setDuration(sheetHeightAnimationDuration)
+                .withStartAction {
+                    behavior.useSingleDetent(newHeight)
+                    requestLayout()
+                }.withEndAction {
+                    onSheetYTranslationChanged()
+                }.start()
```
kligarski, line 221: Won't using `ViewPropertyAnimator` interfere with `ValueAnimator`s used to animate enter/exit + IME animations? E.g. what would happen if content size changed during enter animation?

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -168,6 +175,80 @@ class Screen(
...
+             * In the animation, we're translating the Screen down by the calculated height delta to the position (which will
+             * be new absolute 0 for the Screen, after ending the transition), providing an impression that FormSheet shrinks.
+             * FormSheet's size remains unchanged during the whole animation, therefore there is no view clipping.
+             *
+             * After animation, we can update the layout: the maximum FormSheet height is updated and we're forcing
+             * another layout pass. Additionally, since the actual layout and the target position are equal,
+             * we can reset translationY to 0.
+             *
+             * After animation, we need to send a notification that ShadowTree state should be updated,
+             * as the FormSheet size has change and the positioning of pressables has changed due to the Y translation manipulation.
```
kligarski:
```suggestion
             * as the FormSheet size has changed and the positioning of pressables has changed due to the Y translation manipulation.
```

android/src/main/java/com/swmansion/rnscreens/Screen.kt
t0maboro: Not sure if this is the use case we should cover. I wouldn't consider such a case, as it would require changing content dynamically during enter/exit animation, what would look strange when the content both slides in/out and jumps

android/src/main/java/com/swmansion/rnscreens/Screen.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3484/commits/2ea04ec901823209c2c14abce50e7b2727986344

android/src/main/java/com/swmansion/rnscreens/Screen.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3484/commits/9e0dfab6b9510528ed3c6271b9d503c3a8ae29f9

t0maboro:
> Also, not sure if this is a real concern but I was able to make the form sheet not adapt to content by changing the content quickly:
> 
>  Screen.Recording.2025-12-18.at.09.04.29.mov 
> I used `TestFormSheet` with `fitToContents` and just added some text at the bottom.

checking..

android/src/main/java/com/swmansion/rnscreens/Screen.kt
kligarski: I wonder if we even need this variable, it seems like `ViewPropertyAnimator` also uses default duration from `ValueAnimator`: https://developer.android.com/reference/android/view/ViewPropertyAnimator#setDuration(long).

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3484/commits/5a049bdd3357b7196f1ebd949b1f6592e125834d

t0maboro:
> Also, not sure if this is a real concern but I was able to make the form sheet not adapt to content by changing the content quickly:
> 
>  Screen.Recording.2025-12-18.at.09.04.29.mov 
> I used `TestFormSheet` with `fitToContents` and just added some text at the bottom.

Should be fixed, tested with

```js
import React, { useRef, useState } from 'react';
import {
  Button,
  View,
  Text,
  StyleSheet,
  Animated,
} from 'react-native';
import {
  createNativeStackNavigator,
  NativeStackScreenProps,
import Colors from '../shared/styling/Colors';

type StackParamList = {
  Home: undefined;
  FormSheet: undefined;
};

const HomeScreen = ({ navigation }: NativeStackScreenProps<StackParamList>) => (
  <View style={styles.screen}>
    <Text style={styles.title}>Home Screen</Text>
    <Button
      title="Open Form Sheet"
      onPress={() => navigation.navigate('FormSheet')}
    />
  </View>
);

const FormSheetScreen = () => {
  const [isTextVisible, setIsTextVisible] = useState(true);

  React.useEffect(() => {
    const timer = setTimeout(() => {
      setIsTextVisible(false);
    }, 1000);

    return () => clearTimeout(timer);
  }, []);

  React.useEffect(() => {
    const timer = setTimeout(() => {
      setIsTextVisible(true);
    }, 1016);

    return () => clearTimeout(timer);
  }, []);

  return (
    <View style={styles.formSheetContainer}>
      <Text style={styles.formSheetTitle}>Form Sheet Content</Text>
      {isTextVisible && (
        <Text>
          Lorem, ipsum dolor sit amet consectetur adipisicing elit. Velit deserunt
          qui fugit unde assumenda. Non id facere mollitia sequi aliquid nostrum,
          tenetur eligendi quos at natus eius, corporis quidem eaque libero
          reiciendis, labore sed minima doloremque temporibus! Veritatis harum
          provident molestias incidunt at temporibus quod ipsam totam optio,
          quisquam amet quae molestiae. Exercitationem animi facere iure, nobis
          nulla repudiandae aut nam natus! Qui, id? Nobis quia recusandae
          repellendus illum voluptas, ipsa, doloribus reprehenderit nulla quas
          facere eos! Alias, accusantium voluptatibus doloremque unde eius totam
          et officia asperiores? Voluptate totam provident, quas inventore
          doloribus autem nihil quisquam soluta eum, dolorum nobis.{' '}
        </Text>
      )}
    </View>
  );
};

  return (
    <NavigationContainer>
      <Stack.Navigator>
        <Stack.Screen
          name="FormSheet"
          component={FormSheetScreen}
          options={{
            presentation: 'formSheet',
            sheetAllowedDetents: 'fitToContents',
            contentStyle: {
              backgroundColor: Colors.YellowLight40,
            },
            // TODO(@t0maboro) - add `sheetContentDefaultResizeAnimationEnabled` prop here when possible
          }}
        />
      </Stack.Navigator>
    </NavigationContainer>
  );
}

  screen: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
    padding: 20,
  },
  title: {
    fontSize: 24,
    marginBottom: 20,
  },
  formSheetContainer: {
    alignItems: 'center',
    justifyContent: 'center',
    padding: 20,
    gap: 20,
  },
  formSheetTitle: {
    fontSize: 20,
    fontWeight: 'bold',
    marginBottom: 10,
  },
  text: {
    fontSize: 16,
    marginBottom: 8,
    color: Colors.White,
  },
  rectangle: {
    width: '100%',
    backgroundColor: Colors.NavyLight80,
    height: 200,
    alignItems: 'center',
    justifyContent: 'center',
  },
  input: {
    width: 100,
    height: 20,
    borderColor: Colors.BlueDark100,
    borderWidth: 1,
  },
});
```

android/src/main/java/com/swmansion/rnscreens/Screen.kt
t0maboro: atm, I added a prop to disable this default animation, allowing to handle the content by developer and react only to the Screen size change on our side - having the combination: TextInput with autoFocus + unmounting another component at the same time (which is a pretty wild combo) - we may have some conflict which might be problematic to synchronize: textInput could want to move the formSheet up, and unmounted component might require to move the component down at the same moment

t0maboro: This is how it works with the custom animation:
https://github.com/user-attachments/assets/99d857a6-a124-4778-b16c-3ea57d78b342

t0maboro: This is how it works with the default one (this flicker is junky, I described the root cause for it above, but anyway, the final state is 'stable')
https://github.com/user-attachments/assets/e721e7d0-265d-41d2-8abf-41ed78c535c4

t0maboro: This is how it could look with more advanced custom animator: 

https://github.com/user-attachments/assets/b5f7db6f-8ab3-492e-94f6-d52e55079379

t0maboro: Therefore, I believe that (at least at the early stage, we can recommend this as the proper approach for handling such cases). I'm okay with continuing the development to synchronize it with our default animation, but as for now, I'd propose to add this support in the separate PR as this one already covers at least a few completely different scenarios. Let me know what do you think about the current state.

GaylordP:
Thank you very much for this PR — looking forward to seeing it merged. :)

marcel-happyfloat:
My formSheet doesn't appear at all in Android with fitToContents, maybe this will help 🙏 😄 . Thanks! ❤️

android/src/main/java/com/swmansion/rnscreens/Screen.kt
kligarski: Sounds good.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -224,21 +235,45 @@ class Screen(
...
+         */
+        behavior.updateMetrics(height)
+        layout(this.left, this.bottom - height, this.right, this.bottom)
+        // Force a layout pass on the CoordinatorLayout to synchronize BottomSheetBehavior's
+        // internal offsets with the new maxHeight. This prevents the sheet from snapping back
+        // to its old position when the user starts a gesture.
+        parent.requestLayout()
+        if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
+            updateScreenSizeFabric(width, height, top + translationY.toInt())
+        }
```
kligarski, line 277: Shouldn't we use `onSheetYTranslationChanged()` instead?

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -156,7 +156,11 @@ class Screen(
...
                 if (shouldAnimateContentHeightChange) {
-                    animateSheetContentHeightChange(sheetBehavior, oldHeight, height)
+                    if (sheetContentDefaultResizeAnimationEnabled) {
+                        animateSheetContentHeightChangeWithDefaultAnimation(sheetBehavior, oldHeight, height)
+                    } else {
+                        updateSheetDetentContentHeightAndLayout(sheetBehavior, height)
+                    }
                 } else {
                     updateSheetDetentWithoutHeightChangeAnimation(sheetBehavior, height)
                 }
```
kligarski: I wonder if we should think about renaming variables and update methods. What I don't like is that:
- first condition is `shouldAnimateContentHeightChange`but if second is false, we're not animating anything - this might be misleading
- if I understand correctly, `updateSheetDetentWithoutHeightChangeAnimation` is used for initial layout but the name suggests it's just update without animation (this name would fit better to 2nd function here)

Maybe we should consider a method named like `updateSheetDetentContentHeight` with `animated: Boolean` as param that would call `update...WithoutAnimation` or `update...WithAnimation` and separate method for updating height for initial layout?

android/src/main/java/com/swmansion/rnscreens/Screen.kt
t0maboro: I'm not changing translation anywhere in this code branch, therefore I'd prefer to make an explicit call, even if it's duplication

kligarski: That's fair.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3484/commits/bc2e458629a83eb4ab72f4b03b0b8a1048548270

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -168,6 +178,142 @@ class Screen(
...
+    private fun updateSheetDetentContentHeight(
+        sheetBehavior: BottomSheetBehavior<Screen>,
+        newHeight: Int,
+        animated: Boolean,
+    ) {
+        if (animated) {
+            updateSheetDetentContentHeightAndLayout(sheetBehavior, newHeight)
+        } else {
+            updateSheetDetentWithoutHeightChangeAnimation(sheetBehavior, newHeight)
+        }
```
kligarski: I don't think that this makes sense now - `animated == true` means that we're using custom animation (=> that means that the user will animate it themselves and we're just adjusting the height _without_ any animation)?

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -147,12 +149,20 @@ class Screen(
...
+                val shouldAnimateContentHeightChange = oldHeight > 0 && oldHeight != height
+
+                val sheetContentHeightAnimation =
+                    when {
+                        !shouldAnimateContentHeightChange -> FormSheetSizeChangeAnimationType.NONE
+                        sheetContentDefaultResizeAnimationEnabled -> FormSheetSizeChangeAnimationType.DEFAULT
+                        else -> FormSheetSizeChangeAnimationType.CUSTOM
+                    }
+
+                updateSheetContentHeight(sheetBehavior, oldHeight, height, sheetContentHeightAnimation)
```
kligarski: I'm not sure if we need the enum. My suggestion:

`animateSheetContentHeightChangeWithDefaultAnimation` -> `updateSheetContentHeightWithDefaultAnimation`
`updateSheetDetentContentHeightAndLayout` -> `updateSheetContentHeightWithoutAnimation`
`updateSheetDetentWithoutHeightChangeAnimation` -> I'm not sure what name should we use here, this is called only if `oldHeight == 0` (initial I guess?) or when height did not change.

and then just run those methods with `when`.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3484/commits/f4475fa2909e42a02a02c9425c6708f33c20ba0a

android/src/main/java/com/swmansion/rnscreens/Screen.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3484/commits/f4475fa2909e42a02a02c9425c6708f33c20ba0a

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -147,12 +149,24 @@ class Screen(
...
+                    return
+                }
+
+                if (isInitial) {
+                    setupInitialSheetContentHeight(sheetBehavior, height)
+                } else if (sheetContentDefaultResizeAnimationEnabled) {
+                    updateSheetContentHeightWithDefaultAnimation(sheetBehavior, oldHeight, height)
+                } else {
+                    updateSheetContentHeightWithoutDefaultAnimation(sheetBehavior, height)
+                }
```
kligarski: I'm wondering if we should use `DefaultAnimation` name or just `Animation`. In theory, we're using default values for the animation but on the other hand, we added it so I'm not sure if we should call it "default".

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3484/commits/830cf1a1fb5532434e50f1548b9f48aab1481036

src/types.tsx
```diff
@@ -503,6 +503,23 @@ export interface ScreenProps extends ViewProps {
...
+   *
+   * When set to `false`, the internal animation and translation logic is ignored. This
+   * allows the sheet to adjust its size dynamically based on the current dimensions of
+   * the content provided by the developer, allowing implementing custom resizing animations.
+   *
+   * Defaults to `true`.
+   *
+   * @platform android
+   */
+  sheetContentDefaultResizeAnimationEnabled?: boolean;
```
kkafar:
```suggestion
  sheetDefaultResizeAnimationEnabled?: boolean;
```

I'd remove the `Content` part here, as we do not apply any animation to the sheet contents. We animate only the `Screen` component natively.

kkafar (commented):
One additional remark regarding naming.

src/types.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3484/commits/3794f4de50cec57220479e73063478b305467fa1

t0maboro:
> Before merging I asked internally for clarification what is `absoluteWoBottom` style useful for and we still have some discussion about timing of component unmount. Will progress shortly.

https://github.com/software-mansion/react-native-screens/pull/3484/commits/f69b071c9d91d5bec224895f9cfda5c2cdf88db3

PR #3487, opened 2025-12-16
chore(Split): reorganize Split component exports and naming

kkafar:
## Description

Closes https://github.com/software-mansion/react-native-screens-labs/issues/735

I'm open to discussion on the naming here. 

This PR changes only JS layer of the component naming. If necessary native-side will be handled separately. 

## Changes

- **Export split component as `Split` instead of separate ones**
  I've removed exports for `SplitViewHost` & `SplitViewScreen` in favour
  of `Split.Host` & `Split.Column` / `Split.Inspector`.
  
  The naming here is open for discussion. I prefer short version of `Split`
  instead of `SplitView`, but that's matter of preference, can't find
  reasonable (& clinching) arguments for one or the other.
  
  Most important thing here is to just make it consistent. If we are to
  use `View` suffix, then we should do similar thing for Stack ->
  `StackView.Host` & `StackView.Screen`.
  
  So it's either:
  
  `Stack.Host`, `Stack.Screen`,
  `Tabs.Host`, `Tabs.Screen`,
  `Split.Host`, `Split.Column`, `Split.Inspector`
  
  OR
  
  `StackView.Host`, `StackView.Screen`,
  `TabsView.Host`, `TabsView.Screen`,
  `SplitView.Host`, `SplitView.Column`, `SplitView.Inspector`
  
  OR MAYBE (?)
  
  `Stack.HostView`, `Stack.ScreenView`,
  `Tabs.HostView`, `Tabs.ScreenView`,
  `Split.HostView`, `Split.ColumnView`, `Split.InspectorView`
  

- **Update examples**
  

- **Remove `view` substring from `SplitView` prefix used across symbols**

### DIFF
--- a/src/components/gamma/split/SplitHost.types.ts
+++ b/src/components/gamma/split/SplitHost.types.ts
@@ -296,7 +296,7 @@ export interface SplitHostProps extends ViewProps {
   /**
    * @summary Determines whether inspector column should be displayed.
    *
-   * Inspector will be displayed on the trailing edge of the main (secondary) column (for expanded Split) or as a modal (for collapsed SplitView).
+   * Inspector will be displayed on the trailing edge of the main (secondary) column (for expanded Split) or as a modal (for collapsed Split).
    * The result on the interface for this prop depends on whether the Split is collapsed or expanded.
    *
    * @supported iOS 26 or higher


apps/src/tests/TestSplit/SplitBaseApp.tsx
```diff
@@ -1,25 +1,25 @@
...
 import { StyleSheet, Text, View } from 'react-native';
-import { SplitViewHost, SplitViewScreen } from 'react-native-screens/experimental';
+import { Split } from 'react-native-screens/experimental';
 import { Colors } from '../../shared/styling/Colors';
 import { TestBottomTabs, TestScreenStack } from '..';
-import { SplitViewBaseConfig } from './helpers/types';
+import { SplitBaseConfig } from './helpers/types';
 
-const SplitViewBaseApp = ({ splitViewBaseConfig }: { splitViewBaseConfig: SplitViewBaseConfig }) => {
+const SplitBaseApp = ({ splitViewBaseConfig }: { splitViewBaseConfig: SplitBaseConfig }) => {
```
t0maboro: nit: splitViewBaseConfig -> splitBaseConfig - applies to all, but might be done as a followup

kkafar: Done somewhere after rebase & force push

kkafar:
Yeah, you're right. Thanks for noticing this. I'll fix the warning messages.

apps/src/tests/TestSplit/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const App = () => {
+  const splitViewBaseConfig: SplitBaseConfig = {
+    preferredDisplayMode: 'twoBesideSecondary',
+    preferredSplitBehavior: 'tile',
+  };
+
+  return (
+    <SplitBaseApp splitBaseConfig={splitViewBaseConfig} />
+  );
```
t0maboro:
```suggestion
  const splitBaseConfig: SplitBaseConfig = {
    preferredDisplayMode: 'twoBesideSecondary',
    preferredSplitBehavior: 'tile',
  };

  return (
    <SplitBaseApp splitBaseConfig={splitBaseConfig} />
  );
```

kkafar:
This is back to draft for some time, as thanks to some recent commit the `TestSplit` example stopped to work.

kkafar:
Works now

kkafar:
> SplitHost.tsx, L79 : This enables us to fully recreate the SplitView when necessary, ensuring the correct column configuration is always applied.

@t0maboro this one is left deliberately, as this one refers to the `SplitViewHostNativeComponent` / UIKit view, name of which remains unchanged.

PR #3492, opened 2025-12-17
feat(Tabs): allow full tabs background color styling

kligarski:
## Description

Adds customization of `backgroundColor` for native tabs container (`UITabBarController.view` on iOS, `TabsHost` on Android) and tab screen (`RNSBottomTabsScreenComponentView` on iOS, `TabScreen` on Android).

| Android | iOS |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/6dfb2628-560c-42d3-a380-6a858e7a8534" /> | <video src="https://github.com/user-attachments/assets/545dfcf3-7924-4c2d-ad7f-2ca00d369cb8" /> |

Closes https://github.com/software-mansion/react-native-screens-labs/issues/721.
Supersedes https://github.com/software-mansion/react-native-screens/pull/3478.

### Comparison of different approaches to setting background color on iOS

This PR allows for (at least) 3 kinds of setting background color in tabs. Due to asynchronous nature, different approaches result in different effect.

#### Tab screen + React View has background color

This is similar to current approach where tab screen uses `systemBackgroundColor` and user defines background color via View inside components passed to tab screen. This approach was used to mitigate problems with liquid glass tab bar appearance adaptation (more details: https://github.com/software-mansion/react-native-screens/pull/3279).

https://github.com/user-attachments/assets/49397de6-77d2-4d70-8412-6681fb361ee1

First, background color of tab screen is visible before tab's content is rendered by React and correct color appears.

#### Only React View has background color

This is similar to approach suggested in https://github.com/software-mansion/react-native-screens/pull/3478 (transparent screen). While this seems to work better on first tab renders, switching between similar tabs after they are rendered causes a flash.

https://github.com/user-attachments/assets/816e59ad-6265-4c84-97c5-4ca64ca8341b

This is reproducible in bare UIKit app when child VC's view has clear background color and its subview has different background color.

#### Only Tab screen has background color

This approach is possible thanks to this PR and it is closes to native platform but tab's content still appears after a delay because it needs to be rendered by React - this causes a flash on first render.

https://github.com/user-attachments/assets/8883acfe-d353-4f8c-90f7-462b46f4bf79

## Changes

- add `nativeContainerStyle` prop to `BottomTabs` component with `backgroundColor` field
- add `style` prop to `BottomTabsScreen` component with `backgroundColor` field
- add `Test3492`

Run `Test3492`.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
@@ -9,6 +9,7 @@ import android.view.View
 import android.view.WindowInsets
 import android.widget.FrameLayout
 import androidx.appcompat.view.ContextThemeWrapper
+import androidx.core.graphics.drawable.toDrawable
 import androidx.core.view.children
 import androidx.fragment.app.FragmentManager
 import com.facebook.react.modules.core.ReactChoreographer
@@ -226,6 +227,12 @@ class TabsHost(
         if (newValue != oldValue) {
             updateInterfaceInsets()
             updateNavigationMenuIfNeeded(oldValue, newValue)
+        }
+    }
+
+    var nativeContainerBackgroundColor: Int? by Delegates.observable(null) { _, oldValue, newValue ->
+        if (newValue != oldValue) {
+            background = newValue?.toDrawable()
         }
     }
 
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHostViewManager.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHostViewManager.kt
@@ -150,6 +150,14 @@ class TabsHostViewManager :
         view.tabBarHidden = value
     }
 
+    @ReactProp(name = "nativeContainerBackgroundColor", customType = "Color")
+    override fun setNativeContainerBackgroundColor(
+        view: TabsHost,
+        value: Int?,
+    ) {
+        view.nativeContainerBackgroundColor = value
+    }
+
     // Android additional
 
     @ReactProp(name = "tabBarItemTitleFontColorActive", customType = "Color")
--- a/android/src/paper/java/com/facebook/react/viewmanagers/RNSBottomTabsManagerDelegate.java
+++ b/android/src/paper/java/com/facebook/react/viewmanagers/RNSBottomTabsManagerDelegate.java
@@ -24,6 +24,12 @@ public class RNSBottomTabsManagerDelegate<T extends View, U extends BaseViewMana
   @Override
   public void setProperty(T view, String propName, @Nullable Object value) {
     switch (propName) {
+      case "tabBarHidden":
+        mViewManager.setTabBarHidden(view, value == null ? false : (boolean) value);
+        break;
+      case "nativeContainerBackgroundColor":
+        mViewManager.setNativeContainerBackgroundColor(view, ColorPropConverter.getColor(value, view.getContext()));
+        break;
       case "tabBarBackgroundColor":
         mViewManager.setTabBarBackgroundColor(view, ColorPropConverter.getColor(value, view.getContext()));
         break;
@@ -74,9 +80,6 @@ public class RNSBottomTabsManagerDelegate<T extends View, U extends BaseViewMana
         break;
       case "tabBarControllerMode":
         mViewManager.setTabBarControllerMode(view, (String) value);
-        break;
-      case "tabBarHidden":
-        mViewManager.setTabBarHidden(view, value == null ? false : (boolean) value);
         break;
       case "controlNavigationStateInJS":
         mViewManager.setControlNavigationStateInJS(view, value == null ? false : (boolean) value);
--- a/android/src/paper/java/com/facebook/react/viewmanagers/RNSBottomTabsManagerInterface.java
+++ b/android/src/paper/java/com/facebook/react/viewmanagers/RNSBottomTabsManagerInterface.java
@@ -14,6 +14,8 @@ import androidx.annotation.Nullable;
 import com.facebook.react.uimanager.ViewManagerWithGeneratedInterface;
 
 public interface RNSBottomTabsManagerInterface<T extends View> extends ViewManagerWithGeneratedInterface {
+  void setTabBarHidden(T view, boolean value);
+  void setNativeContainerBackgroundColor(T view, @Nullable Integer value);
   void setTabBarBackgroundColor(T view, @Nullable Integer value);
   void setTabBarItemTitleFontFamily(T view, @Nullable String value);
   void setTabBarItemTitleFontSize(T view, float value);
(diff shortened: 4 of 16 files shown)


t0maboro (approved):

kkafar (commented):
I've watched the videos carefully and here are my initial thoughts:

Please take a look at the "bg_only_react" video. Notice that after you press on "tab 2" (first change to "tab 2"), the view does not change immediately (**or maybe it does, but it is fully transparent, as there is no content?** - this is likely what happens). This leads to situation where we get this seemingly nice transition w/o effect where content appears on previously empty background after a second. I'm pretty sure, that in case the render is heavy (takes like 500 ms) we'll observe that tab changes nicely, but you still can see content from previous screen for those couple hundreds of ms. So this approach does not solve the initial problem, it mitigates it by accidentally displaying transparent content until React mounts rendered views.

I'll give code remarks & suggestions in separate review

kkafar:
Also from these videos & PR description I do not quite understand impact of setting background colour on `UITabBarController.view`. The videos only talk about setting it on `TabScreenComponentView` and inside the react content. Can you describe what's the impact of setting it (background colour) on "main container view"?

kkafar (commented):
My current understanding is that with leaving the main container view transparent and setting the background colour for the "TabScreenView" only after first content render we could mitigate the problems pretty nicely? 

But would glass effects on tab bar behave nicely in such configuration?

kkafar (commented):
The code itself looks good. I don't have any bigger remarks. But let's discuss the approach first ☝🏻

kligarski:
> Please take a look at the "bg_only_react" video. Notice that after you press on "tab 2" (first change to "tab 2"), the view does not change immediately (**or maybe it does, but it is fully transparent, as there is no content?** - this is likely what happens). This leads to situation where we get this seemingly nice transition w/o effect where content appears on previously empty background after a second. I'm pretty sure, that in case the render is heavy (takes like 500 ms) we'll observe that tab changes nicely, but you still can see content from previous screen for those couple hundreds of ms. So this approach does not solve the initial problem, it mitigates it by accidentally displaying transparent content until React mounts rendered views.

You're right here. What happens is:
- config tab starts fading out
- container background becomes therefore visible
- tab2 fades in but because its content isn't rendered yet, there is only transparent background (no visual impact here)
- tab2's content appears after it is rendered & mounted by React

Here is the video with heavy tab2 render:

https://github.com/user-attachments/assets/6abf9b7f-abac-4937-8058-9415ed93c763

> Also from these videos & PR description I do not quite understand impact of setting background colour on `UITabBarController.view`. The videos only talk about setting it on `TabScreenComponentView` and inside the react content. Can you describe what's the impact of setting it (background colour) on "main container view"?

This is relevant when we try "bg-react-only" approach with transparent `TabScreenComponentView`s. Here is the same video as above but with blue background:

https://github.com/user-attachments/assets/35a17f2e-5942-40ae-ae31-4b2d08a25e64

Also, even if the tab content has been rendered, the container background is visible during the transition. Here is the video (container background is blue as well):

https://github.com/user-attachments/assets/bb89c790-7512-458b-b01e-2830638e0305

I haven't tested yet how tab bar appearance adaptation works with each of those approaches but previously having solid background for `TabScreenComponentView` (it is currently `systemBackground`) helped mitigate some issues previously (details: https://github.com/software-mansion/react-native-screens/pull/3279). 

> My current understanding is that with leaving the main container view transparent and setting the background colour for the "TabScreenView" only after first content render we could mitigate the problems pretty nicely?
> 
> But would glass effects on tab bar behave nicely in such configuration?

I'm not exactly sure what you mean here. "bg-only-tab"/"bg-tab-and-react" + preload should behave well.

src/components/bottom-tabs/BottomTabs.types.ts
```diff
@@ -33,7 +33,16 @@ export type TabBarMinimizeBehavior =
 // iOS-specific
 export type TabBarControllerMode = 'automatic' | 'tabBar' | 'tabSidebar';
 
-export interface BottomTabsProps extends ViewProps {
+export type BottomTabsNativeContainerStyleProps = {
```
kkafar: Let's refactor this PR after https://github.com/software-mansion/react-native-screens/pull/3535 is merged.

kkafar (approved):
The code looks good. We must only make sure to coordinate the changes with https://github.com/software-mansion/react-native-screens/pull/3535
Let's wait for that PR first & then adjust the naming here.

kkafar:
I've merged #3535. We can rebase / merge main to this PR and proceed here.

kkafar (approved):
lgtm

Brma1048:
@kligarski is there also a option to set the backgroundColor for the tabBarBadge on iOS 26?

kligarski:
> @kligarski is there also a option to set the backgroundColor for the tabBarBadge on iOS 26?

There is `tabBarItemBadgeBackgroundColor` prop (see here: https://github.com/software-mansion/react-native-screens/blob/main/src/components/tabs/TabsScreen.types.ts#L257). It changes color for all badges in the tab bar when given tab is selected. It should work on iOS 26 as well.

Brma1048:
@kligarski on iPad it fallsback to red, is there a way to change the iPad tabbar badge background color?
SCR-20260617-imjt

kligarski:
Unfortunately, this seems to be a bug in UIKit. The same happens with pure native app (without `react-native` and `react-native-screens`). I also checked iOS 27 and it's still the same.

Screenshot 2026-06-19 at 14 32 00

Here is the native reproduction I checked.

<details>
<summary>UIKit reproduction</summary>

```swift
import UIKit

class MainTabBarController: UITabBarController {

    override func viewDidLoad() {
        super.viewDidLoad()
        
        setupAppearance()
        setupViewControllers()
    }
    
    private func setupAppearance() {
        let appearance = UITabBarAppearance()
        
        let itemAppearance = UITabBarItemAppearance()
        itemAppearance.normal.badgeBackgroundColor = .cyan
        itemAppearance.selected.badgeBackgroundColor = .green
      
        appearance.stackedLayoutAppearance = itemAppearance
        appearance.inlineLayoutAppearance = itemAppearance
        appearance.compactInlineLayoutAppearance = itemAppearance
        
        tabBar.standardAppearance = appearance
        tabBar.scrollEdgeAppearance = appearance
    }
    
    private func setupViewControllers() {
        let homeVC = UIViewController()
        homeVC.view.backgroundColor = .systemBackground
        homeVC.tabBarItem = UITabBarItem(title: "Home", image: UIImage(systemName: "house"), tag: 0)
      
        // test legacy setting as well
        homeVC.tabBarItem.badgeColor = .cyan
        
        homeVC.tabBarItem.badgeValue = "3"
        
        let settingsVC = UIViewController()
        settingsVC.view.backgroundColor = .systemGroupedBackground
        settingsVC.tabBarItem = UITabBarItem(title: "Settings", image: UIImage(systemName: "gear"), tag: 1)
      
        settingsVC.tabBarItem.badgeValue = "15"

        viewControllers = [homeVC, settingsVC]
    }
}

```
</details>

PR #3493, opened 2025-12-17
chore(Stack): refactor component definition & nomenclature on native side

kkafar:
## Description

Closes https://github.com/software-mansion/react-native-screens-labs/issues/743

## Changes

- **Move stack related native components to dedicated directory**
- **Rename native spec file of `StackHost` component**
- **Change `RNSScreenStackHost` component definition to `RNSStackHost`**
- **Rename `RNSScreenStackHostComponentView` -> `RNSStackHostComponentView`**

### DIFF
--- a/ios/gamma/stack/RNSStackScreenComponentView.h
+++ b/ios/gamma/stack/RNSStackScreenComponentView.h
@@ -9,9 +9,9 @@ NS_ASSUME_NONNULL_BEGIN
 @class RNSStackHostComponentView;
 
 typedef NS_ENUM(int, RNSStackScreenLifecycleState) {
-  RNSScreenStackLifecycleInitial = 0,
-  RNSScreenStackLifecycleDetached = 1,
-  RNSScreenStackLifecycleAttached = 2,
+  RNSStackScreenLifecycleInitial = 0,
+  RNSStackScreenLifecycleDetached = 1,
+  RNSStackScreenLifecycleAttached = 2,
 };
 
 @interface RNSStackScreenComponentView : RNSReactBaseView
--- a/ios/gamma/stack/RNSStackScreenComponentView.mm
+++ b/ios/gamma/stack/RNSStackScreenComponentView.mm
@@ -49,7 +49,7 @@ namespace react = facebook::react;
 
   // container state
   _screenKey = nil;
-  _maxLifecycleState = RNSScreenStackLifecycleInitial;
+  _maxLifecycleState = RNSStackScreenLifecycleInitial;
 }
 
 - (void)setupController


src/components/gamma/stack/StackHost.web.tsx
```diff
@@ -1,7 +1,8 @@
 import { View, ViewProps } from 'react-native';
 
+// TODO: Determine what this type is & whether it should not be removed
```
kligarski, line 3: I think that previously we had `ScreenStackNativeProps` which was exported from component file and to keep web version in sync we also exported it from web implementation but after this refactor: https://github.com/software-mansion/react-native-screens/commit/c1ecf0dd8fb05b399c620f97a5d191c5105a2604#diff-ce8d6cce55638ffd308f59d85ff9d75285083b51e8f56ef7ca77fda467bfcdbfR3 this shouldn't be necessary.

ios/gamma/stack/RNSStackScreenComponentView.h
```diff
@@ -6,7 +6,7 @@
 NS_ASSUME_NONNULL_BEGIN
 
 @class RNSStackScreenController;
-@class RNSScreenStackHostComponentView;
+@class RNSStackHostComponentView;
 
 typedef NS_ENUM(int, RNSScreenStackLifecycleState) {
```
kligarski: It's probably going to be changed either way but to keep things consistent we should also update the name of this enum and its values.

```suggestion
typedef NS_ENUM(int, RNSStackScreenLifecycleState) {
```

src/components/gamma/stack/StackHost.web.tsx
kkafar: Nice, thanks for the information here. I'll open ticket for that

kkafar: https://github.com/software-mansion/react-native-screens-labs/issues/748

ios/gamma/stack/RNSStackScreenComponentView.h
kkafar: I've left these deliberately, as they are to be removed soon, but I guess you're right.

kkafar: [ff9a005](https://github.com/software-mansion/react-native-screens/pull/3493/commits/ff9a005cd8660000b82da3f55fd5a4970570d32d)

ios/gamma/stack/RNSStackScreenController.swift
```diff
@@ -3,13 +3,13 @@ import UIKit
 
 @objc
 public class RNSStackScreenController: UIViewController {
-  let screenStackComponentView: RNSStackScreenComponentView
+  let screen: RNSStackScreenComponentView
```
kligarski, line 6: One inconsistency is that in `RNSStackController` we have full name `stackHostComponentView` but here we have `screen`. Not sure what is the best approach here, maybe `hostView` and `screenView`?

kkafar: I see! But we'll rewrite this almost from scratch 🔜 I won't linger now

PR #3497, opened 2025-12-18
feat: Add testId and accessibilityLabel for BottomTabsScreen and Item

kmichalikk:
Closes https://github.com/software-mansion/react-native-screens-labs/issues/503

## Description

This PR adds `testId` and `accessibilityLabel` support for BottomTabsScreen and Item. For iOS, this maps to `accessibilityIdentifier` and `accessibilityLabel`, respectively. For android, things are more complicated when it comes to the item (the screen is handled automatically by react native in BaseViewManager):
- `testId` is mapped to `tag` since this is what appeared to be working with Detox, which uses espresso under the hood
- `accessibilityLabel` is mapped to the part of the contentDescription that overwrites tab title. With tab badge present, the final description is the title and badge description combined. Detox seems to be matching the parts of the description correctly with only the title/label as the selector. **This prop is only available on android on API level 26 and up**.

## Changes

Added 4 props to the bottom tab configuration object:
- screenTestId
- tabBarItemTestId
- screenAccessibilityLabel
- tabBarItemAccessibilityLabel

Added `nativeBottomTabs.e2e.ts` test.

Run the e2e test. Use e2e-utils from https://github.com/software-mansion/react-native-screens/pull/3430/changes#diff-04033cb65a060a8211fbcd7be662a11be0f32c9b98d2f0d510c489e4e74af450, otherwise iOS won't complete.

### DIFF
--- a/ios/bottom-tabs/host/RNSTabBarController.mm
+++ b/ios/bottom-tabs/host/RNSTabBarController.mm
@@ -187,8 +187,8 @@
 
 - (void)updateTabBarA11yIfNeeded
 {
-  for (RNSTabsScreenViewController *tabViewController in self.viewControllers) {
-    auto screenView = static_cast<RNSBottomTabsScreenComponentView *>(tabViewController.tabScreenComponentView);
+  for (UIViewController *tabViewController in self.viewControllers) {
+    auto screenView = static_cast<RNSTabsScreenViewController *>(tabViewController).tabScreenComponentView;
     if (!screenView.tabBarItemNeedsA11yUpdate) {
       continue;
     }
--- a/ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
+++ b/ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
@@ -515,6 +515,7 @@ RNS_IGNORE_SUPER_CALL_END
     _tabBarItemNeedsRecreation = NO;
 
     _tabBarItemNeedsUpdate = YES;
+    _tabBarItemNeedsA11yUpdate = YES;
   }
 
   if (_tabBarItemNeedsUpdate) {
@@ -677,7 +678,6 @@ RNS_IGNORE_SUPER_CALL_END
 {
   _systemItem = systemItem;
   _tabBarItemNeedsRecreation = YES;
-  _tabBarItemNeedsA11yUpdate = YES;
 }
 
 - (void)setSpecialEffects:(NSDictionary *)specialEffects


android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
```diff
@@ -339,6 +341,7 @@ class TabsHost(
     override fun onMenuItemAttributesChange(tabScreen: TabScreen) {
         getMenuItemForTabScreen(tabScreen)?.let { menuItem ->
             appearanceCoordinator.updateMenuItemAppearance(menuItem, tabScreen)
+            a11yCoordinator.setA11yPropertiesToTabItems()
```
kkafar: I think that this call should update the a11y only for the tab screen indicated by the argument passed to the `onMenuItemAttributesChange` function.

Right now you update them for all screens (tabs), ignoring the fact that the attributes could have changed only for a single screen.

ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
```diff
@@ -188,6 +191,9 @@ - (void)createTabBarItem
     tabBarItem = [[UITabBarItem alloc] init];
   }
 
+  tabBarItem.accessibilityIdentifier = _tabBarItemTestId;
```
kkafar: The code looks good, but I think it is in wrong place.

We should keep the flow. Instead of setting these props here, the intention of update should be propagated to tab bar controller (see how appearance is handled) and then actual props should be updated in `RNSTabBarController.reactMoutingTransactionDidMount`, e.g. after appearance pass is done.

ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
```diff
@@ -293,6 +299,16 @@ - (void)updateProps:(const facebook::react::Props::Shared &)props
...
   }
 
+  if (newComponentProps.tabBarItemTestId != oldComponentProps.tabBarItemTestId) {
+    _tabBarItemTestId = RCTNSStringFromStringNilIfEmpty(newComponentProps.tabBarItemTestId);
+    tabBarItemNeedsRecreation = YES;
+  }
+
+  if (newComponentProps.tabBarItemAccessibilityLabel != oldComponentProps.tabBarItemAccessibilityLabel) {
+    _tabBarItemAccessibilityLabel = RCTNSStringFromStringNilIfEmpty(newComponentProps.tabBarItemAccessibilityLabel);
+    tabBarItemNeedsRecreation = YES;
```
kkafar: Is recreation necessary here? Can't we just update these props?

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -384,6 +384,32 @@ export interface BottomTabsScreenProps {
   freezeContents?: boolean;
   // #endregion General
 
+  // #region Accessibility
+
+  /**
+   * @summary testId for the BottomTabScreen
+   */
+  screenTestId?: string;
```
kligarski: Should we use `screen` prefix here if prop is defined on `BottomTabsScreen?`

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
kmichalikk: Reworked it a bit; it turned out that it is a happy coincidence that it worked in the first attempt, because at the time of setting the values first time, menu item for the screen (or vice versa) may be null. Appearance updates work around this by running forEach later. Now I'm running forEach for the first time and single-screen updates later.

974fbac2177b3ab236ad2b30c36a0677de9979a1

ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
kmichalikk: 88ef668cb85f6057270540df1fa22ab11c0ee3ae

let me know if it could be implemented this way?

ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
```diff
@@ -340,6 +353,17 @@ - (void)updateProps:(const facebook::react::Props::Shared &)props
     tabItemNeedsAppearanceUpdate = YES;
   }
 
+  if (newComponentProps.tabBarItemTestID != oldComponentProps.tabBarItemTestID) {
```
t0maboro: This condition is repeated. Any reason for splitting these blocks?

ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
```diff
@@ -340,6 +353,17 @@ - (void)updateProps:(const facebook::react::Props::Shared &)props
...
+    _controller.tabItemTestID = RCTNSStringFromStringNilIfEmpty(newComponentProps.tabBarItemTestID);
+    tabBarItemNeedsA11yUpdate = YES;
+  }
+
+  if (newComponentProps.tabBarItemAccessibilityLabel != oldComponentProps.tabBarItemAccessibilityLabel) {
```
t0maboro: This condition is repeated. Any reason for splitting these blocks?

ios/bottom-tabs/screen/RNSTabsScreenViewController.h
```diff
@@ -51,6 +53,8 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)clearTabsSpecialEffectsDelegateIfNeeded:(nonnull id<RNSBottomTabsSpecialEffectsSupporting>)delegate;
 
+- (void)updateTabItemA11yProps;
```
t0maboro: nit: other methods in the header are documented, we may consider adding 1 sentence here for consistency

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -384,6 +384,32 @@ export interface BottomTabsScreenProps {
   freezeContents?: boolean;
   // #endregion General
 
+  // #region Accessibility
+
+  /**
+   * @summary testId for the BottomTabScreen
```
t0maboro:
```suggestion
   * @summary testID for the BottomTabScreen
```

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -384,6 +384,32 @@ export interface BottomTabsScreenProps {
...
+   */
+  testID?: string;
+
+  /**
+   * @summary accessibilityLabel for the BottomTabScreen
+   */
+  accessibilityLabel?: string;
+
+  /**
+   * @summary testId for the TabBarItem
```
t0maboro:
```suggestion
   * @summary testID for the TabBarItem
```

kmichalikk: [3544f85](https://github.com/software-mansion/react-native-screens/pull/3497/commits/3544f85c8b88c817cd2a10d3a376988696366a7c)

src/components/bottom-tabs/BottomTabsScreen.types.ts
kmichalikk: [3544f85](https://github.com/software-mansion/react-native-screens/pull/3497/commits/3544f85c8b88c817cd2a10d3a376988696366a7c)

ios/bottom-tabs/screen/RNSTabsScreenViewController.h
kmichalikk: [3544f85](https://github.com/software-mansion/react-native-screens/pull/3497/commits/3544f85c8b88c817cd2a10d3a376988696366a7c)

ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
kmichalikk: My mistake, intended leave only the second one, but it turns out both are needed.

[3544f85](https://github.com/software-mansion/react-native-screens/pull/3497/commits/3544f85c8b88c817cd2a10d3a376988696366a7c)

ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
kmichalikk: [3544f85](https://github.com/software-mansion/react-native-screens/pull/3497/commits/3544f85c8b88c817cd2a10d3a376988696366a7c)

kmichalikk:
I'm not 100% positive we don't need to recreate the tab item when changing the id / label.

Please check and let me know if adding the following code to bottom tabs' index.tsx to update the props programmatically after 2s actually changes the props (on iOS, with accessibility inspector).

```ts
  const [tabConfigs, setTabConfigs] = React.useState(TAB_CONFIGS);

  useEffect(() => {
    setTimeout(() => {
      setTabConfigs([
        { ...TAB_CONFIGS[0], tabScreenProps: { ...TAB_CONFIGS[0].tabScreenProps, tabBarItemTestID: 'pomidor0', tabBarItemAccessibilityLabel: 'pomidor0' } },
        { ...TAB_CONFIGS[1], tabScreenProps: { ...TAB_CONFIGS[1].tabScreenProps, tabBarItemTestID: 'pomidor1', tabBarItemAccessibilityLabel: 'pomidor1' } },
        { ...TAB_CONFIGS[2], tabScreenProps: { ...TAB_CONFIGS[2].tabScreenProps, tabBarItemTestID: 'pomidor2', tabBarItemAccessibilityLabel: 'pomidor2' } },
        { ...TAB_CONFIGS[3], tabScreenProps: { ...TAB_CONFIGS[3].tabScreenProps, tabBarItemTestID: 'pomidor3', tabBarItemAccessibilityLabel: 'pomidor3' } },
      ])
    }, 2000);
  }, []);

...

      <BottomTabsContainer
        tabConfigs={tabConfigs}
```

src/components/bottom-tabs/BottomTabsScreen.types.ts
```diff
@@ -384,6 +384,32 @@ export interface BottomTabsScreenProps {
...
+
+  /**
+   * @summary testID for the BottomTabScreen
+   */
+  testID?: string;
+
+  /**
+   * @summary accessibilityLabel for the BottomTabScreen
+   */
+  accessibilityLabel?: string;
```
kkafar, line 397: Here you define `testID` & `accessibilityLabel` for the `TabScreen` itself, however please note that `TabScreen` Android implementation does not inherit form `ReactViewGroup`. How does it work then? Could you describe it in PR description, please? 

Is is the question of `ReactViewManager` using the accessibility delegate & not relying on state defined explicitly in `ReactViewGroup`?

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHostA11yCoordinator.kt
```diff
@@ -0,0 +1,35 @@
...
+            // The setting will overwrite default item title when building a11y description
+            // see https://github.com/material-components/material-components-android/blob/9cc5d57d7b5a41d0ceafa29b98d2cd6d4094d19c/lib/java/com/google/android/material/navigation/NavigationBarItemView.java#L692
+            menuItem.contentDescription = tabScreen.tabBarItemAccessibilityLabel
+        }
+
+        // when matching view by id, espresso driver seems to look for tag property
+        menuView.tag = tabScreen.tabBarItemTestID
+    }
+
+    fun setA11yPropertiesToTabItems() {
```
kkafar: IMO this improves the readability by much. 

```suggestion
    fun setA11yPropertiesToAllTabItems() {
```

ios/bottom-tabs/screen/RNSTabsScreenViewController.h
```diff
@@ -17,6 +17,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, strong, readonly, nullable) RNSBottomTabsScreenComponentView *tabScreenComponentView;
 @property (nonatomic, weak, readonly, nullable) id<RNSBottomTabsSpecialEffectsSupporting> tabsSpecialEffectsDelegate;
+@property (nonatomic, nullable) NSString *tabItemTestID;
+@property (nonatomic, nullable) NSString *tabItemAccessibilityLabel;
```
kkafar: What are these properties for here? If they are already defined on `screen`, then do we need to duplicate those here? 

I see them being set in `- [RNSBottomTabsScreenComponenView updateProps:oldProps]` however I don't see a purpose. Could you explain it please?

ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
```diff
@@ -421,6 +435,10 @@ - (void)updateProps:(const facebook::react::Props::Shared &)props
     tabItemNeedsAppearanceUpdate = YES;
   }
 
+  if (tabBarItemNeedsA11yUpdate) {
+    [_controller updateTabItemA11yProps];
+  }
```
kkafar: Don't you wanna cleanup the invalidated flag here?

kkafar: Ideally this update should be deferred & only signalised to the controller. Same as appearance update or orientation update. Can we please align that? 

I see now, that we haven't done so for scrollview edgeeffects for some reason. It should be subject of refactor in follow up PR, can we create ticket for this?

src/components/bottom-tabs/BottomTabsScreen.types.ts
kmichalikk: AFAIK this is set by [BaseViewManager](https://github.com/facebook/react-native/blob/758a3db449d770214883ba8d15da4856ca997bff/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java#L314) which TabScreenViewManager extends somehow

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHostA11yCoordinator.kt
kmichalikk: [4e91e66](https://github.com/software-mansion/react-native-screens/pull/3497/commits/4e91e66dffff92e89c504014805df934e706f306)

ios/bottom-tabs/screen/RNSTabsScreenViewController.h
kmichalikk: I don't see them being defined anywhere else? I asked if it could be implemented this way here: https://github.com/software-mansion/react-native-screens/pull/3497#discussion_r2639617943

ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
kmichalikk: This is not the same, because here we want to update a specific item, and other deferred ones update the whole tabBar. I'd need to pass some id to get the item again in TabBarController, and I'm not sure it would look good.

src/components/bottom-tabs/BottomTabsScreen.types.ts
kkafar: Okay. So it works provided that `BaseViewManager` does not rely on any state defined explicitly in `ReactViewGroup`. This is not perfect, but I guess we should be fine. 

Can we please add a note to PR description that this is the case? (that we rely on ReactAndroid here).

ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
kkafar: We can add state do the screen view controller (or just expose a property on controller & read the state from screen view) that can be read in the main loop of appearance coordinator to decide whether update is required or not. 

I think this might be the way to go.

ios/bottom-tabs/host/RNSTabBarController.mm
```diff
@@ -184,6 +185,20 @@ - (void)updateTabBarAppearance
                                              imageLoader:[self.tabsHostComponentView reactImageLoader]];
 }
 
+- (void)updateTabBarA11yIfNeeded
+{
+  for (RNSTabsScreenViewController *tabViewController in self.viewControllers) {
+    auto screenView = static_cast<RNSBottomTabsScreenComponentView *>(tabViewController.tabScreenComponentView);
```
kligarski: Why do we need a cast here? Maybe we should iterate over `UIViewController*`, static cast it to `RNSTabsScreenViewController*` and then just use `tabViewController.tabScreenComponentView`.

ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
```diff
@@ -654,6 +677,7 @@ - (void)setSystemItem:(RNSBottomTabsScreenSystemItem)systemItem
 {
   _systemItem = systemItem;
   _tabBarItemNeedsRecreation = YES;
+  _tabBarItemNeedsA11yUpdate = YES;
```
kligarski: Shouldn't we move this to `didSetProps` to match Fabric implementation?

kligarski (approved):
The code looks good, I left 2 suggestions.

ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
kmichalikk: [2a92d90](https://github.com/software-mansion/react-native-screens/pull/3497/commits/2a92d902436ccd502229e3fb189647d7079f1aa2)

ios/bottom-tabs/host/RNSTabBarController.mm
kmichalikk: [2a92d90](https://github.com/software-mansion/react-native-screens/pull/3497/commits/2a92d902436ccd502229e3fb189647d7079f1aa2)

PR #3505, opened 2025-12-22
chore(Stack): rename `maxLifecycleState` to `activityMode`

kkafar:
Rename `maxLifecycleState` to `activityMode`
Following naming from RFC-0591.

This commit also does necessary cleanup of no longer used types such as
`StackScreenLifecycleState`.

RNSConversion-Stack is introduced with a `convert` function template,
which can be possibly later reused for other components as well.

### DIFF
--- a/ios/conversion/RNSConversions-Stack.h
+++ b/ios/conversion/RNSConversions-Stack.h
@@ -1,0 +1,21 @@
+#pragma once
+
+#if RCT_NEW_ARCH_ENABLED && RNS_GAMMA_ENABLED
+
+#import <react/renderer/components/rnscreens/Props.h>
+#import "RNSStackScreenComponentView.h"
+
+namespace rnscreens::conversion {
+
+template <typename TargetType, typename InputType>
+TargetType convert(InputType) {
+  static_assert(
+      false, "[RNScreens] Missing template specialisation for demanded types!");
+}
+
+template <>
+RNSStackScreenActivityMode convert(react::RNSStackScreenActivityMode mode);
+
+}; // namespace rnscreens::conversion
+
+#endif // RCT_NEW_ARCH_ENABLED && RNS_GAMMA_ENABLED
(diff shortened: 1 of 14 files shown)


kkafar (commented):
I'll proceed here, because the changes apply to gamma (not stable) component & I've got too many changes stacked, which leads to hell in PR management.

Please do a review anyway, I'll act on it in follow up PRs.

ios/gamma/stack/screen/RNSStackScreenComponentView.h
```diff
@@ -8,19 +8,24 @@ NS_ASSUME_NONNULL_BEGIN
...
 @class RNSStackHostComponentView;
 
-typedef NS_ENUM(int, RNSStackScreenLifecycleState) {
-  RNSStackScreenLifecycleInitial = 0,
-  RNSStackScreenLifecycleDetached = 1,
-  RNSStackScreenLifecycleAttached = 2,
+typedef NS_ENUM(int, RNSStackScreenActivityMode) {
+  RNSStackScreenActivityModeDetached = 0,
+  RNSStackScreenActivityModeAttached = 1,
 };
```
kligarski, line 14: I wonder if it should be defined in the same file as view? We also have `RNSEnums.h` but I'm not sure if putting all enums in one header is the best idea.

ios/conversion/RNSConversions-Stack.h
```diff
@@ -0,0 +1,21 @@
...
+#import <react/renderer/components/rnscreens/Props.h>
+#import "RNSStackScreenComponentView.h"
+
+namespace rnscreens::conversion {
+
+template <typename TargetType, typename InputType>
+TargetType convert(InputType) {
+  static_assert(
+      false, "[RNScreens] Missing template specialisation for demanded types!");
+}
```
kligarski, line 14: I'm not sure but we might need to look at this: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2593r0.html.

kkafar: Shocking and great at the same time. Thank you, great catch. I'll handle this in follow up.

https://github.com/software-mansion/react-native-screens-labs/issues/763

PR #3508, opened 2025-12-23
feat(Android): Make commit hook condition for resetting screen size more general

kmichalikk:
Closes https://github.com/software-mansion/react-native-screens-labs/issues/598

## Description

This PR aims to generalize the condition in RNSScreenShadowNodeCommitHook to also include the screen resize, without orientation change. Let's discuss the change & decide how "general" should it be.

Even when triggering the mechanism on every screen resize, this shouldn't be a problem, since the change is still rare.

## Changes

Modified the screen size reset condition in RNSScreenShadowNodeCommitHook

## Before & After

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/4f4f562c-dafb-4881-9d27-9439d1375d34" /> | <video src="https://github.com/user-attachments/assets/d136bf0b-b87f-4ff9-866d-e5a4199be39c" /> |

Use Test2933 & try split screen, floating window.

### DIFF
--- a/common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
+++ b/common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNodeCommitHook.cpp
@@ -32,10 +32,9 @@ RootShadowNode::Unshared RNSScreenShadowNodeCommitHook::shadowTreeWillCommit(
   auto newRootProps =
       std::static_pointer_cast<const RootProps>(newRootShadowNode->getProps());
 
-  const bool wasHorizontal = isHorizontal_(*oldRootProps);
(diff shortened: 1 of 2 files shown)


kkafar (commented):
Hmm, this one is interesting and I'll revisit it. 

I'm not sure whether we want to zero the shadow node state every single time when the screen dimensions change. 

I'll look into it after new year. Thanks for the PR.

kmichalikk:
I updated the description. It's actually easy to see on the android tablet emulator.

t0maboro:
I'm observing the regression that the ScreenContentWrapper's size doesn't match Screen's size and the content is misaligned, starting from this commit.

Before:
before

After:
after

Could you check that?

PR #3518, opened 2026-01-05
chore: add reference `StackContainer` implementation

kkafar:
## Description

This PR introduces JS implementation of `StackContainer` component. It is not library component, but rather a container implementation that is a example of a possible downstream implementation.
It is required for us to have a referential container implementation to test on.

This is not a complete implementation. It is rather a starting point, providing implementation of basic operations following [RFC-0591](https://github.com/software-mansion/react-native-screens-labs/blob/main-issue-tracker/rfcs/0591-stack-operation-model.md).

Closes <https://github.com/software-mansion/react-native-screens-labs/issues/771>

## Changes

- **Rename `maxLifecycleState` to `activityMode`**
  Following naming from RFC-0591.
  
  This commit also does necessary cleanup of no longer used types such as
  `StackScreenLifecycleState`.
  
  RNSConversion-Stack is introduced with a `convert` function template,
  which can be possibly later reused for other components as well.
  
- **Make type of `children` prop of `StackHost` not key-optional**
  It does not make sense to not specify children for `StackHost`.
  
- **Reimplement StackContainer**
  The original implementation is hard for me to follow. Moreover it has
  been based upon old operation model based on `maxLifecycleState`.
  
  New operation model introduced in RFC-0591 requires changes in
  implementation of stack operations, therefore I took this opportunity
  and reimplemented all of it.
  
  The implementation is not completed yet (nor was the previous one).
  It introduces single `StackContainer` component that should be used
  to render the new stack.
  
  While implementing I noticed a couple of problems introduced by
  me in RFC-0591. First - there is no easy way to get the navigation state
  in the JS layer. This is especially troublesome when implementing `pop`
  operation, as it is not clear what screen to dismiss.
  
  The exact problem lies in fact that in current model we allow for
  rendering array of `Stack.Screen`s with some in `detached` mode.
  When already rendered `Screen` changes its mode to `attached`,
  it becomes a top screen, but it's index in the state array does not
  change. Therefore, we can not really tell which screen is on top.
  
  `Pop` operation is one problem, but another is simply dumping current
  navigation state e.g. for telemetry purposes / saving app state
  before restart or something. I'm sure there are many use cases for
  reading current navigation state.

- **Add `useRenderDebugInfo` utility hook**
  It logs component renders, mount & unmount.
  We can generalize / move it common code later.
  It seems useful.
  
- **Fix action name & move useRenderDebugInfo hook to `react-native-screens/private`**
  This hook can be reused easily for different components while debugging.
  Seems useful enough to add it to private package.
  
- **Add basic support for dismiss in JS example code**
  
- **Add basic support for dismiss in component code**
  This is rather a PoC than proper support. There are a couple of issues
  with current implementation.
  
- **Pass info on `activityMode` change from StackScreen to StackHost**
  Before this change, the information has been passed from
  `RNSStackScreenComponentView` to `StackController` via
  `StackScreenController`. This is not compliant with RFC-0753.
  
  The controllers should not be aware of activity mode changes. The
  information should be passed directly to `StackHost` and there request
  to container should be dispatched. This is exactly what this commit
  does.
  
- **Cleanup in component code**
  
- **Add navigationStateReducerWithLogging that logs actions & state**
  This is debug-oriented utility.
  
- **Prevent pop action from executing when there is at most one attached screen**
  Previously I checked only whether there is screen at all. This could
  lead to corrupted state in obvious way, where all screens were detached!
  
- **Fix import in component code**

### DIFF
--- a/apps/src/shared/gamma/containers/stack/reducer.tsx
+++ b/apps/src/shared/gamma/containers/stack/reducer.tsx
@@ -139,7 +139,7 @@ function navigationActionPopHandler(
   }
 
   const newState = [...state];
-  // NOTE: This modifes existing state, possibly impacting calculations done before new state is updated.
+  // NOTE: This modifies existing state, possibly impacting calculations done before new state is updated.
   // Consider doing deep copy of the state here.
   // EDIT: not sure really whether this is really a problem or not, since the updates are queued
   // and the original state won't be immediatelly affected.


apps/src/shared/gamma/containers/stack/reducer.tsx
```diff
@@ -0,0 +1,236 @@
...
+    const route = state[renderedRouteIndex];
+
+    if (route.activityMode === 'detached') {
+      console.info(
+        `[Stack] Route ${route.name} already rendered, attaching it`,
+      );
+      const newState = state.toSpliced(renderedRouteIndex, 1);
+      const routeCopy = { ...route };
+      routeCopy.activityMode = 'attached';
+      newState.push(routeCopy);
```
kkafar: Please note that the route is pushed to the end of the state list here, to prevent issues with stack state restoration & element inspector.

ios/gamma/stack/screen/RNSStackScreenComponentEventEmitter.mm
```diff
@@ -52,6 +52,28 @@ - (BOOL)emitOnDidDisappear
...
+- (BOOL)emitOnNativeDismiss
+{
+  if (_reactEventEmitter != nullptr) {
+    _reactEventEmitter->onNativeDismiss({});
+    return YES;
+  } else {
+    RCTLogWarn(@"[RNScreens] Skipped onNativeDismiss event emission due to nullish emitter");
+    return NO;
+  }
+}
```
kkafar, line 75: We could consider emitting only a single type of event but with boolean in payload. I think it'll be a better approach.

ios/gamma/stack/screen/RNSStackScreenController.swift
```diff
@@ -53,4 +48,17 @@ public class RNSStackScreenController: UIViewController {
...
+    super.didMove(toParent: parent)
+
+    if parent == nil {
+      if self.screen.activityMode == .detached {
+        reactEventEmitter.emitOnDismiss()
+      } else {
+        reactEventEmitter.emitOnNativeDismiss()
+      }
+    }
+  }
```
kkafar, line 63: This is absolutely not a correct implementation. 
We need a proper way to distinguish a native dismiss from JS initiated one. 

This is just PoC code, rough-and-ready for demonstrative purposes only. We can land it, cause it does its purpose at the moment, but we'll have to improve it 100%.

src/components/gamma/stack/StackHost.types.ts
```diff
@@ -1,5 +1,10 @@
-import type { ViewProps } from 'react-native';
+import React from 'react';
+import type { ReactNativeElement, ViewProps } from 'react-native';
+import { type NativeProps } from '../../../fabric/gamma/stack/StackHostNativeComponent';
 
 export type StackHostProps = {
-  children?: ViewProps['children'];
+  children: ViewProps['children'];
+  ref?: React.RefObject<
+    (React.Component<NativeProps> & ReactNativeElement) | null
```
kkafar, line 8: I made this type like that just to make things work (and it still does not work correctly). It seemed that there is something incompatible in types between RefObject to React.Component and `HostComponentType` (type of any native component). Dunno, needs to be investigated.

src/private/hooks/useRenderDebugInfo.tsx
```diff
@@ -0,0 +1,43 @@
...
+  return componentRef;
+}
+
+function logMessage(
+  componentName: string,
+  nodeHandle: number,
+  message: string,
+) {
+  console.log(`${componentName} [${nodeHandle}] ${message}`);
+}
```
kkafar, line 46: Tremendously useful thing for logging & debugging I found myself to write multiple times over and over. This time I decided to clean it a bit & add it, so that we can reuse it. 

This clearly needs a doc comment explaining what is a purpose of this hook.

kkafar: https://github.com/software-mansion/react-native-screens/pull/3518/commits/e3a9432ef270091b319470b5b04fcc69a72c96e8

apps/src/shared/gamma/containers/stack/reducer.tsx
```diff
@@ -0,0 +1,238 @@
...
+    if (route.activityMode === 'detached') {
+      console.info(
+        `[Stack] Route ${route.name} already rendered, attaching it`,
+      );
+      const newState = state.toSpliced(renderedRouteIndex, 1);
+      const routeCopy = { ...route };
+      routeCopy.activityMode = 'attached';
+      // Please note that we are pushing the route copy to the end of the array,
+      // to mitigate potential issues with element inspector and state restoration.
+      newState.push(routeCopy);
```
kmichalikk: is there a way to loose some info here that would need to be retrieved from screen below, i.e. back title?

apps/src/shared/gamma/containers/stack/StackContainer.types.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { StackScreenProps } from 'react-native-screens/experimental';
+
+export type StackRouteOptions = Omit<
+  StackScreenProps,
+  'children' | 'activityMode' | 'screenKey'
+>;
+
+export type StackRouteConfig = {
+  name: string;
+  Component: () => React.ReactNode;
```
kligarski: 1. Should prop name `Component` be capitalized? Or should it be `component`?
2. I wonder what type should be used here. For example, in `BottomTabsContainer` we use `ComponentType`. I'm not sure what the right answer is here.

src/components/gamma/stack/StackScreen.tsx
```diff
@@ -18,18 +18,25 @@ function StackScreen({
...
 }: StackScreenProps) {
-  const handleOnDidDisappear = React.useCallback(
-    (e: NativeSyntheticEvent<Record<string, never>>) => {
-      onDidDisappear?.(e);
-      onPop?.(screenKey);
-    },
-    [onDidDisappear, onPop, screenKey],
+  const onDismissWrapper = React.useCallback(() => {
+    onDismiss?.(screenKey);
+  }, [onDismiss, screenKey]);
```
kligarski: I wonder if `onDismiss`/`onNativeDismiss` callbacks should take `screenKey` as a parameter. If someone provides a callback to `Stack.Screen` component with `screenKey="A"`, then their callback must be called with `screenKey === "A"` - this seems redundant. Shouldn't passing appropriate `screenKey` be handled at the level of `StackContainer`, e.g. passing `() => onScreenDismissed(screenKey)` for `onDismissed` prop? Unless there are performance/code complexity reasons I'm missing.

src/components/gamma/stack/StackScreen.tsx
```diff
@@ -38,7 +45,9 @@ function StackScreen({
       onWillAppear={onWillAppear}
       onDidAppear={onDidAppear}
       onWillDisappear={onWillDisappear}
-      onDidDisappear={handleOnDidDisappear}>
+      onDidDisappear={onDidDisappear}
+      onDismiss={onDismissWrapper}
+      onNativeDismiss={onNativeDismissWrapper}>
```
kligarski:
```suggestion
      onDidDisappear={onDidDisappear}
      // Custom events
      onDismiss={onDismissWrapper}
      onNativeDismiss={onNativeDismissWrapper}>
```

If we want to keep comment convention from props.

src/private/types.ts
```diff
@@ -1,3 +1,5 @@
 // Just to keep this as a module. Remove this once
 // any other symbol is exported from this module again.
 export const __MODULE_KEEP__: number = 42;
```
kligarski: Is this still necessary?

ios/gamma/stack/screen/RNSStackScreenComponentEventEmitter.mm
kligarski: +1

ios/gamma/stack/host/RNSStackHostComponentView.mm
```diff
@@ -89,6 +99,13 @@ - (void)mountChildComponentView:(UIView<RCTComponentViewProtocol> *)childCompone
   childScreen.stackHost = self;
   [_reactSubviews insertObject:childScreen atIndex:index];
   _hasModifiedReactSubviewsInCurrentTransaction = true;
+
+  NSLog(
```
kligarski, line 103: We should probably use `RNSLog`.

ios/gamma/stack/screen/RNSStackScreenController.swift
```diff
@@ -53,4 +48,17 @@ public class RNSStackScreenController: UIViewController {
   public override func viewDidDisappear(_ animated: Bool) {
     reactEventEmitter.emitOnDidDisappear()
   }
+
+  public override func didMove(toParent parent: UIViewController?) {
+    print("ScreenCtrl [\(self.screen.tag)] didMoveToParent \(String(describing: parent))")
```
kligarski, line 53: In Obj-C we have RNSLog, not sure what to do here.

ios/gamma/stack/host/RNSStackHostComponentView.mm
```diff
@@ -75,6 +78,13 @@ - (nonnull RNSStackController *)stackController
   return _controller;
 }
 
+#pragma mark - Communication with StackScreen
+
+- (void)stackScreenChangedActivityMode:(nonnull RNSStackScreenComponentView *)stackScreen
+{
+  [_controller setNeedsUpdateOfChildViewControllers];
+}
```
kligarski, line 86: I'm not sure if I like the idea of communication between views. I'm in favor of separating native specific logic and `react-native`/JS specific logic but should the views communicate with each other? Wouldn't the previous `ScreenComponentView` -> `ScreenController` -> `HostController` path be more appropriate? We could maybe change the name of the callback so that it's unrelated to `react-native`/JS specific concepts (e.g. just `setNeedsUpdateOfChildViewControllers` or `invalidateScreenState` or something else)?

By the way, `RNSStackController` depends on `ReactMountingTransactionObserving`, we should probably change it as well if we want to separate JS and native container logic but that's probably for another PR.

apps/src/shared/gamma/containers/stack/reducer.tsx
```diff
@@ -0,0 +1,236 @@
...
+    // Please note that we are pushing the route copy to the end of the array,
+    // to mitigate potential issues with element inspector and state restoration.
+    newState.push(routeCopy);
+    return newState;
+  }
+
+  // 2 - Try to render new route
+  const newRouteConfig = action.ctx.routeConfigs.find(
+    route => route.name === action.routeName,
+  );
```
kligarski, line 87: In `StackContainer`, should we validate that all routes have unique route names?

apps/src/shared/gamma/containers/stack/reducer.tsx
```diff
@@ -0,0 +1,236 @@
...
+  if (routeIndex === NOT_FOUND_INDEX) {
+    console.error(
+      `[Stack] Can not perform 'pop-native' action - popped screen is not in state!`,
+    );
+    return state;
+  }
+
+  const route = state[routeIndex];
+  if (route.activityMode === 'detached') {
+    console.warn('[Stack] natively popped route has "detached" state');
```
kligarski, line 200: nitpick: sometimes we use `"` in logs and sometimes we use `'`. Would be nice to unify it.

apps/src/shared/gamma/containers/stack/reducer.tsx
```diff
@@ -0,0 +1,236 @@
...
+  const newState = state.toSpliced(routeIndex, 1);
+  return newState;
+}
+
+function navigationActionPreloadHandler(
+  state: StackState,
+  action: NavigationActionPreload,
+): StackState {
+  const routeConfig = action.ctx.routeConfigs.find(
+    routeConfig => routeConfig.name === action.routeName,
```
kligarski: nitpick: `'routeConfig' is already declared in the upper scope on line 212 column 9. (eslint @typescript-eslint/no-shadow)`

apps/src/shared/gamma/containers/stack/reducer.tsx
```diff
@@ -0,0 +1,236 @@
...
+  );
+
+  if (!routeConfig) {
+    console.error(
+      `[Stack] Can not perform 'preload' action - route config with name: ${action.routeName} not found!`,
+    );
+    return state;
+  }
+
+  return [...state, createRouteFromConfig(routeConfig)];
```
kligarski, line 222: Should it be allowed to have the same route preloaded multiple times (in `detached` state), e.g. `Abbbb`?

apps/src/shared/gamma/containers/stack/reducer.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+    console.info(
+      `[Stack] Route ${route.name} already rendered, attaching it`,
+    );
+    const newState = state.toSpliced(renderedRouteIndex, 1);
+    const routeCopy = { ...route };
+    routeCopy.activityMode = 'attached';
+    // Please note that we are pushing the route copy to the end of the array,
+    // to mitigate potential issues with element inspector and state restoration.
+    newState.push(routeCopy);
```
kmichalikk: Is there a case where we need to know some props related to our original parent prior to moving the screen on top?

apps/src/shared/gamma/containers/stack/reducer.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+  return [...state, newRoute];
+}
+
+function navigationActionPopHandler(
+  state: StackState,
+  action: NavigationActionPop,
+): StackState {
+  // FIXME: We have a problem here. We can not really determine, which route is currently at the very top!
+  // For now let's just accept routeKey as param here.
```
kmichalikk, line 105: what does it refer to? the action.routeKey? maybe we should be more descriptive here in the comment, have a more descriptive name for the field, i.e. action.topRouteKey or sth like that?

apps/src/shared/gamma/containers/stack/StackContainer.tsx
```diff
@@ -1,167 +1,66 @@
...
+  const navMethods = useStackOperationMethods(navActionDispatch, routeConfigs);
 
-type Path = {
-  name: string;
-  component: () => React.ReactNode;
-  options?: ScreenProps;
-};
+  React.useEffect(() => {
+    if (stackState.length === 0) {
+      navMethods.pushAction(routeConfigs[0].name);
```
kmichalikk: is this always guaranteed to complete before stack is shown?

apps/src/shared/gamma/containers/stack/StackContainer.tsx
```diff
@@ -1,167 +1,66 @@
...
+          pop: navMethods.popAction,
+          preload: navMethods.preloadAction,
+        };
+
+        return (
+          <Stack.Screen key={routeKey} {...options} activityMode={activityMode} screenKey={routeKey}
+            onDismiss={onScreenDismissed}
+            onNativeDismiss={onScreenNativelyDismissed}
+          >
+            <StackNavigationContext.Provider value={stackNavigationContext}>
```
kmichalikk, line 74: Hypothetically, what should happen if nested stack has only one screen & it calls pop(): should it pop the whole stack or warn that there is only one screen? In this implementation, it would probably do the latter

apps/src/shared/gamma/containers/stack/StackContainer.types.tsx
```diff
@@ -0,0 +1,81 @@
...
+  options: StackRouteOptions;
+};
+
+export type StackContainerProps = {
+  routeConfigs: StackRouteConfig[];
+};
+
+// Possibly private?
```
kmichalikk: maybe some of the types here could be

```ts
// exported publicly
type NavigationAction = { push, pop, preload };
// not exported publicly
type NavigationActionInternal = NavigationAction & { popCompleted, popNative };
```

?

ios/gamma/stack/screen/RNSStackScreenComponentEventEmitter.mm
kmichalikk: I'd say two function is *very slightly* more readable (onDismiss could be made even more descriptive, i.e. onCompleteJSDismiss() )

and if those two diverge in the future, we wouldn't need to refactor

ios/gamma/stack/screen/RNSStackScreenController.swift
kmichalikk: there is also NSLog or logger swift api, they may be more adequate here (?)

src/components/gamma/stack/StackScreen.tsx
kkafar: To be honest, I'm no longer sure I know what is the difference between events and "custom events". I've decided to remove the comments altogether.

https://github.com/software-mansion/react-native-screens/pull/3518/commits/c4ef4628b672b119a31ee79bb35104b4d42f00fa

src/components/gamma/stack/StackScreen.tsx
kkafar: Technically functions should not be defined during render - this is an anti pattern leading to the function being redefined (recreated) during every single render. They should be wrapped in appropriate React hook - likely `React.useCallback`. Now, simply to keep things simple ™️ I've avoided creation of an memoized array of memoized callbacks -> one for each screen.

apps/src/shared/gamma/containers/stack/StackContainer.types.tsx
kkafar: Ad 1. 

This is from [react legacy docs](https://legacy.reactjs.org/docs/components-and-props.html) but I believe the rule still applies:

image

kkafar: Ad 2.

I guess `ComponentType` is more appropriate here. It allows for class component to be passed, which I guess it not that big of advantage, but rather nice thing is that it correctly types Suspending components that potentially return a `Promise<React.ReactNode>`.

kkafar: https://github.com/software-mansion/react-native-screens/pull/3518/commits/8770424b5f53a224589395ab0f25865f5d642043

src/private/types.ts
kkafar: https://github.com/software-mansion/react-native-screens/pull/3518/commits/4c237e38fb05bcff5cee583341dfcb42f1e60325

apps/src/shared/gamma/containers/stack/reducer.tsx
kkafar: I think this is a task for a linter & formatter, rather than for manual work. We don't have it configured right now, so I'll defer it to separate PR.

ios/gamma/stack/screen/RNSStackScreenComponentEventEmitter.mm
kkafar: Please see: https://github.com/software-mansion/react-native-screens/pull/3518/commits/c840b04b114b67b523eeb3f0a8cb627192dadfb8

ios/gamma/stack/host/RNSStackHostComponentView.mm
kkafar: I think it is fine right now for the development phase. We can remove these logs once we're closer to releasing alpha version. 

I've created ticket to think through our logging approach in general: https://github.com/software-mansion/react-native-screens-labs/issues/775

ios/gamma/stack/screen/RNSStackScreenController.swift
kkafar: I've created ticket to think through our logging approach in general: https://github.com/software-mansion/react-native-screens-labs/issues/775

ios/gamma/stack/host/RNSStackHostComponentView.mm
kkafar: > By the way, RNSStackController depends on ReactMountingTransactionObserving, we should probably change it as well if we want to separate JS and native container logic but that's probably for another PR.

Exactly

> [...] Wouldn't the previous ScreenComponentView -> ScreenController -> HostController path be more appropriate? [...]

First - the react abstraction & activity mode shenanigans should not be leaked to controller & native container, as described in the RFC. 
Now, we can have multiple approaches, but I decided to pass this information from the screen to host to make the host aware of the incoming change. This could be abstracted in form of some delegate, so that anyone who implements a particular protocol can be notified, however it is out of scope of this PR.

apps/src/shared/gamma/containers/stack/reducer.tsx
kkafar: Yeah, sensible. Thanks.

kkafar: https://github.com/software-mansion/react-native-screens/pull/3518/commits/484c4210f186a9e3efb741fabb3d0b23c4621270

apps/src/shared/gamma/containers/stack/reducer.tsx
kkafar: https://github.com/software-mansion/react-native-screens/pull/3518/commits/043ac1254d4789110a9c9be8abb6dd6c9d4c9cba

apps/src/shared/gamma/containers/stack/reducer.tsx
kkafar: Yeah, that's completely fine. In such case pushing just takes first one available.

In "non testing" implementation it'd be worth to either limit this or allow for pushing based on particular route key, so that we can navigate to a particular one that has interesting data, but in our case this is enough IMO.

apps/src/shared/gamma/containers/stack/reducer.tsx
kkafar: I don't think there is a case, where in JS a screen needs to have information above what other screen it is rendered.

apps/src/shared/gamma/containers/stack/StackContainer.tsx
kkafar: Yes, it'll do the latter. We'll implement nesting behaviour when native side is in a bit better shape. As the PR description states: this PR brings only a base implementation, starting point for future development.

apps/src/shared/gamma/containers/stack/reducer.tsx
kkafar: This refers to the fact, that the `pop` action dismisses screen with key passed in action payload. It does not dismiss the "top screen" necessarily. 

In our model dismissing non-top screen is forbidden, therefore this code needs some more work. I'd like to defer it to another PR, however, as it potentially requires some more native changes I did not want to include in rather JS-focused PR.

apps/src/shared/gamma/containers/stack/StackContainer.tsx
kkafar: Hmm, I guess not. But it works good enough for now. I could also use state initialiser though instead of the hook 🤔

kkafar: https://github.com/software-mansion/react-native-screens/pull/3518/commits/085b39751a094df898efe79d3b4a2ab75d402a5e

apps/src/shared/gamma/containers/stack/StackContainer.types.tsx
kkafar: https://github.com/software-mansion/react-native-screens/pull/3518/commits/e4ead2056d91588586de5b8a861e811b4538d426

I came to conclusion that despite the types being "public", only certain navigation actions are exposed & that is fine. We're not able to effectively hide the types anyway, they can always be imported unless we want to define them in the modele they're used in & I don't wanna to clutter the code that much.

apps/src/shared/gamma/containers/stack/reducer.tsx
```diff
@@ -223,14 +222,22 @@ function navigationActionPreloadHandler(
   return [...state, createRouteFromConfig(routeConfig)];
 }
 
-function createRouteFromConfig(config: StackRouteConfig): StackRoute {
+export function createRouteFromConfig(config: StackRouteConfig): StackRoute {
```
kmichalikk: I think we don't need to export it

apps/src/shared/gamma/containers/stack/StackContainer.tsx
```diff
@@ -55,12 +79,23 @@ export function StackContainer({ routeConfigs }: StackContainerProps) {
...
-function requireRouteConfigs(routeConfigs?: StackRouteConfig[] | undefined | null) {
+function useSanitizeRouteConfigs(
+  routeConfigs?: StackRouteConfig[] | undefined | null,
+) {
   if (!routeConfigs || routeConfigs.length === 0) {
     throw new Error('[RNScreens] There must be at least one route configured');
   }
-}
 
+  // Do not recompute in case the routeConfigs have hot changed
```
kmichalikk: nit: not* changed

apps/src/shared/gamma/containers/stack/StackContainer.types.tsx
kligarski: Okay. We should probably match that convention in `BottomTabsContainer` then. I created a ticket to do so: https://github.com/software-mansion/react-native-screens-labs/issues/776.

ios/gamma/stack/screen/RNSStackScreenComponentEventEmitter.mm
```diff
@@ -52,6 +52,28 @@ - (BOOL)emitOnDidDisappear
   }
 }
 
+- (BOOL)emitOnDismiss
+{
+  if (_reactEventEmitter != nullptr) {
+    _reactEventEmitter->onDismiss({.isNativeDismiss = false});
+    return YES;
+  } else {
+    RCTLogWarn(@"[RNScreens] Skipped OnDismissed event emission due to nullish emitter");
```
kligarski:
```suggestion
    RCTLogWarn(@"[RNScreens] Skipped OnDismiss event emission due to nullish emitter");
```
nitpick: match name of the event from the warn message below

kligarski (approved):
Looks good!

apps/src/shared/gamma/containers/stack/reducer.tsx
kkafar: https://github.com/software-mansion/react-native-screens/pull/3518/commits/c97a767b6b10bb0943968a2fedb59b45fe91cec7

apps/src/shared/gamma/containers/stack/StackContainer.tsx
kkafar: https://github.com/software-mansion/react-native-screens/pull/3518/commits/2385e0bcd42fbaca4e68c3fcca40fada54b57474

ios/gamma/stack/screen/RNSStackScreenComponentEventEmitter.mm
kkafar: https://github.com/software-mansion/react-native-screens/pull/3518/commits/868f14b2f559b4faf8a045bbca287df97e3a2c1c

apps/src/shared/gamma/containers/stack/hooks/useStackOperationMethods.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  const popNativeAction: PopNativeActionMethod = React.useCallback(
+    (routeKey: string) => {
+      dispatch({ type: 'pop-native', routeKey, ctx: actionContext });
+    },
+    [dispatch, actionContext],
+  );
+
+  const preloadAction = React.useCallback((routeKey: string) => {
+    dispatch({ type: 'preload', routeName: routeKey, ctx: actionContext });
```
t0maboro: why in this specific case we're taking routeKey and then we're passing it to routeName, I'm just considering whether there's a chance to make dispatch call consistent with other actions without changing param name

apps/src/shared/gamma/containers/stack/hooks/useStackOperationMethods.tsx
```diff
@@ -0,0 +1,70 @@
...
+  );
+
+  const popNativeAction: PopNativeActionMethod = React.useCallback(
+    (routeKey: string) => {
+      dispatch({ type: 'pop-native', routeKey, ctx: actionContext });
+    },
+    [dispatch, actionContext],
+  );
+
+  const preloadAction = React.useCallback((routeKey: string) => {
```
t0maboro: shouldn't we add `PreloadActionMethod` type here?

apps/src/shared/gamma/containers/stack/reducer.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+  if (route.activityMode === 'detached') {
+    console.warn(
+      `[Stack] Can not perform pop action on route: ${action.routeKey} - already in detached mode!`,
+    );
+    return state;
+  }
+
+  const newState = [...state];
+  // NOTE: This modifes existing state, possibly impacting calculations done before new state is updated.
```
t0maboro:
```suggestion
  // NOTE: This modifies existing state, possibly impacting calculations done before new state is updated.
```

t0maboro (approved):

apps/src/shared/gamma/containers/stack/hooks/useStackOperationMethods.tsx
kkafar: You're right here. This is a mistake. The callback's param should be named `routeName` and not `routeKey`. 

https://github.com/software-mansion/react-native-screens/pull/3518/commits/609a140cdbb50e27fd8af33440be56354fb86263

apps/src/shared/gamma/containers/stack/hooks/useStackOperationMethods.tsx
kkafar: Thanks, added here: https://github.com/software-mansion/react-native-screens/pull/3518/commits/609a140cdbb50e27fd8af33440be56354fb86263

apps/src/shared/gamma/containers/stack/reducer.tsx
kkafar: https://github.com/software-mansion/react-native-screens/pull/3518/commits/746c8cffd7715811d4a9d1d7d54e05fb37457aad

PR #3522, opened 2026-01-07
fix(Android, FormSheet): Fix Y translation when fitToContents content overflows Screen size

t0maboro:
## Description

When `fitToContents` content is larger than the screen size, incorrect layout and translation were being applied. This PR fixes the issue by adding 0 value as a fallback when the content size exceeds the available space. 

Additionally, we force a layout pass to synchronize with the updated BottomSheet size when the top inset is applied.

| Without additional layout pass | With additional layout pass |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/3ece6fa3-162a-457d-a03d-df5fdf8c31d7" /> | <video src="https://github.com/user-attachments/assets/17ce87e8-4909-4c1e-8826-586ba2072475" /> |

This problem does not occur for fractions, since these are based on a value proportional to the `containerHeight` within the [0.0, 1.0] range.

(let me know if local changelog is still mandatory, if it is, I'll update the file, if not we should update PR template asap)

Closes https://github.com/software-mansion/react-native-screens-labs/issues/762

## Changes

- Added 0 as a fallback if the content is too large
- Added extra layout pass when bottom sheet behavior was updated

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/cab53a61-120c-4b32-9917-3f7be877e2d8" /> | <video src="https://github.com/user-attachments/assets/b41a99ba-bf60-4fcd-bb72-e4cd39066569" /> |

## Test plan

Added Test3522

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -218,7 +218,8 @@ class Screen(
         }
 
         if (isSheetFitToContents()) {
-            // Force a layout pass to synchronize BottomSheetBehavior's internal offsets with the new maxHeight.
+            // The maxHeight may have changed due to incoming top inset.
+            // Force a layout pass to sync BottomSheetBehavior's internal offsets with the new value.
             requestLayout()
         }
 


kmichalikk (approved):

kligarski (approved):

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
```diff
@@ -303,7 +303,10 @@ class SheetDelegate(
         if (screen.isSheetFitToContents()) {
             val contentHeight = screen.contentWrapper?.height ?: 0
             val offsetFromTop = containerHeight - contentHeight
-            return minOf(offsetFromTop, keyboardHeight)
+            // If the content is higher than the Screen, offsetFromTop becomes negative.
+            // In such cases, we return 0 because a negative translation would shift the Screen
+            // to the bottom, which is not intended.
+            return minOf(maxOf(0, offsetFromTop), keyboardHeight)
```
kkafar: **Big nitpick**: I'd contain this new requirement in `offsetFromTop` computation:

```kotlin
// <whole comment here>
val offsetFromTop = maxOf(containerHeight - contentHeight, 0)
```

This makes the `offsetFromTop` value "always valid".

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -217,6 +217,11 @@ class Screen(
             return
         }
 
+        if (isSheetFitToContents()) {
+            // Force a layout pass to synchronize BottomSheetBehavior's internal offsets with the new maxHeight.
+            requestLayout()
+        }
```
kkafar, line 224: Can we get behaviour comparison (recordings) with & without these lines of code? 

My intention is here to make this code possible to refactor / remove in the future. To achieve that we need to understand / observe exactly what effect it is meant to mitigate / enable.

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3522/commits/332ef52668e31fd00835636db1a3e2bad7caaa13

android/src/main/java/com/swmansion/rnscreens/Screen.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3522/commits/64ea50c890043dabcc75959b90019b1e9ceb9cf3

t0maboro: I added recordings in the PR description

kkafar: Thanks!

PR #3523, opened 2026-01-07
fix(iOS): fix issue with ill-formed program due to usage of static_assert(false)

kkafar:
## Description

Our iOS code is an ill-formed program in [C++20](<https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2593r0.html#valid-workaround>). This has been introduced in #3505. This PR aims to fix this problem.

I've followed instructions from:

1. <https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2593r0.html#valid-workaround>
2. <https://stackoverflow.com/questions/14637356/static-assert-fails-compilation-even-though-template-function-is-called-nowhere/14637534#14637534>

I've added `rnscreens::always_false` type that evaluates to `false`
for all template arguments but `rnscreens::do_not_use::DoNotUseThis`.

This "DoNotUseThis" is added to assure that there is at least one valid
template specialization that makes the `always_false` evaluate to true.
This is seemingly required to prevent compiler from just removing the
code.

The error message is now indeed triggered when there is no appropriate
type specialization:

image

The file is stored in "conversion" directory, since it concerns only
native stack for now. When migrating rest of the conversion functions
to the new "convert" pattern, we should move it there.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/763

### DIFF
--- a/ios/conversion/always_false.h
+++ b/ios/conversion/always_false.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#if defined(__cplusplus)
+
 namespace rnscreens {
 
 namespace do_not_use {
@@ -25,3 +27,5 @@ template <>
 struct always_false<do_not_use::DoNotUseThis> : std::true_type {};
 
 }; // namespace rnscreens
+
+#endif // defined(__cplusplus)


kligarski (approved):
Incredible.

kmichalikk (commented):
You might also want to see what they did in react-native repo (if I understand correctly, this works around the same problem): https://github.com/facebook/react-native/blob/f34b8702dbb3843429a76779ca0800e99d5cfa10/packages/react-native/ReactCommon/jsinspector-modern/ExecutionContext.cpp#L36

kkafar:
@kmichalikk In [this paragraph](https://swmansion.slack.com/archives/C01UHR6TUNL/p1767795254204759) it is explicitly stated 
that there must be a valid instantiation of the template possible, and this is why we have this `DoNotUse` struct. 

I do not know how the solution in RN repo realises that requirement (likely the `inline constexpr` thingy), therefore I'll stay with the current solution.

PR #3528, opened 2026-01-12
chore: Refactor library examples & tests

kmichalikk:
Closes https://github.com/software-mansion/react-native-screens-labs/issues/781

## Description

This PR is the first in series of example app refactors, as discussed in https://github.com/software-mansion/react-native-screens-labs/pull/759. Two new test categories are added: Single Feature Tests and Component Integration Tests. The former is meant to test separately each single prop exposed by the library, and the latter is meant to test interactions between multiple props / multiple components. This distinction, however, is sometimes blurry, and the membership to each group will be decided on per-case basis. Let's talk over some examples in the discussion below.

## Changes

Added Two entries to the Examples list: Single Feature Tests, Component Integration Tests.

Added a way to quickly write simple tests
- BottomTabsConfigProvider, StackConfigProvider
  - they provide context for `useStackConfig`/`useBottomTabsConfig` and `useDispatchStackConfig`/`useDispatchBottomTabsConfig` - it uses `useReducer` with `useContext` under the hood to configure single feature under test by passing a config object "diff"
- `BottomTabsAutoconfig`, `StackAutoconfig`
  - Uses the contex providers above to apply config diff to the respective component

See the usage in the examples created for this PR. tl;dr: You provide a ConfigScreen that uses dispatch() for configuration, define a simple app with XAutoconfig wrapped in XConfigProvider, and it should work.

## Test plan

Click through the new tests & check if they are working. This is a preparatory commit, so there is little to test.

## Checklist
n/a

### DIFF
--- a/apps/Example.tsx
+++ b/apps/Example.tsx
@@ -198,23 +198,19 @@ const MainScreen = ({ navigation }: MainScreenProps): React.JSX.Element => {
             disabled={!isPlatformReady(name)}
           />
         ))}
-        <ThemedText style={styles.label}>Single Feature Tests</ThemedText>
+        <ThemedText style={styles.label}>Tests</ThemedText>
         <ListItem
           key="SingleFeatureTests"
           testID="root-screen-single-feature-tests"
           title="Single Feature Tests"
           onPress={() => navigation.navigate('SingleFeatureTests')}
         />
-        <ThemedText style={styles.label}>
-          Component Integration Tests
-        </ThemedText>
         <ListItem
           key="ComponentIntegrationTests"
           testID="root-screen-component-integration-tests"
           title="Component Integration Tests"
           onPress={() => navigation.navigate('ComponentIntegrationTests')}
         />
-        <ThemedText style={styles.label}>Issue Tests</ThemedText>
         <ListItem
           key="IssueTests"
           testID="root-screen-issue-tests"


apps/src/tests/component-integration-tests/orientation/index.tsx
kkafar, line 1: We need to introduce naming convention for directories. This is a matter of preference, but as long as the module name is not a component or proper name of a test I think we should use kebab-case. 

But I need to write a internal doc for this, because if you ask me in couple of weeks I might say something different.

kkafar, line 1: We need to have consistent file naming rules: https://github.com/software-mansion/react-native-screens-labs/issues/770

Ad-hoc, for directories I suggest using kebab-case everywhere unless its a proper component name. So `Orientation` becomes `orientation` but `BottomTabsConfigProvider` stays as is.

kmichalikk:
Please help me decide here https://github.com/software-mansion/react-native-screens-labs/issues/781#issuecomment-3748579944

apps/src/tests/component-integration-tests/orientation/index.tsx
```diff
@@ -0,0 +1,10 @@
+import { Scenario } from "../../shared/helpers";
+import StackInTabs from "./StackInTabs";
+import TabsInStack from "./TabsInStack";
+
+const OrientationScenarios: Scenario[] = [
+  { name: "StackInBottomTabs", details: "Configuration in Stack contained within TabScreen always takes precedence", key: "StackInBottomTabs", screen: StackInTabs, platforms: ['ios'] },
+  { name: "BottomTabsInStack", details: "Configuration in BottomTabs contained within StackScreen should have precedence over configuraton in Stack contained within TabScreen", key: "BottomTabsInStack", screen: TabsInStack, platforms: ['ios'] },
+];
+
+export default OrientationScenarios;
```
kkafar: We need to format these files. TIL that we apparently don't have formatter set up for apps code. Our `format-js` task is out of date and does not take neither apps directory int o account nor FabricExample.

We might consider also researching tools such as `biome` to make commit checks smoother.

I'll create tickets for these:

* https://github.com/software-mansion/react-native-screens-labs/issues/785
* https://github.com/software-mansion/react-native-screens-labs/issues/786

kkafar: Please, take care of missing new lines in the file ends.

apps/src/tests/component-integration-tests/orientation/StackInTabs.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { ScrollView } from 'react-native';
+import BottomTabsConfigProvider, { BottomTabsAutoconfig, useBottomTabsConfig, useDispatchBottomTabsConfig } from '../../shared/BottomTabsConfigProvider';
+import { DummyScreen } from '../../shared/DummyScreens';
+import StackConfigProvider, { StackAutoconfig, useDispatchStackConfig, useStackConfig } from '../../shared/StackConfigProvider';
+
+function ConfigScreen() {
+  const tabsConfig = useBottomTabsConfig();
+  const tabsDispatch = useDispatchBottomTabsConfig();
+  const stackConfig = useStackConfig();
+  const stackDispatch = useDispatchStackConfig();
```
kkafar: This seems nice!

apps/src/tests/component-integration-tests/orientation/StackInTabs.tsx
```diff
@@ -0,0 +1,46 @@
+import { SettingsPicker } from '../../../shared/SettingsPicker';
+import React from 'react';
+import { ScrollView } from 'react-native';
+import BottomTabsConfigProvider, { BottomTabsAutoconfig, useBottomTabsConfig, useDispatchBottomTabsConfig } from '../../shared/BottomTabsConfigProvider';
+import { DummyScreen } from '../../shared/DummyScreens';
+import StackConfigProvider, { StackAutoconfig, useDispatchStackConfig, useStackConfig } from '../../shared/StackConfigProvider';
+
+function ConfigScreen() {
+  const tabsConfig = useBottomTabsConfig();
```
kkafar: I believe you can "destructure" the object `const { tabsConfig } = useBottomTabsConfig()` to simplify later calls.

apps/src/tests/component-integration-tests/orientation/StackInTabs.tsx
```diff
@@ -0,0 +1,46 @@
...
+      <SettingsPicker
+        label='Tab Screen orientation'
+        items={['portrait', 'landscape', 'undefined']}
+        value={tabsConfig.tabConfigs.find(c => c.tabScreenProps.tabKey === 'Tab1')?.tabScreenProps.orientation ?? 'undefined'}
+        onValueChange={value => tabsDispatch({ type: 'tabScreen', tabKey: 'Tab1', config: { tabScreenProps: { orientation: value === 'undefined' ? undefined : value } } })}
+      />
+      <SettingsPicker
+        label='Stack Screen orientation'
+        items={['portrait', 'landscape', 'undefined']}
+        value={stackConfig.find(c => c.name === 'Screen1')?.options?.orientation ?? 'undefined'}
```
kkafar: I don't understand yet at this point, why `tabsConfig` & `stackConfig` objects have different structure, so that you have to access a property on `tabsConfig` and on `stackConfig` you do not.

apps/src/tests/component-integration-tests/orientation/TabsInStack.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const tabsDispatch = useDispatchBottomTabsConfig();
+  const stackConfig = useStackConfig();
+  const stackDispatch = useDispatchStackConfig();
+
+  return (
+    <ScrollView style={{ padding: 40 }}>
+      <SettingsPicker
+        label='Stack Screen orientation'
+        items={['portrait', 'landscape', 'undefined']}
+        value={stackConfig.find(c => c.name === 'Screen1')?.options?.orientation ?? 'undefined'}
```
kkafar: I don't like this part. Dunno yet how to organise it better, but there are many magic constants, which it would be awesome to get rid of. I'll try to suggest a patch after I complete the review.

apps/src/tests/component-integration-tests/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <NavigationContainer>
+        <Stack.Navigator screenOptions={{ headerShown: false }}>
+          <Stack.Screen name="Home" component={HomeScreen} options={{ headerShown: true, headerLargeTitleEnabled: true, headerTitle: "Scenarios" }} />
+          <Stack.Screen name="Orientation">{() => <ScenariosScreen title='Orientation' scenarios={OrientationScenarios}/>}</Stack.Screen>
+          <Stack.Screen name="ScrollView">{() => <ScenariosScreen title='ScrollView' scenarios={ScrollViewScenarios}/>}</Stack.Screen>
+        </Stack.Navigator>
+      </NavigationContainer>
+    </NavigationIndependentTree>
+  );
+}
```
kkafar: This code has to be formatted before we land it 😅

apps/src/tests/component-integration-tests/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+function ScenariosScreen(props: { title: string, scenarios: Scenario[] }) {
+  const Stack = createNativeStackNavigator();
+
+  return (
+    <NavigationIndependentTree>
+      <NavigationContainer>
+        <Stack.Navigator screenOptions={{ headerShown: false }}>
+          <Stack.Screen key='home' name='home' options={{ headerShown: true, headerLargeTitleEnabled: true, headerTitle: props.title }}>{() => <ScenarioSelect scenarios={props.scenarios}/>}</Stack.Screen>
+          { props.scenarios.map(({ name, key, screen }) => <Stack.Screen name={name} key={key} component={screen as any}/>) }
```
kkafar: Why is this `screen` casted to `any` here?

apps/src/tests/component-integration-tests/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export default function App() {
+  const Stack = createNativeStackNavigator();
+
+  return (
+    <NavigationIndependentTree>
+      <NavigationContainer>
+        <Stack.Navigator screenOptions={{ headerShown: false }}>
+          <Stack.Screen name="Home" component={HomeScreen} options={{ headerShown: true, headerLargeTitleEnabled: true, headerTitle: "Scenarios" }} />
```
kkafar: Similarly, it'd be worth to extract it to an array and use `map` here to simplify adding next screens.

apps/src/tests/shared/BottomTabsConfigProvider.tsx
```diff
@@ -0,0 +1,86 @@
+import { BottomTabsContainer, BottomTabsContainerProps, TabConfiguration } from "../../shared/gamma/containers/bottom-tabs/BottomTabsContainer";
+import React, { ComponentType, createContext, Dispatch, ReactNode, useContext, useReducer } from "react";
+
+type BottomTabConfigAction = {
+  type: 'tabScreen',
+  tabKey: string,
+  config: Partial<Omit<TabConfiguration, 'tabScreenProps'> & { tabScreenProps: Partial<TabConfiguration['tabScreenProps']> }>,
+} | {
+  type: 'tabBar',
+  config: Omit<Partial<BottomTabsContainerProps>, 'tabConfigs'>
```
kkafar: I'd suggest naming it `BottomTabsConfigUpdate` maybe? I don't see how this is an "action".

apps/src/tests/shared/BottomTabsConfigProvider.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+export function BottomTabsAutoconfig() {
+  const config = useBottomTabsConfig();
+
+  return (
+    <BottomTabsContainer { ...config } />
+  )
+}
+
+export default function BottomTabsConfigProvider(props: { children: ReactNode | ReactNode[], tabs: Record<string, ComponentType> }) {
```
kkafar: Usually we use `ViewProps['children']` to type the children prop correctly.

apps/src/tests/shared/helpers.ts
```diff
@@ -0,0 +1,7 @@
+export interface Scenario {
+  name: string;
+  key: string;
+  details?: string;
+  platforms?: ('android' | 'ios')[],
```
kkafar: Potentially this should include tvOS.

Also I see this type being used in multiple places, why is it not abstracted?

apps/src/tests/single-feature-tests/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+    <NavigationIndependentTree>
+      <NavigationContainer>
+        <Stack.Navigator screenOptions={{ headerShown: false }}>
+          <Stack.Screen name="Home" component={HomeScreen} options={{ headerShown: true, headerLargeTitleEnabled: true, headerTitle: "Scenarios" }} />
+          <Stack.Screen name="BottomTabs">{() => <ScenariosScreen title='BottomTabs' scenarios={BottomTabsScenarios}/>}</Stack.Screen>
+          <Stack.Screen name="BottomTabsScreen">{() => <ScenariosScreen title='BottomTabsScreen' scenarios={BottomTabsScreenScenarios}/>}</Stack.Screen>
+          <Stack.Screen name="SplitHost">{() => <ScenariosScreen title='SplitHost' scenarios={SplitHostScenarios}/>}</Stack.Screen>
+          <Stack.Screen name="SplitScreen">{() => <ScenariosScreen title='SplitScreen' scenarios={SplitScreenScenarios}/>}</Stack.Screen>
+          <Stack.Screen name="StackHost">{() => <ScenariosScreen title='StackHost' scenarios={StackHostScenarios}/>}</Stack.Screen>
+          <Stack.Screen name="StackScreen">{() => <ScenariosScreen title='StackScreen' scenarios={StackScreenScenarios}/>}</Stack.Screen>
```
kkafar: It feels necessary to simplify addition of a screen. Currently if you want to create a new test, you must add it here, above in HomeScreen, and maybe in few more places I don't see right now. We must have it done it such way, that **at most single modification is required**. 

You can e.g. consider exporting standardized configuration (the `Scenario` you defined) from each of the test files and then only aggregate them in index.

Will try to suggest something after review.

apps/src/tests/component-integration-tests/orientation/index.tsx
kmichalikk: I installed prettier vsc extension and pointed it to the rc from our repo, let me know if it worked & looks correct now

apps/src/tests/component-integration-tests/orientation/StackInTabs.tsx
kmichalikk: I tried to make them similar wherever possible, what's left is the actual configuration that differs

apps/src/tests/component-integration-tests/orientation/TabsInStack.tsx
kmichalikk: discussed on call, I added a find method with type checks [e078546](https://github.com/software-mansion/react-native-screens/pull/3528/commits/e0785464161080c408310ef3fd35d5695c5b09dc)

apps/src/tests/component-integration-tests/index.tsx
kmichalikk: because there is some type mismatch which looks like TS takes matching definition from different places (react-navigation vs node_modules?); I removed it, and it still errors but it doesn't prevent the code from running & I don't know how to fix this elegantly. If you have any idea, let me know

apps/src/tests/component-integration-tests/index.tsx
kmichalikk: [d04594f](https://github.com/software-mansion/react-native-screens/pull/3528/commits/d04594f0a5e5388b44624fd84f11ba95159efe69)

apps/src/tests/shared/BottomTabsConfigProvider.tsx
kmichalikk: [e078546](https://github.com/software-mansion/react-native-screens/pull/3528/commits/e0785464161080c408310ef3fd35d5695c5b09dc)

apps/src/tests/shared/helpers.ts
kmichalikk: I don't understand here

apps/src/tests/single-feature-tests/index.tsx
kmichalikk: [d04594f](https://github.com/software-mansion/react-native-screens/pull/3528/commits/d04594f0a5e5388b44624fd84f11ba95159efe69)

apps/src/tests/component-integration-tests/orientation/StackInTabs.tsx
kmichalikk: I'm not sure if it's necessary, this hook shouldn't return anything else & after refactoring I'm using the returned variable in find...(tabsConfig) later

apps/src/tests/component-integration-tests/orientation/StackInTabs.tsx
```diff
@@ -0,0 +1,96 @@
...
+};
+
+type TabsParamsList = {
+  Tab1: undefined;
+  Tab2: undefined;
+};
+
+function ConfigScreen() {
+  const tabsConfig = useTabsConfig<TabsParamsList>();
+  const tabsDispatch = useDispatchTabsConfig<TabsParamsList>();
```
kkafar: This could be aggregated into a single hook that returns a tuple of objects, would be nicer to use.

kkafar: e.g. `const [tabsConfig, updateTabsConfig] = useTabsConfigState<TabsParamList>();`

kkafar (changes requested):
Looks nice, however some modules are cramped with hooks, components, utility functions, etc. 

I've created a PR with review suggestions - you can merge it as is or adapt individual changes:

* https://github.com/software-mansion/react-native-screens/pull/3546

Please apply similar refactors as I've done to stack-related code to tabs-related code.

apps/src/tests/component-integration-tests/orientation/StackInTabs.tsx
```diff
@@ -0,0 +1,97 @@
...
+type TabsParamsList = {
+  Tab1: undefined;
+  Tab2: undefined;
+};
+
+function ConfigScreen() {
+  const [tabsConfig, tabsDispatch] = useTabsConfigState<TabsParamsList>();
+
+  const stackConfig = useStackConfig<StackParamsList>();
+  const stackDispatch = useDispatchStackConfig<StackParamsList>();
```
kkafar: Let's do the same as for tabs, right? Make it into a single hook. There is no reason to have different approaches between stack & tabs here.

apps/src/tests/shared/components/stack/StackAutoconfig.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { useStackConfig } from '../../hooks/stack-config';
+import { createNativeStackNavigator } from '@react-navigation/native-stack';
+import { NavigationIndependentTree } from '@react-navigation/native';
+
+export function StackAutoconfig() {
+  const config = useStackConfig();
+  const Stack = createNativeStackNavigator();
```
kkafar: We're creating new instance of stack navigator every time this component is rendered. Is this intentional? Does is solve any issue? Wouldn't it make sense to either memoize the computation or take it as param of `StackAutoconfig` component?

What do you think?

apps/src/tests/component-integration-tests/index.tsx
kkafar: I don't observe it longer after recent refactors. Resolving the comment here.

apps/src/tests/shared/helpers.ts
kkafar: Usually, when a type, such as `('android' | 'ios')[]` is repeated in multiple places & it represents the same concept, it's worth to abstract it out to a separate type definition / alias, e.g. `type ScenarioPlatforms = 'android' | 'ios'` and later reuse it. 

After the refactors it is used only in a single place, that's fine then.

kkafar: @kmichalikk resolve this thread later please after you read it

kkafar (approved):
Looks good now. Thank you.  Which PR should I look at next?

PR #3531, opened 2026-01-13
feat(Android, Stack v5): add skeleton native implementation

kkafar:
@kligarski:

## Description

Adds skeleton code for Android implementation of Native Stack v5. Adds support for push, pop and native pop operations of single screens. 

Closes https://github.com/software-mansion/react-native-screens-labs/issues/56.

#### Class relations

Here is the diagram with current class relations:

stack_android_relations

#### Preload

Preload (in general, changing activity mode from `detached` to `attached` for screens that are not at the top of the stack) is currently not supported. It results in incorrect state in native view hierarchy. Turns out our approach to managing preloaded screens suggested in [RFC-753](https://github.com/software-mansion/react-native-screens-labs/blob/main-issue-tracker/rfcs/0753-stack-logical-architecture.md#potential-issues-and-considerations) (moving them to the end of the children list when changing activity state from detached to attached) will not work with current `react-native` architecture. 

Here is an example. Uppercase means screen in `attached` state and lowercase means `detached` state.

| JS | Native container |
| -- | --- |
| `A` | `A` |
| `Ab` | `A` |
| `AbC` | `AC` |
| `ACB` | ? |

When children in JS change from `AbC` to `ACB`, React attempts to unmount screen C from `StackHost` and remount it at index `1`. Please note that in native hierarchy, screens are attached to `StackContainer`, not `StackHost` - we're *lying* a little bit to React that they are children of `StackHost`.

When remounting the screen, there is a check if the view already has a parent. In our case, screen C is attached to `StackContainer` so the check is true. If the view has a parent, view is removed from the parent to prevent a crash.

```java
  @UiThread
  public void addViewAt(final int parentTag, final int tag, final int index) {
    // ...
      // We've hit an error case, and `addView` will crash below
      // if we don't take evasive action (it is an error to add a View
      // to the hierarchy if it already has a parent).
      // We don't know /why/ this happens yet, but it does happen
      // very infrequently in production.
      // Thus, we do three things here:
      // (1) We logged a SoftException above, so if there's a crash later
      // on, we might have some hints about what caused it.
      // (2) We remove the View from its parent.
      // (3) In case the View was removed from the hierarchy with the
      // RemoveDeleteTree instruction, and is now being readded - which
      // should be impossible - we mark this as a "readded" View and
      // thus prevent the RemoveDeleteTree worker from deleting this
      // View in the future.
      if (viewParent instanceof ViewGroup) {
        ((ViewGroup) viewParent).removeView(view);
      }
  // ...
}
```
[source](https://github.com/facebook/react-native/blob/v0.83.1/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/fabric/mounting/SurfaceMountingManager.java#L402)

This wouldn't be a problem in our case because screen would remain attached to `StackContainer` but this code prevents us from handling preload this way.

In follow-up PR, we will try to handle this case in another way (but this requires changes in JS implementation).

## Changes

- add `StackHost` and `StackScreen` + managers
- add `StackContainer` basic implementation
- add events for `StackScreen`

## Visual documentation

| Basic stack operations |
| --- |
| <video src="https://github.com/user-attachments/assets/c4996e32-37f3-4505-a951-23340b0938cf" /> |

## Test plan

Run `TestScreenStack`. Test pushing, popping and native pop via back gesture.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
@@ -45,13 +45,20 @@ internal class StackContainer(
             checkNotNull(FragmentManagerHelper.findFragmentManagerForView(this)) {
                 "[RNScreens] Nullish fragment manager - can't run container operations"
             }
+
+        // We run container update to handle any pending updates requested before container was
+        // attached to window.
+        performContainerUpdateIfNeeded()
     }
 
     /**
      * Call this function to trigger container update
      */
     internal fun performContainerUpdateIfNeeded() {
-        if (pendingOperationQueue.isNotEmpty()) {
+        // If container update is requested before container is attached to window, we ignore
+        // the call because we don't have valid fragmentManager yet.
+        // Update will be eventually executed in onAttachedToWindow().
+        if (pendingOperationQueue.isNotEmpty() && isAttachedToWindow) {
             val fragmentManager =
                 checkNotNull(fragmentManager) { "[RNScreens] Fragment manager was null during stack container update" }
             performOperations(fragmentManager, false)


android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
```diff
@@ -0,0 +1,142 @@
...
+        val transaction = fragmentManager.createTransactionWithReordering()
+        pendingOperationQueue.forEach { operation -> performOperation(fragmentManager, transaction, operation) }
+
+        // TODO: refactor + should every push be added as separate back stack entry to maintain history?
+        val lastPushScreenKey =
+            pendingOperationQueue
+                .reversed()
+                .filter { it is AddOperation }
+                .map { operation -> (operation as AddOperation).screen.screenKey }
+                .firstOrNull()
```
kligarski, line 89: I'm thinking whether each push should be a separate transaction, added to back stack so that we can later pop screens one-by-one rather that all at the same time? For now I added this workaround but it's only temporary. Also, pop operations don't use transactions at all.

android/src/main/java/com/swmansion/rnscreens/gamma/stack/screen/StackScreenFragment.kt
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    override fun onStop() {
+        stackScreen.eventEmitter.emitOnDidDisappear()
+        super.onStop()
+    }
+
+    override fun onDestroyView() {
+        stackContainer.get()?.onFragmentDestroyView(this)
+        stackScreen.eventEmitter.emitOnDismiss(stackScreen.activityMode == StackScreen.ActivityMode.ATTACHED)
```
kligarski, line 43: This is tighlty-coupling `StackScreenFragment` with `StackScreen`/"JS world" (activityMode). When we'll introduce abstraction over screens in container, we need to expose event methods including `onDismiss`. Then `stackScreen.activityMode == StackScreen.ActivityMode.ATTACHED` should be handled by `stackScreen` instead of generic fragment component.

kligarski:
@kkafar Ready for review.

android/src/main/java/com/swmansion/rnscreens/gamma/helpers/FragmentManagerHelper.kt
```diff
@@ -74,3 +75,5 @@ object FragmentManagerHelper {
         }
     }
 }
+
+fun FragmentManager.createTransactionWithReordering(): FragmentTransaction = this.beginTransaction().setReorderingAllowed(true)
```
t0maboro: can be internal

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
```diff
@@ -0,0 +1,142 @@
...
+        fragmentManager =
+            checkNotNull(FragmentManagerHelper.findFragmentManagerForView(this)) {
+                "[RNScreens] Nullish fragment manager - can't run container operations"
+            }
+    }
+
+    /**
+     * Call this function to trigger container update
+     */
+    fun performContainerUpdateIfNeeded() {
```
t0maboro: I'd suggest using `internal` wherever possible, but we can revisit it as a follow-up

android/src/main/java/com/swmansion/rnscreens/gamma/stack/screen/event/StackScreenDidDisappearEvent.kt
```diff
@@ -0,0 +1,30 @@
+package com.swmansion.rnscreens.gamma.stack.screen.event
```
t0maboro, line 1: nit: all these classes for event types (at least lifecycle event types) look similar - we may consider creating a follow-up ticket for extracting common logic to some base class

android/src/main/java/com/swmansion/rnscreens/gamma/helpers/FragmentManagerHelper.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3531/commits/a704a0a36cebf768190d2d3fb7fb6770c2f9dcc2

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3531/commits/a704a0a36cebf768190d2d3fb7fb6770c2f9dcc2

I left `StackHost`, `StackScreen` + managers and props public.

android/src/main/java/com/swmansion/rnscreens/gamma/stack/screen/event/StackScreenDidDisappearEvent.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3531/commits/343be1320e7ae56e725a003138853dec1c476863

I'm not sure if it really helps, there is still some boilerplate in each class.

kligarski:
> I'm seeing that all pushed screens are attached. Is this intentional, and will it be handled later?
> Screenshot 2026-01-26 at 14 44 40

Yes, this is intended for now but eventually we are going to change it.

android/src/main/java/com/swmansion/rnscreens/gamma/stack/screen/event/StackScreenDidDisappearEvent.kt
t0maboro: following DRY rule, we can have some base methods implemented in 1 file instead of 5, so for me it's fine, I'm wondering whether getEventName & getEventRegistrationName cannot be stored in a base class, I'll try to do that on my end

android/src/main/java/com/swmansion/rnscreens/gamma/stack/screen/event/StackScreenLifecycleEvent.kt
```diff
@@ -0,0 +1,20 @@
+package com.swmansion.rnscreens.gamma.stack.screen.event
+
+import com.facebook.react.bridge.Arguments
+import com.facebook.react.bridge.WritableMap
+import com.facebook.react.uimanager.events.Event
+import com.swmansion.rnscreens.gamma.common.NamingAwareEventType
+
+internal abstract class StackScreenLifecycleEvent<T : Event<T>>(
+    surfaceId: Int,
+    viewId: Int,
```
t0maboro, line 10: if you add `eventName`, `registrationName` to the c-tor, we may have default implementation for these two here

kligarski: https://github.com/software-mansion/react-native-screens/pull/3531/commits/edd47bcb433b858d970276eb16a790f4ae06670e

Let me know if this is what you meant.

android/src/main/java/com/swmansion/rnscreens/gamma/stack/screen/event/StackScreenDidDisappearEvent.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3531/commits/edd47bcb433b858d970276eb16a790f4ae06670e

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
```diff
@@ -0,0 +1,144 @@
...
+        fragmentManager: FragmentManager,
+        commitSync: Boolean = false,
+    ) {
+        val transaction = fragmentManager.createTransactionWithReordering()
+        pendingOperationQueue.forEach { operation -> performOperation(fragmentManager, transaction, operation) }
+
+        // TODO: refactor + should every push be added as separate back stack entry to maintain history?
+        val lastPushScreenKey =
+            pendingOperationQueue
+                .reversed()
```
kkafar: Call to `reversed` extension method effectively copies whole queue here. 
You can use `asReversed` to only create adapter object.

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
kkafar: I think we've already answered this question in a meeting. 

We need to treat each screen push as an individual transaction, otherwise we won't be able to provide correct "predictive back gesture" behavior.

Pop operations do not use explicit transactions, however they are implemented via transactions by fragment manager.

There is one thing that is currently out of the planned operation model, but this will tie our hands a bit in the future if we wanted to change that. Namely, removing a screen from the middle of the stack, will be hard to support, since removing such fragment manually will potentially desynchronise fragment manager back stack state. Just a note for the future us.

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
```diff
@@ -0,0 +1,144 @@
...
+        pendingOperationQueue.clear()
+
+        // Pop operation does not use transaction
+        if (!transaction.isEmpty) {
+            require(lastPushScreenKey != null) { "[RNScreens] Expected non-null screenKey for back stack entry." }
+
+            // don't add root to back stack to handle exiting from app.
+            if (fragmentManager.fragments.isNotEmpty()) {
+                transaction.addToBackStack(lastPushScreenKey)
+            }
```
kkafar, line 100: This obviously does not make sense. It should be done inside `performAddOperation`.

I don't think I want you fixing it here. Let's land this PR with the PoC and improve in followups.

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
```diff
@@ -0,0 +1,144 @@
...
+
+        val lastBackStackEntry = fragmentManager.getBackStackEntryAt(backStackEntryCount - 1)
+        require(lastBackStackEntry.name == operation.screen.screenKey) { "[RNScreens] Popping is supported only for top screen." }
+
+        fragmentManager.popBackStack(lastBackStackEntry.name, FragmentManager.POP_BACK_STACK_INCLUSIVE)
+    }
+
+    internal fun onFragmentDestroyView(fragment: StackScreenFragment) {
+        delegate.get()?.onDismiss(fragment.stackScreen)
+    }
```
kkafar, line 146: I think it might make more sense to add fragment attach info listener to fragment manager & dispatch the events there. 

Let's do it in future PRs

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackHost.kt
```diff
@@ -0,0 +1,121 @@
...
+    UIManagerListener,
+    StackContainerDelegate {
+    internal val renderedScreens: ArrayList<StackScreen> = arrayListOf()
+    private val container = StackContainer(reactContext, WeakReference(this))
+
+    init {
+        addView(container)
+
+        // We're adding ourselves during a batch, therefore we expect to receive its finalization callbacks
+        UIManagerHelper.getUIManager(reactContext, UIManagerType.FABRIC)?.addUIManagerEventListener(this)
```
kkafar: We should require the UI manager here. In case this fails we'll never update the container, so it does not make sense to ignore the issue. The app would end up being dysfunctional anyway.

android/src/main/java/com/swmansion/rnscreens/gamma/stack/screen/StackScreen.kt
```diff
@@ -0,0 +1,56 @@
...
+        set(value) {
+            require(value) {
+                "[RNScreens] Natively dismissed StackScreen must remain dismissed."
+            }
+            field = true
+        }
+    internal var stackHost: WeakReference<StackHost?> = WeakReference(null)
+
+    var activityMode: ActivityMode by Delegates.observable(ActivityMode.DETACHED) { _, _, _ ->
+        stackHost.get()?.stackScreenChangedActivityMode(this)
```
kkafar: IIRC `observable` is called even if the value has not effectively changed. We do not want to trigger this callback in such case.

android/src/main/java/com/swmansion/rnscreens/gamma/stack/screen/StackScreen.kt
```diff
@@ -0,0 +1,56 @@
...
+        eventEmitter = StackScreenEventEmitter(reactContext, id)
+    }
+
+    override fun onLayout(
+        changed: Boolean,
+        l: Int,
+        t: Int,
+        r: Int,
+        b: Int,
+    ) {
```
kkafar: Nitpick: The convention is to make noop methods just `onLayout(...) = Unit` in Kotlin. We use it in other files.

android/src/main/java/com/swmansion/rnscreens/gamma/stack/screen/StackScreenFragment.kt
kkafar: yeah, I don't like this one either. 

I think that usage of ["lifecycle aware components"](https://developer.android.com/topic/libraries/architecture/lifecycle) might provide us with a elegant solution here. We can talk this through in the office.

android/src/main/java/com/swmansion/rnscreens/gamma/stack/screen/StackScreenViewManager.kt
```diff
@@ -0,0 +1,52 @@
...
+            "detached" -> view.activityMode = StackScreen.ActivityMode.DETACHED
+            else -> throw JSApplicationIllegalArgumentException("[RNScreens] Invalid activity mode: $value.")
+        }
+    }
+
+    override fun setScreenKey(
+        view: StackScreen,
+        value: String?,
+    ) {
+        view.screenKey = value
```
kkafar, line 49: We don't allow it to be a nullable string, right? Might make sense to verify it here (by calling require)

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3531/commits/2ef9c89eeefc86fdfd95062a2074bc1416e04774

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackHost.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3531/commits/2ef9c89eeefc86fdfd95062a2074bc1416e04774

android/src/main/java/com/swmansion/rnscreens/gamma/stack/screen/StackScreen.kt
kligarski: You're right, I got baited once again by the docs.

https://github.com/software-mansion/react-native-screens/pull/3531/commits/2ef9c89eeefc86fdfd95062a2074bc1416e04774

android/src/main/java/com/swmansion/rnscreens/gamma/stack/screen/StackScreen.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3531/commits/2ef9c89eeefc86fdfd95062a2074bc1416e04774

android/src/main/java/com/swmansion/rnscreens/gamma/stack/screen/StackScreenViewManager.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3531/commits/2ef9c89eeefc86fdfd95062a2074bc1416e04774

Also, we want to keep the logic in the view itself that prevents changing `screenKey` after setting it, right?

https://github.com/software-mansion/react-native-screens/pull/3531/changes#diff-d8c980d4ae1181feeaed7874a3e6530f8d650fdf2028af02cde5a4e357bb235fR34-R40

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
```diff
@@ -0,0 +1,151 @@
...
+    internal fun performContainerUpdateIfNeeded() {
+        // If container update is requested before container is attached to window, we ignore
+        // the call because we don't have valid fragmentManager yet.
+        // Update will be eventually executed in onAttachedToWindow().
+        if (pendingOperationQueue.isNotEmpty() && isAttachedToWindow) {
+            val fragmentManager =
+                checkNotNull(fragmentManager) { "[RNScreens] Fragment manager was null during stack container update" }
+            performOperations(fragmentManager, false)
+        }
+    }
```
kligarski, line 66: > When I navigate from Example main screen to TestScreenStack the example app crashes.

Container updates from StackHost's `didMountItems` arrive before container is attached to window. I added this logic to ignore the update and perform updates in `onAttachedToWindow`.

Let me know what you think. I have 2 thoughts here:
- in theory, someone might want to delay performing updates but `onAttachedToWindow` will flush updates early - should I add a flag here to perform `performContainerUpdateIfNeeded` in `onAttachedToWindow` only if `performContainerUpdateIfNeeded` was called before container was attached to window?
- I also considered moving this logic to `StackHost` but I decided that this is closely related to the timing of acquiring `fragmentManager` reference which is handled inside `StackContainer` so the *delay code* should be added in `StackContainer` as well.

---

Now, after those changes you can enter the `TestScreenStack` via our regular Example app but when you try to perform a native pop, you will get an error. There are 2 things that we need to change but I don't want to do it in this PR:

1. We need to handle popping with only root screen present (no entries in back stack) -> this has been already handled in follow-up PR: https://github.com/software-mansion/react-native-screens/pull/3576/changes#diff-4ee668f64f7d365ed1b6aee293bf6227fed0a6bfc91163f85ee895a86d8e1e5aR99-R104 so I don't want to add unnecessary changes here.
2. It turns out that back gesture doesn't work out of the box in nested fragment managers. In Stack v4, this is all handled by JS (https://github.com/react-navigation/react-navigation/blob/2b7ee0d2d1b1ef0390b690f59c0c1dd8bd57e9eb/packages/native/src/useBackButton.native.tsx#L21) so we didn't need to worry about it. In order to handle this on the native side, we will need to use [`setPrimaryNavigationFragment(fragment)`](https://developer.android.com/reference/kotlin/androidx/fragment/app/FragmentTransaction#setPrimaryNavigationFragment(androidx.fragment.app.Fragment)) for both Stack v4 and Stack v5. More information about it is [here](https://developer.android.com/guide/fragments/fragmentmanager). `onBackPressedCallback` is created by the fragment manager that handles our screens but it remains disabled when nested inside other fragment because it requires not only having not-empty back stack but also for the parent to be "primary navigation fragment". Relevant code snippet from `FragmentManager.java`:
```java
    private void updateOnBackPressedCallbackEnabled() {
        // Always enable the callback if we have pending actions
        // as we don't know if they'll change the back stack entry count.
        // See handleOnBackPressed() for more explanation
        synchronized (mPendingActions) {
            if (!mPendingActions.isEmpty()) {
                mOnBackPressedCallback.setEnabled(true);
                return;
            }
        }
        // This FragmentManager needs to have a back stack for this to be enabled
        // And the parent fragment, if it exists, needs to be the primary navigation
        // fragment.
        mOnBackPressedCallback.setEnabled(getBackStackEntryCount() > 0
                && isPrimaryNavigation(mParent));
    }

    /**
     * Recursively check up the FragmentManager hierarchy of primary
     * navigation Fragments to ensure that all of the parent Fragments are the
     * primary navigation Fragment for their associated FragmentManager
     */
    boolean isPrimaryNavigation(@Nullable Fragment parent) {
        // If the parent is null, then we're at the root host
        // and we're always the primary navigation
        if (parent == null) {
            return true;
        }
        FragmentManager parentFragmentManager = parent.mFragmentManager;
        Fragment primaryNavigationFragment = parentFragmentManager
                .getPrimaryNavigationFragment();
        // The parent Fragment needs to be the primary navigation Fragment
        // and, if it has a parent itself, that parent also needs to be
        // the primary navigation fragment, recursively up the stack
        return parent.equals(primaryNavigationFragment)
                && isPrimaryNavigation(parentFragmentManager.mParent);
    }
```

kligarski: Here is a quick patch to verify this:

<details>
<summary>Patch</summary>

```
diff --git a/android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt b/android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
index 2bf18bca7..87ed7df24 100644
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
@@ -279,6 +279,7 @@ class ScreenStack(
             }
 
             topScreenWrapper = newTop as? ScreenStackFragmentWrapper
+            transaction.setPrimaryNavigationFragment(newTop?.fragment)
             stack.clear()
             stack.addAll(screenWrappers.asSequence().map { it as ScreenStackFragmentWrapper })
 
diff --git a/android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt b/android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
index 5259e8f2e..1ca7eb5d0 100644
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
@@ -133,7 +133,9 @@ internal class StackContainer(
         operation: PopOperation,
     ) {
         val backStackEntryCount = fragmentManager.backStackEntryCount
-        require(backStackEntryCount > 0) { "[RNScreens] Back stack must not be empty." }
+        if (backStackEntryCount == 0) {
+            return
+        }
 
         val lastBackStackEntry = fragmentManager.getBackStackEntryAt(backStackEntryCount - 1)
         require(lastBackStackEntry.name == operation.screen.screenKey) { "[RNScreens] Popping is supported only for top screen." }
```

</details>

With this patch, the navigation should work correctly (Stack v5 nested inside Stack v4).

kligarski:
@kkafar Ready for review.

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
kkafar: I see. I don't think we need to patch stack v4, only for sake of testing in the Example app. As long as we have Stack v5 working inside stack v5, then I think we're fine for now.

kkafar: > I also considered moving this logic to StackHost but I decided that this is closely related to the timing of acquiring fragmentManager reference which is handled inside StackContainer so the delay code should be added in StackContainer as well.

This sounds sound to me.

kkafar: > in theory, someone might want to delay performing updates but onAttachedToWindow will flush updates early - should I add a flag here to perform performContainerUpdateIfNeeded in onAttachedToWindow only if performContainerUpdateIfNeeded was called before container was attached to window?

No. Let's avoid complicating things here. It'd be ideal if we didn't have to give any guarantees from `StackContainer` perspective regarding operation timing. Just call id directly in `onAttachedToWindow` as you do right now.

android/src/main/java/com/swmansion/rnscreens/gamma/stack/screen/StackScreenViewManager.kt
kkafar: Yes, leave it. I'd consider making it a lateinit var to remove nullability from the type, but that's a nitpick. Let's just proceed.

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
kligarski: > I see. I don't think we need to patch stack v4, only for sake of testing in the Example app. As long as we have Stack v5 working inside stack v5, then I think we're fine for now.

Nesting Stack v5 inside Stack v5 will not work for now. As discussed internally, I created a ticket to handle this for Stack v5: https://github.com/software-mansion/react-native-screens-labs/issues/821.

Issue #3372, opened 2025-11-05
[iOS] FullWindowOverlay is always displayed on the most recently created window


atlj:
### Description

On a multi-window app, when I mount `FullWindowOverlay` component, it's always displayed on the most recently created window, not the window I originally tried to mount it from.

https://github.com/user-attachments/assets/1bc80972-f005-4f96-8454-7cdaa02adbff

### Steps to reproduce

1. Create a multi-window app
2. Open the app on an iPad (or simulator)
3. Open a second window for the app
4. Mount a `FullWindowOverlay`
5. `FullWindowOverlay` is always displayed on the most recently created window

### Snack or a link to a repository

https://github.com/atlj/rns-fullwindowoverlay-window-bug

### Screens version

4.16.0

### React Native version

0.82.1

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

Real device

### Device model

iPad Air

### Acknowledgements

Yes

kkafar:
Looks sensible. Only thing I'm afraid of is whether we'll be able to detect appropriate window correctly or not. Currently IIRC we use "key window" mechanism, which apparently fails. 

Thanks for the report, someone will look into this.

minhchienwikipedia:
I got an issue when trying to enable Element Inspector, the component `FullWindowOverlay` is always on top, so I can't use that feature. How can we fix this issue? Anyone have solution for it

kkafar:
@minhchienwikipedia please report an issue if you think you've found one, as it seems unrelated to this particular ticket.

PR #3536, opened 2026-01-14
fix(iOS, FullWindowOverlay): FullWindowOverlay shown behind fullScreenModal 

### LINKED ISSUE #3521: iOS: FullWindowOverlay shown behind fullScreenModal of RN
### Description

I need to display content in a `FullWindowOverlay` above the navigation stack. Starting from version 4.19, if a RN `presentation: 'fullScreenModal'` is open, the `FullWindowOverlay` content opens beneath that modal. No problem with `presentation: 'modal'` and no problem until version 4.18. I suspect this misalignment appeared after [this fix](https://github.com/software-mansion/react-native-screens/pull/3379).
Here I show the different behavior between `modal` and `fullScreenModal`.

https://github.com/user-attachments/assets/2e9d1768-a9e7-48c4-b135-1b80f591415e

https://github.com/user-attachments/assets/a726fa78-cc87-436e-9b6d-bc189b81862c

### Steps to reproduce

Here's a minimal example code that reproduces the issue.

```
import { Button, StyleSheet, View } from 'react-native';
import { FullWindowOverlay } from 'react-native-screens';

  return (
      <Button
        title="Go to RNScreens modal"
        onPress={() => navigation.navigate('Modal')}
      />
    </View>
  );
}

function ModalScreen({ navigation, route }) {
  const setIsVisible = route.params.setIsVisible;
  return (
      <Button
        title="Dismiss RNScreens modal"
        onPress={() => navigation.goBack()}
      />
      <Button title="Show overlay" onPress={() => setIsVisible(true)} />
    </View>
  );
}

const NativeStack = createNativeStackNavigator();

  const [isVisible, setIsVisible] = React.useState(false);
  return (
    <>
      <NavigationContainer>
        <NativeStack.Navigator>
          <NativeStack.Screen name="Home" component={Home} />
          <NativeStack.Screen
            name="Modal"
            component={ModalScreen}
            options={{ presentation: 'fullScreenModal' }}
            initialParams={{ setIsVisible }}
          />
        </NativeStack.Navigator>
      </NavigationContainer>
      {isVisible && (
        <FullWindowOverlay>
          <View style={styles.overlay} pointerEvents="box-none">
            <View style={styles.box} />
            <Button title="Close me" onPress={() => setIsVisible(false)} />
          </View>
        </FullWindowOverlay>
      )}
    </>
  );
}

  container: {
    flex: 1,
    alignItems: 'flex-start',
    justifyContent: 'center',
  },
  overlay: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: 'rgba(0,0,0,0.5)',
  },
  box: {
    width: 40,
    height: 40,
    borderRadius: 10,
    borderWidth: 2,
    borderColor: 'black',
    backgroundColor: 'red',
  },
});
```

Example above, inspired by your test https://github.com/software-mansion/react-native-screens/blob/main/apps/src/tests/Test1096.tsx

### Screens version

4.19.0

### React Native version

0.83.1

### Platforms

iOS

### JavaScript runtime

None

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

SvenLinderoth:
Fixes #3521

**Description**

The solution restores the original z-index behavior by unconditionally calling addSubview when showing the overlay. In UIKit, this forces the view to the front of the visual stack, ensuring it covers the modal.

Crucially, this change maintains the fix for Multi-Window applications introduced in #3372. We continue to rely on didMoveToWindow to resolve the correct UIWindowScene context, ensuring the overlay attaches to the correct window in split-screen or multi-scene environments.

**Changes**

- Restored Z-Index Promotion: Removed the containsObject check. This allows addSubview to move the overlay to the top of the window stack if it is already attached.

- Preserved Context Logic: Kept the logic prioritizing self.window to ensure Multi-Window stability.

- Added Safety Fallback: Added a fallback to RCTKeyWindow in case self.window is nil, improving robustness for edge cases.

**Before**

https://github.com/user-attachments/assets/bfa34ec8-32a4-48bb-ad83-757754e5714a

**After**

https://github.com/user-attachments/assets/163799b4-4927-4c74-a41a-203c929f8e0f

**Steps to reproduce**
1. Open FabricExample.
2. Trigger FullWindowOverlay in Test3380.

I have tried to verify that this solution wont break the fix recently added for multiple screen applications by enabling multiple screens in FabricExample, and tested the test3379 before #3372 was introduced and after this solution to confirm that solution keeps working after this change.
**Before**

https://github.com/user-attachments/assets/2d5aa8d7-b417-4a7e-9864-9a9e35152f91

**After**

https://github.com/user-attachments/assets/e98f776d-d14b-4efb-9326-34950796088a

**Checklist**
- [x]  Included code example that can be used to test this change

### DIFF
--- a/ios/RNSFullWindowOverlay.mm
+++ b/ios/RNSFullWindowOverlay.mm
@@ -166,22 +166,29 @@
 
 - (void)didMoveToWindow
 {
-  // Detaching FullWindowOverlay is handled by `didMoveToSuperview`
-  if (self.window != nil) {
-    [self maybeShow];
-  }
+    [super didMoveToWindow];
+    [self show];
 }
 
-- (void)maybeShow
-{
-  UIWindow *window = [self window];
-  if (![[window subviews] containsObject:self]) {
-    [window addSubview:_container];
-  }
+- (void)show
+{    UIWindow *targetWindow = self.window;
+    
+    if (!targetWindow) {
+        targetWindow = RCTKeyWindow();
+    }
+    
+    if (targetWindow && _container) {
+        [targetWindow addSubview:_container];
+    }
 }
 
 #ifdef RCT_NEW_ARCH_ENABLED
 #pragma mark - Fabric Specific
+
+- (void)maybeShow
+{
+    [self show];
+}
 
 + (react::ComponentDescriptorProvider)componentDescriptorProvider
 {
@@ -200,6 +207,7 @@
 
 - (void)mountChildComponentView:(UIView<RCTComponentViewProtocol> *)childComponentView index:(NSInteger)index
 {
+  [self maybeShow];
   [self addSubview:childComponentView];
 }
 
--- a/apps/src/tests/Test3380.tsx
+++ b/apps/src/tests/Test3380.tsx
@@ -1,0 +1,157 @@
+import React, { useState } from 'react';
+import { Button, StyleSheet, Text, View } from 'react-native';
+import { FullWindowOverlay } from 'react-native-screens';
+import { NavigationContainer } from '@react-navigation/native';
+import { createNativeStackNavigator } from '@react-navigation/native-stack';
+
+/**
+ * REPRODUCTION: "Sibling Structure"
+ * Root
+ * ├── NavigationContainer (contains the Modal)
+ * └── FullWindowOverlay (Sibling, outside the Modal)
+ */
+
+const Stack = createNativeStackNavigator();
+
+// 1. The Screen Inside the Modal
+//    It has a button to trigger the "Global" Overlay
+function ModalScreen({ onTriggerOverlay }: { onTriggerOverlay: () => void }) {
+  return (
+    <View style={styles.modalContainer}>
+      <Text style={styles.modalText}>I am a Native Stack Modal</Text>
+      <Text style={styles.modalSubText}>
+        I am inside the NavigationContainer.
+        The Overlay is outside.
+      </Text>
+      
+      <View style={styles.spacer} />
+      
+      <Button 
+        title="FIRE SIBLING OVERLAY" 
+        color="red"
+        onPress={onTriggerOverlay} 
+      />
+    </View>
+  );
+}
+
+// 2. The Home Screen (just to open the modal)
+function HomeScreen({ navigation }: any) {
+  return (
+    <View style={styles.container}>
+      <Text style={styles.header}>Test 3380: Sibling Structure</Text>
+      <Button title="Open Full Screen Modal" onPress={() => navigation.navigate('MyModal')} />
+    </View>
+  );
+}
+
+// 3. The Root Component (The Mini-App)
+export default function Test3380() {
+  const [overlayVisible, setOverlayVisible] = useState(false);
+
+  return (
+    <View style={{ flex: 1 }}>
+      {/* PART A: The Navigation Stack (Where the Modal lives) */}
+      <NavigationContainer independent={true}>
+        <Stack.Navigator>
+          <Stack.Screen name="Home" component={HomeScreen} />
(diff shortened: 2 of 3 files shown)


kkafar:
Hey, thanks for the PR. I'll review it shortly

ios/RNSFullWindowOverlay.mm
```diff
@@ -166,23 +166,30 @@ - (void)didMoveToSuperview
 
 - (void)didMoveToWindow
 {
-  // Detaching FullWindowOverlay is handled by `didMoveToSuperview`
-  if (self.window != nil) {
-    [self maybeShow];
-  }
+    [super didMoveToWindow];
+    [self show];
```
kkafar, line 170: I don't understand this change. If the `FullWindowOverlay` is detached from `UIWindow` instance it should not be shown anymore. 

Is there any use use case that this change covers? This change does not make sense to me as is. I need additional context.

ios/RNSFullWindowOverlay.mm
```diff
@@ -166,23 +166,30 @@ - (void)didMoveToSuperview
...
-  UIWindow *window = [self window];
-  if (![[window subviews] containsObject:self]) {
-    [window addSubview:_container];
-  }
+- (void)show
+{    UIWindow *targetWindow = self.window;
+    
+    if (!targetWindow) {
+        targetWindow = RCTKeyWindow();
+    }
```
kkafar, line 178: Similarly here. This can lead only to FullWindowOverlay being displayed on wrong window, or shown despite the fact that it is detached from a window. 

Seems incorrect & I won't accept this change without additional context. The explanation given in PR description is rather generic and does not point out any particular use cases.

ios/RNSFullWindowOverlay.mm
```diff
@@ -166,23 +166,30 @@ - (void)didMoveToSuperview
...
 }
 
 #ifdef RCT_NEW_ARCH_ENABLED
 #pragma mark - Fabric Specific
 
+- (void)maybeShow
+{
+    [self show];
+}
```
kkafar, line 191: Having this method is now redundant. We should have only `maybeShow` method & deny attaching `FullWindowOverlay` if it is not attached to any window.

ios/RNSFullWindowOverlay.mm
```diff
@@ -166,23 +166,30 @@ - (void)didMoveToSuperview
...
+- (void)show
+{    UIWindow *targetWindow = self.window;
+    
+    if (!targetWindow) {
+        targetWindow = RCTKeyWindow();
+    }
+    
+    if (targetWindow && _container) {
+        [targetWindow addSubview:_container];
+    }
```
kkafar, line 182: This makes sense. Probably omitting check whether we're already added solves the issue you intend to solve. 

I think we should land only this part of the changes.

kkafar (changes requested):
Thanks for the PR!

I haven't checked the runtime yet. I'll try to do it soon. 
Some parts of the code do not make much sense to me. I recommend removing them and leaving only the part that actually fixes the problem or giving an adequate explanation.

In case I test the runtime before you modify the PR, I'll update the PR and land a fix. Thanks again!

kkafar:
I've debugged it for few moments now & I have better context. Thanks for the solution draft. I'll proceed with it in https://github.com/software-mansion/react-native-screens/pull/3553.

SvenLinderoth:
> I've debugged it for few moments now & I have better context. Thanks for the solution draft. I'll proceed with it in #3536.

Thanks for the quick follow-up as well as thorough review. Great work!

PR #3552, opened 2026-01-20
chore: Disable prebuilt RN core for FabricExample

kmichalikk:
## Description

This PR overrides the new default for downloading prebuild binaries for RN core since 0.84. We need the source code for debugging in our Example app.

Relevant commit from RN: https://github.com/facebook/react-native/pull/54895

## Changes

Env vars: RCT_USE_PREBUILT_RNCORE is set to '0' in the Podfile.

From READMEs and comments:

> ## - RCT_USE_PREBUILT_RNCORE: If set to 1, it will use the release tarball from Maven instead of building from source.

## Test plan

Make sure FabricExample builds & the source code is available in XCode.

## Checklist

n/a
- [x] Ensured that CI passes

### DIFF
--- a/FabricExample/ios/Podfile
+++ b/FabricExample/ios/Podfile
@@ -15,7 +15,6 @@ prepare_react_native_project!
 ENV['RCT_NEW_ARCH_ENABLED'] = '1'
 
 # Disable precompiled binaries for RN core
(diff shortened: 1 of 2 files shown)


kmichalikk:
I can see these are the libs you mentioned, so I'm not disabling the env

PR #3553, opened 2026-01-20
fix(iOS): FullWindowOverlay is not mounted when fullScreenModal is presented

kkafar:
## Description

Supersedes https://github.com/software-mansion/react-native-screens/pull/3536
Closes https://github.com/software-mansion/react-native-screens/issues/3521

When a screen with `presentation: 'fullScreenModal'` is presented, the native hierarchy looks as follows:

image

The original hierarchy from under "primary" `UITransitionView` is detached, including react surface view & other rendered views.
Separate `UITransitionView` is created and attached at higher index under `UIWindow` presenting `fullScreenModal` contents.

https://github.com/software-mansion/react-native-screens/pull/3379 changed how the `FullWindowOverlay` is mounted -> the mounting code is now triggered only from `didMoveToWindow` callback, which as demonstrated on above screenshot won't be triggered, because whole react-native hierarchy is detached from native view hierarchy. Hence, the FullWindowOverlay is not attached. It ends up attached only after the modal is closed, because only then the `didMoveToWindow` callback is triggered.

See the video:

https://github.com/user-attachments/assets/0c720cc3-3764-47bc-abee-3d339f10a451

The above situation presents another challange. Since the `FullWindowOverlay` component is not attached to window, neither `didMoveToWindow` callback is called nor `self.window` contains non-null value. This leads to situation where https://github.com/software-mansion/react-native-screens/issues/3372 might happen again if we decide to partially restore previous code, which used `RCTKeyWindow()`.

This PR proposes usage of `RCTKeyWindow()` as a fallback, when `FullWindowOverlay` is attached to a **super view** and `self.window == nil`. This might reintroduce #3372 in some corner cases, however it should fix the more serious issue, where the `FullWindowOverlay` does not appear at all. 

## Changes

`FullWindowOverlay` calls now `maybeShow` from `didMoveToSuperview`, which crucially *is called after `didMoveToWindow` by UIKit*, therefore if presenting the `FullWindowOverlayContainer` fails in the `didMoveToWindow` method, then we present it from `didMoveToSuperview`. 

`maybeShow` fallbacks to value returned by `RCTKeyWindow()` when `self.window` is `nil`.

This PR also fixes reentry test in `maybeShow` method. Previously it tested whether `window.subviews containsObject:self]`, which does not make much sense, since we do insert `container` to window subviews, not `self`, therefore this always returned `false` and therefore whole condition was always `true`, effectively never preventing reentry. 

## Recording

Below recording presents the behaviour after the changes.

https://github.com/user-attachments/assets/afc4c4fc-b0d6-4e77-a613-c4766ebfb9ec

## Test plan

I've added `Test3321`, demonstrated on the recordings.

1. Open the full screen modal.
2. Open the full window overlay.
3. It should be opened & visible.

I've also tested this change on test covering #3372 - it works fine, see the video below.

https://github.com/user-attachments/assets/8e265fa3-0de1-454f-9080-94a1bc760716

✅ I've also tested legacy architecture - seems to work fine. 

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSFullWindowOverlay.mm
+++ b/ios/RNSFullWindowOverlay.mm
@@ -183,6 +183,8 @@
   UIWindow *window = [self window];
 
   if (window == nil) {
+    // This fallback might return wrong window in case of multi-window
+    // apps e.g. on iPad.
     window = RCTKeyWindow();
   }
 


kkafar:
When testing for #3372 regression I've noticed that the FullWindowOverlay component does not resize on iPad, when dimensions of the active scene / window change. 

https://github.com/user-attachments/assets/70502d11-e641-4487-ab1c-1a5fad324d7a

This is rather a weird bug and must be investigated later.

ios/RNSFullWindowOverlay.mm
```diff
@@ -175,7 +181,12 @@ - (void)didMoveToWindow
 - (void)maybeShow
 {
   UIWindow *window = [self window];
-  if (![[window subviews] containsObject:self]) {
+
+  if (window == nil) {
+    window = RCTKeyWindow();
+  }
```
kligarski, line 189: Maybe we should add a comment explaining that this is a fallback and that this might be incorrect for some edge cases.

ios/RNSFullWindowOverlay.mm
```diff
@@ -175,7 +181,12 @@ - (void)didMoveToWindow
 - (void)maybeShow
 {
   UIWindow *window = [self window];
-  if (![[window subviews] containsObject:self]) {
+
+  if (window == nil) {
+    window = RCTKeyWindow();
```
t0maboro, line 188: I'd be very cautious about using `RCTKeyWindow` here, even as a fallback. Its implementation takes the first scene, where `activationState` is set to `UISceneActivationStateForegroundActive`. In a multi-window environment, it won't work well, as many scenes might be in the foreground. I'm not sure whether this should undergo further investigation or be reported to the core team, because the logic behind `RCTKeyWindow` doesn't seem solid enough to cover the multi-window case reliably.

https://github.com/user-attachments/assets/6b69a0c3-8feb-41a7-b6c0-ac0ed41f4e3b

t0maboro (commented):
opening one discussion about RCTKeyWindow, other changes look great, so I'll approve after we resolve this one thread

ios/RNSFullWindowOverlay.mm
kkafar: That's right. I've discussed this topic already in PR description. I can't link to exact section, hence the screenshot: 
image

This PR does not introduce a regression. The FWO with `fullScreenModal` does not work at all. Right now it will work in standard scenarios and still fail in multi-window one. 

The multi-window scenario should work fine with any presentation that does not remove the react native view hierarchy from host tree.

kkafar: Here's ticket for this one: https://github.com/software-mansion/react-native-screens-labs/issues/798

kkafar:
@kligarski, yeah, I've verified legacy architecture. I'll update the PR description, thanks.

ios/RNSFullWindowOverlay.mm
kkafar: Yeah, I guess it won't hurt. 

https://github.com/software-mansion/react-native-screens/pull/3553/commits/0a19c4815fb6cb36285b7a391419e03a9f83a5dc

kligarski: Thanks.

PR #3564, opened 2026-01-22
fix(Android, FormSheet): Account initial translationY for default animation

t0maboro:
## Description

Previously, the `translateY` value of the FormSheet was not accounted for when calculating its positioning in response to content changes - especially when the keyboard was open. As a result, the FormSheet incorrectly used the bottom of the screen as a reference point, causing unexpected layout behavior during content updates.

This PR introduces a fix that ensures the initial translateY is considered during layout recalculations

> [!NOTE]  
> The edge case where `newHeight > maxHeight` (without considering `translateY`) is not addressed in this PR. I will continue working on that as part of https://github.com/software-mansion/react-native-screens-labs/issues/802.

Fixes: https://github.com/software-mansion/react-native-screens-labs/issues/801

## Changes

- Included translationY in animation
- Included overflow support for translationY

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/294218c3-1b2c-473f-84e1-25db46353dbe" /> | <video src="https://github.com/user-attachments/assets/64501f40-fd74-4acf-bb54-934ed8102ce2" /> |

## Test plan

Added new example. Performed regression testing on Test2560 (with an assumption that overflows will be handled later). Performed regression testnig on TestFormSheet.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
@@ -353,19 +353,6 @@ class ScreenStackFragment :
         return sheetDelegate.createSheetExitAnimator(sheetAnimationContext)
     }
 
-    internal fun resolveMaxFormSheetHeight(): Int? {
-        if (screen.usesFormSheetPresentation()) {
-            val sheetDelegate = requireSheetDelegate()
-            return if (screen.sheetShouldOverflowTopInset) {
-                sheetDelegate.tryResolveContainerHeight()
-            } else {
-                sheetDelegate
-                    .tryResolveSafeAreaSpaceForSheet()
-            }
-        }
-        return null
-    }
-
     private fun createBottomSheetBehaviour(): BottomSheetBehavior<Screen> = BottomSheetBehavior<Screen>()
 
     private fun resolveBackgroundColor(screen: Screen): Int? {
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
```diff
@@ -353,6 +353,19 @@ class ScreenStackFragment :
...
+            val sheetDelegate = requireSheetDelegate()
+            return if (screen.sheetShouldOverflowTopInset) {
+                sheetDelegate.tryResolveContainerHeight()
+            } else {
+                sheetDelegate
+                    .tryResolveSafeAreaSpaceForSheet()
+            }
+        }
+        return null
+    }
```
kkafar: Can we move this calculation to the `SheetDelegate`? ALl other sheet height / dimensions method live there and let's keep them there.

kkafar: It can return `Int` then

t0maboro: should it be `Int`? both `tryResolveSafeAreaSpaceForSheet` and `tryResolveContainerHeight` are returning optionals, so maybe we want to keep the consistency

t0maboro: I'll propose this version: https://github.com/software-mansion/react-native-screens/pull/3564/commits/36351f482df76892d75cbfcc900b7952ca6e9f9f

kkafar: Yes, you're right here.

PR #3566, opened 2026-01-22
fix(Android, FormSheet): Constraint calculations by screen size

t0maboro:
## Description

BottomSheetBehavior has a physical limit (maxHeight) defined by the parent container. If the new content height exceeds this limit (by its size or keyboard offset), simply animating translationY back to 'initialTranslationY' would attempt to render the sheet larger than the screen.

We need to have constraint height inside the container's bounds. By including this overflow to our animation, we ensure the sheet stops expanding exactly at the maxHeight, preventing from being pushed off-screen or causing layout synchronization issues with the CoordinatorLayout.

Fixes: https://github.com/software-mansion/react-native-screens-labs/issues/802

## Changes

- Added clampedHeight and visibleDelta for better handling animation.

Original issue:

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/416e0040-04a6-408e-a952-648d4fcb6c3f" /> | <video src="https://github.com/user-attachments/assets/f913768b-3b95-4ab0-81e2-3d3d94e4eb62" /> |

Test2560 with overflow:

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/23377232-288a-4fa3-b2e3-262cf96d662f" /> | <video src="https://github.com/user-attachments/assets/595f5913-e8fa-48c3-b821-8f0889908d03" /> |

Keyboard regression testing:

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/478410e8-238f-4752-a781-0da833d0ad6b" /> | <video src="https://github.com/user-attachments/assets/e44f3536-368c-478e-a692-83c8af3d848c" /> |

## Test plan

Test3566, Test3564, Test2560, TestFormSheet

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -320,7 +320,8 @@ class Screen(
         val maxAvailableVerticalSpace =
             this.fragment
                 ?.asScreenStackFragment()
-                ?.resolveMaxFormSheetHeight() ?: return targetHeight
+                ?.sheetDelegate
+                ?.tryResolveMaxFormSheetHeight() ?: return targetHeight
 
         // Please note that currentTranslationY is rather < 0 here.
         // The translation is included in constraining the available space, because the FormSheet can have some offset, e.g. to


android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -187,8 +187,32 @@ class Screen(
         oldHeight: Int,
         newHeight: Int,
     ) {
-        val delta = (newHeight - oldHeight).toFloat()
-        val isContentExpanding = delta > 0
+        val keyboardOffset = this.translationY
```
kkafar: I think that calling this `keyboardOffset` is bold and might be rather inaccurate. translationY might be changed by:

1. keyboard,
2. dragging gesture,
3. ongoing value animation,
4. who knows what else.

I'd suggest sticking to "currentTranslationY".

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -187,8 +187,32 @@ class Screen(
...
+         * If the new content height exceeds this limit (by its size or keyboard offset), simply
+         * animating translationY back to 'initialTranslationY' would attempt to render the sheet
+         * larger than the screen.
+         *
+         * We need to have constraint height inside the container's bounds.
+         * By including this overflow to our animation, we ensure the sheet stops
+         * expanding exactly at the maxHeight, preventing from being pushed
+         * off-screen or causing layout synchronization issues with the CoordinatorLayout.
+         */
+        val maxHeight =
```
kkafar:
```suggestion
        val maxAvailableHeight =
```

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -187,8 +187,32 @@ class Screen(
...
     ) {
-        val delta = (newHeight - oldHeight).toFloat()
-        val isContentExpanding = delta > 0
+        val keyboardOffset = this.translationY
+
+        /*
+         * WHY OVERFLOW MATTERS:
+         * BottomSheetBehavior has a physical limit (maxHeight) defined by the parent container.
+         * If the new content height exceeds this limit (by its size or keyboard offset), simply
+         * animating translationY back to 'initialTranslationY' would attempt to render the sheet
```
kkafar: you refer in this comment to `initialTranslationY` variable that does no longer exist.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -187,8 +187,32 @@ class Screen(
...
+         * expanding exactly at the maxHeight, preventing from being pushed
+         * off-screen or causing layout synchronization issues with the CoordinatorLayout.
+         */
+        val maxHeight =
+            this.fragment
+                ?.asScreenStackFragment()
+                ?.resolveMaxFormSheetHeight() ?: return
+
+        val clampedOldHeight = oldHeight.coerceAtMost((maxHeight + keyboardOffset).toInt())
+        val clampedNewHeight = newHeight.coerceAtMost((maxHeight + keyboardOffset).toInt())
```
kkafar: I don't understand this part yet. Why do we clamp the value by `maxHeight + keyboardOffset`. Literally from the variable naming, something you compute & name as `maxHeight` should be a max value. Clearly it's not here. Can we fix it please? 

Disregarding the variable naming, I also don't understand what is the meaning of these variables. Can you describe what are those? `clampedOldHeight` can not be "current height of the sheet", because `oldHeight` argument here is `sheetBehaviour.maxHeight` value, which, ..., oh wait... I context of `fitToContents` `maxHeight` is the only height the sheet can have, right? Therefore `oldHeight` is the actual last sheet height, as defined by content wrapper layout. Here you coerce it in case there is an overflow. Am I right?

1. This needs comments with some explanation. 
2. We need better variable naming. We can consult in the office.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
kkafar: or `maxPossibleSheetHeight`?

android/src/main/java/com/swmansion/rnscreens/Screen.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3566/commits/db43a310912deae0cdcb314a22f22b65bc8a4abf

android/src/main/java/com/swmansion/rnscreens/Screen.kt
t0maboro: https://github.com/software-mansion/react-native-screens/commit/db43a310912deae0cdcb314a22f22b65bc8a4abf

android/src/main/java/com/swmansion/rnscreens/Screen.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3566/commits/2dbc4c8e9dfcb785fb0024d621f11954722230b7

android/src/main/java/com/swmansion/rnscreens/Screen.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3566/commits/db43a310912deae0cdcb314a22f22b65bc8a4abf + https://github.com/software-mansion/react-native-screens/pull/3566/commits/2014cc5d6746b442059ef527fe7a8787a3b0bb4d

t0maboro:
> What goes on in case the default resize animation is disabled?
> 
> I see that here, in `updateSheetContentHeightWithAnimation` you actually call `behavior.updateMetrics(**clampedNewHeight**)` with the "clamped height", whereas it does not happen in case when the animation is disabled. Is that intentional?

Let me check that, but I think you're right that we should work with the clampedHeight either, manipulating with the sheet outside the device dimensions doesn't make sense at all.

t0maboro:
> What goes on in case the default resize animation is disabled?
> 
> I see that here, in `updateSheetContentHeightWithAnimation` you actually call `behavior.updateMetrics(**clampedNewHeight**)` with the "clamped height", whereas it does not happen in case when the animation is disabled. Is that intentional?

it should be aligned now: https://github.com/software-mansion/react-native-screens/pull/3566/commits/0028045959384502ec7968c4ed0803937ba87404

PR #3572, opened 2026-01-23
refactor: sheetDefaultResizeAnimationEnabled -> sheetResizeAnimationEnabled

t0maboro:
## Description

Just renaming prop, the "Default" doesn't add any value to it.

> [!NOTE]  
> I'm intentionally omitting the deprecation path, assuming that no one has consumed this prop so far, as it was released recently.

## Changes

- Rename prop

N/A

## Test plan

Just renaming prop, I believe that green CI is enough,

## Checklist

- [ ] Ensured that CI passes


t0maboro:
as discussed internally, we'd like to avoid breaking change here, we'll apply proper update only to react-navigation

PR #3576, opened 2026-01-27
feat(Android, Stack v5): handle preloading screens

kligarski:
## Description

Adds support for preloading screens. Allows for multiple operations in one React transition.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/819.

#### Preload

As described in the description of https://github.com/software-mansion/react-native-screens/pull/3531, previous approach to handling detached screens did not work due to how `react-native` works when children views are reordered and inserted again.

To mitigate this problem, we decided to change the approach. Now, `StackScreen`s will be rendered inside `StackHost` in the following order:

1. First, all attached screens are rendered in the order that reflects current stack state (so ABC always matches to stack with A at the bottom and C at the top)
2. Then, all detached screens are rendered.

This change has 2 main advantages:

1. We will never move a screen in `attached` state therefore we will avoid React removing the screen from parent view when trying to reinsert it.
2. In result of the above, there will never be a situation where an attached screen will be popped and pushed in one transaction (unless someone explicitly does so) therefore we don't need to implement a mechanism for cancelling out those operations.

There is one drawback: this approach will not help with fixing the inspector, it needs to be handled separately.

#### Multiple operations in one React transaction.

When multiple operations are performed in one transaction, we can't rely on the order of updates from React.

For example, if we have `ABCdefg` state (lowercase means detached state) and we perform at the same time:
```
pop(C)
pop(B)
push(G)
push(E)
```

we end up with `AGEbcdf` state. This information is passed to native side via those operations:

```
remove(G)
remove(E)
insert(G, 1)
insert(E, 2)
update(B, 'detached')
update(C, 'detached')
update(E, 'attached')
update(G, 'attached')
```

If we tried to translate those operations directly without any reordering, we would get:
```
pop(B)
pop(C)
push(E)
push(G)
```

1. Order of pops is incorrect - we need to pop the top screen (C) first.
2. Order of pushes is incorrect - we would get `AEG...` instead of desired `AGE...`.

Therefore we need a mechanism to synchronize the order of operations with the state of the stack. We can use `StackHost`'s children to determine this - if we know that our children are `AGEbcdf`, we know that C needs to be popped before B and G needs to be pushed before E.

## Changes

- add `StackContainerUpdateCoordinator` to ensure correct order of push & pop operations
- add separate push & pop queues to `StackContainer` to ensure order (first pops, then pushes) + in the future, allow for simple detection of necessity of using replace animation
- add batch action to `StackContainer`
- create `Test3576` to test batch operations in new stack implementation

### Preload on TestScreenStack

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/52ab058b-7bec-4141-b1d7-939e0ba0619c" /> | <video src="https://github.com/user-attachments/assets/75641233-0bf7-4e2b-a430-be3d71371c90" /> |

### Multiple operations in one transaction

https://github.com/user-attachments/assets/00c61bde-3277-4953-9a06-222953c320bb

## Test plan

1. Run `TestScreenStack`. Preload B, push A, push B. Hierarchy should contain screens ABA.
4. Run `Test3576`. Run scenario 1. 

## Checklist

- [ ] Ensured that CI passes


android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
```diff
@@ -58,87 +49,68 @@ internal class StackContainer(
         // If container update is requested before container is attached to window, we ignore
         // the call because we don't have valid fragmentManager yet.
         // Update will be eventually executed in onAttachedToWindow().
-        if (pendingOperationQueue.isNotEmpty() && isAttachedToWindow) {
+        if ((pendingPopOperations.isNotEmpty() || pendingPushOperations.isNotEmpty()) && isAttachedToWindow) {
```
kkafar: Nitpick: I'd move this to a computed property `hasPendingOperations`

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
```diff
@@ -58,91 +51,75 @@ internal class StackContainer(
...
     private fun performPopOperation(
         fragmentManager: FragmentManager,
         operation: PopOperation,
     ) {
-        val backStackEntryCount = fragmentManager.backStackEntryCount
-        require(backStackEntryCount > 0) { "[RNScreens] Back stack must not be empty." }
+        val associatedFragment = stackScreenFragments.find { it.stackScreen === operation.screen }
+        require(associatedFragment != null) {
+            "[RNScreens] Unable to find a fragment to pop."
+        }
```
kkafar, line 103: This is tricky.

I can imagine following situation: we already scheduled `add` transaction with `performAddOperation`, so the new fragment is already stored in `stackScreenFragments` array. You find it here, therefore, and later you pop wrong screen, before the `add` transaction has still not been executed.

I dunno how to solve it correctly yet.

kkafar: Let's move forward for now, but we need to create ticket for this.

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
```diff
@@ -58,91 +51,75 @@ internal class StackContainer(
...
-    ) {
         val associatedFragment = StackScreenFragment(WeakReference(this), operation.screen)
         stackScreenFragments.add(associatedFragment)
+
         transaction.add(this.id, associatedFragment)
+
+        // Don't add root screen to back stack to handle exiting from app.
+        if (fragmentManager.fragments.isNotEmpty()) {
+            transaction.addToBackStack(operation.screen.screenKey)
+        }
```
kkafar, line 91: Isn't it also risky here? For similar reasons as in [the previous message](https://github.com/software-mansion/react-native-screens/pull/3576/changes#r2741422139).

In case you push multiple screens e.g. in first transaction, it is highly likely that the very first fragment won't be attached to fragment manager yet when you add the second fragment & this check will prevent it from being added to the back stack. 

We need to coordinate this with removal also. Let's move forward now, but create ticket for this.

android/src/main/java/com/swmansion/rnscreens/gamma/stack/screen/StackScreenFragment.kt
```diff
@@ -9,38 +9,26 @@ import com.swmansion.rnscreens.gamma.stack.host.StackContainer
...
+        screenLifecycleEventEmitter = stackScreen.createAppearanceEventsEmitter(viewLifecycleOwner)
     }
 
     override fun onDestroyView() {
-        stackContainer.get()?.onFragmentDestroyView(this)
-        stackScreen.eventEmitter.emitOnDismiss(stackScreen.activityMode == StackScreen.ActivityMode.ATTACHED)
         super.onDestroyView()
+        stackScreen.onDismiss()
+        stackContainer.get()?.onFragmentDestroyView(this)
+        screenLifecycleEventEmitter = null
```
kkafar, line 35: I decided here to pass the information to both container & screen. 

I'm not sure what is the most elegant approach here, but didn't want to spend more time in this place. We can refactor this later if needed.

kkafar (approved):
I've pushed all the changes I had been interested in.

I leave few remarks pointing out potentially serious problems, however I wanna proceed here & handle these later. I'll create standalone tickets for them.

PR #3584, opened 2026-01-29
fix(iOS, Stack): Fix for reattaching Screens when swiping back quickly

t0maboro:
## Description

This PR introduces logic to collect only active (non-dismissed) screen controllers coming from JS state.

When a screen gets dismissed on the native side, its state is sent asynchronously to the JS. In scenarios where multiple screens are dismissed quickly, this can result in stale updates being sent from JS to native. These outdated updates may reference screens that are no longer part of the native view hierarchy.

Since view recycling is disabled, we do not expect the `RNSScreenView` that was dismissed to be reattached; a new instance should be created when we push an update from JS. Then the new controller instance will be initialized, so these associated with detached screens won't be reused again. Therefore, we can safely filter them out from JS state on the native side. On the JS side, the state will stabilize at some point and align with native, once we receive all asynchronous events.

```objective-c
@implementation RNSScreenView {
...
- (void)initCommonProps
{
  _controller = [[RNSScreen alloc] initWithView:self];
```

To address this, we now filter out any dismissed controllers from the list received from JS.

> [!NOTE]  
> I'm leaving the previous behavior as the default, and I'm adding `iosPreventReattachmentOfDismissedScreens` flag that can enable this fix. We're right before a very important release, and this PR is touching the very basic logic; therefore, we don't want to have this fix enabled by default for now. You can opt-in by setting `featureFlags.experiment.iosPreventReattachmentOfDismissedScreens = true`. Any feedback would be valuable here.

Closes: https://github.com/software-mansion/react-native-screens/issues/2559

## Changes

- Added an attribute to mark controllers that were dismissed.
- Added logic for filtering out screens dismissed natively from JS state.

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/684a5b6d-4a1a-4da5-9bbb-3973f2cc342f" /> | <video src="https://github.com/user-attachments/assets/839debc5-7b8c-43b8-8fa6-6ec9595e708c" /> |

## Test plan

Added Test 2559.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -1895,6 +1895,8 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
     // Since view recycling is disabled, we can rely on a flag indicating that the controller
     // has been removed from the hierarchy, as it will not be reused.
     _isRemovedFromParent = YES;
+  } else {
+    _isRemovedFromParent = NO;
   }
 
   [super didMoveToParentViewController:parent];


ios/RNSScreenStack.mm
```diff
@@ -668,8 +668,26 @@ - (void)setModalViewControllers:(NSArray<UIViewController *> *)controllers
...
+  /// We collect only active (non-dismissed) screen controllers.
+  /// This is necessary because when a Screen is dismissed natively, its state is sent **asynchronously** to JS.
+  /// As a result, JS might send a delayed update back to native when multiple screens are being dismissed quickly.
+  /// Since view recycling is disabled, once we detect that a screen has been removed from the view hierarchy,
+  /// it won't be reused. This allows us to safely filter out dismissed screens from screens coming from JS state via
+  /// `controllers`.
+  NSMutableArray<UIViewController *> *activeControllers = [NSMutableArray arrayWithCapacity:controllers.count];
+
+  for (UIViewController *vc in controllers) {
+    BOOL isScreen = [vc isKindOfClass:[RNSScreen class]];
```
kkafar: What else could be inside that array other than RNSScreen?

kkafar: It'd be nice to move all that filtering code into a dedicated function. So that it does not clutter that function even more than it already is.

t0maboro: It's just extra safety before casting, I don't expect anything else than Screen, we may consider throwing an error if sth else is coming

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3584/commits/a47f2cd4eff695db4880bcbd543941bbce6cb8d9 + https://github.com/software-mansion/react-native-screens/pull/3584/changes/99c8cb23055bde16e4a959dc1be34987f48f702a

kkafar (commented):
Since this change impacts rather core & crucial functionality of stack, I'd like it to be put behind a feature flag or delay its rollout, especially that we plan to release a next version soon & it might potentially end up in SDK 55. By allowing to control this change through JS we'll allow to patch it much more easily.

I'd suggest adding a "experiment" feature flag, e.g. `iosPreventReattachmentOfDismissedScreens` and consume it directly on `screenstack` component.

We can make this flag a noop in two or three minor versions, but next minor should be possibly stable. 

What do you think?

t0maboro:
> Since this change impacts rather core & crucial functionality of stack, I'd like it to be put behind a feature flag or **delay its rollout**, especially that we plan to release a next version soon & it might potentially end up in SDK 55. By allowing to control this change through JS we'll allow to patch it much more easily.
> 
> I'd suggest adding a "experiment" feature flag, e.g. `iosPreventReattachmentOfDismissedScreens` and consume it directly on `screenstack` component.
> 
> We can make this flag a noop in two or three minor versions, but next minor should be possibly stable.
> 
> What do you think?

https://github.com/software-mansion/react-native-screens/pull/3584/changes/8e0f9b3291069bd28fb652f78a6a8f771a2d9755 + https://github.com/software-mansion/react-native-screens/pull/3584/changes/c6e5f1c26918ac1c974c70772b5ccf879a9ab9ac

t0maboro:
> We could consider clearing the state in `didMoveToParentViewController` in case the controller is reattached. Beside that, looks good.

https://github.com/software-mansion/react-native-screens/pull/3584/changes/c9e923bf391751c040354ec3c0815569b47d2992

Issue #3592, opened 2026-01-30
[Mac Catalyst] build error in react-native-screens/ios/bottom-tabs/host/RNSBottomTabsHostEventEmitter.mm


dppo:
### Description

[like this](https://github.com/software-mansion/react-native-screens/issues/3197)

but in node_modules/react-native-screens/ios/bottom-tabs/host/RNSBottomTabsHostEventEmitter.mm

<img width="1768" height="432" alt="Image" src="https://github.com/user-attachments/assets/a5c126dc-fd48-45ca-a710-ab14f789cdb6" />

### Steps to reproduce

1. use react-native-screens in Mac Catalyst
2. build

### Snack or a link to a repository

https://github.com/dppo/ReactNavigationExapmle

### Screens version

4.20.0

### React Native version

0.83.1

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

kkafar:
@dppo thanks for the report.

We do not officially support Mac Catalyst, and we do not test on Mac Catalyst nor we build for that target. 

Usually, since it is not a supported platform, I'd ask you to prepare a patch, but since it seemingly requires one line of code changes that seem harmless, I'll follow precedence from https://github.com/software-mansion/react-native-screens/pull/3202 and prepare it 😄

PR #3594, opened 2026-01-30
fix(iOS): fix possible build issues on Mac Catalyst due to implicit conversions

kkafar:
## Description

Closes <https://github.com/software-mansion/react-native-screens/issues/3592>

The fix relies on adding an explicit cast from `BOOL` to `bool`.
Currently, I am not aware of any scenario where described cast could
fail, therefore I determine it to be safe.

This change reportedly fixes an issue with build on Mac Catalyst.
The fix is analogous to the one I did in <https://github.com/software-mansion/react-native-screens/pull/3202>.

N/A

## Test plan

This diff does not change behavior on our end. For us sufficient testing plan would be
for out CIs to pass. I'm relying here on user confirmation that the PR indeed fixes the issue.

## Checklist

- [x] Ensured that CI passes


kkafar:
@dppo, please confirm that this patch solves [the issue](https://github.com/software-mansion/react-native-screens/issues/3592) on your end.

PR #3596, opened 2026-01-30
fix(iOS, Tabs): Request layout pass after loading image from source

t0maboro:
## Description

Before assigning the image, the tab bar might already be attached to the view hierarchy. This can cause an issue where the tab bar layout is not correctly updated, as it might have been added to the window before the image was set. As a result, the layout adjustments (e.g., image positioning) are not applied immediately.

To resolve this, we explicitly request an additional layout pass after the image is assigned. This ensures that the tab bar updates its layout properly without requiring user interaction (e.g., switching tabs) to trigger a recalculation.

The issue is illustrated in the image below.

before-image

## Changes

- Add layout request in the completion callback when the image is loaded from source.
- Add example & assets demonstrating the issue.

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/bcaedb0f-ded8-482e-9402-7c178c50773c" />  | <video src="https://github.com/user-attachments/assets/7e8dd72a-e496-4af4-9606-50cc5b9a1277" /> |

## Test plan

Added Test3596, performed regression testing on TestBottomTabs

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/bottom-tabs/RNSTabBarAppearanceCoordinator.mm
+++ b/ios/bottom-tabs/RNSTabBarAppearanceCoordinator.mm
@@ -138,6 +138,9 @@
 
   // A layout pass is required because the image might be loaded asynchronously,
   // after the tab bar has already been attached to the window.
+  // This code handles case where image passed by the user is not 
+  // of appropriate size & needs to be readjusted. W/o additional 
+  // layout here the icon would be displayed with original dimensions.
   UIViewController *parent = screenView.controller.parentViewController;
   if ([parent isKindOfClass:[UITabBarController class]]) {
     UITabBarController *tabBarVC = (UITabBarController *)parent;


ios/bottom-tabs/RNSTabBarAppearanceCoordinator.mm
```diff
@@ -119,6 +125,26 @@ - (void)setIconsForTabBarItem:(UITabBarItem *)tabBarItem
...
+           forScreenView:(RNSBottomTabsScreenComponentView *)screenView
+{
+  if (isSelected) {
+    tabBarItem.selectedImage = image;
+  } else {
+    tabBarItem.image = image;
+  }
+
+  // A layout pass is required because the image might be loaded asynchronously,
+  // after the tab bar has already been attached to the window.
```
kkafar, line 140:
```suggestion
  // A layout pass is required because the image might be loaded asynchronously,
  // after the tab bar has already been attached to the window.
  // This code handles case where image passed by the user is not 
  // of appropriate size & needs to be readjusted. W/o additional 
  // layout here the icon would be displayed with original dimensions.
```

PR #3597, opened 2026-01-30
feat(Android, Stack v5): nested stack uses childFragmentManager of parent container

kkafar:
## Description

Closes <https://github.com/software-mansion/react-native-screens-labs/issues/921>

`StackScreen` implements now `FragmentProviding` interface. This causes
`StackContainer` of nested stack to use `childFragmentManager` of parent
container instead of reusing `supportFragmentManager`.

## Changes

N/A

## Visual documentation

N/A

## Test plan

Not a fix or change in behavior, this is just new behavior.
To verify you need to attach a debugger in `StackContainer` and verify that it uses
`childFragmentManager` of parent container.

I've added test example for this use case that I'll likely follow developing in
next PRs.


kkafar:
It's only linter. I'll land it & fix it with follow up PRs @t0maboro. 
I want to avoid all the rebase hustle.

PR #3599, opened 2026-01-31
fix(Android, Stack v5): update stack model after native pop

kkafar:
## Description

Currently when doing native pop, the `stackModel` stored in `StackContainer`
is never updated, therefore leading to an incorrect state held (natively
dismissed fragments are never removed from the state).

This PR changes that.

## Changes

Currently I remove them from the state in `onFragmentDestroyView`,
and it seems to work, however I'm not sure, this is the perfect
place timing-wise. This callback is called only after fragment's
view is destroyed, which might be pretty late in component
lifecycle, and we might execute container operations still taking
into consideration possibly removed screens. We need to look into
that later.

N/A

## Test plan

Only sensible way to test this is to add logging in `StackContainer`
and dismiss few screens natively, and see that the state is updated (with this fix).

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackOperation.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackOperation.kt
@@ -2,7 +2,7 @@ package com.swmansion.rnscreens.gamma.stack.host
 
 import com.swmansion.rnscreens.gamma.stack.screen.StackScreen
 
-internal sealed class StackOperation()
+internal sealed class StackOperation
 
 internal class PushOperation(
     val screen: StackScreen,


android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
```diff
@@ -115,10 +125,13 @@ internal class StackContainer(
             transaction.commitNowAllowingStateLoss()
         }
 
-        stackScreenFragments.remove(associatedFragment)
+        stackModel.remove(associatedFragment)
     }
 
     internal fun onFragmentDestroyView(fragment: StackScreenFragment) {
+        if (stackModel.remove(fragment) && !fragment.stackScreen.isNativelyDismissed) {
+            Log.e(TAG, "[RNScreens] StackContainer natively popped a screen that was not in model!")
```
t0maboro, line 133: I might misinterpret sth here, but:
1. when `stackModel.remove(fragment)` returns true, it means that the `fragment` WAS in `stackModel` - the whole condition still might be valid
2. inside the log message, you're mentioning that the `fragment` WAS NOT in `stackModel`
so these two are contradictory for me

kkafar: You're right. This is badly worded.

Will be fixed here: https://github.com/software-mansion/react-native-screens/pull/3601/changes

PR #3608, opened 2026-02-02
fix: `updateSelectedTab` removing unowned fragments

lukmccall:
## Description

Fixes `updateSelectedTab` removing unowned fragments (fragments that were created by other libraries) 

## Changes

- Made sure that the code will only remove `TabScreenFragment` 

## Test plan

This change was tested using the new Expo template in the preview version of Expo Go. 
Before, the `react-native-screens` removed the fragment created by `expo-dev-menu`. With this fix, everything seems to work.

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
@@ -382,8 +382,9 @@ class TabsHost(
     private fun updateSelectedTab() {
         val newFocusedTab = currentFocusedTab
 
-        check(requireFragmentManager.fragments.size <= 1) { "[RNScreens] There can be only a single focused tab" }
-        val oldFocusedTab = requireFragmentManager.fragments.firstOrNull()
+        val tabFragments = requireFragmentManager.fragments.filterIsInstance<TabScreenFragment>()
+        check(tabFragments.size <= 1) { "[RNScreens] There can be only a single focused tab" }
+        val oldFocusedTab = tabFragments.firstOrNull()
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
```diff
@@ -382,8 +382,9 @@ class TabsHost(
     private fun updateSelectedTab() {
         val newFocusedTab = currentFocusedTab
 
-        check(requireFragmentManager.fragments.size <= 1) { "[RNScreens] There can be only a single focused tab" }
-        val oldFocusedTab = requireFragmentManager.fragments.firstOrNull()
+        val tabFragments = requireFragmentManager.fragments.filterIsInstance<TabScreenFragment>()
```
t0maboro, line 385: overall looks good, but I'm wondering whether we could use `tabScreenFragments` and detect it by `isFocusedTab`, then we'd always operate only on our fragments (on the other hand, RN is becoming the source of truth then, instead of native state, so this might not be valid), cc-ing @kkafar as he has better context for the tabs impl. on Android

t0maboro (approved):

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
kkafar: not really. `tabScreenFragments` do hold fragments for all rendered screens, possibly many, while there should be only a single `TabScreenFragment` of ours attached to the fragment manager. And this is what these assertions are for.

PR #3613, opened 2026-02-03
feat(Tabs): add native RTL support for bottom tabs on iOS & Android

ahmedawaad1804:
## Description

This PR adds proper RTL (Right-to-Left) layout support for native bottom tabs on both iOS and Android in react-native-screens.

Previously, bottom tabs did not support the RTL direction, which caused incorrect tab ordering and layout in RTL locales (e.g. Arabic, Hebrew).

With this change bottom tabs now support changing layout direction. This brings native behavior in line with expected platform RTL handling and React Native layout conventions.

### Details

@kligarski:

#### Android
On Android, the direction works out-of-the-box as it's propagated through view hierarchy. We pass the value of the `direction` prop directly to `TabsHost` view.

##### Badges
Setting `semanticContentAttribute` for `_controller.tabBar` and `_controller.view` is not enough, e.g. badges visible through liquid glass lens are still in LTR.

https://github.com/user-attachments/assets/0a96b18b-ddfe-4dd3-99c1-4a08a89ad86d

To handle this, we tried using the same approach as in native stack - we set `UIView`'s `appearanceWhenContainedInInstancesOfClasses` of the tab bar ([details how it's handled in the header are here](https://github.com/software-mansion/react-native-screens/pull/2185/changes#diff-e5ef5b6e29f17bca80b51bc0c5faef1a44bac24e00952b30ac822520213dc6a5R504)).

However, this does not work for tab bar & sidebar on iPad starting from iOS 18 as it is not a part of `controller.tabBar`. `_UITabContentView` is mounted under `controller.view`. Using `appearanceWhenContainedInInstancesOfClasses` for `_UITabContentView` (which is already sketchy as this is an internal UIKit class) helps with the order of items in the tab bar but the sidebar appears on the wrong side of the screen.

That's why I decided to use modern way to handle direction via trait overrides. For iOS prior to 17, you need to apply overrides on parent view controller ([see here](https://developer.apple.com/documentation/uikit/uiviewcontroller/setoverridetraitcollection(_:forchild:)?language=objc)). This isn't the cleanest solution as the controller changes property of the other controller which might not be a controller belonging to `react-native-screens` but I think that this is the lesser evil. If this turns out to be problematic, we can consider introducing some kind of `ScreensRootView` that will ensure that there is a top controller from `screens`.

For iOS 17+, we can use [`traitOverrides`](https://developer.apple.com/documentation/uikit/uiviewcontroller/traitoverrides-8u19n?language=objc) directly on the controller - this works with the top tab bar/sidebar on iPadOS 18+.

##### ScrollView
On iOS, there is a bug with content of the ScrollView being moved off screen after tab changes. I've reported the issue to `react-native`: https://github.com/facebook/react-native/issues/55768. The fix has been merged (https://github.com/facebook/react-native/pull/55804) and should be available in next `react-native` release.

https://github.com/user-attachments/assets/b033d4c7-bfbe-415b-a02c-66fcfed6d0d6

##### Bottom Accessory
There seems to be a bug with bottom accessory in RTL when search role is NOT used for one of the tabs (Apple Music and Apple Podcasts use search role so the bug isn't visible).

| no search role (bug) | search role (no bug) |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/f5a154c0-4e52-40ca-8263-add687c23e65" /> | <video src="https://github.com/user-attachments/assets/cddcfab4-90b5-4387-8bea-e9c97c3b29e7" />

This bug is reproducible in bare UIKit app on iOS 26.2. 

https://github.com/user-attachments/assets/d0a36a07-f6a9-4bfd-9ac9-b5721a70e134

I've added this to our internal board (https://github.com/software-mansion/react-native-screens-labs/issues/986) and we'll check whether it has been fixed in iOS 26.3/26.4 beta.

##### Top tab bar badges

On iPadOS 18+, there seems to be a bug with the initial position of the badges. They move to correct position after a tab change. I added a ticket on our internal board to check whether this is a native bug: https://github.com/software-mansion/react-native-screens-labs/issues/991.

https://github.com/user-attachments/assets/09c25789-b67c-40a1-a3e0-688043cbe17e

##### Native localization vs `react-native` on iOS

> [!IMPORTANT]
>
> When RTL is forced via `I18nManager.forceRTL(true)` but the language of the native app isn't an RTL language, the views related to containers such as the tab bar/sidebar by default will remain in LTR on iOS. This is because we want to rely on native mechanism for layout direction which is the trait system instead of `semanticContentAttribute` (used by regular `react-native` views) which should only define whether the view should be flipped in RTL & does not propagate down the hierarchy. `forceRTL` does not change the trait therefore containers use layout direction of the native app. In order to support `forceRTL`, you should use `direction={I18nManager.isRTL ? 'rtl' : 'ltr'}` (see our example implementation in `BottomTabsContainer.tsx`). This will override the trait from the app with layout direction from `react-native` and propagate it down the hierarchy.

## Changes

- add `direction` prop to `TabsHost` and implement it for both platforms
- add `Test3598.tsx`

### Before
(iOS only because Android works out of the box).
Simulator Screenshot - iPhone 17 Pro Max - 2026-02-03 at 18 39 51

#### Android
| Android Test3598 |
| --- |
| <video src="https://github.com/user-attachments/assets/07fa151c-94b8-465e-892b-af1aa74deb83" /> |

#### iOS
| iOS Test3598 | iOS TestBottomTabs | iOS Test3288 |
| --- | --- | --- |
| <video src="https://github.com/user-attachments/assets/6ab88d38-df51-4869-9698-97d1b2ac2079" /> | <video src="https://github.com/user-attachments/assets/91d4dffc-c7a3-458c-a973-c61fefc6a2e7" /> | <video src="https://github.com/user-attachments/assets/dc775be4-bc07-4a82-8b42-93f54683ba84" /> |

## Test plan
Use `Test3598`, `TestBottomTabs`, `Test3288` (iOS).

Tested on:
- Android (RTL system language enabled)
- iOS (RTL simulator + device)

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/conversion/RNSConversions-Tabs.mm
+++ b/ios/conversion/RNSConversions-Tabs.mm
@@ -501,4 +501,20 @@ UITraitEnvironmentLayoutDirection UITraitEnvironmentLayoutDirectionFromTabsHostC
   }
 }
 
+UIUserInterfaceStyle UIUserInterfaceStyleFromHostProp(react::RNSTabsHostColorScheme colorScheme)
+{
+  using enum facebook::react::RNSTabsHostColorScheme;
+  switch (colorScheme) {
+    case Inherit:
+      return UIUserInterfaceStyleUnspecified;
+    case Light:
+      return UIUserInterfaceStyleLight;
+    case Dark:
+      return UIUserInterfaceStyleDark;
+    default:
+      RCTLogError(@"[RNScreens] unsupported color scheme");
+      return UIUserInterfaceStyleUnspecified;
+  }
+}
+
 }; // namespace rnscreens::conversion
--- a/FabricExample/yarn.lock
+++ b/FabricExample/yarn.lock
@@ -5523,13 +5523,13 @@ __metadata:
   linkType: hard
 
 "fast-xml-parser@npm:^4.4.1":
-  version: 4.5.2
-  resolution: "fast-xml-parser@npm:4.5.2"
+  version: 4.5.4
+  resolution: "fast-xml-parser@npm:4.5.4"
   dependencies:
     strnum: "npm:^1.0.5"
   bin:
     fxparser: src/cli/cli.js
-  checksum: 10c0/972cc5a54f76676485a776952b1e68d765ddca3c0053d34db613152a93e2ee74dfd3b630672b33a978bf3a257ae24d4c017ae92f5c2100e6a5312e679fb591a5
+  checksum: 10c0/7989148650fc1fce988798b62467f7dee0fd5a7ad049373e00e65fbc68f689c119975d03da32c427f0a1f5aad01de2efbd783a48dcdaf3ca41817a8b161ad3e8
   languageName: node
   linkType: hard
 
--- a/TVOSExample/yarn.lock
+++ b/TVOSExample/yarn.lock
@@ -7675,11 +7675,11 @@ __metadata:
   linkType: hard
 
 "minimatch@npm:^3.0.4, minimatch@npm:^3.0.5, minimatch@npm:^3.1.1, minimatch@npm:^3.1.2":
-  version: 3.1.2
-  resolution: "minimatch@npm:3.1.2"
+  version: 3.1.5
+  resolution: "minimatch@npm:3.1.5"
   dependencies:
     brace-expansion: "npm:^1.1.7"
-  checksum: 10c0/0262810a8fc2e72cca45d6fd86bd349eee435eb95ac6aa45c9ea2180e7ee875ef44c32b55b5973ceabe95ea12682f6e3725cbb63d7a2d1da3ae1163c8b210311
+  checksum: 10c0/2ecbdc0d33f07bddb0315a8b5afbcb761307a8778b48f0b312418ccbced99f104a2d17d8aca7573433c70e8ccd1c56823a441897a45e384ea76ef401a26ace70
   languageName: node
   linkType: hard
 
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
@@ -1,13 +1,13 @@
-package com.swmansion.rnscreens.gamma.tabs.host
+package com.swmansion.rnscreens.gamma.tabs.appearance
 
 import android.annotation.SuppressLint
+import android.content.Context
 import android.content.res.ColorStateList
 import android.graphics.drawable.StateListDrawable
 import android.util.TypedValue
 import android.view.MenuItem
 import android.view.ViewGroup
 import android.widget.TextView
-import androidx.appcompat.view.ContextThemeWrapper
 import androidx.core.view.children
 import androidx.core.view.isVisible
 import com.facebook.react.common.assets.ReactFontManager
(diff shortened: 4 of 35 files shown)


kkafar:
Hey, thanks for the PR. This is great, we need it. I hope to review it somewhere this week. Will keep you up to date.

kligarski:
Hi, thank you for the PR. I'll push some changes and update PR description so we can land it soon. I hope you don't mind.

ahmedawaad1804:
@kligarski  sure thank you :D

ahmedawaad1804:
@kkafar welcome , any time :)

ios/bottom-tabs/host/RNSBottomTabsHostComponentView.mm
```diff
@@ -352,6 +352,24 @@ - (void)updateProps:(const facebook::react::Props::Shared &)props
       }
   }
 
+  if (newComponentProps.directionMode != oldComponentProps.directionMode) {
+    _directionMode =
```
kligarski: We can use `[RCTI18nUtil isRTL]` instead of passing a prop, it should have an updated value as it reads it from `[NSUserDefaults standardUserDefaults]` (but it would be unable to react to dynamic changes). I'm not sure if we can consider this a stable API but it has been mentioned in blog post in react-native: https://reactnative.dev/blog/2016/08/19/right-to-left-support-for-react-native-apps.

cc @kkafar - let me know what you think is better

We can also leave the prop for now (it's not exposed as a part of API either way) and rethink our approach to RTL in separate PR - I'm wondering whether RTL should be handled via some top-level wrapper component for the entire hierarchy below it.

ios/bottom-tabs/host/RNSBottomTabsHostComponentView.mm
```diff
@@ -544,6 +562,24 @@ - (void)setTabBarControllerModeFromRNSTabBarControllerMode:(RNSTabBarControllerM
...
+        : UITraitEnvironmentLayoutDirectionLeftToRight;
+  } else
+#endif // RNS_IPHONE_OS_VERSION_AVAILABLE(17_0)
+  {
+    _controller.view.semanticContentAttribute = _directionMode;
+    _controller.tabBar.semanticContentAttribute = _directionMode;
+    [[UIView appearanceWhenContainedInInstancesOfClasses:@[ _controller.tabBar.class ]]
+        setSemanticContentAttribute:_directionMode];
+  }
+}
```
t0maboro: This is repeated from updateProps, can be deduplicated with a helper method

ios/bottom-tabs/host/RNSBottomTabsHostComponentView.mm
satya164: @kligarski in react navigation, we have a [`direction` prop on `NavigationContainer`](https://reactnavigation.org/docs/8.x/navigation-container/#direction) to control layout direction which we will pass it to navigators that can handle it.

ios/bottom-tabs/host/RNSBottomTabsHostComponentView.mm
kligarski: My reasoning here was that we're getting rid of Paper soon but I guess that helper method won't hurt.

https://github.com/software-mansion/react-native-screens/pull/3613/commits/fd3261bef708c954bd5e96360b26d02fcbf3a3bd

src/components/helpers/PlatformUtils.ts
kligarski, line 1: This util was useful in previous version of this PR but now it isn't really used. It's out of scope of this PR but if you don't mind, I'd keep it as it makes the code cleaner.

android/src/paper/java/com/facebook/react/viewmanagers/RNSTabsHostManagerDelegate.java
```diff
@@ -30,6 +30,9 @@ public void setProperty(T view, String propName, @Nullable Object value) {
       case "nativeContainerBackgroundColor":
         mViewManager.setNativeContainerBackgroundColor(view, ColorPropConverter.getColor(value, view.getContext()));
         break;
+      case "directionMode":
+        mViewManager.setDirectionMode(view, (String) value);
+        break;
```
kmichalikk: this is different to `layoutDirection`
did the name change & this hasn't been updated?

common/cpp/react/renderer/components/rnscreens/RNSSplitScreenComponentDescriptor.h
```diff
@@ -14,8 +14,7 @@ class RNSSplitScreenComponentDescriptor final
   using ConcreteComponentDescriptor::ConcreteComponentDescriptor;
 
   void adopt(ShadowNode &shadowNode) const override {
-    react_native_assert(
-        dynamic_cast<RNSSplitScreenShadowNode *>(&shadowNode));
+    react_native_assert(dynamic_cast<RNSSplitScreenShadowNode *>(&shadowNode));
```
kmichalikk: nit: this file is not related to the PR, can you remove it?

ios/conversion/RNSConversions-SplitView.mm
```diff
@@ -22,8 +22,7 @@ UISplitViewControllerSplitBehavior SplitViewPreferredSplitBehaviorFromHostProp(
   }
 }
 
-UISplitViewControllerPrimaryEdge SplitViewPrimaryEdgeFromHostProp(
-    facebook::react::RNSSplitHostPrimaryEdge primaryEdge)
+UISplitViewControllerPrimaryEdge SplitViewPrimaryEdgeFromHostProp(facebook::react::RNSSplitHostPrimaryEdge primaryEdge)
```
kmichalikk: nit: again, not related to the PR

ios/tabs/host/RNSTabBarController.h
```diff
@@ -118,6 +118,30 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)updateOrientation;
 
+/**
+ * Updates the layout direction based on property on host view.
+ *
+ * This method does nothing if the update has not been previously requested.
+ * If needed, the requested update is performed immediately. If you do not need this, consider just raising an
+ * appropriate invalidation signal & let the controller decide when to flush the updates.
```
kmichalikk, line 126: what does "appropriate signal" mean here? could you be more specific?

ios/tabs/host/RNSTabBarController.h
```diff
@@ -118,6 +118,30 @@ NS_ASSUME_NONNULL_BEGIN
...
+ *
+ * This method is necessary only on iOS versions prior to 17.
+ * On iOS 17+, use `traitOverrides.layoutDirection` on the controller directly.
```
kmichalikk, line 129: if this is only for <= iOS 17 then maybe we can be more specific with naming the function? i.e. `setLayoutDirectionTraitIOS17`?

ios/tabs/host/RNSTabBarController.h
```diff
@@ -118,6 +118,30 @@ NS_ASSUME_NONNULL_BEGIN
...
+/**
+ * Updates the layout direction based on property on host view.
+ *
+ * The requested update is performed immediately. If you do not need this, consider just raising an appropriate
+ * invalidation signal & let the controller decide when to flush the updates.
+ *
+ * This method is necessary only on iOS versions prior to 17.
+ * On iOS 17+, use `traitOverrides.layoutDirection` on the controller directly.
+ *
+ * This method can only be called when `parentViewController` is not nil.
```
kmichalikk, line 142: same as above

ios/tabs/host/RNSTabBarController.h
```diff
@@ -161,6 +185,14 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, readwrite) bool needsOrientationUpdate;
 
+/**
+ * Tell the controller that some configuration regarding layout direction has changed & it requires update.
+ *
+ * This flag is necessary only on iOS versions prior to 17.
+ * On iOS 17+, use `traitOverrides.layoutDirection` on the controller directly.
+ */
+@property (nonatomic, readwrite) bool needsLayoutDirectionUpdate;
```
kmichalikk: same as above

src/components/helpers/PlatformUtils.ts
ahmedawaad1804: @kligarski sure

ios/tabs/host/RNSTabBarController.h
kligarski: It's the convention from props above.

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -254,6 +268,29 @@ - (void)updateOrientation
...
+    [self updateLayoutDirection];
+  }
+}
+
+- (void)updateLayoutDirection
+{
+  _needsLayoutDirectionUpdate = false;
+  if (@available(iOS 17.0, *)) {
+    // no-op
+  } else {
```
t0maboro: you can early return to avoid nesting

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -254,6 +268,29 @@ - (void)updateOrientation
...
+    RCTAssert(
+        self.parentViewController != nil,
+        @"[RNScreens] Expected non-null parent view controller for layout direction update.");
+    [self.parentViewController
+        setOverrideTraitCollection:[UITraitCollection
+                                       traitCollectionWithLayoutDirection:self.tabsHostComponentView.layoutDirection]
+            forChildViewController:self];
+  }
```
t0maboro: `RNS_IPHONE_OS_VERSION_AVAILABLE(17_0)` to exclude it from compilation when irrelevant?

kligarski: I applied it but to "if" above it - this is the correct place. In theory, we don't call this method in such situation but for consistency I guess that we should have it.

https://github.com/software-mansion/react-native-screens/pull/3613/commits/1e0cce55d4359e04b4c58d4f910d892d70efeba6

ios/tabs/host/RNSTabBarController.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3613/commits/1e0cce55d4359e04b4c58d4f910d892d70efeba6

android/src/paper/java/com/facebook/react/viewmanagers/RNSTabsHostManagerDelegate.java
kligarski: https://github.com/software-mansion/react-native-screens/pull/3613/commits/5b9bb8af30d09b9b1a79a07dfe36a048fabbfe10

common/cpp/react/renderer/components/rnscreens/RNSSplitScreenComponentDescriptor.h
kligarski: https://github.com/software-mansion/react-native-screens/pull/3613/commits/0624e7e482413d72f65f5140cacf107c4081577a

ios/conversion/RNSConversions-SplitView.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/3613/commits/0624e7e482413d72f65f5140cacf107c4081577a

ios/tabs/host/RNSTabBarController.h
kligarski: https://github.com/software-mansion/react-native-screens/pull/3613/commits/572fabe6e4eb182d4e72772f8990c925c1c3da9b

ios/tabs/host/RNSTabBarController.h
kligarski: https://github.com/software-mansion/react-native-screens/pull/3613/commits/572fabe6e4eb182d4e72772f8990c925c1c3da9b

ios/tabs/host/RNSTabBarController.h
kligarski: https://github.com/software-mansion/react-native-screens/pull/3613/commits/572fabe6e4eb182d4e72772f8990c925c1c3da9b

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -254,6 +268,32 @@ - (void)updateOrientation
...
+#endif // RNS_IPHONE_OS_VERSION_AVAILABLE(17_0)
+
+  RCTAssert(
+      self.parentViewController != nil,
+      @"[RNScreens] Expected non-null parent view controller for layout direction update.");
+  [self.parentViewController
+      setOverrideTraitCollection:[UITraitCollection
+                                     traitCollectionWithLayoutDirection:self.tabsHostComponentView.layoutDirection]
+          forChildViewController:self];
```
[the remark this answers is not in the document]
kligarski: `unspecified` is available on iOS 10.0+: https://developer.apple.com/documentation/uikit/uitraitenvironmentlayoutdirection/unspecified?language=objc so we should be good here.

ios/conversion/RNSConversions-Tabs.mm
```diff
@@ -484,4 +484,21 @@ UIUserInterfaceStyle UIUserInterfaceStyleFromTabsScreenCppEquivalent(
...
+      return UITraitEnvironmentLayoutDirectionUnspecified;
+    case Ltr:
+      return UITraitEnvironmentLayoutDirectionLeftToRight;
+    case Rtl:
+      return UITraitEnvironmentLayoutDirectionRightToLeft;
+    default:
+      RCTLogError(@"[RNScreens] unsupported layout direction");
+      break;
+  }
+}
```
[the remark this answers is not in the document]
kligarski: https://github.com/software-mansion/react-native-screens/pull/3613/commits/45cf252d52f2b695c14e75102a379c296756901f

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -34,6 +34,15 @@ - (instancetype)initWithTabsHostComponentView:(nullable RNSTabsHostComponentView
   return self;
 }
 
+#pragma mark - UIKit callbacks
+
+- (void)didMoveToParentViewController:(UIViewController *)parent
+{
```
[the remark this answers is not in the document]
kligarski: https://github.com/software-mansion/react-native-screens/pull/3613/commits/45cf252d52f2b695c14e75102a379c296756901f

PR #3617, opened 2026-02-04
fix(Android, FormSheet): Add logic for dismissing keyboard when the BottomSheet is presented

t0maboro:
## Description

Add logic to handle keyboard visibility and focus restoration on sheet enter/exit

> [!IMPORTANT]  
> It doesn't work well in some edge cases, especially when we have a Screen with a TextInput and a FormSheet with another TextInput. When using an HW keyboard in such a case, if we submit the Input inside the FormSheet (losing focus) and then want to reopen it, the focus search might select the TextInput on the Screen, as we do not block focus on the Screen behind the FormSheet.
> Creating ticket for that: https://github.com/software-mansion/react-native-screens-labs/issues/929

Implementation details:
- I'm checking the IME inset on the DecorView, there's 1 soft keyboard for the whole application - it's easier to detect whether it was open on the previous Screen
- I'm restoring the focus, only if the keyboard was visible on the previous screen - that might be bad for applications that are designed for keyboard/D-Pad navigation
- I don't want the Screen to be the subject of focus - there is no good reason for that imo, therefore, I've added FOCUS_AFTER_DESCENDANTS to focus on the 1st focusable descendant inside the FormSheet
- I'm using onCreate/Destroy callbacks, instead of onStart/Stop callbacks, because of the timing for autoFocus inside the FormSheet

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/928

## Changes

- added lifecycle callbacks for sheet creation/destruction that will show/hide keyboard from Screen and save the latest focus

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/bd07bc9d-ee8e-4c92-bea8-75117446bb96" /> | <video src="https://github.com/user-attachments/assets/5e8cd8e6-6da9-47d0-b138-48332087cc59" /> |

## Test plan

Test3617

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
@@ -135,8 +135,8 @@ class SheetDelegate(
         val activity = screen.reactContext.currentActivity ?: return
 
         activity.currentFocus?.let { focusedView ->
-            activity.window?.decorView.let { decorView ->
-                if (isSoftKeyboardVisibleOrNull(decorView!!) == true) {
+            activity.window?.decorView?.let { decorView ->
+                if (isSoftKeyboardVisibleOrNull(decorView) == true) {
                     viewToRestoreFocus = focusedView
                 }
             }


kkafar (approved):
Looks good now. Thank you.

PR #3620, opened 2026-02-04
feat(Android, Stack v5): prevent native dismiss support

kkafar:
## Description

**Add initial implementation of prevent native dismiss on Android**
This commit adds support for "prevent native dismiss" behavior on
Android for stack screens.

The added interface consists of prop on `StackScreen`:
`preventNativeDismiss` and event `onNativeDismissPrevented`, which
lets the programmer react to the native dismiss attempt.

This implementation follows [RFC-0717](https://github.com/software-mansion/react-native-screens-labs/blob/main-issue-tracker/rfcs/0717-stack-prevent-dismiss.md).
**It does not handle JS-side of the mechanism.**

Closes https://github.com/software-mansion/react-native-screens-labs/issues/881

## Changes

This is achieved by leveraging a couple of native mechanisms.
First, I lean into usage of `OnBackPressedCallback` AndroidX API, which
lets us handle this very gracefully & hopefully correct on many
SDK levels.

> [!note]
> TODO: The behavior on lower SDK levels has not been tested yet.

Now, for each `StackScreenFragment` we add a dedicated callback, which
is enabled or disabled depending on the value. I do not add/remove the
callbacks because their order in the `OnBackPressedDispatcher` matter,
and we can add a callback only to the end of the list.

> [!caution]
> This implementation assumes that the fragments will move to `created`
> state in the order they are in the fragment manager (stack order).
> If this ever turns out to not be the case, we'll need to modify this
> implementation.

Beside above ☝🏻 the added code also introduces "top-fragment" mechanism. 
It is necessary to prevent non-top-fragment from blocking native dismiss. 
Likely, when we introduce proper attach/detach mechanism for "non visible" fragments we will be able to get rid of this new mechanism, but currently it is necessary. Please see commit description: https://github.com/software-mansion/react-native-screens/pull/3620/commits/1212f36b3b3b9b078b619617bf3d39a222391aff

**Add missing "exported custom direct event type constants"**
I don't know how our events have worked so far. I thought that
overriding this method is obligatory. Apparently not. Maybe its
just hint for the React machinery. I'm not sure right now,
but I've verified that this code is indeed called & React gathers
this information.

## Visual documentation

https://github.com/user-attachments/assets/cfd8ab97-55c6-4d81-b3ec-1b50b4e46580

## Test plan

These do not test "dynamic" case. It requires a bit more setup in `StackContainer` code that is too much outside of the scope of this PR. I'll work on it separately. 

`prevent-native-dismiss-single-stack`
`prevent-native-dismiss-nested-stack`

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
@@ -179,9 +179,9 @@ internal class StackContainer(
         }
 
     private fun updateTopFragment() {
-        // This assumes that only our fragments are present in fragment manager
+        // We try to handle situation where other fragments might be present.
         val fragments = requireFragmentManager().fragments.filterIsInstance<StackScreenFragment>()
-        check(fragments.isNotEmpty()) { "[RNScreens] Empty fragment manager while attmpting to update top fragment" }
+        check(fragments.isNotEmpty()) { "[RNScreens] Empty fragment manager while attempting to update top fragment" }
         fragments.forEach { it.onResignTopFragment() }
         fragments.last().onBecomeTopFragment()
     }


kkafar:
This is how I use the test for now: 

```tsx
import React from 'react';
import { enableFreeze } from 'react-native-screens';
// import * as Test from './src/tests/issue-tests';
import { COMPONENT_SCENARIOS } from './src/tests/single-feature-tests';

enableFreeze(true);

  const Component = COMPONENT_SCENARIOS.StackHost[0].screen;
  return <Component />;
  // return <Test.TestStackNesting />;
}
```

Currently it is not reachable via Example app, because there are some bugs in the example app code I need to investigate. Will do it separately, its out of the scope of this PR

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
```diff
@@ -170,6 +178,14 @@ internal class StackContainer(
             Log.d(TAG, "Created Fragment $it for screen ${screen.screenKey}")
         }
 
+    private fun updateTopFragment() {
+        // This assumes that only our fragments are present in fragment manager
+        val fragments = requireFragmentManager().fragments.filterIsInstance<StackScreenFragment>()
```
t0maboro: This comment should be corrected. We're not assuming that only our fragments are present

kkafar: https://github.com/software-mansion/react-native-screens/pull/3620/commits/be084e8e3f1aafd0d7f6c004c6c30bce7c8acc91

PR #3622, opened 2026-02-05
chore(examples): refactor test structure

kkafar:
## Description

This PR fixes a minor bug, where previously I could not navigate to screens
for scenarios that had their `name !== key`.

**HOWEVER**, it's main role is to adjust our approach to test organization.
I've had now opportunities to work with the system introduced in <https://github.com/software-mansion/react-native-screens/pull/3528>
and I think we can introduce few improvements.

Fixes https://github.com/software-mansion/react-native-screens-labs/issues/931

## Changes

First, I stay for now with tests separation for "component integration tests"
and "single feature tests", however I change intra-group organization.

Instead of describing each test in `index` file, I find it more suitable to
place & export manifest from each test module. *In the future we can enrich
such manifests, e.g. with detailed test steps / expected outcome description
or list or related PRs*. Scenarios are grouped into scenario groups, which have
their own name & details that can be displayed in the test app.

Second, the single-feature-tests are split into component categories. They had been
already, but I think that it was too granular. It does not make sense to test
`stack-host` or `stack-screen` separately, as they are tightly coupled.

Third, I've introduced single convention for test naming. Tests should be named
with *kebab-case*, and the name should reflect the feature being tested.
The idea here is that I want to be able to find e.g. all orientation related tests
by filtering files with "orientation" in name. Example time: prefer verbose names
such as `tests/component-integration-tests/orientation/test-orientation-split-stack.tsx`
over `tests/component-integration-tests/orientation/split-stack.tsx`, despite the fact
that test scope is contained within directory name.

Beside scenario manifest as
default export, each test module exports `App` component, which is the root component
of the test. From now on this is a required convention.

I'll prepare a document summarizing this.

## Visual documentation

N/A

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/apps/src/tests/component-integration-tests/orientation/orientation-stack-in-tabs.tsx
+++ b/apps/src/tests/component-integration-tests/orientation/orientation-stack-in-tabs.tsx
@@ -12,6 +12,18 @@ import {
   createAutoConfiguredTabs,
   findTabScreenOptions,
 } from '../../shared/tabs';
+import { Scenario } from '../../shared/helpers';
+
+const SCENARIO: Scenario = {
+  name: 'StackInTabs',
+  details:
+    'Configuration in Stack contained within TabScreen always takes precedence',
+  key: 'cit-orientation-stack-in-tabs',
+  AppComponent: App,
+  platforms: ['ios'],
+};
+
+export default SCENARIO;
 
 type StackParamsList = {
   Screen1: undefined;
@@ -84,7 +96,7 @@ const Tabs = createAutoConfiguredTabs<TabsParamsList>({
   Tab2: DummyScreen,
 });
 
-export default function TabsAndStack() {
+export function App() {
   return (
     <Tabs.Provider>
       <Tabs.Autoconfig />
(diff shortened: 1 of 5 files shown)


apps/src/tests/component-integration-tests/orientation/orientation-stack-in-tabs.tsx
```diff
@@ -84,7 +96,7 @@ const Tabs = createAutoConfiguredTabs<TabsParamsList>({
   Tab2: DummyScreen,
 });
 
-export default function TabsAndStack() {
+export function App() {
```
t0maboro, line 99: unused export

apps/src/tests/component-integration-tests/orientation/orientation-tabs-in-stack.tsx
```diff
@@ -83,7 +95,7 @@ const Stack = createAutoConfiguredStack<StackParamList>({
   Screen1: TabsScreen,
 });
 
-export default function TabsAndStack() {
+export function Apps() {
```
t0maboro: unused export - applies to all below

t0maboro (approved):

apps/src/tests/component-integration-tests/orientation/orientation-stack-in-tabs.tsx
t0maboro: nevermind, aligned internally that we'd like to import example directly sometimes

PR #3625, opened 2026-02-06
chore(examples): allow to set route options in StackContainer

kkafar:
## Description

Add `SetRouteOptionsActionMethod` allowing to modify route options.

Closes <https://github.com/software-mansion/react-native-screens-labs/issues/936>

It only modifies options of the particular `StackRoute` indicated
by the passed `routeKey`.
Route configuration is an input to the `StackContainer` & it shouldn't
be changed from `StackContainer`. This is sometimes useful however, but
it should be introduced separately.

## Changes

I've added this to `StackContainer` "navigation methods". It does not fit well to the naming,
but I think that code requires a bit of cleanup / refactor anyway.

## Visual documentation

https://github.com/user-attachments/assets/babb79f0-f479-4f02-bb5a-5658c46edb4e

## Test plan

I've added option to dynamically configure `preventNativeDismiss` in `prevent-native-dismiss-single-stack`
& `prevent-native-dismiss-nested-stack`. This has been missing from that tests.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/apps/src/tests/single-feature-tests/stack-v5/prevent-native-dismiss-nested-stack.tsx
+++ b/apps/src/tests/single-feature-tests/stack-v5/prevent-native-dismiss-nested-stack.tsx
@@ -89,6 +89,7 @@ function AScreen() {
       <RouteInformation routeName="A" />
       <PreventNativeDismissInfo />
       <NavigationButtons isPopEnabled routeNames={['A', 'B', 'NestedStack']} />
+      <TogglePreventNativeDismiss />
     </CenteredLayoutView>
   );
 }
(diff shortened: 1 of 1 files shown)


apps/src/tests/single-feature-tests/stack-v5/prevent-native-dismiss-nested-stack.tsx
```diff
@@ -89,6 +89,7 @@ function AScreen() {
       <RouteInformation routeName="A" />
       <PreventNativeDismissInfo />
       <NavigationButtons isPopEnabled routeNames={['A', 'B', 'NestedStack']} />
+      <TogglePreventNativeDismiss />
     </CenteredLayoutView>
   );
 }
```
kligarski, line 95: We allow to toggle native dismiss on `AScreen` & `NestedAScreen` but if you do so, there is no toast as the callback is not configured for those routes - this might be confusing for somebody testing the behavior. I think that we should either add the callbacks to those routes or remove the toggle (and possibly info as well) from those screens.

Same applies to single stack version of this test.

kkafar: Good catch thank you. I'll fix that in #3626

PR #3626, opened 2026-02-07
chore(examples): allow modifying route blueprints passed to `StackContainer`

kkafar:
## Description  

**Expose `StackContainerWithDynamicRouteConfigs` component**
It's a wrapper for `StackContainer`. It adds `StackRouteConfigContext`,
which can be used to modify route options stored in route configurations.

This is different from what <https://github.com/software-mansion/react-native-screens/pull/3625>
introduced, because this does not impact existing route instances. Only blueprints.

I created separate component to keep the `StackContainer` logic thin.

### DIFF
--- a/apps/src/tests/single-feature-tests/stack-v5/prevent-native-dismiss-nested-stack.tsx
+++ b/apps/src/tests/single-feature-tests/stack-v5/prevent-native-dismiss-nested-stack.tsx
@@ -89,7 +89,6 @@ function AScreen() {
       <RouteInformation routeName="A" />
       <PreventNativeDismissInfo />
       <NavigationButtons isPopEnabled routeNames={['A', 'B', 'NestedStack']} />
-      <TogglePreventNativeDismiss />
     </CenteredLayoutView>
   );
 }
@@ -167,7 +166,6 @@ function NestedAScreen() {
       <RouteInformation routeName="NestedA" />
       <PreventNativeDismissInfo />
       <NavigationButtons isPopEnabled routeNames={['NestedA', 'NestedB']} />
-      <TogglePreventNativeDismiss />
     </CenteredLayoutView>
   );
 }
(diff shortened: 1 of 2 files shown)


apps/src/shared/gamma/containers/stack/contexts/StackRouteConfigContext.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
```
t0maboro, line 1: nit: can be `.ts` - no use of JSX syntax, can be considered as a follow-up refactor, because it refers to more files

apps/src/shared/gamma/containers/stack/StackContainer.tsx
```diff
@@ -95,7 +95,7 @@ function useSanitizeRouteConfigs(
   routeConfigs?: StackRouteConfig[] | undefined | null,
 ) {
   if (!routeConfigs || routeConfigs.length === 0) {
-    throw new Error('[RNScreens] There must be at least one route configured');
+    throw new Error('[Stack] There must be at least one route configured');
```
t0maboro, line 98: Q: Should we consider the same for Split https://github.com/software-mansion/react-native-screens/blob/main/src/components/gamma/split/SplitScreen.web.tsx#L3 ?

t0maboro (approved):

apps/src/shared/gamma/containers/stack/StackContainer.tsx
kkafar: Yes. My motivation here is that these are not library errors, but rather "user-space code". Therefore it's misleading even for us.

I'll handle this immediately. Don't want to create separate ticket for this when it can be done in a minute.

kkafar: Nooo. I spoke the truth, but above ☝🏻 does not apply to the code you linked exactly for the described reasons 😄 The code you linked is library code - and I think we should keep things consistent and use the `[RNScreens]` tag. 

The code I've edited is not shipped with the library. Hence the difference in approach.

PR #3629, opened 2026-02-10
feat(Android, Stack v5): make Transition API work

kkafar:
## Description

> [!note]
> To support predictive back gesture on Android, we need to either use `Animator API`
> or `Transition API` for screen animations. The [`Animation API` is not compatible with predictive back gesture](https://developer.android.com/guide/navigation/custom-back/support-animations#fragments).
> In stack v4, with small exception for formsheets we've heavily relied on `Animation API` and we can
> not reuse that code, unfortunately.

This PR intends to make Transition API work on Android for us. Currently, when
attempting to use it we have problems with:

1. disappearing shadows,
2. jumping content,
3. non-continuous animations (e.g. predictive back gesture),
4. native-pop missing animation.

| Most of the bugs described above |
| --- |
| <video src="https://github.com/user-attachments/assets/7f5cf2e8-5b6a-41ec-8dc4-43e8e40578a0" alt="before" /> |
| This video does not showcase issues no. 3 & 4, but trust be bro, they are there | 

This identifies root causes for each of those problems and introduces changes to
fix them.

It also adds an appropriate feature test and sets some Slide animation, so we
can observe anything when testing. This is not meant to be the default animation
or something. Just a placeholder animation.

## Changes

I encourage reading through particular commits, as I tried to give each of them
a meaningful message.

Here I list most important changes:

* CAUTION: bump native android library dependencies & add depencency on `transition`

    I've decided to bump appcompat, fragment deps.
    This is done mostly in hope to get rid of as many problems
    as possible with animations.

    Dependency on `transition` is added because `transition@1.5.0`
    is required for compatibility with predictive back gesture, so I want
    to have that version under control.

    `fragment` dependency is required to at least 1.7.0 by predictive back
    gesture.

    I've bumped other deps to ensure that they do not force lower version
    for `fragment` or `transition`.

    See: <https://developer.android.com/guide/navigation/custom-back/support-animations#use_existing_apis>

* Set `isTransitionGroup` for `StackScreen` to enable Transition API usage

    This is crucial. This assures that whole StackScreen subtree (including
    the StackScreen itself) is animated together. W/o this the animation is
    glitchy - some of the views can disappear, some views can jump to
    different location mid-transition, **shadows are lost**. Setting this
    option fixes things.

    I wonder what will be impact of using this on SET & Reanimated
    integration. Hopefully it'll be possible to animate some views anyway.

* Merge AddOp & SetPrimaryNavFragmentOp + add OnCommitCallbackFragmentOp

    This is kinda complex change.

    First, this commit removes `SetPrimaryNavFragmentOp` and merges it to
    `AddOp` creating new `AddAndSetAsPrimaryOp`. This is done, because
    previously, we've executed a `AddOp` that has been saved to back stack
    with e.g. some fragment A, and then we executed another transaction
    impacting fragment A by executing `SetPrimaryNavFragmentOp` & it has
    not been saved to back stack. This led to situation, where later popping
    A would not be interpreted as `pop` operation (explained below) by
    fragment manager and in consequence this would break predictive back
    gesture & native pop operation - after touch release animation would be
    completed immediately & not by smoothly animating to completion.

    Fragment manager interprets a batch of fragment transactions as "pop"
    only when last operation in a batch is "pop". Setting primary navigation
    fragment is not a "pop" operation. I haven't tested that in runtime, but
    by reading the code of fragment manager I've figured that we can not
    queue the `SetPrimaryNavFragmentOp` before other operations, because
    it'll trigger an exception that we try to make not-yet-attached fragment
    primary navigation fragment. However, maybe it's worth trying.
    Currently, I've gone with different approach. Now, I batch every add operation with
    set-primary-nav-fragment & add them to back stack together. *Thanks to that,
    we no longer manually manage the primary fragment. Fragment manager does
    that for us, when reverting stack transactions.*

    We still have a need to run a callback on transaction commit to update
    the top fragment (prevent native dismiss requirement), therefore
    exactly due to reasons explained above, I've added
    `OnCommitCallbackFragmentOp` and scheduled it before any other
    operations. It can not be batched with other operations, because a
    transaction with `runOnCommit` configured can NOT be added to back stack
    (callback can not be serialized).

* Fix transitions in nested stacks

    Component views on new architecture receive their first layout after the view   hierarchy is
assembled and attached to window. Note, that in case of screen views & their sub-trees
(including nested containers) this does not hold. The container is updated later, after React
mounting transaction ends, therefore the views are attached to the window much later, hence
their `isLaidOut` returns false, breaking transitions & animations.

    NOTE: Currently, screens attached even to the "top" (non-nested)
container return false from `isLaidOut`! Exactly due to mechanism
described above. I do not know whether it causes any problems, but
potentially - yes. We should be aware of that.

    The fix here is to lay out the container. I need to do that
synchronously, before we schedule the transitions. I'm not aware of
API that allows me to achieve that, so I've introduced
`StackContainerParent` interface.

## Visual documentation

https://github.com/user-attachments/assets/39069381-4a99-445d-98be-9c2fa7a2c608

## Checklist

* [x] Updated / created local changelog entries in relevant test files.
* [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
@@ -178,8 +178,6 @@ internal class StackContainer(
         require(stackModel.remove(fragment)) { "[RNScreens] onNativeFragmentPop must be called with the fragment present in stack model" }
         check(stackModel.isNotEmpty()) { "[RNScreens] Stack model should not be empty after a native pop" }
 
-        val fragmentManager = requireFragmentManager()
-
         // The primary navigation fragment should be updated when popping backstack by FragmentManager
         // reversing the back stack record. At this point we need to just update the top fragment.
         check(requireFragmentManager().primaryNavigationFragment !== fragment) {


android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/FragmentOperation.kt
```diff
@@ -57,14 +57,15 @@ internal class FlushNowOp : FragmentOperation() {
     }
 }
 
-internal class SetPrimaryNavFragmentOp(
-    val fragment: StackScreenFragment,
-    val onCommitCallback: Runnable? = null,
+internal class OnCommitCallbackFragmentOp(
```
kligarski:
```suggestion
internal class OnCommitCallbackOp(
```

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
```diff
@@ -200,8 +198,14 @@ internal class StackContainer(
...
-        if (pop) {
-            delegate.get()?.onScreenDismiss(fragment.stackScreen)
+
+        // This callback is called for every fragment involved in the back stack change, even
+        // if its not added or removed, but e.g. set as a primary navigation fragment, hence
+        // we need to check whether the fragment is actually being removed.
+        // I avoid using `pop` parameter here, because transaction might not be classified as `pop`
+        // and still include fragment removal operations.
+        if (fragment.isRemoving) {
+            delegate.get()?.onScreenDismissCommitted(fragment.stackScreen)
```
kligarski, line 247: One thing I noticed that is maybe unrelated to this PR is that in both `StackHost.onScreenDismissCommitted` and `StackScreen.onDismiss`, we're setting `isNativelyDismissed = true`. Both of those methods are called for native pop as well as JS pop. Should we move marking the screen as natively dismissed to only one of the methods? `StackHost.onScreenDismissCommitted` is called first and `StackScreen.onDismiss` seems to be called at the end of transition.

kkafar: Yes, I've also seen that. Yep, I'll try to look into this.

kkafar:
@kligarski see https://github.com/software-mansion/react-native-screens/pull/3629/commits/e878abb8b279b12cdd37ff5b9355a5a0304128c3 - this should solve problems with nested stack.

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/FragmentOperation.kt
kkafar: https://github.com/software-mansion/react-native-screens/pull/3629/commits/aef44f813551a9201d5696ec59c252bb74527e01

kkafar:
When testing I've noticed regression in prevent native dismiss behaviour. This still needs some work then.

kkafar:
Ready

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
```diff
@@ -133,37 +170,22 @@ internal class StackContainer(
...
-            listOf(
-                SetPrimaryNavFragmentOp(stackModel.last(), {
-                    updateTopFragment()
-                }),
-            ),
-        )
+
+        // The primary navigation fragment should be updated when popping backstack by FragmentManager
+        // reversing the back stack record. At this point we need to just update the top fragment.
+        check(requireFragmentManager().primaryNavigationFragment !== fragment) {
```
kligarski, line 183:
```suggestion
        check(fragmentManager.primaryNavigationFragment !== fragment) {
```

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
kligarski: Maybe we should add a ticket for this?

kligarski (approved):
The only thing left is the animation on the root screen - we can create a ticket for this if we don't want to handle this in this PR.

Other than that it looks good.

kkafar:
I do not wait for iOS CI - seems that we've run out of minutes 😅

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
kkafar: https://github.com/software-mansion/react-native-screens/pull/3629/commits/ef9a80fec43b1df3af2ef21cef3ac46702720aa3

android/src/main/java/com/swmansion/rnscreens/gamma/stack/host/StackContainer.kt
kkafar: https://github.com/software-mansion/react-native-screens-labs/issues/977

kkafar:
Ticket for animation on the first screen:

* https://github.com/software-mansion/react-native-screens-labs/issues/976

Ticket for looking into when `isNativelyDismissed` is called:

* https://github.com/software-mansion/react-native-screens-labs/issues/977

PR #3630, opened 2026-02-10
chore(examples): add patch to `MainActivity` to enable predictive back gesture

kkafar:
## Description

Add patch to MainActivity to enable predictive back gesture

This is needed, because for some reason, currently the
`OnBackPressedCallback` from `ReactActivity` is added after the one from
fragment manager, effectively preventing native dismiss & navigation
from working properly.

This should be either fixed directly on RN side or we'll need to make it
into an installation step / alternatively we can create a hacky
workaround & handle this on our end.

This is added only to FabricExample because we do plan to support
predictive back gesture only on new architecture with new stack
implementation.

## Changes

This patch uses reflection to disable the `ReactAcctivity`'s `OnBackPressedCallback`.

It effectively disables `BackPressed` API from `react-native`. We should be
fine with it for now.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/FabricExample/android/app/src/main/java/com/fabricexample/MainActivity.kt
+++ b/FabricExample/android/app/src/main/java/com/fabricexample/MainActivity.kt
@@ -26,6 +26,10 @@ class MainActivity : ReactActivity() {
     supportFragmentManager.fragmentFactory = RNScreensFragmentFactory()
     super.onCreate(savedInstanceState)
 
+    // Remove this once we have sensible workaround to disable the react callback.
+    // Currently it prevents fragment manager's callback from triggering, blocking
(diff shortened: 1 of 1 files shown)


kligarski:
I wonder whether we should wait with this patch for the 5.x branch cut because now Stack v4 examples are broken.

https://github.com/user-attachments/assets/5241cab0-ade3-4a64-8fc8-aae51abcd49c

kkafar:
Duck.

kkafar:
Good catch. I'll revert it & leave it as a comment, so that it is easily "turnable" by people working on new stack.

kkafar:
Changed my mind. We'll proceed in different direction, but we'll fix this.

PR #3638, opened 2026-02-12
fix(android): Fix Activity memory leak in ScreenDummyLayoutHelper due to unreleased view references and missing lifecycle cleanup

### LINKED ISSUE #3636: [Android] ScreenDummyLayoutHelper leaks Activity due to unreleased view references and missing lifecycle cleanup
### Description

`ScreenDummyLayoutHelper` causes an Activity memory leak on Android because it holds strong references to dummy layout views (`CoordinatorLayout`, `AppBarLayout`, `Toolbar`, `View`) that are created with an Activity-based `Context`, but these references are never released when the Activity is destroyed.

The root cause is a combination of flawed lifecycle management:

1. `lateinit var` prevents nullability — The view fields use `lateinit var`, making it impossible to set them to `null` for cleanup.
2. Lifecycle listener is conditionally registered — In the constructor, `addLifecycleEventListener` is only called if `maybeInitDummyLayoutWithHeader` fails. If it succeeds (the common case), `onHostDestroy` is never invoked.
3. Listener is removed prematurely in `onHostResume` — After successful initialization, `removeLifecycleEventListener` is called, preventing `onHostDestroy` from firing on subsequent Activity destruction.
4. `onHostDestroy` does not release view references — Even when called, it only removes the lifecycle listener and does not null out the view fields.

This results in a reference chain: `ScreenDummyLayoutHelper` → `Views` → `ContextThemeWrapper` → `Activity`, preventing the destroyed Activity (and its entire view hierarchy, resources, and window) from being garbage collected.

Expected behavior: When the host Activity is destroyed, all view references in `ScreenDummyLayoutHelper` should be released so the Activity can be garbage collected. On the next `onHostResume`, the helper should re-initialize with the new Activity's context.

### Steps to reproduce

1. Create a React Native app with react-native-screens using native stack navigation.
2. Navigate through several screens to ensure ScreenDummyLayoutHelper is initialized (it measures header heights via dummy layout).
3. Trigger an Activity recreation — this can happen via:
    - React Native dev reload (Fast Refresh / full reload)
    - Android configuration change (e.g., locale change, dark mode toggle) if Activity is recreated
    - Process death and restoration by the OS
4. Use Android Studio Memory Profiler or LeakCanary to inspect retained objects.
5. Observe that the previous Activity instance is retained in memory, held by ScreenDummyLayoutHelper's view fields (coordinatorLayout, appBarLayout, toolbar, dummyContentView).
6. Repeat steps 3-5 to see the leak accumulate with each Activity recreation.

(Will provide a minimal reproduction repository)

### Screens version

4.23.0

### React Native version

0.84.0

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Release mode

### Device

Real device

### Device model

Multiple Android devices (Android 12+)

### Acknowledgements

Yes

l2hyunwoo:
## Description

Fix Activity memory leak in `ScreenDummyLayoutHelper` caused by unreleased view references (`CoordinatorLayout`, `AppBarLayout`, `Toolbar`, `View`) when the host Activity is destroyed.

The helper created dummy views using Activity-scoped `Context` to measure header heights, but never released these references on Activity destruction. This created a strong reference chain (`ScreenDummyLayoutHelper → Views → ContextThemeWrapper → Activity`) that prevents GC of destroyed Activities.

Closes: #3636 

## Changes

- Changed view fields from `lateinit var` to nullable `var? = null` to allow reference release
- Always register `LifecycleEventListener` in constructor (regardless of init success)
- Keep listener registered across Activity lifecycle (never remove it) to ensure `onHostDestroy` is always called
- Release all view references, reset cache, and reset `isLayoutInitialized` in `onHostDestroy` to break the reference chain and allow re-initialization on next `onHostResume`
- Wrap `computeDummyLayout` in `synchronized(this)` block with null-safe local variables, returning `0.0f` as fallback when views are already cleaned up
- Synchronize `onHostDestroy` cleanup on the same monitor to prevent race conditions

### Before

https://github.com/user-attachments/assets/b22c935d-9e22-4ddb-92f5-91be5caee430

### After

https://github.com/user-attachments/assets/c5946a30-3537-46a0-a15d-33183823294b

## Test plan

1. Launch a React Native app with native stack navigation using `react-native-screens`
2. Navigate through screens to trigger `ScreenDummyLayoutHelper` initialization
3. Trigger Activity recreation (dev reload, configuration change, or process restoration)
4. Use Android Studio Memory Profiler or LeakCanary to verify the previous Activity is no longer retained
5. Repeat to confirm no leak accumulation across multiple Activity recreations
6. Verify header height computation still works correctly after Activity recreation (headers render with correct dimensions)

## Checklist

- [X] Included code example that can be used to test this change.
- [X] Updated / created local changelog entries in relevant test files.
- [ ] Ensured that CI passes


t0maboro:
Hi @l2hyunwoo , 
Thanks for submitting this PR - I can confirm that it resolves the memory leak issue in the `ScreenDummyLayoutHelper`.

I hope you don't mind if I add a few additional changes to this PR to clean things up a bit. Also, based on the research I've done, I'm wondering if it might be better to rely on `ActivityLifecycleCallbacks` to listen for activity destruction events directly from the OS, rather than depending on React Native. Let me follow up with this PR.

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
```diff
@@ -269,6 +315,23 @@ internal class ScreenDummyLayoutHelper(
...
+        dummyContentView = null
+        toolbar = null
+
+        isLayoutInitialized = false
+
+        val activity = reactContextRef.get()?.currentActivity
+        activityLifecycleCallbacks?.let {
+            activity?.application?.unregisterActivityLifecycleCallbacks(it)
+        }
+        activityLifecycleCallbacks = null
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3638/changes/4c24fb9752e495d936073bf3f850c0d0e4f14950

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
```diff
@@ -178,6 +187,10 @@ internal class ScreenDummyLayoutHelper(
             return cache.headerHeight
         }
 
+        val currentCoordinatorLayout = coordinatorLayout ?: return 0.0f
+        val currentAppBarLayout = appBarLayout ?: return 0.0f
+        val currentToolbar = toolbar ?: return 0.0f
+
         val topLevelDecorView = requireActivity().window.decorView
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3638/changes/4c24fb9752e495d936073bf3f850c0d0e4f14950

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
```diff
@@ -192,31 +205,64 @@ internal class ScreenDummyLayoutHelper(
...
+
+                override fun onActivitySaveInstanceState(
+                    activity: Activity,
+                    outState: Bundle,
+                ) {}
+            }
+
+        activity.application.registerActivityLifecycleCallbacks(activityLifecycleCallbacks)
+    }
+
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3638/changes/4c24fb9752e495d936073bf3f850c0d0e4f14950

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
```diff
@@ -269,6 +317,24 @@ internal class ScreenDummyLayoutHelper(
...
+        coordinatorLayout = null
+        appBarLayout = null
+        dummyContentView = null
+        toolbar = null
+
+        isLayoutInitialized = false
+
+        val activity = reactContextRef.get()?.currentActivity
+        activityLifecycleCallbacks?.let {
+            activity?.application?.unregisterActivityLifecycleCallbacks(it)
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3638/changes/848a1a6ed49fc81faa17f22bb1193c1aa21eab0d

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
```diff
@@ -269,6 +317,23 @@ internal class ScreenDummyLayoutHelper(
...
+        cleanUpViews()
+    }
+
+    @Synchronized
+    private fun cleanUpViews(application: Application? = reactContextRef.get()?.currentActivity?.application) {
+        coordinatorLayout = null
+        appBarLayout = null
+        dummyContentView = null
+        toolbar = null
+
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3638/commits/c26813419d8f3f06a81c84190a6751b9db741702

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
```diff
@@ -269,6 +317,23 @@ internal class ScreenDummyLayoutHelper(
...
+        isLayoutInitialized = false
+
+        activityLifecycleCallbacks?.let {
+            application?.unregisterActivityLifecycleCallbacks(it)
+        }
+        activityLifecycleCallbacks = null
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3638/commits/c26813419d8f3f06a81c84190a6751b9db741702

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
```diff
@@ -53,10 +58,11 @@ internal class ScreenDummyLayoutHelper(
...
         weakInstance = WeakReference(this)
-
-        if (!(reactContext.hasCurrentActivity() && maybeInitDummyLayoutWithHeader(reactContext))) {
-            reactContext.addLifecycleEventListener(this)
-        }
+        maybeInitDummyLayoutWithHeader(reactContext)
+        // We're adding lifecycleEventListener to have a chance to call onHostDestroy always,
+        // no matter whether activity was living or not - this ensures that onHostDestroy will
+        // consistently clean up resources
+        reactContext.addLifecycleEventListener(this)
```
kligarski: I don't understand this - we add ourselves as a listener, then we get `onHostResume` and we immediately unregister. This way we won't ever get `onHostDestroy`, right? Is this intended?

In original version of this PR, [`onHostResume`](https://github.com/software-mansion/react-native-screens/pull/3638/changes/e5013bdd45125b34d85c85a89e55e984c977a929#diff-5c015deb8f96fa6d5c713cecd9743c910fbe5850429e813cfcef6d22b6a9e8f4R265) was changed but now it was rolled back - is this intentional?

But I guess that the intended behavior here prior to this PR was to make sure we call `maybeInitDummyLayoutWithHeader` when we have the activity available and that was guaranteed when `onHostResume` is called. Then we clear the callback and `onHostDestroy` was meant to be just cleanup of the callback as well in some edge case scenario (but not for view cleanup done in this PR in activity callback).

So maybe the current solution is valid but we should change the comment here?

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
```diff
@@ -192,31 +207,64 @@ internal class ScreenDummyLayoutHelper(
...
         // Include the top inset to account for the extra padding manually applied to the CustomToolbar.
-        val totalAppBarLayoutHeight = appBarLayout.height.toFloat() + topInset
+        val totalAppBarLayoutHeight = currentAppBarLayout.height.toFloat() + topInset
 
         val headerHeight = PixelUtil.toDIPFromPixel(totalAppBarLayoutHeight)
         cache = CacheEntry(CacheKey(fontSize, isTitleEmpty), headerHeight)
         return headerHeight
     }
 
+    private fun registerActivityLifecycleListener(activity: Activity) {
```
kligarski, line 256: I guess that we can rely on activity callbacks also in brownfield apps, right?

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3638/commits/1b917bf001e94162a3736782a73fb5f5e5ee712e

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3638/changes/b9243ede3e9dec635d3ced20cda16036bb828e3a

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
t0maboro: Note: In general, I think that the native callback is more robust regarding the Activity. I haven't tested how it will behave in the brownfield setup, but at 1st thought I'd say that RN might not be the only one managing the activity lifecycle there

kligarski (approved):
Looks good to me but I'd wait for @kkafar's review before merging.

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
```diff
@@ -125,11 +135,12 @@ internal class ScreenDummyLayoutHelper(
                         ).apply { scrollFlags = 0 }
             }
 
+        // We know that the layout hierarchy parts (appBarLayout, toolbar, coordinatorLayout) are non-null at this point.
         // We know the title text view will be there, cause we've just set title.
-        defaultFontSize = ScreenStackHeaderConfig.findTitleTextViewInToolbar(toolbar)!!.textSize
-        defaultContentInsetStartWithNavigation = toolbar.contentInsetStartWithNavigation
+        defaultFontSize = ScreenStackHeaderConfig.findTitleTextViewInToolbar(toolbar!!)!!.textSize
+        defaultContentInsetStartWithNavigation = toolbar!!.contentInsetStartWithNavigation
```
kkafar: Let's add here checkNonNull + appropriate error message

kkafar: This comment applies to all similar places, even to some old code

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
```diff
@@ -178,7 +190,12 @@ internal class ScreenDummyLayoutHelper(
             return cache.headerHeight
         }
 
-        val topLevelDecorView = requireActivity().window.decorView
+        val currentCoordinatorLayout = coordinatorLayout ?: return 0.0f
+        val currentAppBarLayout = appBarLayout ?: return 0.0f
+        val currentToolbar = toolbar ?: return 0.0f
+        val currentActivity = reactContextRef.get()?.currentActivity ?: return 0.0f
```
kkafar: These can not be independently nullish, right? If any of these views is nullish then all of them are. Let's structure the condition that way, it makes the intention much clearer. Reading this code I wondered for a bit, how can these be nullish independently.

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
```diff
@@ -192,42 +209,73 @@ internal class ScreenDummyLayoutHelper(
...
+                override fun onActivityCreated(
+                    activity: Activity,
+                    savedInstanceState: Bundle?,
+                ) {}
+
+                override fun onActivityStarted(activity: Activity) {}
+
+                override fun onActivityResumed(activity: Activity) {}
+
+                override fun onActivityPaused(activity: Activity) {}
```
kkafar:
```suggestion
                override fun onActivityPaused(activity: Activity) = Unit
```

For all of these. Just a convention we use across the Android code.

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
```diff
@@ -156,6 +167,7 @@ internal class ScreenDummyLayoutHelper(
      * @return header height in dp as consumed by Yoga
      */
     @DoNotStrip
+    @Synchronized
```
kkafar, line 182: I do not understand this change. Why does this function need to be *synchronized*?

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
```diff
@@ -270,6 +318,24 @@ internal class ScreenDummyLayoutHelper(
     override fun onHostDestroy() {
         reactContextRef.get()?.removeLifecycleEventListener(this)
     }
+
+    @Synchronized
+    private fun cleanUpViews(application: Application) {
```
kkafar, line 341: I also don't understand this change yet. Why does this function need to be synchronized? What are the concurrent code paths that can lead here? We're missing comments with explanations here.

kkafar (commented):
Hey, thanks for the contribution. I've left few remarks & question. Let's answer them before landing these changes.

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
t0maboro: I'm observing that computeDummyLayout is called from the bg thread, while I'm calling cleanUpViews from onActivityDestroyed, which will be called on the main thread - both are using the same attributes for initializing and destroying them

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3638/changes/b742a9dd38b52c0d8e053ac7481b01e0b58e1ca4

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3638/changes/ab32c2df90fc355afc5d09cae1a8f215ec076f10

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3638/changes/30151c6c2704b7435be71a5cb2b6d2cd8d88783d

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3638/changes/08920437f85c3a9df1060a44282a5befd5d0e2ac

android/src/main/java/com/swmansion/rnscreens/utils/ScreenDummyLayoutHelper.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3638/changes/08920437f85c3a9df1060a44282a5befd5d0e2ac

kkafar (approved):
Looks ok. Let's roll this out and get some feedback.

efstathiosntonas:
thank you guys for working hard on this one!

@kkafar is there an ETA for this? (I’m right on one step before rolling out a big release so sorry for the question 😅)

mduleone:
Thank you all for this! We just discovered this issue in our app and stumbled upon the PR. Delighted to see that it's merged already! We're still on RN 0.81.5 and `react-native-screens` 4.16.0. What are the chances that this will be patched into a version that still supports RN 0.81?

kkafar:
The 4.25.0 release waits for us being ready with Tabs API (stabilisation). I plan to release it ASAP, but first planned release date was ~ 1 month ago -> I can only say: soon ™️ 
In meantime, *nightly* is available.

@mduleone is using a patch a feasible option for you? I don't plan to backport this issue  to any older version. Seemingly it has been there (the bug) for a long long time (unfortunately), and it is no regression introduced in any recent version.

efstathiosntonas:
@kkafar thank you for the update. There's something off with the nightlies though? I see the latest one was submitted on npm 10 days ago, April 1st. 

`4.25.0-nightly-20260401-0636b922d`

edit: yup, all failing: https://github.com/software-mansion/react-native-screens/actions/workflows/npm-screens-publish-nightly.yml

kkafar:
Indeed, I haven't noticed it during last week. Thanks. We'll sort it out on Monday.

PR #3639, opened 2026-02-12
feat(split-view): add show(column) command and topColumnForCollapsing prop

Ubax:
## Description

This will allow split view navigation when it is collapsed on iPhone.

Expose `UISplitViewController.show(_:)` to JS via an imperative ref command and add a `topColumnForCollapsing` prop that controls which column is displayed when the split view collapses.

| topColumnForCollapsing | show command |
| -- | -- |
| <video src="https://github.com/user-attachments/assets/112acfd6-0948-450f-b1fd-2942fe42189f" alt="topColumnForCollapsing" /> | <video src="https://github.com/user-attachments/assets/7de0881e-55cd-469f-b527-952d06e32446" alt="show-command" /> | 

## Test plan

Added `test-command-show-column` & `test-top-column-for-collapsing`.
See videos ☝🏻

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/apps/src/tests/single-feature-tests/split/index.ts
+++ b/apps/src/tests/single-feature-tests/split/index.ts
@@ -1,9 +1,11 @@
 import { ScenarioGroup } from '../../shared/helpers';
+import TestTopColumnForCollapsing from './test-top-column-for-collapsing';
+import TestCommandShowColumn from './test-command-show-column';
 
 const SplitScenarioGroup: ScenarioGroup = {
   name: 'Split',
   details: 'Single feature tests for Split',
-  scenarios: [],
+  scenarios: [TestTopColumnForCollapsing, TestCommandShowColumn],
 };
 
 export default SplitScenarioGroup;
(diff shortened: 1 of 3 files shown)


src/components/gamma/split/SplitHost.tsx
```diff
@@ -38,8 +43,28 @@ const isValidDisplayModeForSplitBehavior = (
 /**
  * EXPERIMENTAL API, MIGHT CHANGE W/O ANY NOTICE
  */
-function SplitHost(props: SplitHostProps) {
+function SplitHost(
+  props: SplitHostProps,
+  forwardedRef: React.Ref<SplitHostCommands>,
```
kkafar: Could we pass the ref via props leveraging React 19 capabilities?

src/components/gamma/split/SplitHost.types.ts
```diff
@@ -114,6 +114,12 @@ export interface SplitColumnMetrics {
    */
   preferredInspectorColumnWidthOrFraction?: number;
 }
+export type SplitColumn = 'primary' | 'supplementary' | 'secondary';
```
kkafar: Let's think of better name for this, not suggesting that these are all possible column types. Something that indicates purpose of this type.

PR #3645, opened 2026-02-13
feat: add imageSource and templateSource icon support to menus

Ubax:
## Description

Extend menu actions and submenus to support all icon types (imageSource, templateSource) matching the top-level bar button item API. This enables using custom React Native images (require('./icon.png')) as menu icons.

This is a follow-up PR to https://github.com/software-mansion/react-native-screens/pull/3643

https://github.com/user-attachments/assets/49ed9d3b-4198-43b3-a0bb-b1421def9133

## Test plan

BarButtonItems screen in Example.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSBarButtonItem.mm
+++ b/ios/RNSBarButtonItem.mm
@@ -21,28 +21,13 @@ static UIMenuOptions RNSMakeUIMenuOptionsFromConfig(NSDictionary *config);
     return self;
   }
 
+  [[self class] resolveImageFromConfig:dict
+                           imageLoader:imageLoader
+                       completionBlock:^(UIImage *img) {
+                         self.image = img;
+                       }];
+
   NSString *title = dict[@"title"];
-  NSDictionary *imageSourceObj = dict[@"imageSource"];
-  NSDictionary *templateSourceObj = dict[@"templateSource"];
-  NSString *sfSymbolName = dict[@"sfSymbolName"];
-  NSString *xcassetName = dict[@"xcassetName"];
-
-  if (imageSourceObj != nil || templateSourceObj != nil) {
-    BOOL isTemplate = imageSourceObj != nil ? NO : YES;
-    NSDictionary *source = imageSourceObj != nil ? imageSourceObj : templateSourceObj;
-
-    [RNSImageLoadingHelper loadImageSyncIfPossibleFromJsonSource:source
-                                                 withImageLoader:imageLoader
-                                                      asTemplate:isTemplate
-                                                 completionBlock:^(UIImage *image) {
-                                                   self.image = image;
-                                                 }];
-  } else if (sfSymbolName != nil) {
-    self.image = [UIImage systemImageNamed:sfSymbolName];
-  } else if (xcassetName != nil) {
-    self.image = [UIImage imageNamed:xcassetName];
-  }
-
   if (title != nil) {
     self.title = title;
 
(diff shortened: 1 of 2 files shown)


src/components/tabs/TabsScreen.tsx
```diff
@@ -306,9 +306,19 @@ function parseIOSIconToNativeProps(icon: PlatformIconIOS | undefined): {
...
+    return {
+      iconType: 'xcassetTinted',
+      iconResourceName: icon.name,
+    };
+  } else if (icon.type === 'xcassetOriginal') {
+    return {
+      iconType: 'xcassetOriginal',
+      iconResourceName: icon.name,
+    };
   } else {
```
satya164, line 309: The image loader should already support xcasset (`{ uri: 'name' }`) - and `imageSource`/`templateSource` support original/tinted - similar to `Image` from React Native. So what's the advantage of a separate `xcasset`/`xcassetTinted`/`xcassetOriginal` type?

Ubax: Good point!

I didn't realize it is supported in the screens version as well. Thanks for pointing this out. 

I'll close the tab changes PR and only leave `imageSource and templateSource icon support in menus` open.

Ubax: @kkafar this means that `xcasset` may not be needed as well

kkafar: thanks. I've added this to our todo list to get rid of drawableResource / xcasset types if necessary.

ios/RNSBarButtonItem.mm
```diff
@@ -202,6 +221,19 @@ + (UIAction *)createActionItemFromConfig:(NSDictionary *)dict menuAction:(RNSBar
...
+  if (image == nil && (imageSourceObj != nil || templateSourceObj != nil)) {
+    BOOL isTemplate = imageSourceObj != nil ? NO : YES;
+    NSDictionary *source = imageSourceObj != nil ? imageSourceObj : templateSourceObj;
+    [RNSImageLoadingHelper loadImageSyncIfPossibleFromJsonSource:source
+                                                 withImageLoader:imageLoader
+                                                      asTemplate:isTemplate
+                                                 completionBlock:^(UIImage *img) {
+                                                   actionElement.image = img;
+                                                 }];
+  }
```
kkafar: This is repeated twice highly conditional & hard to follow code. I'm not willing to land this in current shape. Likely we want to extract this logic & share it together with what's already in bottom tabs implementation.

src/components/helpers/prepareHeaderBarButtonItems.ts
```diff
@@ -18,18 +18,32 @@ const prepareMenu = (
       const xcassetName =
         iconType === 'xcasset' ? menuItem.icon?.name : undefined;
 
+      let imageSource, templateSource;
+      if (iconType === 'imageSource' && menuItem.icon?.type === 'imageSource') {
```
kkafar: What is going on here? From what I can see few lines above:

`const iconType = menuItem.icon?.type`, therefore this condition effectively checks the same twice. Am I right here?

Ubax: Yes, sorry for the overlook

src/components/helpers/prepareHeaderBarButtonItems.ts
```diff
@@ -18,18 +18,29 @@ const prepareMenu = (
       const xcassetName =
         iconType === 'xcasset' ? menuItem.icon?.name : undefined;
 
+      let imageSource, templateSource;
+      if (menuItem.icon?.type === 'imageSource') {
```
Ubax, line 22: I'm using `menuItem.icon?.type` instead of `iconType` to ensure types correctness within the if condition

kkafar (approved):
Thanks!

Issue #3549, opened 2026-01-20
[iOS] Build fails on Xcode 26.2: constexpr with Objective-C literals in RNSScreenStackHeaderConfig.mm


awais-wego:
## Description

iOS builds fail on **Xcode 26.2** (and likely Xcode 26.x) with a compilation error in `RNSScreenStackHeaderConfig.mm`. The error occurs because `constexpr` cannot be used with Objective-C literals, which Xcode 26's stricter C++ compiler now enforces.

## Environment

| Package | Version |
|---------|---------|
| react-native-screens | 4.17.0 - 4.18.0 |
| React Native | 0.77.0 |
| Xcode | 26.2 |
| iOS Deployment Target | 15.1+ |
| Platform | iOS |

**Note:** This issue does NOT affect versions 4.19.0+, which have a different (codegen-related) issue.

## Error

❌ /path/to/node_modules/react-native-screens/ios/RNSScreenStackHeaderConfig.mm:42:23:
Constexpr variable 'DEFAULT_TITLE_FONT_SIZE' must be initialized by a constant expression
static constexpr auto DEFAULT_TITLE_FONT_SIZE = @17;

**Full compilation output:**

CompileC /Users/.../RNSScreenStackHeaderConfig.o
/Users/.../node_modules/react-native-screens/ios/RNSScreenStackHeaderConfig.mm
normal arm64 objective-c++
com.apple.compilers.llvm.clang.1_0.compiler (in target 'RNScreens' from project 'Pods')

## Root Cause

In `ios/RNSScreenStackHeaderConfig.mm` (lines 42-43):

```
static constexpr auto DEFAULT_TITLE_FONT_SIZE = @17;
static constexpr auto DEFAULT_TITLE_LARGE_FONT_SIZE = @34;
```

Problem:
Objective-C literals like @17 and @34 are runtime NSNumber objects
C++ constexpr requires compile-time constant expressions
Xcode 26 enforces this strictly (earlier Xcode versions were lenient)

Proposed Solution
Change the declarations to use standard Objective-C constant pointers:

```
static NSNumber *const DEFAULT_TITLE_FONT_SIZE = @17;
static NSNumber *const DEFAULT_TITLE_LARGE_FONT_SIZE = @34;
```

Why this works:
static NSNumber *const = "constant pointer to NSNumber object"
This is the standard Objective-C pattern for compile-time constant pointers to runtime objects
No runtime behavior changes (same values, same usage)
Reproduction Steps
[...]

PR #3646, opened 2026-02-13
chore: Change constexpr to obj-c NSNumber

kmichalikk:
Fixes https://github.com/software-mansion/react-native-screens/issues/3549

## Description

This PR replaces constexpr declaration with equivalent NSNumber declaration for default font sizes. We weren't able to reproduce the compilation error, but the change doesn't have a negative impact.

## Changes

Replaced the default font size declarations.

## Test plan

Have a header (regular, large) with some non-default font family (this triggers the font size check). Set a breakpoint in `RNSScreenStackHeaderConfig.mm:491` and `:519` and verify the constants are correct.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -38,8 +38,8 @@
 namespace react = facebook::react;
 #endif // RCT_NEW_ARCH_ENABLED
 
-static constexpr auto DEFAULT_TITLE_FONT_SIZE = @17;
-static constexpr auto DEFAULT_TITLE_LARGE_FONT_SIZE = @34;
+static NSNumber *const DEFAULT_TITLE_FONT_SIZE = @17;
+static NSNumber *const DEFAULT_TITLE_LARGE_FONT_SIZE = @34;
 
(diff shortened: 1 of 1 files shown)


kligarski (approved):

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -38,8 +38,8 @@
 namespace react = facebook::react;
 #endif // RCT_NEW_ARCH_ENABLED
 
-static constexpr auto DEFAULT_TITLE_FONT_SIZE = @17;
-static constexpr auto DEFAULT_TITLE_LARGE_FONT_SIZE = @34;
+static NSNumber *const DEFAULT_TITLE_FONT_SIZE = @17;
+static NSNumber *const DEFAULT_TITLE_LARGE_FONT_SIZE = @34;
```
kkafar:
```suggestion
static const NSNumber * const DEFAULT_TITLE_FONT_SIZE = @17;
static const NSNumber * const DEFAULT_TITLE_LARGE_FONT_SIZE = @34;
```

const pointer to a const value. Now you have only const pointer to a mutable value.

PR #3633, opened 2026-02-11
feat(Android, Tabs): Add support for selectedIcon on Android

t0maboro:
## Description

This PR adds support for the `selectedIcon` prop on Android. I'm not sure why it was omitted initially, but it's possible to implement it by checking whether the tab is focused. 

> [!CAUTION]
> The type for `selectedIcon` was intentionally changed from `PlatformIconIOS` to `PlatformIcon`. Since tabs are still an experimental component, we want to avoid additional maintenance overhead. This change will have a compatibility flag, which will be added after shipping all PRs related to the refactor.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/954 (let me split the original issue 773 from labs to 2 separate sub-issues, as they touch completely different mechanisms)

## Changes

- Added props for native component on android.
- Extended type definition for external API
- Updated example

## Before & after - visual documentation

(No before section, new API)

https://github.com/user-attachments/assets/11b71555-1749-41b3-ace7-14b58e516200

## Test plan

I've updated TestBottomTabs to use selectedIcon of type PlatformIcon. All other examples using that prop were intentionally left unchanged for regression testing.

## Checklist

- [x] Included code example that can be used to test this change.
- [ ] Updated / created local changelog entries in relevant test files.
- [x] For visual changes, included screenshots / GIFs / recordings documenting the change.
- [x] For API changes, updated relevant public types.
- [ ] Ensured that CI passes

PR #3649, opened 2026-02-14
feat(android): Add `selectedIcon` support for bottom tabs

Kasendwa:
`selectedIcon` works on iOS but is completely ignored on Android. The native setters (`setSelectedIconImageSource`, `setSelectedIconResourceName`) in `TabScreenViewManager.kt` are stubbed out as `= Unit`, `TabScreen.kt` has no property to hold the drawable, and the JS side in `parseIconsToNativeProps` never even sends `selectedIcon` data down to Android.

This PR wires it up end-to-end.

## What was wrong

I traced the data flow from JS → native and found it breaks in four places:

1. `TabScreenViewManager.kt` — both `setSelectedIconImageSource` and `setSelectedIconResourceName` are no-ops:

```kotlin
override fun setSelectedIconImageSource(
    view: TabScreen?,
    value: ReadableMap?,
) = Unit
```

2. `TabScreen.kt` — no `selectedIcon` property exists, so even if the view manager tried to set it, there's nowhere to store it.

3. `TabsScreen.tsx` — `parseIconsToNativeProps` only calls `parseAndroidIconToNativeProps` for the main icon and returns early, never touching `selectedIcon`:

```typescript
if (Platform.OS === 'android') {
    const androidNativeProps = parseAndroidIconToNativeProps(
      icon?.android || icon?.shared,
    );
    return { ...androidNativeProps };  // selectedIcon is dropped here
}
```

4. `TabsHostAppearanceApplicator.kt` — `updateMenuItemAppearance` just does `menuItem.icon = tabScreen.icon`, no concept of a selected state.

## What this PR does

**`TabScreen.kt`** — Added a `selectedIcon: Drawable?` property with the same `Delegates.observable` pattern as `icon`, so changes trigger `onMenuItemAttributesChange` and the menu item gets updated.

**`TabScreenViewManager.kt`** — `setSelectedIconImageSource` now actually loads the image:

```kotlin
override fun setSelectedIconImageSource(
    view: TabScreen?,
    value: ReadableMap?,
) {
    val uri = value?.getString("uri")
    if (uri != null && view != null) {
        loadTabImage(view.context, uri, view, isSelectedIcon = true)
    }
}
```

**`TabsImageLoader.kt`** — `loadTabImage` takes an `isSelectedIcon` flag. When `true`, the loaded drawable goes to `view.selectedIcon` instead of `view.icon`.

**`TabsHostAppearanceApplicator.kt`** — When both drawables are present, I wrap them in a `StateListDrawable`:

```kotlin
val newIcon = if (tabScreen.selectedIcon != null && tabScreen.icon != null) {
    StateListDrawable().apply {
        addState(intArrayOf(android.R.attr.state_checked), tabScreen.selectedIcon?.mutate())
        addState(intArrayOf(), tabScreen.icon?.mutate())
    }
} else {
    tabScreen.icon
}
menuItem.icon = newIcon
```

`BottomNavigationView` already toggles `state_checked` on the active menu item, so the swap happens for free — no listeners needed. I call `mutate()` on both drawables to avoid shared-state issues across tabs.

**`TabsScreen.tsx`** — Added `parseAndroidSelectedIconToNativeProps` (mirrors the existing `parseAndroidIconToNativeProps`) and updated `parseIconsToNativeProps` to spread the selected icon props into the return value on Android. Also widened the `selectedIcon` param type to accept `PlatformIconAndroid`.

## Notes

- Fully backward compatible. If `selectedIcon` isn't provided, the code path is identical to before — single icon with tinting.
- `itemIconTintList` still applies to both states. This matches the existing iOS behavior where tinting is separate from icon selection.
- Both icons load async via Fresco. Since `selectedIcon` uses the same `Delegates.observable` → `onMenuItemAttributesChange` chain, late-arriving images rebuild the `StateListDrawable` correctly.
- `setSelectedIconResourceName` (for `drawableResource` type) is still a no-op — can be done in a follow-up if needed.

## Companion PR needed in `react-navigation`

This PR handles the native side, but there's a separate issue in `@react-navigation/bottom-tabs`. In `BottomTabViewNativeImpl.tsx`, `selectedIcon` is resolved as:

```typescript
selectedIcon: selectedIcon?.ios ?? selectedIcon?.shared;
```

This always picks the iOS format (`templateSource`), which Android doesn't understand. The fix there is straightforward:

```typescript
selectedIcon: Platform.OS === 'ios'
  ? (selectedIcon?.ios ?? selectedIcon?.shared)
  : (selectedIcon?.android ?? selectedIcon?.shared);
```

Companion PR: https://github.com/react-navigation/react-navigation/pull/12974

## Testing

- Bottom tab navigator with different icons for `focused: true` / `focused: false` — verified icons swap on tap on Android.
- iOS — no regression, same behavior as before.
- Tabs without `selectedIcon` — still work with single-icon tinting as expected.
- Tested with async image loading (remote URLs) — icons appear correctly even when they load after initial render.

I've been running this as a patch (`patch-package`) in production for a while now with no issues.

### DIFF
--- a/FabricExample/e2e/examplesTests/events.e2e.ts
+++ b/FabricExample/e2e/examplesTests/events.e2e.ts
@@ -12,7 +12,7 @@ const pressBack = async () => {
 
 const awaitClassicalEventBehavior = async () => {
   if (device.getPlatform() === 'ios') {
-    // The order of events in this test differs from the Paper version of the same test.
+    // The order of events in this test differs from Paper.
     // Please see https://github.com/software-mansion/react-native-screens/pull/2785 for details.
     await expect(
       element(by.text('9. Chats | transitionStart | closing')),
--- a/FabricExample/ios/Podfile
+++ b/FabricExample/ios/Podfile
@@ -11,9 +11,6 @@ require_relative '../../scripts/ios/rns_set_swift_compilation_flags'
 platform :ios, min_ios_version_supported
 prepare_react_native_project!
 
-# From 0.76 new arch is enabled by default, we just want to have it explicitly set and for consistency with Example app
-ENV['RCT_NEW_ARCH_ENABLED'] = '1'
-
 # Disable precompiled binaries for RN core
 ENV['RCT_USE_PREBUILT_RNCORE'] = '0'
 
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/stack/screen/StackScreenFragment.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/stack/screen/StackScreenFragment.kt
@@ -18,16 +18,15 @@ internal class StackScreenFragment(
      * Since each StackScreenFragment owns a PreventNativeDismissCallback & adds it to the
      * OnBackPressedDispatcher the callback should be enabled only when the top fragment is this fragment.
      */
-    private val preventNativeDismissBackPressedCallback =
-        PreventNativeDismissCallback(this, stackScreen, canBeEnabled = true)
+    private var preventNativeDismissBackPressedCallback: PreventNativeDismissCallback? = null
+    private val requireNativeDismissBackPressedCallback
+        get() = checkNotNull(preventNativeDismissBackPressedCallback) { "[RNScreens] Attempt to require nullish OnBackPressedCallback" }
 
     private var isTopFragment: Boolean = false
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        requireActivity().onBackPressedDispatcher.addCallback(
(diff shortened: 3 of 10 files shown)


t0maboro:
Hi @Kasendwa 

Thanks a lot for your work on this!

I actually started working on a very similar change a few days ago https://github.com/software-mansion/react-native-screens/pull/3633, and in that PR some other aspects (old arch/aligning `icon` & `selectedIcon` types) are already comleted. I really like your idea of using `StateListDrawable` - very clean approach.

I hope you don't mind if I add you as a co-author to that other PR and close this one.

Kasendwa:
Hi @t0maboro that makes a lot of sense. Thank you.

PR #3655, opened 2026-02-16
fix(iOS, Tabs): defer scroll view behavior override until props are set

Ubax:
## Description

In case of the first render when the initial props were not set yet, the `mountChildComponentView` will most likely fire before the `updateProps`, and the `overrideScrollViewContentInsetAdjustmentBehavior` will use the default value `YES`. This makes it impossible to remove the override from `ScrollView` mounted during the first render.

## Changes

Defer the `overrideScrollViewBehaviorInFirstDescendantChainIfNeeded` call when props haven't yet been set (initial render) and apply it in `updateProps` once the prop value is known.

## Test plan

Test all three possible values of the `overrideScrollViewContentInsetAdjustmentBehavior`: `true`, `false`, `undefined`.

I added `apps/src/tests/single-feature-tests/tabs/override-scroll-view-content-inset.tsx` test case. During testing I also modified the order of the tabs.

https://github.com/user-attachments/assets/3823cd70-df40-4b78-8c04-16141f061d1a

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/apps/src/tests/single-feature-tests/tabs/index.ts
+++ b/apps/src/tests/single-feature-tests/tabs/index.ts
@@ -1,6 +1,7 @@
 import type { ScenarioGroup } from '../../shared/helpers';
 
 import BottomAccessoryScenario from './bottom-accessory-layout';
+import OverrideScrollViewContentInsetScenario from './override-scroll-view-content-inset';
 import TabBarHiddenScenario from './tab-bar-hidden';
 import TabsScreenOrientationScenario from './tabs-screen-orientation';
 
@@ -9,6 +10,7 @@ const TabsScenarioGroup: ScenarioGroup = {
   details: 'Single feature tests for tabs',
   scenarios: [
     BottomAccessoryScenario,
+    OverrideScrollViewContentInsetScenario,
     TabBarHiddenScenario,
     TabsScreenOrientationScenario,
   ],
--- a/apps/src/tests/single-feature-tests/tabs/override-scroll-view-content-inset.tsx
+++ b/apps/src/tests/single-feature-tests/tabs/override-scroll-view-content-inset.tsx
@@ -1,0 +1,114 @@
+import React from 'react';
+import { ScrollView, Text, View, StyleSheet } from 'react-native';
+import {
+  NavigationContainer,
+  NavigationIndependentTree,
+} from '@react-navigation/native';
+import { BottomTabsContainer } from '../../../shared/gamma/containers/bottom-tabs/BottomTabsContainer';
+import type { Scenario } from '../../shared/helpers';
+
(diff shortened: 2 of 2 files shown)


ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
```diff
@@ -465,7 +474,13 @@ - (void)mountChildComponentView:(UIView<RCTComponentViewProtocol> *)childCompone
...
   // from screen to find ScrollView, that's why we're only interested in child mounted at index 0.
   if (index == 0) {
-    [self overrideScrollViewBehaviorInFirstDescendantChainIfNeeded];
+    // If the props were not set before the first child was mounted,
+    // we defer the override behavior application until props are set.
+    if (_isOverrideScrollViewContentInsetAdjustmentBehaviorSet) {
+      [self overrideScrollViewBehaviorInFirstDescendantChainIfNeeded];
+    } else {
+      _needsScrollViewBehaviorOverride = YES;
+    }
```
kkafar: Is the case the comment describes ever not true? 😄 

On the new architecture a given component view, receives it's props ALWAYS after its children are mounted (assuming they are mounted in the first render). This is not documented, but unless something changes in (recently very stable) the very core mutation calculating algorithm, this should always be the case AFAIK.

Doing this in head, I'll try to debug later, but I try to think of scenario this if-else actually covers. During the very first mounting, this will go for "else branch" for every single child & will be deferred to `updateProps:oldProps:` - which would naturally happen anyway. 

Oh, I think I see! This prevents from actually setting the option before we receive users decision in `updateProps:oldProps:` during the first render. Makes sense. 

I think that the comment is misleading here. It should be something like this:

```suggestion
    // Before the props are set (first render) defer acting to after
    // we receive props, as user might have decided to disable this feature.
    if (_isOverrideScrollViewContentInsetAdjustmentBehaviorSet) {
      [self overrideScrollViewBehaviorInFirstDescendantChainIfNeeded];
    } else {
      _needsScrollViewBehaviorOverride = YES;
    }
```

kkafar (commented):
Looks good. I've left one remark, please answer it & I think we can proceed.

kkafar:
Also, please update the PR description. You can leave test plan empty - I will fill it when we're ready OR you can point to the test that was introduced with PR that introduced this feature - that's better option.

ios/bottom-tabs/screen/RNSBottomTabsScreenComponentView.mm
Ubax: I made a mental shortcut in the comment 😅 

Good catch

PR #3658, opened 2026-02-17
chore(Tabs): align naming for native components

kligarski:
## Description

Aligns naming for native components in tabs (TabsHost, TabsScreen, TabsBottomAccessory).

Closes https://github.com/software-mansion/react-native-screens-labs/issues/782.

## Changes

- renaming of iOS native components
- renaming of Android native components
- renaming of Fabric spec files
- renaming of `.h`/`.cpp` ShadowTree files

No visual changes.

## Test plan

Tabs should work without any changes on both platforms & both architectures.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsHost.kt
@@ -353,7 +353,11 @@ class TabsHost(
         }
     }
 
-    override fun getFragmentForTabsScreen(tabsScreen: TabsScreen): TabsScreenFragment? = tabsScreenFragments.find { it.tabsScreen === tabsScreen }
+    override fun getFragmentForTabsScreen(tabsScreen: TabsScreen): TabsScreenFragment? =
+        tabsScreenFragments.find {
+            it.tabsScreen ===
+                tabsScreen
+        }
 
     override fun onFragmentConfigurationChange(
         tabsScreen: TabsScreen,


android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabScreenEventEmitter.kt
```diff
@@ -42,5 +41,5 @@ private fun logEventDispatch(
     viewTag: Int,
     eventName: String,
 ) {
-    RNSLog.d(TAG, "TabScreen [$viewTag] emits event: $eventName")
+    RNSLog.d(TabScreenEventEmitter.Companion.TAG, "TabScreen [$viewTag] emits event: $eventName")
```
kligarski: intelliJ did some refactors when moving the files, not sure if I should keep them or revert them.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/host/TabsHostAppearanceApplicator.kt
```diff
@@ -29,7 +31,7 @@ class TabsHostAppearanceApplicator(
         bottomNavigationView.isVisible = !tabsHost.tabBarHidden
         bottomNavigationView.setBackgroundColor(
             tabsHost.tabBarBackgroundColor
-                ?: resolveColorAttr(com.google.android.material.R.attr.colorSurfaceContainer),
+                ?: resolveColorAttr(R.attr.colorSurfaceContainer),
```
kligarski: Same here.

src/components/tabs/TabsAccessory.tsx
```diff
@@ -8,7 +8,7 @@ import { StyleSheet } from 'react-native';
  */
 export default function TabsAccessory(props: TabsAccessoryProps) {
```
kligarski, line 9: I did not change the JS component name here but in order to have full alignment we would need `TabsBottomAccessory` here. Not sure if we want to change it?

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabScreenEventEmitter.kt
kmichalikk: I'd remove it. This doesn't belong to the PR and we don't need to change it.

kligarski: Okay, I removed it.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/host/TabsHostAppearanceApplicator.kt
kligarski: As discussed, will handle this in separate PR.

kligarski:
>     1. Let's maybe move the resource id refactor to separate PR
>     2. You might also create a PR for moving android tabs from `gamma/` folder

Will do so in separate PR(s).

>     3. There are still few places where BottomTabs could be replaced with Tabs, i.e. in GUIDE_FOR_LIBRARY_AUTHORS and code comments (ctrl + f). There is also `controlledBottomTabs` flag - should we change it?

I think this should be handled in separate PRs. I added a ticket to rename `controlledBottomTabs`: https://github.com/software-mansion/react-native-screens-labs/issues/974.

src/components/tabs/TabsAccessory.tsx
kligarski: I created a ticket to do so: https://github.com/software-mansion/react-native-screens-labs/issues/973 (we'll batch some of these breaking changes).

react-native.config.js
```diff
@@ -14,7 +14,7 @@ module.exports = {
           "RNSScreenFooterComponentDescriptor",
           "RNSScreenContentWrapperComponentDescriptor",
           'RNSModalScreenComponentDescriptor',
-          'RNSBottomTabsComponentDescriptor',
+          'RNSTabsHostComponentDescriptor',
           'RNSSafeAreaViewComponentDescriptor'
         ],
         cmakeListsPath: "../android/src/main/jni/CMakeLists.txt"
```
kkafar, line 20: I do not remember how this file is used by CLI exactly, but it seems that we're missing a couple of components here, right?

kkafar: Talking about new stack implementation? `TabsBottomAccessory`? And maybe something else. We should investigate what this has been added for, whether it is still relevant & we either need to remove or update this file. 

Let's create ticket for that please.

kligarski: There is already a ticket for this: https://github.com/software-mansion/react-native-screens-labs/issues/621.

kligarski: Please see: https://github.com/software-mansion/react-native-screens-labs/issues/621#issuecomment-3943738653.

kligarski:
> One thing I'm mising here is usage of `TabsScreen` on Android, right? We should keep it aligned between platforms

Thank you for spotting this, don't know how I've missed that 😅

https://github.com/software-mansion/react-native-screens/pull/3658/commits/09737f96373e5cacc37cce28303d24717daaac52

kligarski:
> I haven't verified runtime behavior. Please make sure you do.

I've checked `TestBottomTabs` on `{Android, iOS} x {Paper, Fabric}` and it looks good. I'll wait for the CI and proceed.

PR #3661, opened 2026-02-17
chore(Android, Tabs): move tabs files to separate packages

kligarski:
## Description

Refactors file structure in tabs directory. 

## Changes

- create `host` and `screen` directories and move appropriate files
- create `event` package `host` and `screen` and move appropriate files
- apply refactors from IntelliJ

No visual changes.

## Test plan

Tabs should work without any changes.

## Checklist

- [x] Ensured that CI passes


t0maboro:
> I wanted to also move tabs directory out of gamma but when I thought about it more, we rely on a lot of helpers from gamma directory. On the other hand, Stack v4 also uses some helpers from gamma directory so I'm not sure how to proceed here.

If stack v4 is relying on certain APIs from gamma, those APIs shouldn't reside in gamma

Issue #3662, opened 2026-02-18
TabsHost fillParent style prevents flex layout with sibling views


kimchi-developer:
## Description

`TabsHost` (formerly `BottomTabsHost`) applies a `fillParent` style with `flex: 1, width: '100%', height: '100%'`. The explicit `width` and `height` values prevent the component from participating in standard React Native flex layout when placed alongside sibling views in a parent container.

This means wrapping `NativeTabs` in a `<View style={{ flex: 1 }}>` with a sibling element (e.g. a banner ad) does not work — TabsHost always takes the full parent size, ignoring the sibling.

## Steps to reproduce

```tsx
// Expected: BannerAd renders above, NativeTabs fills remaining space
// Actual: NativeTabs covers full parent, BannerAd is hidden behind it
<View style={{ flex: 1 }}>
  <BannerAd />
  <NativeTabs>
    {/* tab triggers */}
  </NativeTabs>
</View>
```

## Root cause

In `src/components/tabs/TabsHost.tsx`:

```ts
const styles = StyleSheet.create({
  fillParent: {
    flex: 1,
    width: '100%',  // ← forces full parent width, ignores flex siblings
    height: '100%', // ← forces full parent height, ignores flex siblings
  },
});
```

`flex: 1` alone is sufficient — the native iOS side (`RNSBottomTabsHostComponentView.mm`) pins the controller view to the host view bounds via `NSLayoutConstraint`, and the Android side (`TabsHost.kt`) uses `MATCH_PARENT`. Both correctly size themselves to whatever bounds the JS view provides.

## Proposed fix

Remove `width: '100%'` and `height: '100%'`, keeping only `flex: 1`:

```ts
const styles = StyleSheet.create({
  fillParent: {
    flex: 1,
  },
});
```

## Environment

- react-native-screens: 4.22.0
- expo-router: unstable-native-tabs
- Platform: iOS 26.2 (also affects Android)

kmichalikk:
Hi @kimchi-developer 
I don't see the bug in a minimal example.
more detailed answer in the PR

PR #3663, opened 2026-02-18
fix(Tabs): remove width/height 100% from TabsHost fillParent style

kimchi-developer:
## Summary

Closes https://github.com/software-mansion/react-native-screens/issues/3662

- Remove `width: '100%'` and `height: '100%'` from `TabsHost`'s `fillParent` style
- Keep only `flex: 1`, which is sufficient for the native layout constraints on both iOS and Android

## Problem

`TabsHost` cannot be used as a flex child alongside sibling views. The explicit `width/height: '100%'` forces it to take the full parent size, making patterns like a persistent banner above tabs impossible.

## Why this is safe

- **iOS**: `RNSBottomTabsHostComponentView.mm` pins `controller.view` to the host view's 4 edges via `NSLayoutConstraint` — it fills whatever bounds the JS view provides
- **Android**: `TabsHost.kt` uses `MATCH_PARENT` — same behavior, fills the parent view's bounds
- **`TabsScreen`** is not affected — it uses `position: 'absolute'` (correct for stacking tab content)

## Test plan

- Wrap `NativeTabs` in a `<View style={{ flex: 1 }}>` with a sibling `<View>` above it
- Verify the sibling renders above and NativeTabs fills remaining space
- Verify tabs render correctly when used standalone (no siblings) — `flex: 1` still fills parent

### DIFF
--- a/src/components/tabs/TabsHost.tsx
+++ b/src/components/tabs/TabsHost.tsx
@@ -99,7 +99,5 @@ export default TabsHost;
 const styles = StyleSheet.create({
   fillParent: {
     flex: 1,
-    width: '100%',
-    height: '100%',
   },
 });


kmichalikk:
Hi @kimchi-developer 
Since You mention `NativeTabs` I'm assuming you are using expo.
I created a blank expo project `npx create-expo-app@latest --template tabs@55`, replaced the default implementation with NativeTabs, wrapped it with a view with `flex: 1`, and added some blank views at the top / bottom. It seems to be working without any patching on screens@4.23.

image

Could You please provide more information about the problem and before / after?

(also tested on our FabricExample, works ok without the PR)

kimchi-developer:
Thank you for taking the time to verify this! You're right — I just tested on screens@4.22 with the original `fillParent` style (including `width: '100%'` and `height: '100%'`) and it works fine. The banner renders correctly above NativeTabs without any patching.

It seems the issue I originally observed was caused by something else in my setup. Closing this PR.

Sorry for the noise, and thanks again for the thorough review! 🙏

PR #3664, opened 2026-02-18
chore: Bring back Example for Paper

kmichalikk:
Closes https://github.com/software-mansion/react-native-screens-labs/issues/971

## Description

We're decided to defer the removal of Paper Example. We'll still be needing it for some time. Since the tests are failing on GH runners nondeterministically, we're skipping them. We don't want to put more time to it & the app will only exist for manual checks on PRs.

## Changes

Reverted the removal of Example. Not reverted:
- e2e workflow files
- string instances of ScreensExample

## Test plan

Build the Paper Example and see if it works.

## Checklist

- [x] Ensured that CI passes


Example/e2e/apps/test-butler-app-2.2.1.apk
kligarski, line 1: Should this be kept in Example or FabricExample now?

Example/e2e/examplesTests/bottomTabs.e2e.ts
kligarski, line 1: In theory, we can get rid of entire Example/e2e directory, right? But if we're not gonna run them anyway maybe let's not complicate this and keep the files (there is also e.g. DetoxTest.java, ...).

Example/e2e/apps/test-butler-app-2.2.1.apk
kmichalikk: [87aa676](https://github.com/software-mansion/react-native-screens/pull/3664/commits/87aa676f07a3f36204931ad1ce4d1e70b2ca7e30)

Example/e2e/examplesTests/bottomTabs.e2e.ts
kmichalikk: I'm keeping them because it doesn't matter that much & removing it cleanly would take some time, also you can run them locally. I only removed the workflow files so they are not run on GH actions.

PR #3671, opened 2026-02-20
chore: Move iOS runners to macos-26-xlarge instances

kmichalikk:
## Description

This PR moves iOS workflows to `macos-26-xlarge` runners to see if they are more reliable

## Changes

Updated `runs-on` on iOS workflows

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Check types
         run: yarn check-types
   lint-ios:
-    runs-on: macos-26-xlarge
+    runs-on: macos-26
     concurrency:
       group: install-and-lint-ios-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - name: Check types
         run: yarn check-types
   lint-ios:
-    runs-on: macos-15
+    runs-on: macos-26-xlarge
```
kkafar: Let's swap to 26, but we do not need large instance here. It takes like a minute to run anyway.

.github/workflows/ios-build-test.yml
kkafar, line 1: This workflow no longer exists. Let's rebase.

PR #3672, opened 2026-02-20
refactor(Android, Tabs): Appearance refactor on Android

t0maboro:
## Description

This PR refines the Android tabs' appearance configuration into a more structured, explicit form.

The goal of this change is to:

- Align the Android API more closely with the iOS implementation
- Make appearance configuration more predictable

The appearance configuration for Android tabs has been reorganized into clearly separated sections, describing different aspects of TabBar:

- backgroundColor
- itemRippleColor
- labelVisibilityMode
- itemColors (normal / selected / focused / disabled)
- activeIndicator
- typography
- badge

---

From now on, the appearance of the entire Android tab bar is determined by the currently active Screen, the same way it is handled on iOS. When the active Screen changes, the tab bar appearance updates accordingly.

---

While the API allows updating appearance dynamically, some properties can negatively affect layout when changed at runtime.

Changing these props dynamically may cause visual glitches:
- labelVisibilityMode
- activeIndicator.enabled
- typography props

For this reason, it is not recommended to change these properties at runtime. They should be treated as relatively stable configuration options.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/955

## Changes

- Renamed classes: AppearanceApplicator & AppearanceCoordinator are now less coupled with the Host component, as they depend on the selected Screen, either. Therefore, I decided to remove "Host" from the names of these classes.
- Diffing mechanism before applying tab bar appearance updates was added. This will prevent assigning the same values multiple times if a specific prop remains unchanged.
- ColorsStateLists were extended to support focused and disabled states - parity with UIAppearance states from UIKit.
- New TabsAppearanceModel was defined.
- Old way of passing all props as top-level props was replaced with passing appearance config object from JS. This is a breaking change. We'll add compatibility flag after the full refactor.
- All native prop "legacy" setters were replaced with the specific parser for Appearance object.
- Appearance configuration was moved from Host to Screen.

This PR unlocks some new behaviors, not sure if Before section is needed, as for Tab 1 & Tab 4 - the flow remains unchanged. For Tab2 & Tab3 - I'm adding new features for configuring TabBar appearance, which are covered in `After` video

### After

https://github.com/user-attachments/assets/e8d017bb-7861-4eac-a3ab-5d3668c689eb

## Test plan

Updated TestBottomTabs example to match the new API structure.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
@@ -2,8 +2,8 @@ package com.swmansion.rnscreens.gamma.tabs.appearance
 
 import android.annotation.SuppressLint
 import android.content.res.ColorStateList
-import android.graphics.drawable.StateListDrawable
 import android.graphics.Color
+import android.graphics.drawable.StateListDrawable
 import android.util.Log
 import android.util.TypedValue
 import android.view.MenuItem


src/components/tabs/TabsScreen.types.ts
```diff
@@ -75,6 +82,180 @@ export type TabsScreenOrientation =
...
+export interface BottomNavItemColorsAndroid {
+  /**
+   * Specifies the tab bar item colors when it's enabled and unselected.
+   *
+   * @platform android
+   */
+  normal?: ItemStateColorsAndroid;
+  /**
+   * Specifies the tab bar item colors when it's selected.
+   * Maps to Android `state_checked=true`.
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3672/changes/86b2c4365df4d01fe6a31c2d4940d49ec134daec

src/fabric/tabs/TabsScreenNativeComponent.ts
```diff
@@ -53,6 +52,56 @@ export type Appearance = {
...
+
+export type BadgeAppearanceAndroid = {
+  textColor?: ProcessedColorValue | null;
+  backgroundColor?: ProcessedColorValue | null;
+};
+
+export type AppearanceAndroid = {
+  tabBarBackgroundColor?: ProcessedColorValue | null;
+  tabBarItemRippleColor?: ProcessedColorValue | null;
+  tabBarItemLabelVisibilityMode?: CT.WithDefault<
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3672/changes/e8e84158773c6732fe1e0f3bac97cacfba4fb9e0

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsAppearanceApplicator.kt
```diff
@@ -0,0 +1,315 @@
...
+
+    private val lastBadgeValues = mutableMapOf<Int, String?>()
+    private val lastBadgeTextColors = mutableMapOf<Int, Int>()
+    private val lastBadgeBackgroundColors = mutableMapOf<Int, Int>()
+
+    private val states =
+        arrayOf(
+            intArrayOf(-android.R.attr.state_enabled), // disabled
+            intArrayOf(android.R.attr.state_focused), // focused
+            intArrayOf(android.R.attr.state_selected), // selected
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3672/changes/76604f51fb4c3564604768f39410195c2fe1148b

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsAppearanceApplicator.kt
```diff
@@ -0,0 +1,315 @@
...
+            } as Int
+
+        // https://github.com/material-components/material-components-android/blob/master/docs/getting-started.md#non-transitive-r-classes-referencing-library-resources-programmatically
+        val newBadgeBackgroundColor =
+            badgeAppearance?.backgroundColor
+                ?: resolveColorAttr(androidx.appcompat.R.attr.colorError)
+        lastBadgeBackgroundColors[menuItemIndex] =
+            updatePropIfChanged(lastBadgeBackgroundColors[menuItemIndex], newBadgeBackgroundColor) {
+                badge.backgroundColor = newBadgeBackgroundColor
+            } as Int
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3672/changes/dd2da07e42703610c22cfd0cb849cb5055dcd6e7

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/TabsAppearanceApplicator.kt
```diff
@@ -0,0 +1,315 @@
...
+            }
+
+        // Styling
+        val newBadgeTextColor =
+            badgeAppearance?.textColor
+                ?: resolveColorAttr(com.google.android.material.R.attr.colorOnError)
+        lastBadgeTextColors[menuItemIndex] =
+            updatePropIfChanged(lastBadgeTextColors[menuItemIndex], newBadgeTextColor) {
+                badge.badgeTextColor = newBadgeTextColor
+            } as Int
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3672/changes/dd2da07e42703610c22cfd0cb849cb5055dcd6e7

src/flags.ts
```diff
@@ -43,6 +43,13 @@ export const compatibilityFlags = {
    * is in use or not.
    */
   usesNewAndroidHeaderHeightImplementation: true,
+
+  /**
+   * In https://github.com/software-mansion/react-native-screens/pull/3672, we refactored
+   * appearance props for Tabs on Android. To allow backward compatibility,
+   * we expose a way to check whether the new implementation is in use or not.
+   */
+  usesRefactoredTabsAppearanceApiAndroid: true,
```
t0maboro: note: it should be coordinated with https://github.com/software-mansion/react-native-screens/pull/3633 and hidden behind a single compat. flag

src/components/tabs/TabsScreen.types.ts
```diff
@@ -75,6 +82,180 @@ export type TabsScreenOrientation =
...
+   *
+   * The supported values correspond to the official Material Components documentation:
+   * @see {@link https://github.com/material-components/material-components-android/blob/master/docs/components/BottomNavigation.md#making-navigation-bar-accessible|Material Components documentation}
+   *
+   * @default auto
+   *
+   * @platform android
+   */
+  labelVisibilityMode?: TabBarItemLabelVisibilityMode;
+
```
kligarski:
```suggestion
```

src/components/tabs/TabsScreen.types.ts
```diff
@@ -75,6 +82,180 @@ export type TabsScreenOrientation =
...
+  backgroundColor?: ColorValue;
+  /**
+   * @summary Specifies the text color of the badge.
+   *
+   * @platform android
+   */
+  textColor?: ColorValue;
+}
+
+export interface AndroidTabsAppearance {
```
kligarski: I'm not sure about the naming here, sometimes platform name is at the beginning, sometimes at the end (e.g. `AndroidTabsAppearance`, `BadgeAppearanceAndroid`). We should probably have a look at other names as well, like `TabsScreenAppearance` now without any platform information even if it's iOS-only but it might be for another PR and more related to splitting the component.

src/components/tabs/TabsScreen.types.ts
```diff
@@ -75,6 +82,180 @@ export type TabsScreenOrientation =
...
+  /**
+   * Specifies the tab bar item colors when it's selected.
+   * Maps to Android `state_selected=true`.
+   *
+   * @platform android
+   */
+  selected?: ItemStateColorsAndroid;
+  /**
+   * Specifies the tab bar item colors when it's focused.
+   * Maps to Android `state_focused=true` (Used mostly for Keyboard navigation).
```
kligarski:
```suggestion
   * Maps to Android `state_focused=true` (Used mostly for keyboard navigation).
```

src/components/tabs/TabsScreen.types.ts
```diff
@@ -75,6 +82,180 @@ export type TabsScreenOrientation =
...
+   *
+   * @platform android
+   */
+  focused?: ItemStateColorsAndroid;
+  /**
+   * Specifies the tab bar item colors when it's disabled.
```
kligarski: I'm not sure whether those descriptions are clear for other people (I know that the same applies to iOS). Maybe we should describe it in more detail that the colors apply to entire tab bar but currently selected tab decides about that?

src/components/tabs/TabsScreen.types.ts
```diff
@@ -75,6 +82,180 @@ export type TabsScreenOrientation =
...
+
+export interface TypographyAppearanceAndroid {
+  /**
+   * @summary Specifies the font family used for the title of each tab bar item.
+   *
+   * @platform android
+   */
+  fontFamily?: TextStyle['fontFamily'];
+  /**
+   * @summary Specifies the font size used for the title unselected tab bar items.
```
kligarski:
```suggestion
   * @summary Specifies the font size used for the title of unselected tab bar items.
```

src/components/tabs/TabsScreen.types.ts
```diff
@@ -417,17 +598,14 @@ export interface TabsScreenProps {
...
-   * @summary Specifies the color of the text in the badge.
+   * @summary Specifies the standard tab bar appearance.
    *
-   * @platform android
-   */
-  tabBarItemBadgeTextColor?: ColorValue;
-  /**
-   * @summary Specifies the background color of the badge.
+   * Allows to customize the appearance depending on the tab bar item state
+   * (normal, selected, focused, disabled). Configuration for the Bottom Navigation View.
```
kligarski: t"Configuration for the Bottom Navigation View." sounds a little bit random, maybe we should change it into a full sentence?

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
```diff
@@ -0,0 +1,324 @@
...
+    ): IntArray {
+        if (oldValue == null || !oldValue.contentEquals(newValue)) {
+            updateFn(newValue)
+        }
+        return newValue
+    }
+
+    private fun resolveColorAttr(attr: Int): Int {
+        val typedValue = TypedValue()
+        context.theme.resolveAttribute(attr, typedValue, true)
```
kligarski: In theory, this function can return false - maybe we should throw an error in such case?

src/components/tabs/TabsScreen.types.ts
t0maboro: I'll only rename it here; changing iOS types to be more descriptive fits more to splitting components PR, so I'll keep that in mind in that follow-up PR.

src/components/tabs/TabsScreen.types.ts
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3672/changes/9fbecd403cb3773e86a2d865a9d62b73e9cba031

src/components/tabs/TabsScreen.types.ts
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3672/changes/13a738a4d03bb004431bea8c8d8c850469f0baec

src/components/tabs/TabsScreen.types.ts
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3672/changes/85df7a26c3fd89f7c92a95661ff271b913babdef

src/components/tabs/TabsScreen.types.ts
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3672/changes/4c0c7bcbe84b9f9e2f360acfcff762cf21d5d6f6

src/components/tabs/TabsScreen.types.ts
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3672/changes/2447fb80de30063fad0e99b8b497417ab5a1a4d1

src/components/tabs/TabsScreen.types.ts
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3672/changes/11d015247b4fbdf0d10e34b4487d7c2218a08029

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3672/changes/6ca7afb8060c55ccbc412ffdc8557e9071432b97

src/components/tabs/TabsScreen.types.ts
kligarski: That's great, thank you.

src/components/tabs/TabsScreen.types.ts
kligarski: `TabScreen` or `TabsScreen`? Or maybe `tab screen`?

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3672/commits/721d19a95d81dce0695c3b1394dcb69283801234

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
```diff
@@ -0,0 +1,351 @@
...
+    private var lastFontColors: IntArray? = null
+    private var lastIconColors: IntArray? = null
+    private var lastLabelVisibilityMode: Int? = null
+    private var lastRippleColor: Int? = null
+    private var lastActiveIndicatorColor: Int? = null
+    private var lastIsActiveIndicatorEnabled: Boolean? = null
+
+    private val lastBadgeValues = mutableMapOf<Int, String?>()
+    private val lastBadgeTextColors = mutableMapOf<Int, Int>()
+    private val lastBadgeBackgroundColors = mutableMapOf<Int, Int>()
```
kkafar, line 39: Haven't read through here, but my immediate impression is that this class should be completely stateless. 

Most of what you want to store here should be read directly from the component, instead of duplicating the state here.

I'll revisit this comment later if I find a reason for keeping this state.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
```diff
@@ -0,0 +1,351 @@
...
+import com.facebook.react.uimanager.PixelUtil
+import com.google.android.material.R
+import com.google.android.material.bottomnavigation.BottomNavigationView
+import com.google.android.material.navigation.NavigationBarView
+import com.swmansion.rnscreens.gamma.tabs.host.TabsHost
+import com.swmansion.rnscreens.gamma.tabs.screen.TabsScreen
+
+@SuppressLint("PrivateResource") // We want to use variables from material design for default values
+class TabsAppearanceApplicator(
+    private val context: ContextThemeWrapper,
```
kkafar, line 26: I believe that context retains the view, view retains `TabsAppearanceApplicator` and `TabsAppearanceApplicator` retains the context. Seems like a retain cycle to me. The `context` should be passed directly to the methods instead of being kept in the state.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
```diff
@@ -0,0 +1,351 @@
...
+    private inline fun <T> updatePropIfChanged(
+        oldValue: T,
+        newValue: T,
+        updateFn: (T) -> Unit,
+    ): T {
+        if (oldValue != newValue) {
+            updateFn(newValue)
+        }
+        return newValue
+    }
```
kkafar, line 58: If we move into "stateless" direction then this helper won't bring any value

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
```diff
@@ -0,0 +1,351 @@
...
+    private inline fun updatePropsIfArrayChanged(
+        oldValue: IntArray?,
+        newValue: IntArray,
+        updateFn: (IntArray) -> Unit,
+    ): IntArray {
+        if (oldValue == null || !oldValue.contentEquals(newValue)) {
+            updateFn(newValue)
+        }
+        return newValue
+    }
```
kkafar, line 69: Similarly here

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
```diff
@@ -0,0 +1,351 @@
...
+        val typedValue = TypedValue()
+        val resolved = context.theme.resolveAttribute(attr, typedValue, true)
+
+        if (!resolved) {
+            Log.w(TAG, "[RNScreens] Failed to resolve color attribute. Falling back to Color.TRANSPARENT.")
+            return defaultColor
+        }
+
+        return typedValue.data
+    }
```
kkafar, line 84: This seems like a nice candidate for general purpose utility function.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
```diff
@@ -0,0 +1,351 @@
...
+
+    fun updateSharedAppearance(tabsHost: TabsHost) {
+        val tabBarAppearance = tabsHost.currentFocusedTab.tabsScreen.appearance
+
+        updatePropIfChanged(bottomNavigationView.isVisible, !tabsHost.tabBarHidden) {
+            bottomNavigationView.isVisible = !tabsHost.tabBarHidden
+        }
```
kkafar, line 91: To my taste, this helper function here does not bring any value but only obfuscates the code. 

If we deicide to move into "stateless" approach, then I'd recommend removing this helper.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
```diff
@@ -0,0 +1,351 @@
...
+
+        val newBackgroundColor =
+            tabBarAppearance?.backgroundColor
+                ?: resolveColorAttr(R.attr.colorSurfaceContainer)
+        lastBackgroundColor =
+            updatePropIfChanged(lastBackgroundColor, newBackgroundColor) {
+                bottomNavigationView.setBackgroundColor(newBackgroundColor)
+            }
```
kkafar, line 99: okay, I see here why you might've wanted to keep the state. You can get the background colour though. There is even an extension function we created for this: `maybeBgColor` I believe.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
```diff
@@ -0,0 +1,351 @@
...
+
+        val fontNormalColor =
+            tabBarAppearance?.itemColors?.normal?.titleColor
+                ?: resolveColorAttr(R.attr.colorSecondary)
+
+        val newFontColors = intArrayOf(fontDisabledColor, fontSelectedColor, fontFocusedColor, fontNormalColor)
+        lastFontColors =
+            updatePropsIfArrayChanged(lastFontColors, newFontColors) {
+                bottomNavigationView.itemTextColor = ColorStateList(states, newFontColors)
+            }
```
kkafar, line 123: Okay, this might be a harder nut to crack, but I'll propose an alternative solution in the main review message.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceCoordinator.kt
```diff
@@ -1,44 +1,47 @@
...
     }
 
-    private fun updateMenuItems() {
+    private fun updateMenuItems(tabsHost: TabsHost) {
         if (bottomNavigationView.menu.size != tabsScreenFragments.size) {
             // Most likely first render or some tab has been removed. Let's nuke the menu (easiest option).
             bottomNavigationView.menu.clear()
         }
         tabsScreenFragments.forEachIndexed { index, fragment ->
+            val appearance = tabsHost.currentFocusedTab.tabsScreen.appearance
```
kkafar, line 31: I suggest to move this outside of the loop.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceModel.kt
```diff
@@ -0,0 +1,41 @@
+package com.swmansion.rnscreens.gamma.tabs.appearance
+
+internal data class AndroidTabsAppearance(
```
kkafar, line 3:
```suggestion
internal data class TabsAppearance(
```

I think we don't need to specify such prefix in Android-specific code. What do you think?

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabsScreenDelegate.kt
```diff
@@ -4,6 +4,8 @@ import android.content.res.Configuration
 import androidx.fragment.app.Fragment
 
 internal interface TabsScreenDelegate {
+    fun onAppearanceChanged(tabScreen: TabsScreen)
```
kkafar, line 7:
```suggestion
    fun onAppearanceChanged(tabsScreen: TabsScreen)
```

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabsScreenViewManager.kt
```diff
@@ -259,6 +253,118 @@ class TabsScreenViewManager :
...
+        }
+        val appearanceMap = if (value.type == ReadableType.Map) value.asMap() else null
+
+        if (appearanceMap == null) {
+            view.appearance = null
+            return
+        }
+
+        view.appearance = parseAndroidTabsAppearance(appearanceMap)
+    }
```
kkafar, line 273: Could you explain why you decided to use UnsafeMix & Dynamic instead of properly typed values?

apps/src/tests/issue-tests/TestBottomTabs/index.tsx
kkafar, line 1: I think we need dedicated "new style" single feature test for this, e.g. `test-selected-tab-determines-appearance` or something like that. 

This test is vastly overused.

src/components/tabs/TabsScreen.tsx
```diff
@@ -124,6 +132,9 @@ function TabsScreen(props: TabsScreenProps) {
       isFocused={isFocused}
       {...iconProps}
       standardAppearance={mapAppearanceToNativeProp(standardAppearance)}
+      standardAppearanceAndroid={mapAndroidAppearanceToNativeProp(
+        standardAppearanceAndroid,
+      )}
```
kkafar, line 137: Is this only temporary?

src/components/tabs/TabsScreen.tsx
```diff
@@ -163,6 +174,85 @@ function mapAppearanceToNativeProp(
   };
 }
 
+export function mapAndroidAppearanceToNativeProp(
```
kkafar, line 177: I don't see any usages of this function outside of this module. Why is it exported?

src/components/tabs/TabsScreen.tsx
```diff
@@ -163,6 +174,85 @@ function mapAppearanceToNativeProp(
...
+function mapBottomNavItemColorsAndroid(colors?: BottomNavItemColorsAndroid) {
+  if (!colors) return undefined;
+
+  return {
+    normal: mapItemStateColorsAndroid(colors.normal),
+    selected: mapItemStateColorsAndroid(colors.selected),
+    focused: mapItemStateColorsAndroid(colors.focused),
+    disabled: mapItemStateColorsAndroid(colors.disabled),
+  };
+}
```
kkafar, line 212: I want to avoid a mistake we made when merging code adding header items on iOS functionality. These are very similar functions we added there and they all lack one thing: properly typed return types. Please add them explicitly.

src/components/tabs/TabsScreen.types.ts
```diff
@@ -74,6 +81,183 @@ export type TabsScreenOrientation =
...
+}
+
+export interface BottomNavItemColorsAndroid {
+  /**
+   * Defines the colors for all tab bar items which are in their enabled, unselected and unfocused state.
+   * The color scheme is determined by the configuration of the currently selected tab.
+   *
+   * @platform android
+   */
+  normal?: ItemStateColorsAndroid;
```
kkafar, line 107: We need consistent nomenclature here. First type describes colors of `Item` in a particular state, where second one describes colors of `BottomNavItem`. This implies that `Item` and `BottomNavItem` are two separate entities.

kkafar: I'll try to come up with some suggestions later

t0maboro:
> First I want to say that this PR introduces an intermediary state between current state & what we want to achieve. I don't think we should land it in ints current shape. Let's first merge here other stacked PRs so that it is self-contained and it actually makes sense to land it on main.

I'll hold off on addressing comments here, as some of them will be outdated when we merge https://github.com/software-mansion/react-native-screens/pull/3683 & https://github.com/software-mansion/react-native-screens/pull/3684 into this PR

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabsScreenViewManager.kt
t0maboro: see the PR description for iOS Appearance, where we used UnsafeMixed either to workaround codegen issues: https://github.com/software-mansion/react-native-screens/pull/3078#:~:text=send%20apperance%20props%20via%20UnsafeMixed%20to%20avoid%20codegen%20issues%20(nested%20objects%20are%20set%20to%20null%2C%20no%20optionals)%20or%20having%20192%20props%20inside%20component%20(if%20we%20wanted%20to%20flatten%20the%20hierarchy)

t0maboro:
superseded by: https://github.com/software-mansion/react-native-screens/pull/3714

PR #3674, opened 2026-02-21
feat: add `ScrollViewMarker` experimental component

kkafar:
## Description

This PR adds a PoC of `ScrollViewMarker` experimental native component.

It is introduced to handle two responsibilities:

1. Configuration of `ScrollView` (in particular edge effects)

    I want to move these configuration options off the "screen components",
    and have a single component that can be used to configure the `ScrollView`.

2. Detection of `ScrollView` in native view hierarchy

    Currently, both we & UIKit employ very simple "first descendant chain" heuristics
    to detect the presence of `ScrollView` in the view hierarchy & use it as
    "content ScrollView". This sometimes works, but it is very easy to break by
    user code.

    Now, user will be required to wrap the `ScrollView` in our component, so this
    is an additional step, but it will allow us to robustly detect the presence
    of `ScrollView`.

    There is open question of what do we do, if there are multiple `ScrollViews`
    a user want to configure. There can be only a single "content `ScrollView`".
    Atm. I think that reasonable approach is to have non-required `isContentScrollView`
    property with `true` as default value, and if there are multiple `ScrollView` instances,
    we can throw an error asking to set this property to `false` for all, but the
    actual "content `ScrollView`". We should decide on default value of `isContentScrollView`
    based on what we estimate as more frequent use case.

    Alternatively, we can expose two components: `ScrollViewMarker` & `ContentScrollViewMarker`,
    which will handle that prop configuration for the user? Let's discuss below.

Beside adding a component I've also implemented basic support for scroll edge effects
configuration.

`RNSScrollViewSeeking` protocol is introduced. Other component can implement it to be
notified when `ScrollViewMarker` is attached to view hierarchy. Currently it is not
used anywhere & it's shape is not final, but it serves PoC purposes.

Closes <https://github.com/software-mansion/react-native-screens-labs/issues/981>

## Caveats

* <https://github.com/software-mansion/react-native-screens/pull/3674/commits/a84419e015eaa3fbe1fcc0030173ba762eba0910>

    Similarly to iOS, I use UIManagerListener.didMountItems to detect a
    moment when the view hierarchy is assembled.

    > [!caution]
    > I've tested & views are created in bottom-up order -> that means the
    > `ScrolLViewMarker` will usually be created BEFORE e.g. a `StackHost`,
    > therefore its `UIManagerListener` listener will be added first -> it
    > will be called before listener of `StackHost`. If for some reason order
    > of view creation changes, e.g. `ScrollViewMarker` will be rendered in
    > consecutive render, the order here will also change.

## Visual documentation

N/A

## Test plan

It's rather hard to test this component in isolation & somehow observe an effect
in JS. If someone has an idea - let me know.

Right now, I've introduced `test-svm-configures-scroll-view`, which uses
`StackContainer` to make scroll edge effects visible, but mostly this test is about prepared environment for testing on native side. Especially on Android, where there are no edge effects.

## Checklist

* [x] Updated / created local changelog entries in relevant test files.
* [x] Ensured that CI passes


ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
```diff
@@ -0,0 +1,228 @@
...
+
+- (void)mountChildComponentView:(UIView<RCTComponentViewProtocol> *)childComponentView index:(NSInteger)index
+{
+  [super mountChildComponentView:childComponentView index:index];
+}
+
+- (void)unmountChildComponentView:(UIView<RCTComponentViewProtocol> *)childComponentView index:(NSInteger)index
+{
+  [super unmountChildComponentView:childComponentView index:index];
+}
```
kkafar, line 237: I've left these for easier debugging.

android/src/main/java/com/swmansion/rnscreens/gamma/scrollviewmarker/ScrollViewMarker.kt
```diff
@@ -0,0 +1,88 @@
...
+    private val reactContext: ThemedReactContext,
+) : ReactViewGroup(reactContext),
+    UIManagerListener {
+    init {
+        // We're adding ourselves during a batch, therefore we expect to receive its finalization callbacks
+        val uiManager =
+            checkNotNull(UIManagerHelper.getUIManager(reactContext, UIManagerType.FABRIC)) {
+                "[RNScreens] UIManager must not be null."
+            }
+        uiManager.addUIManagerEventListener(this)
```
t0maboro: I see we're adding self as a listener, but we never remove, is this intentional?

ios/conversion/RNSConversions-ScrollViewMarker.mm
```diff
@@ -0,0 +1,44 @@
+#import "RNSConversions-ScrollViewMarker.h"
+
+#if RCT_NEW_ARCH_ENABLED && RNS_GAMMA_ENABLED
+
+#import <React/RCTLog.h>
+
+namespace react = facebook::react;
+
+#define SWITCH_EDGE_EFFECT(X)                              \
```
t0maboro, line 9: I see the same code across 3 files responsible for conversion, not an issue, because we always undef, but I'll consider another approach that will allow us to extract it to a single private common file for conversions.

FabricExample/metro.config.js
```diff
@@ -9,7 +9,8 @@ const { getDefaultConfig, mergeConfig } = require('@react-native/metro-config');
 
 const fs = require('fs');
 const path = require('path');
-const exclusionList = require('metro-config/private/defaults/exclusionList').default;
+const exclusionList =
```
kmichalikk, line 12: This file has only prettier changes, right? maybe we shouldn't include it in the PR?

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.h
```diff
@@ -0,0 +1,23 @@
...
+
+@protocol RNSScrollViewSeeking <NSObject>
+
+/**
+ *  Call this method to register a ScrollView wrapped by marker with an interested component (receiver).
+ */
+- (void)registerDescendantScrollView:(nonnull UIScrollView *)scrollView
+                          fromMarker:(nonnull RNSScrollViewMarkerComponentView *)marker;
+
+@end
```
kligarski: I'd consider moving this protocol to separate file for consistency.

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
```diff
@@ -0,0 +1,228 @@
...
+      return static_cast<id<RNSScrollViewSeeking>>(superview);
+    }
+    superview = superview.superview;
+  }
+  return nil;
+}
+
+- (void)maybeRegisterWithSeekingAncestor
+{
+  if (_hasAttemptedRegistration) {
```
kmichalikk, line 89: Is this always the case that one attempt is enough? I'm thinking about view recycling, hot reloads, etc

android/src/main/java/com/swmansion/rnscreens/gamma/scrollviewmarker/ScrollViewSeeking.kt
```diff
@@ -0,0 +1,13 @@
+package com.swmansion.rnscreens.gamma.scrollviewmarker
+
+import android.view.ViewGroup
+
+interface ScrollViewSeeking {
+    // scrollView is a ViewGroup, because there is no universal ScrollView component on Android.
```
kligarski, line 6:
```suggestion
    // ScrollView is a ViewGroup, because there is no universal ScrollView component on Android.
```
nit

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
```diff
@@ -0,0 +1,228 @@
...
+ * This method throws an error in debug mode in case it fails to find the ScrollView instance,
+ * as it does not make sense to use this component if the ScrollView is not there.
+ */
+- (nullable UIScrollView *)findScrollView
+{
+  for (UIView *subview in self.subviews) {
+    if ([subview isKindOfClass:RCTScrollViewComponentView.class]) {
+      return static_cast<RCTScrollViewComponentView *>(subview).scrollView;
+    }
+  }
```
kmichalikk: Should we give a warning if multiple scrollviews are present directly under the marker?

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
```diff
@@ -0,0 +1,228 @@
...
+  _hasAttemptedRegistration = YES;
+  [self registerWithSeekingAncestor];
+}
+
+- (void)registerWithSeekingAncestor
+{
+  UIScrollView *scrollView = [self findScrollView];
+
+  if (scrollView == nil) {
+    // Should we crash the app here?
```
kmichalikk: at least we should give a warning, which we do

please, remove the comment before merging, or reword it to something more definite

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
```diff
@@ -0,0 +1,228 @@
...
+    return;
+  }
+
+  id<RNSScrollViewSeeking> seekingAncestor = [self findSeekingParent];
```
kmichalikk: here we assume that there is only one seeking parent, what if there are multiple? for instance, stack and tabs?

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
kmichalikk: Maybe we can wrap them in some debug mode macro?

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
kmichalikk: ok, shouldBeRecycled is false, so I answered myself partially, but still, have you considered other cases?

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
kligarski: 1. Should we search for scroll view in subviews directly below the marker or in first descendant chain from the marker? What if there is a custom list like FlatList or something that has some wrapper components around the actual ScrollView? I know that there usually is a way to pass the ScrollView component and we can wrap that but is it always the case?
2. Do we assume that it's always going to be `RCTScrollViewComponentView`? So we don't allow any custom native component that uses `UIScrollView`? If so, maybe this should be reflected in method name/signature?

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
kmichalikk: For instance, if there was a list that fetches content from server, and has some placeholder until the content loads, and in the meantime, there is no ScrollView rendered? we should make it explicit that the marker needs to have scrollview under it at all times

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
kligarski: I guess so. We can change it back to just returning nil if there is a valid use case for it.

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
```diff
@@ -0,0 +1,228 @@
...
+  }
+  RCTAssert(false, @"[RNScreens] Failed to find ScrollView"); // debug assertion only
+  return nil;
+}
+
+- (nullable id<RNSScrollViewSeeking>)findSeekingParent
+{
+  const UIView *superview = self.superview;
+  while (superview != nil) {
+    if ([superview respondsToSelector:@selector(registerDescendantScrollView:fromMarker:)]) {
```
kligarski, line 79: I wonder if we should use `rnscreens_` prefix for protocol methods if we're using `respondsToSelector` to check conformance.

FabricExample/__tests__/App.test.tsx
```diff
@@ -2,7 +2,6 @@
  * @format
  */
 
-
```
kmichalikk, line 5: this could also be removed from the PR

apps/src/tests/single-feature-tests/scroll-view-marker/test-svm-configures-scroll-view.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+function ContentScreen() {
+  return (
+    <View
+      style={[
+        styles.container,
+        styles.fillParent,
+        { backgroundColor: Colors.White },
+      ]}>
+      <Text>Hello world</Text>
```
kmichalikk: For future us, we could write some explanation in the <text>, i.e. "Text component that appears as the first child and breaks first descendant heuristic for content ScrollView"

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
```diff
@@ -0,0 +1,228 @@
...
+- (void)configureScrollView:(nonnull UIScrollView *)scrollView
+{
+  [RNSScrollEdgeEffectApplicator applyToScrollView:scrollView withProvider:self];
+}
+
+#pragma mark - Override
+
+// TODO: This will be way too late to configure options etc.
+// Potentially we want to run in the end of transaction, before containers are updated.
+- (void)willMoveToWindow:(UIWindow *)newWindow
```
kligarski, line 148: Do we still need it if we're using 
```
- (void)mountingTransactionDidMount:(const facebook::react::MountingTransaction &)transaction
               withSurfaceTelemetry:(const facebook::react::SurfaceTelemetry &)surfaceTelemetry```

File: ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm [hunk 15]
```diff
@@ -0,0 +1,228 @@
...
+
+- (void)finalizeUpdates:(RNComponentViewUpdateMask)updateMask
+{
+  // This assumes that in first render children are mounted before props are updated.
+  // Also this handles only first render & prop value update. It does not handle case
+  // where child scrollview potentially changes. Separate question is whether we even
+  // want to handle such case.
+  if (_needsEdgeEffectUpdate) {
+    _needsEdgeEffectUpdate = NO;
+    [self configureScrollView:[self findScrollView]];
```
kligarski, line 217: Should we perform `findScrollView` each time prop is updated? I think that registering should be synchronized with prop updates - either we handle ScrollView changes for both or not at all.

If we don't want to support ScrollView changes, then maybe we can cache it instead of calling `findScrollView` each time prop is changed?

apps/src/tests/single-feature-tests/scroll-view-marker/test-svm-configures-scroll-view.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { generateNextColor } from '../../../shared/utils/color-generator';
+
+const SCENARIO: Scenario = {
+  name: 'Basic functionality',
+  key: 'test-svm-configures-scroll-view',
+  details:
+    'Allows to test the basic functionality of ScrollViewMarker component.' +
+    'It utilizes the StackContainer, to allow for observation of edge effects' +
+    'applied to the container edges. On Android this test serves only as a setup' +
+    'for native debugging.',
```
kligarski:
```suggestion
    'Allows to test the basic functionality of ScrollViewMarker component. ' +
    'It utilizes the StackContainer, to allow for observation of edge effects ' +
    'applied to the container edges. On Android this test serves only as a setup ' +
    'for native debugging.',
```

apps/src/tests/single-feature-tests/scroll-view-marker/test-svm-configures-scroll-view.tsx
```diff
@@ -0,0 +1,80 @@
...
+  platforms: ['ios'],
```
kligarski:
```suggestion
  platforms: ['android', 'ios'],
```

android/src/main/java/com/swmansion/rnscreens/gamma/scrollviewmarker/ScrollViewSeeking.kt
```diff
@@ -0,0 +1,13 @@
...
+    // It might a be ScrollView, NestedScrollView (different inheritance hierarchies) or any other
+    // ViewGroup that implements appropriate scrolling interfaces.
+    fun registerScrollView(
+        maker: ScrollViewMarker,
```
kligarski:
```suggestion
        marker: ScrollViewMarker,
```

android/src/main/java/com/swmansion/rnscreens/gamma/scrollviewmarker/ScrollViewMarker.kt
```diff
@@ -0,0 +1,88 @@
...
+        findSeekingParent()?.registerScrollView(this, scrollView)
+    }
+
+    private fun maybeRegisterWithSeekingAncestor() {
+        if (hasAttemptedRegistration) {
+            return
+        }
+
+        registerWithSeekingAncestor()
+        hasAttemptedRegistration = true
```
kligarski, line 71: nit: for consistency with iOS implementation, I'd swap those.

android/src/main/java/com/swmansion/rnscreens/gamma/scrollviewmarker/ScrollViewMarker.kt
```diff
@@ -0,0 +1,88 @@
...
+
+        while (currentView != null) {
+            if (currentView is ScrollViewSeeking) {
+                return currentView
+            }
+            currentView = currentView.parent
+        }
+
+        return null
+    }
```
kligarski: I'm not sure if this works as we intend. I tried `test-svm-configures-scroll-view` and we reach only `StackScreen` which isn't attached to container yet.

kligarski: Similar thing happens on iOS.

android/src/main/java/com/swmansion/rnscreens/gamma/scrollviewmarker/ScrollViewMarker.kt
kkafar: No! Good catch. Thanks! I think we have the same problem in `StackHost`. I'll patch that there in separate PR.

kkafar: https://github.com/software-mansion/react-native-screens/pull/3674/commits/871e7d5a1e8e951dfcccf444ef8ce855fbbc5494

I'm not sure whether this constitutes a retain cycle, as it makes the UIManager to retain the view explicitly, but I'm not sure how could view retain view manager (maybe via context)? Anyway, it makes sense to remove the listener once it is no longer needed.z

ios/conversion/RNSConversions-ScrollViewMarker.mm
kkafar: If I move the edge effect props from tabs to the `ScrollViewMarker` component - that code will me removed completely from tabs conversion code.

FabricExample/metro.config.js
kkafar: See description of https://github.com/software-mansion/react-native-screens/pull/3674/commits/038613971885469da126bc66ca55dd4ce6f1bb90

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
kkafar: Yeah, it throws an error in case there is no scrollview found + additionally we'll describe it in comments. The error is thrown only in debug mode though. Not sure whether it is enough. 

I didn't consider any additional cases. This is a PoC. I plan to work further on this component.

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
kkafar: Yeah, good idea.

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
kkafar: We just look for the the first one. I agree, that the method name could reflect that. Thank you!

FabricExample/__tests__/App.test.tsx
kkafar: Linter

apps/src/tests/single-feature-tests/scroll-view-marker/test-svm-configures-scroll-view.tsx
kkafar: Absolutely +1, thanks

kkafar: https://github.com/software-mansion/react-native-screens/pull/3674/commits/6f9c85f962926537cc50dc51524e6af43fe25180

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
kkafar: https://github.com/software-mansion/react-native-screens/pull/3674/commits/1bee8bd3afa5cd80750de4516fc5643ac14706dc

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
kkafar: https://github.com/software-mansion/react-native-screens/pull/3674/commits/2cff5cf6a5ee4317e69219c17365040ae1ff82bd

android/src/main/java/com/swmansion/rnscreens/gamma/scrollviewmarker/ScrollViewSeeking.kt
kkafar: This refers to the method parameter name

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
kkafar: I've updated this code. Please take a look there again. 

The idea is that we'll try to add support for each "type of scroll view" that are commonly used separately. See the code, it should be clear now.

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
kkafar: I decided against. Just not applying edge effects is better than crashing the app in production.

kkafar: Android code expresses different approach, but I wish I had "debug mode only " asserts there.

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.h
kkafar: https://github.com/software-mansion/react-native-screens/pull/3674/commits/e15d99efa0475bec1f9e9dcde9ad1b2feb9d05cd

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
kkafar: It's rather an uncommon practice. Only place where I'm aware this is recommended is method swizzling / categories on "foreign classes".

android/src/main/java/com/swmansion/rnscreens/gamma/scrollviewmarker/ScrollViewSeeking.kt
kligarski: Fair point.

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
kkafar: > Should we perform findScrollView each time prop is updated? I think that registering should be synchronized with prop updates - either we handle ScrollView changes for both or not at all.

I do not understand what you mean in this part. Please elaborate.

kkafar: > If we don't want to support ScrollView changes, then maybe we can cache it instead of calling findScrollView each time prop is changed?

I don't want to be responsible for any reference to the scrollview. Computing it is cheap. We assume there can be at most a single child attached.

android/src/main/java/com/swmansion/rnscreens/gamma/scrollviewmarker/ScrollViewMarker.kt
kkafar: Swapped it in iOS implementation.

android/src/main/java/com/swmansion/rnscreens/gamma/scrollviewmarker/ScrollViewMarker.kt
kkafar: How high you reach, depends on when you call this method. It's outside of the scope of this method to handle it. 

Likely it'll be enough - as all screen components will implement `RNSScrollViewSeeking`, if not we'll change it on demand. There is no RFC for this yet & as the PR description states - it's a PoC.

kkafar:
Regarding support for legacy arch - no, ~I'll add stubs to make it build, but I won't implement any functionality there. Next minor won't support it anyway.~ I don't plan to add any support. I'm gonna ignore the failing CI in this PR. Next minor is not gonna support paper anyway.

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
kligarski: What I'm mostly concerned about it the possible inconsistency where scroll view A has registered with `RNSScrollViewSeeking` parent, then scroll view A is unmounted and scroll view B is mounted. Now, marker can impact scroll view B's scroll edge effects but scroll view A remains "registered" instead of scroll view B.

package.json
```diff
@@ -220,6 +220,9 @@
         },
         "RNSSafeAreaView": {
           "className": "RNSSafeAreaViewComponentView"
+        },
+        "RNSScrollViewMarker": {
+          "className": "RNSScrollViewMarkerComponentView"
```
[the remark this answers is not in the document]
kkafar: https://github.com/software-mansion/react-native-screens/pull/3674/commits/6c1a47ca5820cb09366e80a7b521534c75291c1d

android/src/main/java/com/swmansion/rnscreens/gamma/scrollviewmarker/ScrollViewMarker.kt
```diff
@@ -0,0 +1,88 @@
...
+     * Call it only after the children have been already attached and not yet detached.
+     */
+    private fun findScrollView(): ViewGroup {
+        val childScrollView =
+            checkNotNull(children.find { childView -> childView is ScrollView || childView is NestedScrollView }) {
+                "[RNScreens] Failed to find supported type of ScrollView in children of ScrollViewMarker"
+            }
+
+        return childScrollView as ViewGroup
+    }
```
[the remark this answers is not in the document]
kkafar: https://github.com/software-mansion/react-native-screens/pull/3674/commits/80e92304e0147281e98b2fdd4de874814fe9824a

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
kmichalikk: We have some instances of using it like that

https://github.com/search?q=repo%3Asoftware-mansion%2Freact-native-screens+rnscreens_+language%3AObjective-C+OR+Objective-C%2B%2B&type=code

but maybe in this case the method name is not as common so we're okay probably

ios/gamma/scroll-view-marker/RNSScrollViewMarkerComponentView.mm
kmichalikk: I think it goes together with https://github.com/software-mansion/react-native-screens/pull/3674/changes#r2845771822

Maybe it's safe to assume that if the ScrollView changed, then the other one must have been destroyed?

kmichalikk: It's also up to the view that implements ScrollViewSeeking whether it will store a reference to it, and it could be risky to assume that it won't. Don't know if this is a problem, just writing this out so it gets attention.

PR #3676, opened 2026-02-22
fix(iOS): prevent search bar from reopening keyboard on header button press

### LINKED ISSUE #3677: iOS: Header button reopens keyboard after search bar is dismissed
## Description

When a screen has both a `headerSearchBarOptions` search bar and header buttons (e.g. `headerRight`), tapping a header button after dismissing the search bar keyboard causes the keyboard to reopen unexpectedly.

## Video

https://github.com/user-attachments/assets/ca767a81-25c9-45a7-98a9-dd3f3090680d

## Steps to reproduce

1. Have a screen with both a search bar (`headerSearchBarOptions`) and a `headerRight` button
2. Tap the search bar to focus it (keyboard opens)
3. Dismiss the keyboard (tap Cancel or tap away)
4. Tap the header right button
5. **Bug:** The keyboard opens again instead of just triggering the button action

## Expected behavior

Tapping the header button after dismissing the search bar should only trigger the button's action, without reopening the keyboard.

## Actual behavior

The keyboard reopens when tapping the header button.

## Minimal reproducible

https://github.com/sbkl/sdk-55-header-right-button-opening-keyboard

Also reported on Expo: https://github.com/expo/expo/issues/43198

## Root cause

1. `RNSScreenStackHeaderConfig.mm`: `updateViewController` unconditionally reassigns `navitem.searchController = searchBar.controller` on every React re-render. UIKit interprets this as a new search controller being set and re-activates the search bar, which reopens the keyboard.
2. `RNSSearchBar.mm`: `searchBarCancelButtonClicked:` calls `resignFirstResponder` on the `RNSSearchBar` wrapper but not on the actual `UISearchBar` (`_controller.searchBar`), so the native search bar may retain first responder status internally.

## Fix

PR: #3676

## Environment

```
react-native-screens: 4.23.0
react-native: 0.83.2
iOS: 26.2
Xcode: 26.2/17C52
Expo SDK: 55.0.0-preview.11
expo-router: 55.0.0-preview.8
```

just1and0:
## Description

Closes #3677.

When a screen has both a `headerSearchBarOptions` search bar and header buttons (e.g. a native UIMenu via `unstable_headerRightItems`), tapping a header button after dismissing the search bar keyboard causes the keyboard to reopen unexpectedly.

**Repro steps:**
1. Have a screen with both a search bar and a header right menu button
2. Tap the search bar to focus it (keyboard opens)
3. Dismiss the keyboard (tap Cancel or tap away)
4. Tap the header menu button
5. **Bug:** The keyboard opens again instead of just triggering the button action

Minimal reproduction: https://github.com/sbkl/sdk-55-header-right-button-opening-keyboard

Also reported on Expo: https://github.com/expo/expo/issues/43198

## Root cause

`searchBarTextDidBeginEditing:` calls `[self becomeFirstResponder]` on the `RNSSearchBar` wrapper, but `searchBarTextDidEndEditing:` never called `[self resignFirstResponder]`. When the user dismisses the keyboard by tapping away (not via the cancel button), `RNSSearchBar` remains first responder. UIMenu's pull-down presentation suppresses (rather than resigns) the current first responder, and when the menu dismisses, UIKit restores it — causing the keyboard to reappear.

## Changes

Three fixes in the iOS native code:

1. **`RNSSearchBar.mm`**: In `searchBarTextDidEndEditing:`, call `[self resignFirstResponder]` so the wrapper properly gives up first responder status whenever editing ends, regardless of how it was dismissed. This is the key fix for the UIMenu case.

2. **`RNSSearchBar.mm`**: In `searchBarCancelButtonClicked:`, also call `resignFirstResponder` on the actual `_controller.searchBar` (the native `UISearchBar`), not just on `self` (the `RNSSearchBar` wrapper). This ensures the search bar fully gives up keyboard focus when cancelled.

3. **`RNSScreenStackHeaderConfig.mm`**: Guard the `navitem.searchController` assignment to only occur when the value actually changes. Previously, this was unconditionally reassigned on every `updateViewController` call, which could cause UIKit to re-activate the search bar.

### Before (bug)

https://github.com/user-attachments/assets/ca767a81-25c9-45a7-98a9-dd3f3090680d

After focusing and dismissing the search bar, tapping the header menu button reopens the keyboard.

### After (fix)

After the fix, tapping the header menu button correctly triggers the menu action without reopening the keyboard.

## Test plan

- Added `Test3677.tsx` in `apps/src/tests/issue-tests/`
- The test uses a native UIMenu via `unstable_headerRightItems` to reproduce the original issue
- Steps to test:
  1. Open the `Test3677` test screen (has a search bar + ellipsis menu header right button)
  2. Tap the search bar to focus it
  3. Dismiss the keyboard
  4. Tap the ellipsis menu button in the header
  5. Verify the keyboard does **not** reopen and the menu appears correctly

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/apps/src/tests/issue-tests/index.ts
+++ b/apps/src/tests/issue-tests/index.ts
@@ -201,5 +201,5 @@ export { default as TestBottomTabsOrientation } from './TestBottomTabsOrientatio
 export { default as TestScreenStack } from './TestScreenStack';
 export { default as TestSplit } from './TestSplit';
 export { default as TestSafeAreaViewIOS } from './TestSafeAreaViewIOS';
-export { default as TestSearchBarKeyboardReopen } from './TestSearchBarKeyboardReopen';
+export { default as Test3677 } from './Test3677';
 export { default as TestStackNesting } from './TestStackNesting';
--- a/apps/src/tests/issue-tests/Test3677.tsx
+++ b/apps/src/tests/issue-tests/Test3677.tsx
@@ -40,13 +40,30 @@ function Home({ navigation }: StackNavigationProp) {
         hideWhenScrolling: false,
         onChangeText: event => setSearchQuery(event.nativeEvent.text),
       },
-      headerRight: () => (
-        <Button
-          title="Menu"
-          onPress={() => Alert.alert('Menu', 'Header button pressed')}
-          testID="home-header-right-button"
-        />
-      ),
+      // @ts-ignore – unstable API
+      unstable_headerRightItems: () => [
+        {
+          type: 'menu',
+          label: 'Menu',
+          icon: { type: 'sfSymbol', name: 'ellipsis.circle' },
+          menu: {
+            items: [
+              {
+                label: 'Settings',
+                type: 'action',
+                icon: { type: 'sfSymbol', name: 'gear' },
+                onPress: () => Alert.alert('Settings', 'Settings pressed'),
+              },
+              {
+                label: 'Refresh',
+                type: 'action',
+                icon: { type: 'sfSymbol', name: 'arrow.triangle.2.circlepath' },
+                onPress: () => Alert.alert('Refresh', 'Refresh pressed'),
+              },
+            ],
+          },
+        },
+      ],
     });
   }, [navigation]);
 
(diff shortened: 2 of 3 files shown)


apps/src/tests/issue-tests/index.ts
```diff
@@ -201,4 +201,5 @@ export { default as TestBottomTabsOrientation } from './TestBottomTabsOrientatio
 export { default as TestScreenStack } from './TestScreenStack';
 export { default as TestSplit } from './TestSplit';
 export { default as TestSafeAreaViewIOS } from './TestSafeAreaViewIOS';
+export { default as TestSearchBarKeyboardReopen } from './TestSearchBarKeyboardReopen';
```
kmichalikk: nit: please use Test3676 or Test3677

t0maboro:
As mentioned in the ticket: https://github.com/software-mansion/react-native-screens/issues/3677#issuecomment-4067063758 . I don't suspect `react-native-screens` to cause an issue here, because I can reproduce a similar issue without our search bar component. It seems like the underlying problem is somewhere in the RN core, regarding focus & keyboard handling logic.

I'm closing this PR.

PR #3678, opened 2026-02-23
chore: Simplify single test import in App.tsx

kmichalikk:
## Description

This PR aims to simplify importing of single tests instead of the whole example app. Depending on which test needs to be selected, part of the `apps/App.tsx` file needs to be uncommented & modified. The whole Example is selected by default.

## Changes

Modify typing & imports for tests.

## Checklist

### DIFF
--- a/apps/App.tsx
+++ b/apps/App.tsx
@@ -1,20 +1,17 @@
 import React from 'react';
 import { enableFreeze } from 'react-native-screens';
-// import { NavigationContainer } from '@react-navigation/native';
 
-// import { COMPONENT_SCENARIOS as Integration } from './src/tests/component-integration-tests';
-// import { COMPONENT_SCENARIOS as Feature } from './src/tests/single-feature-tests';
-// import * as Issue from './src/tests/issue-tests';
 import Example from './Example';
+// import { NavigationContainer } from '@react-navigation/native';
+// import { Tests } from './src/tests';
 
 enableFreeze(true);
 
 export default function App() {
   return <Example />;
-  // return <Issue.TestBottomTabs />;
   // return (
   //   <NavigationContainer>
-  //     <Integration.Orientation.scenarios.StackInTabs.AppComponent />
+  //     <Tests.Issue.TestBottomTabs />
   //   </NavigationContainer>
   // );
 }
--- a/apps/src/tests/index.ts
+++ b/apps/src/tests/index.ts
@@ -1,0 +1,9 @@
+import { COMPONENT_SCENARIOS as Feature } from './single-feature-tests';
+import { COMPONENT_SCENARIOS as Integration } from './component-integration-tests';
+import * as Issue from './issue-tests';
+
+export const Tests = {
+  Feature,
+  Integration,
+  Issue,
+};


apps/src/tests/importer.ts
kkafar, line 1: Why separate file and not just reexport these from index?

apps/App.tsx
```diff
@@ -1,11 +1,19 @@
 import React from 'react';
 import { enableFreeze } from 'react-native-screens';
+
 import Example from './Example';
-// import * as Test from './src/tests/issue-tests';
+
+// import * as Test from './src/tests/importer';
+// import { NavigationContainer } from '@react-navigation/native';
+// const Example = Test.CIT.Orientation.scenarios.StackInTabs.AppComponent;
```
kkafar: Maybe let's import & allow for usage of `Tests.Integration`, `Tests.Issue` / `Tests.Feature` objects, so that e.g. I render appropriate component like this:

`return <Tests.Integration.Orientation.scenarios.StackInTabs.AppComponent />`

It's a bit ugly, but at least we can use the tests directly. 

What do you think?

apps/App.tsx
```diff
@@ -1,11 +1,19 @@
...
 enableFreeze(true);
 
 export default function App() {
   return <Example />;
-  // return <Test.TestBottomTabs />;
+  // return (
+  //   <NavigationContainer>
+  //     <Example />
+  //   </NavigationContainer>
+  // );
```
kkafar, line 16: Why do we want to wrap in `NavigationContainer`?

apps/src/tests/importer.ts
kmichalikk: I'm not sure what you mean here, there was no index in this directory. I removed the importer file, and the files are imported directly in App.tsx now

apps/App.tsx
kmichalikk: I don't know. This file is only meant to be functional, to quickly swap tests, so it's not a big deal if it's ugly, I think.

apps/App.tsx
kmichalikk: SFTs and CITs sometimes contain react-navigation components, but they're not wrapped in NavigationContainer, only the selection screens are. Since we bypass the selection here, we need to manually wrap them.

kkafar: I get it but that is really unfortunate. The examples should be autonomous as defined here: https://github.com/software-mansion/react-native-screens/blob/a6a75752fc8c06131f3dd4c315ae878f191b8715/apps/src/tests/shared/helpers.ts#L23-L28

This is clearly with violation & that's why we have the problem. 

Right now, I think our course of action should be to allow rendering `NavigationContainer` here, but down the line, ideally we don't have tests that use `react-navigation` at all. We'll see how it goes. Thanks for now.

apps/App.tsx
```diff
@@ -1,11 +1,20 @@
 import React from 'react';
 import { enableFreeze } from 'react-native-screens';
+// import { NavigationContainer } from '@react-navigation/native';
+
+// import { COMPONENT_SCENARIOS as Integration } from './src/tests/component-integration-tests';
+// import { COMPONENT_SCENARIOS as Feature } from './src/tests/single-feature-tests';
+// import * as Issue from './src/tests/issue-tests';
```
kkafar: I imagined that we can simply export a single `Tests` object form `./src/tests/index.ts`. I'll see to it.

t0maboro:
seems like it doesn't work correctly in some cases

https://github.com/user-attachments/assets/db90be2c-f621-4292-82d9-c088832a40cd

PR #3681, opened 2026-02-23
feat(iOS, Tabs): add userInterfaceStyle prop to enforce native tabs color scheme

AlshehriAli0:
## Description

Adds a `userInterfaceStyle` prop to `Tabs.Host` that forces the tab bar appearance regardless of the system color scheme.

Many apps intentionally use light-only layouts even when the device is in dark mode. Without this prop, the native tab bar briefly flashes with dark mode colors on every navigation before settling to the correct appearance (sometimes needing user input/scroll to settle correctly), since `overrideUserInterfaceStyle` was previously only applied per-screen after the mounting transaction.

This prop gives developers the high control: lock the tab bar to a specific appearance so it never fights the app's intended design, regardless of the system setting.

## Changes

- Added `userInterfaceStyle` prop to `TabsHostProps` TypeScript types (`'unspecified' | 'light' | 'dark'`)
- Added prop to codegen native component spec
- Added iOS conversion function for the new enum
- Set `overrideUserInterfaceStyle` on the controller in `updateProps` (Fabric) and Paper setter
- Added test scario: Tabs > User Interface Style

https://github.com/user-attachments/assets/8ee558ce-4892-420d-a0cf-b8f3e88751de

1. Set device/simulator to **dark mode**
2. Open the test app → Tabs → **User Interface Style**
3. Select **Light** — the tab bar should stay light with no flash
4. Navigate between tabs — no dark mode flash should occur
5. Select **Dark** — tab bar forces dark appearance
6. Select **Unspecified** — tab bar follows system setting

## Platforms

| Platform | Supported |
|----------|-----------|
| iOS      | ✅        |
| Android  | ❌ (N/A)  |

### DIFF
--- a/apps/src/tests/issue-tests/TestBottomTabs/index.tsx
+++ b/apps/src/tests/issue-tests/TestBottomTabs/index.tsx
@@ -236,6 +236,7 @@ function App() {
         tabBarItemTitleFontWeight="700"
         tabBarItemLabelVisibilityMode="auto"
         tabBarMinimizeBehavior="onScrollDown"
+        userInterfaceStyle="light"
       />
     </ConfigWrapperContext.Provider>
   );
--- a/apps/src/tests/issue-tests/TestBottomTabs/tabs/Tab1.tsx
+++ b/apps/src/tests/issue-tests/TestBottomTabs/tabs/Tab1.tsx
@@ -1,11 +1,10 @@
 import React from 'react';
 import { CenteredLayoutView } from '../../../../shared/CenteredLayoutView';
 import { TabContentView } from '../components/TabContentView';
-import Colors from '../../../../shared/styling/Colors';
 
 export function Tab1() {
   return (
-    <CenteredLayoutView style={{ backgroundColor: Colors.OffWhite }}>
+    <CenteredLayoutView style={{ backgroundColor: 'white' }}>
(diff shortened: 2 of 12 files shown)


AlshehriAli0:
@t0maboro @kkafar @kmichalikk

kkafar:
Hey! Thanks for the PR. It'll wait a couple of more days. We're currently in process of Tabs API stabilisation & we're debating exactly this topic - how do we wanna approach theme management. We'll get back to you after we settle our internal discussions. This will land in one form or another.

AlshehriAli0:
> Hey! Thanks for the PR. It'll wait a couple of more days. We're currently in process of Tabs API stabilisation & we're debating exactly this topic - how do we wanna approach theme management. We'll get back to you after we settle our internal discussions. This will land in one form or another.

Thanks for the update, sounds good 👍
Happy to wait. Just let me know once you’ve aligned internally.

kligarski:
Hi, @AlshehriAli0. We researched handling color scheme for Tabs on both Android and iOS. As a result, I created a PR with `colorScheme` property on iOS: https://github.com/software-mansion/react-native-screens/pull/3716. This will work exactly like the change proposed in your PR so I added you as co-author there.

We'll close this PR then. Thank you for contributing.

PR #3683, opened 2026-02-24
refactor(JS, Tabs): JS layer platform separation for TabsHost and TabsScreen

t0maboro:
Please read this [RFC-994](https://github.com/software-mansion/react-native-screens-labs/blob/92ff0f52aad6135730ef55cf7f476051e147e005/rfcs/0994-tabs-platform-separation.md) as an introcution why we're refactoring this code.

## Description

This PR introduces a strict platform separation for the `TabsHost` and `TabsScreen` components. The platform-specific code is now placed in a separate file for each platform.

> [!NOTE]  
> This is the 1st part of the refactor. As a follow-up, the underlying native components will also be split. That second step will allow us to completely filter out props that do not belong to a given platform.

> [!CAUTION]
> This is a breaking change to the public API.

---

(Description copied from #3684)

It focuses on the split of the native components, so that the specs contain only the definitions of shared and platform-specific props.

From now on, platform-specific props should not be passed to the native side of components on other platforms. If a prop will have an empty implementation on the native side, it means that it was defined in the wrong location.

This refactor also unlocks further cleanup of the codebase, as we no longer need to override setters with an empty implementation on Android. 

> [!CAUTION]
> Since we're no longer supporting old architecture, I didn't take any steps to synchronize these changes in it. In fact, the app most likely won't even build on Paper at this point.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/997

## Changes

From this PR:
- Replaced `TabsX.tsx` with `TabsX.ios.tsx` and `TabsX.android.tsx` files. 
- Grouped platform-specific props into `ios` and `android` objects.
- Moved shared logic into custom `useTabsX` hooks.
- Added TS declaration files to satisfy typechecks, to allow importing components without platform extensions
- Updated docs, removed platform mentions, as they do not provide any extra information, when the prop is under the specific section (top-level or inside android/ios objects)
- Updated examples

From #3684 :
- From now on, we have 2 separate native components, each dedicated for a specific platform: RNSTabsHostAndroid & RNSTabsHostIOS (same approach for TabsScreen components)
- All iOS-specific prop setters were removed from ViewManagers on Android.
- `RNSTabsHostShadowNode` requires a platform-specific name - without that, app is crashing because the Host component hasn't been found in the registry.
- All cpp enums and classes generated by codegen were updated for iOS conversions: `react::RNSTabsScreenX` -> `react::RNSTabsScreenIOSX`
[...]

PR #3684, opened 2026-02-25
refactor(Tabs): Separate native components for TabsHost and TabsScreen

t0maboro:
Please read this [RFC-994](https://github.com/software-mansion/react-native-screens-labs/blob/92ff0f52aad6135730ef55cf7f476051e147e005/rfcs/0994-tabs-platform-separation.md) as an introcution why we're refactoring this code.

## Description

This PR is the 2nd part of the refactor for separating the internal implementation for tabs. It focuses on the split of the native components, so that the specs contain only the definitions of shared and platform-specific props.

From now on, platform-specific props should not be passed to the native side of components on other platforms. If a prop will have an empty implementation on the native side, it means that it was defined in the wrong location.

This refactor also unlocks further cleanup of the codebase, as we no longer need to override setters with an empty implementation on Android. 

> [!CAUTION]
> Since we're no longer supporting old architecture, I didn't take any steps to synchronize these changes in it. In fact, the app most likely won't even build on Paper at this point.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/998

## Changes

- From now on, we have 2 separate native components, each dedicated for a specific platform: RNSTabsHostAndroid & RNSTabsHostIOS (same approach for TabsScreen components)
- All iOS-specific prop setters were removed from ViewManagers on Android.
- `RNSTabsHostShadowNode` requires a platform-specific name - without that, app is crashing because the Host component hasn't been found in the registry.
- All cpp enums and classes generated by codegen were updated for iOS conversions: `react::RNSTabsScreenX` -> `react::RNSTabsScreenIOSX`
- `componentNodeRef`, which before this PR was present in shared code for `TabsX` component (via `useTabsX` hook) was moved to the component's implementation, because types have diverged between platforms after splitting native components
- Props were reorganized and sorted in Specs, to have clean separation between common and platform-specifc definitions.
Screenshot 2026-02-26 at 23 02 38

N/A - all examples should behave the same as in the base PR.

## Test plan

Go through examples and verify that no regression has occurred.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/src/fabric/tabs/TabsHostIOSNativeComponent.ts
+++ b/src/fabric/tabs/TabsHostIOSNativeComponent.ts
@@ -21,13 +21,13 @@ type NativeFocusChangeEvent = {
 
 // #region iOS-specific helpers
 
-type TabBarMinimizeBehaviorIOS =
+type TabBarMinimizeBehavior =
   | 'automatic'
   | 'never'
   | 'onScrollDown'
   | 'onScrollUp';
 
-type TabBarControllerModeIOS = 'automatic' | 'tabBar' | 'tabSidebar';
+type TabBarControllerMode = 'automatic' | 'tabBar' | 'tabSidebar';
 
 // #endregion iOS-specific helpers
 
@@ -54,11 +54,8 @@ export interface NativeProps extends ViewProps {
   // #region iOS-specific Props
 
   tabBarTintColor?: ColorValue;
-  tabBarMinimizeBehavior?: CT.WithDefault<
-    TabBarMinimizeBehaviorIOS,
-    'automatic'
-  >;
-  tabBarControllerMode?: CT.WithDefault<TabBarControllerModeIOS, 'automatic'>;
+  tabBarMinimizeBehavior?: CT.WithDefault<TabBarMinimizeBehavior, 'automatic'>;
+  tabBarControllerMode?: CT.WithDefault<TabBarControllerMode, 'automatic'>;
 
   // #endregion iOS-specific Props
 }
--- a/src/fabric/tabs/TabsScreenAndroidNativeComponent.ts
+++ b/src/fabric/tabs/TabsScreenAndroidNativeComponent.ts
@@ -24,44 +24,44 @@ type LifecycleStateChangeEvent = Readonly<{
 
 // #region Android-specific helpers
 
-type TabBarItemLabelVisibilityModeAndroid =
+type TabBarItemLabelVisibilityMode =
   | 'auto'
   | 'selected'
   | 'labeled'
   | 'unlabeled';
 
-export type AppearanceAndroid = {
+export type Appearance = {
   tabBarBackgroundColor?: ProcessedColorValue | null;
   tabBarItemRippleColor?: ProcessedColorValue | null;
   tabBarItemLabelVisibilityMode?: CT.WithDefault<
-    TabBarItemLabelVisibilityModeAndroid,
+    TabBarItemLabelVisibilityMode,
     'auto'
   >;
 
-  itemColors?: BottomNavItemColorsAndroid;
-  activeIndicator?: ActiveIndicatorAppearanceAndroid;
-  typography?: TypographyAppearanceAndroid;
-  badge?: BadgeAppearanceAndroid;
+  itemColors?: BottomNavItemColors;
+  activeIndicator?: ActiveIndicatorAppearance;
+  typography?: TypographyAppearance;
+  badge?: BadgeAppearance;
 };
 
-type ItemStateColorsAndroid = {
+type ItemStateColors = {
   iconColor?: ProcessedColorValue | null;
   titleColor?: ProcessedColorValue | null;
 };
 
-type BottomNavItemColorsAndroid = {
-  normal?: ItemStateColorsAndroid;
-  selected?: ItemStateColorsAndroid;
-  focused?: ItemStateColorsAndroid;
-  disabled?: ItemStateColorsAndroid;
+type BottomNavItemColors = {
+  normal?: ItemStateColors;
+  selected?: ItemStateColors;
+  focused?: ItemStateColors;
+  disabled?: ItemStateColors;
 };
 
-type ActiveIndicatorAppearanceAndroid = {
+type ActiveIndicatorAppearance = {
   color?: ProcessedColorValue | null;
   enabled?: CT.WithDefault<boolean, true>;
 };
 
-type TypographyAppearanceAndroid = {
+type TypographyAppearance = {
   fontFamily?: string;
   fontSizeSmall?: CT.Float;
   fontSizeLarge?: CT.Float;
@@ -69,7 +69,7 @@ type TypographyAppearanceAndroid = {
   fontStyle?: string;
 };
 
-type BadgeAppearanceAndroid = {
+type BadgeAppearance = {
   textColor?: ProcessedColorValue | null;
   backgroundColor?: ProcessedColorValue | null;
 };
@@ -117,7 +117,7 @@ export interface NativeProps extends ViewProps {
   selectedImageIconResource?: ImageSource;
 
   // Appearance
-  standardAppearance?: UnsafeMixed<AppearanceAndroid>;
+  standardAppearance?: UnsafeMixed<Appearance>;
 
   // #endregion Android-specific Props
 }
--- a/src/fabric/tabs/TabsScreenIOSNativeComponent.ts
+++ b/src/fabric/tabs/TabsScreenIOSNativeComponent.ts
@@ -24,9 +24,9 @@ type LifecycleStateChangeEvent = Readonly<{
 
 // #region iOS-specific helpers
 
-type UserInterfaceStyleIOS = 'unspecified' | 'light' | 'dark';
+type UserInterfaceStyle = 'unspecified' | 'light' | 'dark';
 
-type BlurEffectIOS =
+type BlurEffect =
   | 'none'
   | 'systemDefault'
   | 'extraLight'
@@ -50,7 +50,7 @@ type BlurEffectIOS =
   | 'systemThickMaterialDark'
   | 'systemChromeMaterialDark';
 
-type OrientationIOS =
+type Orientation =
   | 'inherit'
   | 'all'
   | 'allButUpsideDown'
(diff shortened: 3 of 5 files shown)


src/fabric/tabs/TabsScreenAndroidNativeComponent.ts
```diff
@@ -0,0 +1,127 @@
...
+  selectedImageIconResource?: ImageSource;
+
+  // Appearance
+  standardAppearance?: UnsafeMixed<AppearanceAndroid>;
+
+  // #endregion Android-specific Props
+}
+
+export default codegenNativeComponent<NativeProps>('RNSTabsScreenAndroid', {
+  excludedPlatforms: ['iOS'],
```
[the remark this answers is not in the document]
t0maboro: https://github.com/facebook/react-native/blob/v0.84.0/packages/react-native/Libraries/Utilities/codegenNativeComponent.d.ts#L16 - the current version is correct according to the type definition in react-native source code

common/cpp/react/renderer/components/rnscreens/RNSTabsHostShadowNode.h
```diff
@@ -12,8 +12,8 @@ JSI_EXPORT extern const char RNSTabsHostComponentName[];
 
 class JSI_EXPORT RNSTabsHostShadowNode final : public ConcreteViewShadowNode<
                                                    RNSTabsHostComponentName,
-                                                   RNSTabsHostProps,
-                                                   RNSTabsHostEventEmitter,
+                                                   RNSTabsHostIOSProps,
+                                                   RNSTabsHostIOSEventEmitter,
```
kkafar, line 17: I think we need to split Android & iOS components here. Maybe It'd be beneficiary to split the definition into two files & include appropriate one depending on a platform? 

Only a suggestion, depending on how vast the differences will be.

ios/tabs/screen/RNSTabsScreenComponentView.mm
```diff
@@ -755,7 +755,7 @@ -(RCTDirectEventBlock)eventName
 #if RCT_NEW_ARCH_ENABLED
 #pragma mark - View class exposure
 
-Class<RCTComponentViewProtocol> RNSTabsScreen(void)
+Class<RCTComponentViewProtocol> RNSTabsScreenIOS(void)
```
kkafar:
```suggestion
Class<RCTComponentViewProtocol> RNSTabsScreenIOSCls(void)
```

It's not related to this PR, but I thought that `Cls` suffix is necessary. I'll investigate that.

kkafar: Seems that these methods are not called at all now. I don't know in what version of react-native things changed here. We need to investigate that. Let's create ticket for this.

src/fabric/tabs/TabsHostIOSNativeComponent.ts
```diff
@@ -0,0 +1,69 @@
...
+type NativeFocusChangeEvent = {
+  tabKey: string;
+  repeatedSelectionHandledBySpecialEffect: boolean;
+};
+
+// #endregion General helpers
+
+// #region iOS-specific helpers
+
+type TabBarMinimizeBehaviorIOS =
```
kkafar:
```suggestion
type TabBarMinimizeBehavior
```

These type definitions in native component spec file are not exported, therefore we do not need the `IOS` / `Android` suffix. This will reduce necessary amount of changes in native code & also simplify naming there.

src/fabric/tabs/TabsScreenAndroidNativeComponent.ts
```diff
@@ -0,0 +1,127 @@
...
+type LifecycleStateChangeEvent = Readonly<{
+  previousState: CT.Int32;
+  newState: CT.Int32;
+}>;
+
+// #endregion General helpers
+
+// #region Android-specific helpers
+
+type TabBarItemLabelVisibilityModeAndroid =
```
kkafar: Similarly in this file.

src/fabric/tabs/TabsScreenAndroidNativeComponent.ts
```diff
@@ -0,0 +1,127 @@
...
+  | 'auto'
+  | 'selected'
+  | 'labeled'
+  | 'unlabeled';
+
+export type AppearanceAndroid = {
```
kkafar: We should not export these types outside. I think that they should stay module-local & all public APIs should be explicitly defined on public types level.

kkafar: Maybe a little bit different: they can be exported & used for definition of the actual public files. But they should not be exported out directly from the module representing the whole component.

src/fabric/tabs/TabsScreenIOSNativeComponent.ts
```diff
@@ -126,7 +50,7 @@ type BlurEffect =
   | 'systemThickMaterialDark'
   | 'systemChromeMaterialDark';
 
-type Orientation =
+type OrientationIOS =
```
kkafar: Let's remove the iOS prefix.

src/fabric/tabs/TabsScreenIOSNativeComponent.ts
```diff
@@ -170,59 +129,56 @@ export interface NativeProps extends ViewProps {
...
 
   // Experimental
-  userInterfaceStyle?: CT.WithDefault<UserInterfaceStyle, 'unspecified'>;
+  userInterfaceStyle?: CT.WithDefault<UserInterfaceStyleIOS, 'unspecified'>;
+
+  // #endregion iOS-specific Props
 }
 
-export default codegenNativeComponent<NativeProps>('RNSTabsScreen', {});
+export default codegenNativeComponent<NativeProps>('RNSTabsScreenIOS', {
```
kkafar, line 182:
```suggestion
export default codegenNativeComponent<NativeProps>('RNSTabsScreen', {
```

can we consider dropping this suffix - it shouldn't be needed in native code.

src/fabric/tabs/TabsScreenIOSNativeComponent.ts
kkafar: From all the props

kkafar (commented):
Left some remarks - done with pairing.

src/fabric/tabs/TabsHostIOSNativeComponent.ts
t0maboro: Currently, I'm using these in conversions to native props. Temporarily, I'd prefer to mention that the type is platform-specific, but we may consider dropping this suffix.

src/fabric/tabs/TabsScreenAndroidNativeComponent.ts
t0maboro: it's needed for conversions to native props, we may consider refactoring this, but I'd prefer to perform that outside the scope of this PR

src/fabric/tabs/TabsScreenAndroidNativeComponent.ts
t0maboro: for the consistency (even if temporal, I prefer to keep that, as we need to export two Appearance objects for iOS and Android from native components to properly type native props conversions

src/fabric/tabs/TabsScreenIOSNativeComponent.ts
t0maboro: same as above, for consistency I'd consider to keep that (see the Appearance case)

src/fabric/tabs/TabsScreenIOSNativeComponent.ts
t0maboro: seems like codegen is parsing files one-by-one, and if I use a shared name, it will override files, using another spec. So, we should keep separate names. Is this okay for you?

common/cpp/react/renderer/components/rnscreens/RNSTabsHostShadowNode.h
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3684/changes/ee273c516bb3b3219fd2dc418f443993f9474416

ios/tabs/screen/RNSTabsScreenComponentView.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3684/changes/da1ed7152049c97b8ad3859885530c77233f2f23

src/fabric/tabs/TabsScreenAndroidNativeComponent.ts
```diff
@@ -0,0 +1,127 @@
...
+  typography?: TypographyAppearance;
+  badge?: BadgeAppearance;
+};
+
+type ItemStateColors = {
+  iconColor?: ProcessedColorValue | null;
+  titleColor?: ProcessedColorValue | null;
+};
+
+type BottomNavItemColors = {
```
kkafar, line 52:
```suggestion
type TabBarItemColors = {
```

Something like that? To align naming between paltforms

src/fabric/tabs/TabsScreenAndroidNativeComponent.ts
```diff
@@ -0,0 +1,127 @@
...
+  tabBarBackgroundColor?: ProcessedColorValue | null;
+  tabBarItemRippleColor?: ProcessedColorValue | null;
+  tabBarItemLabelVisibilityMode?: CT.WithDefault<
+    TabBarItemLabelVisibilityMode,
+    'auto'
+  >;
+
+  itemColors?: BottomNavItemColors;
+  activeIndicator?: ActiveIndicatorAppearance;
+  typography?: TypographyAppearance;
```
kkafar, line 43: Let's discuss this further down the line, but I see a problem here: just from the property hierarchy I have no idea what this prop applies to? Fonts on screen? Fonts in tab bar? tab bar items? Let's keep that in mind & discuss in the root PR.

src/fabric/tabs/TabsScreenAndroidNativeComponent.ts
t0maboro: will be fixed in #3683

src/fabric/tabs/TabsScreenAndroidNativeComponent.ts
t0maboro: will be fixed in #3683

PR #3698, opened 2026-02-26
feat (Android): setTransitionDuration on Android

### LINKED ISSUE #3694: Feature Request: Allow customizing Android transition duration/interpolator in Native Stack
Description:
The default Android transition animation in native-stack feels too slow and heavy (sluggish). In high-performance or utility apps, users expect a snappier, faster screen transition to feel the responsiveness of the UI. Currently, there is no simple way to increase the speed of these animations on Android.

The Problem:
On iOS, native-stack supports the animationDuration prop, allowing us to make transitions faster. However, on Android, this prop is ignored because the animations are hardcoded in native XML files within react-native-screens. This makes the App feel "laggy" or "unnecessarily heavy" compared to the iOS version or other high-performance native Android apps.

Proposed Solution:
Expose a property to control the transition duration and interpolator for Android, allowing developers to set a shorter duration (e.g., 200ms instead of the default) to achieve a "snappier" feel.

Environment :

React Native Screens version: 4.23.0

React Native version: 0.83.2

Platform: Android

Steps to reproduce:

Use @react-navigation/native-stack.

Navigate between screens on a high-refresh-rate Android device.

Notice that the transition takes too long to complete, and there is no JS API to make it faster.

FilPag:
## Description
Closes #3694

On iOS, native-stack supports the animationDuration prop, allowing us to make transitions faster. However, on Android, this prop is ignored because the animations are hardcoded in native XML files within react-native-screens. This makes the App feel "laggy" or "unnecessarily heavy" compared to the iOS version or other high-performance native Android apps.

This PR exposes a property to control the transition duration for Android, allowing developers to set a shorter duration (e.g., 200ms instead of the default) to achieve a "snappier" feel.

## Changes

- Implemented transitionDuration prop on Android by wiring ScreenViewManager to Screen.kt and overriding onCreateAnimation in ScreenStackFragment to apply the custom duration
- Changed sentinel default from 500 to -1 (unset) across Codegen specs and Paper delegate
- Added duration picker to the Animations example screen
- Updated public types to reflect Android support for transitionDuration

https://github.com/user-attachments/assets/bd1d8185-1d8e-4027-ab4c-c7375a0ecbec

## Test plan
Added duration picker to the Animations example screen. Navigate to Animations example screen in FabricExample. Test the duration selector

## Checklist

- [x] Ensured that CI passes


FilPag:
@kkafar Could you please request reviewers on my behalf. I'm unable to do so :)

kkafar:
Thanks for the tag here.

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -65,6 +65,7 @@ class Screen(
     var stackPresentation = StackPresentation.PUSH
     var replaceAnimation = ReplaceAnimation.POP
     var stackAnimation = StackAnimation.DEFAULT
+    var transitionDuration: Int = -1
```
kmichalikk: nit: Let's document here what does -1 mean or introduce some constant & use it later. If you go for a constant, you might want to make sure that -1 is the only valid negative integer (in the setter, for instance).

android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt
```diff
@@ -347,8 +349,31 @@ class ScreenStackFragment :
...
+    }
+
+    private fun scaleAnimationDuration(animation: Animation, targetDuration: Long) {
+        if (animation is AnimationSet) {
+            val naturalDuration = animation.computeDurationHint()
+            if (naturalDuration > 0) {
+                val scale = targetDuration.toDouble() / naturalDuration.toDouble()
+                for (child in animation.animations) {
+                    child.startOffset = (child.startOffset * scale).toLong()
+                    scaleAnimationDuration(child, (child.duration * scale).toLong())
```
kmichalikk: I think we only have one-level deep nesting in our animations, but still, `computeDurationHint` for child would be better

src/fabric/ModalScreenNativeComponent.ts
```diff
@@ -108,7 +108,7 @@ export interface NativeProps extends ViewProps {
   gestureResponseDistance?: GestureResponseDistanceType | undefined;
   stackPresentation?: CT.WithDefault<StackPresentation, 'push'>;
   stackAnimation?: CT.WithDefault<StackAnimation, 'default'>;
-  transitionDuration?: CT.WithDefault<CT.Int32, 500>;
+  transitionDuration?: CT.WithDefault<CT.Int32, -1>;
```
kmichalikk, line 111: You never updated iOS side after this change. It works only because there is a fallback to 500 for duration < 0, but this is not documented anywhere. 

There is outdated comment above `static constexpr NSTimeInterval RNSDefaultTransitionDuration = 0.5;` that you should change. 

Additionally, I'd set the above default as soon as possible to not store the negative value anywhere since it's not a valid value in the iOS flow.

src/fabric/ScreenNativeComponent.ts
```diff
@@ -108,7 +108,7 @@ export interface NativeProps extends ViewProps {
   gestureResponseDistance?: GestureResponseDistanceType | undefined;
   stackPresentation?: CT.WithDefault<StackPresentation, 'push'>;
   stackAnimation?: CT.WithDefault<StackAnimation, 'default'>;
-  transitionDuration?: CT.WithDefault<CT.Int32, 500>;
+  transitionDuration?: CT.WithDefault<CT.Int32, -1>;
```
kmichalikk, line 111: Same as above

kmichalikk (changes requested):
Hi @FilPag 
Thanks for the PR. I left some comments, please answer them.

FilPag:
@kmichalikk Thank you for the comments. did some updates. let me know what you think!

kmichalikk (commented):
Thanks, looks good to me now.
One more thing: please update the doc on the prop in GUIDE_FOR_LIBRARY_AUTHORS.md and let's wait for @kkafar to approve.

kmichalikk:
Oh, and please run the formatter
lint is failing

kmichalikk (approved):

FilPag:
@kkafar :)

PR #3680, opened 2026-02-23
release: 4.24.0

kkafar:
Minor release shipping fixes & new experimental features.

> [!caution]
> We indent for this version to be the last one, that supports
> legacy architecture.

PR #3699, opened 2026-02-26
chore: Update lockfiles after 4.24 release

t0maboro:
it hasn't been covered in: https://github.com/software-mansion/react-native-screens/pull/3680


t0maboro:
e2e was cancelled because of the timeout; this change only updates lockfiles, what should be done earlier, while releasing 4.24, therefore, I think that it can be merged even without that check completed

PR #3704, opened 2026-03-01
feat: Added last visited screen button to Example app

kapobajza:
## Description

When I was trying to work on an issue using the FabricExample app, by adding the reproduction to the "Issue Tests" section of the app, I've noticed a repetitive pattern:
1. Run the FabricExample app from Xcode/Android Studio
2. Go to "Issue Tests" screen
3. Filter the specific screen you want to open, using the search bar
4. Open it
5. Make changes to the native side
6. Repeat the cycle from steps 1 to 5

That's why I decided to make this process a bit easier

## Changes

Added a "Go to last visited screen" button to the example app. This will reopen the last visited screen. This is especially useful for the "Issue Tests" screen. Because when you need to re-run your app every time you make native changes, you don't have to repeat the process of searching for the exact issue tests screen and opening it.

Also: since I am using `@react-native-async-storage/async-storage` for storage, I've added a RN dev menu item to clear async storage.

| Before | After |
| --- | --- |
| image | image |
| image | image |

## Checklist

- [ ] Ensured that CI passes


satya164:
wouldn't it be nicer to persist the navigation state so that steps 1-4 aren't necessary at all?

kapobajza:
> Hey, thanks for the PR & initiative but this does not align with approach we wanna take here.
> 
> First - we don't want to have any additional dependency. FabricExample is an app where we not only showcase the exposed API but also reproduce issues & test the API. We want to have as limited dependencies as possible to avoid bug-noise and false alarms.

Thanks for the insight, this is understandable.

> We just uncomment & render appropriate test case

My personal preference is to avoid doing this, as it often happens that I forget commenting the changes again when pushing. I know it's easily revertible and detected in PR reviews, but it is at least a mild inconvenience.

However, I do respect your approach and, from the perspective of not wanting to introduce additional complexity to the FabricExample app, it's totally reasonable.


Anyways, I will close this PR. But what I could do instead is to add these guidelines about uncommenting & rendering the appropriate test case to the contributing docs and open a PR with that. Since I didn't know about it, and I couldn't find it in the docs. Does that sound good to you?

kkafar:
Yes, I think this is a good idea. I'll gladly accept PR with such update to [this document](https://github.com/software-mansion/react-native-screens/blob/main/guides/CONTRIBUTING.md).

PR #3708, opened 2026-03-02
feat(windows): support react-native-windows new architecture

### LINKED ISSUE #3589: Projects with this package fail to build for Windows
### Description

Projects with the `react-native-screens` package fail to build for windows:

>fatal error LNK1104: cannot open file 'C:\Users\<username>\.nuget\packages\microsoft.windowsappsdk\1.7.250401001\buildTransitive\native\..\..\lib\win10-\Microsoft.WindowsAppRuntime.Bootstrap.lib' [RNWinRepro\node_modules\react-native-windows\Microsoft.ReactNative\Microsoft.ReactNative.vcxproj]. Check your build configuration.

### Steps to reproduce

Run the following scripts to create a new project with this package:
```
npx --yes @react-native-community/cli@latest init RNWinRepro --version 0.81.2
cd RNWinRepro 
yarn add react-native-windows@0.81.2
yarn add @react-native-async-storage/async-storage
npx react-native init-windows --overwrite
npx react-native run-windows
```

See minimal repro

### Screens version

4.20.0

### React Native version

0.81.2

### Platforms

Windows

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

PHILLIPS71:
### Description
The Windows build for `react-native-screens` has been broken for consumers using autolinking, as reported in #3589.

The root cause is that the library's Windows native module was built against `react-native-windows@0.64`, which is long out of date and incompatible with current New Architecture tooling. This manifested as a linker error (`LNK1104 win10-\`) caused by autolinking discovering the legacy `.sln` files that include `Microsoft.ReactNative.vcxproj`, leaving `$(Platform)` unresolved.

This PR fixes the autolinking issue and fully migrates the Windows native layer to the Fabric (New Architecture) component model required by `react-native-windows@0.81`.

Closes #3589.

### Changes

**Autolinking fix**
- Added explicit `windows` platform config in `react-native.config.js` pointing autolinking only at `RNScreens\RNScreens.vcxproj` (no solution file), preventing the legacy `.sln` files from being auto-discovered and causing the `$(Platform)` linker error

**react-native-windows upgrade**
- Upgraded `react-native-windows` from `^0.64.8` to `^0.81.4`
- Regenerated native module scaffold against `react-native-windows@0.81.4`; switches from the legacy UWP property sheets (`Microsoft.ReactNative.Uwp.CppLib.*`) to the new arch ones (`Microsoft.ReactNative.CppLib.*`)

**Fabric migration**
`react-native-windows` 0.81 removes the Paper renderer entirely. The `IViewManager*` interface family no longer exists. All Paper ViewManager classes and their XAML base classes have been replaced with the Fabric `AddViewComponent` / builder / UserData pattern.

*Removed Paper*
All `*ViewManager` classes (`ScreenViewManager`, `ScreenStackViewManager`, `ScreenContainerViewManager`, `ScreenStackHeaderConfigViewManager`, `ScreenStackHeaderSubviewViewManager`, `ModalScreenViewManager`, `SearchBarViewManager`) and their associated XAML base classes.

*Added Fabric*
| File | Role |
|---|---|
| `Screen.h/.cpp` | `ScreenProps`, `ScreenEventEmitter`, `ScreenUserData`, and `RegisterScreenLike` helper shared by Screen and ModalScreen |
| `ModalScreen.h/.cpp` | Thin wrapper, delegates to `RegisterScreenLike` with `"RNSModalScreen"` |
| `ScreenStack.h/.cpp` | Registers `RNSScreenStack` with `onFinishTransitioning` emitter infrastructure |
| `ScreenContainer.h/.cpp` | Registers `RNSScreenContainer`; no custom props per spec |
| `ScreenStackHeaderConfig.h/.cpp` | Stub registration, see Notes |
| `ScreenStackHeaderSubview.h/.cpp` | Stub registration, see Notes |
| `SearchBar.h/.cpp` | Stub registration, see Notes |
| `BaseProps.h` | Shared minimal `IComponentProps` for stub components; required by the Fabric framework to safely deliver base `ViewProps` |

*Updated:*
- `ReactPackageProvider.cpp` replaces `AddViewManager` calls with `IReactPackageBuilderFabric`; falls back gracefully on non-Fabric hosts

**Lifecycle events**
The old implementation dispatched `onWillAppear`, `onAppear`, `onWillDisappear`, and `onDisappear` from XAML `Loading`, `Loaded`, and `Unloaded` events. Fabric has no pre-mount hook, so `onWillAppear` + `onAppear` are both dispatched from `ComponentView::Mounted`, and `onWillDisappear` + `onDisappear` from `ComponentView::Unmounted`.

The `Will*` events are co-incident with their counterparts on Windows. This matches the old behaviour for `Disappear` (the old `Unloaded` handler also fired both synchronously) and is an acknowledged approximation for `WillAppear`.

### Notes
- **Native version downgrade**: `react-native` devDependency is downgraded from `0.84.0` to `0.81.5` to match `react-native-windows@0.81.4`. The peer dependency floor is also lowered from `>=0.82.0` to `>=0.81.0`

- **Stub components**: `ScreenStackHeaderConfig`, `ScreenStackHeaderSubview`, and `SearchBar` are registered as stubs. The previous implementation was also non-functional for these components (empty XAML `StackPanel` with no props applied or events fired). A non-empty stub registration is required, an empty builder lambda causes the Fabric framework to crash when delivering props. Native implementations are tracked as follow-up work.

- **react-native-restart**: was removed as it doesn't support the new architecture

**Known gaps (follow-up work)**:
- `stackAnimation` and `replaceAnimation` are accepted and stored but not applied; no native transition animations exist on Windows yet
- `ScreenStackHeaderConfig` / `ScreenStackHeaderSubview` require a Windows title bar strategy before they can be implemented; `SearchBar` maps naturally to WinUI `AutoSuggestBox` and is a self-contained follow-up


kkafar:
Thanks for the PR, that's awesome. Tag me once it is ready for review.

PHILLIPS71:
@kkafar ready for a first pass whenever you get a chance!

One thing worth discussing: the existing `windows-build-test.yml` dynamically creates a fresh example app on every CI run, it queries npm for the latest `react-native-windows` version, spins up a matching RN project, builds it, then throws it away. It works, but there's no committed `WindowsExample/` in the repo the way `FabricExample/` exists for iOS and Android.

Should we instead commit a concrete `WindowsExample/` and simplify the pipeline to just build it directly? It would be consistent with how the other platforms work and give contributors something to actually run. The trade-off is that `react-native-windows` tends to trail `react-native` releases by a minor version, so it would need to pin a different RN version than `FabricExample/` and require a periodic update when RNW catches up.

src/core.ts
```diff
@@ -16,7 +16,7 @@ export function enableScreens(shouldEnableScreens = true) {
     return;
   }
 
-  if (ENABLE_SCREENS && !UIManager.getViewManagerConfig('RNSScreen')) {
+  if (ENABLE_SCREENS && !UIManager.hasViewManagerConfig('RNSScreen')) {
```
kkafar: Doesn't `getViewManagerConfig` exist in windows runtime? If it does, please revert this change. I'll accept a separate PR with such change, but not in this PR. It's not related.

apps/src/screens/Gestures.tsx
```diff
@@ -13,7 +13,6 @@ import {
   NativeStackNavigationProp,
 } from '@react-navigation/native-stack';
 import { useHeaderHeight } from '@react-navigation/elements';
-import RNRestart from 'react-native-restart';
```
kkafar, line 16: I guess you remove it, because it is not supported on Windows. This is pretty outdated package. 

I see that it's being used together with `I18nManger` - please convince me that it can be safely removed w/o hurting existing examples. Otherwise we need to find a way to replace it.

apps/Example.tsx
```diff
@@ -40,6 +37,23 @@ import SingleFeatureTests from './src/tests/single-feature-tests';
...
+  Platform.OS === 'windows'
+    ? ({ children }) => <React.Fragment>{children}</React.Fragment>
+    : require('react-native-gesture-handler').GestureHandlerRootView;
+
+const GestureDetectorProvider: React.ComponentType<{
+  children: React.ReactNode;
+}> =
+  Platform.OS === 'windows'
+    ? ({ children }) => <React.Fragment>{children}</React.Fragment>
+    : require('react-native-screens/gesture-handler').GestureDetectorProvider;
```
kkafar: This is fine, but let's try to avoid using `require` for conditional module import here - AFAIK this might break ESM support in the application (could be harder to test ESM compat for us later).

Likely we can work around that by leveraging Metro capabilities here.
If we create `gesture-handler-interop.ts` & `gesture-handler-interop.windows.ts` & export these views from there (proper one & mocks respectively) then this might work & be ESM compliant.

TVOSExample/App.tsx
kkafar, line 1: We likely don't want to touch this file.

TVOSExample/package.json
```diff
@@ -25,7 +25,6 @@
     "react-native": "npm:react-native-tvos@0.84.0-0",
     "react-native-gesture-handler": "^2.29.0",
     "react-native-reanimated": "~4.2.2",
-    "react-native-restart": "^0.0.27",
```
kkafar, line 29: This also seems unnecessary. TVOSExample seems unrelated to windows.

windows/RNScreens/codegen/.clang-format
```diff
@@ -0,0 +1,2 @@
+DisableFormat: true
+SortIncludes: false
```
kkafar: Could you provide some rationale here?

kkafar:
```suggestion
DisableFormat: true
SortIncludes: false

```

windows/RNScreens/ModalScreen.cpp
```diff
@@ -0,0 +1,16 @@
...
+
+// RNSModalScreen shares identical native behavior with RNSScreen on Windows.
+// Both components are containers whose screen-level semantics (presentation
+// style, animation) are managed by JS; the native side only needs to forward
+// lifecycle events.
+void RegisterModalScreen(
+    const IReactPackageBuilderFabric &fabricBuilder) noexcept {
+  RegisterScreenLike(fabricBuilder, L"RNSModalScreen");
+}
+} // namespace winrt::RNScreens::implementation
```
kkafar:
```suggestion
} // namespace winrt::RNScreens::implementation

```

windows/RNScreens/ModalScreen.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+#include "pch.h"
+
+namespace winrt::RNScreens::implementation {
+// Delegates to RegisterScreenLike; see Screen.h for the shared behavior.
+void RegisterModalScreen(
+    const Microsoft::ReactNative::IReactPackageBuilderFabric &fabricBuilder)
+  noexcept;
+} // namespace winrt::RNScreens::implementation
```
kkafar:
```suggestion
} // namespace winrt::RNScreens::implementation

```

windows/RNScreens/pch.cpp
```diff
@@ -1 +1 @@
-﻿#include "pch.h"
+#include "pch.h"
```
kkafar:
```suggestion
#include "pch.h"

```

package.json
```diff
@@ -56,7 +56,8 @@
     "react-native.config.js",
     "README.md",
     "!**/__tests__",
-    "!android/src/main/java/com/swmansion/rnscreens/LifecycleHelper.kt"
+    "!android/src/main/java/com/swmansion/rnscreens/LifecycleHelper.kt",
+    "windows"
```
kkafar: `windows/` directory is already included few lines above

package.json
```diff
@@ -78,18 +79,18 @@
   },
   "peerDependencies": {
     "react": "*",
-    "react-native": ">=0.82.0"
+    "react-native": ">=0.81.0"
```
kkafar: This can not be lowered. We've just recently started using `peerDependencies` explicitly (see [here](https://github.com/software-mansion/react-native-screens/pull/3691)) and it's meant to denote lowest supported version according to our policy. It's 0.82 since not-yet-released 4.25.0.

package.json
```diff
@@ -112,12 +113,12 @@
     "prettier": "^2.8.8",
     "react": "19.2.3",
     "react-dom": "^19.1.0",
-    "react-native": "0.84.0",
+    "react-native": "0.81.5",
```
kkafar: This also should not be changed. We keep the dev library dependencies up-to-date so that we can react to changes in the react-native core & report issues there when necessary.

kkafar (commented):
Thanks for the PR! 

I've left initial notes. I haven't gone through the component code yet, but I've reviewed surrounding changes.

package.json
kkafar: I see it might be problematic :/ I'll ask some people internally whether it would be possible for us to go back to `react-native: "*"`.

kkafar:
Regarding the question about `WindowsExample` you asked - I consider it. I do not remember exact rationale, but IIRC we had an windows example at some point (?), but no one to really maintain it. So my answer here depends on whether we can find a maintainer for that part of the codebase. The "core team at SWM" is now heavily invested in the "in-tree platforms" and this won't change for many more months to come.

src/core.ts
PHILLIPS71: It does, I kept running into console errors outlining that `getViewManagerConfig` was no longer supported and to use `hasViewManagerConfig`, so to silence it i swapped it. I've reverted it as it can be addressed separately.

apps/Example.tsx
PHILLIPS71: Good idea! I've gone ahead and done that.

windows/RNScreens/codegen/.clang-format
PHILLIPS71: The `.clang-format` is generated by the RNW toolchain and is there as without this override, any editor or CI clang-format run would reformat `NativeScreensModuleSpec.g.h`.

package.json
PHILLIPS71: The downgrade was necessary because `react-native-windows` trails behind react-native's versioning. `react-native-windows@0.81.4` only supports `react-native@0.81.x`, and `0.84` is not yet supported upstream. This also caused the peer dependency range to drop from `>=0.82.0` to `>=0.81.0`, which I agree is not ideal.

PHILLIPS71:
Thanks @kkafar, I've reverted the removal of `react-native-restart` and used an interop file that will become a non-op in Windows, there is https://github.com/ahmedawaad1804/react-native-restart-newarch that supports the new architecture, though it can probably be handled in a separate pr.

tux2nicolae:
@PHILLIPS71 thank you for your work, and thanks **kkafar** for review-ing it.

With the release of react-native-windows 0.82.0 which targets the 0.82.0 i think the peer dependency min version problem is solved by itself ? 

We're waiting for windows support as well for a while and this PR gives us some hope

PHILLIPS71:
@tux2nicolae I believe so, I've update the branch against `main` and updates to `react-native-windows@0.82.3` so this PR doesn't become stale.

tux2nicolae:
@PHILLIPS71 that's awesome, thx, @kkafar when you find some time can you review this one more time please?

ClaudiuHBann:
@kkafar, could you please take another look at this PR?

marcor-a:
Any updates here?

tux2nicolae:
The 0.84.0 windows version just got released, @PHILLIPS71 can you rebase this branch please?

tux2nicolae:
~@PHILLIPS71 Please, can you help us with updating this PR to latest version and rebase it on the `software-mansion:main`, we are in process of almost releasing the windows build but this one remains a blocker for us.~

~I know you may be busy with something else, this being an opensource PR, but in case you have some spare time, we will really appreciate it.~

~We managed to rebase it, and opened a new PR here~
~https://github.com/software-mansion/react-native-screens/pull/4281~

This PR was updated to react-native-windows 0.84.0

PHILLIPS71:
@tux2nicolae I've updated it, though unsure if it's going to land in `main`

tux2nicolae:
@PHILLIPS71 

Awesome! Thank you! 

I will close mine, it's not relevant anymore

tux2nicolae:
@PHILLIPS71 We managed to run our app on windows because of your PR, thank you, really appreciate it.

But in the process, we also needed to add some stub files for the bundler to be able to load some files correctly, we also created a PR to merge them into your branch.

https://github.com/PHILLIPS71/react-native-screens/pull/1

PHILLIPS71:
> @PHILLIPS71 We managed to run our app on windows because of your PR, thank you, really appreciate it.
> 
> 
> 
> But in the process, we also needed to add some stub files for the bundler to be able to load some files correctly, we also created a PR to merge them into your branch.
> 
> 
> 
> https://github.com/PHILLIPS71/react-native-screens/pull/1

I'm glad it was of use! I've merged in the fixes you've applied, hopefully it get some traction and it's merged upstream 🤞

tux2nicolae:
@kkafar when you have some spare time, can you review this again please?

tux2nicolae:
The `react-native-restart` just released a new version with new architecture support including windows.

https://github.com/avishayil/react-native-restart/releases/tag/v0.0.29

PR #3713, opened 2026-03-02
chore: Update TVOSExample to RN 0.84

kmichalikk:
Closes https://github.com/software-mansion/react-native-screens-labs/issues/1008

## Description

This PR updates TVOSExample to use react-native-tvos@0.84 and react-native@0.84.

## Changes

Updated according to [upgrade helper](https://react-native-community.github.io/upgrade-helper/?from=0.81.0&to=0.84.0&name=TVOSExample) with some differences:
- omitted SUPPORTED_PLATFORMS in .pbxproj: the project won't compile when iphoneos is set as supported (storyboard incompatibility)
- included the fix for orientation (though, it probably won't have a difference here, let me know if we even want to be consistent here, or just remove the keys from Info.plist)

## Test plan

Build the app & see if basic navigation works

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/.github/workflows/tv-os-build-test.yml
+++ b/.github/workflows/tv-os-build-test.yml
@@ -16,7 +16,7 @@ on:
   workflow_dispatch:
 jobs:
   build:
-    runs-on: macos-14
+    runs-on: macos-26
     timeout-minutes: 60
     env:
       WORKING_DIRECTORY: TVOSExample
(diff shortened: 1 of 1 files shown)


kkafar (approved):
Looks good. I've header that there is some work still needed in CI workflows. Let's push them & proceed. Thanks!

PR #3714, opened 2026-03-02
feat(Android, Tabs): State-driven appearance on Android

t0maboro:
## Description

This PR refines the Android tabs' appearance configuration into a more structured, explicit form.

The goal of this change is to:

- Align the Android API more closely with the iOS implementation
- Make appearance configuration more predictable

The appearance configuration for Android tabs has been reorganized into clearly separated sections, describing different aspects of TabBar and TabBarItems:

- tabBarBackgroundColor - prop is separated, because TabBar is an owner; independent of given TabBarItem state
- tabBarItemRippleColor - possibly might have ColorStateList, not sure whether we have this configurable by state or not, I don't have a good use-case to have it configurable by state
- tabBarItemLabelVisibilityMode - related to all TabBarItems labels, so the owner is a TabBar, even if the name doesn't indicate that
- (normal / selected / focused / disabled) - here all state-driven props are defined
- tabBarItemActiveIndicatorColor, tabBarItemActiveIndicatorEnabled - single activeIndicator, no need to nest these props
- tabBarItemTitleFontFamily, tabBarItemTitleSmallLabelFontSize, tabBarItemTitleLargeLabelFontSize, tabBarItemTitleFontWeight, tabBarItemTitleFontStyle - unrelated to the state, as the Material lib doesn't support that explicitly
- tabBarItemBadgeBackgroundColor, tabBarItemBadgeTextColor - unrelated to state, because in Material, these props are taking Color instead of ColorStateList

---

From now on, the appearance of the entire Android tab bar is determined by the currently active Screen, the same way it is handled on iOS. When the active Screen changes, the tab bar appearance updates accordingly.

---

While the API allows updating appearance dynamically, some properties can negatively affect layout when changed at runtime.

Changing these props dynamically may cause visual glitches:
- tabBarItemLabelVisibilityMode
- tabBarItemActiveIndicatorEnabled
- title font-related props

For this reason, it is not recommended to change these properties at runtime. They should be treated as relatively stable configuration options.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/955

## Changes

- Renamed classes: AppearanceApplicator & AppearanceCoordinator are now less coupled with the Host component, as they depend on the selected Screen. Therefore, I decided to remove "Host" from the names of these classes.
- ColorsStateLists were extended to support focused and disabled states - parity with UIAppearance states from UIKit.
- New model for TabsAppearance was defined for android.
- Old way of passing all props as top-level props was replaced with passing an appearance config object from JS. This is a breaking change. We'll add the compatibility flag after the full refactor.
- All native prop "legacy" setters were replaced with the specific parser for the Appearance object.
- Appearance configuration was moved from Host to Screen.

This PR unlocks some new behaviors, not sure if Before section is needed, as for Tab 1 & Tab 4 - the flow remains unchanged. For Tab2 & Tab3 - I'm adding new features for configuring TabBar appearance, which are covered in `After` video

### After

https://github.com/user-attachments/assets/7831a30a-f572-489f-b873-98f6c89da589

## Test plan

Updated `TestBottomTabs` example to match the new API structure.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabsScreenViewManager.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabsScreenViewManager.kt
@@ -253,20 +253,14 @@ class TabsScreenViewManager :
     @ReactProp(name = "standardAppearanceAndroid")
     override fun setStandardAppearanceAndroid(
         view: TabsScreen,
-        value: Dynamic?,
+        value: ReadableMap?,
     ) {
-        if (value == null || value.isNull) {
-            view.appearance = null
-            return
-        }
-        val appearanceMap = if (value.type == ReadableType.Map) value.asMap() else null
-
-        if (appearanceMap == null) {
+        if (value == null) {
             view.appearance = null
             return
         }
 
-        view.appearance = parseAndroidTabsAppearance(appearanceMap)
+        view.appearance = parseAndroidTabsAppearance(value)
     }
 
     private fun parseAndroidTabsAppearance(appearance: ReadableMap): TabsAppearance =
--- a/src/fabric/tabs/TabsScreenNativeComponent.ts
+++ b/src/fabric/tabs/TabsScreenNativeComponent.ts
@@ -187,7 +187,7 @@ export interface NativeProps extends ViewProps {
 
   // Android-specific appearance
   // Note: temporary standardAppearanceAndroid, suffix will be dropped after splitting components
-  standardAppearanceAndroid?: UnsafeMixed<AppearanceAndroid>;
+  standardAppearanceAndroid?: AppearanceAndroid;
 
   // iOS-specific appearance
   standardAppearance?: UnsafeMixed<Appearance>;


android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabsScreenViewManager.kt
```diff
@@ -259,6 +249,82 @@ class TabsScreenViewManager :
...
+                else -> null
+            }
+        } catch (e: Exception) {
+            Log.w("RNScreens", "[RNScreens] Could not parse color for key '$key': ${e.message}")
+            null
+        }
+    }
+
     companion object {
         const val REACT_CLASS = "RNSTabsScreen"
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3714/changes/5061fe7d4b30c62442fc915a6566a64e8d80f113

src/components/tabs/TabsScreen.types.ts
```diff
@@ -410,17 +565,15 @@ export interface TabsScreenProps {
...
-  /**
-   * @summary Specifies the background color of the badge.
+   * Allows to customize the appearance depending on the tab bar item state
+   * (normal, selected, focused, disabled). Configuration for the Bottom Navigation View
+   * is determined by the currently active tab screen.
    *
    * @platform android
    */
-  tabBarItemBadgeBackgroundColor?: ColorValue;
+  standardAppearanceAndroid?: TabsScreenAppearanceAndroid;
```
kligarski, line 576: Are we going to have appearance for Android other than `standard`? I can see this in the future if we decide to e.g. implement some kind of sidebar for tablets to match iOS but I guess that's only an idea for now.

android/src/main/java/com/swmansion/rnscreens/utils/ColorUtils.kt
```diff
@@ -0,0 +1,24 @@
+package com.swmansion.rnscreens.utils
+
+import android.content.Context
+import android.graphics.Color
+import android.util.Log
+import android.util.TypedValue
+import androidx.annotation.ColorInt
+import com.swmansion.rnscreens.gamma.tabs.appearance.TabsAppearanceApplicator.Companion.TAG
```
kligarski: Should we use a tag from `TabsAppearanceApplicator` in `ColorUtils.kt`?

android/src/main/java/com/swmansion/rnscreens/utils/ColorUtils.kt
```diff
@@ -0,0 +1,24 @@
...
+    attr: Int,
+    @ColorInt defaultColor: Int = Color.TRANSPARENT,
+): Int {
+    val typedValue = TypedValue()
+    val resolved = context.theme.resolveAttribute(attr, typedValue, true)
+
+    if (!resolved) {
+        Log.w(TAG, "[RNScreens] Failed to resolve color attribute. Falling back to Color.TRANSPARENT.")
+        return defaultColor
+    }
```
kligarski, line 22: nit: if you pass to `defaultColor` something other than `Color.TRANSPARENT`, this warning message is incorrect.

src/components/tabs/TabsScreen.types.ts
t0maboro: done rather for aligning with iOS nomenclature

android/src/main/java/com/swmansion/rnscreens/utils/ColorUtils.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3714/commits/aa050c1387496196b346bbefc7b700794bcec558

android/src/main/java/com/swmansion/rnscreens/utils/ColorUtils.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3714/changes/f03e34b8840210be1abc821fffd8ecafbe47b8ad

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
```diff
@@ -15,63 +15,80 @@ import com.facebook.react.uimanager.PixelUtil
...
+    private val states =
+        arrayOf(
+            intArrayOf(-android.R.attr.state_enabled), // disabled
+            intArrayOf(android.R.attr.state_selected), // selected
+            intArrayOf(android.R.attr.state_focused), // focused
+            intArrayOf(), // normal
+        )
+
+    fun updateSharedAppearance(
+        context: ContextThemeWrapper,
```
kkafar:
```suggestion
        context: Context,
```

should be enough, I doubt you need the `ContextThemeWrapper` explicitly

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
```diff
@@ -82,21 +99,25 @@ class TabsHostAppearanceApplicator(
...
 
         bottomNavigationView.isItemActiveIndicatorEnabled =
-            tabsHost.isTabBarItemActiveIndicatorEnabled
+            tabBarAppearance?.tabBarItemActiveIndicatorEnabled ?: true
         bottomNavigationView.itemActiveIndicatorColor = ColorStateList.valueOf(activeIndicatorColor)
     }
 
-    fun updateFontStyles(tabsHost: TabsHost) {
+    fun updateFontStyles(
+        context: ContextThemeWrapper,
```
kkafar: similarly here

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceModel.kt
```diff
@@ -0,0 +1,25 @@
...
+    val tabBarItemTitleFontWeight: String? = null,
+    val tabBarItemTitleFontStyle: String? = null,
+    val tabBarItemBadgeBackgroundColor: Int? = null,
+    val tabBarItemBadgeTextColor: Int? = null,
+)
+
+internal data class ItemStateAppearance(
+    val tabBarItemIconColor: Int? = null,
+    val tabBarItemTitleFontColor: Int? = null,
+)
```
kkafar, line 25: Jeeez, so only these two are state dependent? 🤦🏻‍♂️

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabsScreenViewManager.kt
```diff
@@ -259,7 +249,84 @@ class TabsScreenViewManager :
...
+    private fun ReadableMap.getOptionalColor(key: String): Int? {
+        if (!hasKey(key) || isNull(key)) return null
+        return try {
+            when (getType(key)) {
+                ReadableType.Number -> getInt(key)
+                ReadableType.String -> getString(key)?.toColorInt()
+                else -> null
+            }
+        } catch (e: Exception) {
+            Log.w(TAG, "[RNScreens] Could not parse color for key '$key': ${e.message}")
```
kkafar: In Kotlin you can define such helper functions outside of the class, just in the top level of the file. Since they are not strictly related to the view manager / component props I'd recommend moving them to the bottom of this file (still marking them as private).

apps/src/tests/issue-tests/TestBottomTabs/index.tsx
kkafar, line 1: I'd really want separate feature test for it, e.g. `test-tab-bar-appearance-defined-by-selected-tab` or something. 

This test alongside any other using the Tabs API must also be obviously updated in this PR.

src/components/tabs/TabsScreen.types.ts
```diff
@@ -155,25 +306,29 @@ export interface TabsScreenAppearance {
 // iOS-specific
 export interface TabsScreenItemAppearance {
   /**
-   * Specifies the tab bar item appearance when it's enabled, unselected, and not the focused item.
+   * Defines the colors for all tab bar items which are in their enabled, unselected and unfocused state.
+   * The color scheme is determined by the configuration of the currently selected tab.
```
kkafar: I'd refer to these as `appearance` not only colors. On iOS I believe e.g. font size is also dynamic.

src/components/tabs/TabsScreen.types.ts
kkafar: I think the name `standardAppearance` is fine. Just for the sake of alignment where possible.

src/fabric/tabs/TabsHostNativeComponent.ts
```diff
@@ -3,24 +3,11 @@
 import { codegenNativeComponent } from 'react-native';
 import type { CodegenTypes as CT, ColorValue, ViewProps } from 'react-native';
 
-// TODO: Report issue on RN repo, that nesting color value inside a struct does not work.
-// Generated code is ok, but the value is not passed down correctly - whatever color is set
-// host component receives RGBA(0, 0, 0, 0) anyway.
-// type TabBarAppearance = {
-//   backgroundColor?: ColorValue;
-// };
```
kkafar, line 11: I think we have some very very old ticket for this somewhere. Yeah: https://github.com/software-mansion/react-native-screens-labs/issues/12

So it's ok to remove this.

src/fabric/tabs/TabsScreenNativeComponent.ts
```diff
@@ -138,10 +185,11 @@ export interface NativeProps extends ViewProps {
   selectedDrawableIconResourceName?: string;
   selectedImageIconResource?: ImageSource;
 
-  tabBarItemBadgeTextColor?: ColorValue;
-  tabBarItemBadgeBackgroundColor?: ColorValue;
+  // Android-specific appearance
+  // Note: temporary standardAppearanceAndroid, suffix will be dropped after splitting components
+  standardAppearanceAndroid?: UnsafeMixed<AppearanceAndroid>;
```
kkafar: I'd like us to research usage of more stricter codegen types here.

There is this issue: https://github.com/software-mansion/react-native-screens-labs/issues/12

I'd like us to verify whether it is still valid. If things happen to work it would be great to migrate both Android & iOS code for that or at least start discussion, because there are some tradeoffs with each approach (type safety, memory usage (dynamics are heavy)).

apps/src/tests/issue-tests/TestBottomTabs/index.tsx
kligarski: `test-tabs-<...>.tsx` according to our new test naming convention: https://github.com/software-mansion/react-native-screens-labs/discussions/934#discussioncomment-15982360

t0maboro:
> I haven't found discrepancies but want to ask you to make sure that if we have two props (Android & iOS) that refer to & configure the same thing - they are named the same.

already left some notes in the code that's temporary `standardAppearanceAndroid` and it will be changed on components split, so we're in the intermediate state for a while

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceModel.kt
t0maboro: we may consider more, but they don't take ColorStateList as an argument, so it will require taking ownership for TabBarItem state updates

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3714/changes/31300a06cb5ebf24d65cd4baffed8b50b91eb690

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3714/changes/31300a06cb5ebf24d65cd4baffed8b50b91eb690

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabsScreenViewManager.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3714/changes/09443777e90fae4fb4335c9cd1f199e3128b4904

apps/src/tests/issue-tests/TestBottomTabs/index.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3714/changes/256337dec267ca5846699e38bcdfe89904eb2185

src/components/tabs/TabsScreen.types.ts
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3714/changes/a52fc600d93d075262292054cc00873811214bac

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceModel.kt
kkafar: Okay. Let's proceed for now. We might consider adding support for more in the future, but right now let's proceed.

src/fabric/tabs/TabsScreenNativeComponent.ts
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3714/commits/8ff335067cebe9680d121af11207b1ba23dc6a25

t0maboro: note: works fine on Android; haven't checked iOS yet, but even if iOS works either - that's not the PR in which we should cover it
action item on me: I'm reassigning https://github.com/software-mansion/react-native-screens-labs/issues/12 to me and I'll cover that after splitting components

PR #3718, opened 2026-03-03
feat(Android): configure `RNSLog` utility

sgaczol:
## Description

Introduce automatic `RNSLog` activation in debug mode and deactivation in release mode. This behavior can be manually overridden using the `rnsDebugLogsEnabled` flag.

Example usage:
`yarn run android --mode release --extra-params "-PrnsDebugLogsEnabled=true"` activates `RNSLog` in release mode.
`yarn run android --mode debug --extra-params "-PrnsDebugLogsEnabled=false"` deactivates `RNSLog` in debug mode.

Extend `RNSLog` functionality.

## Changes

- update `gradle.properties` and `build.gradle` 
- add more functionality to `RNSLog` analogous to native Android `Log`

Not applicable.

## Test plan

Run app in both modes and check if it works properly.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/FabricExample/android/gradle.properties
+++ b/FabricExample/android/gradle.properties
@@ -44,7 +44,7 @@ hermesEnabled=true
 edgeToEdgeEnabled=true
 
 # Use this property to enable or disable debug logging
-# rnsDebugLogsEnabled=true
+rnsDebugLogsEnabled=true
 
 # Use this property to enable V5-specific code
 rnsGammaEnabled=true


FabricExample/android/gradle.properties
```diff
@@ -43,8 +43,5 @@ hermesEnabled=true
 # Note: Only works with ReactActivity and should not be used with custom Activity.
 edgeToEdgeEnabled=true
 
-# Use this property to enable or disable debug logging
-rnsDebugLogsEnabled=true
-
```
kkafar, line 48: I'd leave this in place, just because it'll be easier to find & modify.

android/src/main/java/com/swmansion/rnscreens/utils/RNSLog.kt
```diff
@@ -22,4 +22,99 @@ object RNSLog {
             Log.d(tag, message.format(*args))
         }
     }
+
+    fun e(
+        tag: String,
+        message: String,
+    ) {
+        if (BuildConfig.RNS_DEBUG_LOGGING) {
```
t0maboro: very nitpicky: we might have a wrapper function that checks that flag and takes a specific logger as an argument

FabricExample/android/gradle.properties
sgaczol: Can I leave it commented out though so setting it to `true/false` would be optional then? Leaving it uncommented permanently will disturb default release mode state.

android/src/main/java/com/swmansion/rnscreens/utils/RNSLog.kt
sgaczol: Done: https://github.com/software-mansion/react-native-screens/pull/3718/commits/8478ef01176b0e070d7a222dee156d8f1890c9af

FabricExample/android/gradle.properties
sgaczol: Nevermind, you're right, it won't interfere there: https://github.com/software-mansion/react-native-screens/pull/3718/commits/1e28b556cfe7121181e55f71e090a92b47a54a88

android/src/main/java/com/swmansion/rnscreens/utils/RNSLog.kt
t0maboro: Q: is change from `vararg` to `Throwable` necessary?

sgaczol: I wanted it to be more consistent with native Android Log.

t0maboro: makes sense, thanks!

t0maboro (approved):

kligarski (approved):
Looks sensible.

PR #3719, opened 2026-03-03
chore(types): enable `exactOptionalPropertyTypes` support

YevheniiKotyrlo:
## Description

This PR enables TypeScript's [`exactOptionalPropertyTypes`](https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes) flag in the repo and adds `| undefined` to every non-required prop so the flag stays on cleanly.

`exactOptionalPropertyTypes` is a strict TypeScript flag (not included in `strict: true`) that distinguishes between "property is missing" and "property is explicitly `undefined`". When enabled, `prop?: T` means the property can be omitted but cannot be set to `undefined` — the fix is `prop?: T | undefined`.

This matters for downstream projects that enable this flag — without this change, they get type errors when passing destructured props from `react-native-screens` components. This is currently blocking [react-navigation from enabling the flag](https://github.com/react-navigation/react-navigation/pull/12995) — 17 of 22 `@ts-expect-error` directives in that PR come from `react-native-screens` types.

## Changes

The initial commit fixed only the 13 concrete eopt errors (14 after the later Tabs refactor). After internal discussion, @kligarski broadened the scope to every non-required prop across the codebase for a consistent public API — commits [`14c3927d8`](https://github.com/software-mansion/react-native-screens/pull/3719/commits/14c3927d8c310968d0a9c67e268f09920c71b991) ("add undefined to all non-required props") and [`bebd04cba`](https://github.com/software-mansion/react-native-screens/pull/3719/commits/bebd04cbadcbd808379edf67d8c74584e0c5b850) ("remove undefined after CT.WithDefault to satisfy codegen"). The merge commit [`bf4c49ab3`](https://github.com/software-mansion/react-native-screens/pull/3719/commits/bf4c49ab3e9e645d56ffcaff3d488e8ad835676b) integrates latest `main` and resolves the 4 conflicts from the `ScrollViewMarker` migration (#3895) and `controlNavigationStateInJS` removal (#3888).

**Current diff: 41 files, +722/-605**

| Category | Files | Note |
|---|---|---|
| Root | 2 | `tsconfig.json` (enables flag), `src/types.tsx` (all non-required props in `ScreenProps`, `GestureProps`, etc.) |
| Public component types | 14 `.types.ts` | SafeAreaView, ScrollViewMarker, Split, Stack, Tabs, TabsHost, TabsScreen, TabsAccessory, TabsBottomAccessory, etc. |
| Component helpers | 10 `.tsx` / `.ts` | parse/helper return types in `DebugContainer`, `FullWindowOverlay`, `ScreenFooter`, `ScreenStackHeaderConfig.web`, `ScreenStackItem`, `SearchBar`, `TabsScreen.android`, `TabsScreen.ios`, `useTabsScreen`, `useTabsHost` |
| Top-level Fabric specs | 6 | `ModalScreenNativeComponent`, `ScreenNativeComponent`, `ScreenStackHeaderConfigNativeComponent`, `ScreenStackHeaderSubviewNativeComponent`, `ScreenStackNativeComponent`, `SearchBarNativeComponent` |
| Nested Fabric specs | 9 | `gamma/split/*` (2), `gamma/stack/StackScreenNativeComponent`, `safe-area/SafeAreaViewNativeComponent`, `tabs/*` (5: TabsBottomAccessory, TabsHostAndroid/IOS, TabsScreenAndroid/IOS) |

All changes are type annotations only — zero runtime changes.

### Codegen compatibility

The `src/fabric/` spec files are consumed by `@react-native/codegen`. Adding `| undefined` to optional properties is safe — the codegen parser ([`parseTopLevelType.js`](https://github.com/facebook/react-native/blob/main/packages/react-native-codegen/src/parsers/typescript/parseTopLevelType.js)) explicitly strips `undefined` from union types and treats the result identically to `prop?: T`.

**Exception:** `CT.WithDefault<T, V>` props are NOT modified — `WithDefault` already expands to `T | undefined | null` so it's already eopt-compatible, and combining it with `| undefined` crashes the codegen parser ("WithDefault<> is optional and does not need to be marked as optional"). Commit `bebd04cba` removes the stray `| undefined` additions after any `CT.WithDefault<>` for exactly this reason.

I verified the codegen schema output is **byte-identical** between `main` and this branch across all 23 current Fabric spec files:

```bash
FABRIC_FILES=$(find src/fabric -name "*NativeComponent.ts" -type f | tr '\n' ' ')
node node_modules/@react-native/codegen/lib/cli/combine/combine-js-to-schema-cli.js \
  /tmp/schema.json --libraryName rnscreens $FABRIC_FILES
```

Run on both `origin/main` and this branch, `diff -q` exits 0.

## Test plan

- `yarn check-types` passes with 0 errors (with `exactOptionalPropertyTypes` enabled)
- `yarn lint-js` passes with 0 errors and the same 18 pre-existing warnings as `origin/main`
- `yarn prettier --check src/` is clean
- Codegen schema byte-identical to `main` across all 23 `src/fabric/**NativeComponent.ts` files

## Checklist

- [x] Ensured that CI passes (pending maintainer approval for fork-PR workflow runs)


src/components/tabs/TabsScreen.tsx
```diff
@@ -280,17 +280,17 @@ function parseIconsToNativeProps(
...
-  imageIconResource?: ImageResolvedAssetSource;
-  drawableIconResourceName?: string;
-  iconType?: IconType;
-  iconImageSource?: ImageSourcePropType;
-  iconResourceName?: string;
+  imageIconResource?: ImageResolvedAssetSource | undefined;
+  drawableIconResourceName?: string | undefined;
+  iconType?: IconType | undefined;
+  iconImageSource?: ImageSourcePropType | undefined;
+  iconResourceName?: string | undefined;
```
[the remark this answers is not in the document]
YevheniiKotyrlo: Good observation, but this is a false positive. `parsedIconResource` is the return value of `Image.resolveAssetSource()`, which returns `ImageResolvedAssetSource` — an object type that is always truthy. TypeScript's control flow analysis correctly narrows `parsedIconResource || undefined` to just `ImageResolvedAssetSource` (the `|| undefined` branch is dead code from TS's perspective), so there is no TS2375 violation here.

`yarn check-types` passes with 0 errors on this branch, confirming no issue.

src/components/safe-area/SafeAreaView.types.ts
```diff
@@ -8,7 +8,7 @@ export type Edge = 'top' | 'right' | 'bottom' | 'left';
 export type InsetType = 'all' | 'system' | 'interface';
 
 export interface SafeAreaViewProps extends ViewProps {
-  edges?: Readonly<Partial<Record<Edge, boolean>>>;
+  edges?: Readonly<Partial<Record<Edge, boolean>>> | undefined;
   // Android-only
   insetType?: InsetType;
```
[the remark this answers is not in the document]
YevheniiKotyrlo: Good catch — fixed. Added `| undefined` to `insetType` in `SafeAreaViewProps`. The fabric file's `insetType` uses `CT.WithDefault<InsetType, 'all'>` which is already eopt-compatible (expands to `InsetType | undefined | null`), so only the public types file needed the change.

Also applied the same consistency fix to `GestureProps` — added `| undefined` to the 2 remaining optional props (`screensRefs`, `screenEdgeGesture`) that didn't cause internal errors but could affect downstream consumers.

Additionally rebased on latest `main` and fixed 7 new eopt errors from the tabs refactoring (#3613, #3714, #3716). All quality gates pass, codegen output is byte-identical to `main`.

kligarski:
Hi, @YevheniiKotyrlo. Thank you for the PR.

After internal discussion we decided to use `prop?: T | undefined` for all non-required props to provide consistent API, not only those that resulted in TS error in the current state. If git allows, I'll push some changes to finalize the PR. I hope you don't mind.

src/components/gamma/scroll-view-marker/ScrollViewMarker.types.ts
```diff
@@ -3,12 +3,14 @@ import type { ScrollEdgeEffect } from '../../shared/types';
 
 export interface ScrollViewMarkerProps {
   children: ViewProps['children'];
```
kligarski: We can consider adding `undefined` to this prop or removing nullability from `ViewProps['children']`.

kligarski (commented):
By the way, we use `in` in `ScreenStackHeaderConfig.tsx` which is sensitive to not defined vs `undefined` but we should be good here. `buttonId` and `menuId` are internal props set to some value and `items` is a required property.

YevheniiKotyrlo:
Thanks @kligarski — really appreciate you taking it further. Broadening to all non-required props is the right call for API consistency, no objections at all.

Just pushed a merge with latest `main` (`bf4c49ab3`) to unblock merge status. The 4 conflicts came from the `ScrollViewMarker` migration (#3895) and the `controlNavigationStateInJS` removal (#3888) — resolved by keeping your `| undefined` markers in `ScrollViewMarker.types.ts` under the new JSDoc, and dropping the now-deleted props from `TabsHost.types.ts` / `useTabsHost.ts` / `TabsScreen.ios.types.ts`.

Quality gates all pass locally: `yarn check-types` (0 errors), `yarn lint-js` (0 errors, 18 warnings — matches `origin/main` exactly), `yarn prettier --check src/` (clean), and codegen schema byte-identical to `main` across all 23 `src/fabric/**NativeComponent.ts` files. Good that you caught the `WithDefault<>` codegen trap — that's exactly the pitfall I flagged in the original PR description and avoided up front; glad your broader-scope commit handled it the same way (skip `| undefined` after any `CT.WithDefault<>`).

On the `in`-operator audit in `ScreenStackHeaderConfig.tsx` — agreed the three usages are safe: `buttonId`/`menuId` are internal props always set to a concrete value, and `items` is required. Good that you checked it explicitly; that's exactly the subtlety that tends to bite under `exactOptionalPropertyTypes` (explicit `undefined` vs missing key), so having it confirmed up front is worth it.

Also — I've refreshed the PR description to match the broader post-takeover scope (41 files, not the original 10) and to update the codegen verification command against the current Fabric file layout. The old description was stale.

CI workflows are currently in `action_required` state (fork PRs need a maintainer click to run) — could you approve the run when you get a chance? Ready for final review whenever you have time.

src/components/gamma/scroll-view-marker/ScrollViewMarker.types.ts
YevheniiKotyrlo: `children` is required here (no `?`), so `exactOptionalPropertyTypes` doesn't actually constrain it — the flag only affects optional props. `ViewProps['children']` is `React.ReactNode | undefined` (and `React.ReactNode` itself already includes `null | undefined`), so forwarded values stay assignable to this required prop regardless. Happy to leave as-is for this PR, or to make it optional (`children?: ViewProps['children']`) for consistency with the rest of the interface — let me know which you prefer.

kligarski: Yeah, for consistency it would probably be better to make it `children?: ViewProps['children']` or use `Exclude<>`.

cc @kkafar, @t0maboro

kligarski:
Thanks for updating the PR! I want to wait for one more review and we'll hopefully proceed.

src/components/gamma/scroll-view-marker/ScrollViewMarker.types.ts
t0maboro: for me the semantic matters here, based on the context why we decided to have children required here (what seems to be correct, for ScrollViewMarker, you're rather expecting to have a ScrollView inside always)
- if we leave it in a current form, we're signaling the caller, that from *our* perspective we're expecting that children will be always present (even if React allows `undefined`)
- adding optional on our side would be better only if we're having a valid use-case for this component to have no children

t0maboro: for this PR, I'd leave it as-is if the only reason for making it optional is to make it consistent with other components

kligarski: > if we leave it in a current form, we're signaling the caller, that from our perspective we're expecting that children will be always present (even if React allows undefined)

Using `Exclude<>` or `NonNullable<>` would make our intention even more clear. From our discussions I think that we wanted to treat `undefined` and not defined property in the same way. Here, we're allowing `undefined` but you'll get an error if you don't pass the property.

t0maboro: Okay, ideally, I'd go for `NonNullable` in this case, because I don't see any use-case for this component to render without children. However, please remember that it will be a breaking change, because `ReactNode` allowed to pass `undefined` explicitly, so we can do that only for experimental components. If it touches more components, I'll be okay with aligning all of them with `children?:`, but we should pay more attention to that regarding v5.

I'll accept both
`children?: ViewProps['children']` 
and
`children: NonNullable<ViewProps['children']>`

t0maboro (approved):
thanks!

src/components/gamma/scroll-view-marker/ScrollViewMarker.types.ts
YevheniiKotyrlo: Went with `children: NonNullable<ViewProps['children']>` — pushed as `166affb28`.

Your preferences converge cleanly here: @kligarski wanted `Exclude<>` / `NonNullable<>` to make the intent explicit, and @t0maboro's caveat was that rejecting explicit `undefined` (which `ReactNode` permits) is a breaking change only acceptable on experimental components. `ScrollViewMarker` lives under `src/components/gamma/` so that gate is satisfied, and the type now expresses the semantic @t0maboro described — a marker for a ScrollView always has children, and the type reflects that rather than relying on `ReactNode`'s permissive branch.

Quality gates green locally: `yarn check-types` (0 errors), `yarn lint-js` (0 errors, 18 pre-existing warnings matching `origin/main` exactly), `yarn prettier --check src/` clean. One-line diff — no `src/fabric/` changes, so codegen schema is unchanged.

kligarski:
CI fails for unrelated reasons, last commit introduces only changes in JS types. CI passed correctly commit before it so I think that we're good to merge here.

PR #3720, opened 2026-03-03
fix(examples): Convert object to array for props.scenarioGroup and props.scenario

t0maboro:
## Description

related to this PR: https://github.com/software-mansion/react-native-screens/pull/3678 .

Now both are objects, so `map` function won't work properly, converting to array.

Note: typechecks aren't satisfied, but:
- weren't satisfied before my fix
- it potentially blocks SFT for Appearance

## Changes

- object => array

N/A

## Test plan

Try launching any SFT

## Checklist

- [ ] Ensured that CI passes


kkafar:
I never launch them via Example app though

PR #3729, opened 2026-03-05
feat(iOS, Tabs): Migrate tabs to new invalidate callback on Fabric

t0maboro:
## Description

We're now aligning the logic to use the invalidate callback from `RCTComponentViewProtocol`. Additionally, I added `dispatch_async` to ensure all transitions are handled before our cleanup, though I'm not sure we really need it. I'd prefer to keep that for safety reasons, since from a memory-management perspective it won't hurt to free some resources a bit later, and we may avoid crashes that would be hard to debug.

In this PR, I'm cleaning up the old method with the components registry, as we can clean up resources inside components' internal implementations.

## Changes

- overriden `invalidate` callback from `RCTComponentViewProtocol`
- moved the direct logic from the callback inside the `dispatch_async` block

N/A - transparent from the perspective of our examples

## Test plan

Really hard to propose a dedicated test for that, I just launched the app navigated to the screen that has Tabs (and BottomAccessory) and verified that breakpoints are hit on screen dismissal.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/tabs/bottom-accessory/RNSTabsBottomAccessoryComponentView.mm
+++ b/ios/tabs/bottom-accessory/RNSTabsBottomAccessoryComponentView.mm
@@ -74,7 +74,7 @@ namespace react = facebook::react;
   if (self.window != nil) {
     [_helper registerForAccessoryFrameChanges];
   } else {
-    [self invalidateImpl];
+    [self invalidate];
   }
 }
 


ios/tabs/bottom-accessory/RNSTabsBottomAccessoryComponentView.mm
```diff
@@ -74,11 +74,7 @@ - (void)didMoveToWindow
   if (self.window != nil) {
     [_helper registerForAccessoryFrameChanges];
   } else {
-#if RCT_NEW_ARCH_ENABLED
-    [self invalidateController];
-#else // RCT_NEW_ARCH_ENABLED
     [self invalidate];
```
kligarski: Should we call `invalidate` or `invalidateImpl` here? I don't have any particular opinion here but I can see both being used.

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3729/commits/256b5af774050e3ee9d04d045c5173cfe7127993

ios/tabs/bottom-accessory/RNSTabsBottomAccessoryComponentView.mm
```diff
@@ -74,11 +74,7 @@ - (void)didMoveToWindow
   if (self.window != nil) {
     [_helper registerForAccessoryFrameChanges];
   } else {
-#if RCT_NEW_ARCH_ENABLED
-    [self invalidateController];
-#else // RCT_NEW_ARCH_ENABLED
-    [self invalidate];
-#endif // RCT_NEW_ARCH_ENABLED
+    [self invalidateImpl];
```
kkafar: Is `invalidateImpl` called intentionally here or should it be `invalidate`?

ios/tabs/host/RNSTabsHostComponentView.mm
```diff
@@ -122,17 +112,21 @@ - (void)resetProps
...
-#endif // RCT_NEW_ARCH_ENABLED
+  // We want to run after container updates are performed (transitions etc.)
+  __weak auto weakSelf = self;
+
+  dispatch_async(dispatch_get_main_queue(), ^{
+    auto strongSelf = weakSelf;
+    if (strongSelf) {
+      strongSelf->_controller = nil;
+    }
+  });
```
kkafar, line 125: I think this is fine, just to be safe.

My only worry here, is that now we call this cleanup when we receive `Delete` mutation from React. Earlier we cleaned things up when the views were moved from window - this is different. Please make sure that e.g. nothing disappears mid transition visually.

ios/tabs/bottom-accessory/RNSTabsBottomAccessoryComponentView.mm
kkafar: I mean calling `invalidateImpl` here is obviously fine, I'd just like consistent usage and usually convention is to not call the `xxxImpl` (or often called `xxxInner`) method directly, but to call the actual function. 

Leaving this as a nitpick. Up to you to decide I will be fine with anything you choose.

t0maboro: https://github.com/software-mansion/react-native-screens/pull/3729/changes/27d543a14ad1ba709d72aee5a5ffaa616144a28c

ios/tabs/host/RNSTabsHostComponentView.mm
t0maboro: Tested within our example app (Tabs examples are nested in Stacks) - the transition is smooth and nothing is unmounted during that.

PR #3733, opened 2026-03-06
feat(JS) - `RNSLog` utility

sgaczol:
## Description

This PR introduces `RNSLog` - JS logger based on JS Console methods.

It can be activated by setting `debugLogging` flag to true:
`featureFlags.stable.debugLogging = true;`

Closes https://github.com/software-mansion/react-native-screens-labs/issues/119

## Changes

- update `flags.ts`, `utils.ts` and `index.tsx` in `src` directory
- set mentioned flag in `App.tsx` to true

N/A

## Test plan

Check if `RNSLog` activates/deactivates, when `debugLogging` is set to true or not.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -31,7 +31,7 @@ export function parseBooleanToOptionalBooleanNativeProp(
 
 export const RNSLog = {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  log: (message?: any, ...args: unknown[]) => {
+  log: (message?: any, ...args: any[]) => {
     if (featureFlags.stable.debugLogging) {
       console.log(message, ...args);
     }


apps/App.tsx
```diff
@@ -1,11 +1,12 @@
 import React from 'react';
-import { enableFreeze } from 'react-native-screens';
+import { enableFreeze, featureFlags } from 'react-native-screens';
 
 import Example from './Example';
 // import { NavigationContainer } from '@react-navigation/native';
 // import { Tests } from './src/tests';
 
 enableFreeze(true);
+featureFlags.stable.debugLogging = true;
```
t0maboro, line 9: Generally, we're adding flags to https://github.com/software-mansion/react-native-screens/blob/4.24.0/FabricExample/App.tsx . Just wanted to confirm, this one is placed here to handle TVOSExample too, right?

src/utils.ts
```diff
@@ -27,3 +28,30 @@ export function parseBooleanToOptionalBooleanNativeProp(
       return 'false';
   }
 }
+
+export const RNSLog = {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  log: (message?: any, ...args: unknown[]) => {
```
t0maboro: Any reason to have `unknown` for `args`, when other methods are using `any`?

apps/App.tsx
sgaczol: yes

src/utils.ts
sgaczol: Oversight, fixed: https://github.com/software-mansion/react-native-screens/pull/3733/commits/fa722947eaa0883f271b011c18eee1614264b5c8

apps/App.tsx
t0maboro: okay, fine

src/index.tsx
```diff
@@ -59,6 +59,11 @@ export { compatibilityFlags, featureFlags } from './flags';
  */
 export { default as useTransitionProgress } from './useTransitionProgress';
 
+/**
+ * RNSLog
+ */
+export { RNSLog } from './utils';
```
kligarski, line 65: Should this be exported in public API? Shouldn't this be in `react-native-screens/private`?

apps/App.tsx
kkafar: This isn't fine. Let's keep the feature flags management in a single place. This introduces inconsistency in an **example** application. If you want to enable this flag in `TVOSExample` please do it there explicitly. 

`FabricExample` & `TVOSExample` apps are not only for internal usage. They're meant to showcase usage of our API & our recommendation for FF management is to put them in `App.tsx` of the application.

src/index.tsx
kkafar: Yes. It MUST NOT be exported as a part of the library public API.

If we want to export it from `react-native-screens/private` for usage in example apps - that's fine.

PR #3739, opened 2026-03-09
chore(JS): Deduplicate `FeatureFlags` exports

sgaczol:
## Description

This PR deduplicates `featureFlags` exports.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/699

## Changes

- remove default export from `flags.ts`
- update `featureFlags` imports

N/A

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/src/flags.ts
+++ b/src/flags.ts
@@ -230,5 +230,3 @@ export const featureFlags = {
     },
   },
 };
-
-export default featureFlags;
--- a/src/components/Screen.tsx
+++ b/src/components/Screen.tsx
@@ -29,7 +29,7 @@ import {
   resolveSheetLargestUndimmedDetent,
 } from './helpers/sheet';
 import { parseBooleanToOptionalBooleanNativeProp } from '../utils';
-import featureFlags from '../flags';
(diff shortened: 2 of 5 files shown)


src/flags.ts
```diff
@@ -230,5 +230,3 @@ export const featureFlags = {
     },
   },
 };
-
-export default featureFlags;
```
t0maboro, line 234: it's breaking change from the perspective of public API, the code looks good, but we shouldn't merge this PR to main as long as we're publishing 4.x from main

kkafar: Maybe you're right actually. TBH I'm no longer sure myself. We need to define in readme what is the public API - only things exported directly from `react-native-screens`.

t0maboro:
> Looks good. I classified this as a breaking change for some reason in the ticket description. Before merging, could you take a look at how Expo & react-navigation uses it. I'm sure that `Expo` does, `react-navigation` might not use the feature flags, but not sure.
> 
> Please check whether this change will break them or not.

Please keep in mind that feature flags can be used inside end-users' apps, not only downstream libs

PR #3744, opened 2026-03-10
chore(JS): `RNSLog` remove from public API

sgaczol:
## Description

This PR moves `RNSLog` utility out of the library's public API as discussed there:
https://github.com/software-mansion/react-native-screens/pull/3733#pullrequestreview-3915092551

## Changes

- move `RNSLog` to `react-native-screens/private`
- remove `bottomTabsDebugLog`
- update `useRenderDebugInfo`
- update logging in `StackContainer` and `BottomTabsContainer`

N/A

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/private/logging.ts
+++ b/src/private/logging.ts
@@ -1,4 +1,4 @@
-import { featureFlags } from 'react-native-screens';
+import { featureFlags } from '../flags';
 
 export const RNSLog = {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any


src/private/logging.ts
```diff
@@ -1,13 +1,28 @@
-let isDetailedLoggingEnabled = false;
+import { featureFlags } from 'react-native-screens';
```
t0maboro: inside the scope of lib sources, we should import using relative paths

kkafar: @t0maboro is absolutely right here, good catch

sgaczol: solved: https://github.com/software-mansion/react-native-screens/pull/3744/commits/34f3c788cecfca79a552cfce8d3a90a03d7553b5

PR #3749, opened 2026-03-11
chore: format SplitView conversions

kligarski:
## Description

Every time I commit something this file gets reformatted and it's annoying.

## Changes

I ran `yarn format` on the repo.


kligarski:
> Okay, I hope now other people won't have this problem, because it e.g. does not happen for me. Let's try it out.

Would you be able to run `yarn format` on this PR and check whether it gets reverted back? Maybe it's some formatter version mismatch one again, I remember similar situation some time ago.

kkafar:
Yeah, gimme a minute

PR #3751, opened 2026-03-11
chore(Android): Remove BottomSheet-related Paper-specific code

kmichalikk:
Closes https://github.com/software-mansion/react-native-screens-labs/issues/1031

## Description

This PR removes Paper-specific code from BottomSheet implementation on android.

> [!note]
> I was 50/50 about changing `Screen.kt` here, but I figured it's better for testing to have most of the changes in one place. I only touched things that directly relate to formSheet presentation. The `dispatchShadowStateUpdate()` remains Fabric/Paper, though it should use Fabric (new arch guard).

## Changes

Removed new arch guards and related code for BottomSheet from `SheetUtils.kt` and `Screen.kt`.

n/a

## Test plan

> [!note]
> Please DO test the runtime.

Compile & run the app. See if various bottomsheet presentation options work same as before.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -38,7 +38,6 @@ import com.swmansion.rnscreens.bottomsheet.usesFormSheetPresentation
 import com.swmansion.rnscreens.events.HeaderHeightChangeEvent
 import com.swmansion.rnscreens.events.SheetDetentChangedEvent
 import com.swmansion.rnscreens.ext.asScreenStackFragment
-import com.swmansion.rnscreens.ext.parentAsViewGroup
 import com.swmansion.rnscreens.gamma.common.FragmentProviding
 import com.swmansion.rnscreens.utils.getDecorViewTopInset
 import kotlin.math.max
@@ -99,10 +98,6 @@ class Screen(
     var sheetDefaultResizeAnimationEnabled = true
 
     /**
-     * On Paper, when using form sheet presentation we want to delay enter transition in order
-     * to wait for initial layout from React, otherwise the animator-based animation will look
-     * glitchy.
-     *
      * On Fabric, the view layout is completed before window insets are applied.
      * To ensure the BottomSheet correctly respects insets during its enter transition,
      * we delay the transition until both layout and insets have been applied.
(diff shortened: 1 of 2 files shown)


kmichalikk:
## Testing
Results look sensible

The recordings below doesn't show it, but there is a not present:

image

## Test3336

> [!note]
> - fitToContent (with and w/out input) doesn't work -> entirely different issue
> - for partiallyCoveredStatusBar it actually looks like:
> image

### SAV: on, edges: on, overflows: off

https://github.com/user-attachments/assets/2c14d59d-6543-4640-a10a-5d24060dfdf2

### SAV: off, edges: on, overflows: off

https://github.com/user-attachments/assets/9b08a84e-c970-40bc-92ac-82443872e5e6

### SAV: off, edges: off, overflows: off

https://github.com/user-attachments/assets/8acebd3c-9b12-4334-addf-420777adbcd9

### SAV: off, edges: off, overflows: on

https://github.com/user-attachments/assets/834c7ed7-274e-470c-93de-6bb019d17434

## Test3617

https://github.com/user-attachments/assets/56ace15b-18c7-40fb-99b9-8478fef020a4

## Test3564

https://github.com/user-attachments/assets/1b1ee8b8-4475-43ff-bcfe-c7dc1f28bdcf

kkafar:
Test3336 ->

> fitToContent (with and w/out input) doesn't work -> entirely different issue

Is this reported somewhere, do we have ticket for this?

kmichalikk:
> Is this reported somewhere, do we have ticket for this?

https://github.com/software-mansion/react-native-screens-labs/issues/968

t0maboro:
> for partiallyCoveredStatusBar it actually looks 

note: this TC makes sense only with enabled overflow

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -300,9 +284,7 @@ class Screen(
         // internal offsets with the new maxHeight. This prevents the sheet from snapping back
         // to its old position when the user starts a gesture.
         parent.requestLayout()
-        if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
-            updateScreenSizeFabric(width, clampedHeight, top + translationY.toInt())
-        }
+        updateScreenSizeFabric(width, clampedHeight, top + translationY.toInt())
```
t0maboro, line 287: just a note, but unrelated to this PR - we may consider dropping Fabric suffix as we have single architecture

kmichalikk: I assume this will be done in https://github.com/software-mansion/react-native-screens-labs/issues/1029

t0maboro: perfect, thanks!

kmichalikk:
I can confirm that https://github.com/software-mansion/react-native-screens-labs/issues/1049 happens also on main, so this is not the issue introduced here.

PR #3759, opened 2026-03-13
refactor(example): align TabsContainer structure with StackContainer

kkafar:
## Description 

Refactor TabsContainer to match the abstractions and directory structure introduced by StackContainer.

### Structure & API refactor

- Extract types into `TabsContainer.types.tsx` (`TabRouteConfig`, `TabRouteOptions`, `TabsContainerProps`, `SetTabOptionsMethod`)
- Add `contexts/TabsNavigationContext.tsx` providing `routeKey`, `routeOptions` and `setRouteOptions` per tab screen instance
- Add `contexts/TabsRouteConfigContext.tsx` for dynamic route config updates
- Add `hooks/useTabsNavigationContext` and `useTabsRouteConfigContext` consumer hooks with null-check throws
- Add `TabsContainerWithDynamicRouteConfigs` wrapper, mirroring `StackContainerWithDynamicRouteConfigs`
- Export clean public API via `index.ts`

### Route abstraction & state management

- Introduce `TabRoute` as a runtime instance type: `TabRouteConfig` (blueprint) extended with a generated `routeKey` string — mirroring the `StackRouteConfig` / `StackRoute` distinction
- Introduce `TabsNavState` holding `routes: TabRoute[]` and `selectedRouteKey: string`
- Replace `useState` with `useReducer` (`tabsNavigationReducer`) supporting two actions:
  - `'change-tab'` — updates `selectedRouteKey` in response to native focus change events
  - `'set-options'` — merges partial options into a specific route instance; wired to `setRouteOptions` in `TabsNavigationContext`, making it functional (was a `throw` placeholder before)
- Add `useSanitizeRouteConfigs` guard (non-empty + unique names), mirroring Stack's pattern
- Move `id-generator` from `stack/utils/` to `containers/shared/` so it is shared between Stack and Tabs

### Fixes & improvements

- Fix `updateRouteConfigWithOptions` in both `TabsContainerWithDynamicRouteConfigs` and `StackContainerWithDynamicRouteConfigs` to merge options instead of replacing them entirely
- Add `[Tabs]`/`[Stack]` prefix to error messages to follow project convention
- Document intentional one-time props initialization in both `WithDynamicRouteConfigs` components

### Breaking changes in TabsContainer API

- `TabConfiguration` → `TabRouteConfig`; `component` → `Component`; `options.screenKey` moves to top-level `name`; `options.isFocused` replaced by `initialFocusedName` prop on the container
- `tabConfigs` prop → `routeConfigs`
- `SetTabOptionsMethod` first param: `name: string` → `routeKey: string`
- `ConfigWrapperContext.controlledBottomTabs` removed; pass `experimentalControlNavigationStateInJS` directly to `TabsContainer`
- `TabConfigUpdate.screenKey` → `name`

Migrate all existing usages (37 files) to the new API.

## Test plan

I've verified that existing tests keep working. I haven't gone through every single one, but used tested few of those with more complicated setup (e.g. `test-tabs-color-scheme`) and verified that it works on both platforms.

I've found that tests dedicated to tabs orientation fail. But it's not a regression in introduced by this PR. I've confirmed that it is also present on main. Opened ticket for this: https://github.com/software-mansion/react-native-screens-labs/issues/1052

https://github.com/user-attachments/assets/69ae22ab-8b91-4b11-9e6b-a7f203cdac00

Also works on Android.


apps/src/shared/gamma/containers/tabs/reducer.tsx
```diff
@@ -0,0 +1,131 @@
...
+    case 'change-tab':
+      return tabsActionChangeTabHandler(state, action);
+    case 'set-options':
+      return tabsActionSetOptionsHandler(state, action);
+  }
+
+  // @ts-ignore
+  throw new Error(
+    `[Tabs] Unhandled navigation action: ${JSON.stringify(action)}`,
+  );
```
t0maboro, line 27: can be moved to `switch` block as `default`

apps/src/tests/single-feature-tests/tabs/test-tabs-simple-nav.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export default SCENARIO;
+
+export function App() {
+  return (
+    <View>
+      <Text>Hello world</Text>
+    </View>
+  );
+}
```
kligarski, line 21: `test-tabs-simple-nav` does not contain tabs?

apps/src/shared/gamma/containers/tabs/reducer.tsx
```diff
@@ -0,0 +1,131 @@
...
+  arg: TabsNavStateInitArg,
+): TabsNavState {
+  const { routeConfigs, initialFocusedName } = arg;
+
+  const routes = routeConfigs.map(createTabRouteFromConfig);
+
+  let selectedRouteKey: string;
+  if (initialFocusedName != null) {
+    const matchingRoute = routes.find(r => r.name === initialFocusedName);
+    if (matchingRoute == null) {
```
kligarski, line 120: I know that this works but I'm not the biggest fan of comparing with `==` to `null` when `find` returns route or `undefined`. Not really that important though.

apps/src/shared/gamma/containers/tabs/reducer.tsx
```diff
@@ -0,0 +1,131 @@
...
+  throw new Error(
+    `[Tabs] Unhandled navigation action: ${JSON.stringify(action)}`,
+  );
+}
+
+export function tabsNavigationReducerWithLogging(
+  state: TabsNavState,
+  action: TabsNavigationAction,
+): TabsNavState {
+  console.debug(`[Tabs] Handling action: ${JSON.stringify(action)}`);
```
kligarski, line 34: In `TabsContainer` we're using `RNSLog`, should we use it here as well? But I see that we don't have `RNSLog.debug` right now.

apps/src/shared/gamma/containers/tabs/reducer.tsx
```diff
@@ -0,0 +1,131 @@
...
+    throw new Error(
+      `[Tabs] Cannot set options. Route with key "${action.routeKey}" not found`,
+    );
+  }
+
+  const routeCopy = { ...state.routes[routeIndex] };
+  routeCopy.options = {
+    ...routeCopy.options,
+    ...action.options,
+  };
```
kligarski, line 90: This is out of scope of this PR but shouldn't we handle nested props, especially if we have `android` and `ios` props?

apps/src/shared/gamma/containers/tabs/TabsContainerWithDynamicRouteConfigs.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { TabsContainer } from './TabsContainer';
+import type { TabsContainerProps, TabRouteOptions } from './TabsContainer.types';
+import { TabsRouteConfigContext } from './contexts/TabsRouteConfigContext';
+
+/**
+ * TabsContainer wrapped in context allowing for modifying options
+ * of route configs.
+ * See TabsRouteConfigContext.
+ */
+export function TabsContainerWithDynamicRouteConfigs(
```
kligarski, line 11: I don't think that I understand why this is useful? `TabsContainer` generates routes at the beginning based on configs, why would we need to change the config? We can't add or remove tabs dynamically either way, right? Also, shouldn't there be some kind of test to check if the implementation works? I don't see `TabsContainerWithDynamicRouteConfigs` being used anywhere in the repo.

apps/src/tests/single-feature-tests/tabs/test-tabs-simple-nav.tsx
kkafar: Will be filled up in later PRs. It's a minor problem, I decide to not rebase for this.

apps/src/shared/gamma/containers/tabs/reducer.tsx
kkafar: https://github.com/software-mansion/react-native-screens/pull/3790

apps/src/shared/gamma/containers/tabs/reducer.tsx
kkafar: @sgaczol can we align it? 

We need to add `RNSLog.debug`, second we need to introduce consistent rules for logging in our applications.

First - all library-code logs should use the `RNSLog` object.

Second - I think that application code should not use the `RNSLog` - this is an internal library utility and does not give us much in context of example applications. If we happen to change our opinion later it's really easy to change.

I won't push changes here - it's a minor issue.

apps/src/shared/gamma/containers/tabs/reducer.tsx
kkafar: Yes, we should. https://github.com/software-mansion/react-native-screens-labs/issues/1066

apps/src/shared/gamma/containers/tabs/TabsContainerWithDynamicRouteConfigs.tsx
kkafar: You're right. Removed here: https://github.com/software-mansion/react-native-screens/pull/3790.

I've initially planned to support dynamic tab count, but deferred the implementation in favour of other features right now. Will come back to this one day.

apps/src/tests/single-feature-tests/tabs/test-tabs-simple-nav.tsx
kligarski: Okay, just wanted to confirm.

apps/src/shared/gamma/containers/tabs/TabsContainerWithDynamicRouteConfigs.tsx
kligarski: Okay. Do we have a ticket for dynamic tab configurations?

PR #3762, opened 2026-03-14
fix: use path-based menuId to prevent collisions in nested submenus

### LINKED ISSUE #3606: Multiple menus in unstable_headerRightItems cause handler index collision
### Description

When using multiple menus in `unstable_headerRightItems`, selecting an item from one menu triggers the handler from a different menu at the same index position. For example, selecting the first item in Menu A triggers the handler for the first item in Menu B.

Menu item handlers appear to be stored in a flat array, causing index collisions between different menus.

**Workaround:** Moving one menu to `unstable_headerLeftItems` resolves the issue, since left and right items appear to have separate handler arrays. Adding `identifier` properties to menu items did **not** resolve the issue.

### Steps to reproduce

1. Create a screen with two menus in `unstable_headerRightItems`:

```tsx
navigation.setOptions({
  unstable_headerRightItems: () => [
    {
      type: "menu",
      label: "Filter",
      icon: { type: "sfSymbol", name: "line.3.horizontal.decrease" },
      menu: {
        items: [
          {
            type: "action",
            label: "Sort by date",
            onPress: () => console.log("Sort by date"), // Index 0 in Menu 1
          },
          {
            type: "action",
            label: "Sort by name",
            onPress: () => console.log("Sort by name"), // Index 1 in Menu 1
          },
        ],
      },
    },
    {
      type: "menu",
      label: "Add",
      icon: { type: "sfSymbol", name: "plus" },
      menu: {
        items: [
          {
            type: "action",
            label: "Import from URL",
            onPress: () => console.log("Import URL"), // Index 0 in Menu 2
          },
          {
            type: "action",
            label: "Enter manually",
            onPress: () => console.log("Enter manually"), // Index 1 in Menu 2
          },
        ],
      },
    },
  ],
});
```

2. Open the Filter menu and tap "Sort by date"
3. Observe that "Import URL" is logged instead of "Sort by date"

_No response_

### Screens version

4.19.0

### React Native version

0.81.5

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo

### Architecture

New Architecture

### Build type

Debug mode

### Device

iOS simulator

### Device model

_No response_

### Acknowledgements

Yes

chrism:
Closes: https://github.com/software-mansion/react-native-screens/issues/3606

## Summary

- `prepareMenu` passes `menuIndex` as `index` when recursing into submenus, losing the original bar button item index
- This causes `menuId` collisions between actions in different submenus and sibling toolbar menus at the same position
- Fix threads a `path` string through recursion to track nesting depth, while preserving the original bar button item index

## Problem

Given two toolbar menu buttons — a filter menu at bar button index 1 containing submenus, and a sort menu at bar button index 2:

```
Toolbar:
  [+] button (index 0)
  [filter] menu (index 1)     ← contains Location, Item Type, Quality submenus
  [sort] menu (index 2)       ← contains Title, Gear Type, Date Added actions
```

When `prepareMenu` recurses into the Quality submenu (at `menuIndex` 2 inside the filter menu), it calls:
```ts
prepareMenu(menuItem, menuIndex, side)  // menuIndex=2, replacing index=1
```

The first action inside Quality then gets `menuId: "0-2-right"`, which collides with the first action in the sort menu (`"0-2-right"`). Selecting a quality filter fires the sort handler instead.

## Fix

Add a `path` parameter that tracks position through nesting, while keeping the original bar button `index`:

```ts
const prepareMenu = (
  menu, index, side,
  path = '',  // NEW
) => {
  return {
    ...menu,
    items: menu.items.map((menuItem, menuIndex) => {
      const currentPath = path ? `${path}.${menuIndex}` : `${menuIndex}`;
      // ...
      if (menuItem.type === 'submenu') {
        return {
          ...menuItem,
          ...prepareMenu(menuItem, index, side, currentPath),
          //                       ^^^^^ preserved (was menuIndex)
        };
      }
      return {
        ...menuItem,
        menuId: `${currentPath}-${index}-${side}`,
        //        ^^^^^^^^^^^^ path-based (was menuIndex)
      };
    }),
  };
};
```

**Before:** Quality "All" = `0-2-right` (collides with Sort "Title" = `0-2-right`)
**After:** Quality "All" = `2.0-1-right`, Sort "Title" = `0-2-right` (unique)

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/44420c7e-7850-45be-b4b2-558c38da7b1a" /> | <video src="https://github.com/user-attachments/assets/a0deac60-ac7e-4197-91fe-6943a875d06c" /> |

## Verified

- Toolbar with multiple menu buttons — actions in each menu fire the correct handler
- Nested submenus (inline sections) within a toolbar menu — no cross-talk with sibling toolbar menus
- Single menu button with no submenus — no regression (path defaults to menuIndex)


t0maboro:
Hi, thanks for fixing that. This change seems to be reasonable. Let me push an example for regression testing, and I think that we could move this PR forward.

t0maboro (approved):

PR #3764, opened 2026-03-15
refactor(example): align TabsContainer with StackContainer & retire TabsConfigProvider

kkafar:
## Description

Further aligns `TabsContainer` with the existing `StackContainer` abstractions
and retires the `TabsConfigProvider` / `TabsAutoconfig` test infrastructure in
favour of direct container APIs.

The old infrastructure required tests to define a typed param list, wrap their
app in a `<Tabs.Provider>`, and dispatch named actions through a shared reducer
to change tab bar or per-screen options. This was indirect, required knowing
route names at dispatch time, and leaked the infra shape into every test.
The new APIs (`useTabsNavigationContext`, `useTabsHostConfig`) let each screen
update its own state directly via React context — no param lists, no dispatch,
no shared reducer.

## Changes

- Move `safeAreaConfiguration` from top-level `TabRouteConfig` into
  `TabRouteOptions` so it is part of the per-route options object and can be
  updated at runtime via `setRouteOptions`; strip it before spreading onto
  `<Tabs.Screen>` as it is a JS-only wrapping concern.
- Use `name` as `routeKey` in `createTabRouteFromConfig` — tab names are
  enforced to be unique, so no separate generated ID is needed.
- Add `TabsHostConfig` type and `TabsContainerWithHostConfigContext` component:
  initialises host config state from props, provides it via
  `TabsHostConfigContext`, and re-passes the merged config to `TabsContainer`.
  Keeps `TabsContainer` itself free of context management concerns.
- Add `useTabsHostConfig` hook and `DEFAULT_TAB_ROUTE_OPTIONS` preset.
- Adopt React 19 context syntax (`<Context value={...}>`) in `TabsContainer`.
- Migrate all 8 affected test files off `TabsConfigProvider`/`TabsAutoconfig`:
  - `tabBar` dispatches → `useTabsHostConfig` + `updateHostConfig`
  - `tabScreen` dispatches → `useTabsNavigationContext` + `setRouteOptions`
  - Orientation integration tests combine both context hooks directly,
    removing name-based lookups (`findTabScreenOptions`, `findStackScreenOptions`).
- Move one-time-init notes into JSDoc on all three `With*` wrapper components.

## Test plan

Tested manually with the FabricExample app on Android and iOS:

- `apps/src/tests/single-feature-tests/tabs/test-tabs-color-scheme.tsx`
- `apps/src/tests/single-feature-tests/tabs/test-tabs-ime-insets.tsx`
- `apps/src/tests/single-feature-tests/tabs/test-tabs-layout-direction.tsx`
- `apps/src/tests/single-feature-tests/tabs/tab-bar-hidden.tsx`
- `apps/src/tests/single-feature-tests/tabs/tabs-screen-orientation.tsx`
- `apps/src/tests/single-feature-tests/tabs/bottom-accessory-layout.tsx`
- `apps/src/tests/component-integration-tests/orientation/orientation-stack-in-tabs.tsx`
- `apps/src/tests/component-integration-tests/orientation/orientation-tabs-in-stack.tsx`

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/apps/src/shared/utils/deep-merge.ts
+++ b/apps/src/shared/utils/deep-merge.ts
@@ -5,16 +5,7 @@ export function deepMerge<T extends object>(base: T, override: Partial<T>): T {
   for (const key of Object.keys(override) as (keyof T)[]) {
     const baseVal = base[key];
     const overrideVal = override[key];
-    if (
-      overrideVal !== null &&
-      typeof overrideVal === 'object' &&
-      !Array.isArray(overrideVal) &&
-      !React.isValidElement(overrideVal) &&
-      baseVal !== null &&
-      typeof baseVal === 'object' &&
-      !Array.isArray(baseVal) &&
-      !React.isValidElement(baseVal)
-    ) {
+    if (isMergeableObject(overrideVal) && isMergeableObject(baseVal)) {
       result[key] = deepMerge(
         baseVal as object,
         overrideVal as object,
@@ -24,5 +15,14 @@ export function deepMerge<T extends object>(base: T, override: Partial<T>): T {
     }
   }
   return result;
+}
+
+function isMergeableObject(val: unknown): val is object {
+  return (
+    val !== null &&
+    typeof val === 'object' &&
+    !Array.isArray(val) &&
+    !React.isValidElement(val)
+  );
 }
 


apps/src/shared/gamma/containers/tabs/TabsContainer.types.tsx
kligarski, line 73: This is already in `TabsHostProps` I think.

apps/src/shared/gamma/containers/tabs/TabsContainerWithHostConfigContext.tsx
```diff
@@ -0,0 +1,54 @@
...
+export function TabsContainerWithHostConfigContext(props: TabsContainerProps) {
+  const {
+    routeConfigs,
+    initialFocusedName,
+    experimentalControlNavigationStateInJS,
+    ...hostProps
+  } = props;
+
+  const [hostConfig, setHostConfig] = React.useState<TabsHostConfig>(hostProps);
+
```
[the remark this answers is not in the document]
kligarski: sounds reasonable?

apps/src/shared/gamma/containers/tabs/TabsContainerWithHostConfigContext.tsx
```diff
@@ -0,0 +1,54 @@
...
+    [],
+  );
+
+  const tabsHostConfigContext: TabsHostConfigContextPayload = React.useMemo(
+    () => ({ hostConfig, updateHostConfig }),
+    [hostConfig, updateHostConfig],
+  );
+
+  return (
+    <TabsHostConfigContext value={tabsHostConfigContext}>
```
t0maboro, line 42: `TabsHostConfigContext.Provider`?

apps/src/shared/utils/deep-merge.ts
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+export function deepMerge<T extends object>(base: T, override: Partial<T>): T {
+  const result = { ...base };
+  for (const key of Object.keys(override) as (keyof T)[]) {
```
t0maboro, line 5: just to be aware - other solutions are filtering out some keys: https://github.com/voodoocreation/ts-deepmerge/blob/master/src/index.ts#L64-L66

apps/src/shared/utils/deep-merge.ts
```diff
@@ -0,0 +1,28 @@
...
+    const baseVal = base[key];
+    const overrideVal = override[key];
+    if (
+      overrideVal !== null &&
+      typeof overrideVal === 'object' &&
+      !Array.isArray(overrideVal) &&
+      !React.isValidElement(overrideVal) &&
```
t0maboro: These checks are repeated for overrideVal and baseVal, we may consider extracting it to helper method

apps/src/shared/gamma/containers/tabs/TabsContainerWithHostConfigContext.tsx
kkafar: You don't need that since React 19 IIRC

apps/src/shared/utils/deep-merge.ts
kkafar: Seems like a great catch. Thank you

kkafar: My Claude boy tells me something like this: 

```
⏺ This is a prototype pollution concern. The attack vector:

  deepMerge({}, JSON.parse('{"__proto__": {"isAdmin": true}}'))

  Object.keys() will include __proto__, constructor, and toString-style keys when they appear as own string properties on a parsed object.
  Writing to result["__proto__"] doesn't set an own property — it walks up and mutates Object.prototype, poisoning every object in the
  process.

  How serious is it here? In this specific codebase the inputs are typed TabsHostConfig objects constructed in-app, not from
  external/untrusted JSON, so the practical risk is low. But it's still worth fixing defensively.

  Recommended fix — skip any key that is a known dangerous prototype property:

  const UNSAFE_KEYS = new Set(['__proto__', 'constructor', 'prototype']);

  // inside the loop:
  if (UNSAFE_KEYS.has(key as string)) continue;

  Alternatively, use Object.hasOwn / Object.defineProperty patterns, but the blocklist is the simplest and clearest for a utility this
  small.

  Object.keys() already returns only own enumerable string keys, so prototype and constructor only appear if explicitly present as plain
  string properties (e.g. from JSON.parse). __proto__ is the most dangerous one — assigning to it via bracket notation is the classic
  pollution vector.

  Want me to apply this fix and amend/commit?
```

So it seems that Object.keys does not return these properties & this is done rather as a security measure than an implementation correctness. It seems that we might get away with the implementation in current shape.

apps/src/shared/utils/deep-merge.ts
kkafar: https://github.com/software-mansion/react-native-screens/pull/3764/commits/2a80499ff4f8bb64dbc3dcfb551a0af2289fae71

apps/src/shared/gamma/containers/tabs/TabsContainerWithHostConfigContext.tsx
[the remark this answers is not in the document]
kkafar: Will be handled here: https://github.com/software-mansion/react-native-screens/pull/3776/changes#diff-95cf96d604f8ee9419b5bb21ef89f3316c60602fd147e60e9a4422b0fdeef93c

apps/src/shared/utils/deep-merge.ts
t0maboro: sure

Issue #3568, opened 2026-01-22
[iOS] usePreventRemove callback not called when tapping formSheet backdrop


iankberry:
### Description

Opening this issue per [this related issue](https://github.com/react-navigation/react-navigation/issues/12940) in the react-navigation repository.

I am building a form sheet that contains an input component for the user to write a note. I am also using the `usePreventRemove` hook to prevent the user from accidentally losing data if they tap the backdrop as this immediately closes and unmounts the sheet.

On Android, the `usePreventRemove` hook properly prevents the sheet from closing and the `callback` is emitted allowing me to alert the user. However on iOS, while the `usePreventRemove` hook **does** prevent sheet from closing, the `callback` param is not called. It's also worth noting that if you attempt to interactively dismiss the sheet, the callback is properly called. So this seems to be related to the backdrop specifically.

Note, the version of RN screens in the snack is `4.16.0` but I also reproduced the issue with the latest version, `4.20.0`. For some reason the snack wasn't letting me upgrade to the latest version because it's not tested with Expo 54.

### Steps to reproduce

1. Open the snack in an iOS simulator
2. Tap "Open sheet"
3. Tap "Toggle prevent remove" in the sheet
4. Tap the backdrop, notice the sheet does not dismiss but nothing is logged
5. Attempt to interactively dismiss the sheet, the sheet does not dismiss and "Prevented sheet removal!" is logged

### Snack or a link to a repository

https://snack.expo.dev/@iankberry/formsheet-prevent-remove-cb

### Screens version

4.16.0

### React Native version

0.81.5

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo bare workflow

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Acknowledgements

Yes

kkafar:
Thanks for the report. We'll add it to the backlog.

t0maboro:
Hi @iankberry ,
Thanks for reporting that.
I'm working on the fix; it can already be tested as a patch. You can track the progress here: https://github.com/software-mansion/react-native-screens/pull/3771

PR #3771, opened 2026-03-17
fix(iOS, Stack v4): Fix `usePreventRemove` callback not called when tapping formSheet backdrop

t0maboro:
## Description

When using a `FormSheet` with `preventNativeDismiss` enabled, swiping down the sheet correctly triggers the event for dismiss cancellation to JS. However, tapping the dimming backdrop fails to trigger it.
When `presentationControllerShouldDismiss:` returns `NO`, UIKit cancels the dismissal. For swipe gestures, it still calls the `presentationControllerDidAttemptToDismiss` method. For backdrop taps, it simply drops the touch event entirely without any notification.
We can fix this by attaching a custom `UITapGestureRecognizer` directly to the `containerView` (covering the whole Screen) and simply ignoring taps that are inside the `FormSheet` or its content, which ensures the gesture only recognizes taps on the dimming view. Then, we can send an event back to JS via `notifyDismissCancelledWithDismissCount` if `preventNativeDismiss` is enabled. 

Closes: https://github.com/software-mansion/react-native-screens/issues/3568

## Changes

- Added `backdropTapGestureRecognizer` and logic for sending proper event to the JS when backdrop was tapped under certain conditions.

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/aca68f50-f4a1-42b2-90ef-6658a074ec2c" /> | <video src="https://github.com/user-attachments/assets/b23f01e6-7a5f-499d-9299-ff945c850785" /> |

## Test plan

Added Test3568, verified that there's no regression in https://github.com/software-mansion/react-native-screens/pull/2129

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -885,7 +885,9 @@ RNS_IGNORE_SUPER_CALL_END
 
 - (void)setupBackdropTapGestureRecognizer
 {
-  if (self.stackPresentation != RNSScreenStackPresentationFormSheet) {
+  if (self.stackPresentation != RNSScreenStackPresentationFormSheet &&
+      self.stackPresentation != RNSScreenStackPresentationPageSheet &&
+      self.stackPresentation != RNSScreenStackPresentationModal) {
     return;
   }
 


ios/RNSScreen.mm
```diff
@@ -880,6 +883,32 @@ - (void)presentationControllerDidDismiss:(UIPresentationController *)presentatio
...
+
+  UIPresentationController *presentationController = _controller.presentationController;
+
+  if (presentationController && presentationController.containerView && !_backdropTapGestureRecognizer) {
+    _backdropTapGestureRecognizer = [[UITapGestureRecognizer alloc] initWithTarget:self
+                                                                            action:@selector(handleBackdropTap:)];
+    _backdropTapGestureRecognizer.delegate = self;
+    _backdropTapGestureRecognizer.cancelsTouchesInView = NO;
+    [presentationController.containerView addGestureRecognizer:_backdropTapGestureRecognizer];
+  }
```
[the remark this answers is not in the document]
t0maboro: I don't see the case in which Screen could be re-presented - we have disabled view recycling on that level, so if the component is dismissed, the view instance will be destroyed.

ios/RNSScreen.mm
```diff
@@ -1336,6 +1365,32 @@ - (void)safeAreaInsetsDidChange
...
+  if (gestureRecognizer == _backdropTapGestureRecognizer) {
+    UIPresentationController *presentationController = _controller.presentationController;
+
+    // Ignore any touches that land inside the actual sheet content.
+    if (presentationController && presentationController.presentedView &&
+        [touch.view isDescendantOfView:presentationController.presentedView]) {
+      return NO;
+    }
+    return YES;
+  }
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3771/commits/ee931def47650067548206056f8467eb5f4d12c4

ios/RNSScreen.mm
```diff
@@ -1336,6 +1365,38 @@ - (void)safeAreaInsetsDidChange
...
+  return YES;
```
kligarski, line 1390: this line in theory should never be called but maybe as a default leave it as `NO`?

ios/RNSScreen.mm
```diff
@@ -880,6 +883,32 @@ - (void)presentationControllerDidDismiss:(UIPresentationController *)presentatio
   }
 }
 
+- (void)setupBackdropTapGestureRecognizer
+{
+  if (self.stackPresentation != RNSScreenStackPresentationFormSheet) {
```
kligarski: What about e.g. `modal`/`pageSheet`/...?

https://github.com/user-attachments/assets/48e7b8cc-d922-4f7a-867c-b4efec3c4246

ios/RNSScreen.mm
t0maboro: if you're adding a recognizer to the delegate, you're rather expecting that it will receive touch, so for consistency with the logic above, where I'm filtering out only these cases, for which the recognizer shouldn't trigger, it'd prefer to keep the default as `YES`

ios/RNSScreen.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3771/commits/a196a6876fa29973d4988d25cd48384bda5a5ecf

kligarski: I'm not sure here but other types of modals don't need it as well?

t0maboro: other types don't have a backdrop, or are covering up the whole screen

Issue #2578, opened 2024-12-17
Multiple transitions between screens causes app to crash


marklawlor:
### Description

Repeated navigation between 2 screens using React Navigation's Native Stack with a `getId()` prop causes the app to "crash". I use the term crash loosely as the behaviour is different per platform but both result in the app becoming unusable.

## iOS

`react-native` elements that use Pressibility become stuck in their "onPress" state, preventing them from receiving further onPress events. The issue does not occur with non-Pressibility elements such as `<TouchableOpacity />`.

You can observe this behaviour with the reproduction snack by repeatedly moving between the two screens using the Forwards/Backwards `<Text />` items. After 3 transitions, the `<Text />` items are no longer pressable and will be rendered with their `onPress` highlighting.

## Android

After 3 transitions, the native stack header will disappear. After 4 transitions, the native stack will no longer render any visible context.

Expo Router issue: https://github.com/expo/expo/issues/33658
React Navigation issue: https://github.com/react-navigation/react-navigation/issues/12346

### Steps to reproduce

1. Open Snack
2. Click "forwards" / "backwards" 3-5 times

### Snack or a link to a repository

https://snack.expo.dev/@mwlawlor/react-navigation-native-stack-stuck

### Screens version

3.3.0

### React Native version

0.76.3

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

Mokhtar25:
Facing the same issue right now

mlukasik-dev:
My app with expo-router is affected by this problem.

kashmiry:
I am also facing this issue on **Android**. 
On iOS everything works fine.

When I navigate back to the same screen with the same `getId`, I observe the described behavior.
However, in my actual app, where I have a lot of integrations, this issue results in a crash with the following error:

```
The specified child already has a parent. You must call removeView() on the child's parent first
```

The temporary fix working for me now on Android is conditionally using `navigateDeprecated` if Platform is Android:
`navigation.navigateDeprecated(name, params);`

| Package                          | Version  |
|----------------------------------|----------|
| react                            | 18.2.0   |
| react-native                     | 0.74.6   |
| @react-navigation/bottom-tabs    | 7.2.0    |
| @react-navigation/native         | ^7.0.14  |
| @react-navigation/native-stack   | 7.2.0    |
| react-native-screens             | ^4.4.0   |
[...]

PR #3777, opened 2026-03-19
fix(android): guard header updates and finish removal transitions

l2hyunwoo:
## Description

Guards Android header rebuilds during screen removal and completes child transitions before the fragment root detaches. This targets a native view lifecycle condition that can produce `IllegalStateException: The specified child already has a parent` in `ScreenStackHeaderConfig.onUpdate()`.

The original description attributed this to ShadowTree reparenting without a verified navigation reproduction. The relevant distinction is that `ScreenStackHeaderConfigViewManager` exposes `configSubviews` as the config's children, while `onUpdate()` attaches those views to an Android `Toolbar`.

During a removal transition, Android can remove a toolbar child from its child array while retaining the child's parent reference until `endViewTransition()`. Rebuilding the toolbar in that state can fail when adding the child again. A second `removeView()` does not release a child already absent from the array.

**The Android condition is covered by isolated tests. The reported navigation crash has not yet reproduced in the local emulator example, so the integration fix remains unverified.**

## Changes

- Move the removal guard into `ScreenStackHeaderConfig.onUpdate()` and check ancestor screens as well as the immediate screen. This covers both direct updates and updates from `maybeUpdate()`.
- Record the parent/child pairs passed to `startViewTransition()` and finish them in reverse order, including children no longer in the current child array. Keep the removal guard active during cleanup and protect against reentrant completion.
- Complete those transitions before `ScreenStack.endViewTransition()` detaches the fragment root. Preserve `blockFrameworkTransitionFinalization` and the separate pending-removal/native-transition state from current `main`.

This replaces the original extra `removeView()` before `toolbar.addView()`. The cleanup also overlaps with the issue discussed in #3250.

## Validation

- Built full FabricExample debug APKs for the baseline (`73591ca`) and PR head (`397e621`), using RN 0.87.0, Fabric/Hermes and arm64-v8a.
- Passed Spotless/ktlint 1.3.0 for the three changed Kotlin files.
- Passed 28 isolated Android cases: 14 scenarios on each of API 35 and 36, using Robolectric 4.16.1 and AppCompat 1.7.1. The tested guard and transition methods match the merged source. The local harness exercises real Android view behavior with extracted methods; it does not run the complete Fabric/Fragment integration and is not included in this PR.

The isolated cases cover ancestor/direct header guards, removed children, cleanup before parent detachment, reentry, repeated completion, a pending removal mark before native transition start, and blocked framework finalization.

## Navigation reproduction status

I built a local example based on the [nested-stack report](https://github.com/software-mansion/react-native-screens/pull/3777#issuecomment-5532659303): three native stacks, custom LEFT (`Image`) and CENTER (`View`/`Text`) header subviews, and two system Back events injected in one device-side adb shell.

The standalone setup uses the reported RN 0.85.3, screens 4.25.2 and native-stack 7.16.0 versions on an Android 16/API 36 arm64 emulator with compile/target SDK 36. All animation scales are 1. With the default native-stack animation, device-side sleeps of 0.432, 0.482 and 0.582 seconds all returned to the middle stack without the reported exception. Sleep duration is not the measured Back callback interval.

Earlier recordings used `slide_from_right`, which has different animation timing; those are exploratory checks only. Successful navigation recordings do not establish that this PR fixes the crash. The reporter's deterministic test or standalone reproduction is still needed for a failing-before/passing-after comparison.

For local FabricExample comparisons, both APKs restore ReactActivity's standard Back callback; the current example host disables it for its own setup. That host adjustment is test setup and is not part of the library patch.

## Checklist

- [x] Compiled the native change and built comparison APKs.
- [x] Exercised the retained-parent condition in isolated Android tests.
- [ ] Verified the reported navigation crash before and after the patch.
- [ ] Added a verified in-repository regression test.
- [ ] Attached recordings demonstrating the reported failure and its fix.
- [ ] Ensured that CI passes.


kkafar (commented):
Hey! Excuse us late response.

What's the plan here? Is there any reproduction for the problem you describe?

I don't really see how the situation you describe could happen. Reparenting and react-native are rather mutually exclusive terms, at least on ShadowTree model layer. Not sure of all scenarios that could happen in HostTree, but proceeding here would require a proper reproduction of the issue.

andreilupu-plt:
Hi @l2hyunwoo, are there any updates on this? I also noticed a similar issue reported [here](https://github.com/software-mansion/react-native-screens/issues/2578).

fracico-tech:
```
Fatal Exception: java.lang.IllegalStateException: The specified child already has a parent. You must call removeView() on the child's parent first.
       at android.view.ViewGroup.addViewInner(ViewGroup.java:5327)
       at android.view.ViewGroup.addView(ViewGroup.java:5156)
       at android.view.ViewGroup.addView(ViewGroup.java:5096)
       at android.view.ViewGroup.addView(ViewGroup.java:5068)
       at com.swmansion.rnscreens.ScreenStackHeaderConfig.onUpdate(ScreenStackHeaderConfig.kt:341)
       at com.swmansion.rnscreens.ScreenStackHeaderConfigViewManager.onAfterUpdateTransaction(ScreenStackHeaderConfigViewManager.kt:91)
       at com.swmansion.rnscreens.ScreenStackHeaderConfigViewManager.onAfterUpdateTransaction(ScreenStackHeaderConfigViewManager.kt:22)
       at com.facebook.react.uimanager.ViewManager.updateProperties(ViewManager.java:104)
       at com.facebook.react.fabric.mounting.SurfaceMountingManager.updateProps(SurfaceMountingManager.kt:627)
       at com.facebook.react.fabric.mounting.SurfaceMountingManager.updateProps(SurfaceMountingManager.kt:597)
       at com.facebook.react.fabric.mounting.mountitems.IntBufferBatchMountItem.execute(IntBufferBatchMountItem.kt:127)
       at com.facebook.react.fabric.mounting.MountItemDispatcher.executeOrEnqueue(MountItemDispatcher.kt:340)
       at com.facebook.react.fabric.mounting.MountItemDispatcher.dispatchMountItems(MountItemDispatcher.kt:246)
       at com.facebook.react.fabric.mounting.MountItemDispatcher.tryDispatchMountItems(MountItemDispatcher.kt:93)
       at com.facebook.react.fabric.FabricUIManager$DispatchUIFrameCallback.doFrameGuarded(FabricUIManager.java:1528)
       at com.facebook.react.uimanager.GuardedFrameCallback.doFrame(GuardedFrameCallback.kt:42)
       at com.facebook.react.modules.core.ReactChoreographer.frameCallback$lambda$1(ReactChoreographer.kt:59)
       at com.facebook.react.modules.core.ReactChoreographer.$r8$lambda$nSkFhrr5T7rop_XKwzlLov4NLLw()
       at com.facebook.react.modules.core.ReactChoreographer$$ExternalSyntheticLambda0.doFrame(D8$$SyntheticClass)
       at android.view.Choreographer$CallbackRecord.run(Choreographer.java:1728)
       at android.view.Choreographer$CallbackRecord.run(Choreographer.java:1739)
       at android.view.Choreographer.doCallbacks(Choreographer.java:1255)
       at android.view.Choreographer.doFrame(Choreographer.java:1145)
       at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:1693)
       at android.os.Handler.handleCallback(Handler.java:959)
       at android.os.Handler.dispatchMessage(Handler.java:100)
       at android.os.Looper.loopOnce(Looper.java:249)
       at android.os.Looper.loop(Looper.java:337)
       at android.app.ActivityThread.main(ActivityThread.java:9593)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:593)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:936)
```

Gerd270597:
Hi @kkafar, @l2hyunwoo — we hit this crash in a production app. @kkafar asked
for a reproduction of the described scenario, so here is ours.

**I can reproduce it reliably and have it automated in a Detox test.** That also
means I can run that test against this PR's patch and report back whether it
actually fixes the crash. Just say the word and I'll do it — or I can put
together a standalone minimal repro repo if that's more useful than the
description below.

On the doubt raised earlier about reparenting: the reparenting here isn't done
by React, it's done by `ScreenStackHeaderConfig` itself. `onUpdate()` moves the
subviews into the toolbar with `toolbar.addView(view)` (line 368) while the view
manager reports `configSubviews` as the config's children to Fabric
(`getChildCount(parent) = parent.configSubviewsCount`). So the Android parent and
the parent Fabric believes in are different objects by design, and the stack
trace below confirms it: the offending child's parent is the toolbar.

## Reproduction

The crash is deterministic given the right timing.

### Environment

| | |
|---|---|
| `react-native-screens` | 4.25.2 |
| `react-native` | 0.85.3 (New Architecture / Fabric, Hermes) |
| `@react-navigation/native-stack` | 7.16.0 |
| Device | Physical, Android 16 (API 36) |
| `compileSdk` / `targetSdk` | 36 |

### Stack trace

```
java.lang.IllegalStateException: Cannot remove child at index 1 from parent ViewGroup [31572],
only 1 children in parent. Warning: childCount may be incorrect!
    at com.facebook.react.fabric.mounting.SurfaceMountingManager.removeViewAt(SurfaceMountingManager.kt:477)
    at com.facebook.react.fabric.mounting.mountitems.IntBufferBatchMountItem.execute(IntBufferBatchMountItem.kt:125)
    at com.facebook.react.fabric.mounting.MountItemDispatcher.executeOrEnqueue(MountItemDispatcher.kt:340)
    at com.facebook.react.fabric.mounting.MountItemDispatcher.dispatchMountItems(MountItemDispatcher.kt:246)
    at com.facebook.react.fabric.mounting.MountItemDispatcher.tryDispatchMountItems(MountItemDispatcher.kt:93)
    at com.facebook.react.fabric.FabricUIManager$DispatchUIFrameCallback.doFrameGuarded(FabricUIManager.java:1528)
    ...
Caused by: java.lang.IllegalStateException: The specified child already has a parent.
You must call removeView() on the child's parent first.
parent=com.swmansion.rnscreens.ScreenStackHeaderConfig$DebugMenuToolbar{13562b9 V.E...... ......ID 0,0-1080,256}
    at android.view.ViewGroup.addViewInner(ViewGroup.java:5301)
    at android.view.ViewGroup.addView(ViewGroup.java:5127)
    at com.swmansion.rnscreens.ScreenStackHeaderConfig.onUpdate(ScreenStackHeaderConfig.kt:368)
    at com.swmansion.rnscreens.ScreenStackHeaderConfig.maybeUpdate(ScreenStackHeaderConfig.kt:377)
    at com.swmansion.rnscreens.ScreenStackHeaderConfig.removeConfigSubview(ScreenStackHeaderConfig.kt:388)
    at com.swmansion.rnscreens.ScreenStackHeaderConfigViewManager.removeViewAt(ScreenStackHeaderConfigViewManager.kt:68)
    at com.facebook.react.fabric.mounting.SurfaceMountingManager.removeViewAt(SurfaceMountingManager.kt:456)
```

`removeConfigSubview` calls `maybeUpdate()` synchronously, re-entrantly, from
inside a Fabric mount transaction. `onUpdate()` then rebuilds the whole toolbar,
and the `toolbar.addView(view)` at line 368 hits a subview that is still
attached to a toolbar.

### View hierarchy at the point of failure

Logged by `SurfaceMountingManager`. The failing config sits in a stack nested
three levels deep:

```
ScreenStackHeaderConfig (31572)   <- crashes
 └ Screen (31574)
   └ ScreenStack (31576)
     └ ScreenContentWrapper (31580)
       └ Screen (31584)
         └ ScreenStack (246)
           └ ScreenContentWrapper (250)
             └ Screen (254)
               └ ScreenStack (256)
```

### Setup

Three nested native stacks. The innermost one sets, in `screenOptions`, both a
custom `headerLeft` (an `<Image>`) and a custom `headerTitle` (a component
rendering `<View><Text/></View>`), so every screen's header config carries two
subviews (LEFT + CENTER):

```jsx
<Stack.Navigator
  screenOptions={{
    headerTitle: () => <CustomHeader />,   // -> CENTER subview
    headerLeft: () => <Logo />,            // -> LEFT subview
    headerTitleAlign: 'center',
  }}>
  <Stack.Screen name="Menu" component={MenuScreen} />
  <Stack.Screen name="Form" component={FormScreen} />
  {/* ~9 more screens */}
</Stack.Navigator>
```

### Steps

1. Navigate into the innermost stack and let everything settle (no pending
   async work — I verified the crash happens with all data loaded).
2. From the menu screen, open any child screen.
3. Press the system back button **twice**, with roughly **450 ms** between the
   two presses.

The second press pops out of the innermost stack, and the crash lands
~380 ms later.

### Timing

The interval is the whole trick, and it is narrow. Chaining presses as fast as
possible does **not** reproduce: the first pop has to get far enough along
before the second arrives. I measured the dispatch times from
`WindowOnBackDispatcher` in logcat across three independent runs:

| Method | Interval between presses | Crash after 2nd press |
|---|---|---|
| Real finger on the nav bar | 462 ms | 398 ms |
| `adb shell input keyevent 4` | 448 ms | 369 ms |
| Automated (Detox) | 445 ms | 390 ms |

A third press is never involved — in every run it dispatched after the crash
had already happened.

Two measurement notes that cost me time and might save yours:

- Measuring at the kernel level with `getevent` is misleading. It reports
  ~392 ms because it timestamps when the finger goes **down**, but back is
  dispatched on **release**. The `WindowOnBackDispatcher` timestamps are the
  ones that matter.
- Detox's `device.pressBack()` does **not** reproduce this, at any interval,
  even with synchronization disabled — it goes through Espresso/UiAutomator,
  which waits for idle internally, so the requested interval is not the one
  that reaches the device. Injecting via adb works:

```js
execSync('adb shell "input keyevent 4; sleep 0.392; input keyevent 4"');
```

(The `sleep` has to run device-side, in a single invocation. The extra ~56 ms
over the nominal 0.392 is the JVM startup of the `input` command.)

Also worth flagging for anyone trying to reproduce this in CI: on an emulator,
Detox zeroes the animation scales via `EmulatorAllocDriver`, transitions become
instantaneous, and the race cannot occur at all. It only reproduces on a
physical device.

### Relation to #3249

This looks like the header-side manifestation of the view recycling problem in
#3249. Relevant here, the per-component recycling flags are on by default in
RN 0.85.3 (`enableViewRecyclingForImage`, `enableViewRecyclingForText` and
`enableViewRecyclingForView` are all `true`, while the master
`enableViewRecycling` is `false`) — and the header subviews in this setup are
exactly an `Image` and a `View`/`Text`.

I have not yet verified whether disabling those flags avoids the crash; I can
run that if it would be useful.

Issue #3747, opened 2026-03-10
(native-tab-navigator/experimental) - weird/unwanted bottom navigator animation on back when using deeplinks


spasmodiumspam-web:
### Description

Repository provided.

I have a Root Stack navigator that contains:
- a Tab navigator (3 tabs), and
- a **Details** screen at the same stack level.

When I navigate manually from Tab 3 to **Details**, everything works as expected.  
Pressing back returns to Tab 3 with normal animation for the screen (so no specific animation for the tab navigator itself).  

https://github.com/user-attachments/assets/35fdbb3b-af8d-413d-88fe-a8ba54269be1

However, when I open **Details** via deep link, pressing back (which should return to Tab 3) shows an unexpected/weird back animation.  

https://github.com/user-attachments/assets/64458b81-0250-4697-8e99-357726093e42

Use command

`adb shell am start -W -a android.intent.action.VIEW -d "buggednav://tab3/details" com.buggednav
`

**Notice the weird/unwanted animation on the tab navigator.**

**This doesn't happen with the JS/non-native tab navigator in my tests.**

### Current behavior

**Expected**: same back behavior/animation as manual navigation (Tab 3 -> Details -> back to Tab 3).  
**Actual**: back transition animation is visually incorrect only in the deep-link flow.

### Expected behavior

No weird animation should appear, basically it should behave like we'd just go normally back from details to 3rd tab.

### Reproduction

https://github.com/spasmodiumspam-web/react-navigation-native-tabs-deeplink-reset-repro

### Platform

- [x] Android
- [ ] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [x] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [x] @react-navigation/native-stack
- [ ] react-native-drawer-layout
- [ ] react-native-tab-view

### Environment

- [x] I've removed the packages that I don't use

| package                             | version                     |
| ----------------------------------- | --------------------------- |
| @react-navigation/native            | 7.1.33                      |
| @react-navigation/bottom-tabs       | 7.15.5                      |
| @react-navigation/native-stack      | 7.14.4                      |
| react-native-screens                | 4.24.0                      |
| react-native-safe-area-context      | 5.7.0                       |
| react-native-gesture-handler        | 2.30.0                      |
| react-native                        | 0.84.1                      |
[...]

PR #3793, opened 2026-03-24
fix(Android, Stack v4): Add heuristic for applying inset only on top-level header

t0maboro:
## Description

This PR fixes multiple applications of top insets when stack navigators are nested, leading to excessive whitespace. We introduced a `TopInsetConsumptionContext` to coordinate top inset consumption across nested navigators. The topmost visible, non-hidden header consumes the inset and notifies nested screens that the inset has already been handled.
Additionally, we implemented a pre-layout inset handling mechanism using `DecorView` to prevent layout jumps before the window insets are dispatched - this is an updated approach relative to: https://github.com/software-mansion/react-native-screens/pull/3442. Additionally, all workarounds that were related to manual frame corrections from that PR were removed, because now we can apply proper layout to the Toolbar earlier, and we don't need to make any manual corrections before the insets come.
We're adding an option to opt out of this behavior with an experimental feature flag `androidLegacyTopInsetBehavior`.

## Changes

- added `TopInsetConsumptionContext` to track whether the inset was already consumed higher up in the hierarchy
- `ScreenStackItem` now calculates whether its nested children should consider the top inset consumed
- `ScreenStackHeaderConfig` reads the context and determines if it should consume the inset.
- `androidLegacyTopInsetBehavior` was added to allow opt-out from this approach
- added a logic to apply pre-layout padding, reading insets from DecorView - with additional logic to clean up applied paddings and insets when the component was detached 
- added new native props
- removed old workarounds related to manual layout calculations on Screen and AppBarLayout level

Test3793

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/2cc3b26a-3768-477b-a040-dff42c3b6e26" /> | <video src="https://github.com/user-attachments/assets/cbac878c-83b0-440a-86f7-9d77104a96b0" /> |

---

Test3006
Note: issues with jumping content in Tabs->Stack(with header)->Stack(with header) combo are still present, don't see an option to resolve that now, because the `DummyLayoutHelper` would need to be aware which header is top-level, for now I'm treating that specific setup as uncommon.

| Before | After |
| --- | --- |
| (shouldn't change) | <video src="https://github.com/user-attachments/assets/186014e0-da9a-4626-bcee-aad9720c9127" /> |

---

## Test plan

Added `Test3793`, performed regression testing on `Test3006`.

## Checklist

- [x] Ensured that CI passes


android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
```diff
@@ -139,6 +158,13 @@ open class CustomToolbar(
                 0,
             )
 
+        val shouldHandleTopInset = if (config.legacyTopInsetBehavior) true else config.consumeTopInset
+
+        if (!shouldHandleTopInset) {
+            clearAppliedInsetsAndPadding()
+            return unhandledInsets
+        }
```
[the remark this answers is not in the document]
t0maboro: following the assumptions from this PR description, only 1 header should handle inset and this one specific header should always handle display cutout

apps/src/tests/issue-tests/Test3793.tsx
```diff
@@ -0,0 +1,247 @@
+import React, { createContext, useContext, useState } from 'react';
+import { View, Text, StyleSheet, Button, Switch } from 'react-native';
+import { NavigationContainer } from '@react-navigation/native';
+import { createNativeStackNavigator } from '@react-navigation/native-stack';
+import Colors from '../../shared/styling/Colors';
+
+const HeaderConfigContext = createContext<any>(null);
+
+const HeaderConfigProvider = ({ children }: { children: React.ReactNode }) => {
+  const [outerShown, setOuterShown] = useState(true);
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3793/commits/fd47bb3fda15a9c8a00cdda7016ca0badfa20acf

src/components/ScreenStackItem.tsx
```diff
@@ -51,6 +52,22 @@ function ScreenStackItem(
...
+  const consumesTopInset = useLegacyBehavior
+    ? true
+    : !insetContext.isTopInsetConsumed && hasVisibleHeader;
+
+  const nextContextValue = React.useMemo(
+    () => ({
+      isTopInsetConsumed: insetContext.isTopInsetConsumed || consumesTopInset,
+    }),
+    [insetContext.isTopInsetConsumed, consumesTopInset],
+  );
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3793/commits/5ac64ddf047d291779aec4011cf770cfba16b525

phazei:
This is really needed, currently the inset on the header intermittently disappears, and is gone in landscape but back in portrait, there's no consistency so there's no way to manually set anything when it's not there.  This seems like it should make all the cases less fragile.

android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
kligarski, line 166: It's not super important but maybe we should avoid calling `clearAppliedInsetsAndPadding` on every `onApplyWindowInsets` call? I thought about moving this to the top of the function and calling `clearAppliedInsetsAndPadding` only when `insetsAppliedFromListener && !shouldHandleTopInset` but we might have an edge case where we apply insets from decor view and then we change `consumeTopInset` to false so we would need separate property for whether any insets were applied.

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.kt
```diff
@@ -34,6 +34,23 @@ class ScreenStackHeaderConfig(
...
+            }
+        }
+
+    var legacyTopInsetBehavior = false
+        set(value) {
+            if (field != value) {
+                field = value
+                if (isAttachedToWindow) toolbar.requestApplyInsets()
+            }
+        }
```
kligarski: We can also use `by Delegates.observable` but it's not important.

kligarski (approved):

android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3793/commits/bcff18806aef9070a0243a987617434406cf8a34 - tbh I disliked this method after reading this comment, because it makes 2 totally different things: resetting the internal state of Toolbar and manipulating view's frame, therefore I decided to split these and add extra check to reset padding only if it had  non-zero values - let me know wdyt think about this shape, the drawback is that I'm repeating twice
```
resetInsetsState()
clearPaddingIfNeeded()
```

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3793/changes/f0eaaeb61d4eed56cfc8b6ff0641267823d1cb9f

android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
```diff
@@ -113,7 +104,36 @@ open class CustomToolbar(
         }
     }
 
+    override fun onAttachedToWindow() {
+        super.onAttachedToWindow()
+
+        applyDecorViewTopInset()
```
kligarski: Can `onAttachedToWindow` run again after first render, e.g. when we hide and show the header? If so, maybe we should call the metod `applyDecorViewTopInsetIfNeeded`/`maybeApplyDecorViewTopInset` or something like that to highlight that it might return early.

android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
```diff
@@ -113,7 +104,36 @@ open class CustomToolbar(
...
+
+    private fun applyDecorViewTopInset() {
+        if (config.legacyTopInsetBehavior || !config.consumeTopInset) return
+
+        val activity = (context as? ThemedReactContext)?.currentActivity ?: return
+        val decorView = activity.window.decorView
+        val topInset = getDecorViewTopInset(decorView)
+
+        if (topInset > 0) {
+            if (!insetsAppliedFromListener) {
```
kligarski: Can we move this to first early return?

android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
kligarski: I think it's okay now.

src/components/contexts/TopInsetConsumptionContext.ts
```diff
@@ -0,0 +1,30 @@
...
+export const TopInsetConsumptionContext = React.createContext({
+  isTopInsetConsumed: false,
+});
+
+export function useTopInsetConsumption(isHeaderHidden: boolean | undefined) {
+  const insetContext = React.useContext(TopInsetConsumptionContext);
+  const useLegacyBehavior =
+    featureFlags.experiment?.androidLegacyTopInsetBehavior ?? false;
+
+  const hasVisibleHeader = isHeaderHidden !== true;
```
kligarski: I guess that we will need to adapt https://github.com/software-mansion/react-native-screens/pull/3405 to also use this context. We could change the name of the `isHeaderHidden` argument to be more general (`isHeaderHidden` -> `isTopInsetConsumed` but with negation I guess). We can also do this later when needed in https://github.com/software-mansion/react-native-screens/pull/3405.

src/components/contexts/TopInsetConsumptionContext.ts
```diff
@@ -0,0 +1,30 @@
...
+  const consumesTopInset = useLegacyBehavior
+    ? true
+    : !insetContext.isTopInsetConsumed && hasVisibleHeader;
+
+  const nextContextValue = React.useMemo(
+    () => ({
+      isTopInsetConsumed: insetContext.isTopInsetConsumed || consumesTopInset,
```
kligarski, line 19: Please see `BottomTabsAndStack` example. The inset is repeated due to JS header in tabs.

Image

This seems like a good example where we should allow the user to disable the top inset maunally, right? This functionality is missing in this PR but I think that we discussed it. Is this going to be handled in separate PR?

But if we're going to add it, we need to think about propagating that information.

In regular case, if the outer header is hidden, then the inner header should handle the inset as is right now in this PR.

Another situation: imagine if we had 2 native stacks in `BottomTabsAndStack`. If the user manually disables outer native header inset, then the inner header shouldn't apply the padding as well, right? The user override indicates that we applied inset where we shouldn't and any header below it shouldn't apply the inset as well? On the other hand, setting the same prop to all stacks isn't that hard and it would simplify our approach.

Just something to think about.

android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3793/changes/4be7bbb0f01eb42ff17c486cf430f27545d07998

android/src/main/java/com/swmansion/rnscreens/CustomToolbar.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3793/changes/9f73e15c0da05ad580ef699158c83c161034803e

src/components/contexts/TopInsetConsumptionContext.ts
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3793/changes/52f93e65655ae40cfbaa0275c334b979e3ba2cfe

src/components/contexts/TopInsetConsumptionContext.ts
t0maboro: let's make that in the follow-up PR, ticket for that: https://github.com/software-mansion/react-native-screens-labs/issues/1096

I think that just exposing a prop is fine in that case, because the prop will be living per screen instance, so the downstream might propagate that information as well

kkafar: I think that initially - especially in stack v4, we should settle with allowing for manual control. 

Describe somewhere how does the heuristic work and when it needs assistance (cases like described above). 

Let's keep as simple as possible for stack v4.

t0maboro: will be followed-up in: https://github.com/software-mansion/react-native-screens/pull/3835

kligarski (commented):
Is this a regression in this PR or does it happen on main as well (look at the navigation bar inset for bottom navigation view)?

https://github.com/user-attachments/assets/93fa60f5-02eb-4631-8fe9-fc4856bbe7e8

t0maboro:
> Is this a regression in this PR or does it happen on main as well (look at the navigation bar inset for bottom navigation view)?
> 
>  Screen_recording_20260407_083008.mp4

Known issue with delayed insets when native tabs are nested in stack: https://github.com/software-mansion/react-native-screens/issues/3747#issuecomment-4080987291

phazei:
Any reason this hasn't been merged yet?

phazei:
Oh, another use case that needs accounting for, likely an option for manual control of disabling the header top inset entirely:

If the user wants a top banner: Right now when adding a <View> or <SafeAreaView> at the very top of the app, since the header always applies the margin no matter what, that extra margin is added under the top banner.

Has this been tested in landscape?  Currently the header doesn't add any inset in landscape mode, and only does in portrait.

Also, wouldn't have never had adding it to the header at all have been the easiest, then it's left to the dev to handle it and it could be managed, having it auto added is what made it unmanagable.

t0maboro:
> Oh, another use case that needs accounting for, likely an option for manual control of disabling the header top inset entirely:
> 
> If the user wants a top banner: Right now when adding a or at the very top of the app, since the header always applies the margin no matter what, that extra margin is added under the top banner.

will be realized in a follow-up PR: https://github.com/software-mansion/react-native-screens/pull/3835

> Has this been tested in landscape? Currently the header doesn't add any inset in landscape mode, and only does in portrait.

landscape:

without content overflow: Screenshot 2026-04-09 at 09 02 49

with content overflow: Screenshot 2026-04-09 at 09 02 23

isaacfilarski:
Hi, when is this going to be released?

t0maboro:
We still have a few important changes to merge for the native tabs, native stack memory leak fixes and a follow-up PR for the header changes before we release 4.25. If this update is important for you right now, I recommend using the nightly release.

isaacfilarski:
> We still have a few important changes to merge for the native tabs, native stack memory leak fixes and a follow-up PR for the header changes before we release 4.25. If this update is important for you right now, I recommend using the nightly release.

Thank you, could you tell me which nightly release has this specific change? I see that the last one was 8 days ago.

t0maboro:
Sorry, we didn't notice that the nightly releases started failing last week - we're fixing it now https://github.com/software-mansion/react-native-screens/pull/3878

PR #3794, opened 2026-03-24
feat(Tabs): update operation model, add `TabSelectionRejectedEvent` and support rejecting stale state updates

kkafar:
## Description

This PR completes refactor of architecture as described in [RFC-1028](https://github.com/software-mansion/react-native-screens-labs/blob/main-issue-tracker/rfcs/1028-tabs-logical-architecture.md).

I add here update rejection mechanism for stale navigation state updates. This
PR handles both platforms.

The PR adds mostly two things:

1. `rejectStaleNavStateUpdates` boolean prop on `TabsHost`,
2. `onTabSelectionRejected` callback on `TabsHost`.

The prop is a kill-switch for the update rejection mechanism.
By default, it is set to `false` (the mechanism is disabled).

When enabled, the native container will reject any *stale state update* i.e.,

If a state update operation is rejected, the host will emit `TabSelectionRejectedEvent`
triggering `onTabSelectionRejected` callback with appropriate event data.

The `TabSelectionRejectedEvent` can be emitted in a couple of more cases, beside
the *stale state update rejection mechanism*:

1. When JS tries to select a tab that is already selected.
2. On iOs, when JS tries to select the `moreNavigationController` while it is not present in the UI.

### Stale state definition

Initially I've defined the *stale state* as follows:

> any state update with provenance number less than or equal to the provenance
> number of native state at the moment of JS state update operation execution.

However, @kligarski pointed out some [valid concerns](https://github.com/software-mansion/react-native-screens/pull/3794#discussion_r3016592537).
Based on that, I decided that the operation model needs to be **slightly** updated.

What this PR changes is also the interpretation of the `TabsHostNavState#provenance`.
Before it denoted "the next state provenance as believed by JS side".
Now, it means "last confirmed / acknowledged state that the JS side is aware of
at the time of requesting the state update".

The change in semantics of `TabsHostNavState#provenance` described above also impacts
definition of what is a *stale state*. Now, a state update is *stale*, if it has
been based of an *invalidated* state. The previously selected state is *invalidated*,
when after it a new state *of different origin* is selected.

To better understand this, let me give you an example. Let's say that all state
updates named $A_{i, j}$ are JS-triggered states, and $B_{i, j}$ are states that are results
of direct user action / programmatic push on UI thread. $i$ denotes desired tab index, $j$
denotes the update provenance. $A_{i, j}$ and $B_{i, j}$ have different origins.
Let's by $S_{i, j}$ denote a state with selected tab $i$ and provenance $j$.

Consider following sequence of state states, where all the states (tab changes) have been
requested in quick succession with the same *provenance* of e.g. 1

> $A_{0, 1}$, $A_{1, 1}$, $A_{2, 1}$

As a result of these updates, we'd observe three distinct tab states being applied to `TabsContainer`:

> $S_{0, 1}$, $S_{1, 2}$, $S_{2, 3}$

Notice that the native container state provenance increments monotonically.

Now, if we requested yet another update to let's say state $A_{1, 1}$ with `provenance == 1`,
it would be performed and, we'd end up with following series updates being applied:

> $A_{0, 1}$, $A_{1, 1}$, $A_{2, 1}$, $A_{1, 1}$

With final state of $S_{1, 4}$.

But let's say, that before the $A_{1, 1}$ is processed on UI thread,
user action triggers state progression to $B_{0, 5}$:

> $A_{0, 1}$, $A_{1, 1}$, $A_{2, 1}$, $B_{0, 5}$

The state is progressed to $S_{0, 5}$ and now the previously scheduled state update
$A_{1, 1}$ is processed. It will be rejected (if `rejectStaleNavStateUpdates === true`),
because a new state of different *origin* had been committed before it has been
processed. Now, JS side will be notified with `onTabSelectionRejected` event, indicating
the most-up-to-date-state. To navigate successfully to tab 1, the JS will have to send
$A_{1, 5}$ to acknowledge the "state of different origin".

## Visual documentation

Observe that I first:

1. Navigate to the third tab & enable the "heavy render" for that tab.
2. Navigate back to the first tab.
3. Navigate via JS to third tab, but select second tab via UI.
4. Tab changes to second, and "stale" navigation to third gets rejected.
5. I then toggle the `rejectStaleNavStateUpdates` option to false (default) and repeat the experiment.
6. This time, the navigation to third tab is not rejected.

<https://github.com/user-attachments/assets/12ee1d44-ebd1-402d-886e-2c8cf17a0286>

## Test plan

`test-tabs-stale-update-rejection`

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsNavState.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsNavState.kt
@@ -1,6 +1,6 @@
 package com.swmansion.rnscreens.gamma.tabs.container
 
-import com.swmansion.rnscreens.gamma.tabs.container.TabsNavStateUpdateRejectionReason.MORE_TAB_NOT_AVAILABLE
+import com.swmansion.rnscreens.gamma.tabs.container.TabsNavStateUpdateRejectionReason.MORE_NAV_CTRL_NOT_AVAILABLE
 import com.swmansion.rnscreens.gamma.tabs.container.TabsNavStateUpdateRejectionReason.REPEATED
 import com.swmansion.rnscreens.gamma.tabs.container.TabsNavStateUpdateRejectionReason.STALE
 
@@ -30,19 +30,19 @@ data class TabsNavState(
  *
  * - [STALE] — the update's provenance indicates that it is based on a stale state.
  * - [REPEATED] — the requested tab is already selected.
- * - [MORE_TAB_NOT_AVAILABLE] — the iOS "More" navigation controller was requested but is not available.
+ * - [MORE_NAV_CTRL_NOT_AVAILABLE] — the iOS "More" navigation controller was requested but is not available.
  *   This should not happen on Android.
  */
 enum class TabsNavStateUpdateRejectionReason {
     STALE,
     REPEATED,
-    MORE_TAB_NOT_AVAILABLE,
+    MORE_NAV_CTRL_NOT_AVAILABLE,
     ;
 
     override fun toString(): String =
         when (this) {
             STALE -> "stale"
             REPEATED -> "repeated"
-            MORE_TAB_NOT_AVAILABLE -> "more-tab-not-available"
+            MORE_NAV_CTRL_NOT_AVAILABLE -> "more-nav-ctrl-not-available"
         }
 }
--- a/ios/tabs/host/RNSTabBarController.mm
+++ b/ios/tabs/host/RNSTabBarController.mm
@@ -375,7 +375,7 @@ static NSString *const kMoreNavigationControllerScreenKey = @"rnscreens_moreNavi
       [self.tabsHostComponentView tabBarController:self
                              rejectedStateUpdateTo:_pendingOperation
                                       currentState:_navigationState
-                                        withReason:RNSTabsNavigationStateRejectionReasonMoreTabNotAvailable];
+                                        withReason:RNSTabsNavigationStateRejectionReasonMoreNavCtrlNotAvailable];
       return;
     }
     nextSelectedViewController = [self resolveMoreNavigationController];
--- a/src/components/tabs/host/TabsHost.types.ts
+++ b/src/components/tabs/host/TabsHost.types.ts
@@ -74,13 +74,13 @@ export type TabSelectedEvent = {
  *   meaning a newer state has already been applied. Only reported when
  *   {@link TabsHostPropsBase#rejectStaleNavStateUpdates} is enabled.
  * - `repeated` — the requested tab is already selected.
- * - `more-tab-not-available` — the iOS "More" navigation controller was requested
+ * - `more-nav-ctrl-not-available` — the iOS "More" navigation controller was requested
  *   but is not available in the current configuration.
  */
 export type TabSelectionRejectionReason =
   | 'stale'
   | 'repeated'
-  | 'more-tab-not-available';
+  | 'more-nav-ctrl-not-available';
 
 /**
  * @summary Payload of the event emitted when the native side rejects a tab selection request.
--- a/ios/conversion/RNSConversions-Tabs.mm
+++ b/ios/conversion/RNSConversions-Tabs.mm
@@ -228,8 +228,8 @@ RNSOnTabSelectionRejectedRejectionReasonFromRNSTabsNavigationStateRejectionReaso
       return Stale;
     case RNSTabsNavigationStateRejectionReasonRepeated:
       return Repeated;
-    case RNSTabsNavigationStateRejectionReasonMoreTabNotAvailable:
-      return MoreTabNotAvailable;
+    case RNSTabsNavigationStateRejectionReasonMoreNavCtrlNotAvailable:
+      return MoreNavCtrlNotAvailable;
     default:
       return Stale;
   }
--- a/ios/tabs/host/RNSTabsNavigationState.h
+++ b/ios/tabs/host/RNSTabsNavigationState.h
@@ -62,7 +62,7 @@ typedef NS_ENUM(NSInteger, RNSTabsNavigationStateRejectionReason) {
   /** The requested tab is already selected. */
   RNSTabsNavigationStateRejectionReasonRepeated,
   /** The iOS "More" navigation controller was requested but is not available. */
-  RNSTabsNavigationStateRejectionReasonMoreTabNotAvailable
+  RNSTabsNavigationStateRejectionReasonMoreNavCtrlNotAvailable
 };
 
 NS_ASSUME_NONNULL_END
--- a/src/fabric/tabs/TabsHostAndroidNativeComponent.ts
+++ b/src/fabric/tabs/TabsHostAndroidNativeComponent.ts
@@ -23,7 +23,7 @@ type TabSelectionRejectedEvent = Readonly<{
   provenance: CT.Int32;
   rejectedScreenKey: string;
   rejectedProvenance: CT.Int32;
-  rejectionReason: 'stale' | 'repeated' | 'more-tab-not-available';
+  rejectionReason: 'stale' | 'repeated' | 'more-nav-ctrl-not-available';
 }>;
 
 type TabsHostColorScheme = 'inherit' | 'light' | 'dark';
--- a/src/fabric/tabs/TabsHostIOSNativeComponent.ts
+++ b/src/fabric/tabs/TabsHostIOSNativeComponent.ts
@@ -23,7 +23,7 @@ type TabSelectionRejectedEvent = Readonly<{
   provenance: CT.Int32;
   rejectedScreenKey: string;
   rejectedProvenance: CT.Int32;
-  rejectionReason: 'stale' | 'repeated' | 'more-tab-not-available';
+  rejectionReason: 'stale' | 'repeated' | 'more-nav-ctrl-not-available';
 }>;
 
 type TabsHostColorScheme = 'inherit' | 'light' | 'dark';


android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
```diff
@@ -259,6 +260,12 @@ internal class TabsContainer(
                 "[RNScreens] Failed to find Menu Item for screenKey: ${tabSelectOp.navState.selectedKey}"
             }
 
+        if (rejectOpsWithStaleNavState && isNavStateStale(tabSelectOp.navState)) {
+            delegate.onNavStateUpdateRejected(navState, tabSelectOp.navState)
+            pendingJsOperation = null
+            return
+        }
+
```
kkafar, line 275: Not really sure whether we need this boolean (prop). We could achieve the same behaviour with sending `provenance == Int.MAX` - such update would never be treated as stale.

Waiting for your opinions here.

kligarski: Maybe such prop will help downstream implementations allow customizing rejection behavior by users without directly changing internal state implementation (changing provenance number sent)?

I think that intention behind the rejection concept (and that we support enabling it or not) is clearer with dedicated property rather than using large provenance number (but if well documented, this is a viable approach as well).

kligarski: And also, as discussed internally, we should think what should happen when you select multiple tabs in succession from JS and before they reach native, we switch tab by tapping it. First one will be rejected but later operations will finish correctly.

kkafar: Yeah, I think that is a great catch. Good news is I believe I know how to patch it. Bad news is it requires a bit of model change.

I'll update the RFC, but the tldr is:

1. We change the meaning of the *provenance* sent by the JS to the native. Instead of it being *JS provenance* like it is right now, it should denote *last received **confirmed** provenance from the native*. 

Now, native container can reliably detect stale updates - last confirmed provenance is older than **last UI-side update**. Please note, it's not *last update*, but **last UI-side** update - this will allow for multiple JS navigations based on the same provenance.

kkafar: I'll first implement it in a separate PR, based on this one, and then open a PR to update the RFC.

ios/tabs/host/RNSTabsHostComponentView.mm
```diff
@@ -609,6 +615,17 @@ - (void)tabBarController:(nonnull RNSTabBarController *)tabBarController
                                              .isNativeAction = context.isNativeAction}];
 }
 
+- (void)tabBarController:(nonnull RNSTabBarController *)tabBarController
+    rejectedStateUpdateTo:(nonnull RNSTabsNavigationState *)rejectedNavState
+             currentState:(nonnull RNSTabsNavigationState *)currentNavState
+{
+  RCTAssert(currentNavState.selectedScreenKey != nil, @"[RNScreens] Current state screenKey MUST NOT be nil");
```
kligarski, line 623: Is `SCREEN_KEY_MORE_NAV_CTRL` allowed to be default route name? In our `TabsContainer` it's not possible due to check in `determineInitialTabsContainerState` but is it prohibited in general?
- If so, we should probably add this to `TabsHost.navState` docs and our `TabsContainer.defaultRouteName` docs.
- If not, then this assertion will crash the app when you try to use more navigation controller as initial route and there are < 5 tabs (at least it crashed when I removed the check from `determineInitialTabsContainerState`).

This is a super niche edge case but I guess that we can determine whether we support it or not. It might be also useful to have this thought out for when we start implementing prevent default.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsNavState.kt
```diff
@@ -13,3 +13,19 @@ data class TabsNavState(
         val EMPTY = TabsNavState("", Int.MIN_VALUE)
     }
 }
+
+enum class TabsNavStateUpdateRejectionReason {
+    OTHER,
```
kligarski: Does `OTHER` option make sense? If in the future there is a new rejection reason, won't we just add it to the enum instead of using `OTHER`?

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsNavState.kt
```diff
@@ -13,3 +13,19 @@ data class TabsNavState(
...
+    STALE,
+    REPEATED,
+    MORE_TAB_NOT_AVAILABLE;
```
kligarski: nit: I'm not sure about calling this `MORE_TAB_...`, maybe just `MORE` or `MORE_NAV_CTRL` as in `SCREEN_KEY_MORE_NAV_CTRL`?

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -24,6 +24,12 @@ @implementation RNSTabBarController {
 
   /// This property is nullable until first container udpate. Later it MUST NOT be nil.
   RNSTabsNavigationState *_Nullable _navigationState;
+
+  /// Holds last state that has been a result of UI-side navigation (user request).
+  ///
+  /// This property is nullable until first container udpate. Later it MUST NOT be nil.
```
kligarski:
```suggestion
  /// This property is nullable until first container update. Later it MUST NOT be nil.
```

kligarski: Same for comment above for `_navigationState`.

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -358,10 +368,11 @@ - (void)updateSelectedViewControllerInner
 
   if ([self isMoreNavigationControllerRequestedByOperation:_pendingOperation]) {
     if (![self isMoreNavigationControllerPresentInTabBar]) {
-      // If the controller is not visible atm. we'll crash the app.
+      // If the controller is not visible atm. we'll crash the app if we try to navigate to it.
```
kligarski, line 371:
```suggestion
      // If the controller is not visible atm, we'll crash the app if we try to navigate to it.
```

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -522,25 +543,31 @@ - (nonnull NSString *)screenKeyForSelectedViewController
...
+  if (_navigationState == nil || _lastUINavigationState == nil) {
     return NO;
-  } else {
-    return newState.provenance <= _navigationState.provenance;
   }
+
+  return newState.provenance < _lastUINavigationState.provenance;
 }
 
 #pragma mark-- More Navigation Controller
```
kligarski, line 576:
```suggestion
#pragma mark - More Navigation Controller
```

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsNavState.kt
```diff
@@ -13,3 +21,28 @@ data class TabsNavState(
...
 }
+
+/**
+ * Reason why a navigation state update was rejected by the container.
+ *
+ * - [OTHER] — unspecified reason.
+ * - [STALE] — the update's provenance indicates that is is based on a stale state.
+ * - [REPEATED] — the requested tab is already selected.
+ * - [MORE_TAB_NOT_AVAILABLE] — the iOS "More" navigation controller was requested but is not available.
+ *  This should not happen on Android.
```
kligarski:
```suggestion
 *   This should not happen on Android.
```

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsNavState.kt
```diff
@@ -13,3 +21,28 @@ data class TabsNavState(
         val EMPTY = TabsNavState("", Int.MIN_VALUE)
     }
 }
+
+/**
+ * Reason why a navigation state update was rejected by the container.
+ *
+ * - [OTHER] — unspecified reason.
+ * - [STALE] — the update's provenance indicates that is is based on a stale state.
```
kligarski:
```suggestion
 * - [STALE] — the update's provenance indicates that it is based on a stale state.
```

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
kkafar: Decided to go with all the changes in this PR. The current implementation does not make sense to land on main anyway.

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -352,8 +359,9 @@ - (void)updateSelectedViewController
   if ([self isMoreNavigationControllerRequestedByOperation:_pendingOperation]) {
     if (![self isMoreNavigationControllerPresentInTabBar]) {
       // If the controller is not visible atm. we'll crash the app.
-      // TODO: Emit rejection event
-      _pendingOperation = nil;
+      [self.tabsHostComponentView tabBarController:self
+                             rejectedStateUpdateTo:_pendingOperation
+                                      currentState:_navigationState];
       return;
```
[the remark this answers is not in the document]
kkafar: If it is nil, and there is no `moreNavigationController` present in the UI, then I think something went really wrong.

I'll add a comment to the types indicating that is MUST NOT be used as the initially selected tab.

ios/tabs/host/RNSTabsHostComponentView.mm
kkafar: No. I've updated the doc comments in our TS types to forbid it. I've also added an explicit assertion that should signal to the users that this is forbidden.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsNavState.kt
kkafar: What is the better name overall? `SCREEN_KEY_MORE_TAB` or `SCREEN_KEY_MORE_NAV_CTRL`? I agree that these should be aligned, but dunno in what direction 😄

ios/tabs/host/RNSTabBarController.mm
kkafar: That's intended. It's a subsection of utils.

kligarski: Ok but at least can we add a space before double dashes? 🥺
```suggestion
#pragma mark -- More Navigation Controller
```

By the way, this is how it looks in Xcode:

Image

You can also do the mark without a dash at all:
```
#pragma mark More Navigation Controller
```

And it looks like this:

Image

I guess that this is way too much discussion about something so insignificant d:

src/components/tabs/host/TabsHost.types.ts
```diff
@@ -134,6 +134,9 @@ export interface TabsHostPropsBase {
    * Depending on configuration and the provenance of the update
    * the update might get accepted or rejected.
    *
+   * `SCREEN_KEY_MORE_NAV_CTRL` MUST NOT be used during initial render to indicate default
+   * selected tab.
```
kligarski, line 136: Can we add similar comment to `TabsContainerProps.defaultRouteName`?

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsNavState.kt
kligarski: Naming is hard. I don't have a strong preference for any of those. Maybe @kmichalikk, @t0maboro, @sgaczol, @LKuchno have any thoughts?

LKuchno: I would choose `SCREEN_KEY_MORE_NAV_CTRL` or `MORE_NAV_CTRL` name as it refers directly to this new prop.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsNavState.kt
kkafar: Yeah, you're right. I've added it initially and haven't cleaned up.

kkafar: https://github.com/software-mansion/react-native-screens/pull/3794/commits/2ed0875719bbb16ab87d63ffe497b8bf8a68d78a

ios/tabs/host/RNSTabBarController.mm
kkafar: The dashes are there to mark it as a subsection. Yeah, let's close this topic

src/components/tabs/host/TabsHost.types.ts
kkafar: https://github.com/software-mansion/react-native-screens/pull/3794/commits/01177794ffddc9b76b7617007776ff105516f89e

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsNavState.kt
kkafar: https://github.com/software-mansion/react-native-screens/pull/3794/commits/e666f66211a9bc5e0d1ed832752ca3296b37ce72

PR #3796, opened 2026-03-24
feat(Android, Stack v5): handle header configuration and custom subviews

kligarski:
## Description

Adds support for header configuration and providing custom subviews for M3 App Bar in Stack v5.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/916.

#### JS API

After some discussions we decided that the API for header configuration will look as follows:

1. `Stack` object will provide a `Stack.HeaderConfig` component only.
2. All header configuration (and header subview configuration) will be done via props on `Stack.HeaderConfig` component.
3. `Stack.HeaderConfig` will be split in similar way to `TabsHost` and `TabsScreen` between platforms:
  - common props will be available on both platforms (`title`, `hidden`),
  - platform-specific props will be available via `android`/`ios` prop.
4. Due to differences between platforms, components related to subviews/bar button items will be separate (e.g. `StackHeaderSubview` will be Android-only).

##### Header attachment

1. `StackHeaderConfig` is attached to `StackScreen`.
2. `StackHeaderCoordinatorLayout` creates `StackHeaderCoordinator` instance. `CL` attaches its own `onHeaderHeightChanged` callback to the coordinator. It will be called when `AppBarLayout` is scrolled and will propagate correct Y offset to `StackScreen`.
3. `StackHeaderCoordinatorLayout` attaches its own callback as `onHeaderConfigAttachListener` on `StackScreen` (weak ref) and if `stackScreen.headerConfig` is available, it uses the config.
3. `StackHeaderCoordinatorLayout.handleHeaderConfigAttach` runs and attaches its own callback as `onHeaderConfigChangeListener` on new config (weak ref). This callback passes all updates to `headerCoordinator` via `applyHeaderConfig` (it also batches the updates). Initial update is performed.

##### Subview attachment

1. `StackHeaderSubview` is attached to `StackHeaderConfig`.
2. `StackHeaderSubview` attaches itself as `onStackHeaderSubviewChangeListener` (weak ref) to the subview (to inform `HeaderCoordinator` about e.g. `collapseMode` prop change on the subview).
3. `HeaderConfig` informs `HeaderCoordinator` via `onConfigChangeListener` about the new subview.

##### Offset shadow tree update

1. When scrolling event happens, `AppBarLayout.ScrollingViewBehavior.onDependentViewChanged` runs.
2. We override this method in `StackHeaderScrollingViewBehavior` to allow `headerCoordinator` to attach its own callback.
3. This callback allows to update `StackScreen`'s offset via `onHeaderHeightChanged` callback.
5. Additionally, we attach layout and offset listeners to `AppBarLayout` (so that it works with both non-transparent and transparent header). They update size & offset for `HeaderConfig` and offsets for `HeaderSubview`s via `syncShadowState` method.

#### Scrolling behaviors

M3 App Bar suppors various scrolling behaviors. In order to use them, we need to use `ScrollingViewBehavior` via `CoordinatorLayout`. Important thing to note is that the content below the header has **constant** size. Scrolling the content scrolls the app bar and the content. The size of the content is calculated by `CoordinatorLayout` so that the content is fully visible when header is collapsed. When header is expanded, some part of the screen is cut off.

Please also note that for this to work we need to use `NestedScrollView` or at least regular `ScrollView` with `nestedScrollEnabled={true}`.

#### "Transparent" header

The intention of this PR is to focus only on layout-related props, not the final API. I added `transparent` that changes the layout but not the background color.

Another thing to consider is whether `transparent` prop should exist or should it be layout-related only & background color can be handled via regular styling `backgroundColor` prop. This might be even more useful for iOS implementation of the header but it needs to be researched (cc @kmichalikk).

"Transparent" header does not make sense with scrolling behaviors because the content will never be rendered below the header. That's why in this mode the screen fills the entire container and does not move. 

#### Custom subviews sizing

In Stack v4 we faced many problems with handling custom subviews. For Stack v5 we decided that the size of `leading` (`left` in LTR, `right` in RTL), `center` and `trailing` (`right` in LTR, `left` in RTL) subviews will be determined fully by Yoga and it must depend on the size of the content inside of the subview. Subviews will not be aware of the position and size of other subviews and flex behaviors will not work (but obviously if you wrap subview content in a view with specific `width` and `height`, inside the wrapper `flex` will work as usual). The layout of the subview (position in the toolbar) will be determined fully by the native layout.

This simplified approach will allow us to reliably layout the subviews without Yoga and native layout fighting each other.

Stack v5 will additionally support `background` subview for collapsing headers (`medium` and `large`). This view will be rendered as the background of the header. The subview will always be resized to the full size of the `AppBarLayout` (using `flex` will be supported).

#### Layout synchronization between Yoga and native

> [!NOTE]
> 
> The background subview might have `collapseMode: 'parallax'` effect applied - then it's position moves slower than `AppBarLayout`. Light blue box represents background subview with `collapseMode: 'parallax'` as this is the most complicated example, best for the explanation.

headerNativeLayout_v2

headerReactLayout1_v2

headerReactLayout2_v2

##### Title in small header

When `small` header is used, we would usually rely on title prop from `Toolbar`. There is however one caveat when using custom views - title is laid out **before** subviews therefore leading subview will always appear **after** the title. This in isolation might not be considered a problem (due to this being a native behavior) but in `medium`/`large` collapsing header, title is managed by `CollapsingToolbarLayout` which doesn't use the title from `Toolbar` but attaches its own custom `dummyView` as a subview of `Toolbar` -> in this configuration if you add a subview with `Gravity.START`, it will be laid out **before** the title. We don't want the behavior to differ in such manner between header types so we decided that we should align this. It's easier to manage title in `small` behavior and leave `medium`/`large` as is (but it turns out that for RTL we need to do some hacky workarounds for collpsing header either way, details below).

To ensure that title is laid out after the leading subview, we don't rely on `Toolbar.title` but add our own view that tries to 1:1 mimic native title view.

##### Ensuring correct subview order in RTL

For `small` header, we need to make sure that leading subview and title view is added in correct order.

For `medium`/`large` header, the situation is more complicated. `dummyView` is always added last, after our subviews. This causes incorrect ordering. In order to fix this, we need to make sure that `dummyView` is created and attached (by forcing `measure`). Then we manually change the order of the subviews (moving `dummyView` to the index 0 fixes the problem).

##### Adding/removing subviews in runtime

The order of inserting subviews is critical for correct layout and it is very fragile. For now, when subviews are added/removed in runtime, we're recreating the hierarchy from scratch.

https://github.com/user-attachments/assets/30da0d3e-30f5-4f10-8a2d-910adcaffecb

In the future we might research whether we can optimize this to prevent unnecessary rebuilds.

##### Background subview collapse mode

Currently, `background` subview supports only 2 of 3 available `collapseMode`s: `off` and `parallax`. As we wanted to allow using `flex: 1` inside the `background` subview, we need to set the size of the `HeaderConfig` component & use absolute fill on the subview so that it matches `AppBarLayout`. This however is problematic for `pin` collapse mode. If subview takes all the space, it behaves exactly like `off` collapse mode - there is no additional space between the bottom edge of the header subview and bottom edge of `AppBarLayout`/`CollapsingToolbarLayout` so it scrolls immediately. 

In the future, we can consider exposing maunal size configuration for the subview (or finding another solution) to support `collapseMode: 'pin'`.

##### Maintaining scroll on rebuild

Currently, when header needs rebuilding, it doesn't preserve `AppBarLayout`'s scroll position and it expands automatically.

https://github.com/user-attachments/assets/7d688874-c060-402f-a631-7cdcb225a310

We should research whether we can cache the scroll offset between rebuilds.

##### `hitSlop`

`hitSlop` works correctly now but in the context of native hierarchy. When in collapsing toolbar you click below `MaterialToolbar` even though the `hitSlop` range would be enough, the touch won't be registered. Additionally, collapsing header uses `dummyView` that does not display any text when header is expanded but it will prevent touches which might be surprising for users.

We should consider supporting `hitSlop` in full `HeaderConfig` area.

##### Margins for header items

Margins between subviews are inconsistent due to differences in layout.

| `small` | `large` |
| --- | --- |
| margin_small | margin_large |

We should consider exposing full margin configuration for subviews (it's a little bit complicated with the margins used in native layout) and maybe provide more consistent defaults.

##### Scroll flags

Android allows full customization of scrolling behavior.

We need to expose the scroll flags in one of the follow-up PRs. They are hard-coded for now.

##### Handling insets

For now, we're relying on `fitsSystemWindow` on `AppBarLayout` to apply padding to avoid system bars and display cutout. This will be problematic for nested stack support, using insets from decor view in first render and maunal control (similar to https://github.com/software-mansion/react-native-screens/pull/3835).

In the future, we should manually apply the inset (but I remember from my research that this might be more difficult than it seems). If insets are read from decor, we might be able to use choreographer for `requestLayout` in `StackContainer` instead of `post` (currently, using choreographer as in tabs causes layout jump on first render after rebuild).

##### Title layout customization

We need to add support for centering the title & handle our managed title view in `small` header.

##### Navigation icon and menu

We need to add navigation icon (back button) and allow its customization. In the future we want to also expose Menu API.

##### RTL text ellipsize bug

There are some problems with text ellipsize in RTL when custom subviews are used.

https://github.com/user-attachments/assets/5531e385-8f44-4932-bf68-a0c4225b57d0

We should check whether this is a native bug or not.

### JS
- refactored file structure to match convention from Tabs
- added shared `StackHeaderConfig` component with Android base implementation and skeleton for iOS
- added Android-only `StackHeaderSubview` component
- added Fabric specs to new components
- updated `react-native.config.js`
- adjusted `StackContainer` to accept header configuration as a prop and render header config component if the prop is non-null
- fixed debug logs to avoid infinite recursive calls

### Native
- added `detachFromCurrentParent` `View` extension
- extracted `ShadowStateProxy` from `StackScreen` to allow re-use between components
- moved header-related files to separate package
- added native implementation of `StackHeaderConfig` and `StackHeaderSubview`
- added `OnHeaderConfigAttachListener`, `OnHeaderConfigChangeListener` interfaces
- added logic to handle subviews in the header
- adjusted layout flow to ensure that `requestLayout` works

### C++
- added custom shadow nodes for `StackHeaderConfig` and `StackHeaderSubview`
- ensured that RTL doesn't change absolute positioning for `HeaderSubview`

## Visual documentation

| General |
| --- |
| <video src="https://github.com/user-attachments/assets/82ce9856-1342-40b2-8812-0aebbfa8d952" /> |

| Small opaque | Small translucent |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/fd644c78-fc47-4d4d-8e92-86c2a5733f63" /> | <video src="https://github.com/user-attachments/assets/520c0e3d-321e-4aed-a415-38ade883ace9" /> |

| Large opaque | Large translucent |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/afeb6720-cd60-4f11-8549-7908856186f1" /> | <video src="https://github.com/user-attachments/assets/e999198c-82cd-4be8-916f-1832ad14bd84" /> |

Medium is analogous to large, just differs in size.

## Test plan

Run `test-stack-subviews.tsx`. This test contains many props as subview layout is depended on many factors but some parts of the test should be extracted and tested separately (e.g. hidden, transparent, ...). With @LKuchno we decided that this SFT will be split and scenarios will be added *in the future* as some parts of the implementation might still change.

## Checklist

- [x] Ensured that CI passes


android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/config/StackHeaderConfigProviding.kt
```diff
@@ -0,0 +1,25 @@
...
+    val type: StackHeaderType
+    val title: String
+    val hidden: Boolean
+    val transparent: Boolean
+    val leadingSubview: StackHeaderSubviewProviding?
+    val centerSubview: StackHeaderSubviewProviding?
+    val trailingSubview: StackHeaderSubviewProviding?
+    val backgroundSubview: StackHeaderSubviewProviding?
+
+    val isRtl: Boolean
```
kmichalikk: In the repo we write this as `RTL` instead of `Rtl` so we should probably stick to that

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/StackHeaderCoordinator.kt
```diff
@@ -0,0 +1,480 @@
...
+                coordinatorLayout.addView(appBar, 0)
+                setContentBehavior(coordinatorLayout)
+            }
+
+            // Make sure that we receive insets, necessary when changing header mode in runtime.
+            appBar.requestApplyInsets()
+            attachAppBarListeners(appBar)
+
+            populateAppBar(appBar, config)
+            maybeApplyRtlCollapsingToolbarLayoutWorkaround(coordinatorLayout, config, appBar)
```
kmichalikk: RTL

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/StackHeaderCoordinator.kt
```diff
@@ -0,0 +1,480 @@
...
+        }
+    }
+
+    // endregion
+
+    // region Content behavior
+
+    private fun setContentBehavior(coordinatorLayout: StackHeaderCoordinatorLayout) {
+        val params = coordinatorLayout.stackScreenWrapper.layoutParams as CoordinatorLayout.LayoutParams
+        if (params.behavior == null) {
```
kmichalikk, line 331: does it mean we can only set the behavior once?

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/StackHeaderCoordinator.kt
```diff
@@ -0,0 +1,480 @@
...
+
+        subview.updateContentOriginOffset(
+            x = subviewPos[0] - appBarPos[0],
+            y = subviewPos[1] - appBarPos[1],
+        )
+    }
+
+    // endregion
+
+    private fun maybeApplyRtlCollapsingToolbarLayoutWorkaround(
```
kmichalikk: RTL

android/src/main/java/com/swmansion/rnscreens/gamma/stack/screen/StackScreenViewManager.kt
```diff
@@ -29,6 +31,55 @@ class StackScreenViewManager :
...
+    // in the React tree, we only need to handle the last index specially.
+    override fun removeViewAt(
+        parent: StackScreen,
+        index: Int,
+    ) {
+        if (index == getChildCount(parent) - 1 && parent.headerConfig != null) {
+            parent.headerConfig?.let { parent.detachHeaderConfig(it) }
+        } else {
+            super.removeViewAt(parent, index)
+        }
```
kmichalikk, line 78: Normally, I wouldn't count on that being true. But since we are controlling it, maybe we could leave it as is

kmichalikk (approved):

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/StackHeaderCoordinator.kt
kligarski: That's what we do right now, we set it once and keep it while it's necessary. I didn't notice any problems with this approach. If something comes up in the future, we can reapply the behavior on each rebuild.

android/src/main/java/com/swmansion/rnscreens/gamma/stack/screen/StackScreenViewManager.kt
kligarski: Well, we're not rendering header config in `StackScreen` directly, we're doing this in our example container implementation (`StackContainer.tsx`).

I decided to add additional checks to throw an exception in runtime if HeaderConfig is not the last child of StackScreen.

https://github.com/software-mansion/react-native-screens/pull/3796/commits/8da43614e891b069edbee93b20ad55bcfeff09df

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/StackHeaderCoordinator.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3796/commits/d181137b21812b6f73baf21c2ae3434287d59766

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/StackHeaderCoordinator.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3796/commits/d181137b21812b6f73baf21c2ae3434287d59766

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/config/StackHeaderConfigProviding.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3796/commits/d181137b21812b6f73baf21c2ae3434287d59766

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/config/StackHeaderConfig.kt
```diff
@@ -0,0 +1,103 @@
...
+
+    internal fun removeConfigSubviewAt(index: Int) {
+        getConfigSubviewAt(index)?.let { removeConfigSubview(it) }
+    }
+
+    internal fun removeAllConfigSubviews() {
+        backgroundSubview?.let { removeConfigSubview(it) }
+        leadingSubview?.let { removeConfigSubview(it) }
+        centerSubview?.let { removeConfigSubview(it) }
+        trailingSubview?.let { removeConfigSubview(it) }
```
t0maboro, line 98: This will call `notifyConfigChanged` 4 times. Should we consider batching these notifications?

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/config/StackHeaderConfigProviding.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    val isRTL: Boolean
+
+    fun updateHeaderFrame(
+        width: Int,
+        height: Int,
+        contentOffsetY: Int,
+    )
+
+    var onConfigChangeListener: WeakReference<OnHeaderConfigChangeListener>?
```
t0maboro: WeakReference should be a part of contract, or rather an implementation detail of the class implementing that interface? Wondering whether we might need more flexibility regarding that variable in the future.

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/config/StackHeaderConfigViewManager.kt
```diff
@@ -0,0 +1,127 @@
+package com.swmansion.rnscreens.gamma.stack.header.config
```
t0maboro, line 1: Q: in the old implementation, in `ScreenStackHeaderConfigViewManager` we had `onDropViewInstance` overridden. Do we need a similar override here?

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/subview/StackHeaderSubviewCollapseMode.kt
```diff
@@ -0,0 +1,15 @@
+package com.swmansion.rnscreens.gamma.stack.header.subview
+
+import com.google.android.material.appbar.CollapsingToolbarLayout
+
+enum class StackHeaderSubviewCollapseMode {
```
t0maboro, line 10: nit: we may leave a comment why we're not supporting `PIN` for now, I know it's in PR description, but it's really long, so we can easily miss that context

apps/src/shared/gamma/containers/shared/safe-stringify.ts
```diff
@@ -0,0 +1,23 @@
+export function safeStringify(value: unknown, space?: number): string {
+  const seen = new WeakSet();
+  return JSON.stringify(
+    value,
+    (_key, val) => {
+      if (typeof val === 'object' && val !== null) {
+        if (seen.has(val)) {
+          return '[Circular]';
+        }
```
t0maboro, line 13: nit: shared ref. will be treated as `[Circular]`, but maybe it's fine in our case and sth that we just need to be aware of
```
const a = {};
const obj = { x: a, y: a };
```

apps/src/shared/gamma/containers/shared/safe-stringify.ts
```diff
@@ -0,0 +1,23 @@
...
+        seen.add(val);
+      }
+      if (typeof val === 'function') {
+        return '[Function]';
+      }
+      // React elements
+      if (val != null && typeof val === 'object' && '$$typeof' in val) {
+        return '[ReactElement]';
```
t0maboro: https://react.dev/reference/react/isValidElement ?

common/cpp/react/renderer/components/rnscreens/RNSStackHeaderSubviewComponentDescriptor.h
```diff
@@ -0,0 +1,19 @@
...
+namespace facebook::react {
+
+class RNSStackHeaderSubviewComponentDescriptor final
+    : public ConcreteComponentDescriptor<RNSStackHeaderSubviewShadowNode> {
+ public:
+  using ConcreteComponentDescriptor::ConcreteComponentDescriptor;
+
+  void adopt(ShadowNode &shadowNode) const override {
+    ConcreteComponentDescriptor::adopt(shadowNode);
+  }
```
t0maboro: Q: is the comment
```
// Note: Be very careful with calling `shadowNode.setSize` here. By doing
// that, you are likely to introduce a regressions on both platforms. See:
// https://github.com/software-mansion/react-native-screens/pull/2905
```

from old impl still valid here?
If not, do we need to override this method?

t0maboro (approved):

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/config/StackHeaderConfigViewManager.kt
kligarski: This was added a long time ago: https://github.com/software-mansion/react-native-screens/pull/284.

With our current stack model we will keep the screen & its children rendered until the transition ends.

I also checked that the subview remains in place.

https://github.com/user-attachments/assets/6c24860f-99a7-44d5-8fb2-b5a4c91ac473

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/config/StackHeaderConfig.kt
kligarski: It's already batched in `StackHeaderCoordinatorLayout` via `post`.

I understand that in isolation `StackHeaderConfig` doesn't really know about it but on the other hand, I guess that if we change `removeAll...` method, it would make sense to also allow using `add/removeHeaderSubview` without a notify because you might want to change only 2 views and now you'll send 2 notifications. But we use those methods only via view manager and I'm not sure if we can run `notify` manually after the entire transaction. `onAfterUpdateTransaction` runs only after changing props, not children.

What do you think?

common/cpp/react/renderer/components/rnscreens/RNSStackHeaderSubviewComponentDescriptor.h
kligarski: We're relying on size from Yoga so it's valid. I removed the redundant override.

https://github.com/software-mansion/react-native-screens/pull/3796/commits/b62258ee2b4fb65ded6678d9a6b1c4ee02e7b8e3 & https://github.com/software-mansion/react-native-screens/pull/3796/commits/4c34b0335f318ad3131aedf17ce830743c4933fd

apps/src/shared/gamma/containers/shared/safe-stringify.ts
kligarski: https://github.com/software-mansion/react-native-screens/pull/3796/commits/5e85b14774f1cb66387d4ecd9fcb5849f0d4f204

apps/src/shared/gamma/containers/shared/safe-stringify.ts
kligarski: I added a comment - do you think it's enough?

https://github.com/software-mansion/react-native-screens/pull/3796/commits/5e85b14774f1cb66387d4ecd9fcb5849f0d4f204

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/subview/StackHeaderSubviewCollapseMode.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3796/commits/3b3e51b6ee698e8df9a4b75a62265684f6a6ae29

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/config/StackHeaderConfigProviding.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/3796/commits/050aa98348789f0dda522091d21b30b3a973eeb1

I also changed the interface to expose only `setOnConfigChangeListener` instead of entire property to avoid confusion, e.g. if you set the listener and then read `null` from the same property, this would be surprising

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/config/StackHeaderConfig.kt
t0maboro: > you might want to change only 2 views and now you'll send 2 notifications

this convinces me, we can leave it in this form, to have a parity between removeALl and remove each subview separately

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/config/StackHeaderConfigViewManager.kt
t0maboro: ok, thanks for checking it

apps/src/shared/gamma/containers/shared/safe-stringify.ts
t0maboro: yep, for now it's fine for me, we can update it at any time if needed, because it's not a part of a lib

kmichalikk (approved):

PR #3776, opened 2026-03-18
feat(Android, Tabs): partially implement RFC-1028 for Tabs on Android

kkafar:
## Description

> [!note]
> I'll wait with landing this PR until I get approvals on https://github.com/software-mansion/react-native-screens/pull/3781. These two should be landed in tandem. 

Partially implements [RFC-1028](https://github.com/software-mansion/react-native-screens/issues/3702) for the Tabs component on Android. This PR changes the communication model between the native and JS sides of the Tabs implementation, switching from *controlled mode* to *managed mode*, and introduces a richer event payload for tab change events.

This is part of a broader refactor series for the Tabs component. State conflict resolution algorithm (as described in RFC-1028) is deferred to a follow-up PR.

Related to #3702.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/990

### Android (native)

- **Switched from *controlled* to *managed* mode**: The container now first updates its model and then sends events to the Element realm, rather than waiting for JS to confirm state.
- **Replaced `TabsHostNativeFocusChangeEvent` with `TabsHostTabChangeEvent`**: The new event carries an extended payload (`selectedScreenKey` + `provenance`) aligned with RFC-1028 model.
- **Moved menu model construction out of appearance update**: Previously, `BottomNavigationMenuView` model was built inside `updateBottomNavigationViewAppearance`. It's now a separate step that happens before any container update is triggered, which is architecturally cleaner and required by the new update flow.
- **Streamlined container update logic**: Native update now begins with `TabsContainer::onMenuItemSelected` (triggered by `OnMenuItemClicked` listener). JS-initiated update begins with `TabsHost` setting an operation on `TabsContainer` and flushing it, which in turn triggers `onMenuItemSelected` via `bottomNavigationView.menu.selectedItemId` assignment.
- **Removed `ContainerUpdateCoordinator`**: Replaced with the stack model — operations are flushed on `UIManagerListener.didMountItems` callback, aligning with how Stack works on Android and iOS.
- **`TabsScreen` no longer tracks its own selection state**: `isFocused` prop has been removed. Selection state is now passed to the native side via the new `navState` prop on `TabsHost`.
- **Added `TabsContainerDelegate`** and **`MenuHelpers`** as supporting infrastructure.

### JS / TypeScript

- **`onNativeFocusChange` replaced with `onTabChange`**: New event name follows RFC-1028. The payload is richer, providing more context. This is a **breaking change** — downstream code will need to adapt.
[...]

PR #3799, opened 2026-03-25
fix(Android, Tabs): fix issue with badges being assigned to wrong menu items

kkafar:
## Description

Fix issue with badges being assigned to wrong menu items
For some reason the badge appearance update code used menuItemIndex
instead of menuItemId for associating badge with menu item.

I've fixed it & the code is more resistant to refactors right now.

The issue has been originally reported here: https://github.com/software-mansion/react-native-screens/pull/3776#issuecomment-4089134458

## Visual documentation

image

## Test plan

Run `TestBottomTabs` and make sure that the badges are displayed as defined in test code.


LKuchno:
Fix checked on android emulator.
Work ok - badges are placed in correct tabs now.

PR #3808, opened 2026-03-26
chore: make all exports from v5 component modules explicit

kkafar:
## Description

While doing https://github.com/software-mansion/react-native-screens/pull/3785 I've 
encountered a problem, where I could not export any non-type symbol from `src/components/tabs/index.ts` because such symbol would be exported twice - once from `src/index.ts:'export type * from './types'` and once from `src/index.ts:<my added symbol export>` resulting in duplicated exports error. This made me notice that our API export structure is wrong. This PR changes that to one consistent structure across all of our new components. It should not be a breaking change unless somebody used deep-imports. 

I'll create a doc describing the new system later, but here me out here first.

1. Each component module should be responsible for defining it's own public API at `src/components/<component-name>/index.ts` boundary. Anything not directly importable from that file is not a part of public API. 
2. Root `src/index.tsx` file just reexports all the symbols exported at the component module level.
3. No `export * from ...` type exports on the component module boundary level. Only explicit exports are allowed. It's better to not ship some symbol and add it later, than publish something by accident and remove it later.
4. No default exports neither from component module boundary level nor component implementation modules.

Such setup will also allow for easier library modularisation down the line.

## Changes

Replaces wildcard (`export *` / `export type *`) re-exports with explicit named exports across all v5 (experimental) component modules: `Tabs`, `Stack`, `Split`, `SafeAreaView`, and `ScrollViewMarker`.

Also converts default exports to named exports for these components so that each module's public surface is clearly declared in one place.

- **`src/components/tabs/index.ts`** — explicit type exports, `export const Tabs` instead of `export default`
- **`src/components/gamma/stack/index.ts`** — explicit type exports, `export const Stack` instead of `export default`
- **`src/components/gamma/split/index.ts`** — explicit type exports, `export const Split` instead of `export default`
- **`src/components/safe-area/SafeAreaView.tsx`** — named export instead of default
- **`src/components/safe-area/index.ts`** — new barrel with explicit exports
- **`src/components/gamma/scroll-view-marker/`** — named export, explicit re-exports
- **`src/experimental/index.ts`** — simplified to `export *` from each component index (which now have explicit surfaces)
- **`src/experimental/types.ts`** — deleted; types are now owned by their respective component index files
- **`src/index.tsx`** — `export type *` for v4 types; `export * from './components/tabs'` to correctly expose `SCREEN_KEY_MORE_NAV_CTRL` and `Tabs` as a named export
- **`src/types.tsx`** — removed stale re-exports of tabs & shared types (moved to `src/index.tsx`)

## Test plan

- Verified that `Tabs`, `Stack`, `Split`, `SafeAreaView`, `ScrollViewMarker` remain importable from their respective entry points.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -66,4 +66,4 @@ export { default as useTransitionProgress } from './useTransitionProgress';
  * EXPERIMENTAL API BELOW. MIGHT CHANGE W/O ANY NOTICE
  */
 export * from './components/tabs';
-export type * from './components/shared/types'; // TODO: Do we need to reexport these types?
+export type * from './components/shared/types';


src/components/gamma/split/index.ts
```diff
@@ -1,16 +1,30 @@
 import SplitHost from './SplitHost';
 import SplitScreen from './SplitScreen';
 
-export * from './SplitHost.types';
-export * from './SplitScreen.types';
+export type {
+  DisplayModeWillChangeEvent, // TODO: This event should be renamed to match the convention
```
kkafar, line 5: @t0maboro asking you for opinion here. I've added this comment while refactor to remember to ask you alter. Do you think we should rename it?

src/components/gamma/stack/index.ts
```diff
@@ -1,15 +1,21 @@
 import StackHost from './StackHost';
 import StackScreen from './StackScreen';
 
-export * from './StackHost.types';
-export * from './StackScreen.types';
+export type { StackHostProps } from './StackHost.types';
+
+export type {
+  OnDismissEventPayload,
+  EmptyEventPayload, // TODO: Remove this from public types (we need one shared type for this)
```
kkafar, line 8: Yeah, this type is defined multiple times across various components code. It should be defined once in the `shared` types & reused from there. I'll do it separately.

src/components/tabs/index.ts
```diff
@@ -1,17 +1,48 @@
...
+} from './host';
+
+export type {
+  TabsScreenEventHandler,
+  TabsScreenOrientation,
+  TabsScreenPropsBase,
+  TabsScreenProps,
+  // Android
+  TabBarItemLabelVisibilityMode,
+  TabsScreenItemStateAppearanceAndroid,
```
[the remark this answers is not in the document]
kkafar: We don't consider deep-imports changes to be a breaking change.

src/experimental/index.ts
```diff
@@ -3,14 +3,9 @@
...
-export { default as Stack } from '../components/gamma/stack';
-
-export { default as Split } from '../components/gamma/split';
-export { default as SafeAreaView } from '../components/safe-area/SafeAreaView';
-
-export { default as ScrollViewMarker } from '../components/gamma/scroll-view-marker/ScrollViewMarker';
+export * from '../components/gamma/stack';
+export * from '../components/gamma/split';
+export * from '../components/safe-area';
+export * from '../components/gamma/scroll-view-marker';
```
kkafar, line 11: top-level index file just forwards public APIs defined by each of the modules.

src/index.tsx
```diff
@@ -62,4 +65,5 @@ export { default as useTransitionProgress } from './useTransitionProgress';
 /**
  * EXPERIMENTAL API BELOW. MIGHT CHANGE W/O ANY NOTICE
  */
-export { default as Tabs } from './components/tabs';
+export * from './components/tabs';
+export type * from './components/shared/types'; // TODO: Do we need to reexport these types?
```
kkafar: I think we do. I'll update this later.

src/components/gamma/split/index.ts
t0maboro: https://github.com/software-mansion/react-native-screens/blob/bf3c55dc14e7bc4c41d3c0d68056fc589dc16958/ios/gamma/split-view/RNSSplitViewHostController.swift#L447-L454 please keep in mind that as for now it's nicely consistent with the native method signature imo, but I'm open for your propositions

kkafar: I think we should value API consistency more, than cross-realm (JS/native) consistency. 

I think that this might also apply to other components.

kkafar: I'll create a ticket for this, because we need to make decision here before tabs stablisation.

kkafar: https://github.com/software-mansion/react-native-screens-labs/issues/1083

src/components/gamma/stack/index.ts
kkafar: https://github.com/software-mansion/react-native-screens-labs/issues/1084

src/components/safe-area/SafeAreaView.tsx
```diff
@@ -9,7 +9,7 @@ import SafeAreaViewNativeComponent, {
 } from '../../fabric/safe-area/SafeAreaViewNativeComponent';
 import { StyleSheet } from 'react-native';
 
-function SafeAreaView(props: SafeAreaViewProps) {
+export function SafeAreaView(props: SafeAreaViewProps) {
```
t0maboro, line 12: Q: Should we treat named exports as some convention at this level?
I mean, you're changing it here, but SplitHost remained untouched, should we align it too?

kkafar:
I've ran the tests locally: 
image

They pass. no idea why CI fails. I will investigate it separately.

PR #3827, opened 2026-03-31
refactor(tabs): extract TabsContainerItem and memoize tab screens

kkafar:
## Description

Refactor `TabsContainer` to avoid unnecessary re-renders of all tab screens when only one tab's state changes. Currently, every tab screen is rendered inline inside `TabsContainer`, meaning any state change (e.g. tab selection) causes all screens to re-render. Extracting each screen into a memoized `TabsContainerItem` component allows React to skip re-rendering tabs whose props haven't changed.

## Changes

- Extracted tab screen rendering logic (including safe area wrapping and context provision) into a new `TabsContainerItem` component wrapped with `React.memo`.
- Consolidated `setRouteOptions` and `selectTab` callbacks into a `useTabsNavigationMethods` hook that returns a stable memoized object, so it can be passed to children without causing unnecessary re-renders.
- Added `TabsNavigationMethods` type to `TabsContainer.types.tsx`.
- Moved `getContent` and `getSafeAreaViewEdges` helper functions from `TabsContainer.tsx` to `TabsContainerItem.tsx` (co-located with their only consumer).

## Test plan

- Verified tabs behavior in the FabricExample app — tab selection, safe area insets, and route options updates work as before.
- Checked render logs (`RNSLog.info`) to confirm that non-selected tabs are no longer re-rendering on tab switch.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/apps/src/shared/gamma/containers/tabs/TabsContainerItem.tsx
+++ b/apps/src/shared/gamma/containers/tabs/TabsContainerItem.tsx
@@ -1,17 +1,11 @@
 import React from 'react';
 import { Tabs } from "react-native-screens";
-import type { TabRoute, TabRouteConfig, TabsNavigationMethods } from './TabsContainer.types';
+import type { TabRouteConfig } from './TabsContainer.types';
+import type { TabsContainerItemProps } from './TabsContainerItem.types';
 import { SafeAreaView, type SafeAreaViewProps } from 'react-native-screens/experimental';
 import { Platform } from 'react-native';
 import { TabsNavigationContext } from './contexts/TabsNavigationContext';
 import { RNSLog } from 'react-native-screens/private';
-
-export type TabsContainerItemProps = {
-  route: TabRoute;
-  navMethods: TabsNavigationMethods;
-  isSelected: boolean;
-  pendingForUpdate: boolean;
-}
 
 export const TabsContainerItem = React.memo(TabsContainerItemImpl);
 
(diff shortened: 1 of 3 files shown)


apps/src/shared/gamma/containers/tabs/TabsContainerItem.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { Platform } from 'react-native';
+import { TabsNavigationContext } from './contexts/TabsNavigationContext';
+import { RNSLog } from 'react-native-screens/private';
+
+export type TabsContainerItemProps = {
+  route: TabRoute;
+  navMethods: TabsNavigationMethods;
+  isSelected: boolean;
+  pendingForUpdate: boolean;
+}
```
kligarski: Maybe we can move this to new file `TabsContainerItem.types.tsx`?

apps/src/shared/gamma/containers/tabs/TabsContainer.tsx
```diff
@@ -211,3 +131,25 @@ function useSanitizeRouteConfigs(routeConfigs: TabRouteConfig[]) {
...
+
+function useTabsNavigationMethods(dispatch: React.Dispatch<TabsNavigationAction>): TabsNavigationMethods {
+  const setRouteOptions = React.useCallback(
+    (routeKey: string, options: Partial<TabRouteOptions>) => {
+      dispatch({ type: 'set-options', routeKey, options });
+    },
+    [dispatch],
+  );
+
+  const selectTab: SelectTabMethod = React.useCallback(
```
t0maboro, line 143: nit: no need to define type explicitly, `TabsNavigationMethods` type is covering that

apps/src/shared/gamma/containers/tabs/TabsContainerItem.tsx
kkafar: https://github.com/software-mansion/react-native-screens/pull/3827/commits/2e2722f1bdafcdc2f834866e42f365791004d5f5

PR #3833, opened 2026-03-31
feat(iOS, Stack v4): allow background color customization

sgaczol:
## Description

This PR adds customization of `backgroundColor` for `ScreenStack`.

Added a `nativeContainerStyle` prop to `ScreenStack` that accepts a `backgroundColor`. This background color is applied to the native stack container itself, which sits behind all screens. When screens have a transparent background, the container color is visible through them. A fast swipe-back gesture can also briefly expose the container background.

Each screen still supports setting its own background color at two levels: natively through the Screen component's style prop, and via a React view rendered inside the screen. The React view sits on top of the native layer, so it effectively overrides the native screen background when both are set. Similarly, a non-transparent native screen background will cover the container background underneath. 

`Test3833` has been added to verify this behavior, allowing independent configuration of the container background, native screen backgrounds, and React view backgrounds.

On iOS, unlike `TabsHost`, where the default `nativeContainerBackgroundColor` is set to `[UIColor systemBackgroundColor]`, the `ScreenStack` defaults to `nil`. This is intentional — a native `UINavigationController` does not set a background color on its view by default, and we want to match that native behavior.

On Android, the native view hierarchy differs — ScreenStack is used directly as the container, so this prop is a noop.

Closes https://github.com/issues/assigned?issue=software-mansion%7Creact-native-screens-labs%7C1054

## Changes

- add `nativeContainerStyle` prop to `ScreenStack` with `backgroundColor` field
- add `nativeContainerBackgroundColor` prop to `ScreenStackNativeComponent`
- add `Test3833`

<video src="https://github.com/user-attachments/assets/8d32950b-1aa0-4447-86fd-9f9156490933" /> 

## Test plan

Run `Test3833`.

## Checklist

- [ ] Ensured that CI passes


apps/src/tests/issue-tests/index.ts
```diff
@@ -193,6 +192,7 @@ export { default as TestPreload } from './TestPreload';
 export { default as TestActivityStateProgression } from './TestActivityStateProgression';
 export { default as TestHeaderTitle } from './TestHeaderTitle';
 export { default as TestHeaderHeight } from './TestHeaderHeight';
+export { default as Test3833 } from './Test3833';
```
kligarski: nit: please move this to line 186, below other issue test screens with numbers

ios/RNSScreenStack.mm
kligarski, line 1: For clarity I'd add `_nativeContainerBackgroundColor = nil;` in `initCommonProps`.

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
```diff
@@ -34,6 +36,11 @@ class ScreenStack(
     private var childrenDrawingOrderStrategy: ChildrenDrawingOrderStrategy? = null
     private var disappearingTransitioningChildren: MutableList<View> = ArrayList()
 
+    var nativeContainerBackgroundColor: Int? by Delegates.observable(null) { _, oldValue, newValue ->
+        if (newValue != oldValue) {
+            background = newValue?.toDrawable()
```
kligarski: I wonder what to do on Android. 

In Tabs, there is a `TabsHost` which is a react view (we don't explicitly support it in types but you can pass `style={ backgroundColor: ... }` and it will work on `TabsHost` view. TabsHost displays `TabsContainer` and `nativeContainerStyle` is applied to the background of `TabsContainer`.

In Stack v4, it seems that we're using `ScreenStack` (which is a react view) directly to display views of the fragments as children (`ScreensCoordinatorLayout`, ...). Setting `background` here is equivalent to `style={ backgroundColor: ... }` in JS. What is more, `ScreenStackProps` extends `ViewProps` so we directly "support" using `style` prop.

I'm not sure what we should do for Stack v4 then:
1. Have `nativeContainerStyle` only supported for iOS
2. Pass `nativeContainerStyle.backgroundColor` directly in JS to `style.backgroundColor`
3. Do what we're doing now but I'm not sure about the interaction between those 2 ways of setting background color (in the example it seems like our setting overrides `style` but I'm not sure if that's going to be consistent)

cc @kkafar

kligarski (commented):
Looks good, thank you. I only have one concern regarding Android.

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
kkafar: Make it iOS only. On Android it make it a noop. Explain in the doc comment of that prop, that the native view hierarchy differs between platforms and this prop configures only a iOS specific view.

apps/src/tests/issue-tests/Test3833.tsx
```diff
@@ -0,0 +1,221 @@
...
+  Pressable,
+  Button,
+} from 'react-native';
+import {
+  Screen,
+  ScreenStack,
+  ScreenStackHeaderConfig,
+} from 'react-native-screens';
+import Colors from '../../shared/styling/Colors';
+import LongText from '../../shared/LongText';
```
t0maboro: nit: you can rebase to main to use aliased imports for consistency

apps/src/tests/issue-tests/index.ts
sgaczol: done: https://github.com/software-mansion/react-native-screens/pull/3833/commits/72475ad75db555da5b74160d4c7ce5a3392bc925

android/src/main/java/com/swmansion/rnscreens/ScreenStack.kt
sgaczol: done: https://github.com/software-mansion/react-native-screens/pull/3833/commits/ab0179691ab2e19342b50102c6118d8b5a5b8514

ios/RNSScreenStack.mm
sgaczol: https://github.com/software-mansion/react-native-screens/pull/3833/commits/02a5dc41db72d80808e5c22baba358cf8860a403
done

apps/src/tests/issue-tests/Test3833.tsx
sgaczol: done: https://github.com/software-mansion/react-native-screens/pull/3833/commits/0505276d3e3ab99746e100bda323157415bd1f5c

kligarski (approved):
Thanks. Looks good to me. 

Please remember to create a ticket to expose this property in `react-navigation` & add it to `react-navigation` docs after next release of `screens`.

PR #3835, opened 2026-04-02
feat(Android, Stack v4): Add prop for manual opt-out from applying inset to header

t0maboro:
## Description

Add a prop for ignoring the potential whitespace caused by the status bar top inset being applied when stack navigators are nested on Android.

`disableTopInsetApplication` prop was introduced on `ScreenStackHeaderConfigProps` to opt out from applying insets in a whole subtree, starting from the topmost stack with a visible header.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1096

## Changes

- added `disableTopInsetApplication` prop

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/b2a81ed6-bbef-4cd2-a093-610db2fe1e8f" /> | <video src="https://github.com/user-attachments/assets/4ebe6b30-f42a-4852-a73f-66b38729019e" />  |

## Test plan

Test3835

requires changes in react-navigation

`native-stack/src/types.tsx`

in `NativeStackNavigationOptions`
```tsx
  /**
   * ...
   */
  disableTopInsetApplication?: boolean;
```

`native-stack/src/views/useHeaderConfigProps.tsx`

```tsx
export function useHeaderConfigProps({
  ...
+  disableTopInsetApplication,
...
return {
    backButtonInCustomView,
...
+ disableTopInsetApplication
...
```

## Checklist

- [ ] Ensured that CI passes


src/types.tsx
```diff
@@ -796,6 +796,18 @@ export interface ScreenStackHeaderConfigProps extends ViewProps {
    * therefore this prop loses its relevance.
    */
   topInsetEnabled?: boolean;
```
kligarski, line 813: I wonder if we should use old `topInsetEnabled` prop? We deprecated it some time ago but it turns out that it's still necessary. In `react-navigation` there is some logic related to this prop, I'm not sure if it would interfere with our new implementation.

src/types.tsx
```diff
@@ -796,6 +796,18 @@ export interface ScreenStackHeaderConfigProps extends ViewProps {
...
+  /**
+   * When set to `true`, this screen's header will not consume the top inset.
+   *
+   * By default the topmost visible header in the hierarchy consumes the top inset.
+   * Use this prop to opt-out from that behavior for this specific header
+   * without disabling it globally.
+   *
+   * Has no effect when `androidLegacyTopInsetBehavior` is enabled.
```
kligarski:
```suggestion
   * Has no effect when `androidLegacyTopInsetBehavior` feature flag is enabled.
```

src/types.tsx
```diff
@@ -796,6 +796,18 @@ export interface ScreenStackHeaderConfigProps extends ViewProps {
    * therefore this prop loses its relevance.
    */
   topInsetEnabled?: boolean;
+  /**
+   * When set to `true`, this screen's header will not consume the top inset.
+   *
+   * By default the topmost visible header in the hierarchy consumes the top inset.
```
kligarski: Maybe we should mention that header consumes the inset **and applies padding** -> this is the visible result for the user.

src/types.tsx
```diff
@@ -796,6 +796,18 @@ export interface ScreenStackHeaderConfigProps extends ViewProps {
...
+   * Use this prop to opt-out from that behavior for this specific header
+   * without disabling it globally.
+   *
+   * Has no effect when `androidLegacyTopInsetBehavior` is enabled.
+   *
+   * @platform android
```
kligarski, line 824: I thought about adding a default value here but it's not a convention in Stack v4 so leaving this up to you.

kligarski (commented):
Also, I'm not sure if we're still keeping `GUIDE_FOR_LIBRARY_AUTHORS.md` updated?

src/types.tsx
t0maboro: My considerations were that:
- topInsetEnabled - on our side, historically, it was describing whether the top is active for *this* header: https://github.com/software-mansion/react-native-screens/pull/545 - every header could enable/disable that independently
- disableHeaderTopInsetConsumption - consuming inset for *this* header will have an impact on the whole subtree, especially for nested header padding

That's a good callout. Whether we should have a new prop, I'm not 100% convinced of the current solution and API shape. Maybe we should just handle `topInsetEnabled` and allow downstream to propagate that information down the hierarchy. If we decide to have a separate mechanism, I think that a new prop is safer than undeprecating the previous one and changing its semantic

src/types.tsx
t0maboro: I'm allowing `undefined`, only explicit `true` from the user is disabling the inset, so no default is needed.

src/types.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3835/changes/583f0b114f4c0ec1cc19f5b2eeda54cf0f496563

src/types.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3835/changes/71778246642cb38aaf8d88a86545163091b8b929

t0maboro:
> Also, I'm not sure if we're still keeping `GUIDE_FOR_LIBRARY_AUTHORS.md` updated?

https://github.com/software-mansion/react-native-screens/pull/3835/changes/b5b25ca266d460fa2082a1f8df46e4ded4469583

src/types.tsx
kligarski: I think that the sentence might be rephrased to highlight that we will not apply the padding. Applies to `GUIDE_FOR_LIBRARY_AUTHORS.md` as well.

kligarski (approved):
Only one more comment from me, let's just make sure that we discuss old prop vs new prop approach (new prop seems reasonable to me).

src/types.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3835/commits/0576591dfcf05fd4afe847436540f8713af6b198

kkafar:
Out of context:

1. If on the most outer stack, somebody sets `disableHeaderTopInsetConsumption=true`, then  it, and the **all** nested stacks do not insert padding to account for *top inset*.
2. If an most outer stack does not have a header, then it is not considered the "most outer stack with header", and logic recurses.
3. In regular scenario, when we have multiple nested stacks, only the most outer one should apply the top inset to its appbarlayout.

src/types.tsx
t0maboro: updated the approach, as mentioned in: https://github.com/software-mansion/react-native-screens/pull/3835#issuecomment-4215158380 - now it must be a completely new prop, as the behavior changed significantly

t0maboro:
> Out of context:
> 
> 1. If on the most outer stack, somebody sets `disableHeaderTopInsetConsumption=true`, then  it, and the **all** nested stacks do not insert padding to account for _top inset_.
> 2. If an most outer stack does not have a header, then it is not considered the "most outer stack with header", and logic recurses.
> 3. In regular scenario, when we have multiple nested stacks, only the most outer one should apply the top inset to its appbarlayout.

refined the approach - these were covered

kmichalikk (approved):
I had an idea that it may break if we had a "real" full screen modal that would be put inside the stack, and need to consume insets after inset consumption was disabled, but with current implementation, I haven't found a way to do it.

src/types.tsx
```diff
@@ -796,6 +796,19 @@ export interface ScreenStackHeaderConfigProps extends ViewProps {
...
+   *
+   * This prop only takes effect on the outermost visible header in the hierarchy.
+   * Setting it on an inner stack has no additional impact because a parent stack
+   * has already made the decision (whether inset should be consumed or not).
+   *
+   * Has no effect when `androidLegacyTopInsetBehavior` feature flag is enabled.
+   *
+   * @platform android
+   */
+  disableTopInsetConsumption?: boolean;
```
kligarski: During our last discussion did we agree on this name? I recall talking about whether we should use `consumption` here. Note that in native tree we should consume the inset but not apply the padding (but it probably is out of scope of this PR, something similar was planned in https://github.com/software-mansion/react-native-screens/pull/3405).

apps/src/tests/issue-tests/Test3835.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+const NestedStack = () => {
+  const config = useHeaderConfig();
+
+  return (
+    <InnerStack.Navigator
+      screenOptions={{
+        title: 'Inner Stack',
+        headerShown: config.innerShown,
+        disableTopInsetConsumption: config.disableInset,
```
kligarski: You use the same value for `disableTopInsetConsumption` for both stacks but I guess that the idea behind propagation of this prop was to just use this prop on the outer stack and it propagating down without any additional prop changes.

If outer stack has `headerShown: false, disableTopInsetConsumption: true` and inner stack has `headerShown: true, disableTopInsetConsumption: 'undefined'`, the padding is applied. Is this intended?

Image

I guess that you mention in the docs that `prop only takes effect on the outermost visible header in the hierarchy` but it still surprised me.

kligarski: LLM suggested something like this:

<details>

<summary>Suggestion</summary>

```tsx
// Context: has any ancestor already claimed the top inset?
const TopInsetClaimedContext = React.createContext(false);

export function useTopInsetConsumption(
  headerVisible: boolean,
  disableTopInsetConsumption: boolean,
) {
  const alreadyClaimed = React.useContext(TopInsetClaimedContext);
  const useLegacyBehavior = featureFlags.experiment?.androidLegacyTopInsetBehavior ?? false;

  // We claim the inset if:
  // - legacy mode (always claim), OR
  // - no ancestor claimed it yet, AND our header is visible (hidden headers don't claim)
  const claimsInset = useLegacyBehavior || (!alreadyClaimed && headerVisible);

  // We apply the padding only if we claim it AND haven't been told to suppress it
  const consumesTopInset = claimsInset && !disableTopInsetConsumption;

  // Once claimed (regardless of suppression), mark it as spoken for downstream
  const nextClaimed = alreadyClaimed || claimsInset;

  return { consumesTopInset, useLegacyBehavior, nextClaimed };
}
```
</details>

Maybe we should consider something in this direction?

src/types.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3835/commits/4002c75084528a60c43240b5f12a6632ce675226

apps/src/tests/issue-tests/Test3835.tsx
t0maboro: > for both stacks

it's just for example, because it avoids extra boilerplate, only 1 header is using this prop

> just use this prop on the outer stack and it propagating down without any additional prop changes

in this example, only the top-level header setting matters, I don't see any value for querying it from the outer stack, if the header is hidden

> `headerShown: false, disableTopInsetConsumption: true`

it's clearly described in the documentation that only visible headers matter, with `headerShown: false`, the prop for top inset doesn't make sense at all, thus it's ignored

> inner stack has `headerShown: true, disableTopInsetConsumption: 'undefined'`, the padding is applied

inner stack header is the topmost visible, so it should be responsible for itself regarding top inset, in that case, it's responsible on its own, so explicit `true` is proper way to handle it here imo

t0maboro: regarding `useTopInsetConsumption` refactor, the boolean matrix seems to be equivalent, so it seems that it makes sense to apply that

src/types.tsx
t0maboro: and https://github.com/software-mansion/react-native-screens/pull/3835/commits/50ade8158c257374b2ad6782ee22c7dc40461803

apps/src/tests/issue-tests/Test3835.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/3835/commits/50ade8158c257374b2ad6782ee22c7dc40461803

src/components/contexts/TopInsetApplicationContext.ts
```diff
@@ -0,0 +1,27 @@
...
+  headerVisible: boolean,
+  disableTopInsetApplication: boolean,
+) {
+  const alreadyApplied = React.useContext(TopInsetApplicationContext);
+  const useLegacyBehavior =
+    featureFlags.experiment?.androidLegacyTopInsetBehavior ?? false;
+
+  // We want to apply the inset if:
+  // - legacy mode (androidLegacyTopInsetBehavior)
+  // - or when no ancestor applied it yet abd our header is visible
```
kligarski:
```suggestion
  // - or when no ancestor applied it yet and our header is visible
```

kligarski (approved):
Looks good. Please make sure it works for older API versions as well.

You might update PR description (`disableTopInsetConsumption` -> `disableTopInsetApplication` in `react-navigation` section).

PR #3838, opened 2026-04-02
feat(Tabs): add preventNativeSelection support

kkafar:
## Description

Adds the ability to prevent native tab selection on a per-screen basis. When `preventNativeSelection` is set to `true` on a `TabsScreen`, tapping that tab in the tab bar (or selecting it from the "More" list on iOS) will be blocked. The `TabsHost` receives an `onTabSelectionPrevented` callback with the key of the prevented screen and the current navigation state, allowing JS to decide how to handle the attempt (e.g. show a confirmation dialog, redirect elsewhere).

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1078

## Changes

- **`TabsScreen`**: New `preventNativeSelection` boolean prop plumbed from JS → codegen specs → native on both Android and iOS.
- **`TabsHost`**: New `onTabSelectionPrevented` event (type + codegen spec + native emitter) fired when selection is blocked.
- **iOS (`RNSTabBarController`)**:
  - `tabBarController:shouldSelectViewController:` now checks per-screen `preventNativeSelection` and emits the prevented event.
  - ISA-swizzles `UIMoreNavigationController` to intercept `pushViewController:animated:` so screens behind the "More" list are also subject to prevention. Includes table-view deselection cleanup when a push is blocked.
  - Dynamic subclass name is derived from the actual runtime class (e.g. `RNS_UIMoreNavigationController`), so each distinct original class gets its own correct subclass — safe even if another library ISA-swizzles first.
  - Clears the `OBJC_ASSOCIATION_ASSIGN` back-reference in `dealloc` as a safety measure against potential dangling pointers.
  - Refactored `screenKeyForSelectedViewController` → `screenKeyForViewController:` and `isSelectedViewControllerTheMoreNavigationController` → `isViewControllerTheMoreNavigationController:` to support querying arbitrary view controllers (not just the selected one).
- **iOS (`RNSTabsScreenViewController`)**: Added `ScreenPropsForwarding` category exposing `isPreventNativeSelectionEnabled`.
- **Android**: Prevention logic in `TabsContainer.onMenuItemSelected` — checks `isPreventNativeSelectionEnabled` before progressing state, delegates to `TabsContainerDelegate.onNavStateUpdatePrevented` which emits the event via `TabsHost`.
- **Example app**: New test scenario (`TestTabsPreventNativeSelection`) with 6 tabs, per-tab toggle, and toast on prevention.

## Known issues

- **`experimental_controlNavigationStateInJS` is not respected in the "More" navigation controller flow on iOS.** When the experimental controlled-mode flag is enabled and the user is already on the "More" tab, tapping items in the More list will not be blocked by the controlled-mode gate (only by `preventNativeSelection`). This is acceptable since the controlled-mode feature is planned for removal before release.

### Rudimentary scenarios

We navigate to a "third" tab natively, toggle `preventNativeSelection`, go back, and navigate again. That attempt is blocked (prevented) and an event is emitted, as it can be observed by appearance of toast. We navigate to the "third" tab via JS - this is allowed, as we only prevent native selection - toggle the option and repeat the experiment. This time navigation to "third" tab is allowed.

| iOS | Android | 
| -- | -- |
| <video src="https://github.com/user-attachments/assets/fed8b101-265e-469a-8e18-ad4cc6148618" alt="s-1-ios" /> | <video src="https://github.com/user-attachments/assets/523b0fbd-9366-4943-95bc-c4f7059a62e3" alt="s-1-android" /> |

### More navigation controller scenarios

In the video on the left I showcase that the navigation from "more list" is correctly prevented.

In the video on the right I cover a edge case, where a user / programmer navigates first the the "fifth" tab, then navigates away, e.g. to "third" and then comes back. W/o prevention mechanism, the "fifth" tab would be shown, as it is pushed onto *more navigation controller stack*. I've implemented logic to cover this case. Now, the "more list" will be displayed. *I've decided that in such case we will emit the `OnTabSelectionPrevented` event*, as we effectively prevent s user from navigating to that tab, even if this is not fully intentional. See [more details here](https://github.com/software-mansion/react-native-screens/pull/3838/commits/610a4df4679d4c70db802fc1890b51a57f217f51).

| Block from more list | Forced pop on navigation | 
| -- | -- |
| <video src="https://github.com/user-attachments/assets/67fb3e0e-b3bf-4ad8-8834-67e9cf8b439e" alt="block-from-more-list" /> | <video src="https://github.com/user-attachments/assets/d7302f29-dc27-4d9e-a335-6af5dcdbf2fc" alt="forced-pop-on-navigation" /> |

## Test plan

- Run `TestTabsPreventNativeSelection` scenario from the example app.
- Tap a tab → verify it switches normally.
- Toggle `preventNativeSelection` on a tab → tap it → verify it does **not** switch and the `onTabSelectionPrevented` toast appears.
- On iOS with 6 tabs: verify prevention also works for tabs listed under the "More" navigation controller.
- Toggle the flag back off → verify the tab becomes selectable again.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
@@ -361,6 +361,12 @@ internal class TabsContainer(
 
         val isRepeated = nextSelectedFragment === currSelectedFragment
 
+        // If this is user action we test whether it should be prevented before we progress the state.
+        if (!isRepeated && !isInExternalOperationContext && nextSelectedFragment.isPreventNativeSelectionEnabled) {
+            delegate.onNavStateUpdatePrevented(navState, nextSelectedFragment.requireScreenKey)
+            return false
+        }
+
         val stateChanged = updateSelectedFragment(nextSelectedFragment)
 
         val hasTriggeredSpecialEffect =
--- a/ios/tabs/host/RNSTabBarController.mm
+++ b/ios/tabs/host/RNSTabBarController.mm
@@ -39,7 +39,8 @@ static NSString *const kMoreNavigationControllerScreenKey = @"rnscreens_moreNavi
 static void
 rns_pushViewController(__unsafe_unretained id self, SEL _cmd, UIViewController *viewController, BOOL animated)
 {
-  UITabBarController *rawTabBarController = ((UIViewController *)self).tabBarController;
+  UITabBarController *rawTabBarController = static_cast<UIViewController *>(self).tabBarController;
+
   RCTAssert(
       [rawTabBarController isKindOfClass:RNSTabBarController.class],
       @"[RNScreens] Expected tabBarController to be of class %@, got: %@",
@@ -52,8 +53,9 @@ rns_pushViewController(__unsafe_unretained id self, SEL _cmd, UIViewController *
         .receiver = self,
         .super_class = class_getSuperclass(object_getClass(self)),
     };
-    ((void (*)(struct objc_super *, SEL, UIViewController *, BOOL))objc_msgSendSuper)(
-        &superInfo, _cmd, viewController, animated);
+    const auto msgSendSuperPushViewController =
+        reinterpret_cast<void (*)(struct objc_super *, SEL, UIViewController *, BOOL)>(objc_msgSendSuper);
+    msgSendSuperPushViewController(&superInfo, _cmd, viewController, animated);
   }
 }
 #endif // RNS_MORE_NAVIGATION_CONTROLLER_AVAILABLE
--- a/ios/tabs/host/RNSTabBarController.mm
+++ b/ios/tabs/host/RNSTabBarController.mm
@@ -356,6 +356,7 @@ rns_pushViewController(__unsafe_unretained id self, SEL _cmd, UIViewController *
 
   if (shouldPreventTabSelection) {
     // Ideally we'd call this AFTER we prevent, but there is no appropriate callback.
+    // As long as we emit the event asynchronously this is rather fine.
     [self onDidPreventUserFromSelectingViewControllerWithKey:[self screenKeyForViewController:viewController]];
     return NO;
   }
--- a/ios/tabs/host/RNSTabBarController.mm
+++ b/ios/tabs/host/RNSTabBarController.mm
@@ -830,27 +830,38 @@ rns_pushViewController(__unsafe_unretained id self, SEL _cmd, UIViewController *
 /// (e.g. `RNS_UIMoreNavigationController`), so if another library ISA-swizzles it first or Apple
 /// changes the private class, each distinct original class gets its own correct dynamic subclass.
 ///
-/// This method is idempotent — safe to call multiple times. The dynamic subclass is created once
-/// and reused; `object_setClass` and `objc_setAssociatedObject` are re-applied on each call.
-/// This is necessary because UIKit resets the isa pointer of `moreNavigationController` during
-/// tab bar reconfiguration (e.g. iPad app resize crossing the >5 tab threshold).
+/// This method is idempotent — safe to call multiple times regardless of whether UIKit has
+/// reset the ISA between calls. When the ISA already carries our `RNS_` prefix, only the
+/// associated-object back-reference is refreshed. When UIKit has reset the ISA (e.g. iPad
+/// app resize crossing the >5 tab threshold), the dynamic subclass is looked up (or created)
+/// and re-applied.
 - (void)ensurePushInterceptorOnMoreNavigationController
 {
 #if RNS_MORE_NAVIGATION_CONTROLLER_AVAILABLE
-  Class originalClass = object_getClass(self.moreNavigationController);
-  const char *originalClassName = class_getName(originalClass);
+  Class currentClass = object_getClass(self.moreNavigationController);
+  const char *currentClassName = class_getName(currentClass);
+
+  // If the ISA already points to our dynamic subclass, the interceptor is in place.
+  // Just refresh the associated object (it uses ASSIGN semantics, so it's cheap)
+  // and return early. Without this guard, repeated calls when UIKit has NOT reset
+  // the ISA would stack `RNS_RNS_...` subclasses, causing infinite recursion in
+  // rns_pushViewController's objc_msgSendSuper call.
+  if (strncmp(currentClassName, "RNS_", 4) == 0) {
+    [self installSelfAssociationWithMoreNavigationController];
+    return;
+  }
 
   // Build a unique subclass name per original runtime class: "RNS_<originalClassName>"
   char dynamicSubclassName[256];
-  snprintf(dynamicSubclassName, sizeof(dynamicSubclassName), "RNS_%s", originalClassName);
+  snprintf(dynamicSubclassName, sizeof(dynamicSubclassName), "RNS_%s", currentClassName);
 
   Class dynamicSubclass = objc_getClass(dynamicSubclassName);
 
   if (dynamicSubclass == nil) {
-    dynamicSubclass = objc_allocateClassPair(originalClass, dynamicSubclassName, 0);
-    RCTAssert(dynamicSubclass != nil, @"[RNScreens] Failed to allocate dynamic subclass of %s", originalClassName);
+    dynamicSubclass = objc_allocateClassPair(currentClass, dynamicSubclassName, 0);
+    RCTAssert(dynamicSubclass != nil, @"[RNScreens] Failed to allocate dynamic subclass of %s", currentClassName);
 
-    Method pushMethod = class_getInstanceMethod(originalClass, @selector(pushViewController:animated:));
+    Method pushMethod = class_getInstanceMethod(currentClass, @selector(pushViewController:animated:));
     class_addMethod(
         dynamicSubclass,
         @selector(pushViewController:animated:),
@@ -861,9 +872,15 @@ rns_pushViewController(__unsafe_unretained id self, SEL _cmd, UIViewController *
   }
 
   object_setClass(self.moreNavigationController, dynamicSubclass);
+  [self installSelfAssociationWithMoreNavigationController];
+#endif // RNS_MORE_NAVIGATION_CONTROLLER_AVAILABLE
+}
 
-  // Store a back-reference so the C function can reach this controller.
-  // OBJC_ASSOCIATION_ASSIGN: no retain cycle — self owns moreNavigationController and outlives it.
+/// Stores a weak (ASSIGN) back-reference from `moreNavigationController` to `self`
+/// so that `rns_pushViewController` can reach the owning `RNSTabBarController`.
+- (void)installSelfAssociationWithMoreNavigationController
+{
+#if RNS_MORE_NAVIGATION_CONTROLLER_AVAILABLE
   objc_setAssociatedObject(
       self.moreNavigationController, kRNSTabBarControllerAssociationKey, self, OBJC_ASSOCIATION_ASSIGN);
   _didAccessMoreNavigationController = YES;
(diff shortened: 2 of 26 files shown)


kkafar:
I've found a crash. 


https://github.com/user-attachments/assets/baa19b4f-fae5-4ffe-b65a-d4a516f3d9dc

Need to investigate it next week.

kkafar:
I've also noticed on @satya164 suggestion, that we don't emit an `OnTabSelected` event in case user is on `moreNavigationController` `morelist` screen, and he resizes the app, so that the tab changes to the one previously selected. This breaks the state!

I do not know how to fix it yet, as we literally get no information from UIKit that the tab has been effectively changed - this sucks.

satya164:
@kkafar I was thinking, could `moreNavigationController` be a separate thing instead of a screen key?

- Since we don't get an event when this changes due to screen resize, the JS state that uses screen key will not get out of date, as it would still be the actual screen that was selected
- It'll represent its ephemeral nature, compared to actual screen keys
- For React Navigation, that's how I'm planning to treat it as - so for us it'll be closer to the truth

LKuchno:
I've checked basic scenarios on iPhone 17pro (ios26.2) simulator:

- switching between tabs with option preventNativeSelection disabled 
- setting preventNativeSelection to true - navigate between tabs selecting them from tab bar and using button - ok 
- reset preventNativeSelection from true to false - ok 
- for More Tabs:
      - I also performed above tests, 
      - Added steps where first I've select 5th or 6th tab set preventNativeSelection to true navigate to different tab and then using More tab bar item when back to more tabs list - correct event was emmited - refering to last displayed in More option tab even both tabs have preventNativeSelection to true.
      -  Also check scenario where I set preventNativeSelection: true for all available tabs and then navigate to More tabs list which result with unusable application as I can't navigate to any of tab
      **NOTE**: In my opinion is the way this test screen was design but maybe we shouldn't let user/developer disable selection of ALL tabs ?


I started test on iPad Pro simulator:
- checked first 2 points from iPhone flow but with side bar - ok

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
```diff
@@ -361,6 +361,12 @@ internal class TabsContainer(
 
         val isRepeated = nextSelectedFragment === currSelectedFragment
 
+        // If this is user action we test whether it should be prevented before we progress the state.
+        if (!isRepeated && !isInExternalOperationContext && nextSelectedFragment.isPreventNativeSelectionEnabled) {
```
kligarski, line 365: I'm thinking about what should happen on repeated selection if `preventNativeSelection` is enabled. My first guess would be that `preventNativeSelection` has priority over repeated selection but on the other hand, I don't see any valid use case for this - you can always disable special effects manually.

ios/tabs/screen/RNSTabsScreenComponentView.h
```diff
@@ -81,6 +81,8 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic) BOOL shouldUseRepeatedTabSelectionPopToRootSpecialEffect;
 @property (nonatomic) BOOL shouldUseRepeatedTabSelectionScrollToTopSpecialEffect;
 
+@property (nonatomic) BOOL preventNativeSelection;
```
kligarski, line 84: Shouldn't this be `readonly`? But we're missing this in a lot of places because of Paper support (now dropped).

Should we add a ticket to handle this to https://github.com/software-mansion/react-native-screens-labs/issues/1001?

ios/tabs/screen/RNSTabsScreenViewController.h
```diff
@@ -50,4 +50,12 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+@interface RNSTabsScreenViewController (ScreenPropsForwarding)
```
kligarski:
```suggestion
@interface RNSTabsScreenViewController (TabsScreenPropsForwarding)
```
Maybe it would be better to be precise here?

ios/tabs/screen/RNSTabsScreenViewController.h
```diff
@@ -10,6 +10,8 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@class RNSTabBarController;
```
kligarski: Is it needed?

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -214,6 +278,18 @@ - (void)userDidRepeatViewControllerSelection:(nonnull UIViewController *)viewCon
   [self.tabsHostComponentView tabBarController:self didUpdateStateTo:_navigationState withContext:updateContext];
 }
 
+- (void)setSelectedViewController:(__kindof UIViewController *)selectedViewController
+{
+  NSLog(@"setSelectedViewController:%@", selectedViewController);
```
kligarski: If we want to keep it, then we should use `RNSLog`. Applies to all logs.

LKuchno:
Continued checks on iOS.
**iPad:** 
1. The fix for the problem with settings resetting while resizing the app on iPad (described by @kkafar above) has been tested on an iOS 26.2 iPad Air (M3) and works fine. Smoke tests on iPhone after this change did not find any regressions.
2. Side bar: when prevented tab is selected from side bar its not displayed but if we open side bar this tab is shown as selected - this behavior is not always reproducible but most of the time:

https://github.com/user-attachments/assets/e517273b-cf4d-4118-8798-e821f331a93c

3. iPad and Iphone: 
App is hangs while we prevent one of tab from more tabs switch to other tabs and then go back to more tabs and want to select the tab which is not prevented:

Screenshot 2026-04-07 at 10 00 21

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -633,29 +756,166 @@ - (BOOL)isMoreNavigationControllerTabBarItemSelected
...
 #if RNS_MORE_NAVIGATION_CONTROLLER_AVAILABLE
-  [self.moreNavigationController setNavigationBarHidden:YES animated:NO];
+  if (!self.moreNavigationController.navigationBar.isHidden) {
+    [self.moreNavigationController setNavigationBarHidden:YES animated:NO];
+  }
 #endif // RNS_MORE_NAVIGATION_CONTROLLER_AVAILABLE
 }
 
-- (void)popToRootInMoreNavigationControllerIfNeededAnimated:(BOOL)isAnimated
+- (nullable UIViewController *)popToRootInMoreNavigationControllerRespectSelectionPrevention:
```
kligarski, line 729: I'm not sure if this is the best name for the method - it seems like it's "pop to root unless we can't pop to root due to selection prevention" (even though it doesn't make sense in context of tabs but still I think it's misleading). What it actually does is "if we should prevent selection, then pop to root". 

LLM suggested something like `popToRootInMoreNavigationController:onlyIfTopTabPreventsSelection:animated:` but I'm still not convinced.

kligarski (commented):
This might be related to the bug you mentioned:

https://github.com/user-attachments/assets/54459177-ebec-4be0-9b29-e8f3ae5b0daf

ios/tabs/host/RNSTabBarController.mm
kkafar: I don't, I need it for debugging right now. Just had stashed the changes before I went do something else.

kkafar:
@satya164 could you elaborate what do you mean by "a separate thing"? 

I've initially considered exposing support for the `moreNavigationController` in "silent way", so that when user taps it we either send no information to JS or just a separate event that such action has just been taken. This would however require either a very special treatment when attempting to navigate to `moreNavigationController` or no support for such navigation at all. I've got no idea, whether people (and how often) will have use cases for such navigation, but UIKit exposes such possibility and I did not want to restrict it. If `react-native-screens` restrict it and don't account for it, it'll be hard for anybody in the downstream to add proper support for it. 

On the other hand, I can see that incorporating navigation to `moreNavigationController` into state changes, might complicate the decision of whether to render and what tab. 

Until we stabilise the API it's shape is modifiable. Waiting for any valid input here. 
I think I might create a dedicated discussion from this & ask on X for comments. 

Edit: Also I think I've found a way to get information from UIKit that it changes the tab from `moreNavigationController` to "previously selected one" in cases where `moreTab` disappears. This will require an substantial refactor to core container update logic to take it into account, but should be doable & rather robust.

LKuchno:
**Android 16.0:**

**PROBLEM**:
- currently, after few tabs switch app is reloaded and direction is changed from ltr to rtl - to be investigated
- this block further test in ltr direction but I make smoke tests before reload and on rtl direction to have some feedback already.

On **Android phone and tablet** 
- smoke test ok

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
kkafar: No, I think it does not make sense to prevent repeated selection. Since you already are on a screen - it is selected.

In case of *special effects* - just as you said - it can be handled separately & I think it is the right way to do it.

satya164:
@kkafar I meant something other than the controlled `screenKey` + `onTabSelected` combination.

potential APIs maybe:
- just some event like `onMoreNavigationSelected` without an accompanying controlled prop/method - which still leaves room to add the ability to add a way to manually navigate to it in the future
- event + method on ref - following how uncontrolled components usually work

i'm unsure if a controlled API is appropriate for more navigation controller given that we can't actually control it consistently.

kkafar:
@kligarski @LKuchno 
https://github.com/software-mansion/react-native-screens/pull/3838/commits/34af6a9c1267d0157a1f1159ce9c925fcec2a3e4 should fix the crash you reported.

ios/tabs/screen/RNSTabsScreenComponentView.h
kkafar: I can make this one readonly immediately. Regarding others - it's very very minor thingy - ideally they should all be readonly that's true.

kkafar: But we also should come up with general policy for property handling on both platforms. I'll create separate ticket for this.

kkafar: https://github.com/software-mansion/react-native-screens-labs/issues/1109

ios/tabs/screen/RNSTabsScreenViewController.h
kkafar: https://github.com/software-mansion/react-native-screens/pull/3838/commits/3ae2472ecf5ba0caa50e1d9741f8a88ebba169e2

ios/tabs/screen/RNSTabsScreenViewController.h
kkafar: https://github.com/software-mansion/react-native-screens/pull/3838/commits/12f9c593ddd53d9737a3e575e8fda57ab283db6a

ios/tabs/host/RNSTabBarController.mm
kkafar: > What it actually does is "if we should prevent selection, then pop to root".

That's not accurate. It's rather: "pop to root unless the prevention is not enabled".

I'll stay with current name. If there is a good name-candidate we can rename it later.

kkafar:
I've opened a separate issue for the problem with sidebar. 
https://github.com/software-mansion/react-native-screens-labs/issues/1112

kligarski (approved):
Code looks good. I can check the runtime tomorrow.

kligarski:
> @kligarski @LKuchno [34af6a9](https://github.com/software-mansion/react-native-screens/commit/34af6a9c1267d0157a1f1159ce9c925fcec2a3e4) should fix the crash you reported.

Seems to be fixed now, the runtime on the iPhone simulator looks good.

We can add a background color to tabs to avoid this glitch (iOS 18):

https://github.com/user-attachments/assets/e12782b4-c436-4f0d-b95c-aaa7fea79f80

t0maboro (approved):

shamilovtim:
Related: https://github.com/react-navigation/react-navigation/issues/13130

PR #3840, opened 2026-04-03
chore(tests): Tab-bar-hidden test refactor with new manual scenario

LKuchno:
## Description
test-tabs-hidden name change to follow rules, text added to screen and manual  scenario created in new folder

## Changes
- Updated `tab-bar-hidden.tsx` file name to - 'test-tabs-hidden.tsx.' to follow test naming convention
- Above file was also updated in Scenario secotion + test on the screen was added
- New folder - scenarios were created to store scenarios related to tabs 
- new scenario was added -  'test-tabs-hidden.md' 

Test above button was added. 

Simulator Screenshot - iPhone 17 Pro - 2026-04-03 at 11 50 51

### DIFF
--- a/apps/src/tests/single-feature-tests/tabs/index.ts
+++ b/apps/src/tests/single-feature-tests/tabs/index.ts
@@ -2,7 +2,7 @@ import type { ScenarioGroup } from '@apps/tests/shared/helpers';
 
 import BottomAccessoryScenario from './bottom-accessory-layout';
 import OverrideScrollViewContentInsetScenario from './override-scroll-view-content-inset';
-import TestTabsTabBarHidden from './test-tabs-tab-bar-hidden/test-tabs-tab-bar-hidden';
+import TestTabsTabBarHidden from './test-tabs-tab-bar-hidden';
 import TabsScreenOrientationScenario from './tabs-screen-orientation';
 import TabBarAppearanceDefinedBySelectedTabScenario from './test-tabs-appearance-defined-by-selected-tab';
 import TestTabsColorScheme from './test-tabs-color-scheme';


apps/src/tests/single-feature-tests/tabs/scenarios/test-tabs-hidden.md
kkafar, line 1: I'd strongly suggest naming it `test-tabs-tab-bar-hidden`, because the `test-tabs` is a mandatory prefix now, and it is not clear what the `hidden` part refers to. Additionally we might have more similar tests in the future, e.g. I can see a `test-tabs-bottom-accessory-hidden`.

apps/src/tests/single-feature-tests/tabs/scenarios/test-tabs-hidden.md
```diff
@@ -0,0 +1,24 @@
...
+
+**File:** `test-tabs-hidden`  
+**Platform:** iOS + Android
+**E2E test:** No - to be automated
+
+## Prerequisites
+- iOS 18+ device or simulator
+- Android emulator
+
+---
```
kkafar: I have mixed feelings about this part. It duplicates information that's already in the code-file itself. The "#Prerequisites" part is fine though.

apps/src/tests/single-feature-tests/tabs/scenarios/test-tabs-hidden.md
kkafar, line 1: I am for the idea of having the test scenarios written down. I'm not sure about the form though. 

I remember that we've talked briefly about this in the office, but I have second thoughts now. Do we need to have these descriptions in separate markdowns? What benefit that gives us? I see that the descriptions of steps to perform are rather short, would it be a problem to put it directly in the code files? 

---

Also another topic - if we indeed decide to go in direction of separate markdown files, then I wanna starty another thread. I see that currently you plan to put the markdown files in a separate directory. My first thought here would be to colocate these files together with code-files, so that they are not separated and kept separately. My idea would be to turn every single test scenario into directory with `index` file (ask me for details next week or one of other crew members if you're unsure what I mean here, also LLM might be helpful) and put the markdown file together with the test code.

LKuchno: Good point, I will update it.

apps/src/tests/single-feature-tests/tabs/scenarios/test-tabs-hidden.md
LKuchno: Since the scenario has the same name as screen test file its no needed. I would keep prerequisites and information if this scenario is automated.

apps/src/tests/single-feature-tests/tabs/scenarios/test-tabs-hidden.md
LKuchno: This scenario is short but there will be longer ones - the color scheme test markdown is already 67 lines, so inline in the code file won't scale well.

I think it really depends on how we want to use these manual scenarios:

- **If we test alongside the code** - adding scenarios at the end of the code file is the best option
- **If we treat them as black-box tests** - a separate markdown file works better. We can execute the test just by looking at the `.md` file, without opening an IDE. For this approach, I'm also thinking the markdown could double as a template - paste it into a GitHub Issue per test run, use the interactive checkboxes there, and keep the file in the repo clean. We can talk about it next week.

I think the organization you mean is like the one applied ie. for issue-tests/TestBottomTabs ? 
If so the directory-per-scenario structure sounds great to me.

kkafar: Yeah exactly. Okay, I can agree for having the descriptions in external files. In any case, if this turns out to be not ergonomic we can refactor it easily. 

Let's colocate the files with the code. Turn the tests into directories (migration should be done in separate PR, but the `test-tabs-tab-bar-hidden` you add in this PR should be adjusted here).

PR #3844, opened 2026-04-07
refactor(tests): reorganizing test-tabs-color-scheme and adding scenario

LKuchno:
## Description

Reorganizing test-tabs-color-scheme and adding scenario

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1111 

## Changes

- test-tabs-color-scheme moved to separate scenario-directory and rename to index.tsx file.
- Manual scenario for this screen was also added to directory.
- Small changes on screen - changing colors to make text be visible in light and dark mode.

### Before
In dark mode text was not visible:

Simulator Screenshot - iPhone 17 Pro - 2026-04-07 at 15 39 09

### After
Color changes:
Simulator Screenshot - iPhone 17 Pro - 2026-04-07 at 15 38 34

### DIFF
--- a/apps/src/tests/single-feature-tests/tabs/test-tabs-color-scheme/scenario.md
+++ b/apps/src/tests/single-feature-tests/tabs/test-tabs-color-scheme/scenario.md
@@ -9,9 +9,10 @@
   - `adb shell "cmd uimode night yes"`
   - `adb shell "cmd uimode night no"`
 
-Assumption: system and RN color scheme settings are working correctly. Here only the `colorScheme` prop on TabsHost is tested, verified against different system/RN combinations.
+Assumption: 
 
----
+- System and RN color scheme settings are working correctly.
+- Here only the `colorScheme` prop on TabsHost is tested, verified against different system/RN combinations.
 
 ## Steps
 


apps/src/tests/single-feature-tests/tabs/test-tabs-color-scheme/index.tsx
```diff
@@ -41,15 +41,15 @@ function ConfigScreen() {
   return (
     <ScrollView style={styles.container} contentContainerStyle={styles.content}>
       <View style={styles.section}>
-        <Text>
+        <Text style={styles.text}>
```
kkafar, line 44: I get it why you did it, but likely we need to commit to something we talked last week during one of our meetings - migrated e.g. `ThemedText` component to some simple mechanism & remove react-navigation from the loop.

This is fine for now. @sgaczol I think you should take a look at this when you're back. cc @t0maboro so you're informed.

apps/src/tests/single-feature-tests/tabs/test-tabs-color-scheme/scenario.md
```diff
@@ -0,0 +1,72 @@
+## Test Scenario: colorScheme
```
kkafar: One small nitpick. We should also update this in the test we've already merged.

I suggest starting from header level 1 -> `#`, not `##`.

apps/src/tests/single-feature-tests/tabs/test-tabs-color-scheme/scenario.md
```diff
@@ -0,0 +1,72 @@
...
+
+**Baseline**
+
+1. Launch the app and navigate to the scenario
+   
+- [ ] Expected: Config tab is shown. Pickers default to `unspecified` / `inherit`
+
+---
+
+**TabsHost `inherit` — follows RN/system**
```
kkafar: This way, these particular test-flows will be just heading level 3 - `###`. What do you think?

apps/src/tests/single-feature-tests/tabs/test-tabs-color-scheme/scenario.md
```diff
@@ -0,0 +1,72 @@
...
+
+8. Set system/RN to **dark**, keep TabsHost colorScheme = `dark`
+   
+- [ ] Expected: Tab bar stays **dark**
+
+9. Cycle through `inherit` → `light` → `dark` → `light` → `inherit`
+
+---
+
+**Keyboard tab — smoke**
```
kkafar: This is marked with "smoke". Is this something we'll recognise such tests with?

apps/src/tests/single-feature-tests/tabs/test-tabs-color-scheme/scenario.md
LKuchno: Make sense, I will change headers in existing files to follow this scheme :)

apps/src/tests/single-feature-tests/tabs/test-tabs-color-scheme/scenario.md
LKuchno: No, I used this word to mark as we just check if it works without checking each setting combination, to make it more clear I can change it to 'quick check' or 'simple check' what do you think?

kkafar: That's completely fine. I just didn't get the intention after first read. 

👍🏻

PR #3845, opened 2026-04-07
refactor(tests): reorganizing test-tabs-layout-direction plus scenario

LKuchno:
## Description

Reorganizing test-tabs-layout-direction plus adding scenario

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1113

## Changes

- test move to separate directory under SFT-> tabs 
- test file renamed to index.tsx
- no other changes in test file
- manual sceanrio .md added to the same directory


kkafar (commented):
The same remarks as in #3844

apps/src/tests/single-feature-tests/tabs/test-tabs-layout-direction/scenario.md
```diff
@@ -0,0 +1,83 @@
+# Test Scenario: direction
+
+**E2E test:** No
```
kkafar: I think that in cases where we reject E2E test coverage / consider it infeasible, we should point out exact reasons why - so that if tools (e.g. Detox) capabilities ever change, we'll be able to identify tests that can be automated.

kkafar: Also likely, if we migrate eventually to AI-driven tools, we'll be able to do this.

apps/src/tests/single-feature-tests/tabs/test-tabs-tab-bar-layout-direction/scenario.md
```diff
@@ -0,0 +1,83 @@
...
+    
+- [ ] Expected: Tab bar stays RTL
+
+9. Cycle through inherit → ltr → rtl → ltr → inherit
+    
+- [ ] Expected: Tab bar direction updates immediately with each change, no crash or layout freeze
+
+---
+
+### Precedence chain verification
```
kkafar, line 71:
```suggestion
### Precedence chain verification

```

kkafar (commented):
Seems good. Thank you!

I have one remark. I think we should also introduce this as a general rule applying to all tests 👇🏻 

Let me know what do you think.

apps/src/tests/single-feature-tests/tabs/test-tabs-layout-direction/scenario.md
LKuchno: Good point I will review already added scenarios and add this informations.

LKuchno: E2E test implementation have to be research as I see possibility to cover this test using detox. Ive changes name of test to include %-tab-bar-% plus I added info about e2e in commit: [3e7cfab](https://github.com/software-mansion/react-native-screens/pull/3845/changes/3e7cfab46c70c6a39a58c6c51a727ab10b0ef4be)

kkafar (approved):
LGTM!

PR #3855, opened 2026-04-08
fix(Android): null Screen.fragmentWrapper on dismiss to fix Fabric retain cycle

### LINKED ISSUE #3755: [Android][Fabric] Screen.fragmentWrapper retains destroyed ScreenFragment via Fabric's mTagToViewState
### Description

On Android with Fabric (New Architecture), destroyed `ScreenStackFragment` / `ScreenFragment` instances are retained in memory because `Screen.fragmentWrapper` is never nulled after the fragment receives `onDestroy()`.

Fabric's `SurfaceMountingManager.mTagToViewState` keeps `Screen` and `ScreenContainer` views alive as long as their surface is mounted — this is normal Fabric behavior. But because `Screen.fragmentWrapper` still points to the destroyed fragment, that fragment (and everything it retains) can never be GC'd.

This is a separate leak path from `ScreenDummyLayoutHelper` (#3636), but both stem from the same root cause: references not being cleaned up after Activity/Fragment destruction.

### Leak traces (LeakCanary 2.14)

**Trace 1 — `ScreenContainer.parentScreenWrapper` → `ScreenStackFragment`**
```
┬───
│ GC Root: Global variable in native code
│
├─ com.facebook.react.fabric.FabricUIManager instance
│    Leaking: NO (ScreenContainer↓ is not leaking)
│    ↓ FabricUIManager.mMountingManager
├─ com.facebook.react.fabric.mounting.MountingManager instance
│    ↓ MountingManager.mostRecentSurfaceMountingManager
├─ com.facebook.react.fabric.mounting.SurfaceMountingManager instance
│    ↓ SurfaceMountingManager.mTagToViewState
├─ java.util.concurrent.ConcurrentHashMap instance
├─ SurfaceMountingManager$ViewState instance
│    ↓ SurfaceMountingManager$ViewState.mView
├─ com.swmansion.rnscreens.ScreenContainer instance
│    Leaking: NO (View attached)
│    ↓ ScreenContainer.parentScreenWrapper
│                      ~~~~~~~~~~~~~~~~~~~
╰→ com.swmansion.rnscreens.ScreenStackFragment instance
     Leaking: YES (Fragment#onDestroy() callback received)
     Retaining 941 B in 27 objects
```

**Trace 2 — `Screen.fragmentWrapper` → `ScreenFragment`**
```
┬───
│ GC Root: Global variable in native code
│
├─ com.facebook.react.fabric.FabricUIManager instance
│    ↓ FabricUIManager.mMountingManager
├─ MountingManager instance
│    ↓ MountingManager.mostRecentSurfaceMountingManager
├─ SurfaceMountingManager instance
│    ↓ SurfaceMountingManager.mTagToViewState
├─ ConcurrentHashMap instance
├─ SurfaceMountingManager$ViewState instance
│    ↓ SurfaceMountingManager$ViewState.mView
├─ com.swmansion.rnscreens.Screen instance
│    Leaking: NO (View attached)
│    ↓ Screen.fragmentWrapper
│             ~~~~~~~~~~~~~~~
╰→ com.swmansion.rnscreens.ScreenFragment instance
     Leaking: YES (Fragment#onDestroy() callback received)
     Retaining 855 B in 26 objects
```

### Root cause

`Screen.fragmentWrapper` ([Screen.kt#L57](https://github.com/software-mansion/react-native-screens/blob/main/android/src/main/java/com/swmansion/rnscreens/Screen.kt#L57)) is set when a fragment is created but **never nulled** on fragment destruction.

Meanwhile, `ScreenContainer.parentScreenWrapper` **is** properly nulled in `onDetachedFromWindow()` ([ScreenContainer.kt#L290](https://github.com/software-mansion/react-native-screens/blob/main/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.kt#L290)), showing the intended cleanup pattern already exists in the codebase — it was just not applied to `Screen.fragmentWrapper`.

| Field | Nulled on cleanup? |
|---|---|
| `ScreenContainer.parentScreenWrapper` | Yes (`onDetachedFromWindow`) |
| `Screen.fragmentWrapper` | **No — never nulled** |

### Suggested fix

Null out `screen.fragmentWrapper` in `ScreenFragment.onDestroy()`:

```kotlin
// ScreenFragment.kt — onDestroy()
override fun onDestroy() {
    super.onDestroy()
    // ... existing code ...
    childScreenContainers.clear()
    screen.fragmentWrapper = null  // break the retention chain
}
```

### Environment

- `react-native-screens`: 4.20.0 (also confirmed unpatched on `main` as of 4.24.0)
- React Native: 0.81 (Fabric / New Architecture enabled)
- Android: API 29 (Samsung Galaxy A7)
- LeakCanary: 2.14

ghenry22:
## Description

Closes #3755.

On Fabric, `SurfaceMountingManager.mTagToViewState` keeps `Screen` views alive for the lifetime of the surface. Because `Screen.fragmentWrapper` still points to the destroyed `ScreenFragment` after dismissal, the fragment — and everything it transitively retains — can never be GC'd. The reporter on #3755 measured ~1.2 MB per back-nav cycle on a Galaxy A7 / API 29 with rnscreens 4.20–4.24.

The fix nulls `Screen.fragmentWrapper` in `ScreenFragment.onDestroy`, gated on the existing `isScreenDismissed` discriminator that `onDestroy` already uses to decide whether to dispatch `ScreenDismissedEvent`:

```kotlin
val isScreenDismissed = container == null || !container.hasScreen(this.screen.fragmentWrapper)
```

This gating matters because `onDestroy` can fire in lifecycle paths where the screen has not actually been removed from its container. The unconditional null proposed in #3755 crashes in those paths with `IllegalStateException: Parent Screen does not have its Fragment attached` at `ScreenContainer.setupFragmentManager`. Gating on the existing `hasScreen()` check leaves the back-reference intact in non-dismissal paths and only breaks the cycle when the screen is actually gone.

A second hunk in `ScreenViewManager.onDropViewInstance` does the same cleanup as a fallback for the case where Fabric drops the view without `onDestroy` firing on schedule, guarded by `view.fragmentWrapper?.fragment?.isAdded != true` so it only nulls when no live fragment is attached.

Library precedent for both pieces:
- `ScreenContainer.parentScreenWrapper` is already nulled in `onDetachedFromWindow()`.
- `ScreenStackHeaderConfigViewManager.onDropViewInstance` already calls `view.destroy()`.

## Changes

- `ScreenFragment.kt` — hoist the existing dismissal check into a named `isScreenDismissed` local, then null `screen.fragmentWrapper` inside that branch with an `=== this` identity guard.
- `ScreenViewManager.kt` — add an `onDropViewInstance(view: Screen)` override that nulls `view.fragmentWrapper` when no fragment is currently attached, then calls `super`.

## Test plan

**Validation done:**

- Built against RN 0.83 / Fabric on a real device with rnscreens 4.24.0 + these two hunks applied as a `patch-package` patch.
- Rapid push/pop on a native stack — no crashes, no blank screens.
- Push a screen → push another on top → pop back through both — the underlying screen renders correctly (this is the case the unconditional null breaks).
- Open and dismiss a modal/sheet-presented screen rapidly — no regressions.
- Cross-stack navigation between a tabbed root and a presented stack — no JS↔native state divergence.

**Reproducer for the leak (from #3755):**

1. Push any screen onto a native stack.
2. Press back to pop it.
3. Repeat several times.
4. Take a heap dump (Android Studio Profiler or LeakCanary 2.14+).
5. Before this fix: retained `ScreenFragment` instances appear with the chain `Screen.fragmentWrapper → ScreenFragment` (and `ScreenContainer.parentScreenWrapper → ScreenStackFragment` for stack fragments). The reporter on #3755 measured ~1.2 MB per cycle on a Galaxy A7 / API 29.
6. After this fix: the chain is broken at `Screen.fragmentWrapper` and the fragments are GC'd on the next collection.

**Failure mode this fix avoids:**

The unconditional `screen.fragmentWrapper = null` proposed in #3755 produces:

```
java.lang.IllegalStateException: Parent Screen does not have its Fragment attached
    at com.swmansion.rnscreens.ScreenContainer.setupFragmentManager
```

…on back-nav from any pushed screen, because `onDestroy` fires for screens whose wrapper is still needed by the in-flight transaction. The discriminator-gating in this PR avoids that path entirely — the `hasScreen()` check returns `true` for those screens, and we don't touch the wrapper.

## Checklist

- [ ] For visual changes, included screenshots / GIFs / recordings documenting the change. *(N/A — pure memory fix, no visual change.)*
- [x] For API changes, updated relevant public types. *(N/A — no public API change.)*
- [ ] Ensured that CI passes. *(Pending CI run.)*

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt
@@ -316,7 +316,8 @@ open class ScreenFragment :
     override fun onDestroy() {
         super.onDestroy()
         val container = screen.container
-        if (container == null || !container.hasScreen(this.screen.fragmentWrapper)) {
+        val isScreenDismissed = container == null || !container.hasScreen(this.screen.fragmentWrapper)
+        if (isScreenDismissed) {
             // we only send dismissed even when the screen has been removed from its container
             val screenContext = screen.context
             if (screenContext is ReactContext) {
@@ -324,6 +325,16 @@ open class ScreenFragment :
                 UIManagerHelper
                     .getEventDispatcherForReactTag(screenContext, screen.id)
                     ?.dispatchEvent(ScreenDismissedEvent(surfaceId, screen.id))
+            }
+            // Break the Screen->fragmentWrapper retain cycle so Fabric's
+            // SurfaceMountingManager.mTagToViewState doesn't keep the destroyed
+            // ScreenFragment alive. Gated on isScreenDismissed since onDestroy
+            // can also fire in non-dismissal lifecycle paths where the wrapper
+            // must remain reachable. Identity guard avoids clobbering a wrapper
+            // that has been re-assigned to a different fragment.
+            // See: https://github.com/software-mansion/react-native-screens/issues/3755
+            if (this.screen.fragmentWrapper === this) {
+                this.screen.fragmentWrapper = null
             }
         }
         childScreenContainers.clear()
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenViewManager.kt
@@ -95,6 +95,19 @@ open class ScreenViewManager :
         view.onFinalizePropsUpdate()
     }
 
+    // Belt-and-braces fragmentWrapper cleanup, in case Fabric drops the view
+    // without ScreenFragment.onDestroy firing on schedule. Breaks the
+    // Screen->fragmentWrapper retain cycle so SurfaceMountingManager doesn't
+    // keep the destroyed fragment alive. Guarded so we only null when no
+    // fragment is currently attached.
+    // See: https://github.com/software-mansion/react-native-screens/issues/3755
+    override fun onDropViewInstance(view: Screen) {
+        if (view.fragmentWrapper?.fragment?.isAdded != true) {
+            view.fragmentWrapper = null
+        }
+        super.onDropViewInstance(view)
+    }
+
     fun setActivityState(
         view: Screen,
         activityState: Int,


kkafar:
Thanks for the report. We recently did an investigation & AFAIK there should not be a leak - we'll investigate this nonetheless. cc @t0maboro can I ask you to take a look here?

t0maboro:
EDIT: https://github.com/software-mansion/react-native-screens/pull/3867 - noticed another issue with headerShown: true/false

t0maboro:
Also, @ghenry22 would you be able to prepare and add to this PR a new example that shows the difference between main and this PR - it'd be really helpful for regression testing in the future

Sukhdip-Sandhu:
+1, also experiencing a memory leak that is traced back to here

josemariaeliasMM:
+1, I was led to this MR as I am experiencing the same memory leak

PR #3856, opened 2026-04-08
chore(test): new test-tabs-tab-bar-experimental-userInterfaceStyle with scenario

LKuchno:
## Description

UPDATE: New branch and PR will be created for this test

New screen for experimental_userInterfaceStyle: dark prop (iOS only) with corresponding manual scenario.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1116 

**Changes**
- Adding new screen and scenario in new directory to cover experimental_userInterfaceStyle prop set to dark
- Screen added to tabs/index.ts to be displayed under SFT category.


apps/src/tests/single-feature-tests/tabs/test-tabs-tab-bar-experimental-userInterfaceStyle/darkScreen.tsx
```diff
@@ -0,0 +1,171 @@
...
+    ThemeProvider,
+    DarkTheme,
+} from '@react-navigation/native';
+
+type RootStackParamList = {
+    Screen1: undefined;
+    DarkScreen: undefined;
+};
+
+const Stack = createNativeStackNavigator<RootStackParamList>();
```
kkafar, line 24: Why do we use Stack & utilities from react-navigation here? 

We don't want direct dependency on react-navigation here.

apps/src/tests/single-feature-tests/tabs/test-tabs-tab-bar-dark-experimental-userInterfaceStyle/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+        textAlign: 'center',
+        color: 'white'
+    },
+    screenHint: {
+        fontSize: 14,
+        color: '#555',
+        textAlign: 'center',
+        lineHeight: 20,
+    },
+};
```
kkafar:
```suggestion
};

```

kkafar (changes requested):
Our `single feature tests` should not depend on `react-navigation` in any way. 

Let's discuss why this is the case here first & plan migration path before we land it.

apps/src/tests/single-feature-tests/tabs/test-tabs-tab-bar-experimental-userInterfaceStyle/darkScreen.tsx
LKuchno: I freeze this PR and test implementation until this change will be provided -> [from comment:](https://github.com/software-mansion/react-native-screens/pull/3844#discussion_r3045684262) "migrated e.g. ThemedText component to some simple mechanism & remove react-navigation from the loop.".

LKuchno:
New PR will be created for this test - with 'merged' test for light and dark value and without using react-navigation stack

PR #3860, opened 2026-04-09
chore(test): new test-tabs-tab-bar-light-experimental-userInterfaceStyle with scenario

LKuchno:
## Description

New screen for experimental_userInterfaceStyle: light prop (iOS only) with corresponding manual scenario.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1117

### Changes

- Adding new screen and scenario in new directory to cover experimental_userInterfaceStyle prop set to light
- Screen added to tabs/index.ts to be displayed under SFT category.


kkafar (commented):
Why this test and the one in introduced in https://github.com/software-mansion/react-native-screens/pull/3856 are separate? What's the reasoning here. If there is any, it should be in PR description.

LKuchno:
After discussion I close this PR and I'm going to merge those two tests for - dark and light experimental_userInterfaceStyle values into one test with nested screens as it seems possible to implement and will be better for test organization structure.

PR #3861, opened 2026-04-09
refactor(tests): test-tabs-tab-bar-more-navigation-controller + scenario

LKuchno:
## Description

Moving test-tabs-tab-bar-more-navigation-controller into separate directory with scenario to it.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1120 

## Changes

- Screen moved into a new directory with a corresponding name, extended with '-tab-bar-'
- Screen modification:
       - Fixed onMoreTabSelected toast message to display only 'onMoreTabSelected' instead of the full serialized event string.
       - Added onTabSelected handler emitting a blue toast with the selected tab's key.
- New manual scenario created in the same directory covering different paths:
       -  moving between More tab screen + tabs in it and tabs visible in tab bar normally, 
       -  check on navigation using buttons directly navigate to tabs
       - repeated More tab selection check 
       - check on iPad when More tab is not displayed and when it is - scenario with app resizing


apps/src/tests/single-feature-tests/tabs/test-tabs-more-navigation-controller/scenario.md
kkafar, line 1: I think we could add here test flows with resizing the app on iPad. 

There is at least one more behaviour worth documenting.

When more navigation controller is selected on iPad (More list), and we resize the app, so that the "bottom tab bar" disappears, the previously selected tab is selected.

It would be ideal if we also were able to test navigation state updates in such case. Let's talk about this.

kkafar (commented):
I have few remarks. Let's resolve them before proceeding.

I have also a remark regarding naming. I think that *more navigation controller* has a broader scope than just `tab bar`, therefore I think that this rename is unnecessary here. What do you think?

apps/src/tests/single-feature-tests/tabs/test-tabs-more-navigation-controller/scenario.md
```diff
@@ -0,0 +1,159 @@
...
+
+Other: Ongoing research.
+
+## Prerequisites
+
+- iOS device or simulator: iPhone and iPad (iOS18.6 and iOS26.2)
+
+## Note
+
+- On iPad: The More tab only appears when the window is resized to an iPhone-like view. For iOS 18 and older, a Split View must be triggered to achieve this.
```
kkafar:
```suggestion
- On iPad: The More tab only appears when the window is resized to a compact width size class. For iOS 18 and older, a Split View must be triggered to achieve this.
```

kkafar (commented):
One remark - I have concerns regarding including console logs into the test scenario description. 

This is mainly for one reason -> the logs are not stable, and they can be removed / added and I don't want this to break the tests. 

I think direction here is to test effects only visible on the screen. If we need to react to some event, then we should show a toast / other visible UI element. 

What do you think?

LKuchno:
> One remark - I have concerns regarding including console logs into the test scenario description.
> 
> This is mainly for one reason -> the logs are not stable, and they can be removed / added and I don't want this to break the tests.
>
> I think direction here is to test effects only visible on the screen. If we need to react to some event, then we should show a toast / other visible UI element.
> 
> What do you think?

After discussion, I’m removing DevTools console validation due to log instability and high maintenance overhead. To compensate, I’m adding an onTabSelected toast to verify the behavior in edge cases.
Commit: https://github.com/software-mansion/react-native-screens/pull/3861/changes/634c61961c7e59a1cf63af33d6508e9270441466

Repeated More tab selection added: https://github.com/software-mansion/react-native-screens/pull/3861/changes/f2d611a646ca0cb6b0f0b6a3ce66cd0be7e94d6b

kkafar (approved):

PR #3866, opened 2026-04-10
refactor(test): defining testID for ScenarioButton an assign scenario key property value to it

LKuchno:
## Description

Defined testID for ScenarioButton. 
This testID is mapped in the ScenarioSelect function (found in ScenarioScreen.tsx) to the scenario's key property. This change is required for single-feature tests to allow Detox to select a specific scenario from the domain-specific list (e.g., Tabs).

Fixes: https://github.com/software-mansion/react-native-screens-labs/issues/1125 

### Changes
- testID was added to ScenarioButton function to make it accessible by detox 
- Updated the ScenarioSelect function (in ScenarioScreen.tsx) to pass the scenario's **key property** as the testID to each ScenarioButton.

### DIFF
--- a/apps/src/tests/shared/helpers.ts
+++ b/apps/src/tests/shared/helpers.ts
@@ -21,6 +21,10 @@ export interface Scenario {
    */
   platforms?: ('android' | 'ios')[];
   /**
+   * Test ID for the scenario
+   */
+  testID?: string;
+  /**
    * Component that will render the test scenario. It should be standalone!
    * That means it should be possible to render this w/o any additional harness
    * as top-level application component & it should remain functional.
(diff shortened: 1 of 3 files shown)


apps/src/tests/shared/helpers.ts
```diff
@@ -20,6 +20,10 @@ export interface Scenario {
    * What platforms does this test cover.
    */
   platforms?: ('android' | 'ios')[];
+  /**
+   * testID for the scenario
+   */
+  testID?: string;
```
kkafar: I don't know whether it is a good idea yet or not, but what about using `Scenario#key` property?

LKuchno: It should works here - changed in code to use key as testID value for ScenarioSelect ->[ Commit dac90a2](https://github.com/software-mansion/react-native-screens/commit/dac90a2b9e6f6c00291b2e0be6ab797657c05b6b)
made small doc update here [commit](https://github.com/software-mansion/react-native-screens/pull/3866/changes/496ce553e2db92274558555f2ef1e8db827b85f3)

kmichalikk (approved):

kkafar (approved):

kkafar:
But let's update the PR description before merge!

PR #3873, opened 2026-04-11
feat(android): implement preventNativeDismiss / onNativeDismissCancelled for formSheet

wildseansy:
## Problem

`preventNativeDismiss` and `onNativeDismissCancelled` were no-ops on Android. Because of this, navigation libraries (e.g. React Navigation) that rely on `usePreventRemove` + `onNativeDismissCancelled` to implement "unsaved changes" guards work correctly on iOS but are completely bypassed on Android `formSheet` — the sheet dismisses natively before the JS layer can intercept it.

> **Note:** `gestureEnabled` / `isDraggable` support is tracked separately in #2937. This PR does not touch that prop.

## Root cause

iOS exposes `UIAdaptivePresentationControllerDelegate.presentationControllerShouldDismiss` — a **synchronous** hook that can return `NO` before any visual motion occurs. Android's `BottomSheetBehavior` has no equivalent. By the time any callback fires, the sheet has already moved.

## Solution

Implement the closest Android equivalent using `BottomSheetBehavior.isHideable`:

| Step | What happens |
|------|-------------|
| `preventNativeDismiss=true` set | `BottomSheetBehavior.isHideable = false` → sheet cannot reach `STATE_HIDDEN` via gesture |
| User drags sheet downward | `onSlide` fires; `slideOffset < -0.05f` flags a dismiss attempt |
| User releases | Behavior snaps sheet back to last stable detent |
| Stable state fires | `onNativeDismissCancelled` is dispatched to JS |
| JS receives event | App responds — e.g. shows a "Discard changes?" confirmation |

Runtime changes to `preventNativeDismiss` (e.g. a form going from clean → dirty) are reflected immediately via a property setter that updates `BottomSheetBehavior.isHideable` without requiring a re-mount.

The sheet moves slightly before bouncing back (unlike iOS where the gesture is cancelled before any visual motion), but the **JS contract is identical on both platforms**: `preventNativeDismiss` blocks the dismissal and `onNativeDismissCancelled` fires so the app can respond.

## Example usage

The following pattern works on iOS today. After this PR, it also works on Android `formSheet` with no changes required in application code:

```tsx
import { usePreventRemove } from '@react-navigation/native';
import { useState } from 'react';
import { Alert } from 'react-native';

function EditProfileScreen() {
  const [isDirty, setIsDirty] = useState(false);

  usePreventRemove(isDirty, ({ data }) => {
    // iOS:     fires because presentationControllerShouldDismiss returned NO
    // Android: fires via onNativeDismissCancelled after the sheet bounces back
    Alert.alert(
      'Discard changes?',
      'You have unsaved changes. Leave anyway?',
      [
        { text: 'Stay', style: 'cancel' },
        {
          text: 'Discard',
          style: 'destructive',
          onPress: () => navigation.dispatch(data.action),
        },
      ],
    );
  });

  return (
    <TextInput onChangeText={() => setIsDirty(true)} />
  );
}
```

## Demo of fix in action

https://github.com/user-attachments/assets/0d0ecd91-164a-483d-be19-fac1b3770711

## Files changed

| File | Change |
|------|--------|
| `Screen.kt` | Add `isPreventNativeDismiss` with a setter that keeps `BottomSheetBehavior.isHideable` in sync at runtime |
| `ScreenViewManager.kt` | Wire up `setPreventNativeDismiss` (was a no-op); register `onNativeDismissCancelled` event |
| `SheetDelegate.kt` | Apply `isHideable = !screen.isPreventNativeDismiss` during setup; `SheetStateObserver` tracks drag-toward-dismiss and dispatches event on bounce-back |
| `ScreenEventEmitter.kt` | Add `dispatchOnNativeDismissCancelled()` |
| `ScreenNativeDismissCancelledEvent.kt` | New event class (`topNativeDismissCancelled`) |
| `src/types.tsx` | Extend `@platform` annotations for `preventNativeDismiss` and `onNativeDismissCancelled` to include Android |

## Known differences from iOS

- The sheet moves slightly before bouncing back. iOS cancels the gesture before any visual motion via `presentationControllerShouldDismiss`. Android has no pre-gesture synchronous interception point.
- `DISMISS_ATTEMPT_SLIDE_THRESHOLD = -0.05f` is a heuristic. Open to feedback on whether this value needs tuning.
- Only `formSheet` presentation is affected. `pageSheet` falls back to `modal` on Android and is already JS-navigation-controlled.

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -80,12 +80,6 @@ class Screen(
      * Only takes effect when `stackPresentation` is `formSheet`.
      */
     var isPreventNativeDismiss: Boolean = false
-        set(value) {
-            field = value
-            // Keep BottomSheetBehavior in sync when the prop changes at runtime
-            // (e.g., when a form becomes dirty after initial mount).
-            sheetBehavior?.isHideable = !value
-        }
 
     var isSheetGrabberVisible: Boolean = false
 
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -69,6 +69,24 @@ class Screen(
     var isBeingRemoved = false
 
     // Props for controlling modal presentation
+
+    /**
+     * When `true`, the `formSheet` presentation will not be dismissed by a drag gesture.
+     * Instead, the gesture will bounce the sheet back to its last stable detent and fire the
+     * `onNativeDismissCancelled` event so the JS layer can respond (e.g., show a confirmation
+     * dialog).  Mirrors the iOS `preventNativeDismiss` behaviour powered by
+     * `UIAdaptivePresentationControllerDelegate.presentationControllerShouldDismiss`.
+     *
+     * Only takes effect when `stackPresentation` is `formSheet`.
+     */
+    var isPreventNativeDismiss: Boolean = false
+        set(value) {
+            field = value
+            // Keep BottomSheetBehavior in sync when the prop changes at runtime
+            // (e.g., when a form becomes dirty after initial mount).
+            sheetBehavior?.isHideable = !value
+        }
+
     var isSheetGrabberVisible: Boolean = false
 
     // corner radius must be updated after all props prop updates from a single transaction
--- a/android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
@@ -197,8 +197,12 @@ class SheetDelegate(
         }
 
         behavior.apply {
-            isHideable = true
-            isDraggable = true
+            // When preventNativeDismiss is true the sheet must not be able to reach STATE_HIDDEN
+            // via a drag gesture.  BottomSheetBehavior will then snap back to the last stable
+            // detent, giving us a window to detect the attempt and notify JS via
+            // onNativeDismissCancelled (mirrors iOS presentationControllerShouldDismiss logic).
+            isHideable = !screen.isPreventNativeDismiss
+            isDraggable = screen.isGestureEnabled
         }
 
         // There is a guard internally that does not allow the callback to be duplicated.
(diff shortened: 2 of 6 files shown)


android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
```diff
@@ -627,17 +627,29 @@ class SheetDelegate(
...
+            slideOffset: Float,
+        ) = Unit
+
         override fun onStateChanged(
             bottomSheet: View,
             newState: Int,
         ) {
+            // When preventNativeDismiss is enabled, intercept STATE_HIDDEN before it is
+            // processed further.  BottomSheetBehavior still reaches STATE_HIDDEN via the
+            // drag gesture (isHideable stays true so STATE_COLLAPSED is never reached),
```
kligarski, line 641: The PR description and prop in-code doc mention that we're setting `isHideable = false` but we don't touch this prop in any way and here we're saying that it stays `true`? I don't understand.

android/src/main/java/com/swmansion/rnscreens/bottomsheet/SheetDelegate.kt
```diff
@@ -627,17 +627,29 @@ class SheetDelegate(
...
+            // but we snap it back to the last stable state and notify JS instead of
+            // actually dismissing.  This mirrors the iOS
+            // UIAdaptivePresentationControllerDelegate.presentationControllerShouldDismiss
+            // synchronous hook.
+            if (screen.isPreventNativeDismiss && newState == BottomSheetBehavior.STATE_HIDDEN) {
+                sheetBehavior?.state = lastStableState
+                ScreenEventEmitter(screen).dispatchOnNativeDismissCancelled()
+                return
+            }
```
kligarski, line 650: I think this logic should live in `SheetDelegate`.

android/src/main/java/com/swmansion/rnscreens/events/ScreenNativeDismissCancelledEvent.kt
```diff
@@ -0,0 +1,24 @@
...
+import com.facebook.react.uimanager.events.Event
+
+class ScreenNativeDismissCancelledEvent(
+    surfaceId: Int,
+    viewId: Int,
+) : Event<ScreenNativeDismissCancelledEvent>(surfaceId, viewId) {
+    override fun getEventName() = EVENT_NAME
+
+    // All events for a given view can be coalesced.
+    override fun getCoalescingKey(): Short = 0
```
kligarski, line 14: Should it be even coalesced?

android/src/main/java/com/swmansion/rnscreens/Screen.kt
```diff
@@ -69,6 +69,18 @@ class Screen(
...
+
+    /**
+     * When `true`, the `formSheet` presentation will not be dismissed by a drag gesture.
+     * Instead, the gesture will bounce the sheet back to its last stable detent and fire the
+     * `onNativeDismissCancelled` event so the JS layer can respond (e.g., show a confirmation
+     * dialog).  Mirrors the iOS `preventNativeDismiss` behaviour powered by
+     * `UIAdaptivePresentationControllerDelegate.presentationControllerShouldDismiss`.
+     *
+     * Only takes effect when `stackPresentation` is `formSheet`.
+     */
```
kligarski, line 81: I don't think that this comment is necessary here.

src/types.tsx
```diff
@@ -326,10 +329,20 @@ export interface ScreenProps extends ViewProps {
    */
   onWillDisappear?: (e: NativeSyntheticEvent<TargetedEvent>) => void;
   /**
-   * Boolean indicating whether to prevent current screen from being dismissed.
-   * Defaults to `false`.
+   * Boolean indicating whether to prevent current screen from being dismissed by a drag
+   * gesture.  Defaults to `false`.
```
kligarski, line 333: This property is defined in `ScreenProps` which also apply to regular Screen props, not only the form sheet -> on iOS `preventNativeDismiss` works also with tapping the back button so the doc comment here is incorrect.

kligarski (changes requested):
Thank you for the PR but I don't think that this can land in current shape.

PR #333, opened 2020-02-15
Implement largeTitle backgroundColor and hideShadow on iOS

dsolimando:
New props allowing to:
- define the background color 
- hide the shadow

of the header when the edge of any scrollable content reaches the matching edge of the navigation bar.

See [scrollEdgeAppearance](https://developer.apple.com/documentation/uikit/uinavigationitem/3198041-scrolledgeappearance)

PR #3884, opened 2026-04-15
fix(iOS, Stack v4): Make largeHeaderBackgroundColor work only when largeTitle is on

kmichalikk:
## Description

This PR changes the behavior of `largeHeaderBackgroundColor` to be used only for largeTitle, as the name suggests. This props should actually be named `scrollEdgeBackgroundColor`, because it configures `scrollEdgeAppearance`. Only thing is that at the time of writing the feature, before iOS 15 this was only used for large title. Now this appearance is used for both 
header sizes when scrolling to top.

See the last paragraph in the discussion https://developer.apple.com/documentation/uikit/uinavigationbar/scrolledgeappearance

The feature: https://github.com/software-mansion/react-native-screens/pull/333

> [!caution]
> This change aligns the runtime behavior with the name, not with the way it is implemented (set on scrollEdgeAppearance)

## Changes

Added `largeTitle` check before configuring `largeHeaderBackgroundColor`

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/b42640f6-2130-4dd7-a9f5-1c8b41771bfb" /> | <video src="https://github.com/user-attachments/assets/be086682-6d43-484e-b526-43dd39dc134a" /> |

## Test plan

Use provided issue test.

## Checklist

- [ ] Ensured that CI passes


kligarski (commented):
Looks good but I'm not sure about the direction with this prop (should we match the name or the description of the prop which are not equivalent on iOS 15+ as you mentioned.

kmichalikk:
Quoting @kkafar 

> The name is a bit misleading, I admit, but the description is clear how it is supposed to work. It behaves like that for a long long time likely. If there are no serious reports that this is a problem, I think we shouldn't modify it to work differently from what's described in the "docs". That would be a breaking change.

We're not moving forward with this PR

PR #3886, opened 2026-04-16
fix(iOS, Stack v4): Fix reattaching screens which have preventNativeDismiss set

### LINKED ISSUE #3885: usePreventRemove fails with swipe back in native-stack on Expo SDK 55, causes JS/native desync and touch block
### Description

This used to work before the update. I use the hook to prevent users from swiping back if they have any pending requirements, but after the update, swiping back on iOS doesn't prevent navigation and causes the screen to freeze when I tap it. Any solutions? I've already reported this in the React Navigation repository, but there hasn't been a response yet.

### Steps to reproduce

1. Tap “Go to Login”
2. Try swiping back
3. Once you tap the alert, you won't be able to tap it again.

https://github.com/Julian03-lab/repro-router-bug

### Screens version

4.23.0

### React Native version

0.83.4

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

Real device

### Device model

iPhone 15

### Acknowledgements

Yes

t0maboro:
## Description

When a user triggers a native back gesture on iOS, UIKit automatically removes the screen from its parent. For screens where `preventNativeDismiss` is set to true, the screen might be reattached to the hierarchy, what's causing state desynchronization between JS and native, because natively we're filtering out this screen from `viewControllers` that was introduced in: https://github.com/software-mansion/react-native-screens/pull/3584

Note: I intentionally added this check only to 'card' presentation. For modals, `preventNativeDismiss` is fired before detaching the modal from the hierarchy, so the issue with reattaching isn't reproducible.

Closes: https://github.com/software-mansion/react-native-screens/issues/3885

## Changes

- excluded screens that have `preventNativeDismiss` option set from our solution for preventing reattaching screens that were removed from the hierarchy

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/3571ff56-e3be-4ea1-ad60-f57974efa59f" /> | <video src="https://github.com/user-attachments/assets/79a79c29-504e-4657-bbf9-ff0e079a6490" /> |

## Test plan

Added Test3885, performed regression testing on Test2559

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenStack.mm
+++ b/ios/RNSScreenStack.mm
@@ -703,7 +703,15 @@ RNS_IGNORE_SUPER_CALL_END
           /// quickly. Since view recycling is disabled, once we detect that a screen has been removed from the view
           /// hierarchy, it won't be reused. This allows us to safely filter out dismissed screens from screens coming
           /// from JS state via `controllers`.
-          if (_iosPreventReattachmentOfDismissedScreens && screen.controller.isRemovedFromParent) {
+          ///
+          /// Note: screens with `preventNativeDismiss` are intentionally excluded from this guard.
+          /// When `preventNativeDismiss` is set and the user triggers a native back gesture, UIKit removes
+          /// the screen from its parent. We then need to reattach it so that the `preventNativeDismiss`
+          /// callback fires correctly on the JS side. This breaks the general assumption that a screen
+          /// removed from the hierarchy will never be reattached.
+          /// See: https://github.com/software-mansion/react-native-screens/issues/3885
+          if (_iosPreventReattachmentOfDismissedScreens && screen.controller.isRemovedFromParent &&
+              !screen.preventNativeDismiss) {
             continue;
           }
           [pushControllers addObject:screen.controller];
--- a/apps/src/tests/issue-tests/Test3885.tsx
+++ b/apps/src/tests/issue-tests/Test3885.tsx
@@ -1,0 +1,87 @@
+import * as React from 'react';
+import { View, Text, StyleSheet, Alert, Pressable } from 'react-native';
+import {
+  NavigationContainer,
+  ParamListBase,
+  usePreventRemove,
+} from '@react-navigation/native';
+import {
+  createNativeStackNavigator,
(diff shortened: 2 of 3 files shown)


ios/RNSScreenStack.mm
```diff
@@ -703,7 +703,15 @@ - (void)updateContainer
...
-          if (_iosPreventReattachmentOfDismissedScreens && screen.controller.isRemovedFromParent) {
+          ///
+          /// Note: screens with `preventNativeDismiss` are intentionally excluded from this guard.
+          /// When `preventNativeDismiss` is set and the user triggers a native back gesture, UIKit removes
+          /// the screen from its parent. We then need to reattach it so that the `preventNativeDismiss`
+          /// callback fires correctly on the JS side. This breaks the general assumption that a screen
+          /// removed from the hierarchy will never be reattached.
+          /// See: https://github.com/software-mansion/react-native-screens/issues/3885
+          if (_iosPreventReattachmentOfDismissedScreens && screen.controller.isRemovedFromParent &&
+              !screen.preventNativeDismiss) {
```
kkafar, line 714: What about scenario, where we:

1. first have `preventNativeDismiss` enabled,
2. we attempt to dismiss the screen with gesture, 
3. we get container update & push it here again, but `screen.controller.isRemovedFromParent` stays set to `true`,
4. we disable `preventNativeDismiss`, 
5. receive container update (some render, or another screen pushed).

Will it push the screen again?

kkafar (commented):
I've got a remark, let's dive into it

ios/RNSScreenStack.mm
t0maboro: 3. if we push the screen again, isRemovedFromParent will receive `NO` https://github.com/software-mansion/react-native-screens/blob/4.24.0/ios/RNSScreen.mm#L1908 

is your case still valid then?

kkafar: Looks sensible then!

kkafar (approved):

PR #3896, opened 2026-04-16
fix(Android): RTL layout for center-aligned custom header title

### LINKED ISSUE #3438: [Android] Missing title and an extra space appear when using RTL
### Description

In **RTL** mode, if `headerTitleAlign` is set to `center` and `headerTitle` is defined as a function, the title isn’t displayed and the header gains extra spacing. 

```
<Stack.Navigator 
    screenOptions={{
      headerTitle: () => (
        <View>
          <Text>Elo elo</Text>
        </View>
      ),
      headerRight: () => (
        <View>
          <Text>Right</Text>
        </View>
      ),
      headerTitleAlign: 'center',
    }}
  >
  (...)
  </Stack.Navigator>
```

rtl no title

Interestingly, using a `string` for `headerTitle` makes the header render properly in **RTL** as well.

```
  <Stack.Navigator 
    screenOptions={{
      headerTitle: 'Elo elo',
      headerRight: () => (
        <View>
          <Text>Right</Text>
        </View>
      ),
      headerTitleAlign: 'center',
    }}
  >
  (...)
  </Stack.Navigator>
```

rtl with static title

On the other hand, header is rendered properly with both types in **LTR**.

ltr works fine

### Steps to reproduce

* Use native stack navigator with `headerTitleAlign: 'center'` and pass a function to`headerTitle`
* Use RTL

https://github.com/kkosobudzki/rtl-navigation

### Screens version

4.18.0

### React Native version

0.82.1

### Platforms

Android

### JavaScript runtime

None

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

HakimMohamed:
## Description

When `headerTitleAlign` is `'center'` and `headerTitle` is a function (custom React component), the header title is invisible on Android in RTL mode. LTR works fine, iOS works fine.

**Root cause:** `onNativeToolbarLayout` sends `paddingStart`/`paddingEnd` (logical, direction-aware) to the C++ shadow tree, which uses them as physical `left`/`right` Yoga padding. In RTL, start=right and end=left, so the paddings were swapped — a large right-side value ended up as left padding, leaving near-zero width for the centered title.

Fixes #3438

## Changes

- Detect RTL via `toolbar.layoutDirection` in `onNativeToolbarLayout`
- Convert logical start/end insets to physical left/right before sending to the shadow tree
- Added `Test3438.tsx` reproducing the issue

## Test plan

- Open `Test3438` in FabricExample
- Set device/emulator language to an RTL language (e.g. Arabic)
- Verify "Custom Title" is visible and centered in the header
- Switch language back to English, verify LTR still works

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.kt
@@ -138,30 +138,40 @@ class ScreenStackHeaderConfig(
         }
 
         val isBackButtonDisplayed = toolbar.navigationIcon != null
+        val isRtl = toolbar.layoutDirection == LAYOUT_DIRECTION_RTL
 
-        val contentInsetStartEstimation =
+        val startInsetEstimation =
             if (isBackButtonDisplayed) {
                 toolbar.currentContentInsetStart + toolbar.paddingStart
             } else {
                 max(toolbar.currentContentInsetStart, toolbar.paddingStart)
             }
 
-        // Assuming that there is nothing to the left of back button here, the content
-        // offset we're interested in in ShadowTree is the `left` of the subview left.
-        // In case it is not available we fallback to approximation.
-        val contentInsetStart =
-            configSubviews.firstOrNull { it.type === ScreenStackHeaderSubview.Type.LEFT }?.left
-                ?: contentInsetStartEstimation
+        val endInset = toolbar.currentContentInsetEnd + toolbar.paddingEnd
+        val leftSubview =
+            configSubviews.firstOrNull { it.type === ScreenStackHeaderSubview.Type.LEFT }
 
-        val contentInsetEnd = toolbar.currentContentInsetEnd + toolbar.paddingEnd
+        // The shadow tree expects physical left/right padding, but the toolbar APIs
+        // return logical start/end values which swap sides in RTL.
+        val paddingLeft: Int
+        val paddingRight: Int
+
+        if (!isRtl) {
+            paddingLeft = leftSubview?.left ?: startInsetEstimation
+            paddingRight = endInset
+        } else {
+            paddingLeft = endInset
+            paddingRight =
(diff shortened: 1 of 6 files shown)


android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.kt
```diff
@@ -138,30 +138,40 @@ class ScreenStackHeaderConfig(
         }
 
         val isBackButtonDisplayed = toolbar.navigationIcon != null
+        val isRtl = toolbar.layoutDirection == LAYOUT_DIRECTION_RTL
 
-        val contentInsetStartEstimation =
+        val startInsetEstimation =
```
kkafar: Unnecessary renaming. This value name should stay untouched.

kkafar: Also applies to other names.

kkafar (commented):
Hey, thanks for the PR!

I'm not convinced this is the right fix yet. 

Yoga is RTL aware & it should swap the paddings on its own, when the layout direction changes. I'll try to investigate it a bit deeper soon.

android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.kt
HakimMohamed: Good point dug into it and Yoga does handle logical edges correctly, but the existing setPadding() call maps paddingStart/paddingEnd to physical Edge::Left/Edge::Right, so the values never reach Yoga as logical. Pushed a commit that routes them through Edge::Start/Edge::End instead; tested in FabricExample and my production app, renders correctly in LTR and RTL.

SimpleCreations:
@kkafar any chance you could take another look?

PR #3897, opened 2026-04-16
docs(Android, Stack v5): add missing docs to StackHeaderConfig

kligarski:
## Description

Adds missing docs to `StackHeaderConfig`.

### DIFF
--- a/src/components/gamma/stack/header/StackHeaderConfig.android.types.ts
+++ b/src/components/gamma/stack/header/StackHeaderConfig.android.types.ts
@@ -9,30 +9,125 @@ export type StackHeaderBackgroundSubviewCollapseModeAndroid =
   StackHeaderSubviewCollapseModeAndroid;
 
 export interface StackHeaderToolbarSubviewAndroid {
+  /**
+   * @summary The React component rendered in this toolbar slot.
+   *
+   * The subview is sized by Yoga but positioned by the native layout. Each
+   * subview is placed independently — subviews do not participate in a shared
+   * flex layout and cannot influence each other's sizing.
+   *
+   * Intrinsic sizing and explicit dimensions work as expected. Avoid
+   * parent-relative sizing (e.g. `flex: 1`) on the root element — it will
+   * produce incorrect dimensions. Flex layout within a root that has a known
+   * size works as expected.
+   *
+   * @platform android
+   */
   Component: ReactNode;
 }
 
 export interface StackHeaderBackgroundSubviewAndroid {
+  /**
+   * @summary Controls how the background subview behaves when the app bar
+   * collapses.
+   *
+   * The following values are available:
+   * - `off` - the subview scrolls away with the app bar,
+   * - `parallax` - the subview scrolls at a slower rate, creating a parallax
+   *   effect.
+   *
+   * `pin` is not currently supported because the background subview is
+   * stretched to match the entire `AppBarLayout`, which causes pinned content
+   * to move immediately rather than staying fixed. Support for `pin` collapse
+   * mode might be added in the future.
+   *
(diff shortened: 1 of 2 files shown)


src/components/gamma/stack/header/StackHeaderConfig.android.types.ts
```diff
@@ -9,30 +9,125 @@ export type StackHeaderBackgroundSubviewCollapseModeAndroid =
   StackHeaderSubviewCollapseModeAndroid;
 
 export interface StackHeaderToolbarSubviewAndroid {
+  /**
+   * @summary The React component rendered in this toolbar slot.
+   *
+   * The subview is sized by Yoga but positioned by the native layout. Each
```
t0maboro: nit: not sure if mentioning Yoga here without extra context wouldn't be too low-level for some app developers, as we're touching RN internals
wdyt about sth like "The subview's size is determined by its React Native styles..." ?

src/components/gamma/stack/header/StackHeaderConfig.android.types.ts
```diff
@@ -9,30 +9,125 @@ export type StackHeaderBackgroundSubviewCollapseModeAndroid =
...
+   * subview is placed independently — subviews do not participate in a shared
+   * flex layout and cannot influence each other's sizing.
+   *
+   * Intrinsic sizing and explicit dimensions work as expected. Avoid
```
t0maboro, line 22: `@remarks` or maybe marking it similarly to `backButtonIcon` description?

src/components/gamma/stack/header/StackHeaderConfig.types.ts
```diff
@@ -2,9 +2,46 @@ import { StackHeaderConfigPropsAndroid } from './StackHeaderConfig.android.types
...
+   * @platform android, ios
+   */
   hidden?: boolean | undefined;
+  /**
+   * @summary Specifies if the content should be rendered behind the header.
+   *
+   * When `true`, content is rendered behind the header instead of starting
+   * below it.
+   *
+   * On Android:
```
t0maboro, line 25: should we mention how it affects the layout, e.g. top inset handling?

t0maboro (approved):

src/components/gamma/stack/header/StackHeaderConfig.types.ts
kligarski: Inset handling will be a part of separate PR. Currently, there isn't much to say.

t0maboro: fine

src/components/gamma/stack/header/StackHeaderConfig.android.types.ts
kligarski: https://github.com/software-mansion/react-native-screens/pull/3897/commits/e69a90698a2b8aa197f7273315fa8be5a43a9361

src/components/gamma/stack/header/StackHeaderConfig.android.types.ts
kligarski: https://github.com/software-mansion/react-native-screens/pull/3897/commits/e69a90698a2b8aa197f7273315fa8be5a43a9361

I used "React Native's layout engine". I don't like "React Native styles" because it might suggest that you need to set explicitly `width` and `height` but that's not the case - you can just add a `<Text>` component inside the subview and it will work just fine.

t0maboro: sure, mainly I wanted to avoid Yoga keyword, looks good

Issue #1140, opened 2021-09-20
Going back does not recalculate screen height when parent height is dynamic


somebody32:
## Description

I have a native stack navigator inside a bottom sheet. The bottom sheet's height is dynamic and depends on the currently focused screen. The problem is when going from a taller screen into a smaller one and then back `UIViewControllerWrapperView` and its children (`RNSScreen & RNSScreenView`) stay the size of the previous screen (check flipper screenshot below). Visually it is not changing anything, but taps are not registered anywhere below the old height.

### Screenshots
Here I have the first screen where the whole area is tappable and navigates to the second small screen. When coming back, I can not tap anymore below the top 100px (the height of the second screen).

https://user-images.githubusercontent.com/25742/133996001-1c2a7f48-0220-40af-b4bf-c71c53cde0a7.mov

<img width="1322" alt="image" src="https://user-images.githubusercontent.com/25742/133996127-1e99bdd5-a05f-44cf-9c88-8df4c47105fa.png">

Currently, I'm changing the height via the `useFocusEffect` hook from react-navigation. The problem can go away if I do that on `transitionEnd`, but that would highly delay the animation.

I would be happy to start working on a snack but maybe there is a fast solution that I'm not aware of?

## Package versions
- React: 17.0.2
- React Native: 0.65.1
- React Native Screens: 3.7.2

WoLewicki:
I think the repro is necessary here since it seems like the native screen height passed to the `RNSScreenView` component is wrong for some reason and I see no quick solution to that.

somebody32:
@WoLewicki here you have it, hope it helps https://github.com/somebody32/RNScreensHeightRepro
tried to put in on snack, but the exact bottom sheet lib version doesn't resolve there.

WoLewicki:
I tested the repro you provided and when I swipe to go back, it works correctly. I think the problem comes from the fact that you use `useFocusEffect`, which is not a best option for `native-stack` since it is fired at the beginning of transition when navigating from JS, and at that moment the correct layouts are not set yet. When navigating natively, it is fired only after the transition is done, and it resolves in the proper behavior. Isn't the usage of `transitionEnd` proper from the visual point of view, where you first navigate, and then the bottomSheet gets the correct view? Also, it makes it consistent with navigating natively.

somebody32:
Thank you, @WoLewicki. 

[...]

PR #3906, opened 2026-04-20
fix(iOS, Tabs): prevent double event on more tab navigation

kkafar:
## Description

When a user had something already pushed onto the stack inside the *more tab*, navigated away and then came back, the navigation state update event was dispatched **twice**:

1. First from `tabBarController:didSelectViewController:` delegate callback,
2. Second from `UINavigationControllerDelegate`'s `navigationController:didShowViewController:animated:`.

This PR fixes that by gating the second dispatch behind a new flag, so state only progresses when navigation was actually triggered via `pushViewController`.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1140

### Known risk

The flag is set in the swizzled `pushViewController`, while what we truly want to detect is navigation originating from `UIMoreList`. This could break if `UIMoreNavigationController` calls `pushViewController` from other internal paths — I have not found such edge cases, but open to suggestions for a more precise detection.

### Considered alternative

Deduplication on the `eventemitter` / `TabsHost` layer was considered and rejected — it would mean tolerating incorrect `RNSTabBarController` behavior and working around it upstream. Fixing it at the source is preferred.

### Side effect worth noting

When the user is in tab `Sixth`, selects the more tab item (navigating to the *more list*), and then selects `Sixth` again from the list — this now triggers another `OnTabSelected` event with fresh provenance (native state progresses) and `OnTabSelected#isRepeated` set to **`false`**.

I consider this correct: we inform the programmer *where* the user navigated, and `isRepeated = false` signals "no way this should trigger a special effect", which is the intended purpose of that payload field. Open for discussion.

## Changes

- Added `shouldProgressStateOnMoreNavigationControllerPush` flag on `RNSTabBarController`.
- Flag set to `YES` inside the swizzled `rns_pushViewController`, reset in `init`.
- Renamed the `UINavigationControllerDelegate` callback override from `didShowViewController:` to `willShowViewController:` and guarded `userDidSelectViewController:` call behind the flag — clearing it after use.

## Test plan

Tested with:

- `apps/src/tests/single-feature-tests/tabs/test-tabs-more-navigation-controller.tsx`

Reproduction steps (before the fix):

1. Build the `FabricExample` iOS app with more than 5 tabs configured (triggers the *More* tab).
2. Navigate into a tab accessible only via the *More* list (e.g. `Sixth`).
3. Push some screen onto its stack.
4. Switch to a different visible tab, then return to the *More* tab.
5. Observe that `onNativeStateChange` / `OnTabSelected` fires twice per return.

After the fix only a single event is dispatched.

Additional scenario to verify the intentional side effect:

1. Select tab `Sixth` via the *More* list.
2. Tap the *More* tab item again (goes back to *More* list).
3. Tap `Sixth` again from the list.
4. Expect a new `OnTabSelected` event with `isRepeated = false`.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/tabs/host/RNSTabBarController.mm
+++ b/ios/tabs/host/RNSTabBarController.mm
@@ -21,6 +21,8 @@
 - (BOOL)moreNavigationController:(UINavigationController *)navigationController
         shouldPushViewController:(UIViewController *)viewController;
 
+@property (nonatomic, readwrite) BOOL shouldProgressStateOnMoreNavigationControllerPush;
+
 @end
 
 #if RNS_MORE_NAVIGATION_CONTROLLER_AVAILABLE
@@ -51,6 +53,8 @@ rns_pushViewController(__unsafe_unretained id self, SEL _cmd, UIViewController *
     };
     const auto msgSendSuperPushViewController =
         reinterpret_cast<void (*)(struct objc_super *, SEL, UIViewController *, BOOL)>(objc_msgSendSuper);
+
+    [tabBarController setShouldProgressStateOnMoreNavigationControllerPush:YES];
     msgSendSuperPushViewController(&superInfo, _cmd, viewController, animated);
   }
 }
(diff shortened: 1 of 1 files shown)


LKuchno:
Checked this fix on iPhone 17 pro with ios26.2. Issue #1140 not occur :)

kligarski (approved):
Looks good.

t0maboro (approved):

PR #3907, opened 2026-04-20
chore(tests): allow forcing tab-select action when already selected

kkafar:
## Description

Extends the `selectTab` navigation method and the `tab-select` reducer action
with an optional `forceAction` flag. Previously, dispatching `tab-select` for
the currently selected route was a no-op (the reducer returned the existing
state). With `forceAction: true`, callers can bypass that short-circuit and
re-run the selection logic even when the target route is already active.

This is useful for scenarios where selecting the already-selected tab should
still produce a side effect (for example, to re-trigger an effect subscribed
to selection changes, or to reset internal state tied to a tab).

## Changes

- Added `forceAction: boolean` to `TabsNavigationActionSelectTab`.
- Extended `SelectTabMethod` signature with an optional `forceAction` flag.
- `selectTab` in `useTabsNavigationMethods` now forwards the flag (defaulting
  to `false`) to the dispatched action.
- Updated the reducer's early-return guard in `tabsActionSelectTabHandler` so
  it no longer short-circuits when `forceAction` is set.

## Test plan

Manual verification in `FabricExample`:

1. Open a tabs-based screen in the example app.
2. Call `selectTab(routeKey)` with the current route's key — confirm the state
   remains unchanged (existing behavior preserved).
3. Call `selectTab(routeKey, true)` with the current route's key — confirm the
   reducer proceeds past the early-return and emits a new state.

No new automated tests added; the change is a non-breaking API extension with
a default-off behavior.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/apps/src/shared/gamma/containers/tabs/reducer.tsx
+++ b/apps/src/shared/gamma/containers/tabs/reducer.tsx
@@ -65,7 +65,7 @@ function tabsActionSelectTabHandler(
     return state;
   }
 
-  if (state.confirmedState.selectedRouteKey === action.routeKey) {
+  if (state.confirmedState.selectedRouteKey === action.routeKey && !action.forceAction) {
     return state;
   }
 
--- a/apps/src/shared/gamma/containers/tabs/TabsContainer.tsx
+++ b/apps/src/shared/gamma/containers/tabs/TabsContainer.tsx
@@ -127,8 +127,9 @@ function useTabsNavigationMethods(dispatch: React.Dispatch<TabsNavigationAction>
   );
 
   const selectTab: SelectTabMethod = React.useCallback(
-    (routeKey: string) => {
-      dispatch({ type: 'tab-select', routeKey });
+    (routeKey: string, forceAction?: boolean) => {
+      const shouldForceAction = forceAction ?? false;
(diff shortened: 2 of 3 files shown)


kligarski (approved):
It would be nice to have some kind of test for regression testing.

apps/src/shared/gamma/containers/tabs/reducer.tsx
t0maboro, line 72: I'm observing an additional render triggering tab reselect with `forceAction`. Looks like this line will create a new state object, and in `useTabsHostNavState` we'll recompute `hostNavState`, even if the state is the same regarding values - is this intended?

kkafar: Yeah, exactly. When forcing the action we want to trigger the render.

There is one problem though, however I do not know how to work around it.
Namely, if we navigated to the tab via JS, and then try to navigate via JS again to that tab -> React render will be triggered, but it won't result in any update on the native side, because when diffing props React will conclude that there were no changes.

t0maboro: tough one, in current shape the only path I'm seeing would be via incrementing the provenance with some workarounds for this specific case

PR #3909, opened 2026-04-20
fix(iOS): fallback to `imageNamed` for custom SF symbols

aaronleopold:
## Description

This PR fixes an issue where custom SF Symbols fail to render in iOS components (e.g., stack menus).

Currently, `react-native-screens` uses `systemImageNamed` to resolve symbols but custom ones bundled into the app's asset catalogs are not loadable this way (reference [here](https://developer.apple.com/documentation/uikit/uiimage/init(systemname:)?language=objc#Discussion)).

The docs pointed me to use [`imageNamed`](https://developer.apple.com/documentation/uikit/uiimage/init(named:)?language=objc), instead.

## Changes

`ios/RNSBarButtonItem.mm`

- Updated `resolveImageFromConfig` to use `imageNamed` when `systemImageNamed` is nil

| Before | After |
| --- | --- |
| Before | After |

## Test plan

I did not add tests. The code change was simple enough to get done with linting and following the docs, but I am not familiar enough with objc to know best practices. I can try to spend more time on this area if needed, but mostly tested the fix via a patch applied to local dev builds. You can see that work [here](https://github.com/stumpapp/stump/commit/7be690d248d749c8aa10efcf46e9f40cb6a6d6c7), if interested or helpful at all

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSBarButtonItem.mm
+++ b/ios/RNSBarButtonItem.mm
@@ -136,7 +136,11 @@ static UIMenuOptions RNSMakeUIMenuOptionsFromConfig(NSDictionary *config);
 {
   NSString *sfSymbolName = dict[@"sfSymbolName"];
   if (sfSymbolName != nil) {
-    completionBlock([UIImage systemImageNamed:sfSymbolName]);
+    UIImage *image = [UIImage systemImageNamed:sfSymbolName];
+    if (!image) {
+      image = [UIImage imageNamed:sfSymbolName];
+    }
+    completionBlock(image);
     return;
   }
 


ios/RNSBarButtonItem.mm
```diff
@@ -136,7 +136,11 @@ + (void)resolveImageFromConfig:(NSDictionary *)dict
 {
   NSString *sfSymbolName = dict[@"sfSymbolName"];
   if (sfSymbolName != nil) {
-    completionBlock([UIImage systemImageNamed:sfSymbolName]);
+    UIImage *image = [UIImage systemImageNamed:sfSymbolName];
+    if (!image) {
+      image = [UIImage imageNamed:sfSymbolName];
+    }
```
satya164, line 142: imo it shouldn't implictly fallback to custom image. it should be explicit.

having the same exact way to specify system sf symbols and custom ones makes it impossible to provide proper type checking, since system sf symbols are a union of known values and custom sf symbols can be any string.

kkafar (commented):
I agree with @satya164 here. Why not use `imageSource` or `xcasset` directly?

aaronleopold:
Thank you both for your time and for giving feedback!

> Why not use imageSource or xcasset directly?

I think I see what you are referring to in the code, I missed that conditional block in my narrow trace through. For my use-case, expo-router doesn't support passing a custom `xcasset`. The `icon` is typed as `SFSymbol | ImageSourcePropType` and, again this isn't quite my domain, I can't quite get any sf symbol, custom or system, to load as image sources.

So, if I understand things correctly, it seems my effort is potentially better directed there to make sure I can wire up the toolbar items with `xcasset` instead.

I'll close this PR. Thanks again for the pointer!

PR #3918, opened 2026-04-21
chore(types): remove nullability from some `children` props in unstable/experimental components

kligarski:
## Description

This PR is a follow-up to https://github.com/software-mansion/react-native-screens/pull/3719. We adjust `children` property to be not nullable for `Host` components (Tabs, Split and Stack v5) and subviews in Stack v5 implementation.

This change is applied only to unstable/experimental components so it shouldn't be considered a breaking change.

## Changes

- wrap `ReactNode`/`ViewProps['children']` in `NonNullable` for:
  - `StackHost`
  - `TabsHost`
  - `SplitHost`
  - `StackHeaderSubview` and related places in `StackHeaderConfig`

N/A

## Test plan

`yarn check-types`

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/src/components/gamma/stack/header/android/StackHeaderSubview.android.types.ts
+++ b/src/components/gamma/stack/header/android/StackHeaderSubview.android.types.ts
@@ -9,7 +9,7 @@ export type StackHeaderSubviewTypeAndroid =
 export type StackHeaderSubviewCollapseModeAndroid = 'off' | 'parallax';
 
 export type StackHeaderSubviewProps = {
-  children?: ViewProps['children'] | undefined;
+  children: NonNullable<ViewProps['children']>;
 
   type?: StackHeaderSubviewTypeAndroid | undefined;
   collapseMode?: StackHeaderSubviewCollapseModeAndroid | undefined;
--- a/src/components/gamma/split/SplitHost.types.ts
+++ b/src/components/gamma/split/SplitHost.types.ts
@@ -113,7 +113,7 @@ export type SplitHostCommands = {
 };
 
 export interface SplitHostProps extends ViewProps {
-  children?: React.ReactNode | undefined;
+  children: NonNullable<React.ReactNode>;
   ref?: React.Ref<SplitHostCommands> | undefined;
 
   /**
--- a/src/components/gamma/stack/header/StackHeaderConfig.android.types.ts
+++ b/src/components/gamma/stack/header/StackHeaderConfig.android.types.ts
@@ -26,7 +26,7 @@ export interface StackHeaderToolbarSubviewAndroid {
    *
    * @platform android
    */
-  Component: ReactNode;
+  Component: NonNullable<ReactNode>;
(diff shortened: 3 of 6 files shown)


kligarski (commented):
I decided not to mess with overriding `children` prop in Fabric specs but it's not a public API so it shouldn't be a problem.

src/components/gamma/stack/header/android/StackHeaderSubview.android.types.ts
```diff
@@ -9,7 +9,7 @@ export type StackHeaderSubviewTypeAndroid =
 export type StackHeaderSubviewCollapseModeAndroid = 'off' | 'parallax';
 
 export type StackHeaderSubviewProps = {
-  children?: ViewProps['children'] | undefined;
+  children: NonNullable<ViewProps['children']>;
```
t0maboro, line 12: shouldn't we allow to pass `undefined` for the subview? for me it seems that it might be rendered conditionally in some cases

t0maboro (approved):

src/components/gamma/stack/header/android/StackHeaderSubview.android.types.ts
kligarski: You can pass `undefined` to the entire `*Subview` prop in header config:

```ts
  /**
   * @summary Custom view rendered behind the header content.
   *
   * @platform android
   */
  backgroundSubview?: StackHeaderBackgroundSubviewAndroid | undefined;
  /**
   * @summary Custom view placed in the leading (start) slot of the toolbar.
   *
   * @platform android
   */
  leadingSubview?: StackHeaderToolbarSubviewAndroid | undefined;
```

But `StackHeaderToolbarSubviewAndroid`/`StackHeaderBackgroundSubviewAndroid` enforces non-nullable component and therefore `StackHeaderSubview` component (which won't be a part of public API, header config is the only place you will be able to configure subview) can require non-nullable component.

kligarski: You will be able to conditionally render the subview in header config - is this enough for the use case you brought up?

t0maboro: okay, if undefined subview on HC level and undefined children inside HeaderSubview props have an equivalent domain, we're good to go

kligarski: `StackHeaderConfig.leadingSubview` - may be `undefined`. If so, we don't render `StackHeaderSubview` at all.

If `StackHeaderConfig.leadingSubview` is defined, we require `StackHeaderConfig.leadingSubview.Component` to be also defined and we render `StackHeaderSubview` with `StackHeaderConfig.leadingSubview.Component` as a child.

Therefore `StackHeaderSubview.children` also requires defined component.

PR #3923, opened 2026-04-22
chore(release): set main sentinel version and guard stable publishes

kkafar:
## Description

Adds two coupled changes for the release-branch workflow.

### 1. Pin `main` to a sentinel version (`1000.0.0`)

Bumps `package.json#version` on `main` to `1000.0.0`, per the release guide (see [labs#726](https://github.com/software-mansion/react-native-screens-labs/discussions/726)).

In the release-branch workflow, the version in `main/package.json` is not a real release version — releases are cut from dedicated `<major>.<minor>-stable` branches where the version is set independently. Keeping `main` pinned to a recently-released value (e.g. `4.24.0`) invites confusion and requires a bump on every release cut. Alternatives have their own trade-offs — `react-native-reanimated` uses a `<version>-main` suffix, which demands format discipline on every bump; `react-native-gesture-handler` does nothing, and the version on `main` silently drifts behind the real latest release. The sentinel `1000.0.0` is deliberately absurd so nobody mistakes it for a real release, requires no maintenance on release cuts, and still parses as valid semver for tooling.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1167.

### 2. Guard the publish workflow against non-nightly publishes from `main`

The sentinel is a clear *signal*, but without enforcement nothing prevents an operator from triggering **Publish release to npm** on `main` and pushing `react-native-screens@1000.0.0` to the registry. Adds a pre-publish validation step in `publish-npm.yml` that invokes `scripts/validate-publish-target.sh` whenever a non-nightly release is triggered via `workflow_dispatch`. The script rejects:

- the sentinel version (`1000.*`),
- runs targeting `main`,
- branches that do not match the documented release-branch patterns (`<major>.<minor>-stable` or `v<major>-main`).

All failures are surfaced in a single run so an operator sees every problem at once, not one at a time.

Nightly publishes from `main` remain intentionally untouched — they are the designed exception per the release guide.

## Changes

- `package.json`: version `4.24.0` → `1000.0.0`.
- `.github/workflows/publish-npm.yml`: new `Validate publish target` step, gated on `workflow_dispatch` with non-nightly `release-type`.
- `scripts/validate-publish-target.sh` (new): runs the three checks; exit `0` on success, `1` on validation failure, `2` on usage error.

## Test plan

- Locally: `./scripts/validate-publish-target.sh main stable` against current repo state → all three checks fail (sentinel + main + bad pattern), exit `1`.
- Locally: `./scripts/validate-publish-target.sh 4.25-stable stable` → sentinel check fires (`1000.0.0` on this branch), branch checks pass, exit `1`.
- Locally: invocation with missing args → usage error, exit `2`.
- Post-merge: run `publish-npm.yml` in dry-run against `main` with `release-type=stable` → expect the guard step to fail the workflow before the publish action runs.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-native-screens",
-  "version": "4.24.0",
+  "version": "1000.0.0",
   "description": "Native navigation primitives for your React Native app.",
   "scripts": {
     "submodules": "git submodule update --init --recursive && (cd react-navigation && yarn && yarn build && cd ../)",
--- a/scripts/validate-publish-target.sh
+++ b/scripts/validate-publish-target.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+# Validates that the current branch and package.json version are appropriate
+# for publishing a non-nightly (stable / beta / rc) release.
+#
+# Exits 0 if the target is valid; non-zero (with one or more error messages)
+# otherwise. All failures are reported before exit so the operator sees every
+# problem in a single run.
+#
+# This script assumes it is run from the top level repo directory.
+#
+# Usage: scripts/validate-publish-target.sh <branch> <release-type>
+
+set -euo pipefail
+
+if [[ $# -ne 2 ]]; then
+  echo "Usage: $0 <branch> <release-type>" >&2
+  exit 2
+fi
+
+BRANCH="$1"
+RELEASE_TYPE="$2"
+VERSION=$(jq -r .version package.json)
+
+echo "ref=$BRANCH version=$VERSION release-type=$RELEASE_TYPE"
+
+emit_error() {
+  if [[ -n "${GITHUB_ACTIONS:-}" ]]; then
+    echo "::error::$1"
+  else
+    echo "error: $1" >&2
+  fi
+}
+
+FAIL=0
+
+if [[ "$VERSION" == 1000.* ]]; then
+  emit_error "package.json version is '$VERSION' — this is the main-branch sentinel. Bump the version on a release branch before publishing a $RELEASE_TYPE release."
+  FAIL=1
+fi
+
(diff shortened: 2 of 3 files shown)


scripts/validate-publish-target.sh
```diff
@@ -0,0 +1,52 @@
...
+
+set -euo pipefail
+
+if [[ $# -ne 2 ]]; then
+  echo "Usage: $0 <branch> <release-type>" >&2
+  exit 2
+fi
+
+BRANCH="$1"
+RELEASE_TYPE="$2"
```
t0maboro, line 34: can we add a guard that the passed value is correct? atm, we can pass anything

t0maboro (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-native-screens",
-  "version": "4.24.0",
+  "version": "1000.0.0",
```
kligarski, line 3: Did you find a way to avoid releasing nightly with `1000.0.0` version?

kligarski (approved):

package.json
kkafar: Yeah. The version in package.json here does not matter for nightly. The `npm-package-publish` action will just use the `@latest` tag on npm to deduce the version and increment its minor part by 1 (by default) & replace that in the package.json. So it seems that everything should work fine.

scripts/validate-publish-target.sh
kkafar: https://github.com/software-mansion/react-native-screens/pull/3923/commits/5f427d3cb8d11efba08f2f70ddc27b56bd69d214

PR #3925, opened 2026-04-22
refactor(Example): TabsContainer and StackContainer refactor

sgaczol:
## Description

Previously, we were storing the actual React Component references directly inside the `useReducer` state (within `TabRoute` and `StackRoute` objects). This PR separates the navigation state from the Components. During the render phase, the containers dynamically resolve the correct Component by matching the route name from the state against a `componentsByName` Map (built via `useMemo` from the `routeConfigs` prop), rather than storing Component references in reducer state.

There is one issue related to `StackContainer`: often in our environment, we wrap `StackContainer` in an intermediate component (e.g. `StackSetup`) to consume contexts such as `useToast()`. If that wrapper is not explicitly exported from its file, Fast Refresh causes `useReducer` inside `StackContainer` to re-fire its initialization function, resetting the stack to the first screen. This issue will be solved in a separate PR.

## Changes

- Update `TabRoute` and `StackRoute` types to omit `Component` from the route state
- Update `TabsContainer` and `StackContainer` to resolve components via a `componentsByName` Map during render

N/A

## Test plan

N/A

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/apps/src/shared/gamma/containers/stack/StackContainer.tsx
+++ b/apps/src/shared/gamma/containers/stack/StackContainer.tsx
@@ -25,6 +25,14 @@ import { useParentNavigationEffect } from './hooks/useParentNavigationEffect';
 export function StackContainer({ routeConfigs }: StackContainerProps) {
   useSanitizeRouteConfigs(routeConfigs);
 
+  const componentsByName = React.useMemo(() => {
+    const map = new Map<string, StackRouteConfig['Component']>();
+    for (const config of routeConfigs) {
+      map.set(config.name, config.Component);
+    }
+    return map;
+  }, [routeConfigs]);
+
   const [stackNavState, navActionDispatch]: [
     StackNavigationState,
     React.Dispatch<NavigationAction>,
@@ -73,16 +81,12 @@ export function StackContainer({ routeConfigs }: StackContainerProps) {
             setRouteOptions: navMethods.setRouteOptions,
           };
 
-          const matchingConfig = routeConfigs.find(
-            config => config.name === name,
-          );
-          if (!matchingConfig) {
+          const Component = componentsByName.get(name);
+          if (!Component) {
             throw new Error(
               `[Stack] No config matches the "${name}" route name`,
             );
           }
-
-          const Component = matchingConfig.Component;
 
           return (
             <Stack.Screen
--- a/apps/src/shared/gamma/containers/stack/StackContainer.tsx
+++ b/apps/src/shared/gamma/containers/stack/StackContainer.tsx
@@ -62,7 +62,7 @@ export function StackContainer({ routeConfigs }: StackContainerProps) {
   return (
     <Stack.Host ref={hostRef}>
       {stackNavState.stack.map(
-        ({ Component, options: { headerConfig, ...options }, activityMode, routeKey }) => {
+        ({ options: { headerConfig, ...options }, activityMode, routeKey, name }) => {
           const stackNavigationContext: StackNavigationContextPayload = {
             routeKey,
             routeOptions: { ...options },
@@ -72,6 +72,17 @@ export function StackContainer({ routeConfigs }: StackContainerProps) {
             batch: navMethods.batchAction,
             setRouteOptions: navMethods.setRouteOptions,
           };
+
+          const matchingConfig = routeConfigs.find(
+            config => config.name === name,
+          );
+          if (!matchingConfig) {
+            throw new Error(
+              `[Stack] No config matches the "${name}" route name`,
+            );
+          }
+
+          const Component = matchingConfig.Component;
 
           return (
             <Stack.Screen
--- a/apps/src/shared/gamma/containers/stack/StackContainer.tsx
+++ b/apps/src/shared/gamma/containers/stack/StackContainer.tsx
@@ -21,17 +21,12 @@ import {
   useRenderDebugInfo,
 } from 'react-native-screens/private';
 import { useParentNavigationEffect } from './hooks/useParentNavigationEffect';
+import { useComponentsByName } from '../shared/use-components-by-name';
 
 export function StackContainer({ routeConfigs }: StackContainerProps) {
   useSanitizeRouteConfigs(routeConfigs);
 
-  const componentsByName = React.useMemo(() => {
-    const map = new Map<string, StackRouteConfig['Component']>();
-    for (const config of routeConfigs) {
-      map.set(config.name, config.Component);
-    }
-    return map;
-  }, [routeConfigs]);
+  const componentsByName = useComponentsByName(routeConfigs);
 
   const [stackNavState, navActionDispatch]: [
     StackNavigationState,
--- a/apps/src/shared/gamma/containers/stack/StackContainer.types.tsx
+++ b/apps/src/shared/gamma/containers/stack/StackContainer.types.tsx
@@ -22,7 +22,7 @@ export type StackRouteConfig = {
   options: StackRouteOptions;
 };
 
-export type StackRoute = StackRouteConfig & {
+export type StackRoute = Omit<StackRouteConfig, 'Component'> & {
   activityMode: StackScreenProps['activityMode'];
   routeKey: StackScreenProps['screenKey'];
   isMarkedForDismissal: Boolean; // whether this route is during or after dismissal process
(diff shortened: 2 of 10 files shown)


t0maboro:
For stack v5 tests, I'm observing that fast refresh is resetting the stack to the initial route. I'm on this commit: ced4f0c50
<video src="https://github.com/user-attachments/assets/fcdc6188-1a7b-4fd7-899d-823dd5ba1416" />

apps/src/shared/gamma/containers/stack/reducer.tsx
```diff
@@ -271,8 +277,10 @@ function createRouteFromConfig(
   config: StackRouteConfig,
   activityMode: StackScreenActivityMode = 'detached',
 ): StackRoute {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const { Component, ...rest } = config;
```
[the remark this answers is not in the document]
t0maboro: I'd leave it in the current form, because it's on the example app's side and we also did the same in other places.

apps/src/shared/gamma/containers/stack/StackContainer.tsx
```diff
@@ -73,6 +78,17 @@ export function StackContainer({ routeConfigs }: StackContainerProps) {
             setRouteOptions: navMethods.setRouteOptions,
           };
 
+          const matchingConfig = routeConfigs.find(
```
t0maboro: I'm not a big fan of calling find for each item separately inside map, would it be possible to prepare a dictionary on StackContainer level and here just lookup for a specific component?

sgaczol: done: https://github.com/software-mansion/react-native-screens/pull/3925/commits/6757f871a837a711e5675ac355821fce4130099f

apps/src/shared/gamma/containers/stack/StackContainer.tsx
```diff
@@ -25,6 +25,14 @@ import { useParentNavigationEffect } from './hooks/useParentNavigationEffect';
 export function StackContainer({ routeConfigs }: StackContainerProps) {
   useSanitizeRouteConfigs(routeConfigs);
 
+  const componentsByName = React.useMemo(() => {
+    const map = new Map<string, StackRouteConfig['Component']>();
+    for (const config of routeConfigs) {
+      map.set(config.name, config.Component);
+    }
+    return map;
+  }, [routeConfigs]);
```
kkafar: Move it to separate helper hook. Similarly to useSanitizeRouteConfigs - you can define it in the bottom of this file. 

If this is reused between different containers (Stack & Tabs) -> then you can move ti to standalone file and share it.

apps/src/shared/gamma/containers/stack/StackContainer.types.tsx
```diff
@@ -17,10 +17,10 @@ export type StackRouteConfig = {
   options: StackRouteOptions;
 };
 
-export type StackRoute = StackRouteConfig & {
+export type StackRoute = Omit<StackRouteConfig, 'Component'> & {
   activityMode: StackScreenProps['activityMode'];
   routeKey: StackScreenProps['screenKey'];
-  isMarkedForDismissal: Boolean, // whether this route is during or after dismissal process
+  isMarkedForDismissal: Boolean; // whether this route is during or after dismissal process
```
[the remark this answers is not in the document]
kkafar: hes right, treat is as a typo and fix it.

apps/src/shared/gamma/containers/tabs/TabsContainer.tsx
```diff
@@ -25,15 +25,18 @@ import { TabsContainerItem } from './TabsContainerItem';
...
 
   useSanitizeRouteConfigs(routeConfigs);
 
+  const componentsByName = React.useMemo(() => {
+    const map = new Map<string, TabRouteConfig['Component']>();
+    for (const config of routeConfigs) {
+      map.set(config.name, config.Component);
+    }
+    return map;
+  }, [routeConfigs]);
```
kkafar: Yep, it is shared. Move it to standalone file and share it between implementations if possible. If you'd have to do some weird things to satisfy type constraints -> then make a hook per container, but definitively move it to a hook.

apps/src/shared/gamma/containers/tabs/TabsContainerItem.types.ts
```diff
@@ -5,5 +5,6 @@ export type TabsContainerItemProps = {
   navMethods: TabsNavigationMethods;
   isSelected: boolean;
   pendingForUpdate: boolean;
+  Component: React.ComponentType;
 }
```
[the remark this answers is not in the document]
kkafar: +1, `import React from 'react';`

apps/src/tests/component-integration-tests/orientation/orientation-stack-in-tabs.tsx
```diff
@@ -62,19 +62,19 @@ function ConfigScreen() {
   );
 }
 
-const STACK_ROUTE_CONFIGS: StackRouteConfig[] = [
+export const STACK_ROUTE_CONFIGS: StackRouteConfig[] = [
```
kkafar: Do we need to export that, or only components? 

I'm aware that we said to literally export everything during the meeting, but realising now that I had hidden assumption and thought only of components.

apps/src/tests/issue-tests/Test2809/index.tsx
kkafar, line 1: Do we need to apply these exports rules also to places where we had no problems? What do you think? 

cc @t0maboro?

eslint-plugin-local-rules/index.js
kkafar, line 1: It would be nice to not make top-level directory out of it. Is it possible to put it inside `apps`?

kkafar (commented):
I haven't gone through each test one by one, but I've left few general remarks below. Please answer them.

apps/src/tests/issue-tests/Test2809/index.tsx
t0maboro: if examples touching stack v4 are working fine, we should consider applying rules only to specific directories

apps/src/tests/component-integration-tests/orientation/orientation-stack-in-tabs.tsx
sgaczol: Ok, I think that exporting components is all we need there, so I'll change it

eslint-plugin-local-rules/index.js
t0maboro: explicitly disabling in top-level .eslintrc and overriding that rule in the nested directory `.eslintrc` should work, we may consider adding it deeper, even for SFTs and CITs only, because issue tests won't be actively updated once written. wdyt?

apps/src/shared/gamma/containers/stack/StackContainer.tsx
t0maboro: +1 for generic helper

apps/src/tests/issue-tests/Test2809/index.tsx
sgaczol: It makes sense, we need to refactor `issue-tests` then, so tests regarding different components are in different directories to implement such approach. However I think that such change should be added in a separate PR and if so, an addition of a new ESLint rule should be moved to that PR. What do you think?

eslint-plugin-local-rules/index.js
sgaczol: The problem is that issue tests regarding Tabs or Stackv5 created in the future may seem to be broken, if we don't apply that rule to at least some of the issue tests. We could divide files in `issue-tests` into a few directories and apply the rule only to some of them.

apps/src/shared/gamma/containers/tabs/TabsContainerItem.types.ts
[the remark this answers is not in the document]
sgaczol: done: https://github.com/software-mansion/react-native-screens/pull/3925/commits/047e2d1a6aa7756e12eeaaa0a7f2721cf72abaf0

apps/src/shared/gamma/containers/stack/StackContainer.types.tsx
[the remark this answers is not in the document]
sgaczol: done: https://github.com/software-mansion/react-native-screens/pull/3925/commits/047e2d1a6aa7756e12eeaaa0a7f2721cf72abaf0

kligarski (approved):
I haven't checked the runtime but it looks good.

Remember to update the PR description.

t0maboro (approved):

PR #3930, opened 2026-04-23
chore(AI): upstream Claude.md

kkafar:
Checking into VCS, so that we have a common base to develop.


kligarski (approved):
Looks sensible. What do you think about using `AGENTS.md` and symlinking it as `CLAUDE.md`?

CLAUDE.md
```diff
@@ -0,0 +1,50 @@
...
+- `./FabricExample/` - contains an example application, we use it to showcase the library
+  capabilities and test the library. It is not published as a part of the package.
+
+- `./apps/` - extracted JS (react-native) code of the `FabricExample` application;
+  this is done to share the code with other example applications in the repository.
+
+- `./react-navigation/` - this is a Git sub-module of one of downstream libraries providing
+  complete navigation solution. It is not a part of the library.
+
+## Code conventions
```
t0maboro: Shouldn't we mention conventions for the native development?

t0maboro (approved):

CLAUDE.md
kkafar: Maybe we should. But I think it should be done separately, outside of this PR.

PR #3931, opened 2026-04-23
feat(iOS): Enable synchronous screen shadow state updates by default

kmichalikk:
## Description

This PR changes the default value for `synchronousScreenUpdatesEnabled` flag to `true`

## Changes

Updated default values for the flag in native hierarchy & JS.

N/A

## Test plan

See https://github.com/software-mansion/react-native-screens/pull/3282

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreen.mm
+++ b/ios/RNSScreen.mm
@@ -103,6 +103,7 @@ struct ContentWrapperBox {
   _sheetsScrollView = nil;
   _sheetContentHeight = 0.0;
   _markedForUnmountInCurrentTransaction = NO;
+  _synchronousShadowStateUpdatesEnabled = YES;
 }
 
 + (RNSViewInteractionManager *)viewInteractionManagerInstance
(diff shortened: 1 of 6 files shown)


ios/RNSScreen.mm
```diff
@@ -103,6 +103,7 @@ - (void)initCommonProps
   _sheetsScrollView = nil;
   _sheetContentHeight = 0.0;
   _markedForUnmountInCurrentTransaction = NO;
+  _synchronousShadowStateUpdatesEnabled = YES;
```
t0maboro, line 106: nit: in HeaderConfig we set that directly in `initWithFrame` https://github.com/software-mansion/react-native-screens/pull/3932/changes maybe we should keep the consistency across components (I'm aware that other variables doesn't have a single convention either, so you can merge this in the current form)

t0maboro (approved):

kkafar (approved):

ios/RNSScreen.mm
kmichalikk: I'm going the other way around, moving the setter to initProps in #3932

PR #3936, opened 2026-04-24
chore(test): test-tabs-events screen with scenario

LKuchno:
## Description
Adds a test scenario for tab lifecycle events - onWillAppear, onDidAppear, onWillDisappear, and onDidDisappear. Each event is surfaced as a color-coded toast so that firing order can be verified visually on both platforms.

The scenario covers the happy-path transition sequence (which differs between iOS and Android), the edge case of re-tapping the already-active tab (no events should fire), rapid switching between tabs, and a full round-trip that produces a deterministic total toast count.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1177

## Changes

**test-tabs-events/index.tsx**
New test screen with three tabs sharing a single TabScreen component. Each tab registers all four lifecycle callbacks via makeCallbacks, which pushes a color-coded toast per event using ToastProvider. 

**test-tabs-events/scenario.md**
Manual test scenario with five sections: baseline initial appearance, three happy-path tab transitions, re-tapping the active tab (no events expected), rapid switching, and a full round-trip. D

**single-feature-tests/tabs/index.ts**
Registered testTabsEvents in the tabs scenario group so the screen is reachable from the test navigator.


apps/src/tests/single-feature-tests/tabs/test-tabs-events/scenario.md
```diff
@@ -0,0 +1,121 @@
...
+
+2. Tap **Tab B** in the tab bar.
+
+- [ ] Expected: The content area switches to show "TabB". Four toast
+  notifications appear in order specific platform:
+  - `TabB: onWillAppear` (green background)
+  - `TabA: onWillDisappear` (light navy background)
+  - `TabB: onDidAppear` (light blue background)
+  - `TabA: onDidDisappear` (dark navy background)
+
```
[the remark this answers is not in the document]
LKuchno: https://github.com/software-mansion/react-native-screens/pull/3936/changes/d6891d7a81676d2160533df266500e79ef81b1dc#diff-c5cb399558d7360315af7b772609bee53983a570e6c86dc43d95649f4b4bcefc

apps/src/tests/single-feature-tests/tabs/test-tabs-events/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+  DEFAULT_TAB_ROUTE_OPTIONS,
+  useTabsNavigationContext,
+} from '@apps/shared/gamma/containers/tabs';
+import { CenteredLayoutView } from '@apps/shared/CenteredLayoutView';
+import { ToastProvider, useToast } from '@apps/shared/';
+import { Colors } from '@apps/shared/styling';
+
+const scenarioDescription: ScenarioDescription = {
+  name: 'Tabs lifecycle events',
+  key: 'test-tabs-events',
```
kkafar: This test does not check all the events. I'd suggest adding a quantifier-part here: `test-tabs-native-view-lifecycle-events` or just `test-tabs-lifecycle-events`. `test-tabs-events` suggest too broad scope.

apps/src/tests/single-feature-tests/tabs/test-tabs-lifecycle-events/scenario.md
```diff
@@ -0,0 +1,149 @@
...
+  1. `TabX: onWillDisappear`
+  2. `TabX: onDidDisappear`
+  3. `TabY: onWillAppear`
+  4. `TabY: onDidAppear`
+- Toasts stack and dismiss automatically. To dismiss a toast manually,
+  tap it. Toast background colors by event type:
+  `onWillAppear` — green, `onWillDisappear` — light navy,
+  `onDidAppear` — light blue, `onDidDisappear` — dark navy.
+- Re-tapping the currently active tab must not fire any lifecycle events.
+
```
kkafar, line 42: Thats problematic.

I believe this is how it works currently, but on Android I believe this might depend on exact transition type. Something to be researched. 

I wonder what we should do right now. Maybe let's just proceed, have this tested and described. Later, if we ever decide to modify the event order here -> we will just update the test.

kkafar (commented):
Just a naming suggestion. 

Looks good overall. Thank you.

apps/src/tests/single-feature-tests/tabs/test-tabs-lifecycle-events/scenario.md
LKuchno: I can provide list of expected event without mention in which order they should appear, maybe just note that onWill% event for particular tab should appear before onDid% - what do you think?
I had problem with this part as well. First is stability of it second clearness during test execution.

LKuchno: After discussion I keep scenario in current shape.

apps/src/tests/single-feature-tests/tabs/test-tabs-events/index.tsx
LKuchno: https://github.com/software-mansion/react-native-screens/pull/3936/changes/077b4ab6c0bdd152a626f97a76870eaf90b635d7

apps/src/tests/single-feature-tests/tabs/index.ts
```diff
@@ -30,6 +31,7 @@ const scenarios = {
   TestTabsStaleStateUpdateRejection,
   TestTabsTabBarMinimizeBehavior,
   TestTabsTabBarControllerMode,
+  TestTabsEvents,
```
kkafar: You've changed the test name, but you also need to change the import symbol name here. It's still `TestTabsEvent`. 

This is exactly why I don't like default exports. If this had been named export, it'd be renamed automatically by LSP.

kkafar (changes requested):

apps/src/tests/single-feature-tests/tabs/index.ts
LKuchno: https://github.com/software-mansion/react-native-screens/pull/3936/changes/3cbad3d15cd98699a3218fc284718b495d5ef1ff

kkafar (approved):
Looks good. Thanks.

PR #3937, opened 2026-04-24
chore(test): e2e test-tabs-lifecycle-events

LKuchno:
## Description

Updates the test screen component and manual scenario. Adds a Detox suite for tab lifecycle events - onWillAppear, onDidAppear, onWillDisappear, onDidDisappear. Events are surfaced as numbered, color-coded toasts so that both firing order and identity can be asserted programmatically.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1123?issue=software-mansion%7Creact-native-screens-labs%7C1183

## Changes

**e2e/single-feature-tests/tabs/test-tabs-events.e2e.ts**
Detox suite with 5 it blocks. A shared dismissToast helper waits for a toast by label and taps it. Toast labels carry numeric prefixes (1. TabA: onWillAppear) so assertions are order-sensitive. Event order assertions are branched by device.getPlatform() to account for the differing iOS / Android firing sequence. The re-tap test is Android only.

**test-tabs-events/index.tsx**
Added tabBarItemTestID to each route config (tab-a-item, tab-b-item, tab-c-item) so tab bar items are directly addressable by the E2E suite. Added testID to the content container of TabScreen, keyed by route as tabContent-{routeKey}, used to assert which tab is currently visible.

**test-tabs-events/scenario.md**
Updated E2E status from "ongoing research" to "partially automated" with notes on what is not automated.


LKuchno:
Android tests are failing because toasts and the Tabs screen list are displayed 'under' the system navigation bar on older Android versions. This test is waiting for a fix or a workaround.

FabricExample/e2e/single-feature-tests/tabs/test-tabs-lifecycle-events.e2e.ts
```diff
@@ -0,0 +1,95 @@
+import { device, expect, element, by } from 'detox';
+import { selectSingleFeatureTestsScreen } from '../../e2e-utils';
+
+async function tapTab(label: string) {
+  if (device.getPlatform() === 'android') {
+    await element(by.text(`Tab ${label}`)).tap();
```
kkafar: doesn't tapping by id work on Android? Is this something we need to work on? I believe @kmichalikk added the testID for tabs on Android also.

FabricExample/e2e/single-feature-tests/tabs/test-tabs-events.e2e.ts
```diff
@@ -0,0 +1,95 @@
...
+  await waitFor(element(by.label(message)))
+    .toBeVisible()
+    .withTimeout(3000);
+  await element(by.label(message)).tap();
+}
+
+describe('Tabs lifecycle events', () => {
+  beforeAll(async () => {
+    await device.reloadReactNative();
+    await selectSingleFeatureTestsScreen('Tabs', 'test-tabs-events');
```
kkafar: If we change name of the test, this will need to be updated.

kkafar (commented):
Looks good. But let's first land #3936 and update this PR before we land it.

FabricExample/e2e/single-feature-tests/tabs/test-tabs-events.e2e.ts
LKuchno: https://github.com/software-mansion/react-native-screens/pull/3937/changes/9868f42d5d0fa0f6f20ddb0bd3de03913a491c12

FabricExample/e2e/single-feature-tests/tabs/test-tabs-lifecycle-events.e2e.ts
LKuchno: When I was writing this test I think there was problem with it, but now it works correctly.
I've changed this part of code to tap element by testID. 
https://github.com/software-mansion/react-native-screens/pull/3937/changes/06824d061255cb13ac3888b9888f7ac964ab6165

kkafar (approved):

PR #3949, opened 2026-04-28
refactor(tabs): replace isNativeAction with actionOrigin enum

kkafar:
## Description

Replaces the boolean `isNativeAction` field on the `onTabSelected` event with a three-valued `actionOrigin` discriminator (`'user' | 'programmatic-js' | 'implicit'`).

Important thing to note here is that this change also enables us to add more action origins in the future, such as native-programmatic actions.

One thing I'm not proud of is that the React model details are leaked into the `TabsContainer` (`TabsContainer` / `RNSTabBarController` on Android & iOS respectively). It now recognizes different sources including the `programmatic-js`.
This is not great, but it shouldn't hinder us later when attempting to test it natively.

The boolean was lossy: it conflated JS-driven updates with implicit UIKit-driven selection changes (e.g. iPad horizontal size-class transitions where UIKit reshuffles the selected tab as a side effect). Consumers that needed to react to JS-originated transitions specifically had no way to distinguish them. Promoting the field to an enum makes the actor explicit at the API boundary; on iOS it also lets us collapse the previous internal `RNSTabsNavigationStateUpdateSource` taxonomy into a single shared enum (`RNSTabsActionOrigin`).

## Changes

- Public TS API: `isNativeAction: boolean` → `actionOrigin: 'user' | 'programmatic-js' | 'implicit'` on `TabSelectedEvent` (and the matching codegen specs for iOS / Android native components).
- iOS:
  - New `RNSTabsActionOrigin` enum (`User`, `ProgrammaticJs`, `Implicit`) declared in `RNSTabsNavigationState.h`.
  - Removed the now-redundant `RNSTabsNavigationStateUpdateSource` enum; `progressNavigationState:withSource:` becomes `progressNavigationState:withOrigin:`.
  - `RNSTabsNavigationStateUpdateContext.isNativeAction` → `actionOrigin`; propagated through `OnTabSelectedPayload` and the event emitter, with an Obj-C → codegen-enum conversion in `RNSConversions-Tabs.mm`.
  - The implicit reconciliation path (More-controller size-class transition) now reports `Implicit` instead of being lumped into `isNativeAction = YES`.
- Android:
  - New `TabsActionOrigin` enum (`USER`, `PROGRAMMATIC_JS`) under `gamma.tabs.container`. `'implicit'` is iOS-only at the moment and not currently produced on Android — documented in the enum's KDoc.
  - Threaded through `TabsContainerDelegate.onNavStateUpdate`, `TabsHostEventEmitter`, and `TabsHostTabSelectedEvent` (event payload key `actionOrigin`, serialized via `enum.toString()`).

## Test plan

No new automated tests; this is a rename/refactor of an existing public field with no behavioral change on the user/JS paths. The implicit path on iOS now reports a distinct origin where it previously reported `isNativeAction = YES`, so manual verification on the affected platforms:

- iOS — tab bar tap → `actionOrigin === 'user'`.
- iOS — JS-driven `navStateRequest` update → `actionOrigin === 'programmatic-js'`.
- iOS — iPad horizontal size-class transition that collapses/expands the More controller → `actionOrigin === 'implicit'` (previously `isNativeAction: true`, now distinguishable).
- Android — tab bar tap → `actionOrigin === 'user'`; JS-driven update → `actionOrigin === 'programmatic-js'`.

Reproduction lives in `FabricExample` Tabs test screens; log `e.nativeEvent.actionOrigin` from `onTabSelected`.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsActionOrigin.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsActionOrigin.kt
@@ -1,25 +1,25 @@
 package com.swmansion.rnscreens.gamma.tabs.container
 
-import com.swmansion.rnscreens.gamma.tabs.container.TabsActionOrigin.JS
+import com.swmansion.rnscreens.gamma.tabs.container.TabsActionOrigin.PROGRAMMATIC_JS
 import com.swmansion.rnscreens.gamma.tabs.container.TabsActionOrigin.USER
 
 /**
  * Origin (actor) that requested a tab transition. Mirrors the public `actionOrigin` event field.
  *
  * - [USER] — direct native UI interaction (tab bar tap).
- * - [JS] — JS-initiated request delivered via the `navStateRequest` prop.
+ * - [PROGRAMMATIC_JS] — JS-initiated request delivered via the `navStateRequest` prop.
  *
  * The `implicit` origin defined on the public TS API is iOS-only at the moment;
  * Android does not currently produce it.
  */
 enum class TabsActionOrigin {
     USER,
-    JS,
+    PROGRAMMATIC_JS,
     ;
 
     override fun toString(): String =
         when (this) {
             USER -> "user"
-            JS -> "js"
+            PROGRAMMATIC_JS -> "programmatic-js"
         }
 }
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsActionOrigin.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsActionOrigin.kt
@@ -1,8 +1,5 @@
 package com.swmansion.rnscreens.gamma.tabs.container
 
-import com.swmansion.rnscreens.gamma.tabs.container.TabsActionOrigin.PROGRAMMATIC_JS
-import com.swmansion.rnscreens.gamma.tabs.container.TabsActionOrigin.USER
-
 /**
  * Origin (actor) that requested a tab transition. Mirrors the public `actionOrigin` event field.
  *
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsActionOrigin.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsActionOrigin.kt
@@ -1,0 +1,25 @@
+package com.swmansion.rnscreens.gamma.tabs.container
+
+import com.swmansion.rnscreens.gamma.tabs.container.TabsActionOrigin.JS
+import com.swmansion.rnscreens.gamma.tabs.container.TabsActionOrigin.USER
+
+/**
+ * Origin (actor) that requested a tab transition. Mirrors the public `actionOrigin` event field.
+ *
+ * - [USER] — direct native UI interaction (tab bar tap).
+ * - [JS] — JS-initiated request delivered via the `navStateRequest` prop.
+ *
+ * The `implicit` origin defined on the public TS API is iOS-only at the moment;
+ * Android does not currently produce it.
+ */
+enum class TabsActionOrigin {
+    USER,
+    JS,
+    ;
+
+    override fun toString(): String =
+        when (this) {
+            USER -> "user"
+            JS -> "js"
+        }
+}
--- a/ios/conversion/RNSConversions-Tabs.mm
+++ b/ios/conversion/RNSConversions-Tabs.mm
@@ -241,7 +241,7 @@ react::RNSTabsHostIOSEventEmitter::OnTabSelectedActionOrigin RNSOnTabSelectedAct
     case RNSTabsActionOriginUser:
       return User;
     case RNSTabsActionOriginProgrammaticJs:
-      return Js;
+      return ProgrammaticJs;
     case RNSTabsActionOriginImplicit:
       return Implicit;
   }
--- a/ios/conversion/RNSConversions-Tabs.mm
+++ b/ios/conversion/RNSConversions-Tabs.mm
@@ -244,7 +244,10 @@ react::RNSTabsHostIOSEventEmitter::OnTabSelectedActionOrigin RNSOnTabSelectedAct
       return ProgrammaticJs;
     case RNSTabsActionOriginImplicit:
       return Implicit;
+    default:
+      RCTLogError(@"[RNScreens] Unexpected actionOrigin: %ld", actionOrigin);
   }
+  return User;
 }
 
 RNSTabsIconType RNSTabsIconTypeFromIcon(react::RNSTabsScreenIOSIconType iconType)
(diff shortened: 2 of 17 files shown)


src/components/tabs/host/TabsHost.types.ts
```diff
@@ -62,10 +62,16 @@ export type TabSelectedEvent = {
...
-   * False in case the event is a result of JS-driven update. True otherwise, e.g. in case of user action (tap)
-   * or implicit UIKit action (app resize, orientation change, etc.).
+   * @summary Origin (actor) that requested this tab transition.
+   *
+   * @description
+   * - `'user'` — direct native UI interaction (e.g. tab bar tap, iOS tab drag-and-drop).
+   * - `'programmatic-js'` — JS-initiated request delivered via the `navStateRequest` prop.
+   * - `'implicit'` — platform side effect not attributable to an explicit actor
+   *   (e.g. UIKit reshuffling the selection during a horizontal size-class transition on iPad).
+   *   Currently only emitted on iOS.
```
kligarski:
```suggestion
   * - `user` — direct native UI interaction (e.g. tab bar tap, iOS tab drag-and-drop).
   * - `programmatic-js` — JS-initiated request delivered via the `navStateRequest` prop.
   * - `implicit` — platform side effect not attributable to an explicit actor
   *   (e.g. UIKit reshuffling the selection during a horizontal size-class transition on iPad).
   *   Currently only emitted on iOS.
```

nit: see section "Formatting Enums" from https://github.com/software-mansion/react-native-screens-labs/discussions/313.

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -236,7 +236,7 @@ - (BOOL)updateSelectedViewControllerTo:(nullable UIViewController *)nextSelected
       ![NSString rnscreens_isBlankOrNull:screenKey],
       @"[RNScreens] The screenKey MUST NOT be null if the view controller is not null");
 
-  [self progressNavigationState:screenKey withSource:RNSTabsNavigationStateUpdateSourceExternal];
+  [self progressNavigationState:screenKey withOrigin:RNSTabsActionOriginProgrammaticJs];
```
kligarski, line 239: Shouldn't we highlight in method signature and/or comment that it should be used only for update from JS? `updateSelectedViewControllerTo` sounds general-purpose. But we should rename the entire path to here so maybe for now we can stick to the current version.

kligarski (approved):

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsActionOrigin.kt
```diff
@@ -0,0 +1,25 @@
+package com.swmansion.rnscreens.gamma.tabs.container
+
+import com.swmansion.rnscreens.gamma.tabs.container.TabsActionOrigin.PROGRAMMATIC_JS
+import com.swmansion.rnscreens.gamma.tabs.container.TabsActionOrigin.USER
```
t0maboro: not needed, this file defines these

src/components/tabs/host/TabsHost.types.ts
```diff
@@ -62,10 +62,16 @@ export type TabSelectedEvent = {
   /** Whether the selection triggered a special effect (e.g. scroll-to-top on repeated selection). */
   hasTriggeredSpecialEffect: boolean;
   /**
-   * False in case the event is a result of JS-driven update. True otherwise, e.g. in case of user action (tap)
-   * or implicit UIKit action (app resize, orientation change, etc.).
+   * @summary Origin (actor) that requested this tab transition.
+   *
+   * @description
+   * - `'user'` — direct native UI interaction (e.g. tab bar tap, iOS tab drag-and-drop).
```
t0maboro: I don't think I like `user` here. My impression is that it's descirbing WHO triggered that, not HOW it happened, whereas others focus on HOW. wdyt about renaming it to `user-interaction`?

t0maboro (approved):

ios/conversion/RNSConversions-Tabs.mm
```diff
@@ -233,6 +233,20 @@ UITabBarControllerMode UITabBarControllerModeFromRNSTabBarControllerMode(RNSTabB
...
+{
+  using enum facebook::react::RNSTabsHostIOSEventEmitter::OnTabSelectedActionOrigin;
+  switch (actionOrigin) {
+    case RNSTabsActionOriginUser:
+      return User;
+    case RNSTabsActionOriginProgrammaticJs:
+      return ProgrammaticJs;
+    case RNSTabsActionOriginImplicit:
+      return Implicit;
+  }
```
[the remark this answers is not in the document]
kkafar: Done

src/components/tabs/host/TabsHost.types.ts
kkafar: You're right. I will change that, but tbh I like what the Claude generated more than our current convention 😄

kkafar: https://github.com/software-mansion/react-native-screens/pull/3949/commits/c10e5ad296cfe9c124e36517df05890efdf848f9

ios/tabs/host/RNSTabBarController.mm
kkafar: This code path is not JS specific. It's for any "external container update". In one of follow-up PRs it'll be also used for programmatic-native actions. 

Another thing is I don't want to leak additional details of RN world into the container.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsActionOrigin.kt
kkafar: At the first glance I agree, but when I run "optimise imports" action in this file - it does nothing - it qualifies these as used. 

I've removed these and the app still builds successfully -> leaving them removed, as you suggested.

kkafar: https://github.com/software-mansion/react-native-screens/pull/3949/commits/7e5a868f9a74a89fca14e49525d984d8ada13fa7

src/components/tabs/host/TabsHost.types.ts
kkafar: Good remark, but I see it the other way around. All of those should describe WHO (the origin) triggered that, and not how. But I don't have better naming ideas for the other enum variants. Do you happen to have any suggestions? 

I'll proceed here & refactor it later if we come up with some nice naming here.

t0maboro: depends on the interpretation tbh, I see your point, my 1st impression is that it diverges, but I see your point, the only option that came to my mind so far was `user-interaction`, but I feel that's not the name I'd like to have in a public API, for now I think I'm okay with the current shape

PR #3953, opened 2026-04-29
chore(test): e2e and scenario for tabs specialEffects scrollToTop

LKuchno:
## Description

This PR adds a new Detox e2e test suite that validates the `specialEffects.repeatedTabSelection.scrollToTop` tab property and update the scenario _e2e test_ section. The Detox e2e tests covers all manual steps.
New functions were added to e2e-utils.ts file to enable selecting already open tab on iOS26.2 and another one to allow run test just for android.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1190

## Changes

- **FabricExample/e2e**: 
Added Detox e2e test suite `test-tabs-special-effects.e2e.ts` covering `scrollToTop: true`, `scrollToTop: false`, and default (no `specialEffects`) behavior; 
Added new functions into e2e-utils: `describeIfAndroid` and `forceTapByLabeliOS ` 
- **apps/tests/tabs**: 
Update`test-tabs-special-effects` scenario screen with e2e covarage information and index with testID values
Update index with testIDs needed in e2e test

### DIFF
--- a/FabricExample/e2e/e2e-utils.ts
+++ b/FabricExample/e2e/e2e-utils.ts
@@ -75,15 +75,11 @@ export async function getElementAttributes(
   return attrs as ElementAttributes;
 }
 
-export async function tapSelectedTab(testLabel: string) {
-  if (device.getPlatform() === 'ios') {
-    const elementAttributes = await getElementAttributes(testLabel);
-    const { x, y, width, height } = elementAttributes.frame;
-    await device.tap({
-      x: x + width / 2,
-      y: y + height / 2,
-    });
-  } else {
-    await element(by.label(testLabel)).tap();
-  }
+export async function forceTapByLabeliOS(testLabel: string) {
+  const elementAttributes = await getElementAttributes(testLabel);
+  const { x, y, width, height } = elementAttributes.frame;
+  await device.tap({
+    x: x + width / 2,
+    y: y + height / 2,
+  });
 }
--- a/FabricExample/e2e/e2e-utils.ts
+++ b/FabricExample/e2e/e2e-utils.ts
@@ -82,6 +82,10 @@ export async function getElementAttributes(
   return attrs as ElementAttributes;
 }
 
+/**
+ * Performs a coordinate-based tap on iOS to interact with an element that may be 
+ * obstructed by other UI layers, bypassing Detox's default visibility checks.
+ */
 export async function forceTapByLabeliOS(testLabel: string) {
   const elementAttributes = await getElementAttributes(testLabel);
   const { x, y, width, height } = elementAttributes.frame;
--- a/FabricExample/e2e/e2e-utils.ts
+++ b/FabricExample/e2e/e2e-utils.ts
@@ -1,7 +1,11 @@
 import { device, expect, element, by } from 'detox';
+import { AndroidElementAttributes, IosElementAttributes } from 'detox/detox';
 
 export const describeIfiOS =
   device.getPlatform() === 'ios' ? describe : describe.skip;
+
+export const describeIfAndroid =
+  device.getPlatform() === 'android' ? describe : describe.skip;
 
 async function scrollUntilVisible(id: string, scrollViewId: string) {
   await waitFor(element(by.id(id)))
@@ -61,4 +65,25 @@ export async function selectSingleFeatureTestsScreen(
     `${scenarioGroup}-scenarios-scrollview`,
   );
   await element(by.id(`${screenKey}`)).tap();
+}
+type ElementAttributes = IosElementAttributes | AndroidElementAttributes;
+
+export async function getElementAttributes(
+  testLabel: string,
+): Promise<ElementAttributes> {
+  const attrs = await element(by.label(testLabel)).getAttributes();
+  return attrs as ElementAttributes;
+}
+
+export async function tapSelectedTab(testLabel: string) {
+  if (device.getPlatform() === 'ios') {
+    const elementAttributes = await getElementAttributes(testLabel);
+    const { x, y, width, height } = elementAttributes.frame;
+    await device.tap({
+      x: x + width / 2,
+      y: y + height / 2,
+    });
+  } else {
+    await element(by.label(testLabel)).tap();
+  }
 }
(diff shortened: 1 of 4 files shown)


FabricExample/e2e/e2e-utils.ts
```diff
@@ -1,8 +1,12 @@
 import { device, expect, element, by } from 'detox';
+import { AndroidElementAttributes, IosElementAttributes } from 'detox/detox';
 
 export const describeIfiOS =
   device.getPlatform() === 'ios' ? describe : describe.skip;
 
+export const describeIfAndroid =
+  device.getPlatform() === 'android' ? describe : describe.skip;
```
kligarski, line 8: In theory we don't use it now after adapting the test to iOS but I would keep it here - I think that we're going to use it soon.

FabricExample/e2e/e2e-utils.ts
```diff
@@ -62,3 +66,24 @@ export async function selectSingleFeatureTestsScreen(
...
+    const elementAttributes = await getElementAttributes(testLabel);
+    const { x, y, width, height } = elementAttributes.frame;
+    await device.tap({
+      x: x + width / 2,
+      y: y + height / 2,
+    });
+  } else {
+    await element(by.label(testLabel)).tap();
+  }
+}
```
kligarski: I think that there might be a situation where we will need this trick to click on something that is not fully visible and this may be something other than a tab. 

What I suggest is:
1. Make this util in `e2e-utils.ts` general-purpose for both platforms, something like:

```ts
/**
 * Add a comment here describing when you might need this function (when view is obstructed by another view).
*/
export async function forceTapByLabel(label: string) {
  const elementAttributes = await getElementAttributes(testLabel);
  const { x, y, width, height } = elementAttributes.frame;
  await device.tap({
    x: x + width / 2,
    y: y + height / 2,
  });
}
```

2. For now, create a helper function in `test-tabs-special-effects-scroll-to-top.e2e.ts` for tapping a tab, something like

```ts
/**
 * Add a comment explaining that on iOS 26+ the liquid glass lens obstructs tab bar item with testID and that we don't have any other way to tap on already selected tab.
*/
async function forceSelectTabByLabel(label: string) {
  if (device.getPlatform() === 'ios') {
    await forceTapByLabel(label);
  } else {
    await element(by.label(label)).tap();
  }
}
```

3. In the future if we need to use `forceSelectTabByLabel` in other test we can consider creating separate file with utils for tabs or adding `forceSelectTabByLabel` to `e2e-utils.ts` as well.

FabricExample/e2e/e2e-utils.ts
```diff
@@ -62,3 +66,24 @@ export async function selectSingleFeatureTestsScreen(
...
   await element(by.id(`${screenKey}`)).tap();
 }
+type ElementAttributes = IosElementAttributes | AndroidElementAttributes;
+
+export async function getElementAttributes(
+  testLabel: string,
+): Promise<ElementAttributes> {
+  const attrs = await element(by.label(testLabel)).getAttributes();
+  return attrs as ElementAttributes;
+}
```
kligarski, line 83: `getAttributes()` might return `{ elements }` object if multiple elements match the label. It shouldn't happen but it might happen so I would throw an error here if that's the case so that we know that something has gone wrong. Something like:


```suggestion
export async function getElementAttributes(
  testLabel: string,
): Promise<ElementAttributes> {
  const attrs = await element(by.label(testLabel)).getAttributes();
  
  if ('elements' in attrs) {
    throw new Error(
      `Multiple elements (${attrs.elements.length}) found for label: "${testLabel}". `
    );
  }
  
  return attrs as ElementAttributes;
}
```

FabricExample/e2e/e2e-utils.ts
```diff
@@ -1,8 +1,12 @@
 import { device, expect, element, by } from 'detox';
+import { AndroidElementAttributes, IosElementAttributes } from 'detox/detox';
```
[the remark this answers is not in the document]
kligarski: I don't think that we can really do anything about it.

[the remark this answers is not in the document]
LKuchno: I was already trying to get rid of it and it didn't work so I kept it in this form.

FabricExample/e2e/e2e-utils.ts
LKuchno: I change it following your suggestion, as forceTabByLabel function in this shape works only for iOS I include platform in it's name.
https://github.com/software-mansion/react-native-screens/pull/3953/changes/fdc7b0af6374b567a34c00ae4589d6a406d35f51

FabricExample/e2e/e2e-utils.ts
```diff
@@ -62,3 +66,27 @@ export async function selectSingleFeatureTestsScreen(
...
+  if ('elements' in attrs) {
+    throw new Error(
+      `Multiple elements (${attrs.elements.length}) found for label: "${testLabel}". `
+    );
+  }
+  
+  return attrs as ElementAttributes;
+}
+
+export async function forceTapByLabeliOS(testLabel: string) {
```
kligarski, line 89:
```suggestion
/**
 * Performs a coordinate-based tap on iOS to interact with an element that may be 
 * obstructed by other UI layers, bypassing Detox's default visibility checks.
 */
export async function forceTapByLabeliOS(testLabel: string) {
```

FabricExample/e2e/single-feature-tests/tabs/test-tabs-special-effects-scroll-to-top.e2e.ts
```diff
@@ -0,0 +1,87 @@
+import { device, expect, element, by } from 'detox';
+import {
+  selectSingleFeatureTestsScreen,
+  forceTapByLabeliOS,
+} from '../../e2e-utils';
+
+async function forceSelectTabByLabel(label: string) {
```
kligarski, line 11:
```suggestion
/**
 * Selects a tab bar item. On iOS, this uses a forced coordinate tap to 
 * ensure the tab is selected even if it is obstructed by the iOS 26 Liquid Glass lens.
 */
async function forceSelectTabByLabel(label: string) {
```

kligarski (approved):
Looks good.

The last suggestion I have is to add some comments to new util methods so that we know why we needed them in the future.

FabricExample/e2e/single-feature-tests/tabs/test-tabs-special-effects-scroll-to-top.e2e.ts
LKuchno: Thank you 😄

PR #3958, opened 2026-04-30
feat(Split): add RTL support

sgaczol:
## Description

This PR introduces `RTL` support to `SplitHost`.

The approach differs on iOS versions prior to iOS 17 and newer ones.

On < iOS 17 versions, we use `setOverrideTraitCollection(_: forChild:)` on the parent view controller to propagate the layout direction, accordingly to https://github.com/software-mansion/react-native-screens-labs/blob/main-issue-tracker/rfcs/0996-rtl-and-dark-mode.md

There is a specific edge case on those iOS versions: if the layout direction is changed while a column is currently open/visible, the UI becomes inconsistent. The sidebar remains on its original side, while the display mode button moves to the new side. The layout only synchronizes correctly after the column is closed and reopened. I checked this issue and it's a native bug: https://github.com/software-mansion/react-native-screens-labs/issues/1237

On ≥ iOS 17 versions, we use `traitOverrides` API on the view controller.

Closes https://github.com/issues/assigned?issue=software-mansion%7Creact-native-screens-labs%7C1027

Currently blocked by https://github.com/software-mansion/react-native-screens-labs/issues/1429 - it needs to be fixed in order to make a stable test.

## Changes

- add `direction` prop to `SplitHost`
- handle `layoutDirection` on iOS side
- add `test-split-direction.tsx`

|  < iOS 17 |  ≥ iOS 17  |
| :---: | :---: |
| <video src="https://github.com/user-attachments/assets/a60ff404-3f6f-44f0-bdf1-58c31191afd2"></video> | <video src="https://github.com/user-attachments/assets/12dbf516-d2f8-46a1-8f6a-0836b91078e1"></video> |

`inherit` option also works properly for locations with `RTL` direction as a default:

https://github.com/user-attachments/assets/720ed82d-045a-4e60-bbf7-e48accf8d117

## Test plan

Run `test-split-direction`.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/apps/src/tests/single-feature-tests/split/test-split-direction.tsx
+++ b/apps/src/tests/single-feature-tests/split/test-split-direction.tsx
@@ -5,8 +5,8 @@ import { Split, SplitHostDirection } from 'react-native-screens/experimental';
 import { StyleSheet, Text, View, Button } from 'react-native';
 
 const scenarioDescription: ScenarioDescription = {
-  name: 'Prop: layoutDirection',
-  key: 'test-split-layout-direction',
+  name: 'Prop: direction',
+  key: 'test-split-direction',
   details: `
     Test the direction prop in Split component.
   `,
--- a/apps/src/tests/single-feature-tests/split/test-split-direction.tsx
+++ b/apps/src/tests/single-feature-tests/split/test-split-direction.tsx
@@ -1,7 +1,7 @@
 import React, { useState } from 'react';
 import type { ScenarioDescription } from '@apps/tests/shared/helpers';
 import { createScenario } from '@apps/tests/shared/helpers';
-import { Split } from 'react-native-screens/experimental';
+import { Split, SplitHostDirection } from 'react-native-screens/experimental';
 import { StyleSheet, Text, View, Button } from 'react-native';
 
 const scenarioDescription: ScenarioDescription = {
@@ -13,10 +13,8 @@ const scenarioDescription: ScenarioDescription = {
   platforms: ['ios'],
 };
 
-type LayoutDirection = 'ltr' | 'rtl' | 'inherit';
-
 export function App() {
-  const [direction, setDirection] = useState<LayoutDirection>('inherit');
+  const [direction, setDirection] = useState<SplitHostDirection>('inherit');
 
   return (
     <Split.Host direction={direction}>
@@ -28,9 +26,6 @@ export function App() {
         />
       </Split.Column>
       <Split.Column>
-        <ColumnContent columnTitle="Supplementary column" />
-      </Split.Column>
-      <Split.Column>
         <ColumnContent columnTitle="Secondary column" />
       </Split.Column>
     </Split.Host>
@@ -39,8 +34,8 @@ export function App() {
 
 export function ColumnContent(props: {
   columnTitle: string;
-  currentDirection?: LayoutDirection;
-  onChangeDirection?: (dir: LayoutDirection) => void;
+  currentDirection?: SplitHostDirection;
+  onChangeDirection?: (dir: SplitHostDirection) => void;
 }) {
   return (
     <View style={styles.container}>
--- a/ios/gamma/split/RNSSplitHostComponentView.mm
+++ b/ios/gamma/split/RNSSplitHostComponentView.mm
@@ -30,6 +30,7 @@ static const CGFloat epsilon = 1e-6;
   bool _needsSplitSecondaryScreenNavBarUpdate;
   bool _needsSplitDisplayModeUpdate;
   bool _needsSplitOrientationUpdate;
+  bool _needsSplitLayoutDirectionUpdate;
   // We need this information to warn users about dynamic changes to behavior being currently unsupported.
   bool _isShowSecondaryToggleButtonSet;
 }
@@ -53,6 +54,7 @@ static const CGFloat epsilon = 1e-6;
   _needsSplitSecondaryScreenNavBarUpdate = false;
   _needsSplitDisplayModeUpdate = false;
   _needsSplitOrientationUpdate = false;
+  _needsSplitLayoutDirectionUpdate = false;
   _reactSubviews = [NSMutableArray new];
 }
 
@@ -92,6 +94,7 @@ static const CGFloat epsilon = 1e-6;
   _topColumnForCollapsingColumn = UISplitViewControllerColumnPrimary;
 
   _orientation = RNSOrientationInherit;
+  _layoutDirection = UITraitEnvironmentLayoutDirectionUnspecified;
 
   _isShowSecondaryToggleButtonSet = false;
 }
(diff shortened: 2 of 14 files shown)


ios/gamma/split/RNSSplitHostComponentView.mm
```diff
@@ -365,6 +374,16 @@ - (void)requestSplitHostControllerForAppearanceUpdate
...
   }
+
+  if (_needsSplitLayoutDirectionUpdate && _controller != nil) {
+    _needsSplitLayoutDirectionUpdate = false;
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(17_0)
+    if (@available(iOS 17.0, *)) {
+      _controller.traitOverrides.layoutDirection = _layoutDirection;
+    } else
+#endif // RNS_IPHONE_OS_VERSION_AVAILABLE(17_0)
+      [_controller setNeedsLayoutDirectionUpdateBelowIOS17];
```
kligarski: In tabs this is also done in similar way but maybe for split we should keep the clean flow where in updateProps we set only the props and flags, here we set only the flag on controller (so more general `[_controller setNeedsLayoutDirectionUpdate]` and then at the level of appearance applicator we have some `updateLayoutDirection` function that will run `updateLayoutDirectionBelowIOS17` / `updateLayoutDirectionAboveIOS17` (traitOverrides.layoutDirection = _layoutDirection) depending on iOS version. What do you think?

src/components/gamma/split/SplitHost.tsx
```diff
@@ -101,6 +101,7 @@ function SplitHost({ ref, ...props }: SplitHostProps) {
       // This enables us to fully recreate the Split when necessary, ensuring the correct column configuration is always applied.
       key={`columns-${columns.length}-inspectors-${inspectors.length}`}
       {...props}
+      layoutDirection={direction}
       style={styles.container}>
```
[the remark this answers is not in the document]
kligarski: Copilot is right here, I had the same exact bug in Tabs and I've spent too much time trying to figure this out. We should destructure props and keep the rest of the props in `filteredProps` and only spread the `filteredProps`, as is done in e.g. `TabsHost.ios.tsx`.

cc @t0maboro

ios/gamma/split/RNSSplitHostController.swift
```diff
@@ -552,4 +562,14 @@ extension RNSSplitHostController: UISplitViewControllerDelegate {
     }
     return proposedTopColumn
   }
+
+  public override func didMove(toParent parent: UIViewController?) {
```
kligarski: I don't think this should be in `UISplitViewControllerDelegate` extension block. It should probably be in main block with `  // MARK: UIKit callbacks`

kligarski: cc @t0maboro

apps/src/tests/single-feature-tests/split/test-split-direction.tsx
```diff
@@ -0,0 +1,102 @@
...
+        <ColumnContent
+          columnTitle="Primary column"
+          currentDirection={direction}
+          onChangeDirection={setDirection}
+        />
+      </Split.Column>
+      <Split.Column>
+        <ColumnContent columnTitle="Secondary column" />
+      </Split.Column>
+    </Split.Host>
```
kligarski, line 31: From what I know, the SplitView layout can be more complicated, you can have more columns, there is an inspector column. Should we maybe test with a more complicated Split layout to test more cases?

cc @t0maboro

apps/src/tests/single-feature-tests/split/test-split-direction.tsx
```diff
@@ -0,0 +1,102 @@
...
+  );
+}
+
+export function ColumnContent(props: {
```
kligarski, line 35: Why is the column cut off below iOS 26? It doesn't look centered.

Also, there seems to be some kind of bug with the shadow? Do you know about this, @t0maboro?

https://github.com/user-attachments/assets/2e4b8a4d-42f5-4e5b-9b17-158126a5c04c

apps/src/tests/single-feature-tests/split/test-split-direction.tsx
kligarski, line 1: I think that we should probably follow new test structure (separate directory, `scenario.md` file) for new tests.

kligarski, line 1: In this test I think that we're missing checking how our configuration behaves with `react-native`'s `I18nManager`. Downstream implementations will probably use `I18nManager.isRTL` as suggested in the docs and how it's done in `TabsContainer`. I know that we don't have something like `SplitContainer` but we should probably add an option `I18nManager.isRTL` in test in addition to current `inherit`/`ltr`/`rtl`. We should also have a way to configure `forceRTL` and `allowRTL`.

src/components/gamma/split/SplitHost.tsx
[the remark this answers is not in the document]
t0maboro: +1

apps/src/tests/single-feature-tests/split/test-split-direction.tsx
t0maboro: +1, afair changing number of columns should work even in runtime, so even conditional rendering could be possilbe to have a setup with 2 or 3 columns

apps/src/tests/single-feature-tests/split/test-split-direction.tsx
t0maboro: the native offset of this column on iOS < 26 has an offset around -100px on X axis, so that's fine; for shadow, not sure about the origin of this issue, will take a look while reviewing this PR

t0maboro: the shadow is because the Primary column hasn't backgroundColor set, when it's transparent, it samples the color from the content behind, natively it works the same


https://github.com/user-attachments/assets/91cc4912-8075-4457-8a46-d9b5a41288ea

kligarski: It's so weird but okay. Would `SafeAreaView` help with the content being cut off 100px?

t0maboro: Yes

kligarski: So we should probably use it here if we can @sgaczol

ios/gamma/split/RNSSplitHostController.swift
sgaczol: done:
https://github.com/software-mansion/react-native-screens/pull/3958/commits/814ab0508ed7e466a1e1354d98d23bc4f4421322

ios/gamma/split/RNSSplitHostComponentView.mm
sgaczol: It makes sense. However, after change in  https://github.com/software-mansion/react-native-screens/pull/3958/commits/b6470e1f4a2876d45a01ae9fd49b306fa5e62c88, it's controller who detects, which iOS version we're currently at (not the applicator). Such approach seems better for me, especially if we are below iOS17 and the `controller.parent` is still `nil`.

src/components/gamma/split/SplitHost.tsx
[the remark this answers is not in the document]
sgaczol: done:
https://github.com/software-mansion/react-native-screens/pull/3958/commits/e342c38819f1f914602903b4f43778f80439b98d

sgaczol:
PR is currently blocked by https://github.com/software-mansion/react-native-screens-labs/issues/1429 - it needs to be fixed in order to make a stable test.

PR #3959, opened 2026-04-30
fix(Example): add ESLint rule to enforce top-level component exports in SFT and CIT

sgaczol:
## Description

This PR introduces a local ESLint rule `require-top-level-exports` to ensure that all top-level React component declarations in our test suites are exported.

This approach fixes an issue with Fast Refresh where some of the unexported components were being fully remounted instead of seamlessly updated. This behavior is tied to the Metro bundler's specification regarding module boundaries and state preservation. By ensuring all top-level components are explicitly exported, Metro can properly track them, which fixes the remounting issue and allows Fast Refresh to work as intended.

## Changes

- add the `eslint-plugin-local-rules` plugin with `require-top-level-exports` rule
- apply the rule to single-feature and component-integration tests
- update `package.json` and add `yarn.lock` in `apps`

### Before

https://github.com/user-attachments/assets/380a615a-a1be-483e-9c47-c5241ffa3530

### After

https://github.com/user-attachments/assets/25c5ad1c-18dc-413f-a406-c06b89990e0e

## Test plan

Make sure that updating components of single-feature and component-integration tests triggers a correct Fast Refresh.

## Checklist

- [ ] Ensured that CI passes


apps/eslint-plugin-local-rules/require-top-level-exports.js
kkafar, line 1: I just took a peek at this PR, it's not a full review yet. My impression here is that this code is not test-related per se, it should not reside in `apps/src/tests/` directory, `apps/src/` will be fine.

What do you think? 

cc @t0maboro

apps/src/tests/.eslintrc.js
```diff
@@ -0,0 +1,30 @@
...
+  path.join(__dirname, 'eslint-plugin-local-rules'),
+);
+const originalResolve = Module._resolveFilename;
+Module._resolveFilename = function (request, ...args) {
+  if (request === 'eslint-plugin-local-rules') {
+    return pluginPath;
+  }
+  return originalResolve.call(this, request, ...args);
+};
+
```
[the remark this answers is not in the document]
sgaczol: A `file:` dependency would require adding it to the root `package.json` or creating `package.json` in `apps\`, so I'd skip this one

apps/eslint-plugin-local-rules/require-top-level-exports.js
```diff
@@ -0,0 +1,101 @@
...
+    if (
+      name &&
+      isPascalCase(name) &&
+      decl.init &&
+      (decl.init.type === 'ArrowFunctionExpression' ||
+        decl.init.type === 'FunctionExpression')
+    ) {
+      return name;
+    }
+  }
```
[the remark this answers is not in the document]
sgaczol: I don't think we're ever using this pattern in the future, so I'd skip.

apps/eslint-plugin-local-rules/require-top-level-exports.js
sgaczol: done: https://github.com/software-mansion/react-native-screens/pull/3959/commits/155cb68b7554f8b1eea86436d1dd4a26e97fd3d4

apps/src/tests/.eslintrc.js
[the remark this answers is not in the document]
kkafar: I think the Copilot is right here. Current approach is unnecessarily hacky. I don't see any motivation to complicate the approach like that. 

I see few options here.

First:

Define the plugin in top level of the package `/eslint-plugin-local-rules` and then just simply add it as dev dependency to the example apps. 

Second:

Define the plugin in `/apps/eslint-plugin-local-rules`, add a dev dependency on it to the example apps.

I like the second option (path) a bit more. I know I suggested `/apps/src/...` initially, but taking another look, the plugin is not application code. It's rather utility stuff. 

What do you think?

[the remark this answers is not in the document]
sgaczol: I checked both options and they don't work well as ESLint resolves plugins from the `.eslintrc.js` location, which is `apps/src/tests/` and example apps' `node_modules/` aren't on that resolution path.

I think the cleanest solution is to define the plugin in `apps/eslint-plugin-local-rules` and create a `package.json` in `apps/` with it as a dev dependency. Wdyt?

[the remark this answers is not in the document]
t0maboro: Ideally, I'd try to keep this on the app-level scope. My findings
- ESLint resolves plugins starting from the directory of the `eslintrc` that names the plugin, in our case: `apps/src/tests/`
- it's walking up through ancestor `node_modules` - FabricExample `node_modules` are ignored, so installing the dep there can never satisfy plugin under `apps/src/tests/`
- The plugin should live in node_modules of `apps` or some ancestor higher, in our case, the root, but adding it to the root should be done as a last resort.

I'm okay with making `apps` a module that will be imported by FabricExample/TVOSExample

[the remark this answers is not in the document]
kkafar: That's sounds fine. Let's turn the apps into a separate module with a list of peer dependencies that are provided by the actual example app. 

I'd recommend doing that in an independent PR, and rebasing these changes on top of it later.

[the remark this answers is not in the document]
t0maboro: +1 ^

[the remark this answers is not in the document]
sgaczol: `apps` has been turned into a separate module in this PR: https://github.com/software-mansion/react-native-screens/pull/4052

As of commit https://github.com/software-mansion/react-native-screens/pull/3959/commits/64233d8654b63244673b9d339bc660f5f93f5753, the plugin is listed as a dev dependency inside `apps`.

apps/eslint-plugin-local-rules/require-top-level-exports.js
```diff
@@ -0,0 +1,101 @@
...
+            });
+          }
+        }
+      },
+    };
+  },
+};
+
+function isPascalCase(name) {
+  return /^[A-Z]/.test(name);
```
t0maboro: what if we pass `SOME_FUNCTION_NAME` to check?

sgaczol: It is flagged as a component then. Should we add a check regarding the underscore _ sign?

t0maboro: as aligned internally sth like this `^[A-Z]([A-Z]|[a-z]|[0-9])*` should work good enough in our case

sgaczol: done: https://github.com/software-mansion/react-native-screens/pull/3959/commits/c0ebd408ff85fdd0d24ba7114eb80637c59cdd2d

apps/eslint-plugin-local-rules/require-top-level-exports.js
```diff
@@ -0,0 +1,101 @@
...
+            });
+          }
+        }
+      },
+    };
+  },
+};
+
+function isPascalCase(name) {
+  return /^[A-Z]([A-Z]|[a-z]|[0-9])*/.test(name);
```
t0maboro: sorry, my bad in the previous suggestion, the regex should be terminated

```suggestion
  return /^[A-Z]([A-Z]|[a-z]|[0-9])*$/.test(name);
```

apps/src/tests/single-feature-tests/tabs/test-tabs-general-appearance-no-liquid-glass-ios/index.tsx
t0maboro: why a new TC is added in this PR?

apps/src/tests/single-feature-tests/tabs/test-tabs-system-item-ios/index.tsx
t0maboro: same as above

ios/RNSModule.mm
t0maboro: okay, seems like sth wrong happened during the merge, this change shouldn't be included in diff

t0maboro (commented):
One last suggestion regarding the content. Additionally, the PR should be cleaned up from the content that doesn't belong to it before I can approve it.

ios/RNSModule.mm
sgaczol: I'm not sure what happened there, but I force pushed the branch state before that merge and there are no conflicts anymore so I guess the problem was about the main branch changes

apps/eslint-plugin-local-rules/require-top-level-exports.js
sgaczol: https://github.com/software-mansion/react-native-screens/pull/3959/commits/ac1a2b8f31ffa61149b12c92e60ad2deff8a4032

t0maboro (approved):

PR #3963, opened 2026-05-04
chore(examples): Apply RNRepo to FabricExample

radoslawrolka:
## Description

Adds RNRepo build-tools integration to `FabricExample` so that CI builds can consume pre-built native dependencies instead of compiling them from scratch, speeding up CI.

The integration is guarded by `CI=true` / `ENV['CI']` environment variables, so local development is unaffected.

## Changes

- `FabricExample/android/build.gradle` — resolves `@rnrepo/build-tools` via Node and adds its Gradle plugin JAR to the buildscript classpath.
- `FabricExample/android/app/build.gradle` — applies `org.rnrepo.tools.prebuilds-plugin` when running in CI.
- `FabricExample/ios/Podfile` — requires the RNRepo CocoaPods plugin and calls `rnrepo_post_install` in the `post_install` hook, both gated on `ENV['CI']`.
- `FabricExample/package.json` — adds `@rnrepo/build-tools ~0.1.3-beta.0` as a dev dependency.

## Performance

CI build time comparison (rnrepo vs baseline):

| Platform | Category | rnrepo | basic | Faster |
| Android | Without runner cache | 6:53 avg | 8:35 avg | **19.8%** |
| Android | With runner cache | 3:17 avg | 5:29 avg | **40.1%** |
| iOS | Without runner cache | 6:22 | 6:16 | -1.6% |
| iOS | With runner cache | 6:28 avg | 6:18 avg | -2.6% |
| iOS | With runner cache and prebuilt rncore | 3:31 avg | 4:01 avg | **12.4%** |

Android sees the biggest wins (~20% standard, ~40% with cache). iOS gains are visible when `rncore` is prebuilt (~12%) (tracked in #3955).

## Test plan

- Verified that a local build (without `CI=true`) still works without any RNRepo-related side effects.
- CI run with `CI=true` should pick up pre-built artifacts via the RNRepo plugin on both Android and iOS.

### DIFF
--- a/FabricExample/android/app/build.gradle
+++ b/FabricExample/android/app/build.gradle
@@ -3,7 +3,7 @@ apply plugin: "org.jetbrains.kotlin.android"
 apply plugin: "com.facebook.react"
 
 // Use RNRepo in CI builds
-if (System.getenv("CI") == "true") {
+if (["1", "true"].contains(System.getenv("CI"))) {
   apply plugin: "org.rnrepo.tools.prebuilds-plugin"
 }
 
--- a/FabricExample/android/app/build.gradle
+++ b/FabricExample/android/app/build.gradle
@@ -2,6 +2,11 @@ apply plugin: "com.android.application"
 apply plugin: "org.jetbrains.kotlin.android"
 apply plugin: "com.facebook.react"
 
+// Use RNRepo in CI builds
+if (System.getenv("CI") == "true") {
+  apply plugin: "org.rnrepo.tools.prebuilds-plugin"
+}
+
 /**
  * This is the configuration block to customize your React Native Android app.
  * By default you don't need to apply any configuration, just uncomment the lines you need.
--- a/FabricExample/android/app/build.gradle
+++ b/FabricExample/android/app/build.gradle
@@ -2,9 +2,19 @@ apply plugin: "com.android.application"
 apply plugin: "org.jetbrains.kotlin.android"
 apply plugin: "com.facebook.react"
 
-// Use RNRepo in CI builds
-if (["1", "true"].contains(System.getenv("CI"))) {
-  apply plugin: "org.rnrepo.tools.prebuilds-plugin"
+def isCIEnabled() {
+    return ["1", "true"].contains(System.getenv("CI")?.toLowerCase())
+}
+
+// System.setProperty("DISABLE_RNREPO", "1") // Uncomment to disable RNRepo even in CI
+def isRNRepoEnabled() {
+    return System.getenv("DISABLE_RNREPO") == null
+}
+
+// Use RNRepo in CI builds. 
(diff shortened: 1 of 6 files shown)


FabricExample/android/build.gradle
```diff
@@ -15,6 +15,14 @@ buildscript {
...
         classpath("com.facebook.react:react-native-gradle-plugin")
         classpath("org.jetbrains.kotlin:kotlin-gradle-plugin")
+        // RNRepo plugin in CI builds (plugin applied in app/build.gradle)
+        def rnrepoDir = new File(
+            providers.exec {
+                workingDir(rootDir)
+                commandLine("node", "--print", "require.resolve('@rnrepo/build-tools/package.json')")
+            }.standardOutput.asText.get().trim()
+        ).getParentFile().absolutePath
+        classpath fileTree(dir: "${rnrepoDir}/gradle-plugin/build/libs", include: ["prebuilds-plugin.jar"])
```
kkafar: Move it out to a single easily removable function

FabricExample/android/app/build.gradle
```diff
@@ -2,6 +2,11 @@ apply plugin: "com.android.application"
 apply plugin: "org.jetbrains.kotlin.android"
 apply plugin: "com.facebook.react"
 
+// Use RNRepo in CI builds
+if (["1", "true"].contains(System.getenv("CI"))) {
+  apply plugin: "org.rnrepo.tools.prebuilds-plugin"
+}
```
kkafar: Let's gate it additionally with an project property from gradle.properties, e.g. `rnsUseRnRepo`

FabricExample/ios/Podfile
```diff
@@ -5,6 +5,15 @@ require Pod::Executable.execute_command('node', ['-p',
     {paths: [process.argv[1]]},
   )', __dir__]).strip
 
+# Use RNRepo in CI builds
+if %w[1 true].include?(ENV['CI'].to_s.downcase)
```
kkafar: Let's extract it to isCIEnabled or something

FabricExample/android/build.gradle
radoslawrolka: done

FabricExample/ios/Podfile
radoslawrolka: done

FabricExample/android/app/build.gradle
radoslawrolka: done - requires to uncomment the
```
// System.setProperty("DISABLE_RNREPO", "1") // Uncomment to disable RNRepo even in CI
def isRNRepoEnabled() {
    return System.getenv("DISABLE_RNREPO") == null
}

```

kkafar (approved):
LGTM, thanks!

PR #3966, opened 2026-05-04
feat(tabs): cohesive cross-platform native API + observer registry

kkafar:
## Description

Stacked on #3951.

Defines a small, cohesive cross-platform public contract on `TabsContainer` (Android) and `RNSTabBarController` (iOS) so downstream native libraries can integrate against the tabs container with a stable surface. Today's API grew organically: the two platforms diverged, the single-delegate model can't fan out to multiple observers, and many host-only methods are visible to third-party consumers (especially on iOS where the public `.h` exposes ~12 individual `update*` flush methods plus `setNeeds*` invalidation properties).

The downstream contract is now exactly four things: read the current navigation state, submit a tab-selection request, flush pending updates, and observe results — and the observer side supports multiple registrations rather than a single delegate. The submission entry point is deliberately narrow: third-party callers cannot specify the action origin or base provenance — both are filled in by the container — so they can only ever produce `programmatic-native`-origin transitions.

This is a design / refactor PR, not a feature add. No new public-API distribution mechanism (no umbrella headers, module maps, or framework targets) per design constraint — public/internal split is enforced by Kotlin visibility on Android and `#pragma mark` sections on iOS.

### Public API (cross-platform, identical shape)

| Capability | Android `TabsContainer` | iOS `RNSTabBarController` |
|---|---|---|
| Read current state | `val navigationState: TabsNavigationState` | `@property (readonly, nullable) RNSTabsNavigationState *navigationState;` |
| Submit tab selection | `fun submitSelectionOfTabsScreenWithKey(screenKey: String)` | `- (void)submitSelectionOfTabsScreenWithKey:(NSString *)screenKey` |
| Flush updates | `fun flushPendingUpdates()` | `- (void)flushPendingUpdates;` |
| Add observer | `fun addNavigationStateObserver(observer): Boolean` | `- (BOOL)addNavigationStateObserver:` |
| Remove observer | `fun removeNavigationStateObserver(observer): Boolean` | `- (BOOL)removeNavigationStateObserver:` |

`submitSelectionOfTabsScreenWithKey` builds a `TabsNavigationStateUpdateRequest` internally with `actionOrigin = PROGRAMMATIC_NATIVE` and the current `navigationState.provenance` (or the `EMPTY` sentinel if no state is established yet) — neither can be fabricated by the caller. The verb `submit` (not `select`) signals queued semantics: the caller still needs `flushPendingUpdates` (or the next React mounting transaction) for the change to take effect.

The fully-flexible `setPendingNavigationStateUpdate(request)` form stays accessible to the host (Kotlin `internal` / iOS Internal API section) because the host needs to dispatch JS-driven requests with `PROGRAMMATIC_JS` origin.

### Observer registry replaces single delegate

- New `TabsNavigationStateObserverRegistry` (Android) / `RNSTabsNavigationStateObserverRegistry` (iOS). Multi-observer fan-out; observers held with strong references.
- The host (`TabsHost` / `RNSTabsHostComponentView`) registers itself as one of many possible observers; downstream native consumers can register additional ones.
- Reentrant emission is forbidden: `emit*` asserts no emission is in flight (recursion would deliver out-of-order events to later observers); `add` / `remove` return `Boolean` / `BOOL` and reject during in-flight emission.

### Cross-platform name parity

Android renames:
- `TabsNavState` → `TabsNavigationState`
- `TabsNavStateUpdateRequest` → `TabsNavigationStateUpdateRequest`
- `TabsNavStateUpdateRejectionReason` → `TabsNavigationStateRejectionReason`
- `TabsContainerDelegate` → `TabsNavigationStateObserver` (now public)
- `performContainerUpdateIfNeeded` → `flushPendingUpdates`
- `rejectOpsWithStaleNavState` → `rejectStaleNavigationStateUpdates`
- Callbacks `onNavStateUpdate*` → `onNavigationStateUpdate*`
- `TabsContainerOp` / `TabSelectOp` sealed-class wrappers removed; the container now stores `TabsNavigationStateUpdateRequest?` directly (matches iOS structure).

iOS renames:
- `RNSTabBarControllerDelegate` → `RNSTabsNavigationStateObserver` (lives in `RNSTabsNavigationState.h` next to its value objects)
- Selectors `tabBarController:` → `tabsContainer:` (sender argument name unified across platforms)
- The iOS-only More-tab callback stays as `@optional` on the same protocol — no separate iOS-only protocol introduced.

### Public/internal boundary

- **Android**: Kotlin `internal` / `public` modifiers, with `// region Public API` / `// region Host-internal API` / `// region Private helpers` markers in `TabsContainer.kt` for IDE folding.
- **iOS**: `#pragma mark - Public API` and `#pragma mark - Internal API (host-only; subject to change without notice)` sections in `RNSTabBarController.h`. Class-level doc explicitly marks the boundary. The ~12 individual `update*IfNeeded` / `setNeeds*` methods, `childViewControllersHaveChangedTo:`, `tabBarAppearanceCoordinator`, `rejectStaleNavigationStateUpdates`, and the host-only `setPendingNavigationStateUpdate:` all live under the Internal API section.

### Misc

- Android `TabsContainer` constructor drops its `delegate` parameter — host registers as observer after construction.
- New internal `tearDown()` method on both platforms — clears the observer registry, releases retained host references, drops pending operation. Called from `TabsHostViewManager.onDropViewInstance` (Android) and `RNSTabsHostComponentView.invalidateImpl` (iOS). Named `tearDown` (not `invalidate`) because `Android.View.invalidate()` already exists and would shadow.
- iOS `submitSelectionOfTabsScreenWithKey:` falls back to `INT_MIN` when `_navigationState` is nil — matches Android's `TabsNavigationState.EMPTY` sentinel (`Int.MIN_VALUE`). Both `isNavigationStateUpdateStale` checks already guard on nil/empty states; the parity is for semantic alignment.
- Doc comments preserved on iOS `RNSTabBarController.h` for all originally-documented members (the structural rewrite kept content, just regrouped under the pragma marks).

### Out of scope (stays as-is)

- JS / TS / codegen — public TS types (`TabSelectedEvent`, codegen specs) unchanged.
- C++ layer — unchanged.
- Stack-related code (different gamma namespace).
- TS-side `apps/src/shared/gamma/containers/tabs/...TabsNavState` type alias — example-app-only, separate from the renamed Kotlin/Obj-C types.

## Test plan

- [x] `yarn check-types` clean.
- [x] Android Kotlin compile: `./gradlew :react-native-screens:compileDebugKotlin` clean.
- [x] Grep audit (zero hits expected, all confirmed): `TabsContainerDelegate`, `TabsNavState`, `TabsNavStateUpdateRequest`, `TabsNavStateUpdateRejectionReason`, `RNSTabBarControllerDelegate`, `performContainerUpdateIfNeeded`, `rejectOpsWithStaleNavState`, `tabBarController:self`, `TabsContainerOp`, `TabSelectOp`.
- [ ] iOS Xcode build via `bundle exec pod install && yarn ios` from `FabricExample/` — pending local Mac toolchain run (Pods not installed in CI sandbox).
- [ ] Manual smoke test on Android (`FabricExample`):
  - Tap-driven tab switches
  - JS-driven `navStateRequest` updates
  - Repeated tap → scroll-to-top / pop-to-root special effects
  - `preventNativeSelection` flow
  - `submitSelectionOfTabsScreenWithKey` from a debug-only native call site → confirm `actionOrigin` reaches JS as `programmatic-native`
- [ ] Manual smoke test on iOS (iPhone + iPad):
  - Same flows as Android
  - iPad More-tab path (verify `@optional` More-tab callback still reaches host)
- [ ] Multi-observer sanity: temporary debug patch registering a second `TabsNavigationStateObserver` (or `RNSTabsNavigationStateObserver`) and confirming fan-out fires for all callbacks.
- [ ] `tearDown` lifecycle: confirm registry is cleared on `onDropViewInstance` (Android) / `invalidateImpl` (iOS) and that subsequent emits are no-ops.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
@@ -95,7 +95,7 @@ class TabsContainer internal constructor(
         checkNotNull(pendingStateUpdateRequest) { "[RNScreens] Attempt to require nullish pendingStateUpdateRequest" }
 
     /**
-     * Denotes whether container is currently performing update triggered by the `pendingOperation`.
+     * Denotes whether container is currently performing update triggered by the `pendingStateUpdateRequest`.
      */
     private var isInExternalOperationContext: Boolean = false
 
@@ -248,7 +248,7 @@ class TabsContainer internal constructor(
 
     /**
      * Idempotent teardown. Releases observer references and clears any pending operation.
-     * Called by the host on view recycle. Note: named `tearDown` (not `invalidate`) to avoid
+     * Called by the host on view lifecycle end. Note: named `tearDown` (not `invalidate`) to avoid
      * shadowing [android.view.View.invalidate].
      */
     internal fun tearDown() {
@@ -539,9 +539,7 @@ class TabsContainer internal constructor(
                 hasTriggeredSpecialEffect = hasTriggeredSpecialEffect,
                 actionOrigin =
                     if (isInExternalOperationContext) {
-                        checkNotNull(pendingStateUpdateRequest) {
-                            "[RNScreens] Unexpected pending operation $pendingStateUpdateRequest while in external operation context"
-                        }.actionOrigin
+                        requirePendingStateUpdateRequest().actionOrigin
                     } else {
                         TabsActionOrigin.USER
                     },
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
@@ -89,9 +89,9 @@ class TabsContainer internal constructor(
 
     internal val invalidationFlags = TabsContainerInvalidationFlags()
 
-    private var pendingOperation: TabsContainerOp? = null
-    internal val hasPendingOperation
-        get() = pendingOperation != null
+    private var pendingStateUpdateRequest: TabsNavigationStateUpdateRequest? = null
+    private val requirePendingStateUpdateRequest: TabsNavigationStateUpdateRequest
+        get() = checkNotNull(pendingStateUpdateRequest) { "[RNScreens] Attempt to require nullish pendingStateUpdateRequest" }
 
     /**
      * Denotes whether container is currently performing update triggered by the `pendingOperation`.
@@ -174,9 +174,14 @@ class TabsContainer internal constructor(
      * Queue a navigation state update. Apply via [flushPendingUpdates] or rely on the
      * host's next render cycle to apply automatically.
      */
-    fun setPendingNavigationStateUpdate(request: TabsNavigationStateUpdateRequest) {
-        pendingOperation = TabSelectOp(request)
-        invalidationFlags.isSelectedTabInvalidated = true
+    fun submitSelectionOfTabsScreenWithKey(screenKey: String) {
+        setPendingNavigationStateUpdate(
+            TabsNavigationStateUpdateRequest(
+                screenKey,
+                navigationState.provenance,
+                TabsActionOrigin.PROGRAMMATIC_NATIVE,
+            ),
+        )
     }
 
     /**
@@ -196,6 +201,15 @@ class TabsContainer internal constructor(
     // endregion
 
     // region Host-internal API
+
+    /**
+     * Queue a navigation state update. Apply via [flushPendingUpdates] or rely on the
+     * host's next render cycle to apply automatically.
+     */
+    internal fun setPendingNavigationStateUpdate(request: TabsNavigationStateUpdateRequest?) {
+        pendingStateUpdateRequest = request
+        invalidationFlags.isSelectedTabInvalidated = request != null
+    }
 
     internal fun addTabsScreenAt(
         index: Int,
@@ -238,8 +252,7 @@ class TabsContainer internal constructor(
      */
     internal fun tearDown() {
         observerRegistry.clear()
-        pendingOperation = null
-        invalidationFlags.isSelectedTabInvalidated = false
+        setPendingNavigationStateUpdate(null)
     }
 
     // endregion
@@ -397,27 +410,25 @@ class TabsContainer internal constructor(
     }
 
     private fun performOperation() {
-        if (pendingOperation == null) {
+        if (pendingStateUpdateRequest == null) {
             RNSLog.w(TAG, "TabsContainer::performOperation called w/o pending operation; skipping update")
             return
         }
 
-        check(hasPendingOperation) { "[RNScreens] Attempt to update container with empty state and no pending update" }
-        check(pendingOperation is TabSelectOp)
-        val tabSelectOp = pendingOperation as TabSelectOp
+        val stateUpdateRequest = requirePendingStateUpdateRequest
 
         val nextSelectedMenuItemId =
-            checkNotNull(getMenuItemIdForFragment(requireFragmentForScreenKey(tabSelectOp.request.selectedScreenKey))) {
-                "[RNScreens] Failed to find Menu Item for screenKey: ${tabSelectOp.request.selectedScreenKey}"
+            checkNotNull(getMenuItemIdForFragment(requireFragmentForScreenKey(stateUpdateRequest.selectedScreenKey))) {
+                "[RNScreens] Failed to find Menu Item for screenKey: ${stateUpdateRequest.selectedScreenKey}"
             }
 
-        if (rejectStaleNavigationStateUpdates && isNavigationStateStale(tabSelectOp.request)) {
+        if (rejectStaleNavigationStateUpdates && isNavigationStateStale(stateUpdateRequest)) {
             observerRegistry.emitOnNavigationStateUpdateRejected(
                 navState,
-                tabSelectOp.request,
+                stateUpdateRequest,
                 TabsNavigationStateRejectionReason.STALE,
             )
-            pendingOperation = null
+            pendingStateUpdateRequest = null
             return
         }
 
@@ -429,12 +440,12 @@ class TabsContainer internal constructor(
         } else {
             observerRegistry.emitOnNavigationStateUpdateRejected(
                 navState,
-                tabSelectOp.request,
+                stateUpdateRequest,
                 TabsNavigationStateRejectionReason.REPEATED,
             )
         }
 
-        pendingOperation = null
+        pendingStateUpdateRequest = null
     }
 
     private fun updateNavigationMenuStructure() {
@@ -464,7 +475,7 @@ class TabsContainer internal constructor(
 
     private fun updateSelectedFragment(nextSelectedFragment: TabsScreenFragment): Boolean {
         if (navState.isEmpty()) {
-            check(isInExternalOperationContext && hasPendingOperation)
+            check(isInExternalOperationContext && pendingStateUpdateRequest != null)
             navState = TabsNavigationState(nextSelectedFragment.requireScreenKey, 0)
             requireFragmentManager
                 .createTransactionWithReordering()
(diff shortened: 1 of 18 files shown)


android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
```diff
@@ -152,12 +162,110 @@ class TabsContainer internal constructor(
...
+    /**
+     * Current navigation state of the container.
+     * Returns [TabsNavigationState.EMPTY] before the first tab is selected.
+     */
+    val navigationState: TabsNavigationState
+        get() = navState
+
+    /**
+     * Queue a navigation state update. Apply via [flushPendingUpdates] or rely on the
+     * host's next render cycle to apply automatically.
```
t0maboro, line 176: Maybe we should highlight that flushPendingUpdates will apply changes immediately?

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
```diff
@@ -152,12 +162,110 @@ class TabsContainer internal constructor(
...
+    internal fun teardownFragmentManager() {
+        fragmentManager = null
+    }
+
+    /**
+     * Idempotent teardown. Releases observer references and clears any pending operation.
+     * Called by the host on view recycle. Note: named `tearDown` (not `invalidate`) to avoid
+     * shadowing [android.view.View.invalidate].
+     */
+    internal fun tearDown() {
```
t0maboro, line 254: nit: above you have `teardown...`, here `tearDown` - should be aligned to a single convention

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsNavigationStateObserverRegistry.kt
```diff
@@ -0,0 +1,91 @@
...
+        isRepeated: Boolean,
+        hasTriggeredSpecialEffect: Boolean,
+        actionOrigin: TabsActionOrigin,
+    ) {
+        check(!isEmitting) { "[RNScreens] Recursive emission on TabsNavigationStateObserverRegistry" }
+        isEmitting = true
+        observers.forEach {
+            it.onNavigationStateUpdate(navState, isRepeated, hasTriggeredSpecialEffect, actionOrigin)
+        }
+        isEmitting = false
```
t0maboro: maybe we should consider wrapping it in finally block, to ensure, that this lock is always cleared

ios/tabs/host/RNSTabsHostComponentView.mm
```diff
@@ -85,6 +85,9 @@ - (void)initState
   [self resetProps];
 
   _controller = [[RNSTabBarController alloc] initWithTabsHostComponentView:self];
+  RCTAssert(
+      [_controller addNavigationStateObserver:self],
```
t0maboro: I don't see any `removeNavigationStateObserver` call for symmetry, only `tearDown` which clears the registry, should we unregister a single observer in any case?

t0maboro (approved):

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
```diff
@@ -70,20 +77,22 @@ class TabsContainer internal constructor(
...
-    private var pendingOperation: TabsContainerOp? = null
-    internal val hasPendingOperation
-        get() = pendingOperation != null
+    private var pendingStateUpdateRequest: TabsNavigationStateUpdateRequest? = null
+
+    private fun requirePendingStateUpdateRequest(): TabsNavigationStateUpdateRequest =
+        checkNotNull(pendingStateUpdateRequest) { "[RNScreens] Attempt to require nullish pendingStateUpdateRequest" }
 
     /**
      * Denotes whether container is currently performing update triggered by the `pendingOperation`.
```
kligarski:
```suggestion
     * Denotes whether container is currently performing update triggered by the `pendingStateUpdateRequest`.
```

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
```diff
@@ -373,16 +533,15 @@ class TabsContainer internal constructor(
...
                 hasTriggeredSpecialEffect = hasTriggeredSpecialEffect,
                 actionOrigin =
                     if (isInExternalOperationContext) {
-                        check(pendingOperation != null && pendingOperation is TabSelectOp) {
-                            "[RNScreens] Unexpected pending operation $pendingOperation while in external operation context"
-                        }
-                        (pendingOperation as TabSelectOp).request.actionOrigin
+                        checkNotNull(pendingStateUpdateRequest) {
+                            "[RNScreens] Unexpected pending operation $pendingStateUpdateRequest while in external operation context"
+                        }.actionOrigin
```
kligarski:
```suggestion
                        requirePendingStateUpdateRequest().actionOrigin
```
or do we want separate message here (but we're only checking not null, previously it was also operation type)

ios/tabs/host/RNSTabBarController.h
```diff
@@ -70,61 +117,85 @@ NS_ASSUME_NONNULL_BEGIN
...
+ * Host-only: third-party native consumers should use `submitSelectionOfTabsScreenWithKey:` instead, which
+ * forces `RNSTabsActionOriginProgrammaticNative`. This method allows the host to dispatch any
+ * origin (e.g. `RNSTabsActionOriginProgrammaticJs` for JS-driven `navStateRequest` updates).
  */
-@property (nonatomic, readwrite) BOOL rejectStaleNavigationStateUpdates;
+- (void)setPendingNavigationStateUpdate:(nullable RNSTabsNavigationStateUpdateRequest *)stateUpdate;
+
+/**
+ * Idempotent teardown. Releases observer references and any retained host references.
+ * Called by the host on view recycle.
```
kligarski: We don't support view recycling for TabsHost currently.

```suggestion
 * Called by the host on view invalidation.
```

```suggestion
 * Called by the host on view teardown.
```

kligarski: Same applies to Android equivalent.

kligarski (approved):
Looks good.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsNavigationStateObserverRegistry.kt
kkafar: https://github.com/software-mansion/react-native-screens/pull/3966/commits/9da01b69cfd848afc91fa271c47d984f12b4e9e4

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
kkafar: I think that's clear from `flushPendingUpdates` description.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
kkafar: Thats correct & intentional here 😄. `teardown` is a noun, and `tearDown` is "two-word" verb.

ios/tabs/host/RNSTabsHostComponentView.mm
kkafar: The method is meant for 3rd parties, e.g. Expo Router. Likely, we won't have any use cases in our codebase most likely.

ios/tabs/host/RNSTabBarController.h
kkafar: https://github.com/software-mansion/react-native-screens/pull/3966/commits/e25ca1e28f3ac931e39836f79f1e19481d2c83f1

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
kkafar: https://github.com/software-mansion/react-native-screens/pull/3966/commits/e25ca1e28f3ac931e39836f79f1e19481d2c83f1

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
kkafar: https://github.com/software-mansion/react-native-screens/pull/3966/commits/e25ca1e28f3ac931e39836f79f1e19481d2c83f1

kkafar:
I'll proceed here. The iOS e2e Ci fails for some `react-native-worklets` related error apparently.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
kligarski: I think @t0maboro meant that the method above is called `teardownFragmentManager` instead of `tearDownFragmentManager` but it's also used as a verb.

PR #3974, opened 2026-05-06
fix(ios): realign large title with custom left item on iOS 26

### LINKED ISSUE #3602: Large Header Title misaligned when using react-navigation/drawer
### Description

First of all, I am not 100% sure if this is an issue with this repo or react-navigation but as it's related to native-stack, I am opening it here.

When using react-navigation/drawer to navigate to a screen that has `headerLargeTitleEnabled: true`, the large title is misaligned with the left header button. When navigating without a drawer, it works fine.

This also seems to only affect iOS 26.x as it works fine with iOS 18.
Tested on real device as well as simulator.

```
import 'react-native-gesture-handler';
import React from 'react';
import { View, Text, Button, ScrollView } from 'react-native';
import { createDrawerNavigator } from '@react-navigation/drawer';
import { GestureHandlerRootView } from 'react-native-gesture-handler';
import { SafeAreaProvider } from 'react-native-safe-area-context';

const Drawer = createDrawerNavigator();

const HomeStack = createNativeStackNavigator();
const LargeStack = createNativeStackNavigator();

function ButtonScreen({ navigation }) {
    return (
            <Text>Open the drawer to navigate to large title screen</Text>
        </View>
    );
}

function ScrollScreen() {
    return (
        <ScrollView contentContainerStyle={{ padding: 16 }} contentInsetAdjustmentBehavior='automatic'>
            <Text style={{ fontSize: 22, marginBottom: 12 }}>Large Title (ScrollView)</Text>
            {Array.from({ length: 30 }).map((_, i) => (
                <Text key={i} style={{ marginBottom: 8 }}>
                    {`Entry ${i + 1} — Example content for scrolling.`}
                </Text>
            ))}
        </ScrollView>
    );
}

function HomeStackScreen() {
    return (
        <HomeStack.Navigator>
            <HomeStack.Screen
                name="Button"
                component={ButtonScreen}
                options={({ navigation }) => ({
                    title: 'Button',
                    headerLeft: () => (
                        <Button title="Menu" onPress={() => navigation.getParent()?.openDrawer()} />
                    ),
                })}
            />
        </HomeStack.Navigator>
    );
}

function LargeStackScreen() {
    return (
        <LargeStack.Navigator>
            <LargeStack.Screen
                name="Scroll"
                component={ScrollScreen}
                options={({ navigation }) => ({
                    title: 'Scroll',
                    headerLargeTitle: true,
                    headerLeft: () => (
                        <Button title="Menu" onPress={() => navigation.getParent()?.openDrawer()} />
                    ),
                })}
            />
        </LargeStack.Navigator>
    );
}

    return (
        <GestureHandlerRootView style={{ flex: 1 }}>
            <SafeAreaProvider>
                <NavigationContainer>
                    <Drawer.Navigator initialRouteName="Home" screenOptions={{ headerShown: false }}>
                        <Drawer.Screen name="Home" component={HomeStackScreen} />
                        <Drawer.Screen name="LargeTitle" component={LargeStackScreen} options={{ title: 'Large Title' }} />
                    </Drawer.Navigator>
                </NavigationContainer>
            </SafeAreaProvider>
        </GestureHandlerRootView>
    );
}
```

iOS 18:

Image

iOS 26:

Image

### Steps to reproduce

1. open drawer menu
2. navigate to "Large Title"

https://gist.github.com/Relax594/b57d4e19ba4007288b8f3ba64a0bccb5

### Screens version

4.21.0

### React Native version

0.81.5

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

Paper (Old Architecture)

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

Jasleen-Kaur96:
**Summary**
Fixes iOS 26 large-title misalignment after drawer-driven navigation by repairing asymmetric navigation bar directional margins during deferred relayout.
The fix now applies to **iOS 26 + headerLargeTitle: true** (with or without custom `headerLeft`), since the same collapsed-leading state can occur in both cases.

**Linked Issue**
Closes #3602

**Reproduction**
Using a drawer navigator:

1. Open drawer
2. Navigate to a screen with:

- `headerLargeTitle: true`
- with or without custom `headerLeft`

3. On iOS 26.x, large title can appear misaligned due to asymmetric nav bar margins after transition
This does not reproduce on iOS 18 in the same setup.

**Root Cause**
On iOS 26, after drawer-driven navigation, `UINavigationBar` can temporarily end up with asymmetric directional margins (collapsed `leading` vs normal `trailing`) during large-title layout. This causes visible title alignment drift.

**Fix**
In `RNSScreenStackHeaderConfig.mm`:

- Updated `scheduleDeferredNavigationBarRelayoutIfNeeded:withConfig`:
- Scoped to `iOS 26 + config.largeTitle` (intentionally not restricted to `hasSubviewLeft`)
- Detects asymmetry dynamically and adjusts **only leading** while preserving trailing
- Simplified deferred relayout flow:
    - single deferred dispatch
    - second layout pass only when repair is applied
- Fixed cancelled-transition handling:
    - computes effective config for completion path
    - uses `fromHeaderConfig` on cancel
    - avoids config shadowing issues

**Why This Scope**
Although initially observed with custom `headerLeft`, the same iOS 26 asymmetry occurs on large-title screens without custom left items. Restricting to `hasSubviewLeft` regresses that case.

**Test Plan**

- Reproduced via drawer → large-title screen
- Verified on iOS 26.1 simulator:
   - `headerLargeTitle: true` + custom `headerLeft`
   - `headerLargeTitle: true` without custom `headerLeft`
- Compared against iOS 18.3 baseline
- Repeated navigation to verify consistency across first and subsequent opens

**Notes**
- Native-only change in header layout handling
- No JS API surface changes

### DIFF
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -285,9 +285,9 @@ RNS_IGNORE_SUPER_CALL_END
       CGFloat trailing = m.trailing;
       // Float slack only (not a screen layout constant like "16pt").
       static const CGFloat kRNSDirectionalMarginAsymmetrySlack = 0.75;
-      BOOL repairedAsymmetricMargins =
+      BOOL shouldRepairAsymmetricMargins =
           (leading + kRNSDirectionalMarginAsymmetrySlack < trailing && trailing > 1.0);
-      if (repairedAsymmetricMargins) {
+      if (shouldRepairAsymmetricMargins) {
         // Preserve trailing; only align collapsed leading to the bar's own trailing inset.
         bar.directionalLayoutMargins = NSDirectionalEdgeInsetsMake(m.top, trailing, m.bottom, m.trailing);
         layoutNavigationChrome();
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -262,27 +262,37 @@ RNS_IGNORE_SUPER_CALL_END
 {
 #if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
   if (@available(iOS 26.0, *)) {
-    if (config.largeTitle && config.hasSubviewLeft) {
-      // iOS 26: after drawer-driven navigation, large title can lay out with collapsed
-      // leading inset while custom left items use the standard 16pt inset. Sync nav bar
-      // directional margins and force a second layout pass on the next runloop tick.
-      dispatch_async(dispatch_get_main_queue(), ^{
-        NSDirectionalEdgeInsets m = navctr.navigationBar.directionalLayoutMargins;
-        navctr.navigationBar.directionalLayoutMargins =
-            NSDirectionalEdgeInsetsMake(m.top, 16.0, m.bottom, 16.0);
-        [navctr.view setNeedsLayout];
-        [navctr.navigationBar setNeedsLayout];
-        [navctr.view layoutIfNeeded];
-        [navctr.navigationBar layoutIfNeeded];
-
-        dispatch_async(dispatch_get_main_queue(), ^{
-          [navctr.view setNeedsLayout];
-          [navctr.navigationBar setNeedsLayout];
-          [navctr.view layoutIfNeeded];
-          [navctr.navigationBar layoutIfNeeded];
-        });
-      });
+    if (config == nil || !config.largeTitle) {
+      return;
     }
+
+    // iOS 26: after drawer-driven navigation, large-title headers can end up with asymmetric
+    // directional margins (collapsed leading vs normal trailing). We avoid hardcoded constants
+    // and preserve trailing: when asymmetry is detected, only leading is aligned to trailing.
+    dispatch_async(dispatch_get_main_queue(), ^{
+      UINavigationBar *bar = navctr.navigationBar;
+      void (^layoutNavigationChrome)(void) = ^{
+        [navctr.view setNeedsLayout];
+        [bar setNeedsLayout];
+        [navctr.view layoutIfNeeded];
+        [bar layoutIfNeeded];
+      };
+
+      layoutNavigationChrome();
+
+      NSDirectionalEdgeInsets m = bar.directionalLayoutMargins;
+      CGFloat leading = m.leading;
+      CGFloat trailing = m.trailing;
+      // Float slack only (not a screen layout constant like "16pt").
+      static const CGFloat kRNSDirectionalMarginAsymmetrySlack = 0.75;
+      BOOL repairedAsymmetricMargins =
+          (leading + kRNSDirectionalMarginAsymmetrySlack < trailing && trailing > 1.0);
+      if (repairedAsymmetricMargins) {
+        // Preserve trailing; only align collapsed leading to the bar's own trailing inset.
+        bar.directionalLayoutMargins = NSDirectionalEdgeInsetsMake(m.top, trailing, m.bottom, m.trailing);
+        layoutNavigationChrome();
+      }
+    });
   }
 #endif // RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
 }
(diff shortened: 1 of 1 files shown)


ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -257,6 +257,36 @@ + (void)setAnimatedConfig:(UIViewController *)vc withConfig:(RNSScreenStackHeade
...
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
+  if (@available(iOS 26.0, *)) {
+    if (config.largeTitle && config.hasSubviewLeft) {
+      // iOS 26: after drawer-driven navigation, large title can lay out with collapsed
+      // leading inset while custom left items use the standard 16pt inset. Sync nav bar
+      // directional margins and force a second layout pass on the next runloop tick.
+      dispatch_async(dispatch_get_main_queue(), ^{
+        NSDirectionalEdgeInsets m = navctr.navigationBar.directionalLayoutMargins;
+        navctr.navigationBar.directionalLayoutMargins =
+            NSDirectionalEdgeInsetsMake(m.top, 16.0, m.bottom, 16.0);
```
kligarski: What is the source of these constants? How can we rely that they're not going to change in next iOS releases?

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -257,6 +257,36 @@ + (void)setAnimatedConfig:(UIViewController *)vc withConfig:(RNSScreenStackHeade
...
+        [navctr.navigationBar setNeedsLayout];
+        [navctr.view layoutIfNeeded];
+        [navctr.navigationBar layoutIfNeeded];
+
+        dispatch_async(dispatch_get_main_queue(), ^{
+          [navctr.view setNeedsLayout];
+          [navctr.navigationBar setNeedsLayout];
+          [navctr.view layoutIfNeeded];
+          [navctr.navigationBar layoutIfNeeded];
+        });
```
kligarski: Requesting layout twice doesn't seem like a valid solution to the problem.

kligarski (commented):
Thank you for the PR. Unfortunately, I don't think we can land it in current shape.

We don't want to rely on hard-coded values and spamming the layout unless it's absolutely necessary. We would need to dive deeper into why this bug happens in the first place.

I've also seen some reports that it is reproducible in bare UIKit apps without `react-native-screens` or `react-native` (see [here](https://developer.apple.com/forums/thread/798256) and [here](https://x.com/luksape/status/1962134689717588459)). If that's the source of the problem, we'd rather rely on Apple fixing the bug instead of applying hacky workarounds in `screens`. 

We're also missing minimal test screen in this PR (following `Test<issue_number>.tsx` convention in our repo) that would help to verify the fix.

ios/RNSScreenStackHeaderConfig.mm
Jasleen-Kaur96: Removed the hardcoded 16pt margin override. The workaround now derives behavior from the nav bar’s runtime directional margins and only repairs asymmetric state (collapsed leading vs non-collapsed trailing), so it is not tied to a fixed iOS constant.

ios/RNSScreenStackHeaderConfig.mm
Jasleen-Kaur96: Updated to avoid the nested “dispatch + dispatch + relayout twice” pattern. The code now performs a single deferred layout pass, then runs a second pass only if margin repair was actually applied.

ios/RNSScreenStackHeaderConfig.mm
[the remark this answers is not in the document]
Jasleen-Kaur96: Fixed. We no longer overwrite both sides or set any fixed value. Trailing is preserved, and only leading is conditionally aligned when we detect asymmetric/collapsed leading margins.

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -672,9 +702,11 @@ + (void)updateViewController:(UIViewController *)vc
             }
             [self setAnimatedConfig:fromVC withConfig:config];
           }
+          [self scheduleDeferredNavigationBarRelayoutIfNeeded:navctr withConfig:config];
         }];
```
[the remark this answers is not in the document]
Jasleen-Kaur96: Fixed. In transition completion, we now compute the effective config for the final visible screen and pass that to deferred relayout. On cancellation, it uses fromHeaderConfig instead of destination config.

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -257,6 +257,46 @@ + (void)setAnimatedConfig:(UIViewController *)vc withConfig:(RNSScreenStackHeade
...
 
++ (void)scheduleDeferredNavigationBarRelayoutIfNeeded:(UINavigationController *)navctr
+                                            withConfig:(RNSScreenStackHeaderConfig *)config
+{
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
+  if (@available(iOS 26.0, *)) {
+    if (config == nil || !config.largeTitle) {
+      return;
+    }
+
```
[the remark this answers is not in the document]
Jasleen-Kaur96: Good catch — the scope is now intentionally broader than “largeTitle + custom headerLeft”. We observed the same collapsed-leading/asymmetric margin state on iOS 26 large-title screens without custom left items as well, so the guard is intentionally !config.largeTitle (not !config.hasSubviewLeft). If we add hasSubviewLeft gating, the no-custom-headerLeft case regresses. I’ll update the PR description/comments to reflect the broadened scope.

kligarski:
Hi, thanks for the PR. Without knowing the exact mechanism of the bug, we're not confident in this change and we're not willing to merge a solution that seems very hacky (using arbitrarily chosen threshold, setting leading margin to trailing margin, requesting layout multiple times). We would need a minimal reproduction example, analysis what is causing the bug and explanation why copying the inset is safe for all possible cases (including e.g. RTL, different orientations, iPad layout, ...).

PR #3975, opened 2026-05-06
chore: fix opt-out comment of RNRepo

radoslawrolka:
## Description

Mistakenly setProperty was added, whch is no longer supported as opt-out flag, so RNRepo can be disabled via uncommenting the `return false` in CI


t0maboro (approved):

kkafar:
Please change the PR title before merging to conform to conventional commits rules!

PR #3980, opened 2026-05-06
iOS: scope singleSelection to leaf header menus

### LINKED ISSUE #3973: unstable_headerRightItems: state: 'on' checkmark not displayed for items in the second of two sibling nested submenus
### Description

When two nested submenus are rendered as siblings inside an inline-grouped menu under unstable_headerRightItems, action items in the first submenu correctly display their state: 'on' checkmark, but action items in the second submenu show no checkmark — even though the JS-side data is structurally identical between the two submenus.

This appears related to the flat-index identifier scheme that issue [#3606](https://github.com/software-mansion/react-native-screens/issues/3606) and PR [#3762](https://github.com/software-mansion/react-native-screens/pull/3762) addressed for onPress handler routing. It looks like the same flat-index identity is also used to associate state with each item, so the second submenu's state is silently lost on the native side.

### Steps to reproduce

1. Create a screen using react-native-screens with `<Stack.Screen options={{ unstable_headerRightItems: () => [...] }} />`.
2. Configure a `type: 'menu'` item that contains an `inline: true` wrapper submenu, which itself contains **two sibling submenus** ("Sort" and "Group"). Each child submenu has a list of `type: 'action'` items, with `state: 'on'` set on the matching item:
```
{
  type: 'menu',
  label: 'More',
  icon: { type: 'sfSymbol', name: 'ellipsis' },
  menu: {
    items: [
      {
        type: 'submenu',
        label: '',
        inline: true,
        items: [
          {
            type: 'submenu',
            label: 'Sort',
            icon: { type: 'sfSymbol', name: 'arrow.up.arrow.down' },
            items: [
              { type: 'action', label: 'Manual',   state: 'off', onPress: () => {} },
              { type: 'action', label: 'Due date', state: 'on',  onPress: () => {} },
              { type: 'action', label: 'Priority', state: 'off', onPress: () => {} },
            ],
          },
          {
            type: 'submenu',
            label: 'Group',
            icon: { type: 'sfSymbol', name: 'square.3.layers.3d' },
            items: [
              { type: 'action', label: 'None',      state: 'on',  onPress: () => {} },
              { type: 'action', label: 'Goal',      state: 'off', onPress: () => {} },
              { type: 'action', label: 'Life area', state: 'off', onPress: () => {} },
            ],
          },
        ],
      },
    ],
  },
}
```
3. Open the "More" menu on the header → tap "Sort" → checkmark is shown next to "Due date" ✅
Back to "More" menu → tap "Group" → no item has a checkmark ❌, even though "None" has `state: 'on'`.
when switching between the 2, the checkmark seems to move from 1 to the other in the reproduction. In my case, both always have a state with a checkmark.

https://snack.expo.dev/@handsometim/react-native-screens-unstable_headerrightitems-checkmark-not-displayed-reproduction

### Screens version

4.24.0

### React Native version

0.83.2

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Architecture

Fabric (New Architecture)

### Build type

Release mode

### Device

Real device

### Device model

iPhone 15 pro

### Acknowledgements

Yes

Jasleen-Kaur96:
## Summary
- Fixes header menu behavior on iOS where selecting an item in one submenu clears the checkmark in a sibling submenu.
- Root cause: `UIMenuOptionsSingleSelection` was applied to menus that contain submenus, causing UIKit to treat sibling submenu actions as a shared selection group.
- This change scopes `singleSelection` to leaf menus only (menus without submenu children).

Fixes #3973.

## Repro
Using `unstable_headerRightItems`, create a header menu with:
- an inline parent submenu
- two sibling submenus (for example `Sort` and `Group`)
- each submenu contains action items with `state: 'on' | 'off'`
Observed before:
- selecting in submenu A removes checkmark in submenu B (and vice versa)
Expected / after:
- each submenu maintains its own selected checkmark independently

## Test plan
- Run app on iOS and open the header menu (`...`)
- Open submenu A (e.g. `Sort`) and select an option
- Open submenu B (e.g. `Group`) and select an option
- Verify both submenus keep their own checkmark state
- Swap submenu order in config and verify behavior remains correct

## Environment
- iOS: 26.3
- Device: iPhone 17 Pro

### DIFF
--- a/ios/RNSBarButtonItem.mm
+++ b/ios/RNSBarButtonItem.mm
@@ -4,7 +4,8 @@
 #import "RNSDefines.h"
 #import "RNSImageLoadingHelper.h"
 
-static UIMenuOptions RNSMakeUIMenuOptionsFromConfig(NSDictionary *config);
+static UIMenuOptions RNSMakeUIMenuOptionsFromConfig(NSDictionary *config, BOOL hasSubmenus);
+static BOOL RNSMenuHasSubmenus(NSArray *items);
 
 @implementation RNSBarButtonItem {
   NSString *_buttonId;
@@ -167,6 +168,7 @@ static UIMenuOptions RNSMakeUIMenuOptionsFromConfig(NSDictionary *config);
 {
   NSArray *items = dict[@"items"];
   NSMutableArray<UIMenuElement *> *elements = [NSMutableArray new];
+  BOOL hasSubmenus = RNSMenuHasSubmenus(items);
   if (items.count > 0) {
     for (NSDictionary *item in items) {
       NSString *menuId = item[@"menuId"];
@@ -191,7 +193,7 @@ static UIMenuOptions RNSMakeUIMenuOptionsFromConfig(NSDictionary *config);
   return [UIMenu menuWithTitle:dict[@"title"]
                          image:image
                     identifier:nil
-                       options:RNSMakeUIMenuOptionsFromConfig(dict)
+                       options:RNSMakeUIMenuOptionsFromConfig(dict, hasSubmenus)
                       children:elements];
 }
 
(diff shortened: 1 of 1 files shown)


kmichalikk:
Hi @Jasleen-Kaur96
Thanks for the PR but I'm afraid that we'll not proceed here. The `hasMenu` checking may be incorrect in some cases. From what I can see, for native iOS having singleSelection on parent menu gets this exact behavior of selecting only one element within all child menus. We'll keep investigating but this issue will probably end up with default values change at most.

PR #2319, opened 2024-08-26
fix(iOS): Change import of RectUtil.h file to react/renderer/components

tboba:
## Description

In the newest version of screens, there's a bug where compiler can't find RectUtil.h file from `utils/RectFile.h` during the compile phase of RNSScreenComponentDescriptor file.
This PR fixes this by changing the path of RectUtil.h file to <react/renderer/components/rnscreens/util/RectFile.h> file.

Fixes #2306.

## Changes

- Changed import of RectUtil.h file in RNSScreenComponentDescriptor.h file

## Test code and steps to reproduce

- Download reproducer from https://github.com/jankosecki/react-native-upgrade-tester/tree/rn-0.75
- Delete patches/react-native-screens+3.34.0.patch
- Clean up node_modules in case patch already applied
- Run yarn setup (it invokes "pod-install": "(cd ./ios ; RCT_NEW_ARCH_ENABLED=1 USE_FRAMEWORKS=static bundle exec pod install --repo-update)" which enables new architecture)
- Open XCode to set up development team (provided repro contains None team in Signing & Capabilities)
- Run yarn start
- Run yarn ios or trigger build from XCode (yarn ios causes a massive error output with different "error" lines but XCode offers a clear reason.

## Checklist

- [ ] Ensured that CI passes

Issue #2306, opened 2024-08-14
[RN 0.75 new arch] 'utils/RectUtil.h' file not found


jankosecki:
### Description

Hi team,

I've been testing upgrading RN 0.74 to 0.75-RC7 with basic stack including `react-navigation` and `react-native-screens`.
The iOS build fails for me with the following error:

```
In file included from /Users/jkosecki/repos/custom/ReactNativeUpgradeTester/node_modules/react-native-screens/ios/RNSScreen.mm:15:
/Users/jkosecki/repos/custom/ReactNativeUpgradeTester/node_modules/react-native-screens/common/cpp/react/renderer/components/rnscreens/RNSScreenComponentDescriptor.h:10:10: fatal error: 'utils/RectUtil.h' file not found
#include "utils/RectUtil.h"
```

I have applied the following patch locally (included in the reproduction repo):
```
diff --git a/node_modules/react-native-screens/RNScreens.podspec b/node_modules/react-native-screens/RNScreens.podspec
index be7cc24..14cd2e4 100644
--- a/node_modules/react-native-screens/RNScreens.podspec
+++ b/node_modules/react-native-screens/RNScreens.podspec
@@ -27,7 +27,7 @@ Pod::Spec.new do |s|
     s.subspec "common" do |ss|
       ss.source_files         = ["common/cpp/**/*.{cpp,h}", "cpp/**/*.{cpp,h}"]
       ss.header_dir           = "rnscreens"
-      ss.pod_target_xcconfig  = { "HEADER_SEARCH_PATHS" => "\"$(PODS_TARGET_SRCROOT)/common/cpp\"" }
+      ss.pod_target_xcconfig  = { "HEADER_SEARCH_PATHS" => "\"$(PODS_TARGET_SRCROOT)/common/cpp/react/renderer/components/\"" }
     end
   end
 
```
but I'm not sure if that's correct way to fix this issue.

### Steps to reproduce

1. Remove `patches/react-native-screens+3.34.0.patch`
2. Run `yarn setup && yarn ios`

### Snack or a link to a repository

https://github.com/jankosecki/react-native-upgrade-tester/tree/rn-0.75

### Screens version

3.34.0

### React Native version

0.75-RC7

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

None

### Acknowledgements

Yes

[...]

PR #3986, opened 2026-05-07
fix(iOS): forward-declare RCTSurfaceTouchHandler in UIView+RNSUtility.h

### LINKED ISSUE #3985: [iOS] UIView+RNSUtility.h leaks RCTSurfaceTouchHandler import, breaks Clang module dependency scanning under use_frameworks!:static
## Description

The public header `ios/utils/UIView+RNSUtility.h` directly imports `<React/RCTSurfaceTouchHandler.h>` and exposes `RCTSurfaceTouchHandler` as a return type in a category interface:

```objc
// ios/utils/UIView+RNSUtility.h
#import <React/RCTSurfaceTouchHandler.h>
// ...
- (nullable RCTSurfaceTouchHandler *)rnscreens_findTouchHandlerInAncestorChain;
```

`RCTSurfaceTouchHandler.h` is not actually shipped in the `React` clang module — it lives in `React-RCTFabric` (`RCTFabric.framework/Headers/React/RCTSurfaceTouchHandler.h`). RNScreens' own xcconfig adds the `RCTFabric` framework search path so RNScreens compiles fine in isolation, but any **consumer module** that imports RNScreens (e.g. ExpoRouter) inherits this transitive header import without inheriting the `RCTFabric` search path. Under `use_frameworks!:linkage => :static` + new arch, Clang's module dependency scanning then fails before compilation even starts:

```
error: Clang dependency scanning failure: While building module 'RNScreens'
imported from RNScreens-6d5e763e.input:1:
  UIView+RNSUtility.h:4:9 'React/RCTSurfaceTouchHandler.h' file not found
error: Compilation search paths unable to resolve module dependency: 'RNScreens'
(in target 'ExpoRouter' from project 'Pods')
```

The root cause is that a public header should not surface fabric-internal types as part of its module interface. A forward declaration in the header plus a real `#import` in the corresponding `.mm` is sufficient for RNScreens' own implementation, and keeps the public surface free of cross-module header dependencies. This matches the pattern already used by `RNSScreenStack.mm` and the sibling `UINavigationBar+RNSUtility.h` (which only imports `<UIKit/UIKit.h>` in its public header).

## Steps to reproduce

1. New Expo project on RN 0.85.3 with new arch (fabric-only) enabled.
2. Add `expo-build-properties` with:
   ```json
   {
     "ios": { "useFrameworks": "static" }
   }
   ```
3. Install `react-native-screens@4.25.0-beta.1` and `expo-router` (or any module that imports the `RNScreens` clang module).
4. `npx expo prebuild --clean && npx expo run:ios`.

## Snack or a link to a repository

Happy to put together a minimal repro repo if needed — the configuration is essentially `bare-minimum` Expo template + the three lines of `expo-build-properties` change above.

## Screens version

4.25.0-beta.1 (also reproduces on current `main` — verified that `ios/utils/UIView+RNSUtility.h` still contains the direct `#import <React/RCTSurfaceTouchHandler.h>` at HEAD `9744779`)

## React Native version

0.85.3

## Platforms

iOS

## JavaScript runtime

Hermes

## Workflow

Expo (bare / prebuild)

## Architecture

Fabric (New Architecture)

## Build type

Debug mode

## Device

iOS Simulator

## Acknowledgements

Yes

---

## Suggested fix

Forward-declare in the header and move the `#import` to the implementation file:

```objc
// ios/utils/UIView+RNSUtility.h
#import <UIKit/UIKit.h>
@class RCTSurfaceTouchHandler;   // instead of #import <React/RCTSurfaceTouchHandler.h>
```

```objc
// ios/utils/UIView+RNSUtility.mm
#import <React/RCTSurfaceTouchHandler.h>   // moved from header
```

Note: `<Foundation/Foundation.h>` is replaced with `<UIKit/UIKit.h>` because `RCTSurfaceTouchHandler.h` was previously providing UIKit transitively, and the file declares a `UIView` category. This matches the convention used by the sibling header `ios/utils/UINavigationBar+RNSUtility.h`.

I verified locally that the change builds cleanly under `use_frameworks!:static` + new arch and unblocks consumer module dependency scanning.

All current internal consumers of `UIView+RNSUtility.h` (`RNSScreenStack.mm`, `RNSScreen.mm`, `UIView+RNSUtility.mm` itself) already `#import <React/RCTSurfaceTouchHandler.h>` directly, so the change is self-contained and does not introduce a new transitive-import requirement on existing call sites.

## Related context

This is in the same family of `use_frameworks!:static` + new arch header-purity issues as expo/expo#39080 (where `expo-router` imports RNScreens internal headers). Different direction (consumer → RNScreens vs. RNScreens → React fabric), same underlying constraint: under static frameworks + Clang module dependency scanning, public-header transitive imports must resolve through the consumer's module search paths, not the producer's.

It is also conceptually adjacent to #2306 / #2319 (`RectUtil.h` import path), where moving an import to a fabric-renderer path resolved a similar consumer-side compilation failure under static frameworks.

## Patch diff

```diff
diff --git a/ios/utils/UIView+RNSUtility.h b/ios/utils/UIView+RNSUtility.h
--- a/ios/utils/UIView+RNSUtility.h
+++ b/ios/utils/UIView+RNSUtility.h
@@ -1,7 +1,8 @@
 #pragma once

-#import <Foundation/Foundation.h>
-#import <React/RCTSurfaceTouchHandler.h>
+#import <UIKit/UIKit.h>
+
+@class RCTSurfaceTouchHandler;

 NS_ASSUME_NONNULL_BEGIN

diff --git a/ios/utils/UIView+RNSUtility.mm b/ios/utils/UIView+RNSUtility.mm
--- a/ios/utils/UIView+RNSUtility.mm
+++ b/ios/utils/UIView+RNSUtility.mm
@@ -1,5 +1,6 @@
 #import "UIView+RNSUtility.h"

+#import <React/RCTSurfaceTouchHandler.h>
 #import <React/RCTSurfaceView.h>
 #import "RNSModalScreen.h"
```

zhe-qi:
## Summary

`ios/utils/UIView+RNSUtility.h` declares methods returning `RCTSurfaceTouchHandler *` and currently `#import`s `<React/RCTSurfaceTouchHandler.h>` directly in the public header. That header is shipped in the `React-RCTFabric` clang module, not `React` — so under `use_frameworks!:linkage => :static` + new arch, any consumer module that imports the `RNScreens` clang module (e.g. ExpoRouter) fails Clang's module dependency scanning before compilation:

```
error: Clang dependency scanning failure: While building module 'RNScreens'
  UIView+RNSUtility.h: 'React/RCTSurfaceTouchHandler.h' file not found
error: Compilation search paths unable to resolve module dependency: 'RNScreens'
(in target 'ExpoRouter' from project 'Pods')
```

This PR moves the import out of the public surface:

- `UIView+RNSUtility.h`: replace `#import <React/RCTSurfaceTouchHandler.h>` with `@class RCTSurfaceTouchHandler;` forward declaration. Replace `<Foundation/Foundation.h>` with `<UIKit/UIKit.h>` (since the fabric header was previously providing UIKit transitively, and this is a `UIView` category — matching the sibling `UINavigationBar+RNSUtility.h`).
- `UIView+RNSUtility.mm`: add the real `#import <React/RCTSurfaceTouchHandler.h>`.

This matches the pattern already used by `RNSScreenStack.mm`, where fabric-internal imports stay inside the implementation file.

Fixes #3985.

## Compatibility

All current internal consumers of `UIView+RNSUtility.h` (`RNSScreenStack.mm`, `RNSScreen.mm`, and `UIView+RNSUtility.mm` itself) already `#import <React/RCTSurfaceTouchHandler.h>` directly, so they retain access to the type without relying on transitive imports from the public header. No call sites need updating.

The change is purely a compile-time refactor — no symbols, return types, ABI, or runtime behavior change.

## Test plan

- [x] Verified on the consumer side: under `use_frameworks!:static` + new arch + RN 0.85.3, the patched header allows ExpoRouter to compile against RNScreens. Without the fix, `expo run:ios` fails on the Clang module dependency scanning error above.
- [ ] Maintainers: existing CI matrix should cover both `use_frameworks!` and non-frameworks builds across new/legacy arch.

## Related

- See #3985 for the full failure trace and reproduction steps.
- Conceptually adjacent to #2306 / #2319 (`RectUtil.h` import path), which fixed a similar consumer-side compilation failure by relocating a header import.
- Same family as expo/expo#39080 (consumer → RNScreens internal headers), but in the opposite direction (RNScreens → React fabric).

### DIFF
--- a/ios/utils/UIView+RNSUtility.h
+++ b/ios/utils/UIView+RNSUtility.h
@@ -1,7 +1,8 @@
 #pragma once
 
-#import <Foundation/Foundation.h>
-#import <React/RCTSurfaceTouchHandler.h>
+#import <UIKit/UIKit.h>
+
+@class RCTSurfaceTouchHandler;
 
 NS_ASSUME_NONNULL_BEGIN
 
--- a/ios/utils/UIView+RNSUtility.mm
+++ b/ios/utils/UIView+RNSUtility.mm
@@ -1,5 +1,6 @@
 #import "UIView+RNSUtility.h"
 
+#import <React/RCTSurfaceTouchHandler.h>
 #import <React/RCTSurfaceView.h>
 #import "RNSModalScreen.h"
 


ios/utils/UIView+RNSUtility.h
```diff
@@ -1,7 +1,8 @@
 #pragma once
 
-#import <Foundation/Foundation.h>
-#import <React/RCTSurfaceTouchHandler.h>
+#import <UIKit/UIKit.h>
```
kkafar, line 3: Why change Foundation import for UIKit import? How is it related to the issue?

kkafar (commented):
Hey, thanks for the PR. Seems like something I'll land. I just have one question.

ios/utils/UIView+RNSUtility.h
kligarski: I think it's explained in https://github.com/software-mansion/react-native-screens/issues/3985 ("Suggested fix" section).

kkafar: Thanks. Makes sense.

kkafar (approved):
I'm good with this change. @kligarski would you mind giving a round of test (build with the configuration the OP suggests) & in case of success landing it?

kligarski:
> I'm good with this change. @kligarski would you mind giving a round of test (build with the configuration the OP suggests) & in case of success landing it?

Okay, I'll test it on Monday.

kkafar (approved):
I gave it a round of testing & seems to work fine. Thank you.

I'll cherry-pick this change to 4.25.0 release.

PR #4024, opened 2026-05-13
refactor(test): emulator set to pixel 8 

LKuchno:
## Description

Pixel 8 is latest supported emulator/device for api level 34. As we are currently no update api level to 36 due to problem with test on stack v4 after changing OnBackPressedDispatcher settings https://github.com/software-mansion/react-native-screens-labs/issues/1459 we need to use emulator for api level 34. 

It will be updated to pixel9 (or newer) when we will change API level for tests.

## Changes

Emulator in workflow updated to Pixel 8.


kkafar (approved):
CI passes. 

✅

PR #4058, opened 2026-05-18
refactor(iOS, FormSheet v5): Move RCTTouchHandler to ContentView

t0maboro:
## Description

Manage RCTSurfaceTouchHandler lifecycle on ContentView attach/detach to the window. Move touch handler content origin synchronization to ContetView. Left only hitTest override on the Host level (strictly related to the Host), the rest of touch handling logic was removed.

## Changes

- RCTSurfaceTouchHandler lifecycle was moved from Host to ContentView.

N/A - refactor

## Test plan

All SFTs should work the same

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/gamma/modals/form-sheet/RNSFormSheetAppearanceApplicator.h
+++ b/ios/gamma/modals/form-sheet/RNSFormSheetAppearanceApplicator.h
@@ -1,0 +1,21 @@
+#pragma once
+
+#import <Foundation/Foundation.h>
+
+@class RNSFormSheetAppearanceCoordinator;
+@class RNSFormSheetContentController;
+@class RNSFormSheetHostComponentView;
+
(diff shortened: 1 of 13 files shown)


kligarski (commented):
Why do we want to keep React-related logic (`RCTSurfaceTouchHandler`) inside native part of implementation (`RNSFormSheetContentView` which is not a React view)?Shouldn't we use delegate pattern here to expose the callbacks and values instead of directly configuring `RCTSurfaceTouchHandler`?

t0maboro:
closing, RCTSurfaceTouchHandler setup needs to be done after implementing fitToContents, I need some native component on which I can setup it, as for now, it should remain on Host to prevent RN logic leak to native

PR #4059, opened 2026-05-18
chore(Example): ScenarioDescription refactor

sgaczol:
## Description

This PR refactors test scenario structure by extraction `scenarioDescription` into dedicated files.

## Changes

- extract scenarioDescription objects from `index.tsx` into separate `scenario-description.ts` files as default exports
- move standalone `.tsx` test files into `<test-name>/index.tsx + scenario-description.ts` directory structure - some of these new directories don't follow the naming convention as they will be verified anyway
- add `smokeTest` and `e2eCoverage` properties to `ScenarioDescription` type - those properties are currently set in all objects to `false` and `'tbd'` respectively
- display `smokeTest` and `e2eCoverage` indicators in the scenario menu

As below, if a scenario is marked as a smoke test, an orange `s` is shown. For e2e coverage, a purple `e2e` indicates full coverage.

smokeTest and e2eCoverage

The image is only an example of how indicators are displayed and does not necessarily reflect actual property values assigned to the scenarios.

## Test plan

N/A

## Checklist

- [ ] Ensured that CI passes


apps/src/tests/shared/helpers.ts
```diff
@@ -21,6 +21,17 @@ export interface ScenarioDescription {
    * What platforms does this test cover.
    */
   platforms?: ('android' | 'ios')[];
+  /**
+   * Indicates the current status of end-to-end (E2E) test coverage for this scenario.
+   * - 'full': The scenario is completely covered by E2E tests.
+   * - 'incomplete': The scenario is only partially covered.
```
LKuchno: This value indicates whether a scenario is partially covered or, for some reason, not covered at all. We should explicitly mention that in either case, developers/testers should familiarize themselves with the scenario and manually complete the missing parts. Scenarios with a complete lack of coverage fall into this category as well, since full automation isn't always feasible. 

My suggestion is: 
'incomplete': The scenario is only partially covered or not covered at all. Familiarize yourself with the scenario and manually complete the relevant parts.

or just first sentence

LKuchno (changes requested):
Thank you!
Just small correction is needed in e2eCoverage documentation :)

apps/src/tests/shared/helpers.ts
sgaczol: done: https://github.com/software-mansion/react-native-screens/pull/4059/commits/0c0ae2efa5e891da22aba26b31a9fd571c6b027c

Also I've made some changes regarding `ScenarioButton` for better transparency, as updated in the visual documentation.

LKuchno (approved):
LGTM! ✅

apps/src/tests/shared/ScenarioButton.tsx
```diff
@@ -67,4 +73,16 @@ const styles = StyleSheet.create({
...
+    padding: 4,
+    fontWeight: '900',
+    color: 'orange',
+  },
+  e2eFull: {
+    fontSize: 18,
+    padding: 4,
+    fontWeight: '900',
+    color: 'purple',
+  },
```
kligarski, line 87: I wonder if we should also have a marker in different color for tests that are partially covered so that you can test them manually after checking tests without e2e.

kligarski (commented):
The code looks good but I'm not up to date with current convention for testing so I don't want to decide anything here.

apps/src/tests/shared/ScenarioButton.tsx
sgaczol: Such approach was implemented in the initial version of this PR, but I thought that it could be a bit confusing. Wdyt?

kkafar: I think that the action required for tests that have partial e2e coverage and no coverage at all is the same - you need to checkout scenario and determine what needs to be tested manually. Therefore I don't think that we need to distinguish between them right now. 

If such need arises in the future, we can consider adding it later. Atm. I find it redundant. 

tldr: show no icon for partial/missing e2e.

sgaczol: alright

PR #4065, opened 2026-05-19
fix(Android, FormSheet): include keyboard translationY in shadow update

### LINKED ISSUE #4064: [Android][Fabric] FormSheet `contentOffsetY` desyncs from keyboard `translationY`
### Description

# [Android][Fabric] FormSheet `contentOffsetY` desyncs from keyboard `translationY` → Pressables become unresponsive

## Summary

On Android with the **New Architecture (Fabric)**, any `Pressable` inside a
FormSheet stops receiving touches once the soft keyboard is open and a layout
pass occurs that does not carry the keyboard `translationY`. Closing the
keyboard makes the controls responsive again.

iOS is unaffected.

## Environment

|                                | Version                                           |
| ------------------------------ | ------------------------------------------------- |
| react-native-screens           | 4.25.1 (also confirmed on 4.24.0)                 |
| react-native                   | 0.83.6                                            |
| react                          | 19.2.0                                            |
| expo                           | SDK 55 (~55.0.24)                                 |
| @react-navigation/native       | ^7.2.4                                            |
| @react-navigation/native-stack | ^7.15.1                                           |
| Architecture                   | New Architecture / Fabric (required to reproduce) |
| Platform                       | Android (iOS not affected)                        |

## Minimal reproduction

Standalone repro project (Expo dev build, `react-native-screens@4.25.1` pinned):
https://github.com/jiroscripts/rns-formsheet-repro

Deterministic scenario — no navigation, no focus race:

1. A native-stack screen presented with `presentation: 'formSheet'`
   (`sheetAllowedDetents: [0.8]`).
2. Its content is a `TextInput` with `autoFocus` + `secureTextEntry`, plus a
   `Pressable` with a visible counter.
3. Open the FormSheet → the field autofocuses, the keyboard opens (the sheet is
   shifted up via `translationY`).
4. Toggle `secureTextEntry` 2–3 times (reconfigures the Android IME).
5. Tap the `Pressable`: **the counter does not increment** — the touch is not
   delivered.
6. Dismiss the keyboard → the `Pressable` works again immediately.

### Expected

The `Pressable` receives touches while the keyboard is open.

### Actual

The `Pressable` (and the area around it) is unresponsive until the keyboard is
dismissed.

## Root cause

On Fabric, react-native-screens reports the FormSheet content position to the
shadow tree so RN's JS touch responder knows where the view is. While the
keyboard is open the sheet is shifted by `translationY`. Two code paths in
`Screen.kt` report that position, and they disagree:

| Path                                                        | Offset sent to Fabric        | Correct?                  |
| ----------------------------------------------------------- | ---------------------------- | ------------------------- |
| `Screen.onSheetYTranslationChanged()` (`Screen.kt:546`)     | `top + translationY.toInt()` | ✅ real visual position   |
| `Screen.onBottomSheetBehaviorDidLayout()` (`Screen.kt:251`) | `top` only                   | ❌ ignores `translationY` |

`onBottomSheetBehaviorDidLayout()` is invoked on every relevant layout pass:

- `ScreensCoordinatorLayout.onLayout(changed)` → `onBottomSheetBehaviorDidLayout(changed)` (`ScreensCoordinatorLayout.kt:96`)
- `ScreenStackFragment.handleInsetsUpdateAndNotifyTransition()` → `onBottomSheetBehaviorDidLayout(true)` (`ScreenStackFragment.kt:572`)

When the keyboard offset (`translationY`) is active and one of these layout
passes fires, Fabric receives `contentOffsetY = top` (without `translationY`).
No keyboard animation follows to re-run `onSheetYTranslationChanged()` and
re-correct it, so the Fabric `contentOffsetY` stays desynced from the real
visual position by the keyboard height. RN routes JS touches using the Fabric
position → taps on the visible `Pressable` miss it.

The `delta = 0.9f` early-return guard in `FabricEnabledViewGroup.updateState()`
does not help — the two values differ by the full keyboard offset, so the
incorrect write goes through (last-writer-wins).

The defect is structurally tied to the keyboard `translationY` mechanism
introduced by **PR #3248** ("Moving formsheet above keyboard", 4.19.0): without
that `translationY`, the two paths would agree. It therefore likely originates
there, but we have **not verified this against 4.19.0 directly** — we only
confirmed the bug on **4.24.0 and 4.25.1** (the method was renamed
`dispatchShadowStateUpdate`/`updateScreenSizeFabric` → `updateShadowNodeScreenSize`
between those versions; the logic and the bug are identical in both).

## Suggested fix

Align `onBottomSheetBehaviorDidLayout()` with `onSheetYTranslationChanged()`:

```kotlin
// Screen.kt, in onBottomSheetBehaviorDidLayout()
if (coordinatorLayoutDidChange) {
    updateShadowNodeScreenSize(width, height, top + translationY.toInt())
}
```

When `translationY == 0` (no keyboard) this is identical to the current
behavior (`top + 0 == top`), so there is no regression for the no-keyboard
case. PR with this change to follow.

## A second, harder-to-reproduce manifestation (nested navigation)

We originally hit this in production via a different trigger: a native stack
**nested inside a FormSheet** (`ScreenA → ScreenB → back`) where the keyboard
stays open across the back navigation. The same Fabric desync occurs and the
**same one-line fix resolves it** (verified in our app).

We were **not able to isolate this path into a standalone repro.** Its
precondition is that the keyboard remains open across the nested pop, which we
traced to focus/teardown ordering: React unmounts the popped screen (clearing
its native focus, without dismissing the IME — detaching an `EditText` does not
hide the keyboard) **before** `ScreenContainer.removeView()` runs, so the
keyboard-dismiss safety-net there (`if (view === focusedChild) hideSoftInputFromWindow(...)`)
sees `focusedChild == null` and is skipped. In a bare project the ordering is
reversed (the popped screen still holds focus at `removeView` time, the
safety-net fires, the keyboard closes) — so the precondition is not met. We
could not pin down the exact factor that flips this ordering (navigation depth /
provider context); we ruled out screen freezing, custom animations,
`detachInactiveScreens`, static vs dynamic navigation API, async navigation,
and `synchronousScreenUpdatesEnabled`.

We mention it for completeness: it is the **same root cause and the same fix**;
the `secureTextEntry` repro above isolates the defect deterministically without
this fragile precondition. As a separate observation, the
`view === focusedChild` guard in `ScreenContainer.removeView()` looks racy for
nested/FormSheet pops, but that is independent of the fix proposed here.

---

> _Disclaimer: the investigation and the write-up of this issue were produced
> with the assistance of Claude (Anthropic's AI). The standalone reproduction
> and the proposed fix were both verified on a physical Android device by a
> human before filing._

### Steps to reproduce

1. App opens on Home. Tap “Open FormSheet” → the sheet opens and the password field autofocuses (keyboard opens).
2. Tap the “Show / Hide” toggle 2–3 times (flips secureTextEntry, which reconfigures the Android IME → triggers the offending layout pass while the keyboard stays open).
3. Tap the red “TAP ME” button.

https://github.com/jiroscripts/rns-formsheet-repro

### Screens version

3.25.1

### React Native version

0.83.6

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

Fabric (New Architecture)

### Build type

None

### Device

Real device

### Device model

_No response_

### Acknowledgements

Yes

jiroscripts:
# fix(android/Fabric): include keyboard `translationY` in `onBottomSheetBehaviorDidLayout` shadow update

Fixes #4064

## Summary

On Android + New Architecture, a `Pressable` inside a FormSheet becomes
unresponsive while the keyboard is open. `Screen.onBottomSheetBehaviorDidLayout()`
reports the sheet content position to Fabric as `top` only, whereas
`Screen.onSheetYTranslationChanged()` reports `top + translationY`. When the
keyboard `translationY` offset is active and a layout pass goes through
`onBottomSheetBehaviorDidLayout()`, Fabric's `contentOffsetY` desyncs from the
real visual position by the keyboard height, so RN's JS touch responder no
longer matches where the view is rendered.

This aligns the two code paths.

## Change

```diff
  // android/.../com/swmansion/rnscreens/Screen.kt — onBottomSheetBehaviorDidLayout()
  if (coordinatorLayoutDidChange) {
-     updateShadowNodeScreenSize(width, height, top)
+     // Keep consistent with onSheetYTranslationChanged(): the FormSheet may be
+     // shifted by translationY to stay above the keyboard. Sending `top` alone
+     // desyncs the Fabric contentOffsetY from the native visual position.
+     updateShadowNodeScreenSize(width, height, top + translationY.toInt())
  }
```

This is exactly the value `onSheetYTranslationChanged()` already sends, so the
two reporting paths now agree. It also fixes the hard-coded
`onBottomSheetBehaviorDidLayout(true)` caller in
`ScreenStackFragment.handleInsetsUpdateAndNotifyTransition()`.

## Why this is safe

- When `translationY == 0` (no keyboard offset), `top + 0 == top` → behavior is
  **identical** to before. No change for the common case.
- The shadow update is `IS_NEW_ARCHITECTURE_ENABLED`-gated → **Paper is
  unaffected**.
- Single line; no API change; no new allocation.
- Likely tied to PR #3248 (4.19.0), which introduced the keyboard `translationY`
  mechanism without propagating the offset to this layout path (not verified
  against 4.19.0 directly; confirmed on 4.24.0 and 4.25.1). This restores
  consistency between the two paths rather than adding a workaround.

## Test plan

Verified with the standalone repro (Expo SDK 55, RN 0.83.6,
`react-native-screens@4.25.1` pinned, New Arch, Android):

- [x] FormSheet + `autoFocus` `secureTextEntry` input + `Pressable`, single
      detent `[0.8]`: **before** — `Pressable` dead after toggling
      `secureTextEntry` with keyboard open; **after** — works normally.
- [x] Same fix independently verified in a production app (both the
      `secureTextEntry` path and a nested-navigation path that exhibits the
      same desync).
- [x] No-keyboard case unaffected (manual check; `top + 0 == top`).

Not yet covered (happy to extend or get reviewer guidance):

- [ ] `sheetAllowedDetents: 'fitToContents'`
- [ ] 2- and 3-detent sheets
- [ ] Additional Android API levels
- [ ] iOS (expected: unaffected — Android-only path)
- [ ] New Architecture disabled (expected: unaffected — Fabric-gated)

## Notes

There is a related, separate observation (not addressed here): the
`view === focusedChild` guard in `ScreenContainer.removeView()` appears racy for
screens popped inside a nested/FormSheet stack (focus can be cleared by React
unmount before `removeView` runs). That only affects whether the keyboard stays
open; the desync itself — and this fix — are independent of it. Happy to file a
follow-up if useful.

---

> _Disclaimer: the root-cause investigation and this description were produced
> with the assistance of Claude (Anthropic's AI). The one-line change was
> verified on a physical Android device — bug reproduced on pristine 4.25.1,
> resolved after the fix — both in a standalone repro and in a production app,
> before opening this PR._

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
@@ -248,7 +248,7 @@ class Screen(
         }
 
         if (coordinatorLayoutDidChange) {
-            updateShadowNodeScreenSize(width, height, top)
+            updateShadowNodeScreenSize(width, height, top + translationY.toInt())
         }
 
         footer?.onParentLayout(coordinatorLayoutDidChange, left, top, right, bottom, container!!.height)
--- a/apps/src/tests/issue-tests/Test4064.tsx
+++ b/apps/src/tests/issue-tests/Test4064.tsx
@@ -1,0 +1,109 @@
+import React, { useState } from 'react';
+import {
+  Button,
+  Platform,
+  StyleSheet,
+  Text,
+  TextInput,
+  View,
+} from 'react-native';
+import {
+  createStaticNavigation,
+  useNavigation,
+} from '@react-navigation/native';
+import { createNativeStackNavigator } from '@react-navigation/native-stack';
+import { SafeAreaProvider } from 'react-native-safe-area-context';
+import { Colors } from '@apps/shared/styling';
+
+export function SecureForm() {
+  const [secure, setSecure] = useState(true);
+  const [count, setCount] = useState(0);
+
+  return (
+    <View style={styles.container}>
+      <TextInput
+        style={styles.input}
+        autoFocus
+        secureTextEntry={secure}
+        placeholder="Password"
+      />
+      <Button
+        title={secure ? 'Show Password' : 'Hide Password'}
+        onPress={() => setSecure(s => !s)}
+      />
+      <Button
+        title={`TAP ME — count: ${count}`}
+        onPress={() => setCount(c => c + 1)}
+      />
+    </View>
+  );
+}
+
+export function Home() {
+  const navigation = useNavigation<any>();
+  return (
+    <View style={styles.container}>
+      <Text style={styles.title}>FormSheet</Text>
+      <Text style={styles.hint}>
+        1. The password field below is autofocused (keyboard open){'\n'}
+        2. Tap "Show / Hide" 2–3 times{'\n'}
+        3. Tap "TAP ME" — counter is frozen until the keyboard closes
+      </Text>
+      <Button
+        title="Open FormSheet"
+        onPress={() => navigation.navigate('Sheet')}
+      />
+    </View>
+  );
+}
+
+const RootNavigator = createNativeStackNavigator({
+  screens: {
+    Home: { screen: Home, options: { headerShown: false } },
+    Sheet: {
+      screen: SecureForm,
+      options: {
+        presentation: 'formSheet',
+        sheetAllowedDetents: [0.8],
+        headerShown: Platform.OS === 'android',
+        headerTitle: 'Auth',
+      },
+    },
+  },
+});
+
+const Navigation = createStaticNavigation(RootNavigator);
+
+export default function App() {
+  return (
+    <SafeAreaProvider>
+      <Navigation />
+    </SafeAreaProvider>
+  );
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    padding: 20,
+    gap: 16,
+    justifyContent: 'center',
+    backgroundColor: Colors.White,
+  },
+  title: {
+    fontSize: 22,
+    fontWeight: '700',
+  },
+  hint: {
+    fontSize: 14,
+    color: Colors.text,
+    lineHeight: 20,
+  },
+  input: {
+    borderWidth: 1,
+    borderColor: Colors.NavyDark100,
+    borderRadius: 8,
+    padding: 12,
+    fontSize: 16,
+  },
+});
--- a/apps/src/tests/issue-tests/index.ts
+++ b/apps/src/tests/issue-tests/index.ts
@@ -193,6 +193,7 @@ export { default as Test3867 } from './Test3867';
 export { default as Test3885 } from './Test3885';
 export { default as Test3910 } from './Test3910';
 export { default as Test4027 } from './Test4027';
+export { default as Test4064 } from './Test4064';
 export { default as TestScreenAnimation } from './TestScreenAnimation';
 // The following test was meant to demo the "go back" gesture using Reanimated
 // but the associated PR in react-navigation is currently put on hold


t0maboro (approved):
Makes sense, thanks @jiroscripts for fixing that, let me add repro to our examples, and we can proceed with that

PR #4072, opened 2026-05-20
refactor(test): expand e2e coverage for tab bar layout direction

LKuchno:
## Summary

- Expands the `test-tabs-tab-bar-layout-direction` Detox suite to cover two previously missing iOS-only scenarios: system LTR + React Native forced RTL, and system RTL + React Native `allowRTL: false` (forced LTR). These scenarios test the interaction between the OS-level writing direction and the React Native `I18nManager` overrides independently, which the previous cross-platform tests could not exercise.
- Fixes the `beforeEach` → `beforeAll` regression in the LTR suite's setup hook, preventing redundant app reloads between tests.
- Corrects label assertions in both default-state tests: the `direction` picker now correctly initialises to `ltr` / `rtl` (reflecting the resolved effective direction) rather than the previously expected `inherit`.
- Removed unnecessary assertions that could be misleading; forceRTL and allowRTL do not reflect the set values, and the checkboxes display default values if not toggled.
- Removed the AppleTextDirection and I18NIsRTL attributes from launchApp, as they are not required for this test.
- Replace launchApp() function with launchAppWithAttributes()
- Updates the scenario's `E2E test` field from `Other: Covers first two manual scenarios` to `Yes: Covers all scenarios`, and removes the stale note about iOS-specific tests being pending.

## Changes

- **E2E/Tabs**: Added two new `describeIfiOS` blocks covering (1) system LTR + RN forced RTL and (2) system RTL + RN `allowRTL: false`, each with default-state, explicit direction override, and cycle-through tests
- **E2E/Tabs**: Changed `beforeEach` to `beforeAll` in the LTR suite to avoid redundant `reloadReactNative` calls between individual test cases
- **E2E/Tabs**: Removed unnecessary assertions that could be misleading; forceRTL and allowRTL do not reflect the set values, and the checkboxes display default values if not toggled.
- **E2E/Tabs**: Removed the AppleTextDirection and I18NIsRTL attributes from launchApp, as they are not required for this test.
- **E2E/Tabs**: Replace launchApp() function with launchAppWithAttributes()
- **E2E/Tabs**: Corrected `toHaveLabel` assertions for the direction picker default state (`'direction: ltr'` / `'direction: rtl'` instead of `'direction: inherit'`)
- **E2E/Tabs**: Renamed describe blocks to `system/RN settings: LTR|RTL` to clarify they cover both OS and React Native direction settings
- **Scenario**: Updated `test-tabs-tab-bar-layout-direction/scenario.md` to mark E2E coverage as complete and remove the "iOS tests pending" note


FabricExample/e2e/single-feature-tests/tabs/test-tabs-tab-bar-layout-direction.e2e.ts
```diff
@@ -159,37 +159,40 @@ describe('Tab Bar Layout Direction - system settings: RTL', () => {
...
+
+    await expect(element(by.id('is-rtl-information'))).toHaveText(
+      'I18nManager.isRTL == true',
+    );
     await expect(element(by.id('react-force-rtl-picker'))).toHaveLabel(
       'forceRTL: false',
     );
     await expect(element(by.id('react-allow-rtl-picker'))).toHaveLabel(
       'allowRTL: true',
     );
```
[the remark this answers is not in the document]
LKuchno: It's as expected, we want to set RN RTL and keep it set for this test suites. afterAll function relaunch app setting direction back to LTR.

FabricExample/e2e/single-feature-tests/tabs/test-tabs-tab-bar-layout-direction.e2e.ts
```diff
@@ -207,3 +210,174 @@ describe('Tab Bar Layout Direction - system settings: RTL', () => {
...
 });
+
+describeIfiOS(
+  'iOS only: Tab Bar Layout Direction - system settings: RTL and RN settings: LTR',
+  () => {
+    beforeAll(async () => {
+      await launchAppWithAttributes({
+        AppleTextDirection: 'YES',
+        NSForceRightToLeftWritingDirection: 'YES',
+        I18NIsRTL: 'YES',
```
kligarski: I'm not sure why do we need this? (`I18NIsRTL`) Does it do anything? In the test we're expecting `I18nManager.isRTL == false` so it's the opposite.

FabricExample/e2e/single-feature-tests/tabs/test-tabs-tab-bar-layout-direction.e2e.ts
```diff
@@ -207,3 +210,174 @@ describe('Tab Bar Layout Direction - system settings: RTL', () => {
...
+
+      await expect(element(by.id('is-rtl-information'))).toHaveText(
+        'I18nManager.isRTL == true',
+      );
+      await expect(element(by.id('react-force-rtl-picker'))).toHaveLabel(
+        'forceRTL: false',
+      );
+      await expect(element(by.id('react-allow-rtl-picker'))).toHaveLabel(
+        'allowRTL: true',
+      );
```
kligarski: Should we assert the value in these pickers if they don't have accurate information? E.g. in this test we're using `forceRTL = true` and here we are asserting something quite opposite which is a bit misleading.

kligarski (commented):
Looks good but I have 2 questions.

FabricExample/e2e/single-feature-tests/tabs/test-tabs-tab-bar-layout-direction.e2e.ts
LKuchno: To avoid confusion, I removed this check from both here and the full System/RN RTL test. It was originally added just to verify if the screen loaded with the 'default' value, but the manual scenario doesn't mention these flags.
https://github.com/software-mansion/react-native-screens/pull/4072/commits/1a6239062fffd10fe2fd4cc24394010094dd94ee

FabricExample/e2e/single-feature-tests/tabs/test-tabs-tab-bar-layout-direction.e2e.ts
LKuchno: I removed this attribute, setting it to NO with other cofnigurations result with final value I18nManager.isRTL == true so next steps were needed anyway.
https://github.com/software-mansion/react-native-screens/pull/4072/commits/1a6239062fffd10fe2fd4cc24394010094dd94ee

apps/src/tests/single-feature-tests/tabs/test-tabs-tab-bar-layout-direction/scenario.md
```diff
@@ -25,16 +25,14 @@ Both platforms validate `inherit`, `ltr`, and `rtl` prop values.
 
 ## E2E test
 
-Other: Covers first two manual scenarios for LTR/RTL configurations:
+Full: Covers all scenarios for LTR/RTL configurations:
```
kligarski, line 28: btw. should we change our `template/scenario.md`? We have `yes/no/other` there but here we're using `full`. But that's for another PR.

kligarski (approved):
Looks good. Let's wait for the CI to finish.

PR #4074, opened 2026-05-21
refactor(iOS, FormSheet v5): Move AppearanceCoordinator/Applicator to ContentController

t0maboro:
## Description

Previously, there was a tight coupling where `RNSFormSheetContentController` and `RNSFormSheetHostComponentView` had a direct dependency. By introducing dedicated provider protocols, the UIKit is now completely agnostic of React.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1454

## Changes

- defining contracts that Host now implements to supply data.
- removed all `RNSFormSheetHostComponentView` dependencies from `RNSFormSheetContentController` and `RNSFormSheetAppearanceApplicator`
- moved invalidation flags logic from the Host to the Controller

N/A - refactor

## Test plan

No regression in existing examples

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/gamma/modals/form-sheet/RNSFormSheetAppearanceApplicator.h
+++ b/ios/gamma/modals/form-sheet/RNSFormSheetAppearanceApplicator.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #import <Foundation/Foundation.h>
+#import "RNSFormSheetProviders.h"
 
 @class RNSFormSheetAppearanceCoordinator;
 @class RNSFormSheetContentController;
@@ -10,9 +11,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface RNSFormSheetAppearanceApplicator : NSObject
 
-- (void)updateAppearanceIfNeededForHost:(RNSFormSheetHostComponentView *)host
-                             controller:(RNSFormSheetContentController *)controller
-                            coordinator:(RNSFormSheetAppearanceCoordinator *)coordinator;
+- (void)updateAppearanceIfNeededForAppearanceProvider:(id<RNSFormSheetAppearanceProvider>)appearanceProvider
+                                     behaviorProvider:(id<RNSFormSheetBehaviorProvider>)behaviorProvider
+                                           controller:(RNSFormSheetContentController *)controller
+                                          coordinator:(RNSFormSheetAppearanceCoordinator *)coordinator;
 
 - (void)resetInitialDetent;
 
--- a/ios/gamma/modals/form-sheet/RNSFormSheetAppearanceApplicator.h
+++ b/ios/gamma/modals/form-sheet/RNSFormSheetAppearanceApplicator.h
@@ -10,10 +10,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface RNSFormSheetAppearanceApplicator : NSObject
 
-- (void)updateAppearanceIfNeededForAppearanceProvider:(id<RNSFormSheetAppearanceProvider>)appearanceProvider
-                                     behaviorProvider:(id<RNSFormSheetBehaviorProvider>)behaviorProvider
-                                           controller:(RNSFormSheetContentController *)controller
-                                          coordinator:(RNSFormSheetAppearanceCoordinator *)coordinator;
+- (void)updateAppearanceIfNeededWithAppearanceProvider:(id<RNSFormSheetAppearanceProvider>)appearanceProvider
+                                      behaviorProvider:(id<RNSFormSheetBehaviorProvider>)behaviorProvider
+                                            controller:(RNSFormSheetContentController *)controller
+                                           coordinator:(RNSFormSheetAppearanceCoordinator *)coordinator;
 
 - (void)resetInitialDetent;
 
--- a/ios/gamma/modals/form-sheet/RNSFormSheetAppearanceApplicator.h
+++ b/ios/gamma/modals/form-sheet/RNSFormSheetAppearanceApplicator.h
@@ -5,7 +5,6 @@
 
 @class RNSFormSheetAppearanceCoordinator;
 @class RNSFormSheetContentController;
-@class RNSFormSheetHostComponentView;
 
 NS_ASSUME_NONNULL_BEGIN
 
--- a/ios/gamma/modals/form-sheet/RNSFormSheetAppearanceApplicator.mm
+++ b/ios/gamma/modals/form-sheet/RNSFormSheetAppearanceApplicator.mm
@@ -24,10 +24,10 @@
   _initialDetentApplied = NO;
 }
 
-- (void)updateAppearanceIfNeededForAppearanceProvider:(id<RNSFormSheetAppearanceProvider>)appearanceProvider
-                                     behaviorProvider:(id<RNSFormSheetBehaviorProvider>)behaviorProvider
-                                           controller:(RNSFormSheetContentController *)controller
-                                          coordinator:(RNSFormSheetAppearanceCoordinator *)coordinator
+- (void)updateAppearanceIfNeededWithAppearanceProvider:(id<RNSFormSheetAppearanceProvider>)appearanceProvider
+                                      behaviorProvider:(id<RNSFormSheetBehaviorProvider>)behaviorProvider
+                                            controller:(RNSFormSheetContentController *)controller
+                                           coordinator:(RNSFormSheetAppearanceCoordinator *)coordinator
 {
   [coordinator updateIfNeeds:RNSFormSheetAppearanceUpdateFlagsConfiguration
            performOperations:^{
--- a/ios/gamma/modals/form-sheet/RNSFormSheetAppearanceApplicator.mm
+++ b/ios/gamma/modals/form-sheet/RNSFormSheetAppearanceApplicator.mm
@@ -6,7 +6,6 @@
 #import "RNSFormSheetAppearanceUpdateFlags.h"
 #import "RNSFormSheetContentController.h"
 #import "RNSFormSheetDetentResolver.h"
-#import "RNSFormSheetHostComponentView.h"
 
 @implementation RNSFormSheetAppearanceApplicator {
   BOOL _initialDetentApplied;
@@ -25,20 +24,24 @@
   _initialDetentApplied = NO;
 }
 
-- (void)updateAppearanceIfNeededForHost:(RNSFormSheetHostComponentView *)host
-                             controller:(RNSFormSheetContentController *)controller
-                            coordinator:(RNSFormSheetAppearanceCoordinator *)coordinator
+- (void)updateAppearanceIfNeededForAppearanceProvider:(id<RNSFormSheetAppearanceProvider>)appearanceProvider
+                                     behaviorProvider:(id<RNSFormSheetBehaviorProvider>)behaviorProvider
+                                           controller:(RNSFormSheetContentController *)controller
+                                          coordinator:(RNSFormSheetAppearanceCoordinator *)coordinator
 {
   [coordinator updateIfNeeds:RNSFormSheetAppearanceUpdateFlagsConfiguration
            performOperations:^{
-             [self updateSheetConfigurationForHost:host controller:controller];
+             [self updateSheetConfigurationForAppearanceProvider:appearanceProvider
+                                                behaviorProvider:behaviorProvider
+                                                      controller:controller];
            }];
 }
 
 #pragma mark - Updaters
 
-- (void)updateSheetConfigurationForHost:(RNSFormSheetHostComponentView *)host
-                             controller:(RNSFormSheetContentController *)controller
+- (void)updateSheetConfigurationForAppearanceProvider:(id<RNSFormSheetAppearanceProvider>)appearanceProvider
+                                     behaviorProvider:(id<RNSFormSheetBehaviorProvider>)behaviorProvider
+                                           controller:(RNSFormSheetContentController *)controller
 {
 #if !TARGET_OS_TV
   UISheetPresentationController *sheet = controller.sheetPresentationController;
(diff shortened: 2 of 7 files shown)


ios/gamma/modals/form-sheet/RNSFormSheetContentController.h
```diff
@@ -20,10 +21,20 @@ NS_ASSUME_NONNULL_BEGIN
...
+#pragma mark - Signals
+
+- (void)setNeedsPresentationUpdate;
+- (void)setNeedsAppearanceUpdate;
+- (void)setNeedsInitialDetentReset;
+
+#pragma mark - Mounting transaction signals
+
+- (void)reactMountingTransactionWillMount;
+- (void)reactMountingTransactionDidMount;
```
kkafar: It was my mistake to name these like this. 

The content controller should ideally not be aware of the react inner workings. 
We also likely don't need `will` callback. 

Single `flushPendingUpdates` should be enough. Just as it is done in tabs.

ios/gamma/modals/form-sheet/RNSFormSheetContentController.mm
```diff
@@ -88,6 +113,55 @@ - (void)dismissIfNeeded
...
 }
 
+#pragma mark - Appearance
+
+- (void)updateAppearanceIfNeededForHost:(nonnull RNSFormSheetHostComponentView *)host
+{
+  if (_needsInitialDetentReset) {
+    _needsInitialDetentReset = NO;
+    [_appearanceApplicator resetInitialDetent];
+  }
```
kkafar, line 143: I do my first reading & don't understand the code fully yet, but detent update in appearance method doesn't fit for me. Aren't detents rather a behaviour? 

Thinking now. Maybe call it configuration or something and then have both detents and appearance here.

ios/gamma/modals/form-sheet/RNSFormSheetContentController.mm
```diff
@@ -88,6 +113,55 @@ - (void)dismissIfNeeded
   [self dismissViewControllerAnimated:YES completion:nil];
 }
 
+#pragma mark - Appearance
+
+- (void)updateAppearanceIfNeededForHost:(nonnull RNSFormSheetHostComponentView *)host
```
kkafar: `...ForHost` suggests that you apply the appearance configuration **on** the host object, which is not the case. I'd name it `...withConfigProvider` or somethig like this.

This method ideally, shouldn't take `RNSFormSheetHostComponentView`, but rather an interface of a delegate / provider. This is not a priority though.

ios/gamma/modals/form-sheet/RNSFormSheetHostComponentView.mm
```diff
@@ -1,34 +1,35 @@
...
-  RNSFormSheetAppearanceCoordinator *_Nonnull _appearanceCoordinator;
-  RNSFormSheetAppearanceApplicator *_Nonnull _appearanceApplicator;
 
   RNSFormSheetContentController *_Nullable _controller;
   RCTSurfaceTouchHandler *_Nullable _touchHandler;
 
+  // Invalidation flags
+  BOOL _needsPresentationUpdate;
+  BOOL _needsAppearanceUpdate;
+  BOOL _needsInitialDetentReset;
```
kkafar: Should these live on the host? 

All the invalidation flags should leave on the content controller. Take a look at how `Tabs` are implemented. Host & Screens report the need (directly or not) of the update to the controller & it holds it, and clears the flag when it runs the update logic. 

We can talk about it some more if that's not fully clear.

ios/gamma/modals/form-sheet/RNSFormSheetHostComponentView.mm
```diff
@@ -201,24 +195,38 @@ - (void)updateProps:(const facebook::react::Props::Shared &)props
...
+    return;
+  }
 
-  [_appearanceCoordinator updateIfNeeds:RNSFormSheetAppearanceUpdateFlagsPresentation
-                      performOperations:^{
-                        [self updatePresentationState];
-                      }];
+  if (_needsPresentationUpdate) {
+    _needsPresentationUpdate = NO;
+    [_controller setNeedsPresentationUpdate];
```
kkafar: Instead of holding these additional flags on the host and passing this information to the controller here we can directly signalise to the controller when we e.g. detect prop change.

kkafar (commented):
I have few remarks. Please answer them before proceeding.

ios/gamma/modals/form-sheet/RNSFormSheetContentController.mm
t0maboro: you're right, there are a few things I don't like here:
- appearance & behavior separation
- that AppearanceApplicator is not stateless

my current plan is to:
- separate appearance and behavior on Providers level for this PR
- refactor AppearanceApplicator to extract Behavior configuration (next PR)
- make AppearanceApplcator stateless (next PR)

let me know if you're okay with that

ios/gamma/modals/form-sheet/RNSFormSheetHostComponentView.mm
```diff
@@ -65,32 +70,70 @@ - (void)resetProps
...
+#pragma mark - RNSFormSheetPresentationProvider
+
+- (BOOL)isOpen
 {
-  [super didMoveToWindow];
-  [self updatePresentationState];
+  return _isOpen;
+}
+
+- (UIWindow *)hostWindow
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4074/changes/9022290ae68410dbb8d249469f21d4089808fcf5

ios/gamma/modals/form-sheet/RNSFormSheetContentController.h
```diff
@@ -20,10 +22,21 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, weak, nullable) id<RNSFormSheetContentControllerDelegate> delegate;
 
+@property (nonatomic, weak, nullable) id<RNSFormSheetPresentationProvider> presentationProvider;
+@property (nonatomic, weak, nullable) id<RNSFormSheetAppearanceProvider> appearanceProvider;
+@property (nonatomic, weak, nullable) id<RNSFormSheetBehaviorProvider> behaviorProvider;
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4074/changes/e6c5928af1be7ad2e1f4922f0ea2f22d923f1d1d

ios/gamma/modals/form-sheet/RNSFormSheetAppearanceApplicator.mm
```diff
@@ -47,22 +50,23 @@ - (void)updateSheetConfigurationForHost:(RNSFormSheetHostComponentView *)host
       @"[RNScreens] sheetPresentationController is nil. Ensure modalPresentationStyle is set to UIModalPresentationFormSheet.");
 
   NSArray<UISheetPresentationControllerDetent *> *nativeDetents =
-      [RNSFormSheetDetentResolver buildSheetDetentsForFractions:host.detents];
+      [RNSFormSheetDetentResolver buildSheetDetentsForFractions:behaviorProvider.detents];
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4074/changes/e6c5928af1be7ad2e1f4922f0ea2f22d923f1d1d

ios/gamma/modals/form-sheet/RNSFormSheetAppearanceApplicator.h
```diff
@@ -10,9 +11,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface RNSFormSheetAppearanceApplicator : NSObject
 
-- (void)updateAppearanceIfNeededForHost:(RNSFormSheetHostComponentView *)host
-                             controller:(RNSFormSheetContentController *)controller
-                            coordinator:(RNSFormSheetAppearanceCoordinator *)coordinator;
+- (void)updateAppearanceIfNeededForAppearanceProvider:(id<RNSFormSheetAppearanceProvider>)appearanceProvider
+                                     behaviorProvider:(id<RNSFormSheetBehaviorProvider>)behaviorProvider
```
t0maboro: note: behavior will have a dedicated path, but I'm extracting it to a separate PR in which I'm going to introduce a dedicated class for that

t0maboro: associated followup PR: https://github.com/software-mansion/react-native-screens/pull/4084

ios/gamma/modals/form-sheet/RNSFormSheetAppearanceApplicator.h
```diff
@@ -1,18 +1,19 @@
...
-@class RNSFormSheetHostComponentView;
 
 NS_ASSUME_NONNULL_BEGIN
 
 @interface RNSFormSheetAppearanceApplicator : NSObject
 
-- (void)updateAppearanceIfNeededForHost:(RNSFormSheetHostComponentView *)host
-                             controller:(RNSFormSheetContentController *)controller
-                            coordinator:(RNSFormSheetAppearanceCoordinator *)coordinator;
+- (void)updateAppearanceIfNeededForAppearanceProvider:(id<RNSFormSheetAppearanceProvider>)appearanceProvider
```
kkafar: I have one remark regarding naming. 

```suggestion
- (void)updateAppearanceIfNeededWithAppearanceProvider:(id<RNSFormSheetAppearanceProvider>)appearanceProvider
```

The `For` part does sound inappropriate and misleading in this context.

ios/gamma/modals/form-sheet/RNSFormSheetHostComponentView.mm
```diff
@@ -65,32 +69,70 @@ - (void)resetProps
...
 {
-  [super didMoveToWindow];
-  [self updatePresentationState];
+  return _isOpen;
+}
+
+- (nullable UIWindow *)hostWindow
+{
+  return self.window;
+}
```
kkafar, line 97: I don't understand what `hostWindow` is for. The window is shared for both host & content hierarchies. Why do we need additional method here? Do you need window earlier than `onWillMoveToWindow` on `RNSFormSheetContentController.view`?

ios/gamma/modals/form-sheet/RNSFormSheetHostComponentView.mm
```diff
@@ -65,32 +69,70 @@ - (void)resetProps
...
+
+- (CGFloat)preferredCornerRadius
+{
+  return _preferredCornerRadius;
+}
+
+- (NSInteger)largestUndimmedDetentIndex
+{
+  return _largestUndimmedDetentIndex;
+}
```
kkafar: Are these necessary? You have declared properites on `RNSFormSheetHostComponentView.h`, aren't they synthesized by the compiler?

kkafar (commented):
Looks good. I still have few remarks / questions. Let's answer them.

ios/gamma/modals/form-sheet/RNSFormSheetHostComponentView.mm
t0maboro: When I navigate to the screen that has FormSheet with initially `isOpen=true` set, what I'm observing is that on the 1st mounting transaction, `flushPendingUpdates` updates the presentation and calls `updatePresentationState` before `willMoveToWindow` is called on ContentView

ios/gamma/modals/form-sheet/RNSFormSheetAppearanceApplicator.h
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4074/commits/c5473d9333235625ac8436e42c00c191d1b0bf66

ios/gamma/modals/form-sheet/RNSFormSheetHostComponentView.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4074/changes/dc3f9ac3a6fd02d193cdb927a0bbbe8bbaa04011 
note: having these providers, I think that we shouldn't expose properties from HostComponent at all, but rather operate on proper providers

ios/gamma/modals/form-sheet/RNSFormSheetHostComponentView.mm
kkafar: Okay, but why do you even need a window for before you receive `willMoveToWindow`?

ios/gamma/modals/form-sheet/RNSFormSheetHostComponentView.mm
kkafar: Sorry, I think I didn't communicate clearly.

These properties SHOULD exist on `RNSFormSheetHostComponentView` interface. We always expose the props. 

My remark is that you don't have to write their getters in `.mm` file. IIRC the getters should be synthesized by the compiler.

ios/gamma/modals/form-sheet/RNSFormSheetHostComponentView.mm
t0maboro: `updatePresentationState` triggers `presentFromWindow` and in that method I'm searching for a VC that will present this `ContentController`

ios/gamma/modals/form-sheet/RNSFormSheetHostComponentView.mm
t0maboro: correct, these getters will be synthesized, okay will expose props

t0maboro: https://github.com/software-mansion/react-native-screens/pull/4074/commits/a8a43b2a087f96ffbc3a22abf7b709d6be1d11b5 + https://github.com/software-mansion/react-native-screens/pull/4074/changes/6fd75c09f63a46c0fb425bd2253dd07596c80490

ios/gamma/modals/form-sheet/RNSFormSheetHostComponentView.mm
kkafar: Okay. I don't like the approach that `RNSFormSheetContentController` is responsible for its own presentation, however it's not crucial atm, it's more important to progress.

Let's create a ticket to clean this up. 

Host should be responsible for presentation of the content controller and the logic should leave there.

ios/gamma/modals/form-sheet/RNSFormSheetHostComponentView.mm
kkafar: Let's fix the code then. Restore the properties on the interface and remove unnecessary implementations.

ios/gamma/modals/form-sheet/RNSFormSheetHostComponentView.mm
t0maboro: https://github.com/software-mansion/react-native-screens-labs/issues/1488

kkafar (approved):

PR #4076, opened 2026-05-21
refactor(iOS, FormSheet v5): Expose reactEventEmitter from Host

t0maboro:
## Description

Similarly to Split, I'm exposing reactEventEmitter from Host, moving the logic to the ContentController.

## Changes

- expose reactEventEmitter from Host
- replace delegate callbacks with dispatching event from the Controller level

N/A - refactor

## Test plan

Focused on OnDetentChanged SFT, performed regression testing, randomly selecting other examples.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/gamma/modals/form-sheet/RNSFormSheetContentController.h
+++ b/ios/gamma/modals/form-sheet/RNSFormSheetContentController.h
@@ -8,13 +8,10 @@ NS_ASSUME_NONNULL_BEGIN
 @class RNSFormSheetContentController;
 @class RNSFormSheetHostComponentView;
 
+// TODO: @t0maboro - now we should get rid of that delegate and handle ShadowStateProxy/TouchHandler origin sync from
+// controller
 @protocol RNSFormSheetContentControllerDelegate <NSObject>
-- (void)sheetControllerDidNativeDismiss:(RNSFormSheetContentController *)controller;
 - (void)sheetControllerViewDidLayoutSubviews:(RNSFormSheetContentController *)controller;
-#if !TARGET_OS_TV
-- (void)sheetController:(RNSFormSheetContentController *)controller
-    didChangeDetentIdentifier:(nullable NSString *)identifier;
-#endif // !TARGET_OS_TV
 @end
 
 @interface RNSFormSheetContentController : UIViewController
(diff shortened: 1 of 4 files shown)


ios/gamma/modals/form-sheet/RNSFormSheetContentController.h
```diff
@@ -8,13 +8,10 @@ NS_ASSUME_NONNULL_BEGIN
 @class RNSFormSheetContentController;
 @class RNSFormSheetHostComponentView;
 
+// TODO: @t0maboro - now we should get rid of that delegate and handle ShadowStateProxy/TouchHandler origin sync from
+// controller
```
kkafar, line 12: No. Controller should not be aware of react-specific piping. Just as it is done in tabs: Controller should pass the necessary information via delegate interfaces to the delegate (host) and it should react appropriately.

ios/gamma/modals/form-sheet/RNSFormSheetContentController.mm
```diff
@@ -166,7 +168,10 @@ - (void)reactMountingTransactionDidMount
 
 - (void)presentationControllerDidDismiss:(UIPresentationController *)presentationController
 {
-  [self.delegate sheetControllerDidNativeDismiss:self];
+  RNSFormSheetHostComponentView *host = self.hostComponentView;
+  RCTAssert(host != nil, @"[RNScreens] hostComponentView must be set before delegate callbacks fire");
+  [host onNativeDismiss];
+  [[host reactEventEmitter] emitOnNativeDismiss];
```
kkafar, line 174: Definitely. Host should be a delegate. It should be notified that the `presentationControllerDidDismiss`, and from its delegate method implementation emit appropriate events to JS. 

This ensures that the `RNSFormSheetContentController` is not aware of react-specific piping and in the future can be tested in isolation. 

This PR tries to do exactly opposite, which is not the right direction.

kkafar (commented):
I've left some remarks, please answer them before proceeding.

This PR in the current shape is partially a regression. We should aim to decouple `RNSFormSheetContentController` from Host & other react specific piping, and here we do something opposite by directly emitting react element events from `RNSFormSheetContentController` code.

t0maboro:
as aligned internally, the implementation in Split & TabsScreen should be refactored, we shouldn't expose event emitter to native controller, closing this PR

PR #4080, opened 2026-05-21
chore(Example): remove default exports from `scenario-description.ts` files

sgaczol:
## Description

This PR removes default exports from `scenario-description.ts` files.

We are making a script mentioned in https://github.com/software-mansion/react-native-screens-labs/issues/1475 and such approach will give us a cleaner solution when it comes to extracting object from `module`.

## Changes

- make exports named instead of default ones
- update imports properly

N/A

## Test plan

N/A

## Checklist

- [ ] Ensured that CI passes


kkafar (commented):
Hey!

General remark: when creating PR description, explain not only "what" is changed, but also "why" behind the changes.

kkafar (approved):
I don't have other remarks. I've skimmed through the code (not every file) and it looks solid. 

Update the PR description and we can proceed.

PR #4085, opened 2026-05-22
chore(Tabs, iOS): add test-tabs-layout-appearances-ios screen and scenario

LKuchno:
## Description

Adds a new manual test screen — **Tab Bar Layout Appearances (iOS)** — to `apps/src/tests/single-feature-tests/tabs/`. The screen exercises `stackedLayoutAppearance`, `inlineLayoutAppearance`, and `compactInlineLayoutAppearance` for `TabsScreenAppearanceIOS`, using `tabBarItemTitleFontColor` as the visual discriminator between buckets (red = stacked, blue = inline, green = compactInline) and `tabBarItemTitleFontStyle: 'italic'` on the selected state as the font-style signal. The scenario documents explicit iOS 18 vs. iOS 26+ differences: the `inline` bucket is not honored on iOS 26+ iPhones, and normal-state (unselected) color falls back to the system default on iOS 26+. This PR is iOS-only; no Android changes are included.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1466

**Note:** `e2eCoverage` is set to `incomplete` because Detox cannot introspect native `UITabBarItem` typographic attributes (font style, font color) — all verification steps are manual.

## Changes

- **Test/Tabs**: Added `test-tabs-layout-appearances-ios/` scenario covering all three layout-appearance buckets (`stacked`, `inline`, `compactInline`), normal vs. selected state styling, and rotation-triggered bucket switching
- **Test/Tabs**: Added `scenario-description.ts` declaring the scenario as iOS-only with `e2eCoverage: 'incomplete'` and `smokeTest: false`
- **Test/Tabs**: Added `scenario.md` with full manual test steps, device/OS prerequisites, and iOS 18 vs. iOS 26+ caveats
- **Test/Tabs**: Registered `TestTabsLayoutAppearances` in `apps/src/tests/single-feature-tests/tabs/index.ts`

## Test visual presentation

|iOS18.6 iPhone Pro|iOS18.6 iPhone Pro Max|iOS26.5 iPhone Pro|
|---|---|---|
|<video src="https://github.com/user-attachments/assets/7e196586-0b39-48e3-bc59-e6e3d835afb4"/>|<video src="https://github.com/user-attachments/assets/a8fdd918-524c-4072-862f-4e7256e6af61"/>|<video src="https://github.com/user-attachments/assets/33e5a686-76dd-49ae-97b9-37b180493b6c"/>|


apps/src/tests/single-feature-tests/tabs/test-tabs-layout-appearances-ios/scenario-description.ts
```diff
@@ -0,0 +1,11 @@
+import type { ScenarioDescription } from "@apps/tests/shared/helpers";
+
+export const scenarioDescription: ScenarioDescription = {
+  name: 'Tab Bar Layout Appearances (iOS)',
+  key: 'test-tabs-layout-appearances-ios',
+  details:
+    'Verifies tab item font styles and colors across all layout appearances. Visual changes are used to confirm responsive layout shifts during device rotation.',
+  platforms: ['ios'],
+  e2eCoverage: 'incomplete',
+  smokeTest: false,
```
kkafar, line 10: These tests are fonts only - seems like. What about taking that into account in the test key? 

Also the meaning of `-layout-` part is not clear to me. I get that the applied properties differ between different width / height size classes of the interface.

What about `test-tabs-item-font-appearance-ios`? What do you think?

kkafar (approved):
Looks good overall. I have only naming related suggestions.

apps/src/tests/single-feature-tests/tabs/test-tabs-layout-appearances-ios/scenario-description.ts
LKuchno: I would keep the current naming, as what we are actually verifying here are the stacked, inline, and compact inline layouts. The fonts and colors differ only to serve as a point of reference, ensuring that the correct layout settings have been applied.

I used the term "layout" following a discussion with @t0maboro, who pointed me to the Apple documentation: https://developer.apple.com/documentation/uikit/uitabbarappearance/stackedlayoutappearance

As for the tab bar item title properties, those are already verified in the test-tabs-item-title test.

PR #4089, opened 2026-05-23
feat: support hintTextColor on iOS search bar

skrtdev:
## Summary
- apply `hintTextColor` to the iOS search bar placeholder
- keep the placeholder string stable when the hint color changes
- update search bar types/docs so `hintTextColor` is no longer marked Android-only

## Test plan
- `git diff --check`

`clang-format` and full package checks were not run locally because this environment does not have the repo toolchain installed.

### DIFF
--- a/apps/src/tests/issue-tests/Test4089.tsx
+++ b/apps/src/tests/issue-tests/Test4089.tsx
@@ -1,0 +1,198 @@
+import { NavigationContainer } from '@react-navigation/native';
+import { createNativeStackNavigator } from '@react-navigation/native-stack';
+import type { NativeStackScreenProps } from '@react-navigation/native-stack';
+import React, { useLayoutEffect, useState } from 'react';
+import { Button, ScrollView, StyleSheet, Text, View } from 'react-native';
+import type { SearchBarPlacement, SearchBarProps } from 'react-native-screens';
+
+type StackParamList = {
+  Home: undefined;
+  SearchBar: undefined;
+};
+
+type PlaceholderMode = 'default' | 'custom' | 'empty';
+type HintColorMode = 'default' | 'red' | 'blue';
+
+type SearchBarConfig = {
+  placement: SearchBarPlacement;
+  placeholderMode: PlaceholderMode;
+  hintColorMode: HintColorMode;
+  allowToolbarIntegration: boolean;
+};
+
+const Stack = createNativeStackNavigator<StackParamList>();
+
+const placements: SearchBarPlacement[] = [
+  'automatic',
+  'inline',
+  'stacked',
+  'integrated',
+  'integratedButton',
+  'integratedCentered',
+];
+
+const placeholderModes: PlaceholderMode[] = ['default', 'custom', 'empty'];
+const hintColorModes: HintColorMode[] = ['default', 'red', 'blue'];
+
+const defaultConfig: SearchBarConfig = {
+  placement: 'automatic',
+  placeholderMode: 'default',
+  hintColorMode: 'default',
+  allowToolbarIntegration: true,
+};
+
+function getNextValue<T>(values: readonly T[], currentValue: T): T {
+  const currentIndex = values.indexOf(currentValue);
+  return values[(currentIndex + 1) % values.length];
+}
+
+function getPlaceholder(mode: PlaceholderMode): SearchBarProps['placeholder'] {
+  switch (mode) {
+    case 'custom':
+      return 'Custom placeholder';
+    case 'empty':
+      return '';
+    default:
+      return undefined;
+  }
+}
+
+function getHintTextColor(
+  mode: HintColorMode,
+): SearchBarProps['hintTextColor'] {
+  switch (mode) {
+    case 'red':
+      return 'red';
+    case 'blue':
+      return 'blue';
+    default:
+      return undefined;
+  }
+}
+
+function Home({ navigation }: NativeStackScreenProps<StackParamList, 'Home'>) {
+  return (
+    <View style={styles.centeredContainer}>
+      <Text style={styles.title}>Test4089</Text>
+      <Button
+        title="Open search bar test"
+        onPress={() => navigation.navigate('SearchBar')}
+      />
+    </View>
+  );
+}
+
+function SearchBarScreen({
+  navigation,
+}: NativeStackScreenProps<StackParamList, 'SearchBar'>) {
+  const [config, setConfig] = useState<SearchBarConfig>(defaultConfig);
+
+  useLayoutEffect(() => {
+    navigation.setOptions({
+      headerSearchBarOptions: {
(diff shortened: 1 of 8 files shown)


src/fabric/SearchBarNativeComponent.ts
```diff
@@ -60,6 +60,7 @@ export interface NativeProps extends ViewProps {
   barTintColor?: ColorValue | undefined;
   tintColor?: ColorValue | undefined;
   textColor?: ColorValue | undefined;
+  hintTextColor?: ColorValue | undefined;
```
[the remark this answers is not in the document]
kligarski: I'd move it above the comment.

ios/RNSSearchBar.mm
```diff
@@ -204,7 +206,14 @@ - (void)setAutoCapitalize:(UITextAutocapitalizationType)autoCapitalize
...
 {
-  [_controller.searchBar setPlaceholder:placeholder];
+  _placeholder = placeholder;
+
+  if (_hintTextColor != nil && _placeholder != nil) {
+    _controller.searchBar.searchTextField.attributedPlaceholder =
+        [[NSAttributedString alloc] initWithString:_placeholder attributes:@{NSForegroundColorAttributeName : _hintTextColor}];
+  } else {
+    [_controller.searchBar setPlaceholder:_placeholder];
+  }
```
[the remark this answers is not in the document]
kligarski: Agreed

ios/RNSSearchBar.mm
```diff
@@ -227,6 +236,14 @@ - (void)setTextColor:(UIColor *)textColor
 #endif
 }
 
+- (void)setHintTextColor:(UIColor *)hintTextColor
+{
+#if !TARGET_OS_TV
```
kligarski: Why do we need tvOS guard here?

ios/RNSSearchBar.mm
```diff
@@ -227,6 +236,14 @@ - (void)setTextColor:(UIColor *)textColor
...
+  _hintTextColor = hintTextColor;
+  [self setPlaceholder:_placeholder];
```
kligarski: I wouldn't call the setter here. I suggest that we keep the setters simple (just set the `_placeholder`/`_hintTextColor` properties, if any of them we're changed (`oldScreenProps.XXX != newScreenProps.XXX`), set new local flag in `updateProps` to YES) and at the end of `updateProps` if props were invalidated, let's call new method like `updatePlaceholder` that will check both props and apply the `placeholder`/`attributedPlaceholder` accordingly.

src/types.tsx
```diff
@@ -1064,8 +1064,6 @@ export interface SearchBarProps {
   textColor?: ColorValue | undefined;
   /**
    * The search hint text color
-   *
-   * @plaform android
```
kligarski, line 1068: We should also remove `(Android only)` from `GUIDE_FOR_LIBRARY_AUTHORS.md`

ios/RNSSearchBar.mm
```diff
@@ -204,7 +206,14 @@ - (void)setAutoCapitalize:(UITextAutocapitalizationType)autoCapitalize
...
 - (void)setPlaceholder:(NSString *)placeholder
 {
-  [_controller.searchBar setPlaceholder:placeholder];
+  _placeholder = placeholder;
+
+  if (_hintTextColor != nil && _placeholder != nil) {
+    _controller.searchBar.searchTextField.attributedPlaceholder =
+        [[NSAttributedString alloc] initWithString:_placeholder attributes:@{NSForegroundColorAttributeName : _hintTextColor}];
+  } else {
+    [_controller.searchBar setPlaceholder:_placeholder];
```
kligarski: If the user provides only `_hintTextColor` and `_placeholder` remains `nil`, we get empty search bar instead of the default text. We need to fix this.

https://github.com/user-attachments/assets/6b7f5f9c-9ac6-4bb6-bb87-616eb2cecead

ios/RNSSearchBar.mm
```diff
@@ -204,7 +206,14 @@ - (void)setAutoCapitalize:(UITextAutocapitalizationType)autoCapitalize
 
 - (void)setPlaceholder:(NSString *)placeholder
 {
-  [_controller.searchBar setPlaceholder:placeholder];
+  _placeholder = placeholder;
+
+  if (_hintTextColor != nil && _placeholder != nil) {
+    _controller.searchBar.searchTextField.attributedPlaceholder =
```
kligarski, line 252: It seems like it doesn't work on iOS 26 for new search bar at the bottom. We should investigate why & if we can't fix it, describe the limitations in the prop docs.

https://github.com/user-attachments/assets/0f5d570f-1663-47cd-9f4e-50df9cf24979

kligarski (changes requested):
Hi, @skrtdev. Thank you for the PR. I left some questions and suggestions.

I also think it would be nice to create a dedicated test for the new prop. For current stack we're using `apps/src/tests/issue-tests` directory. We can create `Test4089.tsx` there. We should check different search bar placements and `placeholder` x `hintTextColor` configurations, including empty values and restoring default values in runtime.

Also, the title of the PR start with `feat:` as we're adding new functionality to iOS implementation rather than fixing something.

skrtdev:
You're right: it doesn't work on iOS 26 for new search bar at the bottom. I also noticed that even on non-bottom search bar, styling is ignored when the search bar is focused. It seems like we can't fix it. I also think there's a bug after the first focus. 2nd and subsequent focuses dont fully focus the search bar. The user can type but it's not the native focused state. I'll investigate more on that and eventually open a separate PR / issue.

apps/src/tests/issue-tests/Test4089.tsx
kligarski, line 1: The search bar doesn't really support runtime changes for e.g. placement so we should probably have configuration on root screen and then apply the config on new pushed screen with the config.

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -614,6 +614,7 @@ + (void)updateViewController:(UIViewController *)vc
             }
           }
 #endif /* Check for iOS 26.0 */
+          [searchBar updatePlaceholder];
```
kligarski, line 617: I'm not sure if we need this. We will be informed about any changes in `updateProps` right?

ios/RNSSearchBar.mm
```diff
@@ -508,10 +581,21 @@ - (UIView *)view
   }
 }
 
-RCT_EXPORT_VIEW_PROPERTY(placeholder, NSString)
+RCT_CUSTOM_VIEW_PROPERTY(placeholder, NSString, RNSSearchBar)
```
kligarski, line 584: We've dropped support for Paper so these are not necessary. I think recently PR removing those landed on main. We should merge/rebase onto current main.

PR #4097, opened 2026-05-25
chore: Pull latest react-navigation@7.x

t0maboro:
## Description

Bump react-navigation to the latest sources on 7.x. At this moment, any example with native bottom tabs from react navigation will crash due to the breaking changes we made.

## Changes

- pull latest react-navigation sources

(create any example with native bottom tabs)

### Before

Crash

### After

No crash

## Test plan

create any example with native bottom tabs

## Checklist

- [ ] Ensured that CI passes


kligarski (approved):
Looks good.

By the way, do you have the same error in your IDEs or is it a problem with my setup?

Image

I've seen it before this PR as well.

t0maboro:
> By the way, do you have the same error in your IDEs or is it a problem with my setup?

yup, I'm seeing the same issue in my IDE as well

kligarski:
> > By the way, do you have the same error in your IDEs or is it a problem with my setup?
> yup, I'm seeing the same issue in my IDE as well

LLM suggested some changes in tsconfig but it didn't really work (but I spent 5 minutes on it so it might need more research).

t0maboro:
> > > By the way, do you have the same error in your IDEs or is it a problem with my setup?
> > > yup, I'm seeing the same issue in my IDE as well
> 
> LLM suggested some changes in tsconfig but it didn't really work (but I spent 5 minutes on it so it might need more research).

I wouldn't spend much time on that, because in our examples we'll be using TabsContainer anyway

PR #4098, opened 2026-05-25
fix(Android, Tabs): Apply bottom inset immediately when the container is attached

t0maboro:
## Description

On Android, applying window insets asynchronously causes a visual jump of the bottom tab bar. This happens because React Native calculates the initial layout and starts the screen transition before the system dispatches the window insets. By the time the insets arrive, the `BottomNavigationView` recalculates its height/padding mid-transition or right after it, resulting in a tab bar content jitter.

> [!NOTE]  
> Visual jumps inside the TabBar's content will be resolved in a followup PR.

## Changes

- In TabsContainer we now fetch the window insets directly from the activity's `DecorView` during `onAttachedToWindow` and manually dispatch them immediately.
- Introduced the `tabBarShouldApplyInsetsSynchronously` prop to the JS layer, allowing developers to opt-out of this behavior if needed.

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/ef6ca723-1b56-4433-a3e1-0e3e3b866d9f" /> | <video src="https://github.com/user-attachments/assets/141faa5e-9513-48ec-9760-0ffa8492bd9d" /> |

## Test plan

Added dedicated SFT

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
@@ -10,10 +10,12 @@ import android.view.WindowInsets
 import android.widget.FrameLayout
 import androidx.appcompat.view.ContextThemeWrapper
 import androidx.core.graphics.Insets
+import androidx.core.view.ViewCompat
 import androidx.core.view.WindowInsetsCompat
 import androidx.core.view.children
 import androidx.core.view.size
 import androidx.fragment.app.FragmentManager
+import com.facebook.react.uimanager.ThemedReactContext
 import com.google.android.material.R
 import com.google.android.material.bottomnavigation.BottomNavigationView
 import com.swmansion.rnscreens.gamma.common.colorscheme.ColorScheme
@@ -126,6 +128,9 @@ class TabsContainer internal constructor(
 
     internal var colorScheme: ColorScheme by colorSchemeCoordinator::colorScheme
     internal var tabBarRespectsIMEInsets: Boolean = false
+
+    internal var tabBarShouldApplyInsetsSynchronously: Boolean = true
+    private var insetsAppliedBySystem = false
 
     private val contentView: FrameLayout =
         FrameLayout(context).apply {
@@ -274,6 +279,8 @@ class TabsContainer internal constructor(
         RNSLog.d(TAG, "TabsContainer [$id] attached to window")
 
         super.onAttachedToWindow()
+
+        maybeApplyDecorViewInsetsSynchronously()
         setupFragmentManager()
 
         // When TabsContainer is reattached to window, it might find new fragment manager (other
@@ -306,6 +313,19 @@ class TabsContainer internal constructor(
         colorSchemeCoordinator.onConfigurationChanged(newConfig)
     }
 
+    private fun maybeApplyDecorViewInsetsSynchronously() {
+        if (insetsAppliedBySystem || !tabBarShouldApplyInsetsSynchronously) return
+
+        val activity = (context as? ThemedReactContext)?.currentActivity ?: return
+        val decorView = activity.window.decorView
+        val insetsCompat = ViewCompat.getRootWindowInsets(decorView) ?: return
+
+        val windowInsets = insetsCompat.toWindowInsets()
+        if (windowInsets != null) {
+            dispatchApplyWindowInsets(windowInsets)
+        }
+    }
+
     override fun dispatchApplyWindowInsets(insets: WindowInsets?): WindowInsets? {
         // On Android versions prior to R, insets dispatch is broken.
         // In order to mitigate this, we override dispatchApplyWindowInsets with
@@ -315,6 +335,8 @@ class TabsContainer internal constructor(
         if (insets?.isConsumed ?: true) {
             return insets
         }
+
+        insetsAppliedBySystem = true
 
         for (child in children) {
             if (child === bottomNavigationView) {
--- a/apps/src/tests/single-feature-tests/tabs/index.ts
+++ b/apps/src/tests/single-feature-tests/tabs/index.ts
@@ -15,6 +15,7 @@ import TestTabsStaleStateUpdateRejection from './test-tabs-stale-update-rejectio
 import TestTabsTabBarMinimizeBehavior from './test-tabs-tab-bar-minimize-behavior-ios';
 import TestTabsTabBarControllerMode from './test-tabs-tab-bar-controller-mode-ios';
 import TestTabsSpecialEffectsScrollToTop from './test-tabs-special-effects-scroll-to-top';
+import TestTabsSynchronousInsetsAndroid from './test-tabs-synchronous-insets-android';
 import TestTabsTabBarExperimentalUserInterfaceStyle from './test-tabs-tab-bar-experimental-user-interface-style-ios';
 import TestTabsLifecycleEvents from './test-tabs-lifecycle-events';
 import TestTabsItemTitle from './test-tabs-item-title-ios';
@@ -35,6 +36,7 @@ const scenarios = {
   TestTabsTabBarMinimizeBehavior,
   TestTabsTabBarControllerMode,
   TestTabsSpecialEffectsScrollToTop,
+  TestTabsSynchronousInsetsAndroid,
   TestTabsTabBarExperimentalUserInterfaceStyle,
   TestTabsLifecycleEvents,
   TestTabsItemTitle,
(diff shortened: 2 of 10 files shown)


apps/src/tests/single-feature-tests/tabs/test-tabs-synchronous-insets-android/index.tsx
```diff
@@ -0,0 +1,153 @@
+import React, { createContext, useContext, useState, useEffect } from 'react';
+import { Button, StyleSheet, Text, View } from 'react-native';
+import { NavigationContainer } from '@react-navigation/native';
+import { createNativeStackNavigator } from '@react-navigation/native-stack';
+
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4098/commits/8d3fe3815b7b6c6665fd2c5c66d9b6de90ba679e

apps/src/tests/single-feature-tests/tabs/test-tabs-synchronous-insets-android/scenario.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Details
+
+**Description:** Verifies the `tabBarShouldApplyInsetsSynchronously` prop on the Android. When set to `true`, window insets are retrieved from the `DecorView` and applied synchronously during the `onAttachedToWindow` lifecycle phase. This prevents visible layout jumps of the bottom tab bar that occur when insets are applied asynchronously by the system during or after a screen transition.
+
+**OS test creation version:** Android: API Level 36.
+
+## E2E test
+
+No. Visual layout jumps during screen transitions are difficult to capture reliably in automated UI tests like Detox without precise frame-by-frame visual regression tools. Manual visual verification is required.
```
LKuchno, line 11: Following new naming ([described in RFC](https://github.com/software-mansion/react-native-screens-labs/pull/1493)), if we are not going to implement e2e test I would change this line to:

```suggestion
Incomplete: Not automated. Visual layout jumps during screen transitions are difficult to capture reliably in automated UI tests like Detox without precise frame-by-frame visual regression tools. Manual visual verification is required.
```

apps/src/tests/single-feature-tests/tabs/test-tabs-synchronous-insets-android/scenario-description.ts
```diff
@@ -0,0 +1,10 @@
+import type { ScenarioDescription } from '@apps/tests/shared/helpers';
+
+export const scenarioDescription: ScenarioDescription = {
+  name: 'Synchronous insets application',
+  key: 'test-tabs-synchronous-insets-android',
+  details: 'Test synchronous application of window insets on Android',
+  platforms: ['android'],
+  e2eCoverage: 'tbd',
```
LKuchno, line 8: Looking at the scenario _E2E test_ section it's already decided that this test won't be automated so e2eCoverage value update should be made:

```suggestion
  e2eCoverage: 'incomplete',
```

apps/src/tests/single-feature-tests/tabs/index.ts
```diff
@@ -15,6 +15,7 @@ import TestTabsStaleStateUpdateRejection from './test-tabs-stale-update-rejectio
 import TestTabsTabBarMinimizeBehavior from './test-tabs-tab-bar-minimize-behavior-ios';
 import TestTabsTabBarControllerMode from './test-tabs-tab-bar-controller-mode-ios';
 import TestTabsSpecialEffectsScrollToTop from './test-tabs-special-effects-scroll-to-top';
+import TestTabsSynchronousInsetsAndroid from './test-tabs-synchronous-insets-android';
```
LKuchno, line 18: If there is no corresponding iOS screen with the same name, we can omit the platform from the scenario name (similar to how it is done for TestTabsTabBarMinimizeBehavior). The fact that a scenario is Android-only is already indicated by the icon on the scenario list.

```suggestion
import TestTabsSynchronousInsets from './test-tabs-synchronous-insets-android';
```

apps/src/tests/single-feature-tests/tabs/index.ts
```diff
@@ -35,6 +36,7 @@ const scenarios = {
   TestTabsTabBarMinimizeBehavior,
   TestTabsTabBarControllerMode,
   TestTabsSpecialEffectsScrollToTop,
+  TestTabsSynchronousInsetsAndroid,
```
LKuchno, line 39:
```suggestion
  TestTabsSynchronousInsets,
```

apps/src/tests/single-feature-tests/tabs/test-tabs-synchronous-insets-android/scenario.md
```diff
@@ -0,0 +1,42 @@
...
+
+1. Launch the app and navigate to **Synchronous insets application**.
+
+- [ ] Expected: The **Setup** screen is visible. The `shouldApplyInsetsSynchronously` button is toggled `true`.
+
+2. Tap the **Push Tabs Screen** button.
+
+- [ ] Expected: A transition to the tabs screen begins. The bottom tab bar renders with the correct height and bottom padding respecting system navigation bars. There is no visible vertical layout jump or resize of the tab bar after the transition completes. 
+
+3. Press the back button to return to the **Setup** screen.
```
LKuchno, line 30: Could you add Button to the screen to go back to previous screen? 
On new Android devices (i.e. Pixel 9) system back button is hidden by default. 

Image

LKuchno (changes requested):
I left few comments and suggestions :)

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
```diff
@@ -316,6 +336,8 @@ class TabsContainer internal constructor(
             return insets
         }
 
+        insetsAppliedBySystem = true
```
kligarski, line 339: First call to this method comes from us in `onAttachedToWindow` -> can we call it "by system" then?

src/components/tabs/host/TabsHost.android.types.ts
```diff
@@ -15,4 +15,24 @@ export interface TabsHostPropsAndroid {
    * @supported API 30 or higher
    */
   tabBarRespectsIMEInsets?: boolean | undefined;
+
```
kligarski, line 18:
```suggestion
```

kligarski: I think we've been using no new line between props. Recently in FormSheet we started using them but it would be nice to keep one convention. Not sure which one.

src/components/tabs/host/TabsHost.android.types.ts
```diff
@@ -15,4 +15,24 @@ export interface TabsHostPropsAndroid {
...
+   *
+   * Setting this to `false` disables this workaround and falls back to the default asynchronous
+   * behavior. This serves as an opt-out mechanism if the synchronous application causes
+   * conflicts with other inset-handling configurations or unexpected UI side-effects.
+   *
+   * @default true
+   *
+   * @platform android
+   */
+  tabBarShouldApplyInsetsSynchronously?: boolean;
```
kligarski, line 37: I'm thinking about how we want to expose controlling the inset in tabs here.
Now we're adding a switch between reading from decor vs not reading from decor and relying on native dispatch.

But should we also expose an option to disable the inset completely? E.g. is somebody uses a footer of some kind that handles the bottom inset visually but doesn't consume it natively (e.g. a react native view)?

If so, do we need an option to control synchronous read? I guess it won't hurt even if we add a prop to control inset application later.

t0maboro:
Investigation for delayed draw of the pill and suspicious animations during transition

I noticed a significant difference in draw passes between this PR and #4161 

The RC is a native Android animation (`ChangeBounds`) interacting with how the layout measurement callbacks are scheduled on the native thread.

The `BottomNavigationView` has a built-in `TransitionManager` that attempts to trigger a `ChangeBounds` animation whenever the menu item's selection is updated. How the screen's layout callback is posted determines whether this animation plays or is completely bypassed.

**1. Animation Triggerring Path**
* The layout callback is posted via `ReactChoreographer`.

1. The initial empty state is measured (width: 0).
2. The menu is built, the item selection is realized, and the `TransitionManager` zeroed layout as `startValues`.
3. The layout callback runs and updates the layout to full width.
4. The `onPreDraw` listener fires and captures the new full-width bounds as `endValues`.
5. Because `startValues` != `endValues`, `ChangeBounds.createAnimator` is triggered (which internally calls `suppressLayout` on the `BottomNavigationView`). An `ObjectAnimator` is created, interpolating the width frame-by-frame with `dispatchDraw`.

**2. Animation Bypassing Path**
* The layout update is posted via `Handler.post`.

1. The initial empty state is measured (width: 0). `TransitionManager` captures this as `startValues`.
2. The menu is built, the item selection is realized, and the `TransitionManager` zeroed layout as `startValues`.
3. Because of the `Handler.post`, the layout has *not* changed yet when the frame is about to draw.
4. The `onPreDraw` listener fires and captures the `endValues`. Since the layout hasn't updated, the `endValues` still show a width of 0.
5. Because `startValues` == `endValues`, no animator is created. 
6. In the next `doFrame` tick (during the `CALLBACK_TRAVERSALS` path), the delayed layout pass finally executes and draws the proper, full-width layout immediately in exactly 1 draw pass.

t0maboro:
superseded by: https://github.com/software-mansion/react-native-screens/pull/4161

PR #4101, opened 2026-05-26
feat(Android, Stack v5): expose showAsAction prop for toolbar menu items

kligarski:
## Description

Adds `showAsAction` property to toolbar menu items in header on Android in Stack v5. 

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1202.

### API shape

The initial API idea was:

```ts
showAsAction?: 'always' | 'ifRoom' | 'never' | undefined;
showAsActionWithText?: boolean | undefined;
```

But this turned out to be problematic on the native side. You can only set the
`showAsAction` flags, you can't read them from the `MenuItem`. If the user
changed only one of these 2 props, we wouldn't know all the flags that need to
be applied. We could add state to the `ToolbarMenuCoordinator` or `HeaderConfig`
but we would like to avoid it.

I decided to merge these 2 props into one enum. For `never`, we don't need the
`withText` modifier so there are only 5 options possible which isn't bad.

We could also use nested object but we've been trying to avoid complicated
nesting.

The final API therefore looks like this:

```ts
showAsAction?: 'always' | 'alwaysWithText' | 'ifRoom' | 'ifRoomWithText' | 'never' | undefined;
```

I'm open to suggestions though.

### Codegen problem

When adding the
`showAsAction?: CT.WithDefault<StackHeaderToolbarMenuItemShowAsActionAndroid, 'never'>;`
prop to `StackHeaderToolbarMenuItemAndroid` interface in native spec on Android,
iOS build stopped working. In `Props.h`, all types are generated (also for
Android-only components). For some reason, if we use nested object as an array
(` toolbarMenuItems?: StackHeaderToolbarMenuItemAndroid[] | undefined;`), the
enum type isn't generated which causes the build to fail. After I added
`DO_NOT_USE?: StackHeaderToolbarMenuItemAndroid | undefined;` property, the enum
is generated correctly. I assumed that it's better to add this workaround
instead of losing the typing completely.

### Limited support for `ifRoom`

`ifRoom` option doesn't react to orientation changes. The maximum width of
action items is determined on first layout and it remains that way even through
layout/orientation changes. This means that if you open the screen in portrait,
more items won't move from overflow menu to the toolbar when rotated to
landscape and if you open the screen in landscape, rotating to portrait won't
make the items move to the overflow menu, usually resulting in the title being
shrunk down. 

| portrait -> landscape | landscape -> portrait |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/ef59d7f6-a171-46b9-9854-ee977989a951" /> | <video src="https://github.com/user-attachments/assets/ecc66cc1-eca0-44ef-86e1-7ad9597bf54d" /> |

This happens due to native implementation. Typically on Android, every
orientation change causes the activity to restart. The maximum width of action
menu items is evaluated only on initialization of `ActionMenuProesenter`
(`initForMenu`). If activity restarts are disabled, as is the case in
`react-native` apps, the initially set width will still apply.

> [!NOTE]
>
> By the way, this value turns out to be half of the width of the device:
> 
> ```java
> public int getEmbeddedMenuWidthLimit() {
>     return mContext.getResources().getDisplayMetrics().widthPixels / 2;
> }
> ```
> 
> This value is then made smaller by subtracting the width of the collapse icon.

I tried looking for a way to force this value to reset. Unfortunately, I didn't
find any way that wouldn't involve heavy reflection usage or completely
resetting the menu, resulting in the loss of state (i.e. changes applied via
view commands would be completely lost). I think that keeping the state is more
important in this case but I'm open for discussion here.

While testing this I also noticed some problems with handling orientation
changes for the screen. I will investigate this separately
(https://github.com/software-mansion/react-native-screens-labs/issues/1499).

https://github.com/user-attachments/assets/a6d64b67-6599-4abc-a154-92bdacfe9c72

## Changes

- add `showAsAction` prop and `StackHeaderToolbarMenuItemShowAsAction(Android)` enums
- add `DO_NOT_USE?: StackHeaderToolbarMenuItemAndroid | undefined;` to native spec to  ensure generation of enum in `Props.h` to fix build on iOS
- minor fixes:
  - expose `StackHeaderToolbarMenuItemClickedEvent`
  - make `clear`, `updateItem` internal in `ToolbarMenuCoordinator`
  - adjust `test-stack-toolbar-menu-commands` to new convention

## Visual documentation

https://github.com/user-attachments/assets/419ef60c-f7af-4a43-8f86-b62640a681bc

## Test plan

Run `test-stack-toolbar-menu-show-as-action-android` SFT.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/toolbar/StackHeaderToolbarMenuItemOptions.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/toolbar/StackHeaderToolbarMenuItemOptions.kt
@@ -9,4 +9,6 @@ package com.swmansion.rnscreens.gamma.stack.header.toolbar
 data class StackHeaderToolbarMenuItemOptions(
     val title: String? = null,
     val hidden: Boolean? = null,
+    val showAsAction: StackHeaderToolbarMenuItemShowAsAction? = null,
+    val showAsActionWithText: Boolean? = null,
 )
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/toolbar/StackHeaderToolbarMenuItemOptions.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/toolbar/StackHeaderToolbarMenuItemOptions.kt
@@ -10,5 +10,4 @@ data class StackHeaderToolbarMenuItemOptions(
     val title: String? = null,
     val hidden: Boolean? = null,
     val showAsAction: StackHeaderToolbarMenuItemShowAsAction? = null,
-    val showAsActionWithText: Boolean? = null,
 )
--- a/src/fabric/gamma/stack/StackHeaderConfigAndroidNativeComponent.ts
+++ b/src/fabric/gamma/stack/StackHeaderConfigAndroidNativeComponent.ts
@@ -17,7 +17,9 @@ type StackHeaderToolbarMenuItemClickedEvent = Readonly<{
 
 type StackHeaderToolbarMenuItemShowAsActionAndroid =
   | 'always'
+  | 'alwaysWithText'
   | 'ifRoom'
+  | 'ifRoomWithText'
   | 'never';
 
 export interface StackHeaderToolbarMenuItemAndroid {
@@ -28,7 +30,6 @@ export interface StackHeaderToolbarMenuItemAndroid {
     StackHeaderToolbarMenuItemShowAsActionAndroid,
     'never'
   >;
-  showAsActionWithText?: CT.WithDefault<boolean, false>;
 }
 
 export interface NativeProps extends ViewProps {
--- a/src/fabric/gamma/stack/StackHeaderConfigAndroidNativeComponent.ts
+++ b/src/fabric/gamma/stack/StackHeaderConfigAndroidNativeComponent.ts
@@ -54,6 +54,13 @@ export interface NativeProps extends ViewProps {
   onToolbarMenuItemClicked?:
     | CT.DirectEventHandler<StackHeaderToolbarMenuItemClickedEvent>
     | undefined;
+
+  // When StackHeaderToolbarMenuItemAndroid is used as an array
+  // in toolbarMenuItems, Codegen doesn't generate an enum in Props.h
+  // which causes a build failure on iOS. By adding a property where
+  // StackHeaderToolbarMenuItemAndroid is used directly, we ensure
+  // that the enum is generated.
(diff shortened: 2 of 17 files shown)


android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/toolbar/StackHeaderToolbarMenuItemOptions.kt
```diff
@@ -9,4 +9,6 @@ package com.swmansion.rnscreens.gamma.stack.header.toolbar
 data class StackHeaderToolbarMenuItemOptions(
     val title: String? = null,
     val hidden: Boolean? = null,
+    val showAsAction: StackHeaderToolbarMenuItemShowAsAction? = null,
+    val showAsActionWithText: Boolean? = null,
```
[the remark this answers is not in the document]
kligarski: https://github.com/software-mansion/react-native-screens/pull/4101/commits/3e01af05ba02ae08ae774b56e05181db68433828

src/fabric/gamma/stack/StackHeaderConfigAndroidNativeComponent.ts
```diff
@@ -44,6 +55,13 @@ export interface NativeProps extends ViewProps {
   onToolbarMenuItemClicked?:
     | CT.DirectEventHandler<StackHeaderToolbarMenuItemClickedEvent>
     | undefined;
+
+  // When StackHeaderToolbarMenuItemAndroid is used as an array
+  // in toolbarMenuItems, Codegen doesn't generate an enum in Props.h
+  // which causes a build failure on iOS. By adding a property where
+  // StackHeaderToolbarMenuItemAndroid is used directly, we ensure
+  // that the enum is generated.
+  DO_NOT_USE?: StackHeaderToolbarMenuItemAndroid | undefined;
```
t0maboro, line 64: I don't like 2 facts:
- that it forces us to create empty stubs in ViewManager on android & it pollutes native props where that attribute will be publicly accessible
- that the content isn't hidden behind some `ifdef` in `Props.h` as we explicitly excluded iOS from platforms

Anyway, this isn't our issue, so we may proceed, but I'm wondering whether we shouldn't report the issue to core or bump if it exists.

t0maboro (approved):

src/fabric/gamma/stack/StackHeaderConfigAndroidNativeComponent.ts
kligarski: https://github.com/software-mansion/react-native-screens-labs/issues/1510

PR #4083, opened 2026-05-22
chore(tabs): add tab-item-title appearance scenario for iOS

LKuchno:
## Description

Adds a new manual test scenario (`test-tabs-item-title-ios`) that exercises 8 iOS-only tab bar item title appearance props exposed on `TabsScreen`: `title` (string label), `tabBarItemTitleFontColor`, `tabBarItemTitleFontFamily`, `tabBarItemTitleFontSize`, `tabBarItemTitleFontStyle`, `tabBarItemTitleFontWeight`, and `tabBarItemTitlePositionAdjustment`. The scenario also validates that `tabBarItemTitleFontColor` correctly overrides the host-level `tabBarTintColor` for the label while leaving the icon tint unaffected. 

This is iOS-only; Detox automated assertions are not feasible because Detox does not expose color or font attributes of native tab bar items. Two known issues are documented in the scenario notes: the `normal`-state color is not applied on iOS 26, and long-title truncation does not work correctly on iOS 18 and lower.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1396 

## Changes

- **Tests/Tabs**: Added `test-tabs-item-title-ios/index.tsx` — new test screen with three tab routes exercising long-title truncation, `tabBarItemTitleFontColor` vs `tabBarTintColor` override, and combined font/position props
- **Tests/Tabs**: Added `scenario-description.ts` registering the scenario as iOS-only, e2eCoverage `incomplete`, `smokeTest: false`
- **Tests/Tabs**: Added `scenario.md` with full manual test steps, known-issue callouts for iOS 26 normal-state color and iOS 18 truncation bug, and checklist items for each testable behavior
- **Tests/Tabs**: Registered `TestTabsItemTitle` in `apps/src/tests/single-feature-tests/tabs/index.ts` so it appears in the scenario list

## Test visual presentation

https://github.com/user-attachments/assets/327fd47c-9cda-46e0-aca0-76c31741e87c

PR #4106, opened 2026-05-28
chore(test): test-tabs-item-title appearance scenario

LKuchno:
## Description

- Replaced the iOS-only `test-tabs-item-title-ios` screen and scenario (added in https://github.com/software-mansion/react-native-screens/pull/4083) with a unified `test-tabs-item-title` that covers both iOS and Android tab bar item title props in a single file.
- Screen components branch on `Platform.OS` to render platform-appropriate explanatory text; each `TabRouteConfig` now carries both `ios` and `android` blocks covering font color (selected/normal/focused states), font family, font size (iOS: single size with position adjustment; Android: small/large label sizes), font style, and font weight.
- Updated the registration in `apps/src/tests/single-feature-tests/tabs/index.ts` to point to the new unified directory, removing the platform-suffixed import.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1403

### Changes:

- iOS updates: Minor changes only—defined normal.tabBarItemTitleFontColor for the Color tab and renamed the Font and Position tab to Font.
- Android updates: Created the screen implementation with Android-specific descriptive text (the first tab remains identical across both operating systems).
- Scenario updates: Updated the test scenario to cover both platforms and specified the expected results for each, as certain behaviors differ between iOS and Android.
- Test directory name update to not point iOS
- Scenario-description updated with android as covered platform, iOS removed from key and extending description.

### Test screen record:

https://github.com/user-attachments/assets/6543d03d-883f-427d-bd4f-912efa7daf9c


apps/src/tests/single-feature-tests/tabs/test-tabs-item-title/scenario.md
```diff
@@ -0,0 +1,109 @@
+# Test Scenario: Tab Bar Item Title
+
+## Details
+
+Validates the tab bar item title configuration across both platforms, verifying
```
t0maboro: `**Description:**` prefix should be added or are we resigning from it?

t0maboro (approved):

apps/src/tests/single-feature-tests/tabs/test-tabs-item-title/scenario.md
LKuchno: We are not I removed it by mistake - good catch thank you!

https://github.com/software-mansion/react-native-screens/pull/4106/changes/23122cec60aaa8fd57fcb6f80eb7bcdb333c94c1

PR #4111, opened 2026-05-29
fix(iOS, Tabs): weak-capture tab bar item and screen view in async image callbacks

cortinico:
## Description

Avoid updating a tab bar item whose internal view hierarchy has been
torn down. This works around an iOS 26 UIKit regression where unowned
refs inside UITabBarItem._updateViewAndPositionItems: cause crashes.

Both the normal icon and selected icon completion blocks now:
1. Weak-capture tabBarItem and screenView before the block
2. Promote to strong refs inside the block and bail if either is nil
3. Guard that the screen's controller is still parented by a UITabBarController

TBA

## Test plan

Discussed offline

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/tabs/RNSTabBarAppearanceCoordinator.mm
+++ b/ios/tabs/RNSTabBarAppearanceCoordinator.mm
@@ -92,16 +92,30 @@
   } else if (imageLoader != nil) {
     bool isTemplate = screenView.iconType == RNSTabsIconTypeTemplate;
 
+    // Weak-capture to avoid updating a tab bar item whose internal
+    // view hierarchy has been torn down (iOS 26 UIKit regression:
+    // unowned refs inside UITabBarItem._updateViewAndPositionItems:).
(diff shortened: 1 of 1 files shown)


kkafar (commented):
Hello!

I can not really reproduce the issue, but it won't hurt to add a null-guard & verify whether we have been not detached from the parent controller since block-schedule-time. 

One thing though -> this slightly changes semantics. Earlier, even if we were detached from the parent controller, the update would go through, effectively setting `tabBarItem.image` / `tabBarItem.selectedImage`. Now it won't. This implies that we might need to take care of reapplying the appearance at the moment we receive `didMoveToParentViewController` on `RNSTabsScreenViewController`. 

I'll look into it.

kkafar (approved):
Okay, we already request appearance update on `didMoveToParentViewController`.

Seems fine then.

PR #4112, opened 2026-05-29
chore(readme): add readme banners

p-malecki:
## Description

Adds banners to the readme. They are currently hidden.


kkafar (approved):
This will allow for dynamic banner serving by simply swapping the contents served by the server.

PR #4116, opened 2026-05-31
fix(Android): fix translucent header first layout correction

### LINKED ISSUE #3530: [Android] headerTransparent: true causes layout glitch (bottom gap) during transition (newArch)
### Description

When using @react-navigation/native-stack on Android with `headerTransparent: true`, a visible gap (appearing to be the height of the header) flashes at the bottom of the screen during the transition. It doesn't consistently happen but it's likely to happen on screens that are expensive to render. This only happens on Android with the New Architecture enabled. 

As seen in the screen recording (slowed down to 0.25x) you can still see the old screen for a short period in the bottom of the screen.

https://github.com/user-attachments/assets/f3929b20-8b00-41e8-84e0-b250677f7b1e

### Steps to reproduce

1. Add a screen with `headerTransparent: true`
2. Navigate to screen (more likely to happen if the screen is a bit more heavy to render
3. Sometimes the glitch appears, sometimes it doesn't)

https://github.com/LeonDvlpmnt/screens-reproducer

### Screens version

4.19.0

### React Native version

0.81.5

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

Expo bare workflow

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

Real device

### Device model

Pixel 9

### Acknowledgements

Yes

sorinc03:
## Summary
- Skip the Android Fabric first-layout header-height correction when the header config is translucent.
- Keep translucent header layout consistent with the native/header-config behavior so the screen is not temporarily shortened during transitions.

## Root cause
On the first Fabric layout, before Android has sent the native screen dimensions back through state, `RNSScreenShadowNode` estimates the native header height and applies padding/frame corrections. Translucent headers do not reserve that vertical space on the native side, so applying the estimate can briefly shrink the screen and reveal the previous screen at the bottom during the transition.

## Validation
- `yarn check-types`
- `yarn lint-android`
- `cd FabricExample/android && ./gradlew :app:assembleDebug --console=plain -PreactNativeArchitectures=arm64-v8a`

Fixes #3530

### DIFF
--- a/common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.cpp
+++ b/common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.cpp
@@ -101,7 +101,10 @@ void RNSScreenShadowNode::appendChild(
           *std::static_pointer_cast<const RNSScreenStackHeaderConfigProps>(
               headerConfigChild->getProps());
 
-      const auto headerHeight = headerProps.hidden
+      // Translucent headers do not reserve vertical space, so applying the
+      // estimated header correction can expose the previous screen at the
+      // bottom during the first Fabric layout.
+      const auto headerHeight = (headerProps.hidden || headerProps.translucent)
           ? 0.f
           : findHeaderHeight(
                 headerProps.titleFontSize, headerProps.title.empty())


kmichalikk (approved):
From what I can see, this fixes the issue by making the bottom padding 0 at first render. Then it also sets header height to 0 and for the case of translucent header is a bit misleading, because it has its height, just it's not considered for screen layout. Still, I wouldn't do any unnecessary refactors to v4, so It's okay for me.

cc @kkafar what do You think?

kkafar (approved):
Looks sensible, thanks!

PR #4117, opened 2026-05-31
Fix iOS orientation updates during transitions

### LINKED ISSUE #3831: enforceDesiredDeviceOrientation silently fails on iOS 16+ during screen transitions
### Description

On iOS 16+, `enforceDesiredDeviceOrientation` in `RNSScreenWindowTraits.mm` has two bugs that cause per-screen orientation changes to silently fail during navigation transitions:

1. **Stale `windowScene.interfaceOrientation`** — The heuristic checks `interfaceOrientation` to decide whether rotation is needed. This value can be stale after scene restoration, app relaunch, or a previous `requestGeometryUpdate` that changed internal metadata without visually rotating. When stale, the method thinks the device is already in the correct orientation and skips the rotation entirely. (Same root cause as #2197)

2. **Empty error handler on `requestGeometryUpdate`** — When called during `viewWillAppear`, UIKit's internal orientation cache may still hold the *outgoing* screen's mask. `requestGeometryUpdate` validates against this stale cache and fails with:
   ```
   Error Domain=UISceneErrorDomain Code=101
   "None of the requested orientations are supported by the view controller.
    Requested: portrait, portraitUpsideDown; Supported: landscapeRight"
   ```
   The error handler is `^(NSError *_Nonnull error){}` — the failure is silently swallowed.

These bugs make it unreliable to navigate between screens with different orientations (e.g. portrait home → landscape play screen → back to portrait). After 1-2 cycles the orientation gets permanently stuck.

Related: [expo/expo#43692](https://github.com/expo/expo/issues/43692), #2197, #1775

### Steps to reproduce

1. Clone the repro: `git clone https://github.com/JeanDes-Code/orientation-lock-repro`
2. `npm install`
3. **Delete `patches/react-native-screens+4.23.0.patch`** (the expo-screen-orientation patch can stay — it's equivalent to [expo/expo#44181](https://github.com/expo/expo/pull/44181))
4. `npm install` again (to revert the react-native-screens patch)
5. `npx expo prebuild --clean && npx expo run:ios`
6. Tap **"Go to Landscape (Option C)"** — uses `Stack.Screen options={{ orientation: 'landscape_right' }}`
7. Tap **"Go Back"** — screen should rotate to portrait but stays stuck in landscape
8. Repeat — orientation locking fails in both directions

https://github.com/JeanDes-Code/orientation-lock-repro

The repo includes a [workaround patch for react-native-screens](https://github.com/JeanDes-Code/orientation-lock-repro/blob/master/patches/react-native-screens%2B4.23.0.patch) that:
- Always calls `setNeedsUpdateOfSupportedInterfaceOrientations` + `requestGeometryUpdate` on iOS 16+ (bypasses the stale `interfaceOrientation` heuristic)
- Falls back to `UIDevice setValue:@(orientation) forKey:@"orientation"` in the error handler
- Moves the old heuristic to the pre-iOS 16 path only

### Screens version

4.23.0

### React Native version

0.83.4

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

iOS simulator, Real device

### Device model

iPhone 17 Pro Simulator (iOS 26.2), iPhone 15 Pro (iOS 18)

### Acknowledgements

Yes

sorinc03:
## Summary
- On iOS 16+, request orientation geometry updates from the desired supported-orientation mask directly instead of first relying on `windowScene.interfaceOrientation`.
- Refresh the top view controller's supported-orientation state before requesting geometry updates.
- Retry a failed geometry update once on the next main-queue turn and warn if UIKit still rejects it, instead of silently swallowing the failure.

## Root cause
`enforceDesiredDeviceOrientation` gated iOS 16+ `requestGeometryUpdate` behind a pre-iOS-16 heuristic that compares the current device and scene interface orientations. During navigation transitions, scene restoration, or after previous geometry updates, `windowScene.interfaceOrientation` can be stale. When the stale orientation still appears to satisfy the new mask, the method skips the iOS 16+ geometry request completely.

UIKit may also reject a transition-time request while its supported-orientation cache still reflects the outgoing screen. The previous error handler was empty, so these failed requests disappeared without another attempt.

## Reproduction
This is easiest to reproduce with the Expo repro from the issue because it exercises native-stack per-screen orientation changes during a real iOS transition.

1. `git clone https://github.com/JeanDes-Code/orientation-lock-repro`
2. `cd orientation-lock-repro`
3. `npm install`
4. Delete `patches/react-native-screens+4.23.0.patch` so the app runs against unpatched react-native-screens. The `expo-screen-orientation` patch can remain.
5. `npm install`
6. `npx expo prebuild --clean`
7. `npx expo run:ios`
8. Tap **Go to Landscape (Option C)**. This navigates to a native-stack screen with `options={{ orientation: 'landscape_right' }}`.
9. Tap **Go Back**. The app should return to portrait, but without this fix it can stay stuck in landscape.
10. Repeat the landscape/back cycle once or twice. The orientation lock can then fail in both directions.

The app-level repro is equivalent to this single-file native-stack setup:

```tsx
import React from 'react';

function PortraitScreen({ navigation }) {
  return (
      <Button
        title="Go to Landscape"
        onPress={() => navigation.push('Landscape')}
      />
    </View>
  );
}

function LandscapeScreen({ navigation }) {
  return (
      <Button title="Go Back" onPress={() => navigation.goBack()} />
    </View>
  );
}

  return (
    <NavigationContainer>
      <Stack.Navigator screenOptions={{ orientation: 'portrait' }}>
        <Stack.Screen name="Portrait" component={PortraitScreen} />
        <Stack.Screen
          name="Landscape"
          component={LandscapeScreen}
          options={{ orientation: 'landscape_right' }}
        />
      </Stack.Navigator>
    </NavigationContainer>
  );
}
```

## Validation
- `yarn check-types`
- `yarn lint-swift`
- `xcodebuild -workspace FabricExample/ios/FabricExample.xcworkspace -scheme FabricExample -configuration Debug -destination "platform=iOS Simulator,name=iPhone 17 Pro Max,OS=latest" -derivedDataPath /Users/sorincioban/Library/Developer/Xcode/DerivedData/RNScreensOSS CODE_SIGNING_ALLOWED=NO build`

Fixes #3831

### DIFF
--- a/ios/RNSScreenWindowTraits.mm
+++ b/ios/RNSScreenWindowTraits.mm
@@ -97,11 +97,83 @@
 }
 #endif
 
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(16_0) && !TARGET_OS_TV && !TARGET_OS_VISION
++ (UIWindowScene *)windowSceneForOrientationUpdate
+{
+  UIWindow *keyWindow = RCTKeyWindow();
+  if (keyWindow.windowScene != nil) {
+    return keyWindow.windowScene;
+  }
+
+  // When an app supports multiple scenes (e.g. CarPlay), it is possible that
+  // UIWindowScene is not the first scene, or it may not be present at all.
+  NSArray *connectedScenes = [[[UIApplication sharedApplication] connectedScenes] allObjects];
+  for (id connectedScene in connectedScenes) {
+    if ([connectedScene isKindOfClass:[UIWindowScene class]]) {
+      return connectedScene;
+    }
+  }
+
+  return nil;
+}
+
++ (void)setNeedsUpdateOfSupportedInterfaceOrientations
+{
+  UIViewController *topController = RCTKeyWindow().rootViewController;
+  while (topController.presentedViewController) {
+    topController = topController.presentedViewController;
+  }
+
+  [topController setNeedsUpdateOfSupportedInterfaceOrientations];
+}
+
++ (void)requestGeometryUpdateForOrientationMask:(UIInterfaceOrientationMask)orientationMask
+{
+  UIWindowScene *scene = [RNSScreenWindowTraits windowSceneForOrientationUpdate];
+  if (scene == nil) {
+    return;
+  }
+
+  [RNSScreenWindowTraits setNeedsUpdateOfSupportedInterfaceOrientations];
+
+  UIWindowSceneGeometryPreferencesIOS *geometryPreferences =
+      [[UIWindowSceneGeometryPreferencesIOS alloc] initWithInterfaceOrientations:orientationMask];
+  [scene
+      requestGeometryUpdateWithPreferences:geometryPreferences
+                              errorHandler:^(NSError *_Nonnull error) {
+                                dispatch_async(dispatch_get_main_queue(), ^{
+                                  [RNSScreenWindowTraits setNeedsUpdateOfSupportedInterfaceOrientations];
+
+                                  UIWindowSceneGeometryPreferencesIOS *retryGeometryPreferences =
+                                      [[UIWindowSceneGeometryPreferencesIOS alloc]
+                                          initWithInterfaceOrientations:orientationMask];
+                                  [scene
+                                      requestGeometryUpdateWithPreferences:retryGeometryPreferences
+                                                              errorHandler:^(NSError *_Nonnull retryError) {
+                                                                RCTLogWarn(
+                                                                    @"[RNScreens] Failed to update interface orientation: %@",
+                                                                    retryError);
+                                                              }];
+                                });
+                              }];
+}
+#endif
+
 + (void)enforceDesiredDeviceOrientation
 {
 #if !TARGET_OS_TV && !TARGET_OS_VISION
   dispatch_async(dispatch_get_main_queue(), ^{
     UIInterfaceOrientationMask orientationMask = [RCTKeyWindow().rootViewController supportedInterfaceOrientations];
+
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(16_0)
+    if (@available(iOS 16.0, *)) {
+      // On iOS 16+ geometry updates are idempotent and should be driven by the
+      // desired mask directly. The scene's current interfaceOrientation may be
+      // stale during navigation transitions or after scene restoration.
+      [RNSScreenWindowTraits requestGeometryUpdateForOrientationMask:orientationMask];
+      return;
+    }
+#endif // Check for iOS 16
 
     UIInterfaceOrientation currentDeviceOrientation =
         [RNSScreenWindowTraits interfaceOrientationFromDeviceOrientation:[[UIDevice currentDevice] orientation]];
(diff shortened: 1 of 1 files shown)


ios/RNSScreenWindowTraits.mm
```diff
@@ -97,12 +97,84 @@ + (UIInterfaceOrientationMask)maskFromOrientation:(UIInterfaceOrientation)orient
...
+}
+
++ (void)setNeedsUpdateOfSupportedInterfaceOrientations
+{
+  UIViewController *topController = RCTKeyWindow().rootViewController;
+  while (topController.presentedViewController) {
+    topController = topController.presentedViewController;
+  }
+
+  [topController setNeedsUpdateOfSupportedInterfaceOrientations];
```
kmichalikk: you should wrap this in if(available iOS16) check, I'm getting warnings here and in multiple places in the requestGeometry... function

ios/RNSScreenWindowTraits.mm
```diff
@@ -97,12 +97,84 @@ + (UIInterfaceOrientationMask)maskFromOrientation:(UIInterfaceOrientation)orient
...
+  for (id connectedScene in connectedScenes) {
+    if ([connectedScene isKindOfClass:[UIWindowScene class]]) {
+      return connectedScene;
+    }
+  }
+
+  return nil;
+}
+
++ (void)setNeedsUpdateOfSupportedInterfaceOrientations
```
kmichalikk: wouldn't it be more correct to infer the controller from the scene you found in `windowSceneForOrientationUpdate`?

kmichalikk (changes requested):
Thanks for the PR. Please answer my comments. Also, I would like to have a step-by-step instructions on how to reproduce the bug. Most preferably, prepare an issue test in our repo with a code that could be put into one file. If expo is required, then paste the file in the PR and tell what to do besides create-expo-app, but please don't add unnecessary dependencies, expo-router, etc, if not necessary.

sorinc03:
Updated the PR description with step-by-step reproduction instructions and a one-file native-stack example. The branch already includes the iOS 16 availability guard and resolves the orientation update controller from the scene used for the geometry request.

kmichalikk (commented):
I tested some more and I can see that on our FabricExample the one-file example works correctly without patches, on expo with react-navigation template it works only if expo-dev-client is uninstalled, and it fails if it is installed, patch doesn't help there. So I am not able to successfully reproduce & see it being fixed, I don't know how to proceed here.

cc @kkafar

sorinc03:
Thanks for retesting. I agree this is a blocker for the PR as currently framed.

I checked the external repro again, and it is not a pure `react-native-screens` case: its README/workaround combines two changes:

- an `expo-screen-orientation` fix for finding the RNScreens controller behind `DevLauncherViewController`
- a `react-native-screens` change around the iOS 16+ geometry update path

This PR only targets the second part, so if the `expo-dev-client` setup still fails with this branch, I do not want to claim that this fixes that full repro. I will try to isolate a smaller case that fails on `main` and passes here, preferably in FabricExample / without `expo-dev-client`. If I cannot produce that, I agree the PR should not be merged as-is and I will close or reframe it.

PR #4122, opened 2026-06-02
feat(iOS, FormSheet v5): Add support for nativeContainerStyle

t0maboro:
## Description

Prep for applying some styling props directly to the native container that's hosting FormSheet's content.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1512

## Changes

- Updated public API
- Added native prop application

| iOS 18 | iOS 26 |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/038c96fe-1314-481f-a8a2-5a0d7cc34822" /> | <video src="https://github.com/user-attachments/assets/037eec80-efb7-40d2-959a-fcb320376c65" /> |

## Test plan

Added dedicated SFT.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/apps/src/tests/single-feature-tests/form-sheet/test-form-sheet-native-container-style-ios/scenario.md
+++ b/apps/src/tests/single-feature-tests/form-sheet/test-form-sheet-native-container-style-ios/scenario.md
@@ -24,7 +24,7 @@ TBD: Planned, but will be implemented separately.
 
 ---
 
-### Verification: Applying Native Background Color
+### Applying Default Native Background Color
 
 2. Tap the "Open FormSheet" button.
 
@@ -33,7 +33,7 @@ TBD: Planned, but will be implemented separately.
 
 ---
 
-### Verification: Layout Stability
+### Layout Stability
 
 3. Tap the "Expand Content" button.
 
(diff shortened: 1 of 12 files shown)


apps/src/tests/single-feature-tests/form-sheet/test-form-sheet-native-container-style-ios/scenario.md
LKuchno, line 1: Just one thing: why do we have the "Purple" Native Background Color option if we don't use it in the scenario? Maybe we could add 2 additional steps just to check if the color is applied correctly?

LKuchno (approved):

ios/gamma/modals/form-sheet/RNSFormSheetConfigurationApplicator.mm
```diff
@@ -90,6 +90,8 @@ - (void)applyConfigurationWithAppearanceProvider:(id<RNSFormSheetAppearanceProvi
     sheet.largestUndimmedDetentIdentifier = largestUndimmedDetentIdentifier;
   }];
 #endif // !TARGET_OS_TV
+
+  controller.view.backgroundColor = appearanceProvider.nativeContainerBackgroundColor ?: UIColor.clearColor;
```
kligarski, line 94: is the native default also transparent/clear/nil or does it have some background by default?

kligarski (approved):
Looks good but I have one question.

ios/gamma/modals/form-sheet/RNSFormSheetConfigurationApplicator.mm
t0maboro: this view is, from react perspective, my ContentView which I am inserting manually to teleport React children there and I explicitly set the bg color to transparent there https://github.com/software-mansion/react-native-screens/blob/main/ios/gamma/modals/form-sheet/RNSFormSheetContentView.mm#L11 to ensure that nothing is sampling from this specific view

kligarski: okay, thanks for the explanation

apps/src/tests/single-feature-tests/form-sheet/test-form-sheet-native-container-style-ios/scenario.md
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4122/commits/2e2aa7494653182ce0b9d829595caa575f9f4c27

PR #3727, opened 2026-03-05
feat(iOS): configure `RNSLog` utility

sgaczol:
## Description

This PR updates `RNSLog` function on iOS. It can be manually activated by using `RNS_DEBUG_LOGGING` flag during pods installation:

`RNS_DEBUG_LOGGING=1 rbenv exec bundle exec pod install`

`RNSLog` in `.mm` environment is based on `NSLog`.
`rnsLog` in `.swift` environment is based on `Swift.print`.

## Changes

- create Swift's `rnsLog` and update Obj-C one
- update `Podfile` and `RNScreens.podspec`
- update Ruby version limit to <= 3.3.6 in `TVOSExample`

## Before & after - visual documentation

No visuals changes.

## Test plan

Run app with set flags and check if it works properly.

## Checklist

- [ ] Included code example that can be used to test this change.
- [ ] Updated / created local changelog entries in relevant test files.
- [ ] For visual changes, included screenshots / GIFs / recordings documenting the change.
- [ ] For API changes, updated relevant public types.
- [ ] Ensured that CI passes

PR #4125, opened 2026-06-02
chore: bump lib and example apps to `react-native@0.86.0`

kligarski:
## Description

Bumps example apps (FabricExample, TVOSExample) and lib to `react-native@0.86.0`. Bumps `reanimated`, `worklets` and `gesture-handler` where applicable.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1503.

## Changes

- bump `react-native` to `0.86.0` for lib and `FabricExample` (`react-native-tvos` is still at rc.3)
- bump `react-native-gesture-handler` to `3.0.1`
- bump `react-native-reanimated` and `react-native-worklets` to `4.5.0-nightly-20260531-e6a8526c5` and `0.10.0-nightly-20260531-e6a8526c5` respectively
  - keep older version for lib to avoid TS problems - should be resolved separately
- add `.ruby-version` to `TVOSExample` (according to https://github.com/software-mansion/react-native-screens/pull/3727 it is necessary to use lower version, cc @sgaczol)

N/A

## Test plan

Run basic Stack v4 test screens & Tabs smoke tests both on Android and iOS. Verified build on Android TV and tvOS.

## Checklist

- [ ] Ensured that CI passes


kligarski:
Moving to draft until nightly with https://github.com/software-mansion/react-native-gesture-handler/pull/4228 is released.

kligarski:
One of the e2e tests fails on the CI but I think it's related to gesture-handler so I created an issue there: https://github.com/software-mansion/react-native-gesture-handler/issues/4246.

EDIT: this should be fixed here: https://github.com/software-mansion/react-native-gesture-handler/pull/4166

We also have these outdated references in our code - we might want to check if we should remove them soon - I'll add this to the ticket with removal of RCTCxxBridge.

FabricExample/package.json
```diff
@@ -44,11 +44,10 @@
     "@react-native-community/cli": "20.1.0",
     "@react-native-community/cli-platform-android": "20.1.0",
     "@react-native-community/cli-platform-ios": "20.1.0",
-    "@react-native/babel-preset": "0.85.0",
-    "@react-native/eslint-config": "0.85.0",
-    "@react-native/jest-preset": "0.85.0",
```
t0maboro, line 49: wondering why they decided to remove that dependency in the upgrade helper, as it was added from 84->85 migration, and in `jest.config.js` it remains - unrelated to the PR's content

t0maboro (approved):
The PR looks good to me.

Do we know the timeline for the RNGH PR merge and release? Unless this is blocking us, I'd rather wait for the CI to pass before merging.

FabricExample/package.json
kligarski: no idea tbh.

PR #4147, opened 2026-06-10
chore(test): add e2e test for test-tabs-system-item-ios scenario

LKuchno:
## Description

This PR implements the Detox e2e test suite for the `test-tabs-system-item-ios` scenario, covering the `systemItem` prop on iOS tab bar items. 

The test covers: static system item rendering, runtime `systemItem` cycling (favorites/history/search), title override cycling (custom/hidden/system), icon override cycling (custom SFSymbols vs. system fallback), and combinations of all three overrides. All test blocks are wrapped in `describeIfiOS` since this feature is iOS-only. The scenario.md is updated to reflect what is and isn't automated.

**⚠️ Notes**

The following are explicitly not automated (documented in `scenario.md`):
- Validating icon vs. selectedIcon differences
- Checking visual icon and label changes.
- "Hidden title" is verified indirectly only; manual verification recommended
- Steps 20-23 (orientation changes).

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1410

## Changes

- **E2E/Tabs**: Add `test-tabs-system-item-ios.e2e.ts` - new Detox test suite 
- **Test Screen**: Add `testID` attributes to config state display labels and section headers; add `tabBarItemTestID` to both tab route configso enable reliable Detox targeting
- **Scenario Docs**: Update `scenario.md` to replace "TBD" with the actual automation status, listing known gaps


FabricExample/e2e/single-feature-tests/tabs/test-tabs-system-item-ios.e2e.ts
```diff
@@ -0,0 +1,415 @@
...
+
+async function tapSystemTitleOption() {
+  await element(by.text('system')).atIndex(0).tap();
+}
+
+async function tapSystemIconOption() {
+  await element(by.text('system')).atIndex(1).tap();
+}
+
+describeIfiOS('Tab Bar System Item (iOS 26)', () => {
```
kligarski: I wonder if we should skip running this e2e completely if iOS < 26 is used but maybe it's not necessary.

FabricExample/e2e/single-feature-tests/tabs/test-tabs-system-item-ios.e2e.ts
```diff
@@ -0,0 +1,415 @@
...
+        element(by.id('star.fill').and(by.label('favorite'))).atIndex(0),
+      ).toExist();
+      await expect(
+        element(by.id('magnifyingglass').and(by.label('Search'))).atIndex(0),
+      ).not.toExist();
+    });
+  });
+
+  describe('Runtime Config tab — title override cycling', () => {
+    it('should update the tab bar item label after tapping custom title', async () => {
```
kligarski:
```suggestion
    it('should update the tab bar item label after switching to custom title', async () => {
```
nit: I think we should focus here on the result/intention of the action, not the literal action of tapping the button

kligarski: applies to other places as well

kligarski (approved):

FabricExample/e2e/single-feature-tests/tabs/test-tabs-system-item-ios.e2e.ts
LKuchno: I rewrite this test and now its suitable for iOs26 and ios18 and lower

FabricExample/e2e/single-feature-tests/tabs/test-tabs-system-item-ios.e2e.ts
```diff
@@ -1,24 +1,52 @@
...
+}
+
 async function tapSystemTitleOption() {
   await element(by.text('system')).atIndex(0).tap();
 }
 
 async function tapSystemIconOption() {
   await element(by.text('system')).atIndex(1).tap();
 }
+function isIOSVersionAtLeast(version: string): boolean {
```
kligarski, line 35:
```suggestion

function isIOSVersionAtLeast(version: string): boolean {
```
nit

FabricExample/e2e/single-feature-tests/tabs/test-tabs-system-item-ios.e2e.ts
```diff
@@ -0,0 +1,534 @@
...
+  return (
+    device.getPlatform() === 'ios' &&
+    isVersionEqualOrHigherThan(getIOSVersionNumber(), version)
+  );
+}
+
+const tabBarButtonType = isIOSVersionAtLeast('26.0')
+  ? '_UITabButton'
+  : 'UITabBarButton';
```
kligarski, line 44: This is something that might break in the future if there are some changes to native hierarchy in next iOS versions but I don't think we have really any other option here so it's okay.

kligarski (approved):

PR #4148, opened 2026-06-10
feat: Handle onPress in iOS Menu

kmichalikk:
Closes https://github.com/software-mansion/react-native-screens-labs/issues/1521

## Description

This PR adds support for `onPress` in iOS Stack v5 header.

## Changes

- added required `id` for all menuElements, currently used for `onPress` in menu items
- added optional `onPress` for menu item - a callback defined per menu item

N/A

## Test plan

Use `test-stack-header-menu-ios`, follow the scenario.md

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/gamma/stack/header/RNSStackHeaderConfigEventEmitter.mm
+++ b/ios/gamma/stack/header/RNSStackHeaderConfigEventEmitter.mm
@@ -1,4 +1,5 @@
 #import "RNSStackHeaderConfigEventEmitter.h"
+#import <React/RCTConversions.h>
 #import <React/RCTLog.h>
 
 @implementation RNSStackHeaderConfigEventEmitter {
@@ -8,7 +9,7 @@
 - (BOOL)emitOnMenuItemPress:(NSString *)menuElementId
 {
   if (_reactEventEmitter != nullptr) {
-    _reactEventEmitter->onMenuItemPress({.menuElementId = menuElementId.cString});
+    _reactEventEmitter->onMenuItemPress({.menuElementId = RCTStringFromNSString(menuElementId)});
     return YES;
   } else {
     RCTLogWarn(@"[RNScreens] Skipped OnMenuItemPress event emission due to nullish emitter");
--- a/ios/gamma/stack/header/RNSStackHeaderConfigEventEmitter.mm
+++ b/ios/gamma/stack/header/RNSStackHeaderConfigEventEmitter.mm
@@ -1,0 +1,24 @@
+#import "RNSStackHeaderConfigEventEmitter.h"
+#import <React/RCTLog.h>
+
+@implementation RNSStackHeaderConfigEventEmitter {
+  std::shared_ptr<const react::RNSStackHeaderConfigIOSEventEmitter> _reactEventEmitter;
+}
+
+- (BOOL)emitOnPressMenuItem:(NSString *)menuElementId
+{
+  if (_reactEventEmitter != nullptr) {
+    _reactEventEmitter->onPressMenuItem({.menuElementId = menuElementId.cString});
+    return YES;
+  } else {
+    RCTLogWarn(@"[RNScreens] Skipped OnPressMenuItem event emission due to nullish emitter");
+    return NO;
+  }
+}
+
+- (void)updateEventEmitter:(const std::shared_ptr<const react::RNSStackHeaderConfigIOSEventEmitter> &)emitter
+{
+  _reactEventEmitter = emitter;
+}
+
+@end
--- a/ios/gamma/stack/header/RNSStackHeaderConfigEventEmitter.mm
+++ b/ios/gamma/stack/header/RNSStackHeaderConfigEventEmitter.mm
@@ -6,10 +6,10 @@
   std::shared_ptr<const react::RNSStackHeaderConfigIOSEventEmitter> _reactEventEmitter;
 }
 
-- (BOOL)emitOnMenuItemPress:(NSString *)menuElementId
+- (BOOL)emitOnMenuItemPress:(NSString *)menuItemId
 {
   if (_reactEventEmitter != nullptr) {
-    _reactEventEmitter->onMenuItemPress({.menuElementId = RCTStringFromNSString(menuElementId)});
+    _reactEventEmitter->onMenuItemPress({.menuItemId = RCTStringFromNSString(menuItemId)});
     return YES;
   } else {
     RCTLogWarn(@"[RNScreens] Skipped OnMenuItemPress event emission due to nullish emitter");
--- a/ios/gamma/stack/header/RNSStackHeaderConfigEventEmitter.mm
+++ b/ios/gamma/stack/header/RNSStackHeaderConfigEventEmitter.mm
@@ -5,13 +5,13 @@
   std::shared_ptr<const react::RNSStackHeaderConfigIOSEventEmitter> _reactEventEmitter;
 }
 
-- (BOOL)emitOnPressMenuItem:(NSString *)menuElementId
+- (BOOL)emitOnMenuItemPress:(NSString *)menuElementId
 {
   if (_reactEventEmitter != nullptr) {
-    _reactEventEmitter->onPressMenuItem({.menuElementId = menuElementId.cString});
+    _reactEventEmitter->onMenuItemPress({.menuElementId = menuElementId.cString});
     return YES;
   } else {
-    RCTLogWarn(@"[RNScreens] Skipped OnPressMenuItem event emission due to nullish emitter");
+    RCTLogWarn(@"[RNScreens] Skipped OnMenuItemPress event emission due to nullish emitter");
     return NO;
   }
 }
--- a/src/components/gamma/stack/header/StackHeaderConfig.ios.tsx
+++ b/src/components/gamma/stack/header/StackHeaderConfig.ios.tsx
@@ -41,7 +41,6 @@ export default function StackHeaderConfig(props: StackHeaderConfigProps) {
         ...(leadingItems ?? []).filter(it => it && it.type === 'item'),
         ...(trailingItems ?? []).filter(it => it && it.type === 'item'),
       );
-      console.log(event.nativeEvent.menuItemId);
       const menuElement = findMenuElementByIdInItems(
         items,
         event.nativeEvent.menuItemId,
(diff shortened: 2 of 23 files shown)


src/components/gamma/stack/header/ios/StackHeaderMenu.ios.types.ts
```diff
@@ -1,9 +1,12 @@
 export interface StackHeaderMenuItem {
+  menuElementId: string;
```
kligarski: 1. do we need `menuElement` prefix if we're already in menu element?
2. should we unify naming with regular items (`key` vs `id`)?

src/components/gamma/stack/header/ios/StackHeaderMenu.ios.types.ts
```diff
@@ -1,9 +1,12 @@
...
   type: 'menuItem';
   title?: string | undefined;
+  onPress?: () => void;
```
kligarski:
```suggestion
  onPress?: () => void | undefined;
```

src/fabric/gamma/stack/StackHeaderConfigIOSNativeComponent.ts
```diff
@@ -14,6 +16,8 @@ export interface NativeProps extends ViewProps {
   largeTitle?: string | undefined;
   largeSubtitle?: string | undefined;
   largeTitleEnabled?: CT.WithDefault<boolean, false>;
+
+  onPressMenuItem?: CT.DirectEventHandler<PressMenuItemEvent> | undefined;
```
kligarski:
```suggestion
  onMenuItemPress?: CT.DirectEventHandler<MenuItemPressEvent> | undefined;
```
nit, up to you

src/components/gamma/stack/header/ios/StackHeaderMenu.ios.types.ts
kligarski: same applies to the content of `PressMenuItemEvent`

src/components/gamma/stack/header/StackHeaderConfig.ios.types.ts
```diff
@@ -6,16 +6,20 @@ export interface StackHeaderBaseItemIOS {
   label?: string | undefined;
 }
 
-export interface StackHeaderInlineItemIOS extends StackHeaderBaseItemIOS {
-  type: 'item';
+export interface SupportsMenuIOS {
```
kligarski, line 9: tbh, do we even need this base type & do we need it exported? I'm not convinced about the name here.

src/components/gamma/stack/header/StackHeaderConfig.ios.tsx
```diff
@@ -32,14 +35,32 @@ export default function StackHeaderConfig(props: StackHeaderConfigProps) {
     largeTitleEnabled,
   } = ios ?? {};
 
+  const handleMenuPress = useCallback(
+    (event: NativeSyntheticEvent<PressMenuItemEvent>) => {
+      const items = Array.of(
+        ...(leadingItems ?? []).filter(it => it && it.type === 'item'),
+        ...(trailingItems ?? []).filter(it => it && it.type === 'item'),
+      );
+      const menu = findMenuElementByIdInItems(
```
kligarski:
```suggestion
      const menuElement = findMenuElementByIdInItems(
```
nit

src/components/gamma/stack/header/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+export function findMenuElementByIdInItems(
+  items: SupportsMenuIOS[],
+  menuElementId: string,
+): StackHeaderMenuElement | null {
+  for (const item of items) {
+    if (item.menu === undefined) {
+      continue;
+    }
+
+    const menu = findMenuElementById(item.menu, menuElementId);
```
kligarski:
```suggestion
    const menuElement = findMenuElementById(item.menu, menuElementId);
```

ios/gamma/stack/header/RNSStackHeaderConfigEventEmitter.mm
```diff
@@ -0,0 +1,24 @@
...
+#import <React/RCTLog.h>
+
+@implementation RNSStackHeaderConfigEventEmitter {
+  std::shared_ptr<const react::RNSStackHeaderConfigIOSEventEmitter> _reactEventEmitter;
+}
+
+- (BOOL)emitOnPressMenuItem:(NSString *)menuElementId
+{
+  if (_reactEventEmitter != nullptr) {
+    _reactEventEmitter->onPressMenuItem({.menuElementId = menuElementId.cString});
```
kligarski: e.g. in tabs we use `RCTStringFromNSString`, maybe we should do so here as well

```
{.selectedScreenKey = RCTStringFromNSString(payload.selectedScreenKey),
```

kligarski (approved):

t0maboro (approved):

src/components/gamma/stack/header/ios/StackHeaderMenu.ios.types.ts
kmichalikk: 1. menuElementId applies to both menu and menu item so having the name with a prefix common to both -for me - should indicate that they're linked

2. wdym "regular items"? I'm not sure about `key` field since it is associated with react internals

src/components/gamma/stack/header/StackHeaderConfig.ios.types.ts
kmichalikk: I like interface segregation here + I'm using this to type find function that searches for menu with specific id

src/components/gamma/stack/header/ios/StackHeaderMenu.ios.types.ts
kmichalikk: [2c35d9a](https://github.com/software-mansion/react-native-screens/pull/4148/commits/2c35d9add05033a3616cfba3905dc5d421b5d1d5)

src/fabric/gamma/stack/StackHeaderConfigIOSNativeComponent.ts
kmichalikk: [2c35d9a](https://github.com/software-mansion/react-native-screens/pull/4148/commits/2c35d9add05033a3616cfba3905dc5d421b5d1d5)

src/components/gamma/stack/header/StackHeaderConfig.ios.tsx
kmichalikk: [2c35d9a](https://github.com/software-mansion/react-native-screens/pull/4148/commits/2c35d9add05033a3616cfba3905dc5d421b5d1d5)

src/components/gamma/stack/header/utils.ts
kmichalikk: [2c35d9a](https://github.com/software-mansion/react-native-screens/pull/4148/commits/2c35d9add05033a3616cfba3905dc5d421b5d1d5)

ios/gamma/stack/header/RNSStackHeaderConfigEventEmitter.mm
kmichalikk: [e04deae](https://github.com/software-mansion/react-native-screens/pull/4148/commits/e04deae18b1adee7e0ebde8f01456a670d83ab03)

src/components/gamma/stack/header/ios/StackHeaderMenu.ios.types.ts
kligarski: 1. But if I pass children to prop called `StackHeaderMenuElement[]` I already know that I'm adding menu elements so I don't see the reason to repeat it and make user type `menuElementId` instead of `id`. But you can ask for other opinions and then decide.

2. Are you going to support runtime changes to bar button items via view commands? E.g. changing a title/icon? If so, then we will need an id. If we need an `id` prop, we should use it for React `key` as well, right? It also removes some confusion for the user about what is the difference between key in bar button items and id in menu items and it matches Android where bar button items and menu items are equivalent.

src/components/gamma/stack/header/StackHeaderConfig.ios.types.ts
kligarski: Is it going to be exposed to public API? If so we should rename it to have `StackHeader` prefix and export it from `gamma/stack/index.ts`. If not, then I guess it's okay.

BTW, check if all types are exported. I think we're missing e.g. `StackHeaderMenu`, right? And the types should have `IOS` suffix to match other types.

kligarski (commented):
Answered some questions in threads from previous review.

src/components/gamma/stack/header/ios/StackHeaderMenu.ios.types.ts
kmichalikk: 1. okay so where would You leave menuElementId? I think the nativeEvent should still have menuElementId because id is nondescriptive there

for 2. I don't think there is a connection since this is inside a plain object passed to _one_ item (which has a key) or to some prop (titleMenu which will be added later), there is no react representation here

kligarski: 1. but menu can't be pressed, right? so maybe it should be menuItemId in native event. Either way, I'm referring to public api mostly.
2. I don't understand what you mean. If we will add support for changing the bar button item via view command, we will need to reference it by some `id` (e.g. `setBarButtonMenuItem("button1", {title: "New Title"})`). If so, we can only use `id` prop on bar button menu item and don't ask the user for `key` prop separately, just use the `id`. For menu items we will use the same name `id` as well. then both menu items and bar button items will require passing `id`.

kmichalikk: I don't understand why would we need a `key` prop in the first place

kmichalikk: [6260b8a](https://github.com/software-mansion/react-native-screens/pull/4148/commits/6260b8ac0b1aaad51c65fba15e69b2e3135e852a)

src/components/gamma/stack/header/StackHeaderConfig.ios.types.ts
kmichalikk: [5cbbbc6](https://github.com/software-mansion/react-native-screens/pull/4148/commits/5cbbbc6d0812ae3691d384ff5ce30e10750235f9)

kligarski (approved):

PR #4149, opened 2026-06-10
fix(Android): drive ScreenFooter keyboard insets from the sheet, not the decor view

stachbial:
## Description

On Android, `ScreenFooter` installs a `WindowInsetsAnimationCompat.Callback` with `DISPATCH_MODE_STOP` on the activity's decor view (in its `init` block) and never removes it. Android keeps a single insets-animation callback slot per view, and `DISPATCH_MODE_STOP` stops dispatch to the entire subtree of the view it is set on — and the decor view is the root of the window. As a result, once a single formSheet with `unstable_sheetFooter` mounts, no other view in the window ever receives `onStart` / `onProgress` / `onEnd` again, and the callback stays installed even after the footer unmounts.

With edge-to-edge enabled (the default since React Native 0.81, enforced on Android 15+), `WindowInsetsAnimation` dispatch is the only keyboard signal libraries like reanimated (`useAnimatedKeyboard`) and `react-native-keyboard-controller` rely on. So mounting one sheet footer permanently breaks keyboard handling/animations in the whole app until the process is restarted.

The interference was already flagged as a known issue in the code itself:

```kotlin
// Note that we do override insets animation on given view. I can see it interfering e.g.
// with reanimated keyboard or even other places in our code. Need to test this.
```

### Minimal reproduction

React Native 0.83 / react-native-screens 4.x, Android only (reproduced on API 33 and API 36). Included in this PR as `apps/src/tests/issue-tests/TestScreenFooterKeyboardInsets.tsx` — a Home screen with a `TextInput` translated by `useAnimatedKeyboard`, plus a formSheet route with `unstable_sheetFooter`.

Steps (against unpatched `main`):

1. Focus the `TextInput` — the input animates up with the keyboard (insets animation dispatch works).
2. Dismiss the keyboard, open the form sheet (the footer mounts), close the sheet.
3. Focus the `TextInput` again — the keyboard animation no longer runs (`useAnimatedKeyboard` reports nothing), and it never recovers for the lifetime of the process.

## Changes

- `ScreenFooter` no longer installs a `WindowInsetsAnimationCompat.Callback` on the decor view — the `init` block and all decor-view access are removed. It exposes three methods the sheet calls instead — `notifyKeyboardAnimationStart()`, `handleKeyboardInsetsProgress(insets)`, `notifyKeyboardAnimationEnd()` — which run the same per-frame work (`layoutFooterOnYAxis`) the old `onStart` / `onProgress` / `onEnd` did.
- The form sheet already observes the IME animation on its own `Screen`-level `WindowInsetsAnimationCompat.Callback` in `ScreenStackFragment` (the one that drives `sheetDelegate.handleKeyboardInsetsProgress`). `SheetDelegate` now forwards each keyboard frame to both the sheet animation coordinator (which moves the sheet body) and the footer. `ScreenStackFragment` is unchanged.

The decor view's single insets-animation callback slot is never touched, so reanimated / keyboard-controller keep receiving keyboard animations.

### Why drive the footer from SheetDelegate

The footer cannot listen on itself or its `Screen` parent: every formSheet's `Screen` view carries its own `DISPATCH_MODE_STOP` callback (`ScreenStackFragment`), so nothing inside the sheet's subtree receives insets-animation dispatch — which is why the footer reached up to the decor view in the first place. Instead of taking that shared slot, the footer is now fed from the one place that already receives the sheet's IME animation: the `Screen`-level callback that drives `SheetDelegate`. `SheetDelegate` already owns the footer (it registers it with the sheet behaviour), so the keyboard hand-off lives next to that.

## Test plan

- `apps/src/tests/issue-tests/TestScreenFooterKeyboardInsets.tsx` (added in this PR): after opening and closing the form sheet (footer mount + unmount), focusing the `TextInput` on the Home screen still animates with the keyboard. On unpatched `main`, step 3 of the reproduction fails. A form sheet **without** a footer never breaks the keyboard, confirming the cause is footer-specific.
- `apps/src/tests/issue-tests/TestFormSheet.tsx` with the commented-out `unstable_sheetFooter: FormSheetFooter` option re-enabled: focusing a `TextInput` inside the sheet still raises the footer above the keyboard — the footer keeps tracking the keyboard, now via the frames relayed from `SheetDelegate`.

## Checklist

- [ ] Ensured that CI passes


t0maboro:
Hi @stachbial , thanks for reporting the issue and opening the PR!

The root cause appears to have been identified correctly. However, while testing the fix in FabricExample environment, I'm still able to reproduce the issue using the provided example.

Could you please confirm how this behaves in your setup and whether you're able to reproduce it as well?

I can continue investigating on my side and do some additional debugging to check what's going on.

https://github.com/user-attachments/assets/5df9656e-6b80-48b9-908d-96ab11a7186a

stachbial:
> Hi @stachbial , thanks for reporting the issue and opening the PR!
> 
> The root cause appears to have been identified correctly. However, while testing the fix in FabricExample environment, I'm still able to reproduce the issue using the provided example.
> 
> Could you please confirm how this behaves in your setup and whether you're able to reproduce it as well?
> 
> I can continue investigating on my side and do some additional debugging to check what's going on.
> 
>  4149.mov



> Hi @stachbial , thanks for reporting the issue and opening the PR!
> 
> The root cause appears to have been identified correctly. However, while testing the fix in FabricExample environment, I'm still able to reproduce the issue using the provided example.
> 
> Could you please confirm how this behaves in your setup and whether you're able to reproduce it as well?
> 
> I can continue investigating on my side and do some additional debugging to check what's going on.
> 
>  4149.mov

Yep, you nailed it — thanks for actually running it 🙏

What the first version did: it kept the footer's callback on the decor view and just (a) switched it from DISPATCH_MODE_STOP to DISPATCH_MODE_CONTINUE_ON_SUBTREE and (b) moved registration into onAttachedToWindow 
/ onDetachedFromWindow, with a WeakReference guard so one footer's detach couldn't clear a newer footer's callback.

However, that only helps consumers listening on a child of the decor view . `useAnimatedKeyboard` registers its own animation callback on the decor view itself (WindowsInsetsManager → 
setWindowInsetsAnimationCallback(decorView, …)), and with a single callback slot per view the footer just overwrites reanimated's — dispatch mode doesn't change that, and on detach we restored the slot to null 
rather than to reanimated's, so it stayed dead. It looked fixed on my end only because my app uses react-native-keyboard-controller, whose observer sits on a child view, so the CONTINUE change was enough there. 
Your reanimated repro is exactly the case the dispatch-mode tweak can't cover.

What changed: the footer no longer touches the decor view at all. The sheet already receives the IME animation on its own Screen-level callback (ScreenStackFragment → SheetDelegate), so I drive the footer from 
there — SheetDelegate forwards each frame to both the coordinator (sheet body) and the footer.

Why this approach seems nicer:

- the footer stops reaching out to a shared window-level view it doesn't own;
- nothing competes for the decor slot, so reanimated / keyboard-controller / anything else on it just keep working — it removes the cause rather than mitigating the dispatch-blocking symptom;
- no slot arbitration to reason about (no WeakReference bookkeeping, no clearing-on-detach, no risk of clobbering another callback);
- marginally lighter too — one fewer insets-animation callback in the per-frame dispatch chain, folded into the screen-level one that already runs 😁

Verified on FabricExample, with the footer useAnimatedKeyboard keeps animating across open/close cycles, and an input inside the sheet still raises the footer above the keyboard. Updated the PR and the description — let me know if you'd rather drive the footer differently aand sorry for the prior incomplete fix 😜

t0maboro (approved):
We discussed with @stachbial  the root cause and impact of this PR internally. We have IME insets callbacks set up per screen (added to move the FormSheet above the keyboard some time ago). I think that it makes sense to associate the sheet footer with the same path rather than relying on the callback registered on the decorview (which may override another insets animation callback).

@stachbial thanks for fixing that!

PR #4161, opened 2026-06-12
fix(Android, Tabs): fix flickering tab bar height & contents when entering on ScreenStackFragment

kkafar:
This is a small commit, however it introduces profound changes and is
based on delicate interactions between our components and the framework.

The intention here is to fix an issue, where the contents of a `BottomNavigationView`,
when nested inside a `ScreenStack` (stack v4) would jump on enter animation,
the activity indicator view would appear mid transition etc.
This very issue is fixed by swapping usage of `postFrameCallback` with
`post` in `TabsHost` **before insets are applied. 
Let's explain first what is the difference between
the two.

`ReactChoreographer` is a `Choreographer.FrameCallback` and registers
itself as the callback on the actual `Choreographer` instance.

`Choreographer`'s frame callbacks are triggered on VSync signal from the
platform's renderer, by posting an asynchronous message to the main
thread's message queue. This means, that when we add a callback to the
`ReactChoreographer` we have a guarantee to be executed before the next
frame is drawn, because the view traversals (measure, layout, draw) are
executed after user-added frame callbacks. This has one big drawback -
the layout we schedule with `postFrameCallback` will be executed BEFORE
system dispatches insets, therefore the result of the layout
calculation won't take them into consideration and effectively won't be
correct. If we draw a frame before we recompute the layout again with insets,
we will observe a `BottomNavigationView` height jump.

The solution here is usage of the `post` to simply defer the layout
computation until the insets are dispatched. Do we have a guarantee
that the insets will be dispatched before our `post` callback is executed?
Yes, provided that before we call `post` we (or anybody else) request
inset dispatch from the hierarchy root (`BottomNavigationView` does this
for us). When insets are requested, the `ViewRootImpl` adds a
*barrier message* to the message queue, which naturally will be
processed before our `post` callback, and will block processing of
subsequent *regular messages* until vsync comes and traversals are executed.

Okay, so the problem of jumping tab bar height is fixed, but there is
still another problem with *activity indicator* (a round view that
emphasizes the currently selected item icon). THe problem is that the
`NavigationBarItemView` instances are not created and added to
`NavigationBarMenuView` until we create & add items. This is done in
`TabsContainer.performContainerUpdate`, just before we trigger a
fragment transaction adding the selected tab, in turn, this happens
synchronously after (sometimes before) we schedule the layout with `post`.
That would be fine if the *activity indicator* had been laid out
synchronously with the rest of the `BottomNavigationView` contents, but
its not the case. The `NavigationBarItemView` schedules with `post` a callback
that updates the *activity indicator* **layout params** (not the
frame!), which triggers `parent.requestLayout()` and effectively
schedules yet another `post` callback from `TabsHost` with the layout callback.
Therefore it seems, that the activity indicator would appear after a
couple frames of the transition have already been drawn, but it is not
the case. **I'm not sure why**, but the activity indicator appears right
now immediately on the very first animation frame in the right place and
has good dimensions. I'm gonna let this be just because it works, but
it could definitely be investigated further (maybe it takes a couple of
frames to start the transition? Maybe it's `SafeAreaView`
`ViewTreeObserver.OnPreDrawListener` that blocks the first frame from
being drawn until the layout is correct? Not sure).

Okay, so the tab bar now seems to be working correctly, but we have
created yet another problem. This time we've broken the exit transition
of the screen we're pushed on top of.

Stack v4 uses the *Animation API* to animate the screens. When, in
result of a *fragment transaction* fragment's view will be removed from
the view hierarchy, the framework will mark the view as disappearing by
calling `screenStack.startViewTransition(child)` and later observing
that the `child` is detached from the hierarchy. This allows the
disappearing view to be still drawn during the transition. It has to
call `endViewTransition` at some point. To ensure it's called "at the
right moment", [just after it starts the view transition it also
schedules an accompanying `Animation` on the view](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:fragment/fragment/src/main/java/androidx/fragment/app/DefaultSpecialEffectsController.kt;l=562-591;drc=c53e65302e39ba824260a5b9e0831d872aa786d1). This is a subclass of
`Animation`: `FragmentAnim.EndViewTransitionAnimation`. It ends the
transition on animation end OR after [it detects that it has to been
queried for transformation between two *post callbacks*.](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:fragment/fragment/src/main/java/androidx/fragment/app/FragmentAnim.java;l=226-268;drc=c53e65302e39ba824260a5b9e0831d872aa786d1).
This way it determines that the animation is "no longer active" and calls
the `endViewTranstion`. This "heuristic" is completely wrong though, in
case ANY `ViewTreeObserver.OnPreDrawListener` blocks the draw process by
returning `false` from `onPreDraw`. In particular our `SafeAreaView`
does exactly that, because it needs to wait until the layout is correct
and insets are applied before it allows the first frame to be drawn.
This means that the transition will be ended prematurely, right after the
first frame is drawn, which causes the screen to simply disappear and
therefore a flash.

To workaround the interaction described above I've added a hack in
`ScreensCoordinatorLayout` & `ScreenStack` that blocks the premature
calls to `endViewTransition` and moves the responsibility of ending the
transition to the `ScreenStack`.

We've considered an alternative approach, but this one, despite being
a hack, seemed like the least disruptive one. Considered alternative:

* Simply return `true` from `SafeAreaView`'s `onPreDraw`.

    This would allow the transition to be ended at the right moment, but it
    might cause a flash of un-inset content during the first frame of the transition,
    which is even worse than the current issue.
    I haven't been able to trigger that effect though. The
    `SafeAreaView` seemed to be working correctly despite that change,
    but decided it's a real risk.

It seems that this is not an end of the problems. We've noticed that
when we set a large font size for the tab bar item label, the safe area
view receives few incorrect layouts. The tab bar is thankfully (:D)
stable, however the safe area jump is visible.

---

## Why we need to keep both scheduling methods - `Handler.post` and `ReactChoreographer` - TransitionManager and Material's Animator (ChangeBounds)

While `post` solves the inset jump during the **initial** mount, relying on it permanently introduces a regression: it breaks the Material tab switch animation.

`forceSubtreeMeasureAndLayoutPass` forces a measure/layout pass across the subtree, ensuring the layout reaches native views embedded deep within the React hierarchy (React views have empty overrides for native layout logic). `TabsHostLayoutCoordinator` chooses between two scheduling methods depending on whether insets were already propagated:

- **`postLayoutToMessageQueue`** - uses `Handler.post`. Executes via the standard message queue. If a layout traversal is already scheduled or ongoing, this will execute after it. Used until insets are propagated.
- **`postLayoutToReactChoreographer`** - uses `ReactChoreographer`'s `NATIVE_ANIMATED_MODULE` queue. Synchronizes with vsync, guaranteeing execution JUST BEFORE the upcoming traversal. Used afterwards.

### The transition mechanism

When a tab is selected, `NavigationBarMenuView.updateMenuView()` triggers an `AutoTransition` (Fade + ChangeBounds + TextScale) via `TransitionManager.beginDelayedTransition()`. The TransitionManager works in three distinct steps:

1. It immediately and synchronously captures `startValues` for the view subtree (before any new layout is performed).
2. It waits for the layout pass to happen.
3. It installs a `ViewTreeObserver.OnPreDrawListener` to capture `endValues` just before the frame is drawn (after measure/layout).

Animators are built based on the diff between `startValues` and `endValues`.

### Path A - layout via `Handler.post` (before insets)

The forced `forceSubtreeMeasureAndLayoutPass` is queued in the Looper. It does NOT run between the transition's `startValues` snapshot and its `endValues` capture.

- **Result:** from the transition's point of view, the subtree bounds are identical. Only properties like `android:fade:transitionAlpha` and visibility change.
- **Effect:** a clean Fade animator is created, NO ChangeBounds animation is performed, preventing the unwanted content jump during screen transition.

### Path B - layout via `ReactChoreographer` (after insets)

The forced layout runs in the SAME frame (before `dispatchOnPreDraw`). The subtree is forcefully laid out with new dimensions BEFORE the transition captures `endValues`.

- **Result:** `android:changeBounds:bounds` for `BottomNavigationItemView` differ significantly. This also affects inner views like `navigation_bar_item_content_container` and `navigation_bar_item_labels_group`.
- **Effect:** ChangeBounds animators are generated and the BottomNavigationView's content is animated on tab selection, restoring the smooth Material animation for tab switching. The active indicator layout (with non-zero layout params) is deferred via `Handler.post` in `NavigationBarItemView.onSizeChanged`. When it happens during the transition from another screen of the native-stack, `ChangeBounds` suppresses layout on animator creation and prevents recalculating the active indicator's layout until the `AutoTransition` of `BottomNavigationView` ends.

Once insets are applied, we switch **permanently** to the `ReactChoreographer` path so subsequent tab-selection changes trigger the full `ChangeBounds` animations.

### Why `onPreDraw` is a good signal that insets were propagated

We flip from path A to path B once the view tree has first been laid out with window insets applied, which we detect via `onPreDraw`. This is a reliable signal: `ViewRootImpl.performTraversals()` runs in the Choreographer `CALLBACK_TRAVERSAL` phase, and within a single traversal it:

1. computes and dispatches window insets down the hierarchy (`dispatchApplyWindowInsets`),
2. runs measure + layout,
3. invokes `ViewTreeObserver.dispatchOnPreDraw()` - immediately before `performDraw()`.

So by the time `onPreDraw` fires, the frame's insets have been applied.


apps/src/tests/issue-tests/index.ts
```diff
@@ -194,6 +194,8 @@ export { default as Test3885 } from './Test3885';
 export { default as Test3910 } from './Test3910';
 export { default as Test4027 } from './Test4027';
 export { default as Test4064 } from './Test4064';
+export { default as Test4090 } from './Test4090';
```
t0maboro, line 197: note: this is the old test placed wrong, I'm moving it

kkafar (commented):
I have a request:

Can we also add a *component-integration-test* `test-stack-tabs-tabs-in-stack-stable-enter-transition`, where we'll test integration of new `Stack` with `Tabs`? 

what do you think?

t0maboro:
> I have a request:
> 
> Can we also add a _component-integration-test_ `test-stack-tabs-tabs-in-stack-stable-enter-transition`, where we'll test integration of new `Stack` with `Tabs`?
> 
> what do you think?

added

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/host/TabsHost.kt
```diff
@@ -118,33 +121,85 @@ class TabsHost(
...
 
-    private val layoutCallback =
-        Choreographer.FrameCallback {
-            isLayoutEnqueued = false
-            forceSubtreeMeasureAndLayoutPass()
-        }
-
+    /**
+     * Forces a measure/layout pass over our subtree after a React-driven update. Chooses between two
+     * scheduling methods depending on whether the insets were already propagated.
```
kkafar: > Forces a measure/layout pass over our subtree after a React-driven update.

I believe it to not be true. 
When e.g. `BottomNavigationView` changes its layout it might call `requestLayout`, which will trigger the `requestLayout` in `TabsHost` and in result call `refreshLayout`. This method is not only called on a React-driven update. It's called in general from our `requestLayout` override, which might be called due to a wide gamut of reasons. 

Is that right?

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/host/TabsHost.kt
```diff
@@ -118,27 +122,79 @@ class TabsHost(
...
+     *
+     * - [TabsHostLayoutCoordinator.postLayout] - `Handler.post`, runs on the NEXT frame. Used until the insets
```
kkafar: That is also not accurate I believe, but to discuss we need to define precisely what do we mean by "run on current frame" and "run on next frame".

By "run on current frame" I mean the callback to be executed BEFORE OR DURING the next or **currently ongoing*** traversal (vsync tick)

By "run on the next frame" I mean the callback to be executed AFTER closest vsync tick.

Now, `postLayout` will run "on the next frame" only in case the traversals are already scheduled or vsync is already in the queue. Otherwise it'll "run on current frame". 

Is that accurate?

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/host/TabsHostLayoutCoordinator.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Defers the layout to the NEXT frame via `Handler.post`. Because this puts the action at the end
+     * of the message queue, it runs after the current frame's `dispatchOnPreDraw`. Consequently, the
+     * forced layout mutations are NOT captured into the BottomNavigationView's `endValues` by the
+     * TransitionManager. This prevents the animator from producing a `ChangeBounds` and the content jump
+     * during the initial inset application.
+     * See [TabsHost.refreshLayout].
+     */
+    internal fun postLayout() {
```
kkafar: I'd name it in a bit more descriptive manner: `postLayoutToMessageQueue` and `postLayoutToChoreographer`. What do you think?

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/host/TabsHost.kt
```diff
@@ -118,27 +122,79 @@ class TabsHost(
...
+     * This also affects inner views like `navigation_bar_item_content_container` and `navigation_bar_item_labels_group`.
+     * - Effect: ChangeBounds animators are generated and the BottomNavigationView's content animated on tab selection.
+     * The active indicator layout with non-zero layout params is deferred via `Handler.post`
+     * in [NavigationBarItemView.onSizeChanged]. When it happens during the transition from another screen of
+     * the native-stack, `ChangeBounds` suppresses layout on animator creation and prevents recalculating the layout
+     * of the active indicator until the `AutoTransition` of `BottomNavigationView` ends.
+     *
+     * Once insets are applied, we switch permanently to `ReactChoreographer` path to ensure subsequent
+     * tab selection changes trigger the full `ChangeBounds` animations.
+     */
```
kkafar: I know I've suggested that we write an essay here, but looking at it - it is too much complexity. This whole explanation should be placed in the PR & PR comments. Here we should only put a summary and a link to the PR. It doesn't bring more value being in-place beside impacting readability of this file.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/host/TabsHost.kt
```diff
@@ -118,27 +122,79 @@ class TabsHost(
...
-            isLayoutEnqueued = true
-            // we use NATIVE_ANIMATED_MODULE choreographer queue because it allows us to catch the current
-            // looper loop instead of enqueueing the update in the next loop causing a one frame delay.
-            ReactChoreographer
-                .getInstance()
-                .postFrameCallback(
-                    ReactChoreographer.CallbackType.NATIVE_ANIMATED_MODULE,
-                    layoutCallback,
-                )
+        @Suppress("SENSELESS_COMPARISON") // layoutCoordinator can be null here since this method can be called in init
```
kkafar, line 137: Is this comment accurate? `layoutCoordinator` is not initialised inline at declaration site, likely can not be null here, can it? Can you verify it? If it can not -> then we'll have different timing for the first layout schedule. Can it matter?

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/host/TabsHost.kt
```diff
@@ -118,27 +122,79 @@ class TabsHost(
...
+     * `onPreDraw` is a reliable signal that insets have already been propagated for this frame:
+     * `ViewRootImpl.performTraversals()` runs in the Choreographer `CALLBACK_TRAVERSAL` phase, and within a
+     * single traversal it
+     * 1. computes and dispatches window insets down the hierarchy (`dispatchApplyWindowInsets`),
+     * 2. runs measure + layout,
+     * 3. invokes `ViewTreeObserver.dispatchOnPreDraw()` - immediately before `performDraw()`. So by the time this
+     * callback fires, the insets for the frame have been applied.
+     */
+    override fun onPreDraw(): Boolean {
+        hasFirstLayoutWithInsets = true
```
kkafar, line 155: We should reset this flag in `onDetachedFromWindow`? If we're removed from the view hierarchy, on reentrance the framework will treat us as not-laid-out anyway, right? we'll still most likely want to schedule layout with `post` on reenter.

kkafar (commented):
Leaving few code comments.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/host/TabsHost.kt
t0maboro: already verified and received null

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/host/TabsHost.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4161/changes/70a384d9f66fd3081305650cd02fd8df40e8d55c - moved later to the PR description

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/host/TabsHost.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4161/changes/3126df65cf61909ffc5ef62ae131317f23938c43 - moved later to the PR description

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/host/TabsHostLayoutCoordinator.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4161/changes/c18f2f22bb5d836fa17cddc3f9df8de77564281f

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/host/TabsHost.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4161/changes/6828c34f920283f14616ba79ffb450dbc7b6313d

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/host/TabsHost.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4161/changes/488e9535e810bf8322005950ab5fbb0b74bccd2d

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/host/TabsHost.kt
t0maboro: this is the 1st backtrace and it's `null` at this time

```
refreshLayout:138, TabsHost (com.swmansion.rnscreens.gamma.tabs.host)
requestLayout:161, TabsHost (com.swmansion.rnscreens.gamma.tabs.host)
setFlags:18702, View (android.view)
initViewGroup:724, ViewGroup (android.view)
<init>:717, ViewGroup (android.view)
<init>:711, ViewGroup (android.view)
<init>:707, ViewGroup (android.view)
<init>:703, ViewGroup (android.view)
<init>:85, FrameLayout (android.widget)
<init>:28, TabsHost (com.swmansion.rnscreens.gamma.tabs.host)
createViewInstance:29, TabsHostViewManager (com.swmansion.rnscreens.gamma.tabs.host)
createViewInstance:21, TabsHostViewManager (com.swmansion.rnscreens.gamma.tabs.host)
createViewInstance:224, ViewManager (com.facebook.react.uimanager)
createView:147, ViewManager (com.facebook.react.uimanager)
createViewUnsafe:565, SurfaceMountingManager (com.facebook.react.fabric.mounting)
preallocateView:1036, SurfaceMountingManager (com.facebook.react.fabric.mounting)
execute:39, PreAllocateViewMountItem (com.facebook.react.fabric.mounting.mountitems)
executeOrEnqueue:340, MountItemDispatcher (com.facebook.react.fabric.mounting)
dispatchPreMountItemsImpl:315, MountItemDispatcher (com.facebook.react.fabric.mounting)
dispatchPreMountItems:294, MountItemDispatcher (com.facebook.react.fabric.mounting)
doFrameGuarded:1527, FabricUIManager$DispatchUIFrameCallback (com.facebook.react.fabric)
doFrame:42, GuardedFrameCallback (com.facebook.react.uimanager)
lambda 'synchronized' in 'frameCallback':59, ReactChoreographer (com.facebook.react.modules.core)
frameCallback$lambda$1:49, ReactChoreographer (com.facebook.react.modules.core)
$r8$lambda$nSkFhrr5T7rop_XKwzlLov4NLLw:-1, ReactChoreographer (com.facebook.react.modules.core)
doFrame:0, ReactChoreographer$$ExternalSyntheticLambda0 (com.facebook.react.modules.core)
run:1568, Choreographer$CallbackRecord (android.view)
run:1579, Choreographer$CallbackRecord (android.view)
doCallbacks:1179, Choreographer (android.view)
doFrame:1104, Choreographer (android.view)
run:1553, Choreographer$FrameDisplayEventReceiver (android.view)
handleCallback:995, Handler (android.os)
dispatchMessage:103, Handler (android.os)
loopOnce:248, Looper (android.os)
loop:338, Looper (android.os)
main:9067, ActivityThread (android.app)
invoke:-1, Method (java.lang.reflect)
run:593, RuntimeInit$MethodAndArgsCaller (com.android.internal.os)
main:932, ZygoteInit (com.android.internal.os)
```

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/host/TabsHost.kt
kkafar: looks good

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/host/TabsHost.kt
kkafar: okay! thanks for checking this!

kkafar (commented):
Okay @t0maboro,
let's solve the code conflicts and land it ✅ 

Thank you

PR #4160, opened 2026-06-12
fix(Android): use per-display density for px→dp conversion in FabricEnabledViewGroup.updateState

ziponia:
## Description

Fabric mounts views using the density of the display the surface is attached to, but `FabricEnabledViewGroup.updateState` converts the natively measured **px** size back to **dp** with the process-global density (`PixelUtil`), captured once from the device's main display.

On a display whose density differs from the main one (Samsung DeX, freeform multi-window, external monitors), every `Screen` pushes `windowSize ÷ mainDisplayDensity` into the Shadow Tree — all stack/tab content collapses into a 1/density-sized top-left box, while plain RN views outside screens track the window correctly.

Closes #4159.

## Changes

- `FabricEnabledViewGroup.updateState` (Android, Fabric): convert px→dp using the view's own context density (`context.resources.displayMetrics.density`), with a defensive fallback to `PixelUtil.toDIPFromPixel` if the density is not positive.
- On single-display devices the context density equals the global density, so behavior there is unchanged.

## Before & after - visual documentation

| Before — content in a `1/density` box | After — content fills the window |
| --- | --- |
| before | after |

(The dark card is a diagnostic overlay that instruments `updateState` pushes. In the "after" screenshot the visible text rendering is correct because the app also applies a workaround for a *separate* react-native core bug — stale global density in text sp→px conversion — which is out of scope for this PR and reported to facebook/react-native.)

Measured `updateState` pushes on a Galaxy S25 Ultra (main display density **2.8125**) in DeX, window resized 6 times:

| window (dp) | pushed state before fix | pushed state after fix |
|---|---|---|
| 394 × 702 | 140 × 235 | 394 × 662 |
| 497 × 730 | 177 × 245 | — |
| 737 × 794 | 262 × 268 | — |
| 421 × 788 | 150 × 266 | 421 × 748 |
| 338 × 727 | — | 338 × 687 |
| 429 × 790 | — | 429 × 750 |

Before the fix every push equals `window ÷ 2.8125`; after the fix it equals the window content size at every tested window size.

## Test plan

- Minimal reproducer (RN 0.86, New Architecture, based on the official reproducer template): https://github.com/ziponia/rn-multidisplay-density-repro — shows live on-screen measurements (outside-stack vs inside-`ScreenStack` size and the `window ÷ content` ratio).
[...]

PR #4169, opened 2026-06-15
fix(Android): route all px→dp Shadow Tree state pushes through per-display density

### LINKED ISSUE #4159: [Android][Fabric] Screen content collapses to 1/density size on displays with a different density (Samsung DeX / freeform) — px→dp in updateState uses global density
### Description

On Android displays whose density differs from the device's **main** display (Samsung DeX, freeform multi-window, external monitors), all `ScreenStack` / `Tabs` content is squeezed into a small top-left box of exactly `windowSize ÷ mainDisplayDensity`, while plain RN views outside any Screen track the window correctly.

**Root cause**: [`FabricEnabledViewGroup.updateState`](https://github.com/software-mansion/react-native-screens/blob/main/android/src/fabric/java/com/swmansion/rnscreens/FabricEnabledViewGroup.kt#L31-L33) converts the natively measured **px** size back to **dp** with `PixelUtil.toDIPFromPixel` — a process-global density captured once from the main display — but Fabric mounted the view using the **per-display** density. The mismatched round-trip pushes a shrunken `frameWidth` / `frameHeight` into the Shadow Tree, so the whole JS subtree lays out at `1/density` scale. The same lines are present on latest `main`.

**Real-device evidence** (Galaxy S25 Ultra, main display density **2.8125**, instrumented `updateState`, DeX window resized 6 times):

| window (dp) | Screen state push (dp) | ratio |
|---|---|---|
| 394 × 702 | 140 × 235 | 2.81 |
| 497 × 730 | 177 × 245 | 2.81 |
| 737 × 794 | 262 × 268 | 2.81 |
| 304 × 452 | 108 × 146 | 2.81 |
| 421 × 788 | 150 × 266 | 2.81 |

Every push is exactly `window ÷ 2.8125`. Screenshots (before/after) are in the reproducer README.

**Verified fix (on device)**: convert using the view's own context density — `context.resources.displayMetrics.density` — which is identical to the global value on single-display devices (no behavior change) and correct on multi-display setups. After this change the pushed state equals the window size at every tested window size, and content fills the window. **A PR follows shortly.**

(Related Fabric size-sync reports, likely same family: #2803, #2933. Note: a *second*, independent core-RN bug — stale global density in text sp→px — also surfaces in this scenario; it is out of scope here and will be reported to facebook/react-native.)

### Steps to reproduce

1. Build the reproducer below on a device with Samsung DeX or freeform windowing.
2. Switch to DeX (or open the app in a freeform window on a display with a different density).
3. The app displays live measurements: `root onLayout (OUTSIDE stack)` ≈ window size, while `INSIDE ScreenStack` content ≈ `window ÷ phoneDensity`, with the ratio shown on screen.

https://github.com/ziponia/rn-multidisplay-density-repro

### Screens version

4.25.2 (bug confirmed present on `main` as well)

### React Native version

0.86.0 (reproducer) / 0.85.3 (production app where it was first hit)

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

Bare (reproducer) — also reproduced in an Expo SDK 56 production app

### Architecture

Fabric (New Architecture)

### Build type

Reproduced in both Debug and Release

### Device

Real device — Samsung Galaxy S25 Ultra (SM-S938N), One UI / Android 16, Samsung DeX

### Disclosure

This issue was authored by **Claude (an AI coding agent)** operating on behalf of and supervised by [@ziponia](https://github.com/ziponia), who hit this bug in a production app. All measurements were taken on @ziponia's physical device, and this report was reviewed and approved by them before submission.

kkafar:
## Description

Follow-up that **completes and refactors** the per-display px→dp conversion fix for Android Fabric. 🥞 **Stacked on #4160** — land that first, then this.

`Fabric` mounts views using the density of the display the surface is attached to, but the screens library converts the natively-measured **px** back to **dp** with the process-global density (`PixelUtil`), captured once from the device's main display. On a display whose density differs (Samsung DeX, freeform multi-window, external monitors), every state pushed to the Shadow Tree (and the header-height event) is mis-scaled. #4160 fixes the first call site (`FabricEnabledViewGroup.updateState`); this PR extracts that into a reusable helper and routes **every remaining px→dp state push** through it.

Closes #4159 (together with #4160).

## Changes

- **`utils/DimensionUtils.kt`**: add an `internal` per-display conversion helper — a context-free core `pxToDp(px, density)` plus a `View.pxToDp(px)` extension that reads the view's own display density fresh on every call (so it stays correct if a view moves between displays of differing density). Documents the `density > 0f` guard and its relationship to `DisplayMetrics.density`.
- Route the remaining Shadow Tree / JS state pushes through the helper:
  - `FabricEnabledViewGroup` (refactor of #4160's inline conversion), `FabricEnabledHeaderConfigViewGroup`, `FabricEnabledHeaderSubviewViewGroup` — header/frame state.
  - `safearea/SafeAreaView` — inset values.
  - `gamma/common/ShadowStateProxy` — gamma stack/header state. Has no `View`, so it takes a **required `density` parameter supplied per call** by its owning views (`StackScreen`, `StackHeaderConfig`, `StackHeaderSubview`).
  - `Screen.notifyHeaderHeightChange` — the header-height event dispatched to JS.

N/A — no visible change on single-display devices, where the per-display density equals the global one (the conversion is a no-op there). The fix only affects rendering on a display whose density differs from the device's main display; that path is validated on hardware in #4160 (Galaxy S25 Ultra in DeX).

## Test plan

- `./gradlew :app:assembleDebug` (FabricExample) — **BUILD SUCCESSFUL**; Kotlin compile + codegen pass with all changed files.
- Ran FabricExample on an emulator (Pixel 9 Pro, API 36) and verified basic navigation has no regression on the single-display (no-op) path: stack push/pop (Simple Native Stack → Detail → back) and bottom-tabs switching all render full-window with correct headers.
- Multi-display behaviour (the actual bug): covered by #4160's on-device validation and reproducer — https://github.com/ziponia/rn-multidisplay-density-repro.

## Checklist

- [x] Included code example that can be used to test this change (reproducer in #4160).
- [ ] For visual changes, included screenshots / GIFs / recordings documenting the change. — N/A
- [x] For API changes, updated relevant public types. — N/A, only `internal` helpers + an `internal` parameter.
- [ ] Ensured that CI passes — pending.

### DIFF
--- a/android/src/fabric/java/com/swmansion/rnscreens/FabricEnabledHeaderConfigViewGroup.kt
+++ b/android/src/fabric/java/com/swmansion/rnscreens/FabricEnabledHeaderConfigViewGroup.kt
@@ -5,8 +5,8 @@ import android.view.ViewGroup
 import androidx.annotation.UiThread
 import com.facebook.react.bridge.WritableMap
 import com.facebook.react.bridge.WritableNativeMap
-import com.facebook.react.uimanager.PixelUtil
 import com.facebook.react.uimanager.StateWrapper
+import com.swmansion.rnscreens.utils.pxToDp
 import kotlin.math.abs
 
 abstract class FabricEnabledHeaderConfigViewGroup(
@@ -39,10 +39,12 @@ abstract class FabricEnabledHeaderConfigViewGroup(
         paddingStart: Int,
         paddingEnd: Int,
     ) {
-        val realWidth: Float = PixelUtil.toDIPFromPixel(width.toFloat())
-        val realHeight: Float = PixelUtil.toDIPFromPixel(height.toFloat())
-        val realPaddingStart: Float = PixelUtil.toDIPFromPixel(paddingStart.toFloat())
-        val realPaddingEnd: Float = PixelUtil.toDIPFromPixel(paddingEnd.toFloat())
+        // Convert px->dp with this view's own display density (see pxToDp); the global density
+        // mis-scales the frame pushed to the Shadow Tree on a non-main-density display. See #4159.
+        val realWidth: Float = pxToDp(width.toFloat())
+        val realHeight: Float = pxToDp(height.toFloat())
+        val realPaddingStart: Float = pxToDp(paddingStart.toFloat())
+        val realPaddingEnd: Float = pxToDp(paddingEnd.toFloat())
 
         // Check incoming state values. If they're already the correct value, return early to prevent
         // infinite UpdateState/SetState loop.
--- a/android/src/fabric/java/com/swmansion/rnscreens/FabricEnabledHeaderSubviewViewGroup.kt
+++ b/android/src/fabric/java/com/swmansion/rnscreens/FabricEnabledHeaderSubviewViewGroup.kt
@@ -5,8 +5,8 @@ import android.view.ViewGroup
 import androidx.annotation.UiThread
 import com.facebook.react.bridge.WritableMap
 import com.facebook.react.bridge.WritableNativeMap
-import com.facebook.react.uimanager.PixelUtil
 import com.facebook.react.uimanager.StateWrapper
+import com.swmansion.rnscreens.utils.pxToDp
 import kotlin.math.abs
 
 abstract class FabricEnabledHeaderSubviewViewGroup(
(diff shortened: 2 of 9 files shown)


t0maboro (approved):
I noticed that `PixelUtil` is still being used in a few places, particularly in `ScreenDummyLayoutHelper`. Could this lead to any inconsistencies?

kkafar:
Yes, it potentially could. The `ScreenDummyLayoutHelper` is never attached to any window though, to handle it correctly we'd need to:

1. somehow get information of current display metrics in the shadow tree layer,
2. pass them via the JNI to the `ScreenDummyLayoutHelper` code,
3. extend the `ScreenDummyLayoutHelper` caching mechanism to integrate `displayDensity` as a part of cache key.

The worst what can happen here now is wrong layout (content height of stack) over a couple of frames. The use case is so niche, that I don't plan to look into it.

PR #4170, opened 2026-06-15
chore(test): e2e: test-tabs-item-badge 

LKuchno:
## Description

This PR adds a new test scenario screen and a Detox e2e test that exercise the `badgeValue` prop on `TabsScreen`, covering default badge appearance, a long numeric string, a custom string badge, and an emoji badge. The e2e test is iOS-only (wrapped in `describeIfiOS`) because badge text content is introspectable via `_UIBarBadgeView` on iOS but not on Android; the scenario screen itself is cross-platform and documents Android behavior as well. To enable Detox assertions, `tabBarItemTestID` is added to each of the four tab route configs, exposing the tab bar items via `by.id()`. The overall `e2eCoverage` is marked `incomplete` because badge color and text color attributes are not yet assertable without a screenshot-diff approach.

Closes:https://github.com/software-mansion/react-native-screens-labs/issues/1415

## Changes

- **FabricExample/e2e**: Added `test-tabs-item-badge.e2e.ts` — Detox iOS-only suite using `describeIfiOS` that asserts badge text on all four tabs via `_UIBarBadgeView` ancestor queries
- **apps/src**: Added `test-tabs-item-badge/` scenario directory with `index.tsx` (cross-platform screen covering four badge cases), `scenario-description.ts`, and `scenario.md` (manual steps for iOS and Android including known iOS 26 badge color regression KI-1072)
- **apps/src/tests/single-feature-tests/tabs/index.ts**: Registered `TestTabsItemBadge` in the scenario map
- **Tabs/testID**: Set `tabBarItemTestID` on all four tab route configs to expose tab bar items as Detox targets

### DIFF
--- a/FabricExample/e2e/single-feature-tests/tabs/test-tabs-item-badge.e2e.ts
+++ b/FabricExample/e2e/single-feature-tests/tabs/test-tabs-item-badge.e2e.ts
@@ -16,7 +16,6 @@ const tabBarBadgeViewType = isIOSVersionAtLeast('26.0')
   ? '_UIBarBadgeView'
   : '_UIBadgeView';
 
-const tabBarBagdeViewType = tabBarBadgeViewType;
 describeIfiOS('Tab Bar Item Badge', () => {
   beforeAll(async () => {
     await device.reloadReactNative();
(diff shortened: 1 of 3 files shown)


kkafar (approved):
Seems to work ok

LGTM ✅

PR #4173, opened 2026-06-16
chore(test): add @smoke tag to tabs e2e test suites

LKuchno:
## Description

This PR introduces `@smoke` tagging to Detox e2e test suites that have been identified as belonging to the smoke test subset for the Tabs component. The `@smoke` suffix is appended to `describe` block names, which allows a Detox test runner configuration to filter and execute only the smoke subset via a `--testNamePattern` or equivalent flag — without any changes to the test logic itself.

The two tagged test files cover critical Tabs behaviour: prevention of programmatic native tab selection, and rejection of stale navigation-state updates. Both already had `smokeTest: true` set in their corresponding `scenario-description.ts` files, so this change brings the Detox tags into alignment with that metadata.

## Changes

- **E2E / Tabs**: Added `@smoke` to the `describe` block in `test-tabs-prevent-native-selection.e2e.ts` and  to both `describe` blocks in `test-tabs-stale-update-rejection.e2e.ts`
- **package.json**: Added two scripts to run smoke test: `test-e2e-android-smoke` and `test-e2e-ios-smoke`

### DIFF
--- a/FabricExample/e2e/single-feature-tests/tabs/test-tabs-prevent-native-selection.e2e.ts
+++ b/FabricExample/e2e/single-feature-tests/tabs/test-tabs-prevent-native-selection.e2e.ts
@@ -1,7 +1,7 @@
 import { device, expect, element, by } from 'detox';
 import { selectSingleFeatureTestsScreen } from '../../e2e-utils';
 
-describe('Tabs: preventNativeSelection @smoke', () => {
+describe('@smoke Tabs: preventNativeSelection', () => {
   beforeAll(async () => {
     await device.reloadReactNative();
     await selectSingleFeatureTestsScreen(
--- a/FabricExample/e2e/single-feature-tests/tabs/test-tabs-prevent-native-selection.e2e.ts
+++ b/FabricExample/e2e/single-feature-tests/tabs/test-tabs-prevent-native-selection.e2e.ts
@@ -1,7 +1,7 @@
 import { device, expect, element, by } from 'detox';
 import { selectSingleFeatureTestsScreen } from '../../e2e-utils';
 
-describe('Tabs: preventNativeSelection', () => {
+describe('Tabs: preventNativeSelection @smoke', () => {
   beforeAll(async () => {
     await device.reloadReactNative();
     await selectSingleFeatureTestsScreen(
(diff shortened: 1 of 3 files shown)


FabricExample/e2e/single-feature-tests/tabs/test-tabs-prevent-native-selection.e2e.ts
```diff
@@ -1,7 +1,7 @@
 import { device, expect, element, by } from 'detox';
 import { selectSingleFeatureTestsScreen } from '../../e2e-utils';
 
-describe('Tabs: preventNativeSelection', () => {
+describe('Tabs: preventNativeSelection @smoke', () => {
```
kligarski: I thought that `@smoke` is some Detox/Jest annotation. If it's just in the name, maybe we can use something like `[SMOKE]` at the beginning of the name of the test? 

Having it at the start of the title might also help with quickly identifying that a *smoke* test has failed which is usually more important that some other tests.

But it's fully a question of what convention we want - we can keep the current `@smoke` or think of some other way.

kligarski (approved):

FabricExample/e2e/single-feature-tests/tabs/test-tabs-prevent-native-selection.e2e.ts
LKuchno: Having it at the beginning makes sense — we can settle on the name at meeting :)

I did a bit of research though, and [SMOKE] isn't a great choice: when used directly in the command it doesn't actually filter the way we'd expect.

**From my Claude and runtime research:**
@ smoke isn't a Detox/Jest annotation - it's just part of the test name. The filtering is done by Jest's --testNamePattern (-t), which Detox forwards to Jest. The important detail is that --testNamePattern is a regular expression, not a literal string match. That's what trips up [SMOKE]:

@ smoke - every character is a literal in regex, so it matches the substring @ smoke in the test name. Works as expected.
[SMOKE] - in regex, [...] is a character class, so [SMOKE] means "any single one of the characters S, M, O, K, or E." It ends up matching almost every test we have (basically any name containing one of those letters), so it doesn't filter to smoke tests at all.

So I'd lean toward a regex-safe prefix instead — something like SMOKE: or keeping @smoke but moving it to the start of the title. Happy to discuss on Thursday.

kkafar (approved):
Tested locally - works all right.

PR #4180, opened 2026-06-18
refactor(test): align tabs test structure with rules from rfc

LKuchno:
## Description

This PR cleans up the **tabs single-feature test suite** so directory names, e2e
file names, and `scenario-description` metadata follow the project's test
organization conventions. It is a test-infrastructure-only change — no library
source or runtime behavior is touched.

## Changes

- **Renamed** the `tabs-screen-orientation` test directory to
  `test-tabs-screen-orientation` (both `index.tsx` and `scenario-description.ts`)
  to match the `test-*` naming convention, and updated its registration `key`
  accordingly.
- **Renamed** the e2e spec `test-tabs-bottom-accessory-ios.e2e.ts` →
  `test-tabs-bottom-accessory-layout-ios.e2e.ts` so the e2e file mirrors its
  test screen directory.
- **Reordered** the scenario imports/registrations in the tabs `index.ts` for a
  consistent, logical order.
- **Added missing `details`** metadata to the `scenario-description.ts` files for
  `test-tabs-tab-bar-hidden` and
  `test-tabs-tab-bar-experimental-user-interface-style-ios`.
- **Fixed the `key`** in the `test-tabs-more-navigation-controller-ios`
  scenario-description to carry the missing `-ios` suffix.
- **Updated `scenario.md`** metadata (e2e test section reference) for
  `test-tabs-layout-appearances-ios` to match the new e2e file naming.

New Tabs screen order:
Simulator Screenshot - iPhone 17 Pro - 2026-06-18 at 10 41 58


apps/src/tests/single-feature-tests/tabs/test-tabs-screen-orientation/scenario-description.ts
```diff
@@ -2,8 +2,8 @@ import type { ScenarioDescription } from '@apps/tests/shared/helpers';
 
 export const scenarioDescription: ScenarioDescription = {
   name: 'Tabs Screen Orientation',
-  key: 'tabs-screen-orientation',
-  platforms: ['ios', 'android'],
+  key: 'test-tabs-screen-orientation',
+  platforms: ['ios'],
```
kkafar: Is it iOS specific? I don't think it is. Maybe it's yet unimplemented for Android, however the test is intended for both platforms, for Android it just fails.

LKuchno: Yes, it's currently unimplemented on Android tracked here: [#868](https://github.com/software-mansion/react-native-screens-labs/issues/868).

I plan to add the 'android' tag to [#1208](https://github.com/software-mansion/react-native-screens-labs/issues/1208) once it's actually working. There are also separate iOS and iPad-specific issues that will require scenario updates. For now, I'm preparing a basic scenario to verify 3 out of the 10 possible options (according to the docs).

I feel it might be misleading to place the Android icon next to a screen that doesn't work on that platform yet. However, I can add it if preferred, trusting that the note about the current iOS-only implementation in the scenario description will be enough context.

LKuchno: https://github.com/software-mansion/react-native-screens/pull/4180/commits/418bcaee9b6cb45bdb68b451690f26858a7a5591

PR #4189, opened 2026-06-18
feat(Android, FormSheet v5): Setup basic structure for standalone FormSheet

t0maboro:
## Description

This PR introduces the base implementation for the `FormSheet` component on Android that wraps the native Material `BottomSheetDialog`.

Because the native `BottomSheetDialog` operates in its own separate window, this implementation relies on a portal pattern. The `FormSheetHost` resides in the standard React Native view hierarchy and acts as the anchor, but it explicitly forwards its React children to a detached `FormSheetContentView` inside the dialog.

To ensure the React shadow tree is aware of the native dialog dimensions, we use a `ShadowStateProxy` to push the container's width and height to Yoga. Since Android currently lacks a direct API for synchronous shadow node updates (unlike iOS), we introduced the `FormSheetSyncFlushEvent`. This custom event utilizes `experimental_isSynchronous = true`.

This PR addressed the jumping content issue upon opening the sheet for the 1st time. While the event emission is realized synchronously, the actual View mounting isn't. If a synchronous flush is triggered during an Android's Choreographer traversal phase, the tick for the `EventBeat` is deferred to the next frame. To prevent this issue, we attached `FormSheetContentView` with React's `UIManagerListener` and native `ViewTreeObserver.OnPreDrawListener`. By returning false in `onPreDraw`, we explicitly block native drawing until `didDispatchMountItems` fires, ensuring Android waits for the correct layout before drawing the sheet.

---

Structure in the Host Tree

1. Main window
```
└── DecorView
    └── ...
        └── ReactSurfaceView
            └── ...
                └── FormSheetHost (Native Fabric component)
```

2. Modal window

```
└── DecorView (New window / modal context)
    └── ...
        └── design_bottom_sheet (Material Component)
            └── FormSheetContainer (Native FrameLayout)
                └── FormSheetContentView (ReactViewGroup)
                    └── [Teleported React Children]
```

---

Component Breakdown & Roles

1. FormSheetHost
- Native Fabric component acting as the initial anchor/parent for children in the React element tree.
- Has an empty layout implementation. Since the children are intercepted and teleported to another part of the hierarchy, `FormSheetHost` doesn't handle their physical layout on this screen.

2. FormSheetContainer
- Native Android FrameLayout serving as the structural wrapper for the sheet's content.
- Expands to take up all available space provided by the `design_bottom_sheet`.

3. FormSheetContentView
- ReactViewGroup that is a target destination for the teleported React children.
- Automatically expands to match the dimensions determined by its `FormSheetContainer`.
- Because it is a `ReactViewGroup`, it does not pass native Android layout down to its children. Instead, the layout of these children is delegated to the Yoga algorithm.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1547

---

Follow-ups:
The following items are intentionally out of scope for this PR and will be handled iteratively:
- The Host currently intercepts touch events from other views in the hierarchy, which can make elements behind it unclickable.
- Custom detents are not supported yet; the sheet falls back to the default detents provided by the native.
- Dedicated `FormSheetHostEventEmitter` will be implemented in a separate PR.
- Remaining props will be added successively.
- The default native dimming animation needs further research as it looks really glitchy.

## Changes

- Setup for native dialog wrapper (`FormSheetHost`, `FormSheetContainer`, `FormSheetContentView`).
- Added ViewManager with stubs for unimplemented props.
- Implemented content size synchronization using `ShadowStateProxy`.
- Implemented frame correction logic using `UIManagerListener` and `OnPreDrawListener` to prevent flickers.
- Added `FormSheetNativeDismissEvent` to handle the native swipe-to-dismiss gesture.

https://github.com/user-attachments/assets/99cc806e-07d4-4641-8afe-38e7d67dce61

## Test plan

This is not a final shape, but the initial implementation can be tested on the base example from SFTs.

## Checklist

- [ ] Ensured that CI passes


android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetHost.kt
```diff
@@ -0,0 +1,77 @@
...
+    internal fun setIsOpen(open: Boolean) {
+        if (this.isOpen == open) return
+        this.isOpen = open
+
+        if (isOpen) {
+            dialogManager.show()
+        } else {
+            dialogManager.dismiss()
+        }
+    }
```
kkafar, line 35: I think this is the source of the problem we've talked about in direct message.

You attempt here to show the dialog at the moment you receive the props for the first time. 

1. You might not have the children yet (I don't remember exactly the order of operations here).
2. You might not have the layout yet. 

To solve this you could: send the dialog manager signal that you want to open the modal, it'll save that information (boolean). When you are first attached to window AND you have the layout (I believe you will receive frame before being attached to window, but not sure) you give that information to the dialog manager. When both conditions are satisfied the dialogManager is able to perform the action.

t0maboro: https://github.com/software-mansion/react-native-screens/pull/4189/commits/693bc479fec9ccfad11420ee42f2935a6c06be07

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
```diff
@@ -0,0 +1,34 @@
...
+    internal val contentView get() = container.contentView
+
+    // Eagerly create the dialog and attach the container
+    private val dialog =
+        BottomSheetDialog(context).apply {
+            setContentView(container)
+            setOnDismissListener {
+                host.onNativeDismiss()
+            }
+        }
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4189/changes/08ea760c3eb88dca976c6e940cebe6897c874af4

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetHost.kt
```diff
@@ -0,0 +1,101 @@
...
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4189/changes/8010685d0f5a8e763f55333b24f4b5736fb5e81e

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetHostViewManager.kt
```diff
@@ -0,0 +1,135 @@
...
+    RNSFormSheetHostManagerInterface<FormSheetHost> {
+    private val delegate: ViewManagerDelegate<FormSheetHost> =
+        RNSFormSheetHostManagerDelegate<FormSheetHost, FormSheetHostViewManager>(this)
+
+    override fun getDelegate(): ViewManagerDelegate<FormSheetHost> = delegate
+
+    override fun getName(): String = REACT_CLASS
+
+    override fun createViewInstance(reactContext: ThemedReactContext) = FormSheetHost(reactContext)
+
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4189/changes/a0a94aa7dd05cd987569f0d74735f2c596e56a6f

common/cpp/react/renderer/components/rnscreens/RNSFormSheetHostState.h
```diff
@@ -12,10 +17,24 @@ class JSI_EXPORT RNSFormSheetHostState final {
...
+#ifdef ANDROID
+  RNSFormSheetHostState(
+      RNSFormSheetHostState const &previousState,
+      folly::dynamic const &data)
+      : frameSize(
+            Size{
+                (Float)data["frameWidth"].getDouble(),
+                (Float)data["frameHeight"].getDouble()}),
+        contentOffset(Point{0, 0}) {}
+
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4189/changes/5f8e7aa32999021cfd71ba58b03a45074bf0f7ad

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetNativeDismissEvent.kt
kligarski, line 1: Should we use `override fun canCoalesce(): Boolean = false` for both events?

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
```diff
@@ -0,0 +1,34 @@
+package com.swmansion.rnscreens.gamma.modals.formsheet
+
+import android.content.Context
+import com.google.android.material.bottomsheet.BottomSheetDialog
+
+class FormSheetDialogManager(
+    context: Context,
+    private val host: FormSheetHost,
```
kligarski: We have a cycle here but I guess it's not a problem for JVM.

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetContentView.kt
```diff
@@ -0,0 +1,76 @@
...
+import com.facebook.react.views.view.ReactViewGroup
+import com.swmansion.rnscreens.gamma.helpers.getFabricUIManagerNotNull
+
+@OptIn(UnstableReactNativeAPI::class)
+@SuppressLint("ViewConstructor")
+class FormSheetContentView(
+    context: Context,
+    private val onSizeChangedCallback: (width: Int, height: Int) -> Unit,
+) : ReactViewGroup(context),
+    UIManagerListener,
```
kligarski, line 20: I guess that we'll try to extract react-related part in some follow-up PRs, right?

kligarski (approved):

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetContentView.kt
t0maboro: Yes, the current goal is to have only some "working" component. I'm not investing much time in domain separation, because I'm still not sure whether this is the final concept. Firstly, I prefer to focus on issues that may need to adjust the approach for this component implementation.

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetNativeDismissEvent.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4189/commits/3411663920c474d418cc0ce3d54d7e1b36057fdc

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
t0maboro: GC would handle that, but as we're on an early stage, it's good to clean up when it's not causing additional problems, thanks for catching that! https://github.com/software-mansion/react-native-screens/pull/4189/changes/5ad8573e5e451d78c85c6ea2a77fdc602c5742d8

PR #4171, opened 2026-06-16
feat(iOS, ContainedModal): add ContainedModal and ContainedModalProvider related files on JS side

sgaczol:
## Description

This PR adds `ContainedModal` and `ContainedModalProvider` related files on JS side.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1358

## Changes

- added codegen files
- added component and types - related files
- update `package.json`

## Before & after - visual documentation

N/A

## Test plan

N/A

## Checklist

- [ ] Included code example that can be used to test this change.
- [ ] For visual changes, included screenshots / GIFs / recordings documenting the change.
- [ ] For API changes, updated relevant public types.
- [ ] Ensured that CI passes

PR #3947, opened 2026-04-28
feat(iOS, FormSheet v5): Add basic setup for standalone FormSheet native component

t0maboro:
## Description

Introducing the iOS implementation of the standalone FormSheet component.

**Key Architectural Decisions:**

- On the JS side, the FormSheet is placed exactly where it is defined in the component tree, but its frame is intentionally cleared. It acts as a logical host component.
- The presentation and hierarchy are split into a clear 3-tier architecture to maintain Separation of Concerns:
  - RNSFormSheetHostComponentView handles the Fabric lifecycle, prop updates, state management and attaching children.
  - RNSFormSheetHostController is responsible for the native `UIModalPresentationFormSheet` presentation.
  - RNSFormSheetContentView is a dedicated clean UIKit view assigned to the controller to safely manage the physical existence of React subviews.
- The actual UI is driven by a dedicated `RNSFormSheetHostController`. This controller creates a transparent top-level UIView and is presented modally over the current UIWindow hierarchy via the closest valid `presentingViewController`.
- The sheet's visibility is entirely driven by the `isOpen` prop. This declarative boolean is translated into imperative `presentViewController` and `dismissViewController` calls. When a user natively dismisses the sheet (e.g., via a swipe-down gesture), the `RNSFormSheetHostControllerDelegate` triggers the `RNSFormSheetHostComponentEventEmitter` to fire the `onNativeDismiss` event, allowing the JS to synchronize its local state.
- To ensure React children are sized correctly within the dynamic sheet, we observe `viewDidLayoutSubviews` inside the presented controller. When the native bounds change, the new size and origin offset are captured and dispatched via a C++ state update. This explicitly forces Yoga to **synchronously** recalculate the content layout using the sheet's actual native dimensions.
- Child components defined in JS are intercepted during Fabric's mount/unmount lifecycle. Instead of being added to the FormSheet's Host view, they are explicitly reparented into the presented controller's `RNSFormSheetContentView`.

**Implementation details:**

- Fully supports iOS 16+ custom detents, with a safe fallback to system detents for iOS 15. Detents are strictly validated: an empty array gracefully defaults to a large detent, and provided values **MUST be in strictly ascending order**.
- Implements an independent `RCTSurfaceTouchHandler` attached to the controller's view, utilizing `viewOriginOffset` to correctly align the React touch coordinate space with the natively presented window.
- Delays the zeroing of the Shadow Node size until the dismissal animation completes to prevent the layout from collapsing while sliding down.

> [!CAUTION]
[...]

PR #4190, opened 2026-06-18
feat(iOS, ContainedModal): Add basic setup for ContainedModal native component

sgaczol:
## Things to do

- add a prop regarding animation
- add an event emitter and events handling
- add a test regarding nesting providers

## Description

Introducing the iOS implementation of the standalone `ContainedModal` component, together with its `ContainedModalProvider`.

A `ContainedModal` is presented **within the bounds of a provider** instead of over the whole window. The two are matched by id: the modal's `targetContainerId` is compared against a provider's `containerId`, and the modal is presented inside the matching provider's frame. This makes it possible to scope a modal to an arbitrary region of the screen (e.g. one column of a split layout) rather than the full screen.

Key architectural decisions:

- On the JS side, the `ContainedModal` host is a **zero-sized, absolutely-positioned logical anchor** — it never participates in layout of its siblings. The actual React children are "teleported" and mounted inside a separate `UIViewController` hierarchy (`RNSContainedModalContentController` / `RNSContainedModalContentView`), so the host's `hitTest:` always returns `nil` and never intercepts touches meant for the content behind it.
- The `ContainedModalProvider` is, by contrast, a regular space-filling container. Its bounds define the presentation area, and it hosts the `UIViewController` the modal is presented from (resolved once, lazily, by walking up to the provider whose `containerId` matches).
- Presentation is **state-driven**: flipping `isOpen` from `false → true` presents the modal, and `true → false` dismisses it. The present/dismiss lifecycle and its state machine live in `RNSContainedModalPresentationManager`, kept separate from the controller and the component view.
- Presentation style is selected from JS via the `transparent` prop: `true` (the default) uses `UIModalPresentationOverCurrentContext`, so the provider's content stays visible behind the presented modal, while `false` uses `UIModalPresentationCurrentContext`, replacing it. The transition style will eventually be exposed as a prop too, but for now it is hardcoded to `UIModalTransitionStyleCrossDissolve`.

This mirrors the architecture established for the standalone `FormSheet` component (#3947), reusing the same host-anchor + teleported-content + presentation-manager split.

> [!NOTE]
> `ContainedModal` is currently an **iOS-only** component. The Android and Web entry points are present but no-ops for now. It is exported from `react-native-screens/experimental`.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1359
Closes https://github.com/software-mansion/react-native-screens-labs/issues/1361
Closes https://github.com/software-mansion/react-native-screens-labs/issues/1364
Closes https://github.com/software-mansion/react-native-screens-labs/issues/1367
Closes https://github.com/software-mansion/react-native-screens-labs/issues/1356
Closes https://github.com/software-mansion/react-native-screens-labs/issues/1362
Closes https://github.com/software-mansion/react-native-screens-labs/issues/1369

## Changes

- **JS API & exports**
  - Added `ContainedModal` (`isOpen`, `targetContainerId`) and `ContainedModalProvider` (`containerId`, `style`) components, types, and platform splits (`.tsx`, `.android.tsx`, `.web.tsx`).
  - Exposed both components via `src/experimental`.
- **Codegen**
  - Added Fabric native-component specs: `ContainedModalHostNativeComponent` and `ContainedModalProviderNativeComponent`.
- **C++ / shared layer**
  - Added the host shadow node, state, and component descriptor (`RNSContainedModalHostShadowNode`, `RNSContainedModalHostState`, `RNSContainedModalHostComponentDescriptor`).
- **iOS native implementation** (`ios/gamma/modals/contained-modal/`)
  - Host & provider component views, content view & content controller, providers protocol.
  - Presentation manager (present/dismiss state machine), update coordinator & flags, configuration applicator, shadow-state proxy.
  - Wired the gamma stubs (`RNSGammaStubs`).
- **Touch handling**
  - The content view's `RCTSurfaceTouchHandler` is attached lazily on present/layout and **detached on `viewDidDisappear:`** (via the controller→host delegate), in addition to `invalidate`. This prevents a dismissed modal from leaving behind a dead touch zone that silently swallows taps on the content behind it, and it re-attaches correctly on re-present.
- **Build**
  - Minor `package.json` wiring for the new sources.

> [!CAUTION]
> Dynamic content-origin updates aren't supported in the context of synchronization with the ShadowTree state. If an ancestor's offset changes, the frame of our host view might not be updated at all — from the host's perspective the frame doesn't change (`syncShadowNodeState` only schedules a new state update when the computed origin or content bounds actually change), so no content re-layout is triggered.

> [!NOTE]
> The JavaScript files unrelated to the single-feature test (a separate batch of `.js`/`.ts` changes) are tracked in their own PR — see https://github.com/software-mansion/react-native-screens/pull/4171.

https://github.com/user-attachments/assets/0cf09ec2-5166-4222-88e9-0ad5f3a85172

## Test plan

Run single feature tests regarding `ContainedModal` and verify if the modal's behaviors matches the expected ones in `scenario.md` files.

## Checklist

- [x] Included code example that can be used to test this change (single-feature test).
- [ ] Ensured that CI passes.


apps/src/tests/single-feature-tests/contained-modal/index.ts
```diff
@@ -0,0 +1,14 @@
+import type { ScenarioGroup } from '@apps/tests/shared/helpers';
+import TestContainedModalBase from './test-contained-modal-base-ios';
+
+const scenarios = {
+  TestContainedModalBase,
+};
+
+const ContainedModalScenarioGroup: ScenarioGroup<keyof typeof scenarios> = {
+  name: 'ContainedModal',
+  details: 'Single feature tests for ContainedModals',
```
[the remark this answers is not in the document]
sgaczol: 'ContainedModals' makes more sense, especially when we add a different style (CurrentContext) in the future

apps/src/tests/single-feature-tests/contained-modal/test-contained-modal-base-ios/scenario.md
```diff
@@ -0,0 +1,119 @@
...
+- **"Outside count" must always be tappable** - before, while, and after a modal is presented. It sits outside the provider, so the modal must never block it.
+- **"Inside count" must be blocked _only while the modal is open_.** While the modal is presented it sits underneath the (transparent) modal content and must not receive taps. As soon as the modal is dismissed it must be tappable again.
+
+## Steps
+
+### Baseline
+
+1. Launch the app and navigate to the **Basic functionality** screen.
+
+- [ ] Expected: Content is shown with the "Provider: full screen (tap for partial)" button, an "Open contained modal" button, an "Outside count: 0" button, and (inside the provider region) an "Inside count: 0" button.
```
LKuchno, line 40: I would set a background color for the ContainedModalProvider (e.g., backgroundColor: Colors.NavyLightTransparent) to make the provider region visible. Right now, "(inside the provider region)" isn't visually testable.

apps/src/tests/single-feature-tests/contained-modal/test-contained-modal-base-ios/scenario.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Background pressable works again after dismiss (touch-handler cleanup)
+
+> Regression check: after dismissal, the modal's touch handler must be detached.
+> If it is not, it keeps consuming touches in the area the modal used to occupy and
+> silently swallows taps meant for the "Inside count" button behind it, even though
+> the modal is no longer visible.
+
+6. Note the current "Inside count" value, then tap the "Inside count" button (the one
+   that sits in the region the modal was just covering) several times. Tap "Outside
```
LKuchno: I think there is not need to explain what "Inside count" button is.

```suggestion
6. Note the current "Inside count" value, then tap the "Inside count" button several times. Tap "Outside
```

apps/src/tests/single-feature-tests/contained-modal/test-contained-modal-base-ios/scenario.md
```diff
@@ -0,0 +1,119 @@
...
+
+- [ ] Expected: Same invariant as the full-screen case - "Inside count" does **not** increment while the modal is open, "Outside count" **does**. The smaller provider size must not change this behavior.
+
+---
+
+### Dismiss from partial provider
+
+10. Tap "Close".
+
+- [ ] Expected: The modal dismisses smoothly and the screen returns to the partial-provider layout. "Inside count" is tappable again and "Outside count" keeps working.
```
LKuchno: I would add separate step here to check tapping on buttons or add this action to step 10 description.

LKuchno (commented):
Looks good, but I left few comments how we can optimize this to make manual testing a bit more convenient.

ios/gamma/modals/contained-modal/RNSContainedModalConfigurationApplicator.mm
t0maboro, line 1: Some code cleanup is needed here. I wouldn't copy the whole FormSheet implementation unless the specific part is required right now.

sgaczol: done: https://github.com/software-mansion/react-native-screens/pull/4190/commits/dc57862556656afad5f88157d32489fd367d02a6

ios/stubs/RNSGammaStubs.h
```diff
@@ -27,6 +27,12 @@ NS_ASSUME_NONNULL_BEGIN
 @interface RNSFormSheetHostComponentView : NSObject
 @end
 
+@interface RNSContainedModalHostComponentView : NSObject
+@end
+
+@interface RNSContainedModalProviderHostComponentView : NSObject
+@end
```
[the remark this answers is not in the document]
sgaczol: https://github.com/software-mansion/react-native-screens/pull/4190/commits/ff3d9b99b11c4127bf538d3f79b7436f39450df3

apps/src/tests/single-feature-tests/contained-modal/test-contained-modal-base-ios/scenario.md
sgaczol: Okay: https://github.com/software-mansion/react-native-screens/pull/4190/commits/aa4a87013e1015febb8be556bfed92e5a025ca0c

apps/src/tests/single-feature-tests/contained-modal/test-contained-modal-base-ios/scenario.md
sgaczol: updated: https://github.com/software-mansion/react-native-screens/pull/4190/commits/aa4a87013e1015febb8be556bfed92e5a025ca0c

apps/src/tests/single-feature-tests/contained-modal/test-contained-modal-base-ios/scenario.md
sgaczol: done: https://github.com/software-mansion/react-native-screens/pull/4190/commits/aa4a87013e1015febb8be556bfed92e5a025ca0c

apps/src/tests/single-feature-tests/contained-modal/test-contained-modal-base-ios/scenario.md
LKuchno: Information about color should be provided also in scenario imo.

apps/src/tests/single-feature-tests/contained-modal/index.ts
```diff
@@ -0,0 +1,16 @@
+import type { ScenarioGroup } from '@apps/tests/shared/helpers';
+import TestContainedModalBase from './test-contained-modal-base-ios';
+import TestContainedModalPresentationStyle from './test-contained-modal-presentation-style-ios';
+
```
LKuchno, line 4: After PR #4201  we should also add this exports:

```suggestion

export { default as TestContainedModalBase } from './test-contained-modal-base-ios';
export { default as TestContainedModalPresentationStyle } from './test-contained-modal-presentation-style-ios';

```

apps/src/tests/single-feature-tests/contained-modal/test-contained-modal-presentation-style-ios/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+  ContainedModal,
+  ContainedModalProvider,
+} from 'react-native-screens/experimental';
+import { scenarioDescription } from './scenario-description';
+import { createScenario } from '@apps/tests/shared/helpers';
+import { Colors } from '@apps/shared/styling';
+
+const CONTAINER_ID = 'contained-modal-transparent';
+
+export function App() {
```
LKuchno, line 14: After PR#4201 we follow this function name convention: 

```suggestion
function TestContainedModalPresentationStyle() {
```

apps/src/tests/single-feature-tests/contained-modal/test-contained-modal-presentation-style-ios/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+  sheetSubtitle: {
+    fontSize: 16,
+    color: Colors.text,
+  },
+  spacing: {
+    height: 32,
+  },
+});
+
+export default createScenario(App, scenarioDescription);
```
LKuchno, line 131: After PR mentioned above:

```suggestion
export default createScenario(TestContainedModalPresentationStyle, scenarioDescription);
```

apps/src/tests/single-feature-tests/contained-modal/test-contained-modal-base-ios/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+  ContainedModal,
+  ContainedModalProvider,
+} from 'react-native-screens/experimental';
+import { scenarioDescription } from './scenario-description';
+import { createScenario } from '@apps/tests/shared/helpers';
+import { Colors } from '@apps/shared/styling';
+
+const CONTAINER_ID = 'contained-modal-base';
+
+export function App() {
```
LKuchno, line 14: After PR#4201 we should follow this function name convention:

```suggestion
function TestContainedModalBase() {
```

apps/src/tests/single-feature-tests/contained-modal/test-contained-modal-base-ios/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+    fontWeight: '600',
+    marginBottom: 12,
+    color: Colors.text,
+  },
+  spacing: {
+    height: 32,
+  },
+});
+
+export default createScenario(App, scenarioDescription);
```
LKuchno, line 143: Following above changes:

```suggestion
export default createScenario(TestContainedModalBase, scenarioDescription);
```

LKuchno (changes requested):
Few more remarks, mostly suggestion to align with structure after PR#4201

PR #4192, opened 2026-06-18
feat(Android, Tabs): integrate with ScrollViewMarker for special effects

kkafar:
## Description

Repeated selection of the already-active tab runs "special effects" — pop the
stack to root and/or scroll the content to the top. Until now the scroll view
backing the scroll-to-top effect was located purely by walking the first
descendant of the tab content, which is fragile and only ever matched
`android.widget.ScrollView`.

This PR integrates that lookup with the `ScrollViewMarker` mechanism: a marker
mounted around the content's scroll view registers it with the nearest seeking
ancestor, giving the tab a reliable handle to the correct scroll view instead
of relying solely on tree traversal. It also adds `NestedScrollView` support.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1538
Closes https://github.com/software-mansion/react-native-screens-labs/issues/1423

> [!note]
> The pop-to-root is not-yet-implemented due to modal constraints. It requires changes in the tabs model (simplifying here). 

## Changes

- `TabsScreen` now implements `ScrollViewSeeking`; it stores the scroll view
  registered by a `ScrollViewMarker` in a `WeakReference` and exposes it via
  `contentScrollView()`.
- `SpecialEffectsHandler` resolves the content scroll view from the registered
  marker first, falling back to the descendant-chain search. Both
  repeated-selection effects (pop-to-root and scroll-to-top) now anchor on the
  selected `TabsScreen` rather than the container content view.
- Scroll-to-top now handles both `ScrollView` and `NestedScrollView`.
- `ViewFinder.findScrollViewInFirstDescendantChain` widened to return
  `ViewGroup?` and match `NestedScrollView`, so the fallback path can resolve
  it too.

## Visual docs

| before | after |
| -- | -- | 
| <video src="https://github.com/user-attachments/assets/eb0a9e7a-f37c-4178-a15c-a777c63163c3" alt="android-tabs-special-effects-before" /> | <video src="https://github.com/user-attachments/assets/555b03b1-01b0-4864-ba79-69b35e051902" alt="android-tabs-special-effects-after" /> |

## Test plan

Tested on Android via the example app scenario:
`apps/src/tests/component-integration-tests/scroll-view-marker/test-stack-svm-tabs-special-effects/`

Steps:
1. Open the scenario, select a tab, scroll its content down and/or push screens
   onto the stack.
2. Tap the already-selected tab again.
3. Verify the stack pops to root and the content smoothly scrolls to the top,
   for both `ScrollView`- and `NestedScrollView`-backed content.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
@@ -6,13 +6,16 @@ import android.content.res.Configuration
 import android.view.Gravity
 import android.view.MenuItem
 import android.view.View
+import android.view.ViewGroup
 import android.view.WindowInsets
 import android.widget.FrameLayout
+import android.widget.ScrollView
 import androidx.appcompat.view.ContextThemeWrapper
 import androidx.core.graphics.Insets
 import androidx.core.view.WindowInsetsCompat
 import androidx.core.view.children
 import androidx.core.view.size
+import androidx.core.widget.NestedScrollView
 import androidx.fragment.app.FragmentManager
 import com.google.android.material.R
 import com.google.android.material.bottomnavigation.BottomNavigationView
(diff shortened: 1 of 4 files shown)


t0maboro (approved):

kligarski (approved):

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
```diff
@@ -58,23 +61,39 @@ class TabsContainer internal constructor(
...
+        private fun resolveContentScrollViewForScreen(tabsScreen: TabsScreen): ViewGroup? {
+            tabsScreen.contentScrollView()?.let { return it }
+            ViewFinder.findScrollViewInFirstDescendantChain(tabsScreen)?.let { return it }
+            return null
+        }
+
+        /**
+         * Attempts to scroll to top if the passed view is a `ScrollView` or `NestedScrollView`
+         */
+        private fun trySmoothScrollToTop(maybeScrollView: ViewGroup): Boolean {
```
kmichalikk, line 88: Instead of ViewGroup, you could use FrameLayout which is closer to the actual classes

Also, gemini came up with sth like that

```kotlin
interface ScrollableContainer {
    fun scrollToTop()
}

// Extension functions to easily wrap them
fun ScrollView.asScrollable() = object : ScrollableContainer {
    override fun scrollToTop() = smoothScrollTo(0, 0)
}

fun NestedScrollView.asScrollable() = object : ScrollableContainer {
    override fun scrollToTop() = smoothScrollTo(0, 0)
}
```

this is probably overkill, but you could make the typing more strict this way

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabsScreen.kt
```diff
@@ -139,6 +143,16 @@ class TabsScreen(
         tabsScreenDelegate.get()?.onFragmentConfigurationChange(this, config)
     }
 
+    // ScrollViewSeeking
```
kmichalikk: maybe region / endregion instead?

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
kkafar: Yeah, this is nice, but it's more code that what we already have & the current code is straightforward to read.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabsScreen.kt
kkafar: +1

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabsScreen.kt
```diff
@@ -18,16 +20,10 @@ import kotlin.properties.Delegates
...
-    override fun onLayout(
-        changed: Boolean,
-        l: Int,
-        t: Int,
-        r: Int,
-        b: Int,
-    ) = Unit
-
+    FragmentProviding,
+    ScrollViewSeeking {
```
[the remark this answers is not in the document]
kkafar: I admit this is a valid concern, HOWEVER it seems to be working & building just fine. Maybe we finally got support for such setup in more recent version of Kotlin? Dunno. Leaving it as it right now. Will patch once it proves necessary.

PR #4195, opened 2026-06-19
refactor(test, iOS): resolve Detox simulator target so device name and OS version stay in sync

LKuchno:
## Description

Detox iOS e2e runs previously resolved the simulator **device name** and **iOS version** independently, each reading its own env var, with nothing guaranteeing the resulting pair belonged to the same — or even an existing — simulator.

This PR unifies resolution into a single `SimulatorTarget` (`{ name, os }`) so both values always come from the same simulator. Either env var can now be supplied on its own — a lone model resolves to its latest installed OS, a lone version to the first installed device on that OS — and it adds simulator auto-detection (querying what is installed and booted) so local runs adapt to whatever is already available.

Discussion was updated adding subsection:`How the target simulator is chosen`
https://github.com/software-mansion/react-native-screens-labs/discussions/572

## Changelog

`scripts/e2e/ios-devices.js`:

- **Unified resolution** — introduces `resolveSimulatorTarget()`, returning a single `{ name, os }` so the device name and iOS version can never come from different simulators. `resolveAppleSimulatorName()` and `getIOSVersion()` now derive from it. The resolved target is memoized for the process lifetime.
- **Resolution order:**
  1. `RNS_APPLE_SIM_NAME` / `RNS_IOS_VERSION` env vars. Either may be given on its own:
     - both provided → the exact `name` + `os` instance is validated against the installed simulators. If the combination doesn't exist — e.g. `iPhone 15` with `iOS 26` (a runtime that model can't run) — the script **throws a clear error** listing the OS versions that model is actually installed with, instead of letting Detox/`applesimutils` fail later. (When `simctl` returns nothing — failed or no simulators installed — validation can't run, so the script **warns** and returns the pair verbatim, letting Detox surface any error.)
     - model only → its latest installed OS, else the default OS;
     - version only → the first installed device on that OS (preferring a booted one), else the default device.
  2. An already-booted simulator — when several are booted, the newest iOS version wins (preferring iPhone models on a tie), even over the default device.
  3. The default hardcoded device/version, when an instance of it is actually installed.
  4. The newest installed simulator (preferring iPhone models) when the default isn't installed. Falls back to the defaults when nothing is installed, letting Detox surface the error.
- **`simctl` probing** — `listInstalledSimulators()` queries `xcrun simctl list devices available --json` (memoized) and maps CoreSimulator runtime ids (`...SimRuntime.iOS-26-2`) to human versions (`iOS 26.2`). Only created, available instances are returned, because Detox matches existing simulators through `applesimutils` and does **not** create them from a device-type + runtime pairing. Degrades gracefully to defaults (with a warning) if `xcrun` is unavailable.
- **Config guard** — the `simctl` probe only runs for iOS simulator configurations. The active Detox config is read from the `--configuration` flag and the `DETOX_CONFIGURATION` env var, with an argv-scan fallback for indirect invocations.
- **Validation & types** — env-var formats are validated (`iPhone|iPad …`, `iOS …`) and, when both are given, the device + OS combination is checked for existence before use; JSDoc uses the `` `iOS ${string}` `` template-literal type so callers can't drop the prefix.

## Test:
Check if script follows the order of simulator and requirements mentioned in comment:

- [x] Run script with user input
      -   provide iPhone model and os version - different than default   
      -  provide iPhone model and os version not supported on that model
      -  iPhone model provided w/o version - take the latest OS version if feasible
      -  iOS provided w/o iPhone model - select something with appropriate SDK from the top of the list - first match.
 - [x] Run simulator other than default one and run script without args when default simulator is installed
 - [x] Run simulator other than default one and run script without args when default simulator is booted -> booted device with the newest version is taken
 - [x] Make sure default simulator is installed and no simulators are booted, run script without args -> default device is run
 - [x]  Make sure default simulator is NOT installed and no simulator are booted and run script without args -> iPhone with newest version is run

### DIFF
--- a/scripts/e2e/ios-devices.js
+++ b/scripts/e2e/ios-devices.js
@@ -1,46 +1,212 @@
+const { getCommandLineResponse } = require('./command-line-helpers');
+
 const DEFAULT_APPLE_SIMULATOR_NAME = 'iPhone 17';
 const DEFAULT_IOS_VERSION = 'iOS 26.2';
+
+const envVarKeys = /** @type {const} */ ({
+  simName: 'RNS_APPLE_SIM_NAME',
+  iosVersion: 'RNS_IOS_VERSION',
+});
+
+/**
+ * @return {string | undefined} validated simulator name from env, or undefined
+ */
+function readSimulatorNameEnv() {
+  const passedDevice = process.env[envVarKeys.simName];
+  if (!passedDevice) {
+    return undefined;
+  }
+  if (/^(iPhone|iPad)\s.+/.test(passedDevice)) {
+    return passedDevice;
+  }
+  throw new Error(
+    `Environment variable ${envVarKeys.simName} should be "iPhone xyz" or "iPad xyz".`,
+  );
+}
+
+/**
+ * @return {`iOS ${string}` | undefined} validated iOS version from env, or undefined
+ */
+function readIOSVersionEnv() {
+  const passedVersion = process.env[envVarKeys.iosVersion];
+  if (!passedVersion) {
+    return undefined;
+  }
+  if (/^iOS\s\S+/.test(passedVersion)) {
+    return /** @type {`iOS ${string}`} */ (passedVersion);
+  }
+  throw new Error(
+    `Environment variable ${envVarKeys.iosVersion} should be "iOS xyz".`,
+  );
+}
+
+/**
+ * Maps a CoreSimulator runtime identifier to a human iOS version string.
+ * "com.apple.CoreSimulator.SimRuntime.iOS-26-2" -> "iOS 26.2"
+ * @param {string} runtimeId
+ * @return {`iOS ${string}` | undefined}
+ */
+function runtimeIdToIOSVersion(runtimeId) {
+  const match = /SimRuntime\.iOS-(\d+(?:-\d+)*)$/.exec(runtimeId);
+  if (!match) {
+    return undefined;
+  }
+  return /** @type {`iOS ${string}`} */ (`iOS ${match[1].replace(/-/g, '.')}`);
+}
+
+/**
+ * @typedef {{ name: string, os: `iOS ${string}` }} SimulatorTarget
+ */
+
+/**
+ * Queries `simctl` for currently booted, available iOS simulators.
+ * @return {SimulatorTarget[]}
+ */
+function listBootedAppleSimulators() {
+  let parsed;
+  try {
+    parsed = JSON.parse(
+      getCommandLineResponse('xcrun simctl list devices booted --json'),
+    );
+  } catch (error) {
+    console.warn(
+      `Could not query booted iOS simulators, falling back to defaults. Cause:\n${error}`,
+    );
+    return [];
+  }
+  const devicesByRuntime = parsed.devices ?? {};
+  /** @type {SimulatorTarget[]} */
+  const booted = [];
+  for (const [runtimeId, devices] of Object.entries(devicesByRuntime)) {
+    const os = runtimeIdToIOSVersion(runtimeId);
+    if (!os) {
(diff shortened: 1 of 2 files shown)


kkafar:
The desired order of simulator / emulator resolving (descending priority):

1. User input,
2. Already running instance of simulator / emulator,
3. default hardcoded settings,
4. first from the list of installed simulators / emulators.

If user specifies only a iPhone model w/o version - take the latest OS version if feasible. In case that is problematic to implement then we can fallback to some default OS version, but let's try first. 

If user specifies only a SDK version (iOS / Android version) w/o device model then just select something with appropriate SDK from the top of the list - first match.

scripts/e2e/ios-devices.js
```diff
@@ -1,46 +1,372 @@
...
+ */
+function readIOSVersionEnv() {
+  const passedVersion = process.env[envVarKeys.iosVersion]?.trim();
+  if (!passedVersion) {
+    return undefined;
+  }
+  // Accept only "iOS <digits>(.<digits>)*" so typos fail here, not downstream.
+  if (/^iOS\s\d+(\.\d+)*$/.test(passedVersion)) {
+    return /** @type {`iOS ${string}`} */ (passedVersion);
+  }
```
kkafar: Can we actually drop that `iOS` prefix? It doesn't bring any value and is annoying to type. 

What do you think?

scripts/e2e/ios-devices.js
```diff
@@ -1,46 +1,372 @@
...
+  const fromEnv = process.env.DETOX_CONFIGURATION;
+  if (fromEnv) {
+    return fromEnv;
+  }
+  const flagIndex = process.argv.indexOf('--configuration');
+  if (flagIndex !== -1 && flagIndex + 1 < process.argv.length) {
+    return process.argv[flagIndex + 1];
+  }
+  return undefined;
+}
```
kkafar, line 67: What are the accepted values here? Can we document them or point the reader to the place where they are defined (our config file I assume).

scripts/e2e/ios-devices.js
```diff
@@ -1,46 +1,372 @@
...
+
+/**
+ * Queries `simctl` once (memoized) for created, available simulator instances.
+ *
+ * Only existing instances are returned: Detox finds devices via `applesimutils`,
+ * which matches existing simulators and never creates new ones from a device
+ * type + runtime pairing.
+ * @return {InstalledSimulator[]}
+ */
+function listInstalledSimulators() {
```
kkafar, line 112: Do I read this correctly - we'll create a simulator from a device / SDK version pair?

kkafar (commented):
Seems mostly fine. I have just a few questions.

scripts/e2e/ios-devices.js
LKuchno: No 😅 the opposite.
It is written here that we never create new simulators. The only simctl call here is simctl list devices available, which just enumerates the simulators that already exist. The device-name/SDK-version pair is matched against that existing list (and validated against it - we throw "(..) is not installed" if it's missing).

scripts/e2e/ios-devices.js
LKuchno: https://github.com/software-mansion/react-native-screens/pull/4195/changes/0cdf19b68482e17ec0f3b37064b52ad28386f74a

scripts/e2e/ios-devices.js
LKuchno: https://github.com/software-mansion/react-native-screens/pull/4195/changes/0cdf19b68482e17ec0f3b37064b52ad28386f74a

kkafar (approved):
Oki doki, I've skimmed through the changes - looks ok ✅

PR #4196, opened 2026-06-19
fix(iOS, TabsBottomAccessory): use hidden instead of opacity to switch bottom accessory content views

sgaczol:
## Description
 
This PR changes the way the `RNSTabsBottomAccessory` content-view switching workaround hides/shows its two content views: instead of mutating `layer.opacity`, it now toggles `view.hidden`.

Such approach fixes a bug where, during the bottom accessory's appearance animation, both content views (regular & inline) showed through each other instead of just the one matching the current environment. We don't fully understand why `opacity` failed here (the values were already set before the animation started, yet both views still bled through) - most likely `hidden` simply doesn't participate in the animation at all, which is what fixes it.

Switching to `hidden` introduces a new wrinkle: `view.hidden` is also written by React Native itself. Fabric's mounting layer sets `self.hidden = layoutMetrics.displayType == DisplayType::None` in `-[UIView(ComponentViewProtocol) updateLayoutMetrics:oldLayoutMetrics:]`, which runs on every `display` change from JS and on the first layout after a (re)mount (`forceUpdate`). So RN can clobber the visibility the workaround just set, without going through our code. To stay in control, the helper now observes `hidden` via KVO on both content views and re-applies the correct visibility whenever something external toggles it.

## Changes

- `RNSTabsBottomAccessoryHelper`: in `handleContentViewVisibilityForEnvironmentIfNeeded`, swap `layer.opacity` assignments for `hidden` (invisible → `hidden = YES`, visible → `hidden = NO`).  Also observe `hidden` via KVO on both content views and re-apply the correct visibility when something external toggles it (with a reentrancy guard for our own writes). The observed views are held strongly so the observer is always removed from a live instance.
- `RNSTabsBottomAccessoryContentComponentView`: remove the `traitCollectionDidChange:` and `finalizeUpdates:` overrides that only existed to re-assert opacity after `invalidateLayer` reset it. These are redundant once visibility no longer relies on `layer.opacity`.

### Before

https://github.com/user-attachments/assets/b429b655-42db-4da9-b373-0492c4b12a12

https://github.com/user-attachments/assets/90d7e594-e4b9-46a7-b77b-5f0ce2f1e2fa

### After

https://github.com/user-attachments/assets/e5e1baac-3b3d-4d4a-824c-20c1ce62e0f5

https://github.com/user-attachments/assets/3f7c120c-5f5e-4bcb-87a2-78caf9247269

## Test plan
Run the 3288 issue test. Toggle `shown` from `true` to `false` and back to `true`. Confirm the two content views (regular & inline) don't overlap / show through each other.

## Checklist

- [x] Ensured that CI passes


ios/tabs/bottom-accessory/RNSTabsBottomAccessoryHelper.mm
```diff
@@ -16,6 +17,11 @@ @implementation RNSTabsBottomAccessoryHelper {
   RNSTabsBottomAccessoryContentComponentView *__weak _regularContentView;
   RNSTabsBottomAccessoryContentComponentView *__weak _inlineContentView;
 
+  RNSTabsBottomAccessoryContentComponentView *_hiddenObservedRegularContentView;
+  RNSTabsBottomAccessoryContentComponentView *_hiddenObservedInlineContentView;
+
```
kligarski: If we need to retain strongly the reference, let's just use one reference per slot (`_regularContentView` and `_inlineContentView`, just drop `__weak` and new `_hiddenObserved*ContentView`).

Then we can do something like:
```obj-c
- (void)setContentView:(RNSTabsBottomAccessoryContentComponentView *)contentView
        forEnvironment:(RNSTabsBottomAccessoryEnvironment)environment
{
  switch (environment) {
    case RNSTabsBottomAccessoryEnvironmentRegular:
      [self replaceContentView:&_regularContentView with:contentView];
      break;

    case RNSTabsBottomAccessoryEnvironmentInline:
      [self replaceContentView:&_inlineContentView with:contentView];
      break;

    default:
      RCTLogError(@"[RNScreens] Unsupported TabsBottomAccessory environment");
      return;
  }

  [self handleContentViewVisibilityForEnvironmentIfNeeded];
}

- (void)replaceContentView:(RNSTabsBottomAccessoryContentComponentView *__strong *)slot
                      with:(nullable RNSTabsBottomAccessoryContentComponentView *)newView
{
  RNSTabsBottomAccessoryContentComponentView *old = *slot;
  if (old == newView) {
    return;
  }

  if (old != nil) {
    [old removeObserver:self
             forKeyPath:@"hidden"
                context:RNSTabsBottomAccessoryContentViewHiddenContext];
  }

  if (newView != nil) {
    [newView addObserver:self
              forKeyPath:@"hidden"
                 options:NSKeyValueObservingOptionNew
                 context:RNSTabsBottomAccessoryContentViewHiddenContext];
  }

  *slot = newView; // strong assignment — retains newView, releases old
}
```

plus clearing KVO in invalidate.

kligarski (changes requested):

ios/tabs/bottom-accessory/RNSTabsBottomAccessoryHelper.mm
sgaczol: done: https://github.com/software-mansion/react-native-screens/pull/4196/commits/202b2a13c4ff93128747f55fb9f76474ab83e7a4

ios/tabs/bottom-accessory/RNSTabsBottomAccessoryHelper.mm
```diff
@@ -69,16 +73,53 @@ - (void)handleContentViewVisibilityForEnvironmentIfNeeded
...
+- (void)unregisterForContentViewHiddenChanges:(RNSTabsBottomAccessoryContentComponentView *__strong *)observedView
+{
+  RNSTabsBottomAccessoryContentComponentView *currentlyObserved = *observedView;
+  if (currentlyObserved == nil) {
+    return;
+  }
+
+  [currentlyObserved removeObserver:self forKeyPath:@"hidden" context:RNSTabsBottomAccessoryContentViewHiddenContext];
+  *observedView = nil;
```
kligarski, line 102: nit: I wouldn't expect unregister to also clear the reference to be honest. Maybe we can remove clearing the reference from here - `replace...` will clear the reference with last `*slot = newView;`. We would need to restore clearing reference in `invalidate`.

What do you think?

kligarski (approved):

ios/tabs/bottom-accessory/RNSTabsBottomAccessoryHelper.mm
sgaczol: I kinda agree but at the same time I would keep the current state in order to be consistent with already existing `unregisterForAccessoryFrameChanges`. It is called in `didMoveToWindow` and `invalidate` so `_observedNativeWrapperView = nil;` is present so we avoid double unregister. Wdyt?

kligarski: Fair point. Let's keep it then.

kligarski:
CI failed on bottom accessory tests - we should have a look but maybe we just need to adjust the e2e after the changes

sgaczol:
> CI failed on bottom accessory tests - we should have a look but maybe we just need to adjust the e2e after the changes

We probably need some changes  `toBeVisible` -> `notToBeVisible`

LKuchno:
> > CI failed on bottom accessory tests - we should have a look but maybe we just need to adjust the e2e after the changes
> 
> We probably need some changes `toBeVisible` -> `notToBeVisible`

Actually we want bottomAccessory to be visible in this test - thats what we are checking here. 
As detox doesn't see bottomAccessory as visible the correct fix I found is to change function expectBottomAccessoryVisible to **check not the visibility** of bottomAccessory but **if it exist** - only selected bottomAccessory option appear in ViewHierarchy 


```diff --git a/FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-ios.e2e.ts b/FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-ios.e2e.ts
index 7e65c4665..625ebde19 100644
--- a/FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-ios.e2e.ts
+++ b/FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-ios.e2e.ts
@@ -12,8 +12,8 @@ const bottomAccessoryElement = (testID: string) =>
     by.id(testID).withAncestor(by.type('RNSTabsBottomAccessoryComponentView')),
   ).atIndex(0);
 
-async function expectBottomAccessoryVisible(testID: string) {
-  await expect(bottomAccessoryElement(testID)).toBeVisible();
+async function expectBottomAccessoryExist(testID: string) {
+  await expect(bottomAccessoryElement(testID)).toExist();
 }
 
 async function expectBottomAccessoryText(testID: string, text: string) {
@@ -73,32 +73,32 @@ describeIfiOS('Tabs bottomAccessory (iOS)', () => {
   });
 
   it('should show the Upper Left accessory variant on initial load', async () => {
-    await expectBottomAccessoryVisible('accessory-upper-left');
+    await expectBottomAccessoryExist('accessory-upper-left');
     await expectBottomAccessoryText('accessory-upper-left', 'Upper Left');
   });
 
   it('should update the accessory when Center variant card is tapped', async () => {
     await element(by.id('variant-center')).tap();
-    await expectBottomAccessoryVisible('accessory-center');
+    await expectBottomAccessoryExist('accessory-center');
     await expectBottomAccessoryText('accessory-center', 'Center');
   });
 
   it('should update the accessory when Lower Right variant card is tapped', async () => {
     await element(by.id('variant-lower-right')).tap();
-    await expectBottomAccessoryVisible('accessory-lower-right');
+    await expectBottomAccessoryExist('accessory-lower-right');
     await expectBottomAccessoryText('accessory-lower-right', 'Lower Right');
   });
 
   it('should update the accessory when Long variant card is tapped', async () => {
     await element(by.id('variant-long')).tap();
-    await expectBottomAccessoryVisible('accessory-long');
+    await expectBottomAccessoryExist('accessory-long');
   });
 
   it('should update the accessory when RGB variant card is tapped', async () => {
     await element(by.id('variant-rgb')).tap();
-    await expectBottomAccessoryVisible('rgb-strip-0');
-    await expectBottomAccessoryVisible('rgb-strip-1');
-    await expectBottomAccessoryVisible('rgb-strip-2');
+    await expectBottomAccessoryExist('rgb-strip-0');
+    await expectBottomAccessoryExist('rgb-strip-1');
+    await expectBottomAccessoryExist('rgb-strip-2');
   });
 
   // ---------------------------------------------------------------------------
@@ -107,17 +107,17 @@ describeIfiOS('Tabs bottomAccessory (iOS)', () => {
 
   it('should preserve the accessory when switching to the ScrollDown tab and back', async () => {
     await element(by.id('variant-center')).tap();
-    await expectBottomAccessoryVisible('accessory-center');
+    await expectBottomAccessoryExist('accessory-center');
     await expectBottomAccessoryText('accessory-center', 'Center');
 
     await forceTapByLabeliOS('scroll-down-tab-item-label');
     await expect(element(by.id('scroll-down-scrollview'))).toBeVisible();
-    await expectBottomAccessoryVisible('accessory-center');
+    await expectBottomAccessoryExist('accessory-center');
     await expectBottomAccessoryText('accessory-center', 'Center');
 
     await forceTapByLabeliOS('config-tab-item-label');
     await expect(element(by.id('config-scrollview'))).toBeVisible();
-    await expectBottomAccessoryVisible('accessory-center');
+    await expectBottomAccessoryExist('accessory-center');
     await expectBottomAccessoryText('accessory-center', 'Center');
   });
 
@@ -140,7 +140,7 @@ describeIfiOS('Tabs bottomAccessory (iOS)', () => {
       by.type('UITabBar'),
     ).getAttributes()) as IosElementAttributes;
 
-    await expectBottomAccessoryVisible('accessory-center');
+    await expectBottomAccessoryExist('accessory-center');
     await expectBottomAccessoryText('accessory-center', 'Center');
     expectBottomAccessoryExtended(extendedBottomAccessory, extendedTabBar);
   });
@@ -168,7 +168,7 @@ describeIfiOS('Tabs bottomAccessory (iOS)', () => {
       .atIndex(0)
       .getAttributes()) as IosElementAttributes;
 
-    await expectBottomAccessoryVisible('accessory-center');
+    await expectBottomAccessoryExist('accessory-center');
     await expectBottomAccessoryText('accessory-center', 'Center');
     expectBottomAccessoryInline(
       inlineBottomAccessory,
@@ -211,7 +211,7 @@ describeIfiOS('Tabs bottomAccessory (iOS)', () => {
       by.type('UITabBar'),
     ).getAttributes()) as IosElementAttributes;
 
-    await expectBottomAccessoryVisible('accessory-center');
+    await expectBottomAccessoryExist('accessory-center');
     await expectBottomAccessoryText('accessory-center', 'Center');
     expectBottomAccessoryExtended(extendedBottomAccessory, extendedTabBar);
   });
@@ -239,7 +239,7 @@ describeIfiOS('Tabs bottomAccessory (iOS)', () => {
       .atIndex(0)
       .getAttributes()) as IosElementAttributes;
 
-    await expectBottomAccessoryVisible('accessory-center');
+    await expectBottomAccessoryExist('accessory-center');
     await expectBottomAccessoryText('accessory-center', 'Center');
     expectBottomAccessoryInline(
       inlineBottomAccessory,
```

PR #4202, opened 2026-06-23
fix(Android, Stack, Tabs): fix crash on simple navigation with stack nested in tabs

kkafar:
## Description

This PR touches only the Android side. iOS is yet untested.

Closes <https://github.com/software-mansion/react-native-screens-labs/issues/1543>

> [!caution]
> This PR requires careful testing! It touches core `TabsContainer` update logic.

The error mechanism is straightforward. When we tap back on the First
tab, the `TabsContainer` removes the `TabsScreenFragment` associated
with the `Stack` tab from *FragmentManager* and this forces the
*childFragmentManager* and its children (including
`StackScreenFragment`) to be destroyed. `StackScreenFragment` emits
`onDismiss(isNative=true)` from its `onDestroy` lifecycle callback,
which is obviously triggered -> this forces an undesired state update in
JS leading to an empty Stack & later crash.

The desired behavior here is to have the nested stack (nested container
in general) state preserved between tab switches. One way to achieve
this is to actually not remove the `TabsScreenFragment` from
*FragmentManager* entirely as we do now. Instead, we can just detach it
and re-attach it later. Such change shouldn't be that much intrusive,
since we keep all the tab fragments allocated anyway, so having them
attached to the *FragmentManager* should be painful. When we do so, tab
change no longer forces destruction of fragments in
*childFragmentManager* - effectively preserving the state of the nested
stack on the native side.

Alternative here is to keep the current behavior, but:

1. Somehow prevent changes of state in native stack (state should be
   left unmodified),
2. Create fragment manager state reconstruction mechanism that will
   re-apply the state preserved in navigation state back to the
   *FragmentManager*.

For now I decide to go with attach/detach on tabs fragment. This change
however requires careful testing.

> [!important]
> This patch fixes the problem only in this particular scenario. 
> Consider a stack nested in yet another stack - if we decide to unmount the screen
> with nested stack inside - the nested stack will still lose state.
> Similarly, in any case where the fragment hierarchy is being restored (e.g. after backgrounding the app?) this won't work. We still are in need of proper state restoration mechanism. This is however outside of scope of this PR. 

## Changes

`TabsContainer` now does not add/remove a single fragment per tab change. Now it initially adds all the fragments as "detached", and then only attaches & detaches appropriate fragments. 

I've also updated the code responsible for state restoration after window re-attachment added in: https://github.com/software-mansion/react-native-screens/pull/4035.
I've tested the `Test4027` it continues to work correctly. 

| before | after | 
| --- | --- |
| <video src="https://github.com/user-attachments/assets/b84e57cd-a9e6-451a-a26b-4e6ad36b750c" alt="before" /> | <video src="https://github.com/user-attachments/assets/6ae71cd5-4f89-4b2b-9691-3f03123d5dc5" alt="after" /> |

## Test plan

See `test-stack-tabs-stack-in-tabs-base-navigation` and scenario description there.

`Test4027` for regression. 

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
@@ -540,7 +540,7 @@ class TabsContainer internal constructor(
         if (navState.isEmpty()) {
             check(isInExternalOperationContext && pendingStateUpdateRequest != null)
             initNavigationState(nextSelectedFragment.requireScreenKey)
-            applyInitialStateToFragmentManager(nextSelectedFragment)
+            applyInitialStateToFragmentManagerSync(nextSelectedFragment)
             return true
         }
 
@@ -552,11 +552,11 @@ class TabsContainer internal constructor(
         }
 
         progressNavigationState(nextSelectedFragment.requireScreenKey, actionOrigin)
-        applyNextSelectedFragmentToFragmentManager(currentSelectedFragment, nextSelectedFragment)
+        applyNextSelectedFragmentToFragmentManagerSync(currentSelectedFragment, nextSelectedFragment)
         return true
     }
 
-    private fun applyInitialStateToFragmentManager(nextSelectedFragment: TabsScreenFragment) {
+    private fun applyInitialStateToFragmentManagerSync(nextSelectedFragment: TabsScreenFragment) {
         requireFragmentManager
             .createTransactionWithReordering()
             .let {
@@ -568,7 +568,7 @@ class TabsContainer internal constructor(
             }.commitNowAllowingStateLoss()
     }
 
-    private fun applyNextSelectedFragmentToFragmentManager(
+    private fun applyNextSelectedFragmentToFragmentManagerSync(
         currSelectedFragment: TabsScreenFragment,
         nextSelectedFragment: TabsScreenFragment,
     ) {
@@ -665,13 +665,15 @@ class TabsContainer internal constructor(
                 .filter { it in tabsModel }
                 .toList()
 
-        if (currentFragments.size == 1 && currentFragments[0] === selectedTab) {
+        val fragmentsWithAttachedUI = currentFragments.filterNot { it.isDetached }
+
+        if (currentFragments.size == tabsModel.size &&
+            fragmentsWithAttachedUI.size == 1 &&
+            fragmentsWithAttachedUI.first() === selectedTab
+        ) {
             return
         } else if (currentFragments.isEmpty()) {
-            requireFragmentManager
-                .createTransactionWithReordering()
-                .add(contentView.id, selectedTab)
-                .commitNowAllowingStateLoss()
+            applyInitialStateToFragmentManagerSync(selectedTab)
         } else {
             error("[RNScreens] Unexpected fragment manager state.")
         }
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
@@ -533,17 +533,14 @@ class TabsContainer internal constructor(
         }
     }
 
-    private fun updateSelectedFragment(
+    private fun updateNavigationStateAndSelectedFragment(
         nextSelectedFragment: TabsScreenFragment,
         actionOrigin: TabsActionOrigin,
     ): Boolean {
         if (navState.isEmpty()) {
             check(isInExternalOperationContext && pendingStateUpdateRequest != null)
-            navState = TabsNavigationState(nextSelectedFragment.requireScreenKey, 0)
-            requireFragmentManager
-                .createTransactionWithReordering()
-                .add(contentView.id, nextSelectedFragment)
-                .commitNowAllowingStateLoss()
+            initNavigationState(nextSelectedFragment.requireScreenKey)
+            applyInitialStateToFragmentManager(nextSelectedFragment)
             return true
         }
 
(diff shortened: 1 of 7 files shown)


LKuchno (approved):
I went through all defined test for Tabs domain.
No issues found but wait until Android e2e test finishes. 

Checked on Pixel 9 with Android API Level 36.0

apps/src/tests/component-integration-tests/tabs-stack-v5/test-stack-tabs-stack-in-tabs-base-navigation/scenario.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Prerequisites
+
+- iOS or Android device / simulator.
+
+## Steps
+
+### Baseline
+
+1. Launch the app and navigate to the test screen.
```
LKuchno, line 21:
```suggestion
1. Launch the app and navigate to the **Stack in Tabs - basic navigation scenarios** screen.
```

apps/src/tests/component-integration-tests/tabs-stack-v5/test-stack-tabs-stack-in-tabs-base-navigation/scenario.md
```diff
@@ -0,0 +1,56 @@
+# Test Scenario: Stack in Tabs - basic navigation scenarios
+
+## Details
+
+**Description:** Test common navigation flows in Stack in Tabs configuration
+
+**OS test creation version:** Android SDK 36, iOS 26.5
```
LKuchno, line 7: Small change to unify it with other scenarios.

```suggestion
**OS test creation version:** iOS: 26.5, Android: API Level 36.
```

apps/src/tests/component-integration-tests/tabs-stack-v5/test-stack-tabs-stack-in-tabs-base-navigation/scenario.md
```diff
@@ -0,0 +1,56 @@
...
+
+2. Push *Second* screen.
+
+3. Push *Third* screen.
+
+- [ ] Both *Second* and *Third* are pushed onto the stack.
+
+4. Toggle between *First* and *Stack* tabs.
+
+- [ ] *Stack* tab displays nested stack with *Third* route on top.
```
LKuchno, line 56: It looks like the numbering here might be broken unintentionally.

```suggestion
5. Navigate to the *Stack* tab.

6. Push *Second* screen.

7. Push *Third* screen.

- [ ] Both *Second* and *Third* are pushed onto the stack.

8. Toggle between *First* and *Stack* tabs.

- [ ] *Stack* tab displays nested stack with *Third* route on top.
```

kligarski (approved):
I haven't checked the runtime but the code looks good.

apps/src/tests/component-integration-tests/tabs-stack-v5/test-stack-tabs-stack-in-tabs-base-navigation/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+  },
+];
+
+export function TestStackTabsStackInTabsBaseNavigation() {
+  return <App />;
+}
+
+function App() {
+  return <TabsNavigation />;
+}
```
LKuchno, line 71: It looks like we have some component redundancy happening here. TestStackTabsStackInTabsBaseNavigation renders <App/>, which just renders <TabsNavigation/> without adding any additional logic, state, or wrappers. We can flatten this to keep the component tree clean:

```suggestion
export function TestStackTabsStackInTabsBaseNavigation() {
  return <TabsNavigation />;
}
```

apps/src/tests/component-integration-tests/tabs-stack-v5/index.ts
```diff
@@ -1,7 +1,11 @@
 import type { ScenarioGroup } from '@apps/tests/shared/helpers';
 import TabsInStackStableEnterTransition from './test-stack-tabs-tabs-in-stack-stable-enter-transition';
+import { default as TestStackTabsStackInTabsBaseNavigation } from './test-stack-tabs-stack-in-tabs-base-navigation';
```
LKuchno, line 3: Is there a reason TestStackTabsStackInTabsBaseNavigation is being imported differently than TabsInStackStableEnterTransition here? Just wondering if the alias is necessary or if we can make them consistent.

LKuchno (commented):
Test looks good, thank you!
I just left a couple of minor comments regarding component redundancy, import consistency for some quick cleanup and scenario format.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
```diff
@@ -555,14 +552,37 @@ class TabsContainer internal constructor(
...
+    private fun applyInitialStateToFragmentManager(nextSelectedFragment: TabsScreenFragment) {
         requireFragmentManager
             .createTransactionWithReordering()
             .let {
-                it.remove(currentSelectedFragment)
-                it.add(contentView.id, nextSelectedFragment)
+                tabsModel.forEach { fragment ->
+                    it.add(contentView.id, fragment)
+                    it.detach(fragment)
+                }
```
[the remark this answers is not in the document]
kkafar: No. This function is for applying the *initial* state. If the fragment is already attached before an attempt to call this function - it is indeed an unexpected state and we can throw there.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt
```diff
@@ -555,14 +552,37 @@ class TabsContainer internal constructor(
...
+    private fun applyNextSelectedFragmentToFragmentManager(
+        currSelectedFragment: TabsScreenFragment,
+        nextSelectedFragment: TabsScreenFragment,
+    ) {
+        requireFragmentManager
+            .createTransactionWithReordering()
+            .let {
+                it.detach(currSelectedFragment)
+                it.attach(nextSelectedFragment)
+            }.commitNowAllowingStateLoss()
```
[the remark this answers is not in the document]
kkafar: We currently DO NOT support dynamic tab updates, but this is an asked for feature - we might want to support. Here's ticket for it: https://github.com/software-mansion/react-native-screens-labs/issues/1071

kkafar:
Excuse me @LKuchno, I rushed a bit. I'll respond to the review of yours n a follow-up.

apps/src/tests/component-integration-tests/tabs-stack-v5/index.ts
kkafar: Yes there is. This module also exposes exactly the same named object - the top level component. Default object is the scenario. This does not happen in the above.

kkafar:
Done here: https://github.com/software-mansion/react-native-screens/pull/4205

PR #4209, opened 2026-06-24
feat(tabs): custom drawable/symbol tab-bar icons — tinted, per-tab size, indicator sizing, iOS fallback

stachbial:
## Description

Extends the native tab bar for custom, app-provided icons.

**Android** (Tabs v5):

- `tinted?: boolean` (default `true`) on `drawableResource` and `imageSource` icons — `false` keeps the icon's own colors (wraps it in `NoTintDrawable` so the bar's tint can't repaint it). Settable per state (normal / selected).
- `drawableIconSize` (dp) — per-tab icon size. Material has one bar-wide `itemIconSize`, so the box becomes the max across tabs and each icon is inset to its own size; unset uses the Material default (resolved from library resources, not hardcoded).
- `tabBarItemActiveIndicatorWidth` / `tabBarItemActiveIndicatorHeight` (dp) on the appearance — bar-wide indicator size, from the focused tab; unset auto-scales to wrap an enlarged icon, falling back to the Material resource defaults.
- Icon-size changes arriving while the container is detached from window are dropped (same guard as #4271) — the full appearance pass re-applies them on reattach, preventing the #4258-style crash.

**iOS**:

- Custom SF Symbol fallback on the tab bar: `systemImageNamed:` → `imageNamed:`, so a `sfSymbol` icon resolves an app asset-catalog symbol when it isn't a built-in.
- `tinted?: boolean` on `imageSource` icons — `tinted: true` renders the image as a template (follows the item icon color). Supersedes `templateSource`, which is now deprecated in favor of `imageSource + tinted: true`.
  Note: `icon` and `selectedIcon` must resolve to the same native icon type, so their `tinted` values must match (single `iconType` in the native spec).

## Before & after

**Per-tab icon size — `drawableIconSize`.** The wide logo renders at its intended size; the system star stays at the default.

| Before | After |
| --- | --- |
|default 24dp box|per-tab drawableIconSize|

**Active-indicator size.** Reshape the pill to wrap the enlarged icon; shared across tabs.

| Before | After |
| --- | --- |
|default indicator|reshaped indicator|

**Tint — `tinted: false`.** The focused tab keeps the icon's own colors; unfocused falls back to the template tint.

https://github.com/user-attachments/assets/00b13fae-2da7-4a28-8b0a-043e95b3a343

**iOS showcase: `type: 'sfSymbol'` with custom icon name supplied via xcasset `.symbolset` and `.imageset` (vector image)**

https://github.com/user-attachments/assets/09b79ae8-f0b7-43a1-b8a3-c1d4f94fa5dd

## Test plan

- Render the Tabs v5 bar with custom icons on Android: `tinted: false` keeps a multicolor VectorDrawable's colors; `drawableIconSize` scales one tab while others stay at the default; the indicator wraps the enlarged icon; an `imageSource` icon tints by default and keeps its own colors with `tinted: false` ("Image Tint" tab).
- iOS: a custom asset-catalog symbol renders as a tab icon via `sfSymbol`; an `imageSource` icon with `tinted: true` follows the item tint (gray/blue) instead of its original colors ("Image" tab).
- Regression: `test-tabs-general-appearance-android` renders stock icon sizes and stock/recolored/disabled indicators correctly; a `Test4258`-style detach/remount cycle with an icon-size change while detached does not crash and re-applies the size on reattach.
- Built and ran on Android + iOS. Minimal repro: `CustomNativeTabsIcons.{ios,android}.tsx`.


stachbial:
Follow-up react-navigation PRs that consume this:
- https://github.com/react-navigation/react-navigation/pull/13167
- https://github.com/react-navigation/react-navigation/pull/13168

ios/tabs/RNSTabBarAppearanceCoordinator.mm
```diff
@@ -62,7 +62,9 @@ - (void)setIconsForTabBarItem:(UITabBarItem *)tabBarItem
   if (screenView.iconType == RNSTabsIconTypeSfSymbol || screenView.iconType == RNSTabsIconTypeXcasset) {
     if (screenView.iconResourceName != nil) {
       if (screenView.iconType == RNSTabsIconTypeSfSymbol) {
-        tabBarItem.image = [UIImage systemImageNamed:screenView.iconResourceName];
+        // Fall back to a custom symbol from the app's asset catalog.
+        tabBarItem.image = [UIImage systemImageNamed:screenView.iconResourceName]
+            ?: [UIImage imageNamed:screenView.iconResourceName];
```
kkafar: Let's extract it into a helper method on `RNSImageLoadingHelper`

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabsScreen.kt
```diff
@@ -180,5 +216,8 @@ class TabsScreen(
 
     companion object {
         const val TAG = "TabsScreen"
+
+        // Material's default bottom-navigation icon size (dp).
+        internal const val DEFAULT_ICON_SIZE_DP = 24f
```
kkafar: This is not a good approach. The value might change between material versions. I'm sure the dimension is encoded somewhere in Android resources - please resolve it from context. You should be able to find that resource id in material design documentation.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabsScreen.kt
```diff
@@ -79,15 +80,50 @@ class TabsScreen(
...
+        if (newValue != oldValue) rebuildSelectedIcon()
+    }
+
+    var selectedDrawableIconTinted: Boolean = true
+        set(value) {
+            if (field != value) {
+                field = value
+                rebuildSelectedIcon()
+            }
         }
```
kkafar: Notice that we use Delegates.observable instead in multiple other places. Please stick to the convention.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
```diff
@@ -22,6 +24,55 @@ import com.swmansion.rnscreens.utils.resolveColorAttr
...
+    // Explicit dp wins; else auto-scale to the enlarged icon box; else Material
+    // default. Reads iconBoxDp, so applyIconBox must run first.
+    private fun applyActiveIndicatorSize(
+        widthDp: Float?,
+        heightDp: Float?,
+    ) {
+        if (defaultIndicatorWidthPx < 0) {
+            defaultIndicatorWidthPx = bottomNavigationView.itemActiveIndicatorWidth
+            defaultIndicatorHeightPx = bottomNavigationView.itemActiveIndicatorHeight
+        }
```
kkafar: Have you considered whether these values are constant? Don't they change depending on context environment? What about interface orientation changes?

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
```diff
@@ -22,6 +24,55 @@ import com.swmansion.rnscreens.utils.resolveColorAttr
...
+                widthDp != null && widthDp > 0f -> PixelUtil.toPixelFromDIP(widthDp).toInt()
+                enlarged -> PixelUtil.toPixelFromDIP(iconBoxDp + 16f).toInt()
+                else -> defaultIndicatorWidthPx
+            }
+        bottomNavigationView.itemActiveIndicatorHeight =
+            when {
+                heightDp != null && heightDp > 0f -> PixelUtil.toPixelFromDIP(heightDp).toInt()
+                enlarged -> PixelUtil.toPixelFromDIP(iconBoxDp + 8f).toInt()
+                else -> defaultIndicatorHeightPx
+            }
```
kkafar: What are these `16` & `8` magic numbers here?

kkafar: Also this looks like a proper helper method candidate.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
kkafar: Also, how are these values determined? What if this method is called before the `bottomNavigationView` is laid out?

src/components/tabs/screen/TabsScreen.android.types.ts
```diff
@@ -99,6 +99,14 @@ export interface TabsScreenAppearanceAndroid {
...
    */
   tabBarItemActiveIndicatorEnabled?: boolean | undefined;
+  /**
+   * @summary Active-indicator pill size in dp. If unset, it auto-scales to wrap
+   * the icon box when icons are enlarged via `drawableIconSize`.
+   *
+   * @platform android
+   */
+  tabBarItemActiveIndicatorWidth?: number | undefined;
+  tabBarItemActiveIndicatorHeight?: number | undefined;
```
kkafar, line 115: Each prop needs it's own JSDoc for proper documentation generation.

src/fabric/tabs/TabsScreenAndroidNativeComponent.ts
```diff
@@ -102,6 +105,10 @@ export interface NativeProps extends ViewProps {
   imageIconResource?: ImageSource | undefined;
   selectedDrawableIconResourceName?: string | undefined;
   selectedImageIconResource?: ImageSource | undefined;
+  drawableIconTinted?: CT.WithDefault<boolean, true>;
+  selectedDrawableIconTinted?: CT.WithDefault<boolean, true>;
+  // Per-tab icon size (dp); 0/unset = default 24dp.
```
kkafar: Don't put `24` here. It's down to what Material Design decides and those values are not set in stone.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
```diff
@@ -22,6 +24,55 @@ import com.swmansion.rnscreens.utils.resolveColorAttr
...
+    private var iconBoxDp: Float = TabsScreen.DEFAULT_ICON_SIZE_DP
+
+    fun applyIconBox(boxDp: Float) {
+        iconBoxDp = boxDp
+        bottomNavigationView.itemIconSize = PixelUtil.toPixelFromDIP(boxDp).toInt()
+    }
+
+    // Material defaults, captured before we override them so unset bars restore exactly.
+    private var defaultIndicatorWidthPx: Int = -1
+    private var defaultIndicatorHeightPx: Int = -1
```
kkafar: Previously, beside few constants, we've deliberately made `TabsAppearanceApplicator` as stateless as possible. I don't like the idea of runtime sourced values here. 

I don't know yet whether this is necessary or not, I've spent to little time on digging, just pointing out our preference here.

kkafar (commented):
I like the idea of wider support for various icon-related features and overall I'm willing to proceed here. 

I have left few remarks & also Copilot did a review. Please answer it.

kkafar (commented):
I've got some additional concerns!

Please make sure that the bug described in https://github.com/software-mansion/react-native-screens/pull/4161 is not re-introduced. The PR added two tests that can detect the regression

`Test4161` and `test-stack-tabs-stack-in-tabs-stable-enter-transition`.

src/components/tabs/screen/TabsScreen.android.types.ts
```diff
@@ -186,4 +194,14 @@ export interface TabsScreenPropsAndroid {
    * @platform android
    */
   selectedIcon?: PlatformIconAndroid | undefined;
+  /**
+   * @summary Per-tab icon size in dp.
+   *
+   * The bottom bar's icon box is the largest `drawableIconSize` across all tabs;
+   * each tab's icon is inset to its own size within that box. Tabs without a value
+   * use the default 24dp.
```
kligarski:
```suggestion
   * use the system default.
```

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabsScreen.kt
```diff
@@ -79,15 +80,50 @@ class TabsScreen(
...
+
+    // Per-tab icon size in dp; 0 means the default.
+    var drawableIconSize: Float = 0f
+
+    val effectiveIconSizeDp: Float
+        get() = if (drawableIconSize > 0f) drawableIconSize else DEFAULT_ICON_SIZE_DP
```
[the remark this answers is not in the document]
kligarski: +1

Also, wouldn't it be better to use `ViewManager.onAfterUpdateTransaction` to parse icons there instead of doing `rebuildIcon` on every single prop related to icons?

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabsScreen.kt
kligarski: There is `@dimen/mtrl_navigation_bar_item_default_icon_size`.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
```diff
@@ -22,6 +24,55 @@ import com.swmansion.rnscreens.utils.resolveColorAttr
 internal class TabsAppearanceApplicator(
     private val bottomNavigationView: BottomNavigationView,
 ) {
+    // Largest effective per-tab icon size (dp); Material allows only one size for all items.
+    private var iconBoxDp: Float = TabsScreen.DEFAULT_ICON_SIZE_DP
+
+    fun applyIconBox(boxDp: Float) {
+        iconBoxDp = boxDp
+        bottomNavigationView.itemIconSize = PixelUtil.toPixelFromDIP(boxDp).toInt()
```
kligarski: Shouldn't we use the same approach as in https://github.com/software-mansion/react-native-screens/pull/4169 and https://github.com/software-mansion/react-native-screens/pull/4160 for unit conversions?

kligarski: Applies to all unit conversions in this PR.

ios/tabs/RNSTabBarAppearanceCoordinator.mm
kligarski: I'm not sure that I get why we want to fallback to regular assets for `sfsymbol` if we have `xcasset` type?

satya164: @kligarski i mentioned it on yesterday's sync, the main idea is: `type: 'sfSymbol'` can be used for both built-in symbols and custom symbols:

- it'll let us provide a consistent API, e.g. `<SFSymbol>` component in React Navigation, `type: 'sfSymbol'` in native tabs as well as custom navigators (which internally use `<SFSymbol>`), all can have a `name` property that accepts built-in and custom symbols
- by having custom symbols on `type: 'sfSymbol'`, any future options specific to symbols (e.g. effects) stay on a single type instead of duplicating into `type: 'xcasset'` when it won't work for just any asset

while custom symbols are still asset catalog, the precedent is there in screens: `type: 'imageSource'` and `type: 'templateSource'` both render images but have different types solely because the image is created with a different configuration, and `type: 'xcasset'` exposes assets that `type: 'imageSource'` and `type: 'templateSource'` already expose. in case of custom symbols, we should be able to pass symbol configuration as well in the future.

satya164: here's what we support in react navigation for sf symbols: https://reactnavigation.org/docs/8.x/icons/#sf-symbols

satya164: also we don't expose `type: 'xcasset'` in react navigation, as same can be done with `imageSource` and `templateSource` for images, and with more customization (i.e. you can't control rendering mode with `type: 'xcasset'`). for now it looks redundant.

kligarski: Makes sense.

For v5 we might need to rethink the API for icons though, e.g. regarding whether `xcasset` is necessary as you mentioned + having `imageSource` + `templateSource` split on iOS vs `drawableResource` type with `tinted` prop on Android.

satya164: iirc `drawableResource` is necessary as image loader doesn't expose everything from my discussion with @t0maboro . we'll also need to expose it in react navigation, but need to add some native code to make it work for other navigators as well

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabsScreen.kt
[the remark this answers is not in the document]
stachbial: Done, all icon resolution lives in TabsScreenIcon.resolveIfNeeded(), invoked once per transaction from TabsScreenViewManager.onAfterUpdateTransaction ✅

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
```diff
@@ -22,6 +24,55 @@ import com.swmansion.rnscreens.utils.resolveColorAttr
...
+    private fun sizeIcon(
+        icon: Drawable?,
+        effectiveDp: Float,
+    ): Drawable? {
+        if (icon == null || effectiveDp >= iconBoxDp) return icon
+        val larger = maxOf(icon.intrinsicWidth, icon.intrinsicHeight)
+        if (larger <= 0) return icon
+        val insetPx = (larger * (iconBoxDp - effectiveDp) / (2f * effectiveDp)).toInt()
+        return if (insetPx > 0) InsetDrawable(icon, insetPx) else icon
+    }
```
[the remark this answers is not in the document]
stachbial: The intrinsic-based inset is deliberate 🤨
NavigationBarItemView renders the icon in an ImageView sized to itemIconSize with `FIT_CENTER`, so the drawable is scaled to the box rather than drawn at intrinsic bounds. 
The inset ratio (`iconBoxDp − effectiveDp)/(2·effectiveDp`) is dimensionless — the intrinsic px factor cancels against that scaling, leaving the glyph at exactly effectiveDp/iconBoxDp of the box for any drawable. Computing the inset in absolute px would double-apply the scale. 
Added a comment in sizeIcon documenting this 🫡

ios/tabs/RNSTabBarAppearanceCoordinator.mm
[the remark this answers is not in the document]
stachbial: Answered in the following comment

ios/tabs/RNSTabBarAppearanceCoordinator.mm
```diff
@@ -77,7 +79,8 @@ - (void)setIconsForTabBarItem:(UITabBarItem *)tabBarItem
 
     if (screenView.selectedIconResourceName != nil) {
       if (screenView.iconType == RNSTabsIconTypeSfSymbol) {
-        tabBarItem.selectedImage = [UIImage systemImageNamed:screenView.selectedIconResourceName];
+        tabBarItem.selectedImage = [UIImage systemImageNamed:screenView.selectedIconResourceName]
+            ?: [UIImage imageNamed:screenView.selectedIconResourceName];
```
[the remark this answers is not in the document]
stachbial: ✅ Both call sites in RNSTabBarAppearanceCoordinator.mm now call [RNSImageLoadingHelper symbolImageNamed:...]; the helper uses an explicit if (image == nil) check.

ios/tabs/RNSTabBarAppearanceCoordinator.mm
stachbial: Good point, done ✅

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabsScreen.kt
stachbial: Thanks! Done ✅ I deleted `DEFAULT_ICON_SIZE_DP `; TabsAppearanceApplicator.defaultIconSizeDp resolves `R.dimen.mtrl_navigation_bar_item_default_icon_size`  fresh on every access, converted px→dp with the per-display-density helper 🫡

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
stachbial: fair point, switched to using the defaults which are immutable vals captured at applicator construction. They're theme attrs applied in the view constructor — available before any layout, no orientation variants, captured before anything can override them 🫡

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
stachbial: nice catch; switched to using theme-derived auto-scale padding and moved to `resolveIndicatorDimensionPx(explicitDp, autoScaledDp, defaultPx)`, so it tracks Material versions and app themes.

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabsScreen.kt
stachbial: ✅

src/components/tabs/screen/TabsScreen.android.types.ts
stachbial: ✅

src/fabric/tabs/TabsScreenAndroidNativeComponent.ts
stachbial: ✅

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
stachbial: `dpToPx(dp, density)` + `View.dpToPx` added to `DimensionUtils.kt` 🫡

android/src/main/java/com/swmansion/rnscreens/gamma/tabs/appearance/TabsAppearanceApplicator.kt
stachbial: ✅ stateless now — icon box comes in as a parameter, defaults resolve per call. The only vals left are the constructor-captured themed indicator sizes (we overwrite them on the view, so they can't be re-read later)

ios/tabs/RNSTabBarAppearanceCoordinator.mm
stachbial: keeping the fallback per the reasoning above for now; the lookup got extracted into `RNSImageLoadingHelper.symbolImageNamed:`  with an explicit nil check 🫡

stachbial: Thanks @satya164 !

stachbial:
Thanks a lot for the review and guidance @kkafar, @kligarski !

everything should be finally addressed now, in two commits:

[5dccf6c](https://github.com/software-mansion/react-native-screens/pull/4209/commits/5dccf6c1a054f2513f03b028198839c35c8dd0dc) has the direct review fixes: no more hardcoded Material values (default icon size comes from `mtrl_navigation_bar_item_default_icon_size`, indicator defaults are snapshotted from the themed view, auto-scale padding is derived from them), dp conversions go through per-display density helpers like in #4169 , the iOS symbol lookup got extracted to RNSImageLoadingHelper and the docs are per-prop now.

in [d5ba7ac](https://github.com/software-mansion/react-native-screens/pull/4209/commits/d5ba7ac77044b85ac2890e8f3f3296ec8bed265d) I moved the whole icon logic into a `TabsScreenIcon` class - one per normal/selected slot, staging props resolve once per transaction via `onAfterUpdateTransaction` (same pattern as `StackHeaderConfig`) and reuse the shared resolver. this also puts `imageSource` icons through the same path as drawables (they used to bypass staging and race the async load 🤨) and makes `TabsAppearanceApplicator` fully stateless - the icon box is computed by the coordinator and passed in.

merged main on top: files follow the new layout, resolver is `PropIconResolver` after the rename, tinted lives in the shared icon types now. checked on Material 1.14, renders the same.

Let me know, if you find anything else You'd like to be fixed guys! Cheers

src/components/shared/types.ts
```diff
@@ -49,6 +49,11 @@ export type PlatformIconShared = {
 export type PlatformIconAndroidDrawableResource = {
   type: 'drawableResource';
   name: string;
+  /**
+   * Whether the bottom tab bar tints the icon with the item icon color.
+   * Defaults to `true`; `false` keeps the drawable's own colors.
```
kligarski: This is a shared type used across different components, not only tabs - let's change the doc comment to be more general.

apps/src/screens/CustomNativeBottomTabsIcons.ios.tsx
```diff
@@ -0,0 +1,112 @@
...
+      title: 'System',
+      ios: {
+        icon: { type: 'sfSymbol', name: 'star' },
+        selectedIcon: { type: 'sfSymbol', name: 'star.fill' },
+      },
+    },
+  },
+];
+
+export default function CustomNativeBottomTabsIcons() {
```
kligarski: nit: we're trying to use `Tabs` name instead of `Bottom Tabs` because e.g. on iPad the tab bar appears at the top and can transform into a sidebar. It would be nice to keep this convention in this test as well (component, file name, entry in `Example.tsx`)

android/src/main/java/com/swmansion/rnscreens/tabs/container/TabsContainer.kt
```diff
@@ -420,6 +420,14 @@ class TabsContainer internal constructor(
         }
     }
 
+    override fun onIconSizeChange(tabsScreen: TabsScreen) {
+        // Icon box is bar-wide, so this must run even when the changed tab is not selected.
+        invalidationFlags.isNavigationMenuAppearanceInvalidated = true
+        post {
+            this.flushPendingUpdates()
+        }
```
kligarski: We might need to guard this with `isAttachedToWindow` - please see #4271

android/src/main/java/com/swmansion/rnscreens/tabs/appearance/TabsAppearanceApplicator.kt
```diff
@@ -16,12 +18,80 @@ import com.google.android.material.R
...
+        if (tabsScreen.drawableIconSize > 0f) tabsScreen.drawableIconSize else defaultIconSizeDp
+
+    // Material allows only one icon size for all items; iconBoxDp is the largest effective per-tab size.
+    fun applyIconBox(iconBoxDp: Float) {
+        bottomNavigationView.itemIconSize = bottomNavigationView.dpToPx(iconBoxDp).toInt()
+    }
+
+    // Theme-resolved defaults, captured at construction, before this class (the only writer) overrides them.
+    private val defaultIndicatorWidthPx: Int = bottomNavigationView.itemActiveIndicatorWidth
+    private val defaultIndicatorHeightPx: Int = bottomNavigationView.itemActiveIndicatorHeight
```
kligarski: Can't we resolve them similar to `defaultIconSizeDp`? I found these:

```xml
    <dimen name="m3_bottom_nav_item_active_indicator_height">32dp</dimen>
    <dimen name="m3_bottom_nav_item_active_indicator_width">64dp</dimen>
```

android/src/main/java/com/swmansion/rnscreens/tabs/appearance/TabsAppearanceApplicator.kt
```diff
@@ -16,12 +18,80 @@ import com.google.android.material.R
...
+    // Inset the icon so it renders at effectiveDp, centered within iconBoxDp.
+    // Intrinsic-relative on purpose: Material FIT_CENTER-scales the drawable to the icon box,
+    // so the intrinsic factor cancels and the glyph lands at effectiveDp/iconBoxDp of the box.
+    private fun sizeIcon(
+        icon: Drawable?,
+        effectiveDp: Float,
+        iconBoxDp: Float,
+    ): Drawable? {
+        if (icon == null || effectiveDp >= iconBoxDp) return icon
+        val larger = maxOf(icon.intrinsicWidth, icon.intrinsicHeight)
```
kligarski, line 92: Claude pointed out something like this:

<details>

```
1. Selected icon can be mis-sized when normal/selected differ in resolution (real, narrow). When a tab has both icons, targetIcon is a StateListDrawable. Its intrinsicWidth/Height (hence larger) resolves to the wildcard/normal child — the checked state isn't current at sizing time. But InsetDrawable takes a fixed insetPx. So on the checked state the glyph renders at larger_selected · P / (larger_selected + 2i), which only equals effectiveDp if larger_selected == larger_normal. Two icons of the same source usually match, but a resource icon + a differently-sized image-URI icon would make the selected icon slightly off.
```

</details>

I tested it quickly with:

```tsx
      android: {
        icon: {
          type: 'imageSource',
          imageSource: require('@assets/variableIcons/icon.png'),
        },
        selectedIcon: {
          type: 'imageSource',
          imageSource: require('@assets/trees.jpg'),
        },
      },
```

and it seems to be working correctly but it might be worth to have a deeper look.

src/components/tabs/screen/TabsScreen.android.types.ts
```diff
@@ -186,4 +200,14 @@ export interface TabsScreenPropsAndroid {
...
+  /**
+   * @summary Per-tab icon size in dp.
+   *
+   * The bottom bar's icon box is the largest `drawableIconSize` across all tabs;
+   * each tab's icon is inset to its own size within that box. Tabs without a value
+   * use the system default.
+   *
+   * @platform android
+   */
+  drawableIconSize?: number | undefined;
```
kligarski, line 214: I'm wondering about one thing - if material exposes `BottomNavigationView.itemIconSize` and it resizes all icons to that size (that's the intended native behavior), do we want to implement on `react-native-screens` side additional resizing logic for independent icon sizes?

@stachbial do you have any specific use cases for this prop being per-icon or why it's needed on Android? I guess that you don't need to cut out the padding manually from the icon and you can just change the size via prop but is there anything else I'm missing?

But on the other hand, if we're adding per-icon no-tinting behavior which is also not supported directly in BottomNavigationView API then I guess we can handle the sizing as well.

I wonder also about iOS, where icons AFAIK are not scaled in portrait mode at all but they are scaled automatically in landscape mode - if we're going to be able to support something similar there and if it would be necessary. I guess that for portrait, exposing width and height with the image source would be enough but not sure about landscape. But that's out of scope of this PR.

cc @kkafar

apps/src/screens/CustomNativeTabsIcons.ios.tsx
kligarski, line 1: Ideally we would add new/modify existing single feature tests for Tabs with new props but we can do this ourselves later in a separate PR.

android/src/main/java/com/swmansion/rnscreens/tabs/screen/TabsScreenIcon.kt
```diff
@@ -0,0 +1,69 @@
...
+        isInvalidated = false
+        resolver.resolve(context, drawableResourceName, imageUri) { result ->
+            when (result) {
+                IconResolution.Unchanged -> if (tinted != appliedTinted) emit()
+                is IconResolution.Resolved -> {
+                    rawDrawable = result.drawable
+                    emit()
+                }
+            }
+        }
```
kligarski, line 57: This should fix https://github.com/software-mansion/react-native-screens-labs/issues/975 as well but I guess it's okay to do it here even accidentally d:

src/components/shared/types.ts
```diff
@@ -49,6 +49,11 @@ export type PlatformIconShared = {
...
+   */
+  tinted?: boolean;
```
kligarski, line 62: This type is shared across components so I wonder if this prop should be named `tinted` or `tintable`. For now, in Android tabs implementation you can't remove the tint color via `tabBarItemIconColor` but e.g. in Stack v5, icons are not tinted by default. In Stack v5, icon is tinted only if you provide `iconTintColorNormal` so setting `tinted: true` will not make the icon tinted and this might surprise the user.

Another discussion is that this prop would also work with regular images on Android but exposing it might be tricky because on iOS we have separate type for this (`templateSource`).

I wonder if we should add this prop to shared `imageSource` type & deprecate `templateSource` on iOS in favor of `imageSource + tinted: true`. But in that case discussion `tinted` vs `tintable` is even more difficult because on iOS we use `UIImageRenderingModeAlwaysOriginal`/`UIImageRenderingModeAlwaysTemplate` which determines if original colors are used AFAIK.

cc @kkafar

android/src/main/java/com/swmansion/rnscreens/tabs/appearance/TabsAppearanceApplicator.kt
```diff
@@ -174,23 +251,28 @@ internal class TabsAppearanceApplicator(
...
             } else {
-                tabsScreen.icon
+                iconDrawable
             }
 
-        if (menuItem.icon != targetIcon) {
-            menuItem.icon = targetIcon
+        val sizedIcon = sizeIcon(targetIcon, effectiveIconSizeDp(tabsScreen), iconBoxDp)
+
+        if (menuItem.icon != sizedIcon) {
```
kligarski: This will usually be true now because in `sizeIcon` we're creating a new `InsetDrawable` so maybe we should remove it. Hopefully there are no icon flashes when only title is changed.

ios/helpers/image/RNSImageLoadingHelper.mm
```diff
@@ -59,6 +59,15 @@ + (void)loadImageFromSource:(nonnull RCTImageSource *)imageSource
       }];
 }
 
++ (nullable UIImage *)symbolImageNamed:(nonnull NSString *)name
+{
+  UIImage *image = [UIImage systemImageNamed:name];
+  if (image == nil) {
+    image = [UIImage imageNamed:name];
+  }
```
kligarski, line 67: Should we document falling back to `imageNamed` if no `systemImageNamed` is found in `TabsScreenPropsIOS.icon` JS doc?

android/src/main/java/com/swmansion/rnscreens/tabs/appearance/TabsAppearanceApplicator.kt
kligarski: Resolving these using resources should also prevent any problems with different screen densities as in https://github.com/software-mansion/react-native-screens/pull/4209#discussion_r3586692538.

android/src/main/java/com/swmansion/rnscreens/tabs/appearance/TabsAppearanceCoordinator.kt
```diff
@@ -45,7 +52,7 @@ internal class TabsAppearanceCoordinator(
         tabsScreen: TabsScreen,
         appearance: TabsAppearance?,
     ) {
-        appearanceApplicator.updateMenuItemAppearance(menuItem, tabsScreen)
+        appearanceApplicator.updateMenuItemAppearance(menuItem, tabsScreen, resolveIconBoxDp())
```
kligarski: nit: maybe we should move `resolveIconBoxDp()` call to `updateMenuItems` and pass it as an argument to `updateMenuItemAppearance` to avoid resolving this for each menu item separately.

kligarski (commented):
Thanks for the changes, we're going in a good direction.

I left some comments and open questions for which I don't have an answer to yet. I'll try to discuss them with the team during our next meeting but we might need to wait a bit until @kkafar is back from his PTO.

src/components/shared/types.ts
stachbial: ✅ reworded to "whether the host component tints the icon with its state-dependent icon color"

apps/src/screens/CustomNativeBottomTabsIcons.ios.tsx
stachbial: ✅ renamed to `CustomNativeTabsIcons` everywhere

android/src/main/java/com/swmansion/rnscreens/tabs/container/TabsContainer.kt
stachbial: Guarded and verified with a `Test4258` variant toggling drawableIconSize while detached; good point 🫡

android/src/main/java/com/swmansion/rnscreens/tabs/appearance/TabsAppearanceApplicator.kt
stachbial: nice catch 🙏 replaced them with per-access resource lookups of exactly those

android/src/main/java/com/swmansion/rnscreens/tabs/appearance/TabsAppearanceApplicator.kt
stachbial: claude was right; each slot is now inset from its own intrinsic size before the StateListDrawable is built; thanks!

src/components/tabs/screen/TabsScreen.android.types.ts
stachbial: Use case: mixed icon sources (full-bleed brand drawable next to padded Material glyphs) where the bar-wide itemIconSize can't compensate without re-exporting assets that would match. The examples in demo (SWM logo) try to demonstrate exactly this by using the same icon twice 🫡

src/components/shared/types.ts
stachbial: Went for it: tinted on shared `imageSource` (Android both icon types, iOS tinted: true → template path, `templateSource` deprecated) and tested both directions on device ✅ 

on naming I'd keep `tinted` (intent, not capability; docs spell out the per-platform defaults), but happy to rename if the team lands on `tintable`. Lemme know 🫡

android/src/main/java/com/swmansion/rnscreens/tabs/appearance/TabsAppearanceApplicator.kt
stachbial: fair point, removed ✅

ios/helpers/image/RNSImageLoadingHelper.mm
stachbial: added to the `sfSymbol` bullet in `TabsScreenPropsIOS.icon` 🫡

android/src/main/java/com/swmansion/rnscreens/tabs/appearance/TabsAppearanceCoordinator.kt
stachbial: Good point, done ✅ `updateTabAppearance` passes `iconBoxDp` down through `updateMenuItems`; the single-item path keeps a `resolveIconBoxDp()` default argument

stachbial:
Thanks for the review @kligarski !

all addressed in two commits:

[4d3c52d](https://github.com/software-mansion/react-native-screens/pull/4209/commits/4d3c52d81364ed4a08d31070da7365a074ca0b33) — direct fixes: `isAttachedToWindow` guard like #4271 (verified with a `Test4258`-style detach/remount cycle), indicator defaults from `m3_bottom_nav_item_active_indicator_*` (applicator fully stateless again), per-slot icon sizing, dead `!=` check removed, `resolveIconBoxDp()` hoisted, docs + `Tabs` naming.

[176adb2](https://github.com/software-mansion/react-native-screens/pull/4209/commits/176adb2fbbb443759d282367426f6885daf02489) — went ahead with the `imageSource` unification: `tinted` on the shared type, works on both platforms, `templateSource` deprecated. checked on device both ways 🫡

`tinted` vs `tintable` + per-tab sizing left open for the team — replied in-thread.

Cheers

react-navigation
kligarski, line 1: We should revert changing the submodule in this PR.

kligarski (commented):
Thank you for the changes. Only one thing - we should restore `react-navigation` submodule.

I think the code looks good but we still need to consider the open questions regarding icon API for `screens` - let's wait for @kkafar here, he should be back from his PTO at the end of this week/next week.

react-navigation
stachbial: Done!

PR #4211, opened 2026-06-25
feat(iOS, gamma): build gamma as separate ObjC/Swift modules for SwiftPM

chrfalch:
## Description

When working on building RNScreens using SwiftPM, we found that there were some cyclic references between Swift- and ObjC-code - which won't work in SwiftPM since it doesn't support mixed languages in a target - and targets can reference each other in both directions.

### Details:

Gamma's split-view layer mixes ObjC++ component views with Swift controllers that are mutually coupled: the .mm files import the generated RNScreens-Swift.h, and the Swift controllers reference the ObjC view classes. CocoaPods handles this as a single mixed-language module (DEFINES_MODULE), but SwiftPM does not allow mixed-language targets (nor the circular dependency a mixed module would otherwise require), so gamma cannot currently be compiled with SwiftPM.

### Explanation:

This PR breaks the ObjC → Swift compile-time edge so the ObjC and Swift sources can be built as two separate SwiftPM targets, while remaining fully compatible with the existing CocoaPods (mixed-module) build. The change is gated entirely by which sources a build compiles — there is no behavior change when gamma is off, and the CocoaPods path is unaffected.

## Changes

- Introduce pure-ObjC protocols `RNSSplitHostControlling` / `RNSSplitScreenControlling` describing what the component views call on their controllers.
- Instantiate the Swift controllers via `NSClassFromString(...) `instead of `[[ConcreteSwiftClass alloc] init...]`, so no .mm imports `RNScreens-Swift.h`. If the class can't be resolved at runtime, raise `NSInternalInconsistencyException` (visible in Debug and Release).
- Annotate the Swift controllers with `@objc(RNSSplitHostController)` / `@objc(RNSSplitScreenController)` for stable runtime names and conform them to the new protocols.
- Type the controllers' public properties as protocol compositions (`UISplitViewController<RNSSplitHostControlling> *`, `UIViewController<RNSSplitScreenControlling> *`) so Swift's importer keeps them when the module is imported.
- Make `Swift-Bridging.h` a no-op when `RNScreens-Swift.h is` absent (separate-module / non-gamma builds).
- Make `RCTImageSource+AccessHiddenMembers.h` self-contained by importing `<React/RCTImageSource.h>` (Required to build as Clang module)
- Remove now-unused `Swift-Bridging.h` imports from the stack component views.

##  Test code and steps to reproduce

Verified by building RNScreens as a prebuilt XCFramework through Expo's SwiftPM precompile pipeline:
- Gamma enabled (separate ObjC + Swift targets): builds and composes for both Debug and Release.
- Gamma disabled (existing default): still builds and composes — confirming the source changes are mode-agnostic.

| Note: I have not run RNScreens' own example app / CocoaPods build for this change; the edits are designed to be inert in the mixed-module (CocoaPods) path, but a maintainer build there is the authoritative check.

### DIFF
--- a/ios/bridging/Swift-Bridging.h
+++ b/ios/bridging/Swift-Bridging.h
@@ -1,7 +1,10 @@
 #pragma once
 
+// `RNScreens-Swift.h` (the generated ObjC interface of the Swift half) only exists
+// in a single mixed-language module (CocoaPods `DEFINES_MODULE`). Under separate
+// SwiftPM targets it's absent, so `__has_include` makes this header a no-op.
 #if __has_include(<RNScreens/RNScreens-Swift.h>)
 #import <RNScreens/RNScreens-Swift.h>
-#else
+#elif __has_include("RNScreens-Swift.h")
 #import "RNScreens-Swift.h"
 #endif
--- a/ios/gamma/split/RNSSplitHostComponentView.h
+++ b/ios/gamma/split/RNSSplitHostComponentView.h
@@ -7,8 +7,25 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@class RNSSplitHostController;
+@class RNSSplitHostComponentView;
 @class RNSSplitScreenComponentView;
+
+// ObjC-facing surface of the Swift `RNSSplitHostController`, resolved at runtime
+// via NSClassFromString. Lets the ObjC sources drive the controller without
+// importing `RNScreens-Swift.h`, so ObjC and Swift can build as separate SwiftPM
+// targets (SwiftPM has no mixed-language target).
+@protocol RNSSplitHostControlling <NSObject>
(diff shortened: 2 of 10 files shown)


kkafar (commented):
Hey @chrfalch, thanks for the PR!

We've talked with @Ubax and I came to conclusion that we want to pursue different direction here. We'll just completely get rid of Swift from our iOS codebase. This should also solve the problem. 

The earlier intention here behind implementing some of the features with Swift was broader will of migration to the Swift language. We've encountered some problems, however, especially with integration with C++ layer, where we rely in multiple places on React's C++ mounting layer symbols, and came to conclusion that we won't be able to fully migrate anyway.

chrfalch:
Good point - C++ interaction with Swift can become a cascading issue with other libraries.

chrfalch:
Close due to ObjC-rewrite

PR #4226, opened 2026-06-29
refactor(Android, FormSheet v5): Decouple `FormSheetDialogManager` from React's `FormSheetContentView`

t0maboro:
## Description

Previously, the native `FormSheetDialogManager` was tightly coupled with RN, as it was responsible for instantiating the `FormSheetContentView` - a `ReactViewGroup` which is a destination for teleported react children. This PR extracts this React Native dependency out of the native manager.

## Changes

- Removed the FormSheetContentView instantiation from the dialog manager. The manager now accepts an Android View via its constructor. It no longer knows what the content is (ContentView from RN or some other Android View).
- Moved the creation of `FormSheetContentView` to `FormSheetHost` to integrate it directly with RN lifecycle. `FormSheetHost` now adds and removes React subviews directly to/from its instantiated `sheetContentView`, bypassing the native manager completely.

N/A - refactor 

## Test plan

N/A - refactor 

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
@@ -12,7 +12,7 @@ import com.swmansion.rnscreens.gamma.modals.dimmingview.DimmingViewManager
 
 class FormSheetDialogManager(
     context: Context,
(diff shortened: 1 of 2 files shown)


kligarski (approved):
Looks good!

I wonder if we can also handle the `DimmingView`'s dependency on `ReactCompoundViewGroup`/`ReactPointerEventsView` in the future.

PR #4227, opened 2026-06-29
feat(iOS, Stack, Tabs): improve nested container integration -> introduce `Container` & `ContainerItem`

kkafar:
## Description

Port the container-nesting protocol from Android to iOS so nesting information
propagates between containers, and use it to resolve the content scroll view for
special effects (e.g. scroll-to-top) across a nesting boundary.

On iOS the protocol is expressed at the view-controller level, mirroring the
Android side. Shared state/logic lives in two composition holders:
`RNSContainerItemSupport` (item side) and `RNSParentContainerItemRegistry`
(container side).

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1573
Closes https://github.com/software-mansion/react-native-screens-labs/issues/1424

## Changes

- Introduce the `RNSContainer` and `RNSContainerItem` protocols and the two
  composition holders (`RNSContainerItemSupport`, `RNSParentContainerItemRegistry`)
  mirroring the Android implementation.
- `RNSStackNavigationController` and `RNSTabBarController` adopt `RNSContainer`
  and self-register / unregister with the nearest parent `RNSContainerItem` from
  their own `didMoveToParentViewController:` lifecycle (i.e. as the container is
  added to / removed from the view-controller hierarchy). The host views only
  mount the controller.
- `RNSStackScreenController` and `RNSTabsScreenViewController` adopt
  `RNSContainerItem`. Content-scroll-view resolution now follows
  cached -> nested container -> descendant-chain heuristic.
- The parent walk follows the view-controller containment chain
  (`parentViewController`) upwards to the nearest `RNSContainerItem`.
- `RNSTabsScreenViewController.resolveContentScrollView` delegates to
  `findContentScrollView`, removing the duplicated edge / heuristic lookup.
- Wire the stack screen's content scroll view via `setContentScrollView:forEdge:`
  (native `UINavigationBar` scroll edge appearance, top-screen-tap scroll-to-top
  on iPad), matching the tabs screen.

There is no change in behaviour. It should work correctly right now, just not "by accident". 

## Test plan

Use the `test-stack-svm-tabs-special-effects` to confirm that indeed, special effects now do work in nested container scenario, even when the "standard UIKit heuristic" is broken. 

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/gamma/stack/host/RNSStackHostComponentView.mm
+++ b/ios/gamma/stack/host/RNSStackHostComponentView.mm
@@ -47,13 +47,19 @@ namespace react = facebook::react;
 - (void)didMoveToWindow
 {
   RNSLog(@"[RNScreens] StackHost [%ld] attached to window", self.tag);
-  if (self.window != nil && _stackNavigationController.parentViewController == nil) {
-    BOOL mountResult = [RNSContainerHelpers addChildViewController:_stackNavigationController
-                                          toViewControllerManaging:self.reactSuperview
-                                                 withContainerView:self];
-    if (mountResult) {
-      [self setupViewConstraintsForController:_stackNavigationController];
+  if (self.window != nil) {
+    if (_stackNavigationController.parentViewController == nil) {
+      BOOL mountResult = [RNSContainerHelpers addChildViewController:_stackNavigationController
+                                            toViewControllerManaging:self.reactSuperview
+                                                   withContainerView:self];
+      if (mountResult) {
+        [self setupViewConstraintsForController:_stackNavigationController];
+      }
     }
+    
+    [_stackNavigationController attachToParentContainerItem];
+  } else {
+    [_stackNavigationController detachFromParentContainerItem];
   }
 }
 
--- a/ios/gamma/stack/host/RNSStackHostComponentView.mm
+++ b/ios/gamma/stack/host/RNSStackHostComponentView.mm
@@ -56,7 +56,7 @@ namespace react = facebook::react;
         [self setupViewConstraintsForController:_stackNavigationController];
       }
     }
-    
+
     [_stackNavigationController attachToParentContainerItem];
   } else {
     [_stackNavigationController detachFromParentContainerItem];
--- a/ios/gamma/stack/host/RNSStackHostComponentView.mm
+++ b/ios/gamma/stack/host/RNSStackHostComponentView.mm
@@ -47,19 +47,13 @@ namespace react = facebook::react;
 - (void)didMoveToWindow
 {
   RNSLog(@"[RNScreens] StackHost [%ld] attached to window", self.tag);
-  if (self.window != nil) {
-    if (_stackNavigationController.parentViewController == nil) {
-      BOOL mountResult = [RNSContainerHelpers addChildViewController:_stackNavigationController
-                                            toViewControllerManaging:self.reactSuperview
-                                                   withContainerView:self];
-      if (mountResult) {
-        [self setupViewConstraintsForController:_stackNavigationController];
-      }
+  if (self.window != nil && _stackNavigationController.parentViewController == nil) {
+    BOOL mountResult = [RNSContainerHelpers addChildViewController:_stackNavigationController
+                                          toViewControllerManaging:self.reactSuperview
+                                                 withContainerView:self];
+    if (mountResult) {
+      [self setupViewConstraintsForController:_stackNavigationController];
     }
-
-    [_stackNavigationController attachToParentContainerItem];
-  } else {
-    [_stackNavigationController detachFromParentContainerItem];
   }
 }
 
--- a/ios/gamma/stack/host/RNSStackNavigationController.h
+++ b/ios/gamma/stack/host/RNSStackNavigationController.h
@@ -18,16 +18,4 @@
 
 - (void)performContainerUpdateIfNeeded;
 
-/**
- * Register this container with the nearest parent `RNSContainerItem` in the view-controller
- * hierarchy (if any). Call when the owning host view is attached to a window.
- */
-- (void)attachToParentContainerItem;
-
-/**
- * Unregister this container from its parent `RNSContainerItem`. Call when the owning host view is
- * detached from its window.
- */
-- (void)detachFromParentContainerItem;
-
 @end
--- a/ios/gamma/stack/host/RNSStackNavigationController.h
+++ b/ios/gamma/stack/host/RNSStackNavigationController.h
@@ -1,11 +1,12 @@
 #pragma once
 
+#import "RNSContainer.h"
 #include "RNSStackNavigationBarCoordinator.h"
 #include "RNSStackScreenComponentView.h"
 
 @protocol RNSViewFrameChangeDelegate;
 
-@interface RNSStackNavigationController : UINavigationController
+@interface RNSStackNavigationController : UINavigationController <RNSContainer>
 
 @property (nonatomic, weak, nullable) id<RNSViewFrameChangeDelegate> navigationBarFrameChangeDelegate;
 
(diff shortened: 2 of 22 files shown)


ios/gamma/stack/host/RNSStackHostComponentView.mm
```diff
@@ -47,13 +47,19 @@ - (void)initState
...
+      if (mountResult) {
+        [self setupViewConstraintsForController:_stackNavigationController];
+      }
     }
+
+    [_stackNavigationController attachToParentContainerItem];
+  } else {
+    [_stackNavigationController detachFromParentContainerItem];
   }
 }
```
kkafar: I dislike the fact, that this logic is not self contained, i.e. stack navigation controller does not handle this internally, but instead the trigger for attach/detach actions comes from the host. IMO this is wrong and needs to be improved.

ios/gamma/stack/host/RNSStackNavigationController.h
```diff
@@ -17,4 +18,16 @@
...
+ * hierarchy (if any). Call when the owning host view is attached to a window.
+ */
+- (void)attachToParentContainerItem;
+
+/**
+ * Unregister this container from its parent `RNSContainerItem`. Call when the owning host view is
+ * detached from its window.
+ */
+- (void)detachFromParentContainerItem;
+
```
kkafar: I wonder whether these should be "public" methods declared on the interface. Provided that we manage to make the attach/detach actions contained inside the stack navigation controller logic, these do not have to be declared in the header file.

ios/gamma/stack/host/RNSStackNavigationController.mm
```diff
@@ -24,6 +27,49 @@ - (void)initState
...
+
+- (void)willMoveToParentViewController:(UIViewController *)parent
+{
+  [super willMoveToParentViewController:parent];
+}
+
+- (void)didMoveToParentViewController:(UIViewController *)parent
+{
+  [super didMoveToParentViewController:parent];
+}
```
kkafar: Debugging setup. I'll remove this before landing the PR.

ios/helpers/container/RNSParentContainerItemRegistry.mm
```diff
@@ -0,0 +1,47 @@
...
+
+- (void)attachContainer:(UIViewController<RNSContainer> *)container
+{
+  id<RNSContainerItem> parentItem = [self findParentContainerItemFrom:container];
+  if (parentItem == nil) {
+    return;
+  }
+  [parentItem registerNestedContainer:container];
+  _parentItem = parentItem;
+}
```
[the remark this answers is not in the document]
kkafar: No, that's wrong. It's down to the API user to ensure that detach is called at appropriate moment of component lifecycle.

ios/tabs/host/RNSTabsHostComponentView.mm
```diff
@@ -116,13 +116,19 @@ - (void)invalidateImpl
...
+                                                   withContainerView:self];
+      if (mountResult) {
+        [self setupViewConstraintsForController:_controller];
+      }
     }
+
+    [_controller attachToParentContainerItem];
+  } else {
+    [_controller detachFromParentContainerItem];
   }
```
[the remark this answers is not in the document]
kkafar: Couple of remarks:

when the controller registration fails, there are more serious problems at hand that an idempotent redundant call, e.g. nothing will be displayed at all.

Wrapping this call in an if achieves literally nothing. It turns a no-op into a no-op.

ios/gamma/stack/host/RNSStackNavigationController.mm
```diff
@@ -24,6 +27,49 @@ - (void)initState
...
+
+#pragma mark-- Layout
```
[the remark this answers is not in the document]
kkafar: Nope. It makes it easier to scan / search, as it looks like a nested section.

ios/helpers/container/RNSParentContainerItemRegistry.mm
```diff
@@ -0,0 +1,47 @@
...
+- (void)detachContainer:(UIViewController<RNSContainer> *)container
+{
+  [_parentItem unregisterNestedContainer:container];
+  _parentItem = nil;
+}
+
+/**
+ * Walks the view-controller containment chain (`parentViewController`) upwards looking for the
+ * nearest `RNSContainerItem`. We test via `respondsToSelector:` rather than `conformsToProtocol:`
+ * to match the lock-free idiom used elsewhere (e.g. `RNSScrollViewMarkerComponentView`).
```
kligarski, line 32: nit: I'm not sure if we need to mention here that we're using `respondsToSelector` because we're using it in other files.

ios/tabs/screen/RNSTabsScreenViewController.mm
```diff
@@ -105,13 +116,31 @@ - (bool)tabScreenSelectedRepeatedly
...
 {
-  if (auto sv = [self contentScrollViewForEdge:NSDirectionalRectEdgeBottom]; sv != nil) {
-    return sv;
-  }
-  if (auto sv = [self contentScrollViewForEdge:NSDirectionalRectEdgeTop]; sv != nil) {
-    return sv;
-  }
-  return [RNSScrollViewFinder findScrollViewInFirstDescendantChainFrom:[self tabScreenComponentView]];
+  return [self findContentScrollView];
+}
```
kligarski: nit: not sure if we need the intermediate method, we can use `findContentScrollView` directly

ios/gamma/stack/host/RNSStackHostComponentView.mm
kligarski: Why can't we use `didMoveToParentViewController`?

kligarski (commented):
What about `SplitView`? Should it also implement `Container`/`ContainerItem` and handle SVM integration?

kkafar:
@kligarski AFAIK UISplitViewController does not bring any additional interactions, so it seems atm. that integration with that component is not required.

ios/gamma/stack/host/RNSStackHostComponentView.mm
```diff
@@ -47,13 +47,19 @@ - (void)initState
...
+                                                   withContainerView:self];
+      if (mountResult) {
+        [self setupViewConstraintsForController:_stackNavigationController];
+      }
     }
+
+    [_stackNavigationController attachToParentContainerItem];
+  } else {
+    [_stackNavigationController detachFromParentContainerItem];
   }
```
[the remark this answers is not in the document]
kkafar: That is fine. If the registration fails we'll have much bigger problems (nothing will display).

ios/tabs/screen/RNSTabsScreenViewController.mm
kkafar: +1 

https://github.com/software-mansion/react-native-screens/pull/4227/commits/5d627272c1ad214a64afe7eb41817c3bf692a10e

ios/gamma/stack/host/RNSStackHostComponentView.mm
kkafar: Claude sees some objections that I do not fully understand yet 😄

kkafar: Done. I think it's good enough. I have a single concern regarding whether the top-container is ever removed from the root view controller (seems that they are not).

kligarski:
> @kligarski AFAIK UISplitViewController does not bring any additional interactions, so it seems atm. that integration with that component is not required.

Makes sense. If we manage to integrate Stack v5 with SplitView's per-column navigation controllers then everything should work fine.

kkafar:
I'm gonna land this one after the CI passes.

kligarski (approved):
I haven't checked the runtime after the changes but the code looks good.

PR #4228, opened 2026-06-29
feat(Android, Stack v5): add support for groups in toolbar menu

kligarski:
## Description

Adds support for defining groups in (sub)menus which allows creating radio and multiselect menus.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1203.
Closes https://github.com/software-mansion/react-native-screens-labs/issues/1476.

#### API difference between Android and iOS

On iOS, there is no concept of groups. `UIMenu` defines a "group" and its display can be set to `displayInline` to show multiple submenus in one menu. On Android, to achieve similar effect we need to use multiple groups for one (sub)menu. 

For now, I decided to expose groups mechanism directly. Later we'll add an abstraction in JS that will map common JS API (similar to iOS) to Android-compatible groups (e.g. submenu with `displayInline` will be parsed to items in parent menu with additional group).

https://github.com/software-mansion/react-native-screens-labs/issues/1583

#### `singleSelection` difference between Android and iOS

On iOS, if you set `singleSelection` on a menu, all submenus are also single selection. On Android, we map `singleSelection` to `exclusive` group prop. Groups are limited to single submenu only therefore radio across multiple submenus isn't currently supported. 

We would need a custom implementation to handle it - not sure if we want to do so.

#### `groupDivider`

Enabling group divider is possible on every (sub)menu level but I decided not to expose it on all menus as setting it on a submenu just forwards the setting to parent menu - that's how it's implemented in `appcompat`'s `SubMenuBuilder`:

```java
@Override
public void setGroupDividerEnabled(boolean groupDividerEnabled) {
    mParentMenu.setGroupDividerEnabled(groupDividerEnabled);
}
```

#### `checked` option in view command

`checked` option is exposed via view command to allow toggling selection in runtime from JS. It respects `singleSelection`'s behavior but if `checked: false` is provided, the action is no-op (radios can't have all items deselected after first selection).

I was wondering if we should expose separate `toggle` mechanism so that the caller doesn't need to know the current state of the item but I'm not sure if there is any valid use case for this.

## Changes

- JS:
  - expose group configuration (`MenuBase.groups` with `groupId`, `singleSelection`, `onSelectionChange`)
  - add `itemType`, `groupId` and `initialToggleState` to `MenuItem`
  - map `onToolbarMenuGroupSelectionChange` native event to `onSelectionChange` callbacks
  - add assertions for item uniqueness, group uniqueness, item type, valid groupIds on items, valid callbacks (`onPress` for action, `onSelectionChange` for groups/toggles) and valid radio initial selection
  - add `toolbarMenuGroupDividerEnabled` prop for header config
  - add `checked` field in view command's options to allow changing selection in runtime
- tests:
  - add `test-stack-toolbar-menu-groups` SFT
  - adapt `menu-groups` and `nested-menu` SFTs to new test export convention
- native:
  - handle new props and callback in `StackHeaderConfig`
  - add new event for group selection
  - update menu mapper to handle groups
  - add logic for configuring groups and assertions in `StackHeaderApplicator`

## Visual documentation

https://github.com/user-attachments/assets/c60be37d-083f-4587-a8e5-aba05510cf0d

## Test plan

Run `test-stack-toolbar-menu-groups` SFT.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/StackHeaderApplicator.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/StackHeaderApplicator.kt
@@ -300,6 +300,17 @@ internal class StackHeaderApplicator(
         }
     }
 
+    /**
+     * Recursively traverses menu elements and maps user-friendly string item IDs to integers
+     * expected by Android.
+     *
+     * @param elements List of menu elements.
+     * @param forwardIdMap Reference to String->Int ID map to which ID entries will be added.
+     * @param reverseIdMap Reference to Int->String ID map to which ID entries will be added.
+     * @param nextId Function that returns next ID integer. New unique integer should be returned
+     *               each time the function is called. The function is used to handle recursive
+     *               element traversal.
+     */
     private fun assignElementIds(
         elements: List<StackHeaderToolbarMenuElementConfig>,
         forwardIdMap: MutableMap<String, Int>,
@@ -319,6 +330,17 @@ internal class StackHeaderApplicator(
         }
     }
 
+    /**
+     * Recursively traverses menu elements and maps user-friendly string group IDs to integers
+     * expected by Android.
+     *
+     * @param elements List of menu elements.
+     * @param forwardIdMap Reference to String->Int ID map to which ID entries will be added.
+     * @param reverseIdMap Reference to Int->String ID map to which ID entries will be added.
+     * @param nextId Function that returns next ID integer. New unique integer should be returned
+     *               each time the function is called. The function is used to handle recursive
+     *               element traversal.
+     */
     private fun assignGroupIds(
         menuConfig: StackHeaderToolbarMenuConfig,
         forwardMap: MutableMap<String, Int>,
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/StackHeaderApplicator.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/StackHeaderApplicator.kt
@@ -2,6 +2,7 @@ package com.swmansion.rnscreens.gamma.stack.header
 
 import android.content.res.ColorStateList
 import android.graphics.drawable.Drawable
+import android.os.Build
 import android.text.TextUtils
 import android.util.Log
 import android.view.Gravity
@@ -279,6 +280,27 @@ internal class StackHeaderApplicator(
         )
     }
 
+    fun validateRadioInitialSelection(menuConfig: StackHeaderToolbarMenuConfig) {
+        for (group in menuConfig.groups) {
+            if (!group.singleSelection) continue
+            var count = 0
+            for (element in menuConfig.children) {
+                if (element.item.groupId == group.groupId && element.item.initialToggleState) {
+                    count++
+                }
+            }
+            require(count <= 1) {
+                "[RNScreens] Radio group '${group.groupId}' has $count items with " +
+                    "initialToggleState=true. At most 1 is allowed for single-selection groups."
+            }
+        }
+        for (element in menuConfig.children) {
+            if (element is StackHeaderToolbarMenuElementConfig.Submenu) {
+                validateRadioInitialSelection(element.menu)
+            }
+        }
+    }
+
     private fun assignElementIds(
         elements: List<StackHeaderToolbarMenuElementConfig>,
         forwardIdMap: MutableMap<String, Int>,
@@ -286,6 +308,9 @@ internal class StackHeaderApplicator(
         nextId: () -> Int,
     ) {
         for (element in elements) {
+            require(element.item.id !in forwardIdMap) {
+                "[RNScreens] Duplicate toolbar menu item id: '${element.item.id}'. Item IDs must be unique across the entire menu."
+            }
             val nativeId = nextId()
             forwardIdMap[element.item.id] = nativeId
             reverseIdMap[nativeId] = element.item.id
@@ -302,11 +327,12 @@ internal class StackHeaderApplicator(
         nextId: () -> Int,
     ) {
         for (group in menuConfig.groups) {
-            if (group.groupId !in forwardMap) {
-                val nativeId = nextId()
-                forwardMap[group.groupId] = nativeId
-                reverseMap[nativeId] = group.groupId
+            require(group.groupId !in forwardMap) {
+                "[RNScreens] Duplicate toolbar menu group id: '${group.groupId}'. Group IDs must be unique across the entire menu."
             }
+            val nativeId = nextId()
+            forwardMap[group.groupId] = nativeId
+            reverseMap[nativeId] = group.groupId
         }
         for (element in menuConfig.children) {
             if (element is StackHeaderToolbarMenuElementConfig.Submenu) {
@@ -321,12 +347,18 @@ internal class StackHeaderApplicator(
         groupSingleSelection: MutableMap<String, Boolean>,
         groupMemberItems: MutableMap<String, MutableList<String>>,
     ) {
+        val localGroupIds = config.groups.map { it.groupId }.toSet()
         for (group in config.groups) {
             groupSingleSelection[group.groupId] = group.singleSelection
             groupMemberItems.getOrPut(group.groupId) { mutableListOf() }
         }
         for (element in config.children) {
             element.item.groupId?.let { gid ->
+                require(gid in localGroupIds) {
+                    "[RNScreens] Menu item '${element.item.id}' references group '$gid' " +
+                        "which is not defined at the same menu level. " +
+                        "Groups cannot span submenus."
+                }
                 itemGroupMap[element.item.id] = gid
                 groupMemberItems.getOrPut(gid) { mutableListOf() }.add(element.item.id)
             }
@@ -342,18 +374,18 @@ internal class StackHeaderApplicator(
         forwardIdMap: Map<String, Int>,
         reverseIdMap: Map<Int, String>,
         forwardGroupIdMap: Map<String, Int>,
-        onItemClicked: (id: String) -> Unit,
-        onGroupSelectionChanged: (itemId: String) -> Unit,
+        groupDividerEnabled: Boolean,
+        onItemClicked: (id: String, menuItem: MenuItem) -> Unit,
     ) {
         toolbar.menu.clear()
         addElements(toolbar, toolbar.menu, menuConfig, forwardIdMap, forwardGroupIdMap)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            toolbar.menu.setGroupDividerEnabled(groupDividerEnabled)
+        }
         toolbar.setOnMenuItemClickListener { menuItem ->
             val stringId = reverseIdMap[menuItem.itemId]
             if (stringId != null) {
-                onItemClicked(stringId)
-                if (menuItem.isCheckable && menuItem.groupId != Menu.NONE) {
-                    onGroupSelectionChanged(stringId)
-                }
+                onItemClicked(stringId, menuItem)
             }
             true
         }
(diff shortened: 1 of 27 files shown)


src/components/gamma/stack/header/StackHeaderConfig.android.types.ts
```diff
@@ -196,14 +205,100 @@ export interface StackHeaderToolbarMenuItemAndroid
...
    */
   onPress?: (() => void) | undefined;
 }
 
+export interface StackHeaderToolbarMenuGroupAndroid {
+  /**
+   * @summary Unique identifier of the group.
+   *
+   * @platform android
+   */
```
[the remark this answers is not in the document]
kligarski: https://github.com/software-mansion/react-native-screens/pull/4228/commits/20ed848c0676a618d4bc4cbca70e2872cc0d1714

src/components/gamma/stack/header/StackHeaderConfig.android.types.ts
```diff
@@ -229,7 +328,18 @@ export type StackHeaderToolbarMenuElementAndroid =
...
+  /**
+   * @summary Sets the checked state of the menu item.
+   *
+   * @description
+   * In single-selection groups, setting `checked: true`
+   * automatically unchecks other group members.
+   *
+   * @platform android
+   */
+  checked?: boolean | undefined;
```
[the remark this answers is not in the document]
kligarski: https://github.com/software-mansion/react-native-screens/pull/4228/commits/20ed848c0676a618d4bc4cbca70e2872cc0d1714

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/StackHeaderCoordinatorLayout.kt
```diff
@@ -231,6 +257,54 @@ internal class StackHeaderCoordinatorLayout(
...
+        val groupId = toolbarMenuGroupMetadata.itemGroupMap[itemId] ?: return
+        val singleSelection = toolbarMenuGroupMetadata.groupSingleSelection[groupId] ?: return
+        val intId = toolbarMenuForwardIdMap[itemId] ?: return
+        val menuItem = toolbar.menu.findItem(intId) ?: return
+
+        if (singleSelection && explicitCheckedValue == false) {
+            Log.w(
+                TAG,
+                "[RNScreens] Cannot uncheck item '$itemId' in single-selection group '$groupId'. " +
+                    "Check a different item instead.",
```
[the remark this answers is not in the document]
kligarski: I don't think that's a major issue. You shouldn't set radio to false + we're logging this only on the native side with level warn.

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/toolbar/StackHeaderToolbarMenuMapper.kt
```diff
@@ -66,30 +66,66 @@ internal object StackHeaderToolbarMenuMapper {
 
     // region Menu tree parsing
 
-    private fun parseChildren(map: ReadableMap): List<StackHeaderToolbarMenuElementConfig> {
+    private fun parseGroups(map: ReadableMap): List<StackHeaderToolbarMenuGroupConfig> {
+        val array = map.getArray("groups") ?: return emptyList()
+        return (0 until array.size()).map { i ->
+            val group =
+                requireNotNull(array.getMap(i)) {
+                    "[RNScreens] Menu groups array must contain objects."
```
kmichalikk: nit
```suggestion
                    "[RNScreens] Menu groups array must not be null and contain objects with group specification."
```
or sth like that

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/StackHeaderApplicator.kt
```diff
@@ -272,44 +319,143 @@ internal class StackHeaderApplicator(
         }
     }
 
+    private fun assignGroupIds(
+        menuConfig: StackHeaderToolbarMenuConfig,
+        forwardMap: MutableMap<String, Int>,
+        nextId: () -> Int,
```
kmichalikk, line 347: I took me a bit to realize why function is needed here, maybe we should leave a comment, similar in assignElementIds

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/StackHeaderApplicator.kt
```diff
@@ -272,44 +319,143 @@ internal class StackHeaderApplicator(
...
+        }
+        for (element in config.children) {
+            element.item.groupId?.let { gid ->
+                require(gid in localGroupIds) {
+                    "[RNScreens] Menu item '${element.item.id}' references group '$gid' " +
+                        "which is not defined at the same menu level. " +
+                        "Groups cannot span submenus."
+                }
+                itemGroupMap[element.item.id] = gid
+                groupMemberItems.getOrPut(gid) { mutableListOf() }.add(element.item.id)
```
kmichalikk: I think we don't need getOrPut here since we prefilled it before and asserted gid is within those prefilled groups

apps/src/tests/single-feature-tests/stack-v5/test-stack-toolbar-menu-groups-android/index.tsx
```diff
@@ -0,0 +1,360 @@
...
+};
+
+function buildMenu(
+  config: MenuConfig,
+  onItemPress: (id: string) => void,
+  onGroupChange: (groupId: string, selectedIds: string[]) => void,
+): StackHeaderToolbarMenuBaseAndroid {
+  const colorItems = [
+    {
+      type: 'menuItem' as const,
```
kmichalikk: nit: last time I believe we removed `as const`

apps/src/tests/single-feature-tests/stack-v5/test-stack-toolbar-menu-groups-android/scenario.md
```diff
@@ -0,0 +1,302 @@
...
+gamma stack header for Android. It verifies multi-toggle (checkbox)
+and single-selection (radio) group behavior, `onSelectionChange`
+callbacks with correct IDs, `onPress` on ungrouped action items,
+groups scoped to nested submenus, `initialToggleState`,
+`toolbarMenuGroupDividerEnabled`, runtime props updates (group
+type change, adding/removing items, full rebuild resetting command
+state), and imperative `setToolbarMenuItemOptions` commands
+(`checked`, `title`, `hidden`) — including auto-uncheck in
+single-selection groups, no-op on `checked=false` for radio,
+and hidden items preserving their selection in callbacks.
```
kmichalikk:
```suggestion
**Description:** This test focuses on toolbar menu groups on the
Stack v5 header for Android. It verifies multi-toggle (checkbox)
and single-selection (radio) group behavior, `onSelectionChange`
callbacks with correct IDs on groups of items, `onPress` on action items,
`initialToggleState`, `toolbarMenuGroupDividerEnabled`, runtime props updates 
(group type change, adding/removing items, full rebuild resetting command
state), and imperative `setToolbarMenuItemOptions` commands for setting
`checked`, `title`, `hidden`, and hidden items preserving their selection in callbacks.
```

apps/src/tests/single-feature-tests/stack-v5/test-stack-toolbar-menu-groups-android/scenario.md
```diff
@@ -0,0 +1,302 @@
...
+  - Size group (single-selection, `groupId: 'size'`):
+    - `small`: "Small"
+    - `medium`: "Medium" (checked via `initialToggleState`)
+    - `large`: "Large"
+  - `share`: "Share" (ungrouped action item)
+  - `sub`: "More" (submenu containing):
+    - Theme group (single-selection, `groupId: 'theme'`):
+      - `light`: "Light" (checked via `initialToggleState`)
+      - `dark`: "Dark"
+    - `info`: "Info" (ungrouped action item)
```
kmichalikk: I think we don't need this, but up to you

apps/src/tests/single-feature-tests/stack-v5/test-stack-toolbar-menu-groups-android/scenario.md
```diff
@@ -0,0 +1,302 @@
...
+      unchecked.
+
+9. Open the overflow menu and tap "Large" again.
+
+- [ ] No toast is displayed (already selected, single-selection
+      keeps it). Large remains checked.
+
+---
+
+### Ungrouped action item
```
kmichalikk: ungrouped redundant

kmichalikk (commented):
You can mix toggles with `showAsAction: always`; first - should we allow for toggling showAsAction other than never, second - what if some toggles are in header and others are in popup menu? From what I can see this works correctly but has no visual indication to what is toggled.

We may just let users hurt themselves but let's document such scenarios somewhere

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/toolbar/StackHeaderToolbarMenuMapper.kt
kligarski: It might be null - you can define a submenu without any groups (selection behavior), just actions. This require checks that the items inside the array are maps (JS objects).

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/StackHeaderApplicator.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/4228/commits/f1d361b0455f2c03d666b9359fcd15f3a45c6591

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/StackHeaderApplicator.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/4228/commits/f1d5233e2dde7192ce13a87e173960f16040a30c

apps/src/tests/single-feature-tests/stack-v5/test-stack-toolbar-menu-groups-android/index.tsx
kligarski: https://github.com/software-mansion/react-native-screens/pull/4228/commits/7747671ddec98fab357e4f1c3bc099bc7e87b76c

apps/src/tests/single-feature-tests/stack-v5/test-stack-toolbar-menu-groups-android/scenario.md
kligarski: https://github.com/software-mansion/react-native-screens/pull/4228/commits/60d7b166b24214f25a0724f2edf7d866f1ad94c3

apps/src/tests/single-feature-tests/stack-v5/test-stack-toolbar-menu-groups-android/scenario.md
kligarski: https://github.com/software-mansion/react-native-screens/pull/4228/commits/ad103c57a3bfe7ddc1a7416f83424085f1cf9de8

apps/src/tests/single-feature-tests/stack-v5/test-stack-toolbar-menu-groups-android/scenario.md
kligarski: https://github.com/software-mansion/react-native-screens/pull/4228/commits/ad103c57a3bfe7ddc1a7416f83424085f1cf9de8

kligarski:
> You can mix toggles with `showAsAction: always`; first - should we allow for toggling showAsAction other than never, second - what if some toggles are in header and others are in popup menu? From what I can see this works correctly but has no visual indication to what is toggled.
> 
> We may just let users hurt themselves but let's document such scenarios somewhere

https://github.com/software-mansion/react-native-screens/pull/4228/commits/92f26cce7fc7b5c23b6d6ae4be9a004015096993

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/toolbar/StackHeaderToolbarMenuMapper.kt
kmichalikk: I thought that the "must contain objects" and assumed not null. I'd reword that still, but up to you.

kmichalikk (approved):

android/src/main/java/com/swmansion/rnscreens/gamma/stack/header/toolbar/StackHeaderToolbarMenuMapper.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/4228/commits/2724494130e335a8a35cd9b59e4334b6c386a629

PR #4212, opened 2026-06-25
chore(test): Add iPad e2e coverage for tabBarControllerMode

LKuchno:
## Description

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1569

This PR extends e2e test infrastructure to support iPad-specific test runs and adds comprehensive iPad coverage for the tabBarControllerMode prop. The tabBarControllerMode prop exhibits platform-specific behavior (automatic/tabBar modes on iPhone vs. automatic/tabBar/tabSidebar modes on iPad), so dedicated iPad e2e tests are essential to verify the sidebar toggle and layout modes deterministically.

**Infrastructure improvements:**
- Added `describeIfiPad` utility for iPad-only test suites, inferring device idiom from the `RNS_APPLE_SIM_NAME` environment variable
- Extracted shared e2e helpers (`selectComponentIntegrationTestsScreen`, `forceSelectTabByLabel`) to reduce code duplication across tests
- Upgraded iOS version to 26.5 for test stability
- Added bash script and npm task to run iPad-tagged tests on the default iPad simulator

Platforms: iOS (iPhone and iPad) and Android.

## Changes

- **e2e-utils**: Added `isIPadTarget` and `describeIfiPad` helpers for iPad-only test suites; added `selectComponentIntegrationTestsScreen()` and `forceSelectTabByLabel()` utilities
- **ios-devices**: Added `DEFAULT_APPLE_IPAD_SIMULATOR_NAME` constant set to "iPad Pro 13-inch (M4)"; upgraded `DEFAULT_IOS_VERSION` from 26.2 to 26.5; exported device constants for use in npm scripts
- **test-e2e-ios-ipad.sh**: New bash script to run `@ipad`-tagged Detox tests on iPad simulators
- **package.json**: Added `test-e2e-ios-ipad` npm script to invoke iPad e2e tests
- **test-tabs-tab-bar-controller-mode-ios (e2e)**: New iOS-specific e2e test with dual describe blocks: iPad suite verifying sidebar toggle in `tabSidebar` mode; iPhone suite verifying `tabSidebar` behaves like `tabBar` on compact width
- **test-tabs-tab-bar-controller-mode-ios (test screen)**: Added test IDs for e2e element targeting
- **Scenario documentation**: Updated tabBarControllerMode scenario-description.ts and scenario.md files to reflect e2e coverage

PR #4233, opened 2026-06-30
refactor(test): e2e test: tab-bar-hidden implementation for iPad

LKuchno:
## Description

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1584

This PR extends the `test-tabs-tab-bar-hidden` e2e test suite with iPad-specific coverage. The existing test covered the `tabBarHidden` functionality on iPhone and Android, but lacked iPad-specific test assertions. On iPad, the tab bar is rendered as a `_UIFloatingTabBarCollectionView`, requiring dedicated visibility assertions that differ from the iPhone/Android implementations. This PR adds those iPad tests to ensure the tab bar toggles correctly on iPad devices.

## Changes

- **e2e/e2e-utils**: Added `isIPadTarget` flag and `describeIfiPad` helper for conditionally executing iPad-specific e2e test blocks, following the existing pattern of `describeIfiOS` and `describeIfAndroid` - this flad and helper are added also in PR #4212 
- **e2e/test-tabs-tab-bar-hidden.e2e.ts**: Added dedicated iPad test describe block with assertions for `_UIFloatingTabBarCollectionView` visibility toggling behavior


FabricExample/e2e/single-feature-tests/tabs/test-tabs-tab-bar-hidden.e2e.ts
```diff
@@ -1,5 +1,8 @@
 import { device, expect, element, by } from 'detox';
-import { selectSingleFeatureTestsScreen } from '../../e2e-utils';
+import {
+  describeIfiPad,
+  selectSingleFeatureTestsScreen,
+} from '../../e2e-utils';
 
 describe('Tab Bar Hidden', () => {
```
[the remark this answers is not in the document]
LKuchno: iPad run won't execute base suite -> Test for ipad are tagged `@ipad` and will be run against iPad using -t flag. describeIfiPad is set to be skipped if test are run on device different than iPad.

FabricExample/e2e/e2e-utils.ts
```diff
@@ -7,6 +7,12 @@ export const describeIfiOS =
 export const describeIfAndroid =
   device.getPlatform() === 'android' ? describe : describe.skip;
 
+export const isIPadTarget =
+  device.getPlatform() === 'ios' &&
+  /^iPad\s/i.test(process.env.RNS_APPLE_SIM_NAME ?? '');
+
+export const describeIfiPad = isIPadTarget ? describe : describe.skip;
```
kkafar, line 22: Seems that this needs to be rebased after #4212 landed, right?

kkafar (commented):
Code looks ok. Just make sure to rebase. I'll test the runtime now.

FabricExample/e2e/e2e-utils.ts
LKuchno: Yes, thats true, in third PR will be the same :)

kkafar (approved):
Okay, let's rebase this one before proceeding. 

I've checked the runtime, seems to be working fine.

FYI: I've learned that instead of passing `--testNamePattern` to the `jest` and having all the tests but the desired skipped, you can pass `--testPathPattern` and run only the actually matching test.

Issue #3580, opened 2026-01-27
"formSheet" presented screens account/not account for bottom safe area inset


santitopo:
### Description

This issue is a copy from the [original one](https://github.com/react-navigation/react-navigation/issues/12868). Opening it here after I was requested to do so.

### Current behavior

Screens presented with "formSheet" presentation have inconsistent behavior between Android and iOS. 

In iOS the content is rendered above the safe area inset, while in Android it is not, as it can be seen in the image below.

<img width="1696" height="949" alt="Image" src="https://github.com/user-attachments/assets/8e2932cf-713f-4434-bf30-3ef6bdae1135" />

Here's a minimal reproducible example (branch `react-navigation-form-sheet-bug` https://github.com/santitopo/expo-router-stack-header-bug/tree/react-navigation-form-sheet-bug)

### Expected behavior

I think the most flexible behavior would be for both platforms to behave like they do in Android, where no inset is considered. In the hypothetical event that we need some content to render below the bottom system bar in iOS that wouldn't be possible within a screen presented with this technique.

### Reproduction

https://github.com/santitopo/expo-router-stack-header-bug/tree/react-navigation-form-sheet-bug

### Platform

- [x] Android
- [x] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [x] @react-navigation/native-stack
- [ ] react-native-drawer-layout
- [ ] react-native-tab-view

### Environment

- [X] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | "7.1.21"
| @react-navigation/native-stack         | "7.8.0"
| react-native                           | "0.81.5"
| expo                                   | "~54.0.25",
| node                                   | 22.15.1
| npm or yarn                            | 10.9.2

### Steps to reproduce

https://github.com/santitopo/expo-router-stack-header-bug/tree/react-navigation-form-sheet-bug

### Snack or a link to a repository

https://github.com/santitopo/expo-router-stack-header-bug/tree/react-navigation-form-sheet-bug

### Screens version

4.16.0

### React Native version

0.81.5

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

[...]

PR #4238, opened 2026-06-30
fix(android): inset formSheet content above the navigation bar

fabriziocucci:
## What

On Android a content-sized formSheet (`sheetAllowedDetents: 'fitToContents'`) draws its content under the system navigation bar when the keyboard is down. iOS already insets the content for the bottom safe area, so Android ends up clipping the bottom of the content (a `TextInput`, the last row and so on) behind the nav bar.

Keyboard-up is already fine (the sheet sits on top of the keyboard) and this PR leaves it alone.

| | keyboard up | keyboard down |
| --- | --- | --- |
| Android before | ✅ | ❌ |
| Android after | ✅ | ✅ |
| iOS | ✅ | ✅ |

Before and after on Android (content-sized formSheet). In the before the bottom row is clipped under the navigation bar. After it sits above it.

| before | after |
| --- | --- |
| android-before | android-after |

Related to #3580 (iOS insets formSheet content for the bottom safe area, Android does not).

## Why it happens

RNS already computes the right keyboard-aware bottom inset for the sheet in `SheetDelegate.onApplyWindowInsets`:

```kotlin
val newBottomInset = if (!isImeVisible) systemBarsInsets.bottom else 0
// dispatched down via WindowInsetsCompat: nav bar when the keyboard is down, 0 when it is up
```

So the inset value is correct. The problem is that nothing consumes it on Android. `ScreenStackItem` wraps the content in our `SafeAreaView` only when `isIOS26OrHigher`, otherwise on Android it renders the content raw so it runs edge to edge under the nav bar. iOS applies this inset natively, which is why the bug is Android only.

## The change

Extend the existing `SafeAreaView` wrap to Android formSheets with a bottom edge:

```diff
-  const shouldUseSafeAreaView = isIOS26OrHigher;
+  const isAndroidFormSheet =
+    Platform.OS === 'android' &&
+    stackPresentationWithDefault === 'formSheet' &&
+    !unstable_sheetFooter;
+  const shouldUseSafeAreaView = isIOS26OrHigher || isAndroidFormSheet;
   ...
-            <SafeAreaView edges={getSafeAreaEdges(headerConfig)}>
+            <SafeAreaView
+              edges={
+                isAndroidFormSheet
+                  ? { bottom: true }
+                  : getSafeAreaEdges(headerConfig)
+              }>
               {children}
             </SafeAreaView>
```

A native `setPadding` on the content does not work here. `Screen`'s children are laid out by Yoga, so they ignore native parent padding (I tried that first). Reusing our `SafeAreaView` keeps it Fabric-safe.

Because `@react-navigation/native-stack` renders content through `ScreenStackItem` (`NativeStackView.native.tsx`), this also covers react-navigation and expo-router, not just the raw `ScreenStackItem` API.

## Notes on the approach

I put this in `ScreenStackItem`, though I'm not sure it's the best layer. `Screen.tsx` does not wrap children and native-stack goes through `ScreenStackItem`, so it looks like the one spot that reaches every consumer. It also sits next to the existing iOS26 wrap. Raw `<Screen stackPresentation="formSheet">` usage bypasses it, but that already bypasses the iOS26 wrap too. If you'd prefer a shared content path or a native fix, I can move it.

I only apply the bottom edge. `getSafeAreaEdges` only ever returns `top`, since iOS handles the bottom natively. Adding `top` on Android would double up with the native header and the sheet height calc (`lastTopInset`). Left and right (landscape nav bar or display cutout) are 0 in portrait, so I left them for a follow-up.

I skip the wrap when a footer is set. `unstable_sheetFooter` renders outside the wrap and owns the bottom edge, so wrapping the content as well would push it up above the footer.

I don't think detents, scroll content or edge-to-edge are affected. The inset does not depend on sheet height, so fixed detents behave like `fitToContents`. The wrap pads a scroll container the same way the iOS26 wrap already does. The `BottomSheetDialog` is edge-to-edge regardless of the app setting. If the window already consumed the inset the dispatched value is 0 and the wrap is a no-op.

## Tests

This is an inset overlap and the e2e setup here cannot assert insets. The existing formSheet e2e only check which detent is visible. There are already cases skipped for this reason (for example `Test750`). I added a manual repro screen at `apps/src/tests/issue-tests/Test3580.tsx`: a content-sized formSheet with a bottom row that has to clear the nav bar.

Verified by hand on an Android emulator with plain content, on the raw API and on `@react-navigation/native-stack`: keyboard up sits flush, keyboard down clears the nav bar. Before and after screenshots are in the What section above.

If you want an automated check, one option is asserting that a bottom-anchored `testID` is `toBeVisible()` (hidden behind the nav bar when broken, visible once fixed), but that is a new pattern for this repo and would need validating.

## expo-router NativeTabs (coordination)

This fixes the standalone case. It does not fix a separate expo-router symptom where a formSheet under `NativeTabs` shows a gap above the keyboard the size of the tab bar, because expo-router wraps each tab's content in its own `SafeAreaView edges={{ bottom: true }}` (in `NativeTabsView.android`) that is not keyboard-aware. That one is an expo-router change.

Once this lands, a sheet under `NativeTabs` gets wrapped by both this `SafeAreaView` and expo-router's, so the bottom inset can double up. The expo-router side should drop its wrapper for presented sheets (or make it keyboard-aware) so the inset lives in one place.

expo-router follow-up: TODO add the link once the expo-router PR or issue is opened.

### DIFF
--- a/apps/src/tests/issue-tests/Test3580.tsx
+++ b/apps/src/tests/issue-tests/Test3580.tsx
@@ -1,0 +1,70 @@
+import React from 'react';
+import { NavigationContainer, ParamListBase } from '@react-navigation/native';
+import {
+  NativeStackNavigationProp,
+  createNativeStackNavigator,
+} from '@react-navigation/native-stack';
+import { Button, Text, View } from 'react-native';
+
+type RouteParamList = {
+  Home: undefined;
+  FormSheet: undefined;
+};
+
+type NavigationProp<ParamList extends ParamListBase> = {
+  navigation: NativeStackNavigationProp<ParamList>;
+};
+
+type StackNavigationProp = NavigationProp<RouteParamList>;
+
+const Stack = createNativeStackNavigator<RouteParamList>();
+
+function Home({ navigation }: StackNavigationProp) {
+  return (
+    <View style={{ flex: 1, alignItems: 'center', justifyContent: 'center' }}>
+      <Button
+        title="Open FormSheet"
+        onPress={() => navigation.navigate('FormSheet')}
+      />
+    </View>
+  );
+}
+
+function FormSheet() {
+  return (
+    <View style={{ padding: 20, gap: 16 }}>
+      <Text style={{ fontSize: 16, fontWeight: '600' }}>
+        Android formSheet bottom inset
+      </Text>
+      <Text>
+        On Android the content of a content-sized formSheet used to render under the
+        system navigation bar. The row below should sit fully above the nav bar, the
+        same way it already does on iOS.
+      </Text>
+      <View style={{ backgroundColor: '#208AEF', padding: 16, borderRadius: 12 }}>
+        <Text style={{ color: 'white', fontWeight: '600' }}>
+          Bottom row - must clear the navigation bar
+        </Text>
+      </View>
+    </View>
+  );
+}
+
+export default function Test3580() {
+  return (
+    <NavigationContainer>
+      <Stack.Navigator>
+        <Stack.Screen name="Home" component={Home} />
+        <Stack.Screen
+          name="FormSheet"
+          component={FormSheet}
+          options={{
+            presentation: 'formSheet',
+            headerShown: false,
+            sheetAllowedDetents: 'fitToContents',
+          }}
+        />
+      </Stack.Navigator>
+    </NavigationContainer>
+  );
+}
--- a/apps/src/tests/issue-tests/index.ts
+++ b/apps/src/tests/issue-tests/index.ts
@@ -178,6 +178,7 @@ export { default as Test3564 } from './Test3564';
 export { default as Test3566 } from './Test3566';
 export { default as Test3568 } from './Test3568';
 export { default as Test3576 } from './Test3576';
+export { default as Test3580 } from './Test3580';
 export { default as Test3596 } from './Test3596';
 export { default as Test3606 } from './Test3606';
 export { default as Test3611 } from './Test3611';
--- a/src/components/ScreenStackItem.tsx
+++ b/src/components/ScreenStackItem.tsx
@@ -124,7 +124,18 @@ function ScreenStackItem(
     contentStyle = contentWrapperStyles;
   }
 
-  const shouldUseSafeAreaView = isIOS26OrHigher;
+  // Android does not inset formSheet content for the bottom safe area the way
+  // iOS does natively, so content runs under the navigation bar. RNS already
+  // dispatches the keyboard-aware bottom inset (SheetDelegate); we just need the
+  // content to consume it via SafeAreaView. Skipped when a sheet footer is used,
+  // since the footer owns the bottom edge in that case.
+  // eslint-disable-next-line camelcase
+  const hasSheetFooter = unstable_sheetFooter != null;
+  const isAndroidFormSheet =
+    Platform.OS === 'android' &&
+    stackPresentationWithDefault === 'formSheet' &&
+    !hasSheetFooter;
+  const shouldUseSafeAreaView = isIOS26OrHigher || isAndroidFormSheet;
 
   const content = (
     <>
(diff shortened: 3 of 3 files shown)


t0maboro:
Hi @fabriziocucci , thank you for your contribution!

In general, our motivation is to provide the simplest possible components that can be composed as needed. Because of that, I'm not convinced that we should extend the internal use of SAV - which was introduced as a workaround for a specific issue related to iOS 26 rather than as the intended solution for this layer.

If the same result can be achieved by wrapping the content in SAV on the app side, I'd prefer to keep this implementation scoped to its original purpose and avoid expanding its internal responsibilities.

fabriziocucci:
Thanks for taking a look @t0maboro.

The composability principle makes sense to me, so I'll defer to you on where this should live.

Two things I wanted to flag, mostly so I understand the trade-off better. I could easily be missing something.

The first is the iOS/Android difference, which is what nudged me toward doing it in the component. As far as I can tell, on iOS the formSheet already insets its content for the bottom safe area, with no app-side code. On Android it does not, so the same screen that looks right on iOS has its bottom content clipped under the nav bar. In my head this was about making Android match iOS for this component, rather than adding a new internal responsibility. I might be wrong about how iOS is handling it though.

The second is that, as far as I understand from the codebase (but I could be wrong), the app-side wrap is a bit of a footgun. It seems to give the right result only if you wrap with react-native-screens' own `SafeAreaView` (the `experimental` one), because that one consumes the keyboard-aware bottom inset the sheet dispatches in `SheetDelegate` (nav bar when the keyboard is down, 0 when it is up). If someone reaches for `react-native-safe-area-context` instead (which is probably what most people would try), that inset is static and does not drop when the keyboard opens, so it fixes the keyboard-down case but leaves a gap above the keyboard when it is up. In practice "wrap it on the app side" might mean "wrap it with this specific SAV, on Android only, not on iOS", which feels easy to get wrong, but I may be misreading how it behaves.

The main thing for me is that Android and iOS behave consistently here. Right now the same code looks right on iOS but breaks on Android, which is a rough developer experience. Ideally the app author would not have to do anything for this case. It would just be handled. I'm fine with wherever you want the fix to live. I just think consistent cross-platform behaviour with no app-side workaround is worth aiming for.

t0maboro:
@fabriziocucci , I fully understand your perspective from a developer experience. Having both platforms behave 1:1 out of the box is highly desirable. However, adding these layout constraints directly inside the component limits its flexibility. Some developers might not want this safe area behavior, preferring to handle insets manually (e.g., using `useSafeAreaInsets` to apply custom margins or padding). 
Additionally, if stack container is nested within another layout that already accounts for the navigation bar, an embedded `SafeAreaView` would apply its padding on top of it. This creates a "double-offset" issue with no way for the developer to opt out. Following the Single Responsibility Principle, the component should focus on rendering its children within its bounds, without managing system insets or forcing layout side effects.
While it's true that `react-native-safe-area-context` has issues with keyboard-aware insets in this specific scenario, we shouldn't introduce patches into this library to workaround external integration issues.

Instead of embedding this logic into the component, the best approach is to keep it unaware about safe areas as much as possible. We should rather focus on adding a proper section to our documentation with a clear guidance on how to implement the app-side patterns correctly. This will give developers the exact solution they need without compromising the library's architecture.

fabriziocucci:
@t0maboro, quick question on the end state: if the component stays inset-unaware and the app owns insets, will the automatic bottom inset on iOS be removed too so both platforms match?

Also, is there a rough timeline for this? The RN 0.87 branch cut is next week and Expo 57 is coming soon after, so if the plan is to change the iOS behavior or the recommended approach, knowing the timing would help in case the expo-router NativeTabs side needs adjusting to stay consistent.

t0maboro:
@fabriziocucci iOS applies this inset automatically. When we resolve `fitToContents`, we pass the exact content height calculated by Yoga as a custom detent. iOS then internally adjusts that height for the native container, which unfortunately isn't something we can easily override or control from our side.

On the other hand, native Android does not account for the safe area unless insets are handled manually. The root cause of the inconsistency lies entirely at this native platform level, rather than how the component itself is built.

PR #4242, opened 2026-07-01
fix(Android): handle nullable Fresco bitmap

cortinico:
> [!IMPORTANT]  
> This is needed to unblock the react-native@nightly job against react-native-screens.

## Description

RN 0.87 will ship with a Fresco 3.7.0 bump.
Fresco 3.7.0 marks `CloseableBitmap.getUnderlyingBitmap()` as nullable. 

With this bump, `react-native-screens` started failing Android Kotlin compilation because `ImageLoader.kt` was calling `toDrawable()` on the returned bitmap without a null check.

This guards the bitmap before creating the drawable, while preserving compatibility with older Fresco versions where Kotlin sees the Java return type as a platform type.

## Test plan

¯\_(ツ)_/¯

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/helpers/ImageLoader.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/helpers/ImageLoader.kt
@@ -56,8 +56,10 @@ private fun loadImageInternal(
 
                 if (closeableImage is CloseableStaticBitmap) {
                     val bitmap = closeableImage.underlyingBitmap
-                    val drawable = bitmap.toDrawable(context.resources)
-                    onLoaded(drawable)
+                    if (bitmap != null) {
+                        val drawable = bitmap.toDrawable(context.resources)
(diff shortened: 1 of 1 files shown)


cortinico:
cc @kacperkapusciak if you could review and do a new release would be amazing 🙏

t0maboro (approved):
thanks for this PR
internal note: we may need to update the logic in IconResolver when this bitmap is null cc @kligarski

kligarski (approved):
LGTM. We'll wait for CI and merge the changes.

> internal note: we may need to update the logic in IconResolver when this bitmap is null cc @kligarski

I'll handle it in a follow up PR then.

cortinico:
Great 👍 can you let me know when a new release is out also?

kligarski:
Nightly will be out around 12 AM UTC. As for 4.26.0 release, we don’t have a specific release date yet. It should happen in upcoming days/weeks.

PR #4244, opened 2026-07-01
fix(Android, FormSheet v4): refine keyboard avoidance to prevent layout reflows with SafeAreaView

t0maboro:
## Description

Previously, handling keyboard appearance involved modifying window insets on the DecorView level, what caused issues with layout jumping in the whole native hierarchy. We have reworked this model to keep the sheet size stable, relying entirely on `translationY` adjustments driven by SheetAnimationCoordinator. The logic dynamically calculates `keyboardHeightInContainerBounds` and `navigationBarHeightWithinContainer`. This ensures the translation works when the sheet is placed inside a nested container with its own bottom offset.

## Changes

- Background screens and the sheet itself retain their insets, eliminating the layout reflow.
- Introduced `calculateKeyboardShiftHidingBottomInset()` to compute the exact `translationY` correction to properly avoid the keyboard. Instead of lifting the entire sheet, which would push the content too high and leave a content gap caused by the navigation bar, the new logic subtracts the navigation bar overlap from the maximum possible/required keyboard lift.

Test4240

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/cceb0f5f-0ae3-4f67-818e-a2a2c5810710" /> | <video src="https://github.com/user-attachments/assets/174f1da5-17df-4eb8-a994-ab27c607e276" /> |

Test4244

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/71ef0d10-6fa0-42a1-8d8d-def775c1d625" /> | <video src="https://github.com/user-attachments/assets/40583d5a-467a-4fbb-bcfb-a7fe96fc39ce" /> |

## Test plan

Added Test4244.

## Checklist

- [ ] Ensured that CI passes


kmichalikk (approved):
Look good.

There is a bug when rotating the screen but it's also present on main, so it's not blocking here.

PR #4251, opened 2026-07-02
feat(Android, FormSheet v5): Add support for fractional detents

t0maboro:
## Description

Adds configurable detents to the new FormSheets on Android. Due to the native limitations, it allows passing 0-3 detents from JS, which are parsed to BottomSheetBehavior params (peek/half-expanded/expanded/max-height).
I am introducing a `FormSheetBehaviorController` which is a stateless mapper from a resolved detent config to `BottomSheetBehavior`, forcing a fresh layout pass so the behavior re-settles to the new detent configuration.

Additionally, I am making some bugfixes related to or exposed by introducing the sheet detents mechanism.

- Enter-animation flicker on API < 30 - Material's `BottomSheetBehavior` registers a `WindowInsetsAnimationCallback` which updates `translationY` and is causing the sheet to jump during the enter animation (I am modifying translation on a custom slide-in animation). We clear that callback **after** the first layout, because we manage insets ourselves via a fixed container height, preventing the conflict.
- Material only draws the `BottomSheetDialog` edge-to-edge when the theme opts in, and the nav bar is translucent; otherwise, the CoordinatorLayout (and our full-bleed dimming view) is inset below the status bar. We now force edge-to-edge on the dialog window on every API level.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1551

## Changes

- Introduced `FormSheetBehaviorController`  responsible for mapping the raw JS detents array into specific `BottomSheetBehavior` properties.
- Disabled native insets animation to explicitly clear Material's `WindowInsetsAnimationCallback`, unblocking our custom `translationY` enter-animation.
- Forced Edge-to-Edge in `onAttachedToWindow()` in `FormSheetDialog` on both the window decor and the internal Material containers (`R.id.container`, `R.id.coordinator`), ensuring the dimming view covers the status bar area and the FormSheetContainer space is predictable across different API levels.

https://github.com/user-attachments/assets/80c8ed00-643d-4d83-8b35-ebab36bf5fb8

## Test plan

Tested on the base example & 

```tsx
import { Button, ScrollView, StyleSheet, Text, View } from 'react-native';
import { FormSheet, SafeAreaView } from 'react-native-screens/experimental';
import { createScenario } from '@apps/tests/shared/helpers';
import { Colors } from '@apps/shared/styling';

const DETENT_CONFIGS = [
  { label: 'Single detent (0.5)', detents: [0.5] },
  { label: 'Full screen (1.0)', detents: [1.0] },
  { label: 'Two detents (0.3, 0.8)', detents: [0.3, 0.8] },
  { label: 'Half and full (0.5, 1.0)', detents: [0.5, 1.0] },
  { label: 'Three detents (0.2, 0.5, 0.9)', detents: [0.2, 0.5, 0.9] },
  { label: 'Three tall (0.35, 0.65, 1.0)', detents: [0.35, 0.65, 1.0] },
  { label: 'Default fallback (empty array)', detents: [] },
];

function SheetScenario({ label, detents }) {
  const [isOpen, setIsOpen] = useState(false);

  return (
    <View style={styles.scenarioContainer}>
      <Button
        title={`Open: ${label}`}
        color={Colors.primary}
        onPress={() => setIsOpen(true)}
      />

      <FormSheet
        isOpen={isOpen}
        onNativeDismiss={() => setIsOpen(false)}
        detents={detents}>
        <View style={styles.sheetContent}>
          <Text style={styles.sheetTitle}>FormSheet Content</Text>
          <Text style={styles.sheetSubtitle}>
            Static configuration: {JSON.stringify(detents)}
          </Text>
          <View style={styles.spacing} />
          <Button
            title="Close sheet from JS"
            color={Colors.primary}
            onPress={() => setIsOpen(false)}
          />
        </View>
      </FormSheet>
    </View>
  );
}

function TestFormSheetBase() {
  return (
    <SafeAreaView edges={{top: true, bottom: true}} style={styles.container}>
      <Text style={styles.title}>FormSheet Configuration Tests</Text>

      <ScrollView
        style={styles.scrollContainer}
        contentContainerStyle={styles.scrollContent}>
        {DETENT_CONFIGS.map((config, index) => (
          <SheetScenario
            key={index}
            label={config.label}
            detents={config.detents}
          />
        ))}
      </ScrollView>
    </SafeAreaView>
  );
}

  container: {
    flex: 1,
    backgroundColor: Colors.offBackground,
  },
  scrollContainer: {
    flex: 1,
  },
  scrollContent: {
    padding: 16,
    paddingBottom: 40,
  },
  title: {
    fontSize: 20,
    fontWeight: 'bold',
    marginVertical: 20,
    color: Colors.text,
    textAlign: 'center',
  },
  scenarioContainer: {
    marginBottom: 16,
    backgroundColor: 'rgba(0,0,0,0.03)',
    padding: 12,
    borderRadius: 8,
    borderWidth: 1,
    borderColor: 'rgba(0,0,0,0.1)',
  },
  sheetContent: {
    flex: 1,
    backgroundColor: Colors.background,
    padding: 24,
    justifyContent: 'center',
    alignItems: 'center',
  },
  sheetTitle: {
    fontSize: 22,
    fontWeight: '600',
    marginBottom: 8,
    color: Colors.text,
  },
  sheetSubtitle: {
    fontSize: 14,
    color: Colors.text,
    opacity: 0.7,
    fontFamily: 'monospace',
  },
  spacing: {
    height: 32,
  },
});

export default createScenario(TestFormSheetBase, scenarioDescription);
```

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/modals/dimmingview/DimmingViewManager.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/modals/dimmingview/DimmingViewManager.kt
@@ -30,14 +30,8 @@ class DimmingViewManager(
                 )
         }
 
-    internal fun setOnBackdropClickListener(listener: (() -> Unit)?) {
-        if (listener != null) {
-            dimmingView.setOnClickListener {
-                listener()
-            }
-        } else {
-            dimmingView.setOnClickListener(null)
-        }
+    internal fun setOnBackdropClickListener(listener: View.OnClickListener?) {
+        dimmingView.setOnClickListener(listener)
     }
 
     internal fun onDialogShown() {
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/modals/dimmingview/DimmingViewManager.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/modals/dimmingview/DimmingViewManager.kt
@@ -30,9 +30,13 @@ class DimmingViewManager(
                 )
         }
 
-    internal fun setOnBackdropClickListener(listener: () -> Unit) {
-        dimmingView.setOnClickListener {
-            listener()
+    internal fun setOnBackdropClickListener(listener: (() -> Unit)?) {
+        if (listener != null) {
+            dimmingView.setOnClickListener {
+                listener()
+            }
+        } else {
+            dimmingView.setOnClickListener(null)
         }
     }
 
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetBehaviorController.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetBehaviorController.kt
@@ -1,0 +1,100 @@
+package com.swmansion.rnscreens.gamma.modals.formsheet
+
+import android.widget.FrameLayout
+import com.google.android.material.bottomsheet.BottomSheetBehavior
+
+internal class FormSheetBehaviorController(
+    private val sheetView: FrameLayout,
+) {
+    private val behavior = BottomSheetBehavior.from(sheetView)
+
+    private var detents: FormSheetDetents? = null
+    private var sheetAvailableSpace: Int = 0
+    private var shouldApplyInitialState = false
+
+    internal fun updateSheetDetents(detents: FormSheetDetents) {
+        this.detents = detents
+        shouldApplyInitialState = true
+        reconfigureSheetBehavior()
+    }
+
+    /**
+     * @param maxHeight - the full window height that detent fractions are measured against.
+     * Using the full height lets a BottomSheet with large detent configured to slide
+     * behind the status bar, where Material pads insets for us.
+     */
+    internal fun updateSheetAvailableSpace(maxHeight: Int) {
+        if (maxHeight <= 0 || maxHeight == sheetAvailableSpace) {
+            return
+        }
+        sheetAvailableSpace = maxHeight
+        reconfigureSheetBehavior()
+    }
+
+    private fun reconfigureSheetBehavior() {
+        val detents = detents ?: return
+        if (sheetAvailableSpace <= 0) {
+            return
+        }
+
+        val applyInitialState = shouldApplyInitialState
+
+        when (detents.count) {
+            1 -> configureSingleDetent(detents, applyInitialState)
+            2 -> configureTwoDetents(detents, applyInitialState)
+            3 -> configureThreeDetents(detents, applyInitialState)
+            else -> throw IllegalStateException(
+                "[RNScreens] Unsupported detent count ${detents.count}.",
+            )
+        }
+
+        shouldApplyInitialState = false
+
+        // Metrics were mutated on an already-measured sheet; force a fresh layout
+        // pass so the behavior re-settles to the new detent instead of keeping its
+        // stale height.
+        sheetView.requestLayout()
+    }
+
+    private fun configureSingleDetent(
+        detents: FormSheetDetents,
+        applyInitialState: Boolean,
+    ) = behavior.apply {
+        skipCollapsed = true
+        isFitToContents = true
+        maxHeight = detents.maxAllowedHeight(sheetAvailableSpace)
+        if (applyInitialState) {
+            state = BottomSheetBehavior.STATE_EXPANDED
+        }
+    }
+
+    private fun configureTwoDetents(
+        detents: FormSheetDetents,
+        applyInitialState: Boolean,
+    ) = behavior.apply {
+        skipCollapsed = false
+        isFitToContents = true
+        peekHeight = detents.firstHeight(sheetAvailableSpace)
+        maxHeight = detents.maxAllowedHeight(sheetAvailableSpace)
+        // TODO: @t0maboro - in v4 impl the state was passed as a param, consider the same approach
+        if (applyInitialState) {
+            state = BottomSheetBehavior.STATE_COLLAPSED
+        }
+    }
+
+    private fun configureThreeDetents(
+        detents: FormSheetDetents,
+        applyInitialState: Boolean,
+    ) = behavior.apply {
+        skipCollapsed = false
+        isFitToContents = false
+        peekHeight = detents.firstHeight(sheetAvailableSpace)
+        halfExpandedRatio = detents.halfExpandedRatio()
+        expandedOffset = detents.expandedOffsetFromTop(sheetAvailableSpace)
+        maxHeight = detents.maxAllowedHeight(sheetAvailableSpace)
+        // TODO: @t0maboro - in v4 impl the state was passed as a param, consider the same approach
+        if (applyInitialState) {
+            state = BottomSheetBehavior.STATE_COLLAPSED
+        }
+    }
+}
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetBehaviorController.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetBehaviorController.kt
@@ -8,47 +8,31 @@ internal class FormSheetBehaviorController(
 ) {
     private val behavior = BottomSheetBehavior.from(sheetView)
 
-    private var detents: FormSheetDetents? = null
-    private var sheetAvailableSpace: Int = 0
-    private var shouldApplyInitialState = false
-
-    internal fun updateSheetDetents(detents: FormSheetDetents) {
-        this.detents = detents
-        shouldApplyInitialState = true
-        reconfigureSheetBehavior()
-    }
-
     /**
-     * @param maxHeight - the full window height that detent fractions are measured against.
+     * @param detents - parsed detents configuration.
+     * @param sheetAvailableSpace - the full window height that detent fractions are measured against.
      * Using the full height lets a BottomSheet with large detent configured to slide
      * behind the status bar, where Material pads insets for us.
+     * @param applyInitialState - whether to force the sheet to snap to its default starting position
+     * Should be set to `true` when the detents configuration changes.
      */
-    internal fun updateSheetAvailableSpace(maxHeight: Int) {
-        if (maxHeight <= 0 || maxHeight == sheetAvailableSpace) {
-            return
-        }
-        sheetAvailableSpace = maxHeight
-        reconfigureSheetBehavior()
-    }
-
-    private fun reconfigureSheetBehavior() {
-        val detents = detents ?: return
+    internal fun updateSheetBehavior(
+        detents: FormSheetDetents,
+        sheetAvailableSpace: Int,
+        applyInitialState: Boolean = false,
+    ) {
         if (sheetAvailableSpace <= 0) {
             return
         }
 
-        val applyInitialState = shouldApplyInitialState
-
         when (detents.count) {
-            1 -> configureSingleDetent(detents, applyInitialState)
-            2 -> configureTwoDetents(detents, applyInitialState)
-            3 -> configureThreeDetents(detents, applyInitialState)
+            1 -> configureSingleDetent(detents, sheetAvailableSpace, applyInitialState)
+            2 -> configureTwoDetents(detents, sheetAvailableSpace, applyInitialState)
+            3 -> configureThreeDetents(detents, sheetAvailableSpace, applyInitialState)
             else -> throw IllegalStateException(
                 "[RNScreens] Unsupported detent count ${detents.count}.",
             )
         }
-
-        shouldApplyInitialState = false
 
         // Metrics were mutated on an already-measured sheet; force a fresh layout
         // pass so the behavior re-settles to the new detent instead of keeping its
@@ -58,6 +42,7 @@ internal class FormSheetBehaviorController(
 
     private fun configureSingleDetent(
         detents: FormSheetDetents,
+        sheetAvailableSpace: Int,
         applyInitialState: Boolean,
     ) = behavior.apply {
         skipCollapsed = true
@@ -70,6 +55,7 @@ internal class FormSheetBehaviorController(
 
     private fun configureTwoDetents(
         detents: FormSheetDetents,
+        sheetAvailableSpace: Int,
         applyInitialState: Boolean,
     ) = behavior.apply {
         skipCollapsed = false
@@ -84,6 +70,7 @@ internal class FormSheetBehaviorController(
 
     private fun configureThreeDetents(
         detents: FormSheetDetents,
+        sheetAvailableSpace: Int,
         applyInitialState: Boolean,
     ) = behavior.apply {
         skipCollapsed = false
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDetents.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDetents.kt
@@ -22,12 +22,12 @@ internal class FormSheetDetents(
 
     internal val count: Int get() = detents.size
 
-    internal fun at(index: Int): Double = detents[index]
+    private fun heightFractionAt(index: Int): Double = detents[index]
 
-    internal fun heightAt(
+    private fun heightAt(
         index: Int,
         containerHeight: Int,
-    ): Int = (at(index) * containerHeight).toInt()
+    ): Int = (heightFractionAt(index) * containerHeight).toInt()
 
     internal fun firstHeight(containerHeight: Int): Int = heightAt(0, containerHeight)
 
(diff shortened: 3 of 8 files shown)


android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDetents.kt
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        require(detents == detents.distinct().sorted()) {
+            "[RNScreens] Detents must be sorted in strictly ascending order."
+        }
+    }
+
+    internal val count: Int get() = detents.size
+
+    internal fun at(index: Int): Double = detents[index]
```
kmichalikk: tiny nit: sounds a bit like a keyword, maybe `heightFractionAt` ?

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetBehaviorController.kt
```diff
@@ -0,0 +1,87 @@
...
+
+    private fun configureThreeDetents(
+        detents: FormSheetDetents,
+        sheetAvailableSpace: Int,
+        applyInitialState: Boolean,
+    ) = behavior.apply {
+        skipCollapsed = false
+        isFitToContents = false
+        peekHeight = detents.firstHeight(sheetAvailableSpace)
+        halfExpandedRatio = detents.halfExpandedRatio()
```
kmichalikk, line 79: should we reset this for other configure calls?

kmichalikk (approved):

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetBehaviorController.kt
t0maboro: I'm following the v4 implementation here: https://github.com/software-mansion/react-native-screens/blob/main/android/src/main/java/com/swmansion/rnscreens/bottomsheet/BottomSheetBehaviorExt.kt . Looks like leaving it undefined doesn't configure snap point at all, so I think it's fine

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDetents.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4251/commits/1c1aa2ee98a0eca77d9c5541c2776c2594c741ad

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDetents.kt
kkafar, line 1: I really like the `FormSheetDetents` utility class - good organisation.

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetBehaviorController.kt
```diff
@@ -0,0 +1,87 @@
...
+    private fun configureTwoDetents(
+        detents: FormSheetDetents,
+        sheetAvailableSpace: Int,
+        applyInitialState: Boolean,
+    ) = behavior.apply {
+        skipCollapsed = false
+        isFitToContents = true
+        peekHeight = detents.firstHeight(sheetAvailableSpace)
+        maxHeight = detents.maxAllowedHeight(sheetAvailableSpace)
+        // TODO: @t0maboro - in v4 impl the state was passed as a param, consider the same approach
```
kkafar, line 65: Yeah, it might be useful to support `initialDetentIndex` or whatever prop.

kkafar: Not needed in this very moment though. Looks good. You might modify that on is-needed basis.

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialog.kt
kkafar, line 27: These lines have not beed modified in this PR, but I notice them here for the first time. 

They are risky. AFAIK the window object you modify in their implementation is shared between FormSheetDialog view hierarchy, "original view hierarchy" (where react root view resides), and any other dialog opened before/after the form sheet. 

Therefore modifying global window settings will affect behaviour of not only the formsheet, but also all other dialogs. 

E.g. if you disable the dimming view this way, my prediction is that the `<Modal />` component from `react-native` won't have dimming view anymore. Similarly the animations will be disabled.

Please verify this. In case I'm right (let's hope I'm wrong).

kkafar: Okay! We did some research together with @t0maboro and it turns out that each Dialog on Android creates it's own window instance! That's great then.

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialog.kt
```diff
@@ -20,6 +22,27 @@ internal class FormSheetDialog(
...
+     *
+     * We force edge-to-edge on every API level so the custom dimming view covers the whole screen.
+     * The sheet content is sized manually against the system insets, so drawing behind the bars is safe here.
+     */
+    private fun forceEdgeToEdge() {
+        val window = window ?: return
+        WindowCompat.setDecorFitsSystemWindows(window, false)
+        findViewById<View>(com.google.android.material.R.id.container)?.fitsSystemWindows = false
+        findViewById<View>(com.google.android.material.R.id.coordinator)?.fitsSystemWindows = false
+    }
```
kkafar, line 50: This is an important assumption. We MUST emphasise this in the component description that right now we enforce / assume edge-to-edge mode being enabled.

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
```diff
@@ -81,12 +126,27 @@ class FormSheetDialogManager(
                 override fun onPreDraw(): Boolean {
                     view.viewTreeObserver.removeOnPreDrawListener(this)
                     view.translationY = view.height.toFloat()
```
kkafar: Why do we set the `view.translationY` here? You already do that in `runEnterAnimation`.

kkafar: I've removed this line & it seems to be working just fine. Please see to it.

kkafar: I've noticed this, because setting translationY for animation purposes in `onPreDraw` seems rather out of place.

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
```diff
@@ -81,12 +126,27 @@ class FormSheetDialogManager(
...
+                    disableMaterialInsetsAnimationCallback(view)
```
kkafar, line 127: Also this one: why do we do it in `onPreDraw`? Can't we do it much earlier, e.g. after the `BottomSheetBehavior` of the formsheet is initialized for the very first time? When does `BottomSheetBehavior` / `BottomSheetDialog` register this callback?

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
```diff
@@ -117,24 +179,30 @@ class FormSheetDialogManager(
...
      * the system insets, we completely bypass these redundant layout passes.
      */
-    private fun updateNativeContainerHeight(
-        view: View,
-        insets: WindowInsetsCompat,
-    ) {
-        val topInset = getTopInset(insets)
-        val bottomInset = getBottomInset(insets)
+    private fun updateNativeContainerHeight() {
         val dialogDecorHeight = dialog.window?.decorView?.height ?: 0
```
kkafar, line 182: Just a style guide -> if there is nothing to do if `dialogDecorHeight == 0` then just early return and remove one level of indentation from the rest of the code.

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
kkafar: This would make more sense to take `null` as any other API in Android, instead of "empty callback".

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
```diff
@@ -160,4 +228,8 @@ class FormSheetDialogManager(
         dialog.setOnCancelListener(null)
         dialog.dismiss()
     }
+
+    companion object {
+        private const val LARGE_DETENT = 1.0
```
kkafar:
```suggestion
        private const val LARGE_DETENT_FRACTION = 1.0
```

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
```diff
@@ -66,9 +77,43 @@ class FormSheetDialogManager(
...
 
+        if (resolvedDetents == null || formSheetConfig.detents != newConfig.detents) {
+            resolvedDetents = resolveDetents(newConfig.detents)
+            shouldReconfigureDetents = true
+        }
+
+        // TODO: @t0maboro
+        // - a dedicated presentation manager should be introduced as on iOS,
+        // - invalidation flags logic should be implemented following other components convention
+        val isOpening = newConfig.isOpen && !formSheetConfig.isOpen
```
kkafar: You already test for that earlier in this function. Please move this value to the top of the scope and reuse it.

kkafar (commented):
Great job. I think that the abstractions you've introduced make sense. 

I have series of remarks regarding some details. Let's iron them out before proceeding.

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDetents.kt
t0maboro: 👍

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetBehaviorController.kt
t0maboro: There was also some logic in v4 SheetDelegate impl. around the keyboard handling - I haven't verified whether it still applies in the new implementation, so leaving myself a note to revisit this line

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialog.kt
t0maboro: https://cs.android.com/android/platform/superproject/+/android-16.0.0_r4:frameworks/base/core/java/android/app/Dialog.java;l=208-209 pasting here a link to the source code if we need to revisit it for some reason

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
t0maboro: it's registered on the first `onLayoutChild` call here: https://github.com/material-components/material-components-android/blob/release-1.13/lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java#L556

t0maboro: I'm already changing it in the followup PR to use `doOnLayout` https://github.com/software-mansion/react-native-screens/blob/e3bd3dbf520d570e62cff03639eefe3e13671e46/android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDimensionsCoordinator.kt#L46-L50 - let me know if that's okay regarding timing, I did it on pre-draw here to ensure that I am after the layout pass

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialog.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4251/changes/269d2072c805bf3031037b5699968cb02438854c

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4251/changes/d78d65a58599fa7f6053963eb86c23af7bdc683d

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4251/changes/c97048f6d5ecaf0831985ee064e9ce26071d6ea5

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4251/changes/1d512b552a2197642b7c43c6655c5567e9e732a0

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4251/changes/5026fddc766e730f540a812e7b2be326d059d66f

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4251/changes/63ca1dc894fd47952819ccdd85d2d13f012ddada

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
kkafar: I think it is & I like it more in comparison to a global event listener. 

Sounds good. If you modify it in follow-up PR I'm willing to merge current shape of thing here.

android/src/main/java/com/swmansion/rnscreens/gamma/modals/dimmingview/DimmingViewManager.kt
```diff
@@ -30,9 +30,13 @@ class DimmingViewManager(
...
-    internal fun setOnBackdropClickListener(listener: () -> Unit) {
-        dimmingView.setOnClickListener {
-            listener()
+    internal fun setOnBackdropClickListener(listener: (() -> Unit)?) {
+        if (listener != null) {
+            dimmingView.setOnClickListener {
+                listener()
+            }
+        } else {
+            dimmingView.setOnClickListener(null)
```
kkafar: One more nit:

why don't we just use `OnClickListener` as the parameter type & instead define our own? If all we do is pass that value through I guess this is the only thing that makes sense.

kkafar (commented):
Looks really good. I have just a fe more remarks & we can land it.

android/src/main/java/com/swmansion/rnscreens/gamma/modals/dimmingview/DimmingViewManager.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4251/commits/cd383c5d2082f6532d232781e0e649a4864d780a

kkafar (approved):

PR #4253, opened 2026-07-02
feat(iOS): implement unstable_sheetFooter for formSheet

birkir:
## Description

`unstable_sheetFooter` is currently Android-only — `RNSScreenFooter.mm` ships as a stub with the positioning logic commented out. This PR implements the iOS side, mirroring the contract of Android's `ScreenFooter`: the footer overlays the sheet content, pinned to the bottom edge of the sheet, and lifts above the keyboard when one appears. Consumers reserve space for it (when desired) by padding their scroll content; since it overlays, it can also serve as a surface for blur / scroll-edge effects layered over the content.

The abandoned constraint-based approach (the commented-out code) was replaced with frame pinning: Fabric assigns frames directly and fights Auto Layout, so the footer instead accepts Yoga's size and re-pins `origin.y` after every layout pass, from both directions:

- `updateLayoutMetrics:` — React-driven writes (footer content resize, state updates), and
- the parent `RNSScreenView`'s `layoutSubviews` — UIKit-driven writes, which makes the footer track **interactive detent drags** natively since `UISheetPresentationController` resizes the presented view continuously during the gesture.

Keyboard handling is the counterpart of Android's `WindowInsetsAnimationCompat` callback: the footer computes the keyboard's overlap with the sheet (converted through the window, so floating / centered iPad sheets get correct geometry) and animates with the keyboard's own duration and curve.

## Changes

- `ios/RNSScreenFooter.mm` — implemented pinning, keyboard tracking, `prepareForRecycle` cleanup (replaces the stub).
- `ios/RNSScreenFooter.h` — exposed `updateFooterPosition`.
- `ios/RNSScreen.h` / `ios/RNSScreen.mm` — added `findSheetFooter` (analogous to `findHeaderConfig`) and a `layoutSubviews` override on `RNSScreenView` that re-pins the footer during UIKit-driven resizes.
- `src/types.tsx` — `@platform android` → `@platform android, ios` on `unstable_sheetFooter`.
- `guides/GUIDE_FOR_LIBRARY_AUTHORS.md` — removed the Android-only note, documented the overlay contract.

No new props, no codegen changes — the existing Fabric component registration is reused.

## Test plan

- Existing `apps/src/tests/issue-tests/TestScreenFooterKeyboardInsets.tsx` — now exercisable on iOS.
- Existing `apps/src/tests/issue-tests/TestFormSheet.tsx` — the commented-out `unstable_sheetFooter` options can be enabled on iOS.
- Field-tested in a production app (patched into 4.25.2) on iPhone 17 Pro and iPad Pro 11" (iOS 26): `formSheet` with `fitToContents` and detent arrays, footer pinned correctly at rest, during interactive detent drags, during sheet resize when footer content height changes, and above the keyboard; buttons at the sheet bottom remain tappable (frame-based pinning keeps UIKit hit-testing correct).

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
+++ b/guides/GUIDE_FOR_LIBRARY_AUTHORS.md
@@ -447,17 +447,21 @@ function Home() {
 }
 ```
 
-### `unstable_sheetFooter` (Android only)
+### `unstable_sheetFooter`
 
 Footer component that can be used alongside form sheet stack presentation style.
 
 This option is provided, because due to implementation details it might be problematic
 to implement such layout with JS-only code.
 
+The footer overlays the sheet content and is pinned to the bottom edge of the sheet,
+moving above the keyboard when one appears. Reserve space for it (if needed) by padding
+the sheet's scroll content.
+
 Please note that this prop is marked as unstable and might be subject of breaking changes,
 even removal.
 
-Currently supported on Android only.
+Supported on Android and iOS.
 
 ### `contentStyle`
 
--- a/ios/RNSScreen.h
+++ b/ios/RNSScreen.h
@@ -54,6 +54,7 @@ NS_ASSUME_NONNULL_BEGIN
 @end
 
 @class RNSScreenStackHeaderConfig;
+@class RNSScreenFooter;
 
 @interface RNSScreenView : RNSReactBaseView <RNSScreenContentWrapperDelegate,
                                              RNSScrollViewBehaviorOverriding,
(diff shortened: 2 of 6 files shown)


kligarski (commented):
Hi, @birkir. Thank you for the PR.

As we're focused on development of v5 of the library for which we're implementing new native-stack and new FormSheets from scratch, we have limited resources to add/review new features to v4 implementation and we're most likely not going to be able to proceed with the PR. Please also see the [contributing guide](https://github.com/software-mansion/react-native-screens/blob/main/guides/CONTRIBUTING.md#contributing-to-code).

Additionally, the footer should be easily achievable in JS as form sheets on iOS support synchronous state updates now.

For v5, the footer is planned for both platforms.

Also, I haven't reviewed the PR in detail but I think that we will have a problem with handling Pressables when keyboard is shown because they rely on Shadow Tree and we're not updating the position of the footer. Please see the following video:

https://github.com/user-attachments/assets/32a31051-ce9c-4217-a007-fd2869367b52

PR #4255, opened 2026-07-03
chore(test): Add iPad e2e coverage for tabs bottomAccessory layout

LKuchno:
## Description

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1592

Adds iPad end-to-end coverage for the tabs `bottomAccessory` layout test and
refactors the shared `getElementAttributes` helper so elements can be matched
by label, testID, or native type (with optional index disambiguation).

Detox cannot resize the app window to the compact-width size class, so the
iPad block covers the full-size app case only: the initial load and all five
content variants, asserting the accessory renders and stays anchored to the
bottom of the screen. The scenario doc is updated to reflect what is and is
not covered on each platform.

## Changes

- `FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-layout-ios.e2e.ts`:
  added a `describeIfiPad` block covering the initial load and all five
  accessory variants on iPad. Each test is declared with its own inline
  `it(...)`; the shared variant data (card + expected accessories) lives in a
  `VARIANTS` object and the per-variant body behind
  `switchToVariantAndVerifyAccessory`, with `verifyConfigTabInitialContent` for
  the initial-load check. The iPad tests additionally assert the accessory is
  anchored to the bottom via `expectBottomAccessoryAtTheBottom`, which derives
  the window's bottom safe-area band at runtime from a window-spanning element
  rather than a hardcoded inset, so it stays correct across devices / OS
  versions.
- `FabricExample/e2e/e2e-utils.ts`: `getElementAttributes` now takes an
  `{ by, value, index }` matcher (label / id / type) instead of a bare label
  string; `forceTapByLabeliOS` updated accordingly.
- `apps/src/tests/.../scenario.md`: updated the **E2E test** section to
  describe iPhone vs iPad coverage and the remaining iPad compact-width gap.


FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-layout-ios.e2e.ts
```diff
@@ -12,8 +14,22 @@ const bottomAccessoryElement = (testID: string) =>
...
+const getExtendedTabBarAttributes = async () =>
+  getElementAttributes({
+    by: 'type',
+    value: 'UITabBar',
+    index: 0,
+  }) as Promise<IosElementAttributes>;
+
+async function expectBottomAccessoryExist(testID: string) {
+  await expect(bottomAccessoryElement(testID)).toExist();
 }
```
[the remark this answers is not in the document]
LKuchno: It's the way it has to be check after changes https://github.com/software-mansion/react-native-screens/pull/4196 as it's not detected as visible now.

FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-layout-ios.e2e.ts
```diff
@@ -40,6 +56,23 @@ function expectBottomAccessoryInline(
...
+  scrollView: IosElementAttributes,
+) {
+  // The accessory sits above the scroll view's bottom edge by the 20pt padding
+  // set on the scroll view's style in the test screen.
+  const SCROLL_VIEW_BOTTOM_PADDING = 20;
+  const bottomAccessoryBottom =
+    bottomAccessory.frame.y +
+    bottomAccessory.frame.height +
+    SCROLL_VIEW_BOTTOM_PADDING;
+  const scrollViewBottom = scrollView.frame.y + scrollView.frame.height;
```
kkafar: What kind of magic is this? 😄 

What if the padding changes to 22 in next patch version of the system? Will it still work? Why do we even need this?

FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-layout-ios.e2e.ts
```diff
@@ -71,35 +167,20 @@ describeIfiOS('Tabs bottomAccessory (iOS)', () => {
       element(by.type('RNSTabsBottomAccessoryComponentView')),
     ).toBeVisible();
   });
+}
```
kkafar: This is compact ok, but it's hard to read. 

Let's keep convention that we use `it('test name')` to declare each test **inline**, so that it is trivial to see where it is defined and to modify. 

You have to declare all the names any way. 
Then, the test implementation can be put behind an abstraction.

kkafar (changes requested):
The test works. I have few concerns though. Let's iron them out before we proceed.

FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-layout-ios.e2e.ts
LKuchno: Container padding is defined in the index file, so imo it will work unless we change it. -> I was wrong, you were right padding of Scroll View is accidentally the same as the iPad's bottom safe-area inset. 
I didn't find a better way to check that bottomAccessory is actual at the bottom of the screen. If you have another idea, feel free to share - I'm open to suggestions.

LKuchno: This approach seems to be more stable https://github.com/software-mansion/react-native-screens/pull/4255/changes/1e4a25c570fe85b43761cf7c9ddeb2267e7295e2

FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-layout-ios.e2e.ts
LKuchno: https://github.com/software-mansion/react-native-screens/pull/4255/changes/dd2866b3e5697b63af6af275918277a0593c3bd8

FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-layout-ios.e2e.ts
```diff
@@ -256,16 +300,74 @@ describeIfiOS('Tabs bottomAccessory (iOS)', () => {
...
+    await selectSingleFeatureTestsScreen(
+      'Tabs',
+      'test-tabs-bottom-accessory-layout-ios',
+    );
+    await expect(element(by.id('config-scrollview'))).toBeVisible();
+    configScrollView = (await getElementAttributes({
+      by: 'id',
+      value: 'config-scrollview',
+      index: 0,
+    })) as IosElementAttributes;
```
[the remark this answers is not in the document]
LKuchno: Not applying this. Since config-scrollview is a full-window view, its frame doesn't drift, only safeAreaInsets.bottom might. However, the assertion has enough slack that a cached inset still passes comfortably. A false failure would mean the accessory rose above the safe area, which violates the exact behavior we're testing.

FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-layout-ios.e2e.ts
```diff
@@ -12,6 +14,20 @@ const bottomAccessoryElement = (testID: string) =>
     by.id(testID).withAncestor(by.type('RNSTabsBottomAccessoryComponentView')),
   ).atIndex(0);
 
+const getBottomAccessoryAttributes = () =>
+  getElementAttributes({
+    by: 'type',
+    value: 'RNSTabsBottomAccessoryComponentView',
```
kkafar: Please, extract the native symbol type to a shared top-level variable. One less place to refactor in case we rename it.

FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-layout-ios.e2e.ts
kkafar: Seems better. Good enough, thanks.

kkafar (approved):
Looks good. I have just one nit. I think that these component names you use across different tests should be put inside one single module shared between e2e tests, so that there is a single truth for e2e tests for that. What do you think? Do you think it makes sense to refactor all the tests?

FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-layout-ios.e2e.ts
LKuchno: I moved it to  native-type-names.ts file (name already changed in commit there is another one) I would extend this file with other native types like UITabBar, _UIButtonBarButton etc. but in separate PR. I just think if it won't be better to place it inside e2e/helpers directory?

https://github.com/software-mansion/react-native-screens/pull/4255/changes/b533910ec0ce39b0549f38c2da4027d9d3c2cda3

kkafar: Yeah, definitely in a separate PR. 

> I just think if it won't be better to place it inside e2e/helpers directory?

Might be. I leave it up to you to organise this. I'd just like to have it separated and not inline-defined in every single file.

PR #4259, opened 2026-07-03
feat(Android, FormSheet v5): Handle presentation state updates during transitions

t0maboro:
## Description

This PR addresses animation race conditions caused by rapid toggles of the `isOpen` prop from JS and refactors the `FormSheetDialogManager`.
Following the iOS implementation, I am introducing a state machine for the presentation lifecycle, ensuring that asynchronous commands from JS wait for ongoing native transitions to finish before executing.

## Changes

- Introduced `FormSheetPresentationManager` which is a state machine that enqueues incoming presentation commands. If JS requests to open the sheet while it is dismissing, the manager safely waits for the exit animation to complete before triggering the entrance.
- Refined the approach with a `LifecycleCoordinator` - I want to set up an `onShow` listener while the manager is about to present the dialog to prepare all necessary callbacks to be scheduled right before the dialog is shown.
- Renamed and refactored the previous `AnimationCoordinator` into a factory. The animation state is no longer hidden; the factory simply yields configured Animator objects, while the `PresentationManager` controls their execution and cancellation.
- Introduced `NativeDismissCoordinator` which replaces `LifecycleCoordinator`, ensuring that both real native dismissal (e.g., swiping down) and tapping our custom dimming view follow the same dismissal paths.
- Removed `applyInitialState` from sheet detents update - when we need to reopen the same formsheet or detents are changed - in both cases we should reapply the configuration to enustre that the native state is well synchronized after the last dismissal.

https://github.com/user-attachments/assets/6dfdff4b-85ed-4803-b87b-59e6e1fd6685

## Test plan

Updated `test-form-sheet-presentation-state` to be supported on Android.

## Checklist

- [ ] Ensured that CI passes


android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
```diff
@@ -52,59 +52,65 @@ class FormSheetDialogManager(
...
+                    animationCoordinator.runEnterAnimation(bottomSheetView, onComplete)
+                }
+                dialog.show()
+            },
+            performDismiss = { onComplete ->
+                animationCoordinator.runExitAnimation(bottomSheetView) {
+                    dialog.dismiss()
+                    onComplete()
+                }
+            },
```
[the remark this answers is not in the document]
t0maboro: we're calling this from the UI thread, so on `dismissDialog` will be called synchronously
```
 if (Looper.myLooper() == mHandler.getLooper()) {
            dismissDialog();
```
and it will call `mWindowManager.removeViewImmediate(mDecor);` so the view will be torn down immediately, imo no race condition here

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetPresentationState.kt
```diff
@@ -0,0 +1,8 @@
+package com.swmansion.rnscreens.gamma.modals.formsheet
+
+internal enum class FormSheetPresentationState {
+    DISMISSED,
+    DISMISSING,
+    PRESENTED,
+    PRESENTING,
+}
```
kkafar, line 35: Describe these states with comments.

1. What happens with sheet in each of these states.
2. Which states are stable, which are transition only (and between what other states).

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetAnimatorFactory.kt
```diff
@@ -0,0 +1,70 @@
...
+    val animationDuration = 250L
+
+    internal fun createEnterAnimator(
+        view: View,
+        isInterrupting: Boolean = false,
+    ): Animator {
+        val startY = if (isInterrupting) view.translationY else view.height.toFloat()
+        val startAlpha = if (isInterrupting) dimmingManager.dimmingViewAlpha else 0f
+
+        view.translationY = startY
```
kkafar: In situation where this is a factory method, it should not modify any state on the view immediately. 

Instead, maybe you can apply this translation in `onAnimationStart`? Won't it work?

kkafar (commented):
Okay, this looks good. I have just a request for some additional comments & I think we're good.

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetPresentationState.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4259/changes/31e696457b8ba7f0e435113c0ad9d7ac7801054c

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetAnimatorFactory.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4259/changes/56341515f77426b6ef12e092de7ba8bf719315bc

kkafar (approved):
Looks good!

t0maboro:
linter passed locally, e2e not affected, merging to unstack the next PR

PR #4262, opened 2026-07-03
refactor(Android, FormSheet v5): Introduce invalidation flags for updates

t0maboro:
## Description

Cleanup the `applyConfig` method, which now reevaluates what changed while reopening the sheet. All appearance-related and behavior-related props should now be considered for reapplying their configuration while sheet is reopening to ensure the initial state is consistent between presentation cycles.

## Changes

- refactored `applyConfig`

N/A - refactor

## Test plan

N/A - refactor

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
@@ -14,8 +14,6 @@ class FormSheetDialogManager(
 ) {
     private var formSheetConfig = FormSheetConfig()
 
-    internal val invalidationFlags = FormSheetInvalidationFlags()
-
     private val themedContext =
         ContextThemeWrapper(
             context,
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
```diff
@@ -62,29 +62,57 @@ class FormSheetDialogManager(
...
+            updateSheetPresentation()
+        }
+    }
+
+    private fun updateSheetBehavior() {
+        dimensionsCoordinator.updateFormSheetDetents(resolveDetents(formSheetConfig.detents))
+    }
+
+    private fun updateSheetAppearance() {
+        container.setGrabberVisible(formSheetConfig.prefersGrabberVisible)
```
t0maboro: note: a dedicated AppearanceCoordinator will be introduced later, once I add support for e.g. corner radius

kmichalikk (changes requested):
Something's off. On basic functionality test, origin/main starts with detent 0.3 and here it's like 2/3 and it doesn't respond to changes on hot reload whereas on main it does.

Grabber visibility works fine.

kmichalikk:
I'm also wondering if this change here is not a bit redundant since we're not calling this anywhere else and just have flags set which is immediately followed by flush and this is the only one code path. Maybe this will grow later, then it does make sense.

t0maboro:
> Something's off. On basic functionality test, origin/main starts with detent 0.3 and here it's like 2/3 and it doesn't respond to changes on hot reload whereas on main it does.
> 
> Grabber visibility works fine.

~0.3 detent should never happen, on both main & here, most of the examples are configured with detents [0.6, 1.0] - that's how it works on my end, also haven't observed issues with fast refresh on this branch.~

Aligned internally, I misunderstood this comment, fixed now.

t0maboro:
> I'm also wondering if this change here is not a bit redundant since we're not calling this anywhere else and just have flags set which is immediately followed by flush and this is the only one code path. Maybe this will grow later, then it does make sense.

Yep, I think you're right. I tried to match the iOS approach and have props logically grouped, but here there's no need to do that on Android. I think that there might be a use case to have props grouped, e.g. to introduce some AppearanceApplicator class, but before that, it would be good to have more than one appearance prop. As for now, I'm dropping the idea of invalidation flags, and I'm proposing a cleaner logic in applyConfig: https://github.com/software-mansion/react-native-screens/pull/4262/commits/6f569d7c7a69a440784dbdd9eba788df9bb04061

kmichalikk (approved):

PR #4264, opened 2026-07-04
fix(Android): prevent Fabric layout updates from resurrecting header subviews hidden by an expanded search bar

Dwynr:
## Description

On Android (Fabric), opening the header search bar hides the sibling header subviews
natively: `SearchBarView.setToolbarElementsVisibility()` sets `View.visibility = GONE`
on every non-`SEARCH_BAR` `ScreenStackHeaderSubview`, and restores them on close.

That hidden state is imperative — React knows nothing about it — and Fabric's
`SurfaceMountingManager.updateLayout` force-syncs `View.visibility` from the ShadowTree
display type on **every layout mount item** for a view:

```kotlin
// react-native/ReactAndroid/.../fabric/mounting/SurfaceMountingManager.kt (updateLayout)
// displayType: 0 represents display: 'none'
val visibility = if (displayType == 0) View.INVISIBLE else View.VISIBLE
if (viewToUpdate.visibility != visibility) {
  viewToUpdate.visibility = visibility
}
```

So any event that changes a hidden subview's shadow layout makes the subview visible
again while the search bar is still open. The easiest real-world trigger is **rotating a
device with a display cutout**: the inset change relayouts the toolbar,
`onNativeToolbarLayout` pushes new `contentInsetStart/End` into the header-config state,
the subviews' shadow nodes re-layout, and Fabric dispatches `updateLayout` for them —
resurrecting the hidden `headerRight` subview. (Window resizes on foldables or in
split-screen can trigger the same path.)

The toolbar then lays the resurrected END-gravity subview out in the space left over by
the full-width expanded `SearchView` — i.e. at the **start edge of the header, overlapping
the back button** (see screenshots below).

Stack trace of the resurrection, captured with an instrumented
`ScreenStackHeaderSubview.setVisibility` override on RN 0.86 / RNS 4.25.2:

```
subview(type=RIGHT tag=974) visibility 8 -> 0        // GONE -> VISIBLE
  at com.swmansion.rnscreens.ScreenStackHeaderSubview.setVisibility
  at com.facebook.react.fabric.mounting.SurfaceMountingManager.updateLayout(SurfaceMountingManager.kt:865)
  at com.facebook.react.fabric.mounting.mountitems.IntBufferBatchMountItem.execute(IntBufferBatchMountItem.kt:139)
  at com.facebook.react.fabric.mounting.MountItemDispatcher.executeOrEnqueue(MountItemDispatcher.kt:379)
  ...
```

`dumpsys activity top` confirms it is the same view instance flipping visibility (no
remount), and that the misplaced position is the toolbar's own layout of the
now-visible child, not a Fabric frame:

```
search open, before rotation:   ScreenStackHeaderSubview{63e252d G.E......  980,186-1136,294 #7b0}
after rotate to landscape:      ScreenStackHeaderSubview{63e252d V.E......  196,186- 352,294 #7b0}
after rotate back to portrait:  ScreenStackHeaderSubview{63e252d V.E......   60,186- 216,294 #7b0}
```

The bug reproduces deterministically on cutout devices (reported on a Pixel 9a;
reproduced on a Pixel 9 Pro API 36 emulator with the punch-hole cutout overlay). It does
**not** reproduce on cutout-less emulators — the header metrics come out identical after
the rotation round-trip, so no `updateLayout` is dispatched for the hidden subviews —
which is likely why it has gone unnoticed.

## Changes

- `ScreenStackHeaderSubview`: new internal pinned state `isHiddenBySearchBar`;
  `setVisibility` ignores attempts to un-hide the view while pinned (Fabric's
  force-sync in particular) — `setHiddenBySearchBar(false)` is the only way back to
  visible.
- `SearchBarView.setToolbarElementsVisibility`: routes the open/close visibility toggle
  through the pin instead of writing raw `visibility`.
- `ScreenStackHeaderConfig.onUpdate()`: subviews attached or re-attached while the
  search action view is already expanded inherit the hidden state (SearchBarView only
  applies it at open/close time); a stale pin from a previous search session is cleared.

Portrait → landscape → portrait with the header search bar open, on a device with a
display cutout (`headerSearchBarOptions` + a custom `headerRight`):

| Before | After |
| --- | --- |
| upstream-rns-searchbar-before| upstream-rns-searchbar-after|

Before: the `headerRight` item ("⋯") reappears at the top-left, overlapping the back
button, while the search bar is still open. After: the subview stays hidden; closing the
search restores it at its correct END position.

## Test plan

Reproduction (deterministic, any Android emulator):

1. Simulate a display cutout: `adb shell cmd overlay enable com.android.internal.display.cutout.emulation.hole`
   (and make sure auto-rotate is on: `adb shell settings put system accelerometer_rotation 1`)
2. Open a native-stack screen that sets `headerSearchBarOptions` **and** a custom `headerRight`
3. Tap the search icon to expand the search bar (the headerRight item disappears — expected)
4. Rotate to landscape, then back to portrait
5. Before this fix: the headerRight item is visible at the top-left of the header while
   the search bar is still open. With the fix: it stays hidden.

Regression checks performed (Pixel 9 Pro API 36 emulator, RN 0.86, Fabric):

- open/close search still hides/restores all header items, at correct positions
- the restored headerRight item stays functional (menu opens, correctly anchored)
- rotation round-trip with search open on a cutout-less device: unchanged
- subview visibility verified at the native level via `dumpsys activity top`
  (stays `G.E` through the rotation round-trip with the fix)

## Checklist

- [ ] For API changes, updated relevant public types. *(no public API changes)*
- [x] Ensured that CI passes

*Friendly disclosure: this fix was developed with AI assistance (Claude Fable 5) - the root-cause
trace (instrumented `setVisibility`/`setLayoutParams` stack captures), the patch, and the emulator
verification were AI-driven, fully human-reviewed and human-tested before submission.*


kligarski (approved):
Hi, @Dwynr. Thanks for the PR!

The changes look good to me.

We're only missing a reproduction test screen that we use to test the change and check for regressions in the future (they live in `apps/src/tests/issue-tests`). If it will be possible, I'll push the commit with the test screen to your branch. I hope you don't mind.

Dwynr:
> Hi, @Dwynr. Thanks for the PR!
> 
> The changes look good to me.
> 
> We're only missing a reproduction test screen that we use to test the change and check for regressions in the future (they live in `apps/src/tests/issue-tests`). If it will be possible, I'll push the commit with the test screen to your branch. I hope you don't mind.

No that's fine, thanks for taking a look. Looking forward to remove our custom patch from our app in the future.

PR #4286, opened 2026-07-08
fix(iOS, Stack v4): fix header positioning in shadow tree

kligarski:
## Description

Fixes positioning of header confing in Shadow Tree in Stack v4 on iOS.

Fixes Pressables in transparent header.

> [!WARNING]
> 
> Pressables in the screen's content in landscape orientation when opaque header is used still don't work. 
> Here is the ticket: https://github.com/software-mansion/react-native-screens-labs/issues/1622.
> 
> https://github.com/user-attachments/assets/f80cbba0-ff49-4e0f-9667-70a66d8a6d74

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1594.

### Details

On Android, Pressables in the header work correctly after https://github.com/software-mansion/react-native-screens/pull/3239. In transparent header, we're skipping applying any offset and it solves the problem.

android_v4_hc

On iOS, transparent header wasn't handled in any other way - we would still move the header config above the screen which is incorrect (the HC is moved off the screen completely). The correct value of the offset is always the relative position of header config in relation to the screen. It will adapt automatically to every situation (instead of relying on assumption that the toolbar/navigation bar is always just above the screen/just at the top of the screen). Same approach is/will be used in Stack v5.

ios_v4_hc

The difference in platforms comes down to handling the top inset. On Android, the toolbar applies padding therefore the inset is included in its height. On iOS, navigation bar is offset by the inset value therefore its height doesn't contain the offset.

## Changes

- send frame origin via header config state
- calculate frame origin based on position of navigation bar in relation to screen
- apply iOS 26 landscape `edgesForExtendedLayout` correction when header is opaque
- apply frame origin in shadow node

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/7bb654ee-fba3-418d-9965-3cd78ade53a0" /> | <video src="https://github.com/user-attachments/assets/6564cc67-f661-48f1-9235-15b7edec4977" /> |

## Test plan

- `Test3446` + `headerTransparent: true` - test {portrait,landscape}x{ios18,ios26}x{opaque,transparent}
- `Test3111`

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderConfigShadowNode.cpp
+++ b/common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderConfigShadowNode.cpp
@@ -26,7 +26,8 @@ void RNSScreenStackHeaderConfigShadowNode::applyFrameCorrections() {
   ensureUnsealed();
 
   const auto &stateData = getStateData();
-  layoutMetrics_.frame.origin.y = -stateData.frameSize.height;
+  layoutMetrics_.frame.origin.x = stateData.frameOrigin.x;
+  layoutMetrics_.frame.origin.y = stateData.frameOrigin.y;
 }
 
 #if !defined(ANDROID)
--- a/common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderConfigShadowNode.cpp
+++ b/common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderConfigShadowNode.cpp
@@ -26,8 +26,12 @@ void RNSScreenStackHeaderConfigShadowNode::applyFrameCorrections() {
   ensureUnsealed();
 
   const auto &stateData = getStateData();
+#if defined(ANDROID)
+  layoutMetrics_.frame.origin.y = -stateData.frameSize.height;
+#else // defined(ANDROID)
   layoutMetrics_.frame.origin.x = stateData.frameOrigin.x;
   layoutMetrics_.frame.origin.y = stateData.frameOrigin.y;
+#endif // defined(ANDROID)
 }
 
 #if !defined(ANDROID)
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -69,7 +69,7 @@ static const NSNumber *const DEFAULT_TITLE_LARGE_FONT_SIZE = @34;
     _show = YES;
     _translucent = NO;
     _addedReactSubviewsInCurrentTransaction = false;
-    _lastSendState = react::RNSScreenStackHeaderConfigState(react::Size{}, react::EdgeInsets{});
+    _lastSendState = react::RNSScreenStackHeaderConfigState(react::Size{}, react::EdgeInsets{}, react::Point{});
     [self initProps];
   }
   return self;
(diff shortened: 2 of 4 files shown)


ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -190,8 +193,20 @@ - (void)updateHeaderStateInShadowTreeInContextOfNavigationBar:(nullable UINaviga
...
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
+  if (!_translucent && self.shouldHeaderBeVisible) {
+    navBarFrameInScreenView.origin.y -= _screenView.safeAreaInsets.top;
+  }
+#endif // RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
+
   [self updateShadowStateWithSize:navigationBar.frame.size
-                       edgeInsets:[self computeEdgeInsetsOfNavigationBar:navigationBar]];
+                       edgeInsets:[self computeEdgeInsetsOfNavigationBar:navigationBar]
+                      frameOrigin:navBarFrameInScreenView.origin];
```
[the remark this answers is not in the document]
kligarski: In old implementation I don't want to change anything unless it's absolutely necessary. Everything seems to be working now.

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -190,8 +193,20 @@ - (void)updateHeaderStateInShadowTreeInContextOfNavigationBar:(nullable UINaviga
     return;
   }
 
+  CGRect navBarFrameInScreenView = [navigationBar convertRect:navigationBar.bounds toView:_screenView];
+
+  // On iOS 26+ in landscape mode there is a bug with `edgesForExtendedLayout` for non-transparent
+  // header. We're applying SAV to fix it but we need to take it into account for HeaderConfig's
+  // origin as well.
```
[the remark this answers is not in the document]
kligarski: we're applying SafeAreaView component in JS so it's correct.

kmichalikk (approved):

kkafar (changes requested):
The PR description misses explanation of why the bug was happening.

common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderConfigShadowNode.cpp
```diff
@@ -26,7 +26,12 @@ void RNSScreenStackHeaderConfigShadowNode::applyFrameCorrections() {
   ensureUnsealed();
 
   const auto &stateData = getStateData();
+#if defined(ANDROID)
   layoutMetrics_.frame.origin.y = -stateData.frameSize.height;
+#else // defined(ANDROID)
+  layoutMetrics_.frame.origin.x = stateData.frameOrigin.x;
+  layoutMetrics_.frame.origin.y = stateData.frameOrigin.y;
+#endif // defined(ANDROID)
```
kkafar, line 34: Is this problem iOS specific? Why does it differ between Android & iOS?

kkafar: Basically that is a request for error mechanism explanation in PR description.

ios/RNSScreenStackHeaderConfig.mm
```diff
@@ -167,13 +167,16 @@ - (void)layoutNavigationControllerView
...
+                       edgeInsets:(NSDirectionalEdgeInsets)edgeInsets
+                      frameOrigin:(CGPoint)frameOrigin
 {
   // I believe Yoga handles RTL internally & .left will be treated as .right in RTL etc.
   react::EdgeInsets convertedEdgeInsets{
       .left = edgeInsets.leading, .top = edgeInsets.top, .right = edgeInsets.trailing, .bottom = edgeInsets.bottom};
   react::Size convertedSize = RCTSizeFromCGSize(size);
-  auto newState = react::RNSScreenStackHeaderConfigState(convertedSize, convertedEdgeInsets);
+  auto newState =
+      react::RNSScreenStackHeaderConfigState(convertedSize, convertedEdgeInsets, RCTPointFromCGPoint(frameOrigin));
```
kkafar: I mean, just for sake of keeping method convention I'd suggest either inlining the `convertedSize` or creating `react::Point convertedOrigin`. Right now it's smelly.

kkafar (commented):
Okay, beside putting explanation in PR description & some code style everything looks good.

ios/RNSScreenStackHeaderConfig.mm
kligarski: https://github.com/software-mansion/react-native-screens/pull/4286/commits/b3a30eeb951e479984da2d18fdfb0035de09bf83

common/cpp/react/renderer/components/rnscreens/RNSScreenStackHeaderConfigShadowNode.cpp
kligarski: I updated the PR description.

kkafar (approved):
Let's roll. Thank you.

PR #4292, opened 2026-07-08
feat(Android, FormSheet v5): Add `onDetentChanged` support

t0maboro:
## Description

This PR introduces support for the `onDetentChanged` event in the FormSheet. It reports the index of the newly settled detent.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1561

## Changes

- Android's BottomSheetBehavior operates on strict states, so the dynamic mapping in `FormSheetBehaviorController` safely translates these states into the corresponding detents array index depending on the current detent count configuration.
- Bound `FormSheetDetentChangedEvent` with JS

https://github.com/user-attachments/assets/bb44e052-9441-463b-95bc-23fba145d5b9

## Test plan

Marked SFT as applicable on Android.

## Checklist

- [ ] Ensured that CI passes


android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetBehaviorController.kt
```diff
@@ -1,13 +1,45 @@
...
 import com.google.android.material.bottomsheet.BottomSheetBehavior
 
 internal class FormSheetBehaviorController(
     private val sheetView: FrameLayout,
+    private val onDetentChanged: ((index: Int) -> Unit)? = null,
 ) {
     private val behavior = BottomSheetBehavior.from(sheetView)
 
+    private var currentDetentsCount: Int = 1
+    private var lastEmittedDetentIndex: Int = -1
```
kligarski: nit: maybe it would be nice to define a const for -1 so that we know the interpretation of this value?

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
```diff
@@ -51,17 +52,20 @@ class FormSheetDialogManager(
             dialog = dialog,
             bottomSheetView = bottomSheetView,
             dimmingManager = dimmingManager,
-            onNativeDismiss = onDismissRequest,
+            onNativeDismiss = { eventEmitter?.emitOnNativeDismissEvent() },
         )
 
-    internal var appearanceEventEmitter: ViewAppearanceEventEmitter? by presentationManager::appearanceEventEmitter
+    internal var eventEmitter: FormSheetDialogEventEmitter? by Delegates.observable(null) { _, _, newValue ->
```
kligarski, line 58: I'm 50/50 about storing `FormSheetDialogEventEmitter` inside `presentationManager.appearanceEventEmitter` which is `ViewAppearanceEventEmitter` but it works so it's fine I guess.

kligarski (approved):
There are some problems when changing orientation but they might be handled in separate PR with other orientation problems.

https://github.com/user-attachments/assets/99b8c08a-e502-4463-a967-b84440e6a705

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetDialogManager.kt
t0maboro: By passing it as a `ViewAppearanceEventEmitter` I can keep a single event emitter instance and have the presentation manager be aware of the lifecycle events only.

android/src/main/java/com/swmansion/rnscreens/gamma/modals/formsheet/FormSheetBehaviorController.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4292/commits/0cf2dd9c73e40278ab3d1c2a5d38cded93ba0a58

PR #4305, opened 2026-07-09
chore(test): add stack lifecycle events screen test with scenario

LKuchno:
## Description

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1607

Adds a new single-feature test screen exercising the gamma `StackContainer` screen lifecycle callbacks — `onWillAppear`, `onDidAppear`, `onWillDisappear`, `onDidDisappear` — across push, pop, and a nested stack. It mirrors the existing `test-tabs-lifecycle-events` screen so stack and tabs lifecycle behaviour can be checked with the same toast-based visualization.

Each transition pushes a color-coded toast (`onWillAppear` - green, `onWillDisappear` - light navy, `onDidAppear` - light blue, `onDidDisappear` — dark navy) so the fired events and their order are visible on device.

## Changes

- `test-stack-lifecycle-events/index.tsx` — test screen: an outer stack (`Home → A → NestedStack`) and a nested stack (`NestedHome → NestedA`). A shared `useMakeLifecycleCallbacks` hook attaches the four lifecycle callbacks to every route. Every route sets a header `title`; the `NestedStack` route's header is intentionally shown, so on iOS the outer `NestedStack` header and the inner nested-stack header stack together — letting the scenario exercise popping the whole container via the **outer** `NestedStack` header back button as well as the inner one.
- `test-stack-lifecycle-events/scenario.md` — manual scenario covering push/pop (Pop button / header back button / native gesture / system back) at the top level and inside the nested stack, plus crossing the nested-stack boundary — popping the whole container back to `Home` from both `NestedHome` and `NestedA` via the outer `NestedStack` header back button. Documents per-platform event orderings, including the iOS **outer-first** vs Android **inner-first** difference on the container pop.
- `test-stack-lifecycle-events/scenario-description.ts` — scenario metadata (`e2eCoverage: 'tbd'`, platforms iOS + Android).
- `stack-v5/index.ts` — registers the scenario in the Stack v5 group.

### Note:
This test will have to be updated if https://github.com/software-mansion/react-native-screens-labs/issues/1649 is fixed, and the changes mentioned by @kligarski in https://github.com/software-mansion/react-native-screens/pull/4305#discussion_r3570504855 are implemented.


apps/src/tests/single-feature-tests/stack-v5/test-stack-lifecycle-events/scenario.md
```diff
@@ -0,0 +1,301 @@
...
+
+- **iOS and Android fire a fundamentally different set of events**, so verify
+  each platform against its own column below.
+
+  **Android** — only the screen entering or leaving the stack fires; the
+  screen underneath stays silent:
+  - **Push (Y pushed over X):** only the pushed screen fires —
+    `Y: onWillAppear`, then `Y: onDidAppear`. X fires nothing.
+  - **Pop (Y popped, back to X):** only the popped screen fires —
+    `Y: onWillDisappear`, then `Y: onDidDisappear`. X fires nothing.
```
kligarski, line 56: This should change soon when we handle `detach` and `attach` in fragment manager. Not sure if we should merge the test in its current form or wait for the changes?

cc @kkafar

apps/src/tests/single-feature-tests/stack-v5/test-stack-lifecycle-events/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+  return (
+    <StackContainer
+      routeConfigs={[
+        {
+          name: 'NestedHome',
+          Component: NestedHomeScreen,
+          options: {
+            ...makeCallbacks('NestedHome'),
+          },
```
kligarski, line 128: tbh, for consistency I'd use the header on all screens (even though it looks odd, especially now before we've handled the doubled top inset) - then it's clear that the screens are nested. But that's not a strong opinion, up to you.

apps/src/tests/single-feature-tests/stack-v5/test-stack-lifecycle-events/scenario.md
```diff
@@ -0,0 +1,301 @@
...
+- [ ] Screen **Home** is shown again. Because `NestedHome` is the nested
+  stack's only (root) screen, the Pop button pops the whole **NestedStack**
+  container rather than a screen within it, so the same container-pop toasts
+  appear as in step 11 (iOS: six; Android: four).
+
+  The Pop button and the native gesture (step 11) produce the identical
+  container-pop event set. There is **no header back button** at this
+  boundary — the `NestedStack` route's header is hidden and `NestedHome` is
+  the inner root — so the container pop is exercised only via these two
+  methods.
```
kligarski: Actually, in the current version there is a header back button because `NestedHome`'s header is hidden and `NestedStack`'s header is visible.

Image

Same applies to second-to-last note.

apps/src/tests/single-feature-tests/stack-v5/test-stack-lifecycle-events/index.tsx
LKuchno: https://github.com/software-mansion/react-native-screens/pull/4305/changes/ea3c538efbc58960c6a22685658904b08c17bf01

apps/src/tests/single-feature-tests/stack-v5/test-stack-lifecycle-events/scenario.md
LKuchno: https://github.com/software-mansion/react-native-screens/pull/4305/changes/cc47599c74f14a2690ac1b8d63f911ded663c1a6

kligarski (approved):

PR #4309, opened 2026-07-10
fix(iOS, Split): Fix ShadowState inside SplitView columns after transition

t0maboro:
## Description

Fixes an issue where `Pressable` components inside a newly revealed column (e.g., the Primary column) in `UISplitViewController` fail to register touches correctly on Fabric.

When attempting to interact with the newly revealed column, the touches were immediately cancelled.

Note: I don't know the exact reason why KVO for the frame hasn't detected that change. For now, I am not investigating it further, if any other problem with desynchronization occurs, I'll consider researching a different approach.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1636

## Changes

- Registered on `transitionCooridnator` completion callback to ensure that the shadow state is synchronized after the display mode change.

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/bceb5215-0550-421c-bcc5-47f077bf5384" /> | <video src="https://github.com/user-attachments/assets/f8f32d3a-0554-42d8-852b-c818257de529" /> |

## Test plan

Added dedicated SFT for pressables.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/apps/App.tsx
+++ b/apps/App.tsx
@@ -1,9 +1,9 @@
 import React from 'react';
 import { enableFreeze } from 'react-native-screens';
 
-// import Example from './Example';
+import Example from './Example';
 
-import { TestSplitPressables as Example } from './src/tests/single-feature-tests';
+// import { TestTabsSimpleNav as Example } from './src/tests/single-feature-tests';
 // import { TestTabsInStackStableEnterTransition as Example } from './src/tests/component-integration-tests';
 // import { Test42 as Example } from './src/tests/issue-tests';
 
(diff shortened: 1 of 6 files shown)


apps/App.tsx
kligarski, line 1: this should be reverted

kligarski (approved):
The bug seems to be fixed, looks good. We should just revert `App.tsx`.

apps/App.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4309/commits/0acbfe4781d71e5aed7ccc5a9ca46b857917bd3f

PR #4318, opened 2026-07-13
chore(test): add script to create test screen directories

LKuchno:
## Summary

Adds `scripts/create-test-directory.cts`, a Node helper that scaffolds a new test screen directory under `apps/src/tests/<category>/<domain>/` with the three files every test screen needs.

Running it creates `apps/src/tests/<category>/<domain>/<test-dir-name>/` containing:
- `index.tsx` — empty, ready for the screen implementation
- `scenario-description.ts` — generated from the standard shape with `name` and `key` filled in
- `scenario.md` — copied from `apps/src/tests/template/scenario.md` (placeholders left intact)

The `<domain>` directory is created automatically if it doesn't exist yet.

## Usage

```bash
node scripts/create-test-directory.cts <test-dir-name> <screen-name> <domain> <category>

# Example
node scripts/create-test-directory.cts \
  test-form-sheet-grabber-visible \
  "Grabber visible" \
  form-sheet \
  single-feature-tests
```

Arguments:

| Argument | Meaning |
| --- | --- |
| `test-dir-name` | Directory name for the new test (e.g. `test-form-sheet-grabber-visible`). Also used as the `key` in `scenario-description.ts`. |
| `screen-name` | Human-readable screen name (e.g. `"Grabber visible"`). Used as the `name` in `scenario-description.ts`. Quote if it contains spaces. |
| `domain` | Domain sub-directory (e.g. `form-sheet`). Created if missing. |
| `category` | Category directory under `apps/src/tests` (e.g. `single-feature-tests`). |

## Details

- Written as a `.cts` script run directly by Node via strip-only TypeScript support, so no build step or `ts-node` is needed. That mode erases types without rewriting syntax, hence the CommonJS `require` calls instead of `import`.
- Validates arguments (expects exactly 4) and prints usage on `-h`/`--help` or misuse.
- Fails fast with clear errors if the category directory doesn't exist (listing the valid categories), if the `scenario.md` template is missing, or if the target directory already exists.
- Resolves paths relative to the repo root, so it works from any working directory.
- Prints a summary of what was created on success.


kkafar (changes requested):
I applaud the idea. Can I ask to migrate this script to TypeScript (runnable with node)? Use `.cts` file extension for proper TypeScript and CommonJS support. 

## Why

Maintainability. I know that we already have multiple bash scripts present - we should migrate all of them that are above just a few lines of bash code.

kkafar (approved):
Good starting point. We'll patch it as pain points emerge. Thanks.

PR #4321, opened 2026-07-13
fix(iOS, Stack v4): Prevent accounting header in contentHeight when it's transparent

### LINKED ISSUE #4276: iOS formSheet fitToContents includes header height when header is transparent
### Description

When using a `formSheet` with `sheetAllowedDetents: 'fitToContents'` on iOS, setting the native header to transparent appears to leave extra unused height in the sheet.

With an opaque header, the sheet fits the header plus content. With a transparent header, the header overlays the content as expected, but the sheet still appears to reserve the native header height, leaving dead space below the content.

Image

Image

> In the transparent example you can see space below elements which is `inset + header`

**Native context**:

Looking at the iOS implementation, `fitToContents` is based on the `RNSScreenContentWrapper` frame height, then adjusted by `contentHeightErrata`.

That errata is accumulated while walking parent view controllers and adds the height of visible `RNSNavigationController.navigationBar` instances. The check appears to account for `navigationBar.isHidden`, but not whether the header is transparent.

### Steps to reproduce

1. Clone and run the reproduction repository.
2. Open the "Opaque header" sheet.
3. Open the "Transparent header" sheet.
4. Open the "Transparent + header padding" sheet.
5. Compare the sheet content and the empty space below the last row.

https://github.com/fabriziocucci/sheet-header-transparent-repro

### Screens version

4.25.2

### React Native version

0.86.0

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

iPhone 17 Pro Max simulator, iOS 26.2, Xcode 26.2

### Acknowledgements

Yes

t0maboro:
## Description

> [!IMPORTANT]  
> As stack nesting (native header creates nested stack) wasn't planned for FormSheets in v4, I am opening this PR for further discussion whether we should consider applying this fix to v4 or focus on making it work properly in v5.

For 'fitToContents', the sheet's detent height is computed as:
```
detentHeight = reactFrame.size.height + contentHeightErrata
```
`contentHeightErrata` is accumulated while walking up the ViewController tree, where the navigation bar's height was unconditionally added whenever the bar was visible.
In `RNSScreenContentWrapper.mm`, I gate the navbar-height addition to the visible screen's header not being transparent.

Closes: https://github.com/software-mansion/react-native-screens/issues/4276

## Changes

- prevent calculating headerHeightErrata when the header is transparent

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/f32a48f6-deeb-44c8-9c31-8e417d162103" /> | <video src="https://github.com/user-attachments/assets/24b06607-5a5c-4263-bf2d-42798a5d772e" /> |

## Test plan

Added a dedicated example.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/RNSScreenContentWrapper.mm
+++ b/ios/RNSScreenContentWrapper.mm
@@ -56,7 +56,7 @@ namespace react = facebook::react;
       UINavigationBar *navigationBar = navigationController.navigationBar;
 
       RNSScreenStackHeaderConfig *headerConfig = [self headerConfigForNavigationController:navigationController];
-      if (!headerConfig.translucent) {
+      if (headerConfig != nil && !headerConfig.translucent) {
         headerHeightErrata += navigationBar.frame.size.height * !navigationBar.isHidden;
       }
     }
--- a/ios/RNSScreenContentWrapper.mm
+++ b/ios/RNSScreenContentWrapper.mm
@@ -10,6 +10,7 @@
 #import "RNSSafeAreaViewComponentView.h"
 #import "RNSScreen.h"
 #import "RNSScreenStack.h"
+#import "RNSScreenStackHeaderConfig.h"
 
 namespace react = facebook::react;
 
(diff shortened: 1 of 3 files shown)


ios/RNSScreenContentWrapper.mm
```diff
@@ -51,14 +52,29 @@ - (void)attachToAncestorScreenViewStartingFrom:(nonnull RNSScreen *)screenCtrl
         break;
       }
     } else if ([controller isKindOfClass:RNSNavigationController.class]) {
-      UINavigationBar *navigationBar = static_cast<RNSNavigationController *>(controller).navigationBar;
-      headerHeightErrata += navigationBar.frame.size.height * !navigationBar.isHidden;
+      RNSNavigationController *navigationController = static_cast<RNSNavigationController *>(controller);
+      UINavigationBar *navigationBar = navigationController.navigationBar;
+
+      RNSScreenStackHeaderConfig *headerConfig = [self headerConfigForNavigationController:navigationController];
+      if (!headerConfig.translucent) {
```
kligarski: Maybe add a nil check just to be sure? On the other hand, if we have `RNSNavigationController`, there should be `headerConfig`.

```suggestion
      if (headerConfig != nil && !headerConfig.translucent) {
```

kligarski (approved):
Seems reasonable to me.

PR #4220, opened 2026-06-26
feat(Android, Stack v4): add `disable*InsetApplication` props

sgaczol:
## Description

This PR adds per-edge control over header inset application on Android. Previously only the
top inset could be opted out, via the already-existing `disableTopInsetApplication` prop.
This PR introduces three new sibling props - `disableLeftInsetApplication`,
`disableRightInsetApplication` and `disableBottomInsetApplication` — so that an app can disable
inset handling on any edge of the native header.

The top inset is a single shared space at the top of the screen, so in nested stacks it must
be coordinated - only the topmost visible header consumes it (to avoid doubled spacing). The top
inset is additionally anchored to the decor view in `applyDecorViewTopInsetIfNeeded()`, so it
survives a custom `SafeAreaView` consuming the incoming `WindowInsets` somewhere in the subtree.
Left/right/bottom are per-edge: each header applies them on its own edges, so there is nothing
to coordinate, and — unlike the top — they have no decor-view fallback. They are read solely from
the incoming `WindowInsets` and applied per-header, which is why a `SafeAreaView` in the subtree
can consume them per-subtree. The context only propagates the opt-out down the subtree (set the
prop once on the outer stack, inner headers inherit it).

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1535

## Changes

- Added public props `disableLeftInsetApplication`, `disableRightInsetApplication` and
  `disableBottomInsetApplication` to `ScreenStackHeaderConfigProps` (`src/types.tsx`)
- Added `consumeLeftInset` / `consumeRightInset` / `consumeBottomInset` to the native component
  spec (`ScreenStackHeaderConfigNativeComponent.ts`) and the corresponding setters on the Android
  view manager / config (`ScreenStackHeaderConfigViewManager.kt`, `ScreenStackHeaderConfig.kt`)
- Introduced `EdgeInsetApplicationContext` carrying the left/right/bottom opt-out down the subtree, and wired the providers in `ScreenStackItem.tsx` and the consumer in `ScreenStackHeaderConfig.tsx`
- Reworked the inset computation in `CustomToolbar.kt`: each edge (top/left/right/bottom) is now
  evaluated independently
- Added an issue test

## After - visual documentation

https://github.com/user-attachments/assets/1f747d3b-00d2-46f1-9058-b0bce3f02d61

https://github.com/user-attachments/assets/279e1b58-429d-4183-947a-ffa68e4129bb

[...]

PR #4322, opened 2026-07-13
fix(Android, Stack v4): default consume*Inset props to true

patrycjakalinska:
Follow-up to #4220 . The `consume*Inset` props on `ScreenStackHeaderConfig` are declared without a `WithDefault`, so Fabric codegen defaults them to `false`, while the Android native view manager (`ScreenStackHeaderConfig.kt`) defaults them to `true` (`Delegates.observable(true)`).

under prop-update reconciliation (`enablePropsUpdateReconciliationAndroid`) a prop equal to the codegen default is ommited on initial mount, so a `disable*InsetApplication` opt-out is dropped

### DIFF
--- a/src/fabric/ScreenStackHeaderConfigNativeComponent.ts
+++ b/src/fabric/ScreenStackHeaderConfigNativeComponent.ts
@@ -84,7 +84,7 @@ export interface NativeProps extends ViewProps {
 
   // Experimental
   userInterfaceStyle?: CT.WithDefault<UserInterfaceStyle, 'unspecified'>;
-  consumeTopInset?: CT.WithDefault<boolean, true>;
+  consumeTopInset?: CT.WithDefault<boolean, false>;
   consumeLeftInset?: CT.WithDefault<boolean, true>;
   consumeRightInset?: CT.WithDefault<boolean, true>;
   consumeBottomInset?: CT.WithDefault<boolean, true>;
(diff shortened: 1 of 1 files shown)


src/fabric/ScreenStackHeaderConfigNativeComponent.ts
```diff
@@ -84,10 +84,10 @@ export interface NativeProps extends ViewProps {
 
   // Experimental
   userInterfaceStyle?: CT.WithDefault<UserInterfaceStyle, 'unspecified'>;
-  consumeTopInset?: boolean | undefined;
-  consumeLeftInset?: boolean | undefined;
-  consumeRightInset?: boolean | undefined;
-  consumeBottomInset?: boolean | undefined;
+  consumeTopInset?: CT.WithDefault<boolean, true>;
```
t0maboro: Top inset consumption has a different path from others and is set by default to `false` natively. Shouldn't we align that?

t0maboro (commented):
makes sense, leaving 1 comment to align and we're good to go

src/fabric/ScreenStackHeaderConfigNativeComponent.ts
patrycjakalinska: @t0maboro yep, sorry! tested and fixed!

t0maboro (approved):

PR #4332, opened 2026-07-14
feat(Android, FormSheet v5): Add support for preventNativeDismiss

t0maboro:
## Description

Adding the `preventNativeDismiss` support on Android together with `onNativeDismissPrevented` event. When `preventNativeDismiss` is enabled, all native dismissal paths - tapping the backdrop, swiping the sheet down, and pressing the hardware back button - are intercepted. Instead of closing, the sheet stays in the latest stable (visible) state and fires an event that 3p developer might handle. Programmatic dismissal keeps working as before.
When the prop is disabled, native dismissal behaves as it did previously.

Also, I needed to refine the approach for native dismissal handling. The "test-form-sheet-presentation-state" had a regression that was dismissing the FormSheet again after the quick state toggle. The extra dismiss came from Material's `BottomSheetDialog`, which installs an internal callback that calls `cancel()` whenever the sheet reaches `STATE_HIDDEN`. Our own exit animation ends by setting `STATE_HIDDEN`. Previously, `cancel()` was not overridden, so it was going through the `Dialog.cancel()`, which dispatched the `OnCancelListener` **asynchronously**. During a rapid dismiss-present toggle:
1. `setIsOpen(false)` caused the presentation state update to `DISMISSING` and the exit animation starts.
2. `setIsOpen(true)` calls `presentIfNeeded()` is seeing the `DISMISSING` state, so it returns early and just records `targetIsOpen = true`.
3. The exit animation ends; behavior state is set to `STATE_HIDDEN`, so Material calls `cancel()` internally and posts the `OnCancelListener` message.
4. In the meantime, `onDismissComplete` changes the presentation state to `DISMISSED` but our `targetIsOpen` is set to `true`, so the next FormSheet presentation is started, the presentation state updates to `PRESENTING`.
5. When the Looper picks the queued `OnCancelListener` message, it finally runs `handleNativeDismiss()`, so the state will flip to`DISMISSING/DISMISSED` after the presentation completes.

With this PR, I am introducing the cancel interceptor for preventNativeDismiss, which also runs `handleNativeDismiss` **synchronously** at the `STATE_HIDDEN` trigger, preventing the race condition where someone has opened the sheet again, while the **asynchronous** `handleNativeDismiss` callback hasn't been processed yet.

> [!CAUTION]
> I started noticing an occasional content flickering, probably because of wrong timing of the translateY application before starting the enter transition. Noticed that it's reproducible on the main branch as well, so I'll handle that separately: https://github.com/software-mansion/react-native-screens-labs/issues/1689

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1563
Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1669

## Changes

- Added a `CancelRequestInterceptor` to override `cancel()`; when an interceptor is attached, it fully owns the
  outcome and decides whether to allow or prevent the dismissal request.
- `FormSheetNativeDismissCoordinator` was integrated with that interceptor. It routes backdrop taps and swipe-down and hardware back button (via`OnBackPressedCallback`) through the same path, and decides about the outcome. When prevented, emit an event and restore the sheet's last stable detent. Otherwise, drive the animated dismissal via the presentation manager.

https://github.com/user-attachments/assets/38ca81f4-32ee-4f5c-a655-633df71ecff0

## Test plan

Marked SFT as applicable on Android

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/modals/formsheet/FormSheetBehaviorController.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/modals/formsheet/FormSheetBehaviorController.kt
@@ -12,6 +12,7 @@ internal class FormSheetBehaviorController(
 
     private var currentDetentsCount: Int = 1
     private var lastEmittedDetentIndex: Int = FORM_SHEET_UNKNOWN_DETENT_INDEX
+    private var lastStableState: Int = BottomSheetBehavior.STATE_COLLAPSED
 
     private val bottomSheetCallback =
         object : BottomSheetBehavior.BottomSheetCallback() {
@@ -19,6 +20,8 @@ internal class FormSheetBehaviorController(
                 bottomSheet: View,
                 newState: Int,
             ) {
+                rememberStateIfStable(newState)
+
                 val index = mapStateToDetentIndex(newState)
                 if (index != FORM_SHEET_UNKNOWN_DETENT_INDEX && index != lastEmittedDetentIndex) {
                     lastEmittedDetentIndex = index
@@ -34,6 +37,7 @@ internal class FormSheetBehaviorController(
 
     init {
         behavior.isHideable = true
+        rememberStateIfStable(behavior.state)
     }
 
     internal fun setup() {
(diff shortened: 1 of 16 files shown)


kligarski (approved):
LGTM

PR #4334, opened 2026-07-15
refactor(test): render Toast above the system navigation bar

LKuchno:
## Description

The `ToastProvider` used in the example apps wrapped the **whole view hierarchy** in a `SafeAreaView` (provided in https://github.com/software-mansion/react-native-screens/pull/3944), which applied the bottom inset to the entire app and pushed the content (e.g. a tab bar) up — leaving an ugly gap above the system navigation bar on Android.

This reworks the provider to render the toasts in an absolutely positioned **sibling overlay** instead of padding the whole app, as suggested in the issue. It also adds an `anchorSide` prop so a screen can choose whether the toast stack sits at the top or the bottom of the screen.

Closes software-mansion/react-native-screens-labs#1657

## Changes

- Render `children` directly (no longer wrapped in a `SafeAreaView`), so the app is no longer padded and the tab bar sits flush above the navigation bar — no gap.
- Extract a `ToastContainer` component that renders the toasts in a full-screen `box-none` overlay `View` (touches pass straight through to the app). The overlay bottom- or top-anchors a **content-sized** `SafeAreaView` (`flex: 0`) with `top`/`bottom` edges enabled, so the toasts are lifted above the system navigation bar (and below the status bar).
- Add an `anchorSide?: 'top' | 'bottom'` prop to `ToastProvider` (defaults to `'bottom'`) that controls whether the toast stack is anchored to the top or bottom of the screen.
- Keep the overlay permanently mounted so the native `SafeAreaView`'s asynchronously-resolved inset is ready before the first toast appears (otherwise the first toast would render behind the navigation bar until a re-layout).
- Use `anchorSide='top'` in the `test-tabs-lifecycle-events` screen so its toasts appear at the top, clear of the tab bar.

## Test 

Existing e2e test using toast are passing

### DIFF
--- a/apps/src/shared/Toast.tsx
+++ b/apps/src/shared/Toast.tsx
@@ -6,6 +6,7 @@ import {
   Dimensions,
   View,
   Platform,
+  ViewStyle,
 } from 'react-native';
 import { nanoid } from 'nanoid/non-secure';
 import { SafeAreaView } from 'react-native-screens/experimental';
@@ -15,6 +16,7 @@ interface ToastProps {
   id: string;
   backgroundColor: string;
   message: string;
+  style?: ViewStyle;
   remove: (_: string) => void;
 }
 
@@ -25,6 +27,7 @@ const Toast = ({
   id,
   backgroundColor,
   message,
+  style,
   remove,
 }: ToastProps): React.JSX.Element => {
   useEffect(() => {
@@ -35,7 +38,9 @@ const Toast = ({
   }, []);
 
   return (
-    <TouchableOpacity style={styles.container} onPress={() => remove(id)}>
+    <TouchableOpacity
+      style={[styles.container, style]}
+      onPress={() => remove(id)}>
       <View style={{ ...styles.alert, backgroundColor }}>
         <Text style={styles.text}>
           {`${index + 1}. `}
@@ -80,15 +85,13 @@ export const ToastProvider = ({ children }: ToastProviderProps) => {
     <ToastContext.Provider value={{ push }}>
       {children}
       {/*
-        Toasts live in a full-screen `box-none` overlay (so touches pass through
-        to `children`) that bottom-anchors them in a content-sized SafeAreaView
-        (`flex: 0`), whose Android `bottom` inset lifts them above the system
-        navigation bar.
-        The overlay stays permanently mounted so the native SafeAreaView's
-        asynchronously-resolved inset is ready before the first toast appears.
-        Note: While a toast is visible, the SafeAreaView spans the screen and 
-        swallows taps on the content behind it (such as the tab bar). When idle, the container only 
-        spans the safe navigation area where there is no interactive content.
+        Toasts render in a bottom-anchored, content-sized SafeAreaView whose
+        Android `bottom` inset keeps them above the system navigation bar. It
+        stays mounted even with no toasts: the native SafeAreaView needs a
+        layout pass to apply its bottom inset, so mounting it together with the
+        first toast renders that toast over the navigation bar (it only lifts
+        above once a later re-layout applies the inset). Keeping it mounted
+        means the inset is already applied when the first toast appears.
       */}
       <View style={styles.overlay} pointerEvents="box-none">
         <SafeAreaView
--- a/apps/src/shared/Toast.tsx
+++ b/apps/src/shared/Toast.tsx
@@ -70,12 +70,12 @@ interface ToastProviderProps {
 
 function ToastContainer({ toasts, remove, anchorSide }: { toasts: IToast[]; remove: (id: string) => void; anchorSide?: 'top' | 'bottom' }) {
   return (
-    <View style={styles.overlay} pointerEvents="box-none">
+    <View style={[styles.overlay, anchorSide === 'top' ? { justifyContent: 'flex-start' } : { justifyContent: 'flex-end' }]} pointerEvents="box-none">
       <SafeAreaView
         pointerEvents='box-none'
         collapsable={false}
         edges={{ top: true, bottom: true, }}
-        style={[styles.toastArea, anchorSide === 'top' ? { justifyContent: 'flex-start' } : { justifyContent: 'flex-end' }]}
+        style={[styles.toastArea]}
       >
         {toasts.map((toast, i) => (
           <Toast index={i} key={toast.id} {...toast} remove={remove} />
--- a/apps/src/shared/Toast.tsx
+++ b/apps/src/shared/Toast.tsx
@@ -5,6 +5,7 @@ import {
   StyleSheet,
   Dimensions,
   View,
+  ViewStyle,
 } from 'react-native';
 import { nanoid } from 'nanoid/non-secure';
 import { SafeAreaView } from 'react-native-screens/experimental';
@@ -25,7 +26,7 @@ const Toast = ({
   id,
   backgroundColor,
   message,
-  style,
+  style ={},
   remove,
 }: ToastProps): React.JSX.Element => {
   useEffect(() => {
--- a/apps/src/shared/Toast.tsx
+++ b/apps/src/shared/Toast.tsx
@@ -86,10 +86,9 @@ export const ToastProvider = ({ children }: ToastProviderProps) => {
         navigation bar.
         The overlay stays permanently mounted so the native SafeAreaView's
         asynchronously-resolved inset is ready before the first toast appears.
-        Note: the native SafeAreaView ignores `pointerEvents`, so while a toast
-        is visible it swallows taps on whatever is directly behind it (e.g. the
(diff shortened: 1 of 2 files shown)


apps/src/shared/Toast.tsx
```diff
@@ -82,33 +77,51 @@ export const ToastProvider = ({ children }: ToastProviderProps) => {
...
+        asynchronously-resolved inset is ready before the first toast appears.
+        Note: the native SafeAreaView ignores `pointerEvents`, so while a toast
+        is visible it swallows taps on whatever is directly behind it (e.g. the
+        tab bar) until the toast is tapped or auto-dismissed. When idle it only
+        spans the navigation bar area, which has no interactive content.
+      */}
+      <View style={styles.overlay} pointerEvents="box-none">
+        <SafeAreaView
+          edges={{ bottom: Platform.OS === 'android' }}
+          style={styles.toastArea}>
```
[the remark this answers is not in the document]
LKuchno: Thanks for the suggestion! I tried adding pointerEvents="box-none" to the SafeAreaView, but the tab bar buttons underneath are still not touchable.
Since the physical gaps between the toasts are so small anyway, a user wouldn't realistically be able to tap between them to hit the tab bar. It works the same way on iOS as well. Given this behavior and the fact that this functionality isn't needed for our tests, I'll leave the current behavior as is.

apps/src/shared/Toast.tsx
```diff
@@ -27,7 +25,6 @@ const Toast = ({
   id,
   backgroundColor,
   message,
-  style = {},
   remove,
```
kkafar, line 30: This doesn't seem right nor related. Why do you remove the `style` prop here? It's okay to have ability to modify the style of the Toast locally from the use site.

kkafar: One thing to improve though, is instead using `...` spread syntax, you might actually just pass an array of styles.

apps/src/shared/Toast.tsx
```diff
@@ -82,33 +77,50 @@ export const ToastProvider = ({ children }: ToastProviderProps) => {
...
+        to `children`) that bottom-anchors them in a content-sized SafeAreaView
+        (`flex: 0`), whose Android `bottom` inset lifts them above the system
+        navigation bar.
+        The overlay stays permanently mounted so the native SafeAreaView's
+        asynchronously-resolved inset is ready before the first toast appears.
+        Note: While a toast is visible, the SafeAreaView spans the screen and 
+        swallows taps on the content behind it (such as the tab bar). When idle, the container only 
+        spans the safe navigation area where there is no interactive content.
+      */}
+      <View style={styles.overlay} pointerEvents="box-none">
```
kkafar: This comment is redundant. 

One thing that worries me is that it says that `SafeAreaView` "[...] spand the screen and swallows taps on the content behind it [...]". If that is really the case, then this setup introduces a bug. Please verify it.

apps/src/shared/Toast.tsx
```diff
@@ -82,33 +77,50 @@ export const ToastProvider = ({ children }: ToastProviderProps) => {
...
+        <SafeAreaView
+          edges={{ bottom: Platform.OS === 'android' }}
```
kkafar: Why do we only apply it on Android? Isn't it also a problem on iOS?

apps/src/shared/Toast.tsx
```diff
@@ -82,33 +77,50 @@ export const ToastProvider = ({ children }: ToastProviderProps) => {
...
+          style={styles.toastArea}>
           {toasts.map((toast, i) => (
-            <Toast
-              index={i}
-              key={toast.id}
-              style={{ marginBottom: i * 25 }}
-              {...toast}
-              remove={remove}
-            />
+            <Toast index={i} key={toast.id} {...toast} remove={remove} />
```
kkafar: Why is tyle `style` property removed here? Please explain.

kkafar (changes requested):
Left some remarks below. Please answer them before proceeding.

apps/src/shared/Toast.tsx
LKuchno: Problem was android specific. 
Toast appear behind navigation bar and in result couldn't be checked in e2e tests. On iOS toast are visible normally, not covered by nav bar. 

Before fix it was displayed this way:
Android | iOS
-- | --
image| Simulator Screenshot - iPhone 17 Pro - 2026-07-15 at 12 26 14

apps/src/shared/Toast.tsx
LKuchno: In current implementation for Android using this style results with toast rendered the way visible on screenshot below, thats why this style was removed and spacing now comes from styles.container (marginBottom: 5). If we want keep style here, it can be define as marginBottom: 5 but I think it will be code duplication.

Screenshot_1784115427

apps/src/shared/Toast.tsx
LKuchno: I revert this change and apply your suggestion. I will remove style in another PR.

apps/src/shared/Toast.tsx
```diff
@@ -67,9 +66,27 @@ const ToastContext = createContext({
 
 interface ToastProviderProps {
   children: React.ReactNode;
+  anchorSide?: 'top' | 'bottom';
 }
 
-export const ToastProvider = ({ children }: ToastProviderProps) => {
+function ToastContainer({ toasts, remove, anchorSide }: { toasts: IToast[]; remove: (id: string) => void; anchorSide?: 'top' | 'bottom' }) {
```
kkafar: Let's move the param type to a separate type definition, just like `ToastProviderProps` to make it more readable.

apps/src/shared/Toast.tsx
```diff
@@ -67,9 +66,27 @@ const ToastContext = createContext({
...
+  return (
+    <View style={[styles.overlay, anchorSide === 'top' ? { justifyContent: 'flex-start' } : { justifyContent: 'flex-end' }]} pointerEvents="box-none">
+      <SafeAreaView
+        pointerEvents='box-none'
```
kkafar: The `pointerEvents='box-none'` does not currently work on Android, however let's leave it be. Maybe someday we'll fix it in core.

kkafar (dismissed):
Looks good. I just have a minor request regarding type definition.

apps/src/shared/Toast.tsx
LKuchno: I added comment about this limitation https://github.com/software-mansion/react-native-screens/pull/4334/changes/d753a0878bbadbb7c91b91667b15f66b65a90449

apps/src/shared/Toast.tsx
LKuchno: Params moved to separate definition https://github.com/software-mansion/react-native-screens/pull/4334/changes/d753a0878bbadbb7c91b91667b15f66b65a90449

apps/src/shared/Toast.tsx
```diff
@@ -67,9 +66,27 @@ const ToastContext = createContext({
...
+        collapsable={false}
+        edges={{ top: true, bottom: true, }}
+        style={styles.toastArea}
+      >
```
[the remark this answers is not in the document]
t0maboro: +1, just run `yarn format-js` from the repo's root

apps/src/shared/Toast.tsx
```diff
@@ -82,33 +106,31 @@ export const ToastProvider = ({ children }: ToastProviderProps) => {
...
-              {...toast}
-              remove={remove}
-            />
-          ))}
-        </>
-      </ToastContext.Provider>
-    </SafeAreaView>
+    <ToastContext.Provider value={{ push }}>
+      {children}
+      <ToastContainer toasts={toasts} remove={remove} anchorSide={anchorSide ?? 'bottom'} />
```
t0maboro: 1. `style` is no longer passed to Toast, so I'd consider removing it from `ToastProps`
2. nit: `anchorSide` can be assigned by default to 'bottom' in the function's signature; then, in `ToastContainerProps` it can be a non-optional prop - I like this approach a bit more, because here 
```
anchorSide === 'top' ? { justifyContent: 'flex-start' } : { justifyContent: 'flex-end' }]
```
'bottom' and nothing means 'bottom', and we can easily achieve an exact differentiation between 'top' and 'bottom' only

t0maboro (commented):
Please run `yarn format-js` from the root of the repo. I left some comments about minor cleanups, but I already tested the runtime; after resolving these, we'll ship it

apps/src/shared/Toast.tsx
LKuchno: https://github.com/software-mansion/react-native-screens/pull/4334/changes/551c2d1570c2ba9ce507c702d07e0d8b4ccaadd0

t0maboro (approved):

PR #4335, opened 2026-07-15
fix(Android): defensively guard against an unexpected menuItem Id

### LINKED ISSUE #4285: [expo-router] NativeTabs: ClassCastException on Android when CustomBottomNavigationView is assigned a colliding view id
### Description

On Android, `NativeTabs` (the gamma native bottom tabs, via `expo-router/unstable-native-tabs`) crashes with a `ClassCastException` when a tab is selected after a third-party SDK assigns a view id to the (id-less) `CustomBottomNavigationView`.

**Real-world trigger:** the Pendo Mobile SDK (`rn-pendo-sdk`). Its codeless analytics walks the view hierarchy and assigns its own small sequential ids to views that have none (`View.NO_ID`), as part of a screen-scoped dynamic-element scan. `react-native-screens` never assigns `CustomBottomNavigationView` an id at construction, so it is `View.NO_ID` by default — and Pendo was observed stamping **`id=2`** onto it, colliding with menu-item id 2. Confirmed with an instrumented `setId` override; the caller was `sdk.pendo.io.x6.h` (stack trace below). This is not Pendo-specific — any SDK or app code that tags id-less views the same way (a common codeless-analytics pattern) will hit the same collision.

A minimal, runnable reproduction (drops in the real Pendo SDK, no fakes) is here: **https://github.com/ekrapfl/rn-screens-nativetabs-repro**

**Stack trace:**

```
java.lang.ClassCastException: com.swmansion.rnscreens.gamma.tabs.container.CustomBottomNavigationView cannot be cast to com.google.android.material.navigation.NavigationBarItemView
	at com.swmansion.rnscreens.gamma.tabs.host.TabsHostA11yCoordinator.setA11yPropertiesToTabItem(TabsHostA11yCoordinator.kt:19)
	at com.swmansion.rnscreens.gamma.tabs.host.TabsHostA11yCoordinator.setA11yPropertiesToAllTabItems(TabsHostA11yCoordinator.kt:34)
	at com.swmansion.rnscreens.gamma.tabs.container.TabsContainer.updateBottomNavigationViewAppearanceIfNeeded(TabsContainer.kt:468)
	at com.swmansion.rnscreens.gamma.tabs.container.TabsContainer.performPostSelectedTabUpdateActions(TabsContainer.kt:447)
	at com.swmansion.rnscreens.gamma.tabs.container.TabsContainer.performContainerUpdate(TabsContainer.kt:439)
	at com.swmansion.rnscreens.gamma.tabs.container.TabsContainer.flushPendingUpdates(TabsContainer.kt:194)
	at com.swmansion.rnscreens.gamma.tabs.host.TabsHost.didMountItems(TabsHost.kt:197)
	...
```

**Root cause (confirmed with on-device instrumentation).** In [`TabsHostA11yCoordinator.setA11yPropertiesToTabItem`](https://github.com/software-mansion/react-native-screens/blob/main/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/host/TabsHostA11yCoordinator.kt):

```kotlin
val menuView = bottomNavigationView.findViewById<NavigationBarItemView>(menuItem.itemId)
```

- `View.findViewById(id)` returns the **receiver itself** when `receiver.id == id`.
- Menu-item ids are small ints (`fragmentIndex + 1`, i.e. `1..N`, per `MenuHelpers.menuItemIdForFragmentAtIndex`).
- `CustomBottomNavigationView` is never given an id (see [`CustomBottomNavigationView.kt`](https://github.com/software-mansion/react-native-screens/blob/main/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/CustomBottomNavigationView.kt) — no `id = ...` anywhere), unlike `TabsContainer`, which generates one for itself via `ViewIdGenerator` ([`TabsContainer.kt:56`](https://github.com/software-mansion/react-native-screens/blob/main/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/container/TabsContainer.kt#L56)).
- Once something else assigns the bar an id that happens to fall in `1..N`, `findViewById(n)` returns the bar itself instead of the `n`-th menu item, and the inline cast to `NavigationBarItemView` throws.
- `setA11yPropertiesToAllTabItems()` runs on **every** selection change (`invalidateOnSelectedTabChanged()` → `updateBottomNavigationViewAppearanceIfNeeded()`), so the next tab tap after the id is stamped crashes.

Instrumented `setId` override + logcat (app with `rn-pendo-sdk` 3.13.1 / Pendo Android SDK 3.13.7):

```
W RNS_SETID: CBNV.setId(2)
W RNS_SETID: java.lang.Exception: who-set-the-bnv-id
W RNS_SETID: 	at com.swmansion.rnscreens...CustomBottomNavigationView.setId(CustomBottomNavigationView.kt:16)
W RNS_SETID: 	at sdk.pendo.io.x6.h.a(SourceFile:5)
W RNS_SETID: 	at sdk.pendo.io.x6.h$f.invokeSuspend(SourceFile:18)
W RNS_SETID: 	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:34)
	...
```

Corresponding before/after diagnostic (`bnv.id` = `CustomBottomNavigationView.id`, logged from `TabsHostA11yCoordinator`):

```
# before Pendo tags the bar
bnv.id=-1 itemId=1 found=BottomNavigationItemView   # no collision
bnv.id=-1 itemId=2 found=BottomNavigationItemView

# after Pendo's scan stamps id=2 on the bar
bnv.id=2  itemId=1 found=BottomNavigationItemView
bnv.id=2  itemId=2 found=CustomBottomNavigationView  # collision → crash on next select
```

**Expected:** selecting a tab should not crash.

**Suggested fix** — two independent hardenings:

1. Look the item view up as a `View` and type-check before use (and set `contentDescription`, which does not need the view, unconditionally):

   ```kotlin
   if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
       menuItem.contentDescription = tabsScreen.tabBarItemAccessibilityLabel
   }
   val menuView = bottomNavigationView.findViewById<View>(menuItem.itemId)
   if (menuView is NavigationBarItemView) {
       menuView.tag = tabsScreen.tabBarItemTestID
   }
   ```

2. Assign `CustomBottomNavigationView` a generated id at construction, the same way `TabsContainer` already does for itself (`ViewIdGenerator`), so third-party id-tagging can never collide with a menu-item id lookup.

Happy to open a PR for either or both.

### Steps to reproduce

Full instructions (including native-build steps) are in the [repro repo](https://github.com/ekrapfl/rn-screens-nativetabs-repro); summary:

1. `NativeTabs` (`expo-router/unstable-native-tabs`) with a couple of triggers, each a nested `Stack` route.
2. Wire in `rn-pendo-sdk` (`PendoSDK.setup()` + `WithPendoExpoRouter` for screen tracking) — any Pendo app key works, it's a public send-only value.
3. Run on Android (New Architecture), start a Pendo session, then switch tabs a few times (a short pause between taps — Pendo's scan is debounced ~750ms after a screen change).
4. → Crash.

https://github.com/ekrapfl/rn-screens-nativetabs-repro

### Screens version

4.25.2

### React Native version

0.85.3

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

Android emulator

### Device model

Pixel 10 Pro XL emulator, Android 17 (API 37)

### Acknowledgements

Yes

ekrapfl:
## Description

`TabsHostA11yCoordinator` looked up a tab's item view via `bottomNavigationView.findViewById(menuItem.itemId)`. Material's menu item IDs are small ints - the same range third-party SDKs might stamp onto id-less views. `CustomBottomNavigationView` had no id, so with the bar itself stamped with a small id, the lookup returned the bar, and the cast threw `ClassCastException` on tab select.

Closes #4285

## Changes

- We now match the id only among the item views Material built, via the public `getMenuViewGroup()

## Test plan

Fixes the issue mentioned in #4285 (verified in the repro).

## Checklist

- [ ] Ensured that CI passes


t0maboro:
Thank you for opening this PR. The one concern I have is that `NavigationBarItemView` is restricted Material API; we prefer to avoid it as much as possible - I'll try to propose a different approach.

t0maboro:
@ekrapfl I am proposing a slightly different approach. May I ask you to verify whether this resolves the issue?

android/src/main/java/com/swmansion/rnscreens/tabs/host/TabsHostA11yCoordinator.kt
```diff
@@ -16,18 +17,20 @@ class TabsHostA11yCoordinator(
...
         }
 
         // when matching view by id, espresso driver seems to look for tag property
-        menuView.tag = tabsScreen.tabBarItemTestID
+        findTabItemView(menuItem.itemId)?.tag = tabsScreen.tabBarItemTestID
     }
 
+    // Third-party SDKs can freely assign their own ids to id-less views in CustomBottomNavigationView's subtree and there is no way to
+    // coordinate that id space with them, so we should search ids only inside Material's menu items scope.
+    private fun findTabItemView(menuItemId: Int): View? = bottomNavigationView.menuViewGroup.children.firstOrNull { it.id == menuItemId }
```
kkafar, line 32: Let's use `findViewById` here also. If material implementation ever adds a layer this code will break.

kkafar (commented):
I want it also to be said that I do not consider this to be issue on our side , but rather the third party SDK, which assigns conflicting ids. The problem the bahaviour of the third party SDK introduces is universal and might arise in different part of the app as easily as it has here. 

I believe though, that on our side we can tighten our search space - it won't hurt, and provides one less place for corruption by the third party behaviour.

android/src/main/java/com/swmansion/rnscreens/tabs/host/TabsHostA11yCoordinator.kt
t0maboro: The problem with `findViewById` is that `menuViewGroup` hasn't been assigned an ID at all, so we'll just defer the issue one layer down. I see the reasoning why this code is flaky, but I think that we can easily catch any regression after bumping material, as I saw that our e2e tests are relying on that `tabBarItemTestID`

t0maboro: ofc, I am still open for developing a more robust solution if we can find some time in the future

ekrapfl:
> @ekrapfl I am proposing a slightly different approach. May I ask you to verify whether this resolves the issue?

Hey @t0maboro, I did some testing in my repro, and I believe this fixes it just as well. Thank you!


> I want it also to be said that I do not consider this to be issue on our side , but rather the third party SDK, which assigns conflicting ids. The problem the bahaviour of the third party SDK introduces is universal and might arise in different part of the app as easily as it has here.
> 
> I believe though, that on our side we can tighten our search space - it won't hurt, and provides one less place for corruption by the third party behaviour.

@kkafar, I agree that the other library is the one doing something a bit dubious, and it is just manifesting in this library. I also feel it is good practice to be a bit defensive in a situation like this, and hopefully no other issues like that come back here ¯\\\_(ツ)_/¯

t0maboro (approved):
note: I'll merge this PR after we finish the repo structure reorganization

PR #4336, opened 2026-07-15
fix(android): Fix stale padding after orientation change in screens below the top screen

kmichalikk:
## Description

The issue that PR tries to patch is that the commit hook introduced in https://github.com/software-mansion/react-native-screens/pull/3295 resets the size information on orientation change on all screens in stack, but only the ones that have native `onLayout` called can respond correctly.

### More detailed explanation:

The original case that the commit hook fixes focused on currently displayed screen. We're setting the screen size by hand to account for header and the set size is outdated for a brief time after rotating, until native calls `onLayout` which calls state update. The commit hook's purpose is to reset the size to undefined when it detects that the orientation of the screen's parent views (they are not set by hand and therefore not affected) are different than the stale orientation of the screen. Yoga can then immediately relayout the contents of the screen as if it was the first render, with padding for the header and no offset and the content looks good this way. Later `onLayout` is called and the values are set again correctly.

The above depends on the `onLayout` call success. But for the following scenario:

- Home screen initially rendered
- push Second screen
- push Third screen which forces landscape

`onLayout` is called on the Third screen so it goes as expected. After leaving the landscape we're back to portrait, so `onLayout` triggers for the Second screen and it goes as expected. However, going back to the Home screen, we're detecting no change - it was portrait earlier, and it is portrait now, but the commit hook reset has happened and we're having undefined size and top padding, with `onLayout` call skipped due to no apparent change. We're left with incorrect padding and misplacement of shadow nodes which breaks pressables.

To solve this, we need to somehow differentiate between a screen that is on top and rotated, which requires commit hook treatment, and the screen below that may not need it. It can't be for 100% because user might stay in landscape, but that should trigger `onLayout` anyways and things should be good.

There is however no simple way to differentiate the screens, so instead of skipping the commit hook, we're resetting the last known sizes for the screen that is put below the new screen so when going back to it, the size change is actually detected and the code runs correctly.

## Changes

Added a way to force state update in `onLayout` for screens that user goes back to

n/a

## Test plan

Use the provided test, observe the positions of pressables in Home screen after going to fullscreen in third screen and going back to Home

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/fabric/java/com/swmansion/rnscreens/legacy/FabricEnabledViewGroup.kt
+++ b/android/src/fabric/java/com/swmansion/rnscreens/legacy/FabricEnabledViewGroup.kt
@@ -22,6 +22,12 @@ abstract class FabricEnabledViewGroup(
         mStateWrapper = wrapper
     }
 
+    protected fun resetLastSizes() {
+        lastWidth = 0f
+        lastHeight = 0f
+        lastHeaderHeight = 0f
+    }
+
     @UiThread
     fun updateState(
         width: Int,
--- a/android/src/main/java/com/swmansion/rnscreens/legacy/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/legacy/Screen.kt
@@ -236,7 +236,7 @@ class Screen(
     ) {
         // In case of form sheet we get layout notification a bit later, in `onBottomSheetBehaviorDidLayout`
         // after the attached behaviour laid out this view.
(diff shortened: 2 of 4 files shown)


kmichalikk:
Irrespective of what happens with the linked issue, this is a bug worth patching, but we may have to refactor it before landing.

Zhiyilang074811:
Great fix! Ready for review.

Zhiyilang074811:
Great fix! Ready for review.

Zhiyilang074811:
Great fix! Ready for review.

PR #4343, opened 2026-07-16
chore(test, iOS): add e2e test for Stack Header Icons

LKuchno:
## Description

Adds an end-to-end (Detox) test for the **Stack Header Icons (iOS)** single-feature test screen (`test-stack-header-icon-ios`). The spec covers the full manual scenario: every icon variant (`sfSymbol`, `xcasset`, `imageSource`, `templateSource`) is asserted positively — on the bar button item, on the header item menu rows, and inside the nested submenu. The test screen is wired with the `testID`s the spec relies on and the coverage is recorded in `scenario.md` / `scenario-description.ts`.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1663

## Changes

- New e2e spec: `FabricExample/e2e/single-feature-tests/stack-v5/test-stack-header-icon-ios.e2e.ts`.
  - Asserts the header title renders and the trailing bar button item shows the `star.fill` `sfSymbol` icon.
  - Verifies bar-button icon cycling (`sfSymbol → xcasset → imageSource → templateSource → sfSymbol`) directly: after each tap of `cycle-item-icon-button` the new icon must be visible **and** the previous one must no longer exist, anchored on the on-screen `current-item-icon` text.
  - Icons are matched by their icon id (SF Symbol name or asset path) scoped to a native ancestor — `_UIModernBarButton` for the bar button, `_UIContextMenuCellContentView` for menu rows — instead of relying on accessibility labels.
  - Opens the header item menu on long press, asserts all four rows (`Toggle 1..3`, `Submenu`) are visible and that **every** row renders the current icon variant.
  - Opens the nested submenu and asserts each submenu row (`Sub Toggle 1..3`) renders its icon; submenu rows are matched by their own title rather than by index, because the parent menu's cells share the same native class and stay in the hierarchy while the submenu is open.
  - Verifies the in-menu "Cycle icons" action advances the menu icon variant through all four values and back, re-checking every menu row's icon at each step.
  - Matches native views through shared constants from the `native-class-names` registry rather than inline string literals.
- `native-class-names.ts`: registered the native classes this spec queries — `_UIModernBarButton`, `_UIContextMenuCellContentView`, `_UIContextMenuSubmenuTitleView` and `UILabel` — so a native symbol rename only has to be applied in one place.
- `apps/.../test-stack-header-icon-ios/index.tsx`: added `testID`s (`cycle-item-icon-button`, `cycle-menu-icons-button`, `current-item-icon`, `current-menu-icon`, `push-another-screen-button`) needed for stable Detox matching.
- `scenario-description.ts`: `e2eCoverage` changed from `tbd` to `full`.
- `scenario.md`: E2E section updated to "Full: all steps covered."


FabricExample/e2e/single-feature-tests/stack-v5/test-stack-header-icon-ios.e2e.ts
```diff
@@ -0,0 +1,150 @@
...
+    );
+  });
+
+  it('should display the header with a star sfSymbol icon on the trailing item', async () => {
+    await expect(
+      element(by.type('UILabel').and(by.text('Header Icons'))),
+    ).toExist();
+    await expect(
+      element(by.type('_UIModernBarButton').and(by.label('favorite'))),
+    ).toBeVisible();
```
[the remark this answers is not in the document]
LKuchno: Updated
https://github.com/software-mansion/react-native-screens/pull/4343/changes/9fb7fd5d6db47b5b9824d411fac46e16225309c3

kmichalikk (approved):
Looks good. We could add some comment about the indirect checking when cycling directly in the e2e code, there would be less confusion when looking at only the code in the future, but up to you.

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-header-icon-ios.e2e.ts
```diff
@@ -0,0 +1,160 @@
...
+          .and(by.label(rowTitle)),
+      ),
+  );
+
+// Asserts that every submenu row renders the `sfSymbol` icon.
+const expectAllSubmenuRowIconsToBeVisible = async () => {
+  for (const rowTitle of SUBMENU_ROWS) {
+    await expect(submenuRowIcon(rowTitle)).toBeVisible();
+  }
+};
```
kmichalikk: just for consistency, I'd add iconId to `submenuRowIcon()` and pass sfSymbol to it; then the comment about `sfSymbol` would make more sense because there would be an `sfSymbol` used in the function.

kmichalikk (approved):
👍

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-header-icon-ios.e2e.ts
LKuchno: Changed https://github.com/software-mansion/react-native-screens/pull/4343/changes/f6ff69a281823c4a262a79165f58ca082a3daaef
I also added comment about imageSource and templateSource icon id`s

kmichalikk (approved):

PR #4357, opened 2026-07-17
fix(Android, Stack): don't add top inset to measured header height when it's disabled

phazei:
## Description

When a header opts out of the top inset on Android (`disableTopInsetApplication`, which sets `consumeTopInset = false`), the measured header height still includes the status bar / cutout inset, so it comes back too tall.

The native `CustomToolbar` only pads itself with the top inset when `legacyTopInsetBehavior || consumeTopInset` is true. But the dummy layout used to report the header height back to the shadow tree always adds the decor view top inset, regardless of that prop. So the height it reports doesn't match what the toolbar actually renders, and the screen content ends up pushed down by roughly the status bar height on the first Fabric layout.

This looks like it was just missed when the `disable*InsetApplication` props landed in #4220 — the toolbar learned about `consumeTopInset` but the measurement path didn't.

## Changes

- `RNSScreenShadowNode.cpp`: read the same header props the toolbar uses (`legacyTopInsetBehavior || consumeTopInset`) and pass the result down to the measurement as `applyTopInset`. The JNI signature changes from `(IZ)F` to `(IZZ)F`.
- `ScreenDummyLayoutHelper.kt`: only add the decor view top inset when `applyTopInset` is true, and include the flag in the measurement cache key so screens with different settings don't share a stale cached height.

## Test plan

Reproducible with the existing `Test4220` scenario:

1. Open `Test4220` on Android.
2. Go to the Details screen and toggle `disableTopInsetApplication` on.

Before this change the screen content sits about a status bar's height too low (the header reserves space it isn't actually using). After, the content sits directly under the header.

## Checklist

- [x] Included code example that can be used to test this change. (existing `Test4220`)
- [x] For API changes, updated relevant public types. (no public API change)
- [ ] Ensured that CI passes


t0maboro:
Acknowledging the issue, I am adding MRE to this PR, and I'll land it soon, but this needs to wait until we complete the repo structure reorganization.
@phazei thank you for fixing that!

t0maboro (approved):

PR #4366, opened 2026-07-20
chore: block the possibility to create plain-text issues and update agents instructions

kkafar:
## Description

I remove here blank issue template. This will be less convenient for us, e.g. when opening new
pick request thread (~but we can create another dedicated template for that purpose~ - I've done that).

People (mostly via LLMs) are abusing it. There are numerous cases
recently with LLM produced issue reports that do not adhere to our template
and use the blank one instead, failing to provide the required
information.

I've also updated instructions in AGENTS.md with hope that in certain cases this
will guide agents and result in better output.


kligarski (approved):
Seems like a good idea.

kmichalikk (approved):

PR #4370, opened 2026-07-20
chore: Move v4 iOS code to legacy directory

kmichalikk:
Closes https://github.com/software-mansion/react-native-screens-labs/issues/1675

## Description

This PR moves legacy iOS code to legacy/ directory. Enums and conversion methods are split between gamma / legacy so that removing the legacy dir in the future gets rid of the old enums, too.

## Changes

- moved legacy files to separate legacy/ directory
- updated imports

n/a

## Test plan

Build the application and see examples load

## Checklist

- [ ] Ensured that CI passes


ios/legacy/RNSLegacyEnums.h
```diff
@@ -0,0 +1,88 @@
+#pragma once
```
kmichalikk, line 1: having RNSEnums.h both in top-level and legacy/ dirs breaks the build, so I need to rename the file

ios/legacy/conversion/RNSConversions.h
kkafar, line 1: I think that after creation of this file, we'll now end up with duplicated RNSConversions.h files. This might lead to problems, since the headers are flattened in the final output of a built framework. Let's be careful here and void name collisions, even across different project directories.

ios/legacy/RNSLegacyEnums.h
kkafar: makes sense!

kkafar (commented):
RNSUtility categories shouldn't be placed in the legacy directories. They seem useful for future use. I assume that we don't use them yet, but I believe we'll, as we'll encounter the very same problems we did back in stack v4.

Also I don't like how the conversions method are placed. This is outside of the scope of this PR, but let's create a ticket please, where we'll come up with a better approach than the current one. I don't think that putting all conversion methods and enums in a single file was ever a good decision (it was my decision though :D).
Each component should have they own. Let's create a ticket.

ios/legacy/conversion/RNSConversions.h
kmichalikk: I actually backed off from moving things to legacy. There are only 2 candidates, of which one is still being used in UIViewController+RNScreens and the second one is complementary to the first one, so I think we can skip this one.

kkafar (approved):
Looks good. Thanks!

PR #4174, opened 2026-06-16
feat: Adopt AGP v9

hurali97:
## Description

This PR is rasied following the second phase of the RFC around AGP v9 adoption:

RFC: https://github.com/react-native-community/discussions-and-proposals/pull/1006

## Changes

The gist of this PR is to make RNScreens AGP v9 compliant with backward compatibility. The main scope of changes is the `react-native-screens/android/build.gradle`. The rest of the changes can be considered temporary.

The rest changes include:

- Making Fabric Example App AGP9 compliant
- Upgrade `Gradle` to `v9.4.1`
- Use `proguard-android-optimize` proguard file
- Enable opt outs in the `gradle.properties`

Ideally, we should not enable the opt outs and leverage the AGP 9 built-in kotlin and newDSL. However, if we do not do it, then other libraries which are not yet AGP v9 compliant starts to fail. Hence, we need to keep the opt outs enabled for a while. For the context, react-native starting from 0.87.x will ship with AGP v9 and opt outs enabled by default for the new apps, which is the first phase of AGP v9 adoption.

With the second phase when the libraries starts adoption AGP v9, we can eventually remove the opt outs from Fabric Example.

## Test plan

To test this PR, we need to do a few steps:

- Set the `agp` version to `9.2.1` and `kotlin` to `2.2.0` in `react-native-gradle-plugin`
```diff
--- a/node_modules/@react-native/gradle-plugin/gradle/libs.versions.toml
+++ b/node_modules/@react-native/gradle-plugin/gradle/libs.versions.toml
@@ -1,10 +1,10 @@
 [versions]
-agp = "8.12.0"
+agp = "9.2.1"
 gson = "2.8.9"

-kotlin = "2.1.20"
+kotlin = "2.2.0"
 assertj = "3.25.1"
```

- Comment out the following functions in `react-native-gradle-plugin`
```diff
--- a/node_modules/@react-native/gradle-plugin/react-native-gradle-plugin/ReactPlugin.kt
+++ b/node_modules/@react-native/gradle-plugin/react-native-gradle-plugin/ReactPlugin.kt
@@ -1,10 +1,10 @@
      configureBuildTypesForApp(project)
    }

    // Library Only Configuration
-    configureBuildConfigFieldsForLibraries(project)
-    configureNamespaceForLibraries(project)
+    // configureBuildConfigFieldsForLibraries(project)
+    // configureNamespaceForLibraries(project)
    project.pluginManager.withPlugin("com.android.library") {
```

[...]

PR #4375, opened 2026-07-21
chore: add support for react-native@0.87.0-rc.3

t0maboro:
## Description

Upgrade based on the RN upgrade helper: https://react-native-community.github.io/upgrade-helper/?from=0.86.0&to=0.87.0-rc.3

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1666

### Baseline

- Bumped RN.packages to `0.87.0-rc.3` and TypeScript to TS6
- Bumped AGP on Android to 9
- Bumped `react-native-safe-area-context` to 5.8.0 - https://github.com/AppAndFlow/react-native-safe-area-context/releases/tag/v5.8.0 - the PR released in this version is fixing the build error for symbols that were removed in RN 0.87.

### TypeScript - RN 0.87 strict TS API + TS 6 support

- In RN 0.87's generated types, `View` is a function component, so `React.Ref<View>` / `useRef<View>` no longer describe a host instance. All ref sites (e.g., `Screen`, `ScreenStackItem`, `ScreenStackHeaderConfig`) now use the public `ViewInstance` type.
- RN 0.87's real types live under the `types_generated/` (treated as non-public) path, so any *inferred* exported type referencing them fails `.d.ts` emit during `npm pack` (`bob build`). Fixes:
  - `as HostComponent<NativeProps>` on ~30 `codegenNativeComponent` default exports (`HostComponent` is publicly exported; other libs are following that pattern with an explicit cast: https://github.com/software-mansion/react-native-reanimated/blame/4.5.3/packages/react-native-reanimated/src/specs/SharedTransitionBoundaryNativeComponent.ts#L9-L12
  - Nominal instance interfaces for `forwardRef` components (`export interface ScreenInstance extends ViewInstance {}`, same for `ScreenStackHeaderConfigInstance`) - inference stops at a nominal interface, so the fix propagates without further annotations (was resolving to `ReactNativeElement`, which is a non-public type from RN)
  - Exported `TransitionProgressContextBody` and used it as the explicit return type of `useTransitionProgress`.
- Set `rootDir: "./src"` in `tsconfig.build.json`
- `moduleResolution: "bundler"`, dropped `baseUrl`.** `bundler` resolution is required to resolve RN 0.87's `exports`-based type entry points; `baseUrl` is deprecated in TS 6 and the `paths` mappings work relative to the config without it. For more context, see: https://github.com/microsoft/TypeScript/issues/62508
- Restored the removed `TextInputFocusEventData`. RN changed the exported type to `TextInputFocusEvent` (now the full `NativeSyntheticEvent`). A local backward-compat interface is introduced.

### SearchBar

- Migrated from `forwardRef` to React 19 ref-as-prop
- Retyped the internal ref against the native host instance (`React.ComponentRef<typeof SearchBarNativeComponent>`) instead of the previous double-casted `NativeSearchBar` wrapper.

### iOS

- Added missing `_props` default initializers to Fabric component views (e.g., `RNSScreenContentWrapper`). `RCTViewComponentView`'s `updateProps:oldProps:` asserts in debug builds that every subclass initializes `_props` with its own concrete default props in its initializer, starting from 0.87.

### Android

- AGP 9 has built-in Kotlin support.
- AGP 9 compiles Kotlin itself by default, and additionally applying `kotlin-android` plugin fails the build - so the plugin is now applied only when built-in Kotlin is inactive (AGP < 9, or the consumer opted out via `android.builtInKotlin=false`). 
- Registered `src/fabric/java` in the Kotlin source set so the Fabric sources compile under built-in Kotlin.

N/A

## Test plan

- `yarn check-types` - passed
- `npm pack` - passed
- codegen was verified
- `FabricExample` - built and run on iOS and Android.
- Checked the compatibility with the lowest supported RN version (0.84)

Action items yet to be covered:
- test it with SwiftPM
- full testing coverage of the most important components on (0.84, 0.87) x (Android, iOS) x (Debug, Release) combos
- more FabricExample testing - both Debug and Release
- consider adding some commit-level check for detecting issues related to non-public types and TS inference.

## Checklist

- [ ] Ensured that CI passes


src/components/Screen.tsx
```diff
@@ -225,7 +230,14 @@ export const InnerScreen = React.forwardRef<View, ScreenProps>(
...
             // Detailed information can be found here https://github.com/software-mansion/react-native-screens/pull/2351
-            style={[style, { zIndex: undefined }]}
+            style={[
+              style,
+              // The cast is needed because `exactOptionalPropertyTypes` forbids an explicit `undefined` on `zIndex`
+              // after enabling strict types in RN 0.87 where it is typed as `zIndex?: number`.
+              { zIndex: undefined } as {
+                zIndex?: number | undefined;
+              } as ViewStyle,
+            ]}
```
t0maboro: note: workaround, because `undefined` was removed from public type for `zIndex`

src/legacy/types.ts
```diff
@@ -16,6 +16,16 @@ import type {
...
 
+/**
+ * `react-native` 0.87 removed the `TextInputFocusEventData` export (in favor of `TextInputFocusEvent`).
+ * We keep a local copy of the previous data shape to keep `SearchBarProps['onChangeText']`
+ * backward compatible.
+ */
+export interface TextInputFocusEventData extends TargetedEvent {
+  eventCount: number;
+  text: string;
+}
```
t0maboro: for further discussion: I haven't found an equivalent in strict types on 0.87, I decided to copy the type from the core to avoid breaking changes in public API

src/legacy/components/SearchBar.tsx
```diff
@@ -10,70 +10,58 @@ import { View } from 'react-native';
...
 
-function SearchBar(
-  props: SearchBarProps,
-  forwardedRef: React.Ref<SearchBarCommands>,
-) {
-  const searchBarRef = React.useRef<SearchBarCommands>(null);
-
-  React.useImperativeHandle(forwardedRef, () => ({
+  React.useImperativeHandle(props.ref, () => ({
     blur: () => {
```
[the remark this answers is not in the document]
t0maboro: fresh RN 0.84 example (lowest version we're supporting) comes with React 19, imo no action item
```
    "react": "19.2.3",
    "react-native": "0.84.1",
```

src/legacy/components/SearchBar.tsx
```diff
@@ -124,4 +114,4 @@ function SearchBar(
   );
 }
 
-export default React.forwardRef<SearchBarCommands, SearchBarProps>(SearchBar);
+export default SearchBar;
```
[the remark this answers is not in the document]
t0maboro: fresh RN 0.84 example (lowest version we're supporting) comes with React 19, imo no action item
```
    "react": "19.2.3",
    "react-native": "0.84.1",
```

src/components/Screen.web.tsx
```diff
@@ -18,9 +18,12 @@ export class NativeScreen extends React.Component<ScreenProps> {
       activityState,
       style,
       enabled = screensEnabled(),
+      ref,
       ...rest
     } = this.props;
 
+    const refProp = ref !== undefined ? { ref } : {};
+
```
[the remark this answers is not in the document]
t0maboro: fresh RN 0.84 example (lowest version we're supporting) comes with React 19, imo no action item
```
    "react": "19.2.3",
    "react-native": "0.84.1",
```

FabricExample/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "extends": "@react-native/typescript-config",
   "compilerOptions": {
     "baseUrl": ".",
-    "types": ["jest", "node"],
+    "types": ["node"],
     "paths": {
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4375/commits/5790409083d0ab12f7d04f2b6daa973cca5f889d

tsconfig.json
kligarski, line 24: In upgrade helper, `jest` is removed. Not sure if this should be applied in our library as well?

Image

FabricExample/android/gradle/wrapper/gradle-wrapper.properties
kligarski, line 1: Don't we also need to apply the same update to library-level `gradle-wrapper.properties`?

android/build.gradle
```diff
@@ -165,6 +177,11 @@ android {
     }
     sourceSets.main {
         ext.androidResDir = "src/main/res"
+        kotlin {
+            srcDirs += [
+                "src/fabric/java",
+            ]
+        }
```
kligarski: We can consider applying also some changes from https://github.com/software-mansion/react-native-screens/pull/4174 like using `.directories.add` instead of `srcDirs`.

ios/stack/host/RNSStackHostComponentView.mm
```diff
@@ -31,6 +31,8 @@ @implementation RNSStackHostComponentView {
 - (instancetype)initWithFrame:(CGRect)frame
 {
   if (self = [super initWithFrame:frame]) {
+    static const auto defaultProps = std::make_shared<const react::RNSStackHostProps>();
+    _props = defaultProps;
     [self initState];
```
kligarski: nit: in component when we already have `initState`/`resetProps`, maybe we should move the `... _props = defaultProps;` to there?

src/legacy/types.ts
kligarski: If we're going to export this symbol, maybe we should also rename it now? Because `TextInputFocusEventData` doesn't really make much sense for `onChangeText`. Or we can inline the type to the prop. What do you think?

src/fabric/split/SplitHostNativeComponent.ts
kligarski: Should this be also changed to `ComponentRef`?

src/legacy/components/ScreenStackHeaderConfig.tsx
```diff
@@ -243,8 +248,6 @@ const styles = StyleSheet.create({
     width: '100%',
     flexDirection: 'row',
     justifyContent: 'space-between',
-    // We only want to center align the subviews on iOS.
-    // See https://github.com/software-mansion/react-native-screens/pull/2456
```
kligarski, line 247: Is the removal of this comment intended?

kligarski (commented):
Looks good but I left a few minor comments that we should look at before proceeding.

tsconfig.json
t0maboro: The RN 0.87 upgrade helper removed `"types": ["jest"]`, because it's included in `@react-native/typescript-config`. Defining a types array in tsconfig overwrites the base config instead of merging with it, and we additionally need `node` here (using `process` in e2e tests), so we need to leave `jest` on our side.

FabricExample/android/gradle/wrapper/gradle-wrapper.properties
t0maboro: I decided not to bump because it hasn't been bumped in many previous RN upgrades; tbh, not sure whether we should consider bumping that, but I think this might be another topic - the Kotlin version for the library is still set to `1.8.0` while on the example app's side it is 2+. I'd leave it as-is for now.

android/build.gradle
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4375/changes/afe0ffbdaf7564d19fdcae24dbe9d96b9485563b

ios/stack/host/RNSStackHostComponentView.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4375/changes/cd63f7b9d19b6d34aabaf63e63b21d78c1d8f385

src/fabric/split/SplitHostNativeComponent.ts
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4375/changes/0aa37442e86962dfdbe73e683f0aa6b9304456fa

src/legacy/components/ScreenStackHeaderConfig.tsx
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4375/changes/663cd363121a144b7a56c0c34d8f201599f7e9e3

src/legacy/types.ts
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4375/commits/681763e766be9e4c1d8774039c5f06b704432286

kligarski (approved):
Looks good. I haven't checked the runtime after the changes.

Issue #1459, opened 2022-05-26
The type NativeStackNavigationProp has a missing type argument


Yupeng-li:
## Description

The type `NativeStackNavigationProp` has a missing type argument.  [(see the code) ](https://github.com/software-mansion/react-native-screens/blob/90e548739f35b5ded2524a9d6410033fc233f586/src/native-stack/types.tsx#L41)

`NavigationProp` [(see the code)](https://github.com/react-navigation/react-navigation/blob/4e4935ac2584bc1a00209609cc026fa73e12c10a/packages/core/src/types.tsx#L305) expects 6 arguments, but only 5 are given. `NavigatorID` is missing. This caused an issue that `navigation.setOptions` doesn't pickup correct type of screen options. 

### Screenshots
![image](https://user-images.githubusercontent.com/31921080/170520590-f774f050-81ae-4370-bb56-bf22b0b30fe3.png)

## Steps To Reproduce
The details are included in the description section above. 

### Expected behavior
I expect `navigation.setOptions` to use correct type

### Actual behavior
`navigation.setOptions` always shows type error because of the issue.

## Reproduction

I think the error is obvious enough without having a minimal repo.

## Platform

- [x] iOS
- [x] Android
- [ ] Web
- [ ] Windows
- [ ] tvOS

## Workflow

- [ ] Managed workflow
- [x] Bare workflow

## Package versions

| package                                | version |
| -------------------------------------- | ------- |
| react-native                           |    0.66.3     |
| @react-navigation/native               |    6.0.10     |
| @react-navigation/native-stack         |    6.6.2     |
| react-native-screens                   |    3.13.1     |
| react-native-safe-area-context         |    4.2.5     |
| react-native-gesture-handler           |    2.4.2     |
| react-native-reanimated                |         |
| expo                                   |         |

WoLewicki:
Are you using `@react-navigation/native-stack` or `react-native-screens/native-stack`? You should use the first one if you use `react-navigation` v6 and the second one for v5.

Yupeng-li:
Thank you for pointing this out. I used `react-native-screens/native-stack`. Once I switched to `@react-navigation/native-stack`, it worked like a charm 👍 . I'll close this issue.

PR #4393, opened 2026-07-23
chore(test, Stack v5): add simple navigation e2e test

LKuchno:
## Summary

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1687

Adds Detox end-to-end coverage for the **Stack v5 simple-navigation** scenario, plus the small test-screen/shared-component changes needed to make it observable. The e2e lives in a single file with two `describeIf*` suites because the two platforms behave differently in the example-app harness and need different matchers.

### Coverage

- **iOS** (`describeIfiOS`) — full scenario: push/pop, `routeKey` uniqueness and preservation, the native header back button, and the edge-swipe pop. react-native-screens detaches covered screens on iOS, so every matcher resolves to the top screen.
- **Android** (`describeIfAndroid`) — button-driven push/pop and `routeKey` uniqueness only. Covered screens stay attached on Android, so reads/taps/waits target the **topmost** match. The native header back button and system gesture-back do not pop the nested gamma `StackContainer` when the screen is opened from the example app (see [#1459](https://github.com/software-mansion/react-native-screens-labs/issues/1459)); those are verified manually via the direct `App.tsx` launch documented in the scenario.

## Changelog

- **New** `FabricExample/e2e/single-feature-tests/stack-v5/test-stack-simple-nav.e2e.ts` — per-platform e2e suites.
- `StackRouteInformation.tsx` — add `stack-route-name` / `stack-route-key` testIDs so the current route and its `routeKey` can be read.
- `test-stack-simple-nav/index.tsx` — `screenB-layout-view` testID (for the iOS edge-swipe).
- `scenario.md` / `scenario-description.ts` — document automated vs manual coverage; set `e2eCoverage` from `tbd` to `incomplete`.
- `native-class-names.ts` — constant used by the iOS back-button matcher.

## Notes on robustness

- Android's core `<Button>` uppercases titles (`PUSH A`, `POP`), so the Android suite matches them in that form.
- Same-route transitions (`A -> A` push/pop) are gated on the `routeKey` changing (`waitForKeyChange`) rather than the route name, which is unchanged across such transitions — this avoids a race where a read/tap lands before the transition settles.


apps/src/tests/single-feature-tests/stack-v5/test-stack-simple-nav/scenario.md
```diff
@@ -18,11 +18,26 @@ Notes for `routeKey` behavior and how the two platforms are launched
 
 ## E2E test
 
-TBD: Automation is planned and should be straightforward for the
-button-driven push/pop steps (similar in scope to the tabs
-`test-tabs-simple-nav` suite). Native back button, iOS edge-swipe gesture,
-and Android system gesture-back steps may need platform-specific handling
-and are not yet implemented.
+Incomplete: Each block contains a per-platform suite, as the two platforms behave
+differently within the example app harness and require distinct matchers.
```
kligarski:
```suggestion
Incomplete.
```

I don't think we need the details here. From tester's perspective, the most important here is what steps are automated and which aren't and it is described below.

apps/src/tests/single-feature-tests/stack-v5/test-stack-simple-nav/scenario.md
```diff
@@ -18,11 +18,26 @@ Notes for `routeKey` behavior and how the two platforms are launched
...
+
+- iOS: steps 1–10, including the native header back button
+  (step 8) and the edge-swipe pop (step 9). Since react-native-screens detaches
+  covered screens on iOS, `Name:` / `Key:` or button matcher resolves to
+  a single element - the top screen.
```
kligarski:
```suggestion
  (step 8) and the edge-swipe pop (step 9).
```
Same here - I don't think we need these details here.

apps/src/tests/single-feature-tests/stack-v5/test-stack-simple-nav/scenario.md
```diff
@@ -18,11 +18,26 @@ Notes for `routeKey` behavior and how the two platforms are launched
...
-TBD: Automation is planned and should be straightforward for the
-button-driven push/pop steps (similar in scope to the tabs
-`test-tabs-simple-nav` suite). Native back button, iOS edge-swipe gesture,
-and Android system gesture-back steps may need platform-specific handling
-and are not yet implemented.
+Incomplete: Each block contains a per-platform suite, as the two platforms behave
+differently within the example app harness and require distinct matchers.
+
+- iOS: steps 1–10, including the native header back button
+  (step 8) and the edge-swipe pop (step 9). Since react-native-screens detaches
```
kligarski: Here we have information that step 9 is automated (I guess at least, maybe we should change first part to `steps 1–10 are automated`) but below in `Manual only` section step 9 is described not to be automated.

apps/src/tests/single-feature-tests/stack-v5/test-stack-simple-nav/scenario.md
```diff
@@ -18,11 +18,26 @@ Notes for `routeKey` behavior and how the two platforms are launched
...
+  covered screens on iOS, `Name:` / `Key:` or button matcher resolves to
+  a single element - the top screen.
+- Android (`describeIfAndroid`): steps 1–7 and 10 only. Unlike iOS, covered screens
+  remain attached on Android, causing stacked screens to render duplicate buttons and labels.
+  Consequently, the suite inspects the current route, taps buttons, and asserts on
+  the topmost match (the last item in the hierarchy). Additionally, `<Button>`
+  titles also render uppercased (`PUSH A`, `POP`) and are matched in that form.
```
kligarski:
```suggestion
- Android (`describeIfAndroid`): steps 1–7 and 10 only.
```
Same here - do we need these details in scenario file? I think that it would be enough to have a comment in `.e2e.ts`

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-simple-nav.e2e.ts
```diff
@@ -0,0 +1,384 @@
...
+   * unambiguously to the current screen.
+   */
+  async function readRouteKey(): Promise<string> {
+    const attrs = (await element(
+      by.id('stack-route-key'),
+    ).getAttributes()) as AnyAttributes;
+    const value =
+      (attrs as { text?: string }).text ??
+      (attrs as { label?: string }).label ??
+      '';
```
kligarski: Wouldn't it be better to use `getElementAttributes` from `e2e-utils`? Then we have type safety and nice warning instead of unsafe casting.


```suggestion
    const attrs = await getElementAttributes({
      by: 'id',
      value: 'stack-route-key',
    });
    const value = attrs.text ?? attrs.label ?? '';
```

plus import (not included in suggestion)

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-simple-nav.e2e.ts
```diff
@@ -0,0 +1,384 @@
...
+      const key = await readRouteKey();
+      if (key !== '' && key !== previousKey) {
+        return;
+      }
+      await new Promise(resolve => setTimeout(resolve, 100));
+    }
+    throw new Error(
+      `waitForKeyChange timed out; topmost key is still "${previousKey}"`,
+    );
+  }
```
kligarski, line 90: Detox supports matching by regex (https://wix.github.io/Detox/docs/api/matchers#regex-matching) - can't we use that instead?

This is how route keys are generated:

```
// apps/src/shared/containers/stack/reducer.tsx

function generateRouteKeyForRouteName(routeName: string): string {
  return `r-${routeName}-${generateID()}`;
}
```

Claude suggested something like this (but it might not work):
```
await waitFor(element(by.id(new RegExp(`^r-${routeName}-(?!${previousId}$).*$`))))
  .toBeVisible()
  .withTimeout(3000);
```

It should wait for element with same `routeName` but different id than `previousId`.

At the end of the day if current version works and there are some problems with the regex we can keep the current version.

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-simple-nav.e2e.ts
```diff
@@ -0,0 +1,384 @@
...
+
+  /** Reads the `Key`/`Name` label text from the topmost stacked screen. */
+  async function readTopmostText(testID: string): Promise<string> {
+    const matches = await getMatches(by.id(testID));
+    const top = matches[matches.length - 1];
+    return (
+      (top as { text?: string }).text ??
+      (top as { label?: string }).label ??
+      ''
+    ).trim();
```
kligarski:
```suggestion
    return (top.text ?? top.label ?? '').trim();
```
unnecessary cast

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-simple-nav.e2e.ts
kligarski: I guess that it might be hard to use it for Android version so I'm not sure, maybe we can keep the current version.

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-simple-nav.e2e.ts
```diff
@@ -0,0 +1,384 @@
...
+
+  it('should push A with a new key and reveal the Pop button', async () => {
+    await tapTopmostButton(PUSH_A);
+    await waitForRoute('A');
+    firstAKey = await readRouteKey();
+    jestExpect(firstAKey).not.toBe(homeKey);
+    await expect(element(by.text(POP))).toBeVisible();
+  });
+
+  it('should push B on top of A with a new key', async () => {
```
kligarski, line 330:
```suggestion
  // eslint-disable-next-line jest/no-identical-title -- Android and iOS have separate test suites
  it('should push B on top of A with a new key', async () => {
```

We might ignore the warning here and below if you want to.

Image

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-simple-nav.e2e.ts
```diff
@@ -0,0 +1,384 @@
...
+ *   except rapid tapping.
+ * - Android: covered screens stay attached, so a matcher can resolve to one
+ *   element per stacked screen and must be normalized to the topmost match.
+ *   In addition, this screen is opened through the example app's own
+ *   navigation (not launched directly via `App.tsx`), so the native header
+ *   back button and the system gesture-back do not pop the nested gamma
+ *   `StackContainer` — see issue #1459. The Android suite therefore covers
+ *   only navigation driven by the on-screen Push/Pop buttons; native-back and
+ *   gesture-back are verified on iOS and manually on Android via the direct
+ *   launch documented in the scenario.
```
kligarski, line 34: We did the "branch cut" today, maybe we should consider either migrating example app to Stack v5 or maybe thinking of some other solution for e2e? I remember that I looked into the topic some time ago but maybe we can just do a "search bar" but purely in JS - screen with text input and "go" button. After the button is tapped, we would display only the test example with key from the input. Maybe this would reduce CI times as well (the time that we spend on scrolling/native search bar).

Obviously it's out of scope of this PR.

kligarski (approved):
I left some suggestions. The one with regex maybe isn't necessary as we won't be able to handle Android the same way I think.

apps/src/tests/single-feature-tests/stack-v5/test-stack-simple-nav/scenario.md
LKuchno: https://github.com/software-mansion/react-native-screens/pull/4393/changes/b138faa80786364ed63ed1d7f91f26a8db3dafc9

apps/src/tests/single-feature-tests/stack-v5/test-stack-simple-nav/scenario.md
LKuchno: https://github.com/software-mansion/react-native-screens/pull/4393/changes/b138faa80786364ed63ed1d7f91f26a8db3dafc9

apps/src/tests/single-feature-tests/stack-v5/test-stack-simple-nav/scenario.md
LKuchno: I shorten details and add information about what was covered by e2e for step9 on iOS
https://github.com/software-mansion/react-native-screens/pull/4393/changes/b138faa80786364ed63ed1d7f91f26a8db3dafc9

apps/src/tests/single-feature-tests/stack-v5/test-stack-simple-nav/scenario.md
LKuchno: https://github.com/software-mansion/react-native-screens/pull/4393/changes/b138faa80786364ed63ed1d7f91f26a8db3dafc9

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-simple-nav.e2e.ts
LKuchno: Good catch, thank you!!
[applying suggestion to use getAttribute from e2e-utils](https://github.com/software-mansion/react-native-screens/pull/4393/changes/442c094babe38a77b570861a01131c536f05ea85)

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-simple-nav.e2e.ts
LKuchno: [Comment about duplicated test titles](https://github.com/software-mansion/react-native-screens/pull/4393/changes/74c2d5540e894c1abe352fdf3a1a64a28f57fdb7)

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-simple-nav.e2e.ts
LKuchno: I applied this suggestion to function used in iOS for Android function `waitForKeyChange` stayed unchanged [changing function for routeKey change to use regex for iOS](https://github.com/software-mansion/react-native-screens/pull/4393/changes/ae7e95bab1891da0d468317c2b7fcdc08de1b2d5)

PR #4401, opened 2026-07-24
chore: Remove nightly publishing flows from v4 and bump npm-package-publish

kmichalikk:
## Description

This PR removes nightly publishing from v4 branch. I'm not refactoring this entirely, just removing what we don't need, in order to not introduce bugs.

I decided to also bump npm-package-publish to match current (as of writing) version on main branch. I don't think something crucial changed, but let's have it consistent.

## Changes

- update publishing workflow
- bumped npm-package-publish

n/a

## Test plan

n/a
## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/.github/workflows/publish-npm.yml
+++ b/.github/workflows/publish-npm.yml
@@ -1,7 +1,5 @@
 name: Publish release to npm
 on:
-  schedule:
-    - cron: "27 23 * * *" # at 23:27 every day
   workflow_dispatch:
     inputs:
       dry-run:
@@ -15,7 +13,6 @@ on:
         type: choice
         options:
           - stable
-          - nightly
           - beta
           - rc
         default: stable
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-npm.yml
```diff
@@ -69,7 +66,7 @@ jobs:
 
       - name: Publish manual release
         if: ${{ github.event_name == 'workflow_dispatch' }}
-        uses: software-mansion/npm-package-publish@264013301cc21350186b190f1f6dd10ae4c8ee04
+        uses: software-mansion/npm-package-publish@bbfbcdf3f27e2dc288d7d56f5989a731efc79e6f
```
kkafar, line 69: My only concern is why do we update the version of the workflow? Do we have any reasons to do so? 

Even if we don't - I'm for it, but let's state it clearly in the PR description.

kkafar (commented):
Left some comment. Please answer it before proceeding.

.github/workflows/publish-npm.yml
kmichalikk: yes, this is just package bump since we left it stale when moving to 5.0

I updated the PR

kkafar: Okay! thanks

kkafar (approved):
Thanks!

PR #4403, opened 2026-07-24
chore: Proposal for the file structure organization for FormSheet component on android

t0maboro:
## Description

This PR is a proposal to reorganize the Android form-sheet implementation from a flat directory structure into nested architectural layers.

Currently, the library separates other features into per-component directories. This PR introduces a per-domain (layered) separation instead. The primary driver for this shift is that "react" is just one of many potential consumers of the "native" container. By restructuring, we establish strict boundaries that clearly decouple the core native logic from the React Native integrations.

Proposed Structure
- `native/`: Core Native BottomSheet logic (the dialog, lifecycle manager, behavior/appearance coordinators, enter/exit animations). **Rule: No React Native dependencies should ever be introduced in this directory.**
- `react/`: React Native bridging (host view, view managers, event emitters). This handles the integration of RN-based views with the native layer.

Future Scope
- Introduce an additional layer/directory for native test setup.
- Apply this same structural convention to other modal components across the library.

I've marked this as a proposal because I'd like to use this PR to align on the proper directory structure in the modals context. Since the intent is to adopt this pattern for other modals moving forward, first I want to hear your feedback and discuss potential risks before proceeding.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1620

## Changes

- Reorganizing the flat structure into a hierarchical one

N/A

## Test plan

No functional changes

## Checklist

- [ ] Ensured that CI passes


kligarski (approved):
Seems like a good idea.

The only thing I'm thinking about is whether we need `shared` directory. I think that this should just be part of `native`. I see this as "our React implementation uses native implementation" so native implementation shouldn't be aware of React implementation but React implementation can use native as needed.

What do you think?

t0maboro:
> Seems like a good idea.
> 
> The only thing I'm thinking about is whether we need `shared` directory. I think that this should just be part of `native`. I see this as "our React implementation uses native implementation" so native implementation shouldn't be aware of React implementation but React implementation can use native as needed.
> 
> What do you think?

Yep, moving that shared inside `native/interfaces` and `native/model` seems like a good idea, as you've already pointed out that I need to import "native" components to "react" context anyway. I'll update that PR soon.

PR #4404, opened 2026-07-24
feat(Android, FormSheet v5): Separate onDismiss and onNativeDismiss events

t0maboro:
## Description

Splits the Android FormSheet dismissal callback into two distinct events, matching the origin:
- onDismiss - the sheet was dismissed programmatically (JS prop change).
- onNativeDismiss - the sheet was dismissed by an user action (swiping down, pressing back, or tapping the backdrop).

> [!CAUTION]
> I started noticing an occasional content flickering, probably because of wrong timing of the translateY application before starting the enter transition. Noticed that it's reproducible on the main branch as well, so I'll handle that separately: https://github.com/software-mansion/react-native-screens-labs/issues/1689

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1559

## Changes

- new `FormSheetDismissalOrigin` enum;
- `FormSheetPresentationManager` holds `dismissalOrigin` (UNSPECIFIED by default); detected dismissal requests are updating the value to desired dismissal origin
- onDismissComplete() emits the event relatively to the origin

https://github.com/user-attachments/assets/6d5ca1d1-40ae-44dc-a541-117bf76b84c3

## Test plan

Marked SFT as applicable on android.

## Checklist

- [ ] Ensured that CI passes


android/src/main/java/com/swmansion/rnscreens/modals/formsheet/native/presentation/FormSheetDismissalOrigin.kt
```diff
@@ -0,0 +1,16 @@
...
+     */
+    PROGRAMMATIC,
+
+    /**
+     * The dismissal was initiated by the user (swiping the sheet down,
+     * triggering the native back button, or tapping the backdrop).
+     * Emits `onNativeDismiss`.
+     */
+    NATIVE,
+}
```
kligarski, line 22: In Tabs we use these origins:

```kotlin
enum class TabsActionOrigin {
    USER,
    PROGRAMMATIC_JS,
    PROGRAMMATIC_NATIVE,
    ;
    // ...
}
```

Maybe in FormSheet we should use `PROGRAMMATIC_JS, USER` (for now, with possibility to add `PROGRAMMATIC_NATIVE` in the future)?

kligarski: But that will be odd with `onDismiss` and `onNativeDismiss` events I guess. Maybe the current version is okay.

kligarski (approved):
Looks good, we should only have a look at the one edge case we discussed internally.

android/src/main/java/com/swmansion/rnscreens/modals/formsheet/native/presentation/FormSheetDismissalOrigin.kt
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4404/commits/4d71bdbb73eb21411c56948b455a31c13dcef60b

t0maboro:
> Looks good, we should only have a look at the one edge case we discussed internally.

https://github.com/software-mansion/react-native-screens/pull/4404/changes/e0ef97ad5861a379dc380d572ea62c4e84309f3a

PR #4407, opened 2026-07-27
fix(iOS): Don't swap view controller's view during transition

### LINKED ISSUE #4361: [iOS 26][Fabric] Tab press during native-stack pop finalization leaves tab content permanently untouchable (touches never reach JS)
### Description

On iOS 26 (New Architecture), pressing a bottom-tab while a native-stack **pop transition is still finalizing** permanently breaks touch handling for the tab's content. The tab screen renders normally and animations keep running, but **no touch ever reaches the JS responder system again** (verified: not even a root `onTouchStart` fires). The tab bar itself keeps working because it lives outside the `ScreenContainer`. Switching tabs away and back does **not** recover — the screen stays untouchable until the app is restarted.

This appears to be enabled by the iOS 26 `ios26AllowInteractionsDuringTransition` behavior (interactions no longer blocked during stack transitions): the tab press now lands *during* the pop finalization, and the resulting `ScreenContainer` activity-state update (detach previous tab / attach pressed tab) races the Fabric unmount of the popped stack screen, leaving the container's native state inconsistent for hit-testing.

Setting `featureFlags.experiment.ios26AllowInteractionsDuringTransition = false` on 4.25.2 (where the flag is still configurable) **completely eliminates the bug** — which is strong evidence the interaction-during-transition path is the trigger. On 4.26.2 the flag is hardcoded to `true`, so there is no way to opt out anymore.

Likely related (same reentrancy family, crashes instead of hanging): #4168 (`RNSScreenContainer.updateContainer` mutated-while-enumerating under reentrant Fabric mounting).

#### Navigation structure

```
native-stack (RootStack)
├── Screen: bottom-tabs navigator  ← JS @react-navigation/bottom-tabs
│   ├── Tab A (Home)
│   ├── Tab B (MyPets)
│   └── Tab C (Profile)
└── Screen: CalendarScreen (regular push)
```

#### Timing evidence (device console timestamps)

Round that **worked** — tab press arrived ~30 ms *after* the popped screen unmounted:

```
20:13:30.938  CalendarScreen unmount
20:13:30.968  tabPress MyPetsScreen
20:13:30.972  transitionEnd (tabs host screen)
20:13:31.240  root onTouchStart fires  ✅ touches OK
```

Round that **froze** — tab press raced the pop finalization (1 ms after unmount, *before* transitionEnd):

```
20:13:34.154  CalendarScreen unmount
20:13:34.155  tabPress MyPetsScreen      ← lands during transition finalization
20:13:34.177  transitionEnd (tabs host screen)
(afterwards)  taps on tab content produce ZERO JS touch events  ❌ frozen forever
```

#### Ruled out during bisection

- `@react-navigation/*` versions — reproduced identically on bottom-tabs 7.15.9 and 7.18.11 (native 7.2.2 → 7.3.11, native-stack 7.14.11 → 7.18.3)
- react-native-screens 4.25.2 **and** 4.26.2 — both reproduce
- `detachInactiveScreens={false}` on the tab navigator — still reproduces
- App JS code — in the frozen state no touch event reaches JS at all (root `onTouchStart` interceptors never fire), so the block is native-side

### Steps to reproduce

1. Native-stack with a bottom-tabs navigator as one screen and a regular pushed screen on top (structure above), running on an **iOS 26** device (New Architecture).
2. From tab A, push the stack screen.
3. Go back (header back button).
4. **Immediately** (within ~200 ms, while the pop transition is finalizing) press tab B.
5. Tab B renders but its content no longer responds to any touch. Tab bar still works; switching tabs back and forth does not recover.

Timing-sensitive: if the tab press lands after `transitionEnd`, everything works. Pressing the tab as fast as possible after the back navigation reproduces reliably within a few attempts on a physical device.

(can provide a minimal repro repo if needed — the structure is exactly the tree above)

### Screens version

4.25.2 and 4.26.2 (both reproduce)

### React Native version

0.86.0 (Expo SDK 57, New Architecture / Fabric, React Compiler enabled)

### Platforms

iOS

### Additional info

- Device: iPhone 15 Pro Max, iOS 26.5.2 (physical device; could not reproduce on simulator)
- Built with Xcode 26.6
- `@react-navigation/native` 7.3.11, `@react-navigation/bottom-tabs` 7.18.11 (also reproduced on 7.15.9)
- Workarounds found:
  - screens 4.25.2 + `featureFlags.experiment.ios26AllowInteractionsDuringTransition = false` → bug gone, but touches are blocked during **all** transitions (pre-iOS-26 UX) and the flag was removed in 4.26
  - App-level: `preventDefault()` on `tabPress` while the root stack is between `transitionStart`/`transitionEnd`, replaying the press when the transition ends — this is what we shipped

kmichalikk:
Fixes https://github.com/software-mansion/react-native-screens/issues/4361

## Description

This PR fixes empty native stack screen view hanging after overlapping stack / tabs transitions. For full problem description see the linked issue. Running the provided repro with added lifecycle logs, the following happens when trying to pop the screen with back button and quickly tapping the second tab before the stack dismissed the top screen fully:

```
** 1. **
screen 'Calendar-S-6H596y9u5jAlqieB7Of' willMoveToParentViewController (null)!
Unable to simultaneously satisfy constraints.
[...]
[DEBUG] screen 'Calendar-S-6H596y9u5jAlqieB7Of' viewWillDisappear!
[DEBUG] screen 'Tabs-r0nvm3owqzZ0ngsCQLvpW' viewWillAppear!
[DEBUG] screen 'Tabs-r0nvm3owqzZ0ngsCQLvpW' did move to window <UIWindow: 0x104b9b660; frame = (0 0; 390 844); gestureRecognizers = <NSArray: 0x1263c5080>; layer = <UIWindowLayer: 0x1282d6340>>, superview <UIView: 0x1263ce680; frame = (0 0; 390 844); clipsToBounds = YES; autoresize = W+H; layer = <CALayer: 0x1263c6c10>>
TabsScreen [2066] onWillAppear received
TabBar: <UITabBar: 0x128381680; frame = (0 761; 390 83); autoresize = W+TM; tintColor = UIExtendedSRGBColorSpace 0 0.478431 1 1; gestureRecognizers = <NSArray: 0x1256e2080>; layer = <CALayer: 0x125d31ce0>> standardAppearance=0x11b0b2520 didSelectItem: <UITabBarItem: 0x12835fa80> title='My Pets' standardAppearance=0x11b0b22e0 scrollEdgeAppearance=0x11b0b2340 selected
TabsHost [2072] onTabSelected: {"selectedScreenKey":"MyPets-lV7gQMdoOKmHOwiYuK8BO","provenance":1,"isRepeated":false,"hasTriggeredSpecialEffect":false,"actionOrigin":"user","target":2072,"timeStamp":76803034.52725}
TabsScreen [2068] onWillAppear received
TabsScreen [2066] onWillDisappear received
TabsHost render
TabsScreen [2066] render; screenKey: Home-EJSjwXWA38st7IPuzYkHS
TabsScreen [2068] render; screenKey: MyPets-lV7gQMdoOKmHOwiYuK8BO
TabsScreen [2070] render; screenKey: Profile-CneTd9bfupq-UefLAhFf5
I0727 11:22:56.968006 1837248512 UIManagerBinding.cpp:143] instanceHandle is null, event of type topTransitionProgress will be dropped
TabBarCtrl mountintTransactionWillMount
** 2. **
[DEBUG] screen 'Calendar-S-6H596y9u5jAlqieB7Of' did move to window (null), superview (null)
[DEBUG] screen snapshot taken!
TabBarCtrl mountintTransactionDidMount running updates
** 3. **
TabBarCtrl updateSelectedViewController
[DEBUG] screen 'Calendar-S-6H596y9u5jAlqieB7Of' did move to window <UIWindow: 0x104b9b660; frame = (0 0; 390 844); autoresize = W+H; gestureRecognizers = <NSArray: 0x1263c5080>; layer = <UIWindowLayer: 0x1282d6340>>, superview <UIViewControllerWrapperView: 0x12620d500; frame = (0 0; 390 844); layer = <CALayer: 0x125786c40>>
** 4. **
[DEBUG] screen 'Calendar-S-6H596y9u5jAlqieB7Of' viewDidDisappear!
[DEBUG] screen 'Calendar-S-6H596y9u5jAlqieB7Of' dismissed event sent!
screen 'Calendar-S-6H596y9u5jAlqieB7Of' didMoveToParentViewController (null)!
** 5. **
[DEBUG] screen 'Tabs-r0nvm3owqzZ0ngsCQLvpW' viewDidAppear!
TabsScreen [2068] onDidAppear received
TabsScreen [2066] onDidDisappear received
** 6. **
I0727 11:22:59.416013 1837248512 UIManagerBinding.cpp:143] instanceHandle is null, event of type topTouchEnd will be dropped
I0727 11:23:00.200285 1837248512 UIManagerBinding.cpp:143] instanceHandle is null, event of type topTouchEnd will be dropped
I0727 11:23:01.017321 1837248512 UIManagerBinding.cpp:143] instanceHandle is null, event of type topTouchEnd will be dropped
```
1. top screen is being popped with back button, we get lifecycle events for Calendar screen moving out and Tabs screen moving in
2. Calendar screen is detached from window and replaced with snapshot as the controller.view
3. active tab is switched - mid screen transition; **but the screen view is attached to window again, in another parent and this is the view that causes the bug**
4. Calendar screen is fully dismissed from the stack point of view and dismiss event is sent (viewDidX, viewWillX are methods on the controller)
5. transition between tabs is fully completed
6. we can see the bug - clicks are not registered

---

The bug seems to be caused by the snapshot function. If controller.view is not replaced with a snapshot, there is no bug. With a breakpoint inside the problematic `didMoveToWindow` we see the following trace:

image

These are UIKit internals. Supposedly, UIKit tracks both "from" and "to" views during transition, and changing them causes UIKit to readd them in the wrong place. I found out that adding the snapshot inside the already empty screen view instead of replacing it is a possible solution and I decided to move forward with it.

## Changes

- instead of replacing screen controller's view with snapshot directly, the snapshot is added as a child of the view

No visual changes. See the logs in the repro instead: after performing the navigation, each press should be logged correctly.

## Test plan

Use Test4361.

> [!important]
> To test this you need physical device.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/legacy/RNSScreen.mm
+++ b/ios/legacy/RNSScreen.mm
@@ -1991,27 +1991,22 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
 
 #pragma mark - Fabric specific
 
-- (BOOL)setViewToSnapshot
+- (void)setViewToSnapshot
 {
   // if we dismissed the view natively, it will already be detached from view hierarchy
   if (self.view.window == nil) {
-    return NO;
-  }
-
-  UIView *_Nullable snapshot = [self.view snapshotViewAfterScreenUpdates:self.screenView.snapshotAfterUpdates];
-  if (snapshot == nil) {
-    return NO;
+    return;
   }
 
   // The snapshot is added inside the screen view instead of replacing it, so `self.view` remains
   // the view UIKit captured for the ongoing transition and is torn down (with the snapshot)
   // consistently. Replacing the view mid-transition made iOS 26 teardown re-insert the original
   // view into the hierarchy and remove only the snapshot, leaving a view that blocked all touches.
+  UIView *snapshot = [self.view snapshotViewAfterScreenUpdates:self.screenView.snapshotAfterUpdates];
   snapshot.frame = self.view.bounds;
   snapshot.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
   snapshot.userInteractionEnabled = NO;
   [self.view addSubview:snapshot];
-  return YES;
 }
 
 @end
--- a/ios/legacy/RNSScreen.mm
+++ b/ios/legacy/RNSScreen.mm
@@ -1753,9 +1753,13 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
 
 - (void)notifyTransitionProgress:(double)progress closing:(BOOL)closing goingForward:(BOOL)goingForward
 {
+  if (![self.view isKindOfClass:[RNSScreenView class]]) {
+    return;
+  }
+
   // if the screen was already deleted by React, there is no sense in sending progress
   // since on JS side the component is already not present
-  if ([self.view isKindOfClass:[RNSScreenView class]] && !((RNSScreenView *)self.view).isInvalidated) {
+  if (!static_cast<RNSScreenView *>(self.view).isInvalidated) {
     [(RNSScreenView *)self.view notifyTransitionProgress:progress closing:closing goingForward:goingForward];
   }
 }
@@ -2004,6 +2008,7 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
   // view into the hierarchy and remove only the snapshot, leaving a view that blocked all touches.
   UIView *snapshot = [self.view snapshotViewAfterScreenUpdates:self.screenView.snapshotAfterUpdates];
   snapshot.frame = self.view.bounds;
+  // fill the whole screen view with the snapshot
   snapshot.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
   [self.view addSubview:snapshot];
 }
--- a/ios/legacy/RNSScreen.mm
+++ b/ios/legacy/RNSScreen.mm
@@ -769,6 +769,8 @@ RNS_IGNORE_SUPER_CALL_END
 
 - (void)invalidateImpl
 {
+  _invalidated = YES;
+
   // Since the scroll view might get immediately recycled we remove ourselves
   // immediately.
   if (_sheetsScrollView != nil) {
@@ -1751,9 +1753,9 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
 
 - (void)notifyTransitionProgress:(double)progress closing:(BOOL)closing goingForward:(BOOL)goingForward
 {
-  if ([self.view isKindOfClass:[RNSScreenView class]]) {
-    // if the view is already snapshot, there is not sense in sending progress since on JS side
-    // the component is already not present
+  // if the screen was already deleted by React, there is no sense in sending progress
+  // since on JS side the component is already not present
+  if ([self.view isKindOfClass:[RNSScreenView class]] && !((RNSScreenView *)self.view).isInvalidated) {
     [(RNSScreenView *)self.view notifyTransitionProgress:progress closing:closing goingForward:goingForward];
   }
 }
@@ -1907,8 +1909,6 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
   return (int)[[self.screenView.reactSuperview reactSubviews] indexOfObject:view];
 }
 
-// Since the view of the controller can be a snapshot of type `UIView`,
-// when we want to check props of ScreenView, we need to get them from _initialView
 - (RNSScreenView *)screenView
 {
   return _initialView;
@@ -1991,18 +1991,27 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
 
 #pragma mark - Fabric specific
 
-- (void)setViewToSnapshot
+- (BOOL)setViewToSnapshot
 {
-  UIView *superView = self.view.superview;
   // if we dismissed the view natively, it will already be detached from view hierarchy
-  if (self.view.window != nil) {
-    auto afterUpdates = self.screenView.snapshotAfterUpdates;
-    UIView *snapshot = [self.view snapshotViewAfterScreenUpdates:afterUpdates];
-    snapshot.frame = self.view.frame;
-    [self.view removeFromSuperview];
-    self.view = snapshot;
-    [superView addSubview:snapshot];
+  if (self.view.window == nil) {
+    return NO;
   }
+
+  UIView *_Nullable snapshot = [self.view snapshotViewAfterScreenUpdates:self.screenView.snapshotAfterUpdates];
+  if (snapshot == nil) {
+    return NO;
+  }
+
+  // The snapshot is added inside the screen view instead of replacing it, so `self.view` remains
+  // the view UIKit captured for the ongoing transition and is torn down (with the snapshot)
+  // consistently. Replacing the view mid-transition made iOS 26 teardown re-insert the original
+  // view into the hierarchy and remove only the snapshot, leaving a view that blocked all touches.
+  snapshot.frame = self.view.bounds;
+  snapshot.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
+  snapshot.userInteractionEnabled = NO;
+  [self.view addSubview:snapshot];
+  return YES;
 }
 
 @end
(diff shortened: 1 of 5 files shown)


ios/legacy/RNSScreen.mm
```diff
@@ -1991,18 +1991,21 @@ - (RNSOrientation)evaluateOrientation
...
+    return;
   }
+
+  // The snapshot is added inside the screen view instead of replacing it, so `self.view` remains
+  // the view UIKit captured for the ongoing transition and is torn down (with the snapshot)
+  // consistently. Replacing the view mid-transition made iOS 26 teardown re-insert the original
+  // view into the hierarchy and remove only the snapshot, leaving a view that blocked all touches.
+  UIView *snapshot = [self.view snapshotViewAfterScreenUpdates:self.screenView.snapshotAfterUpdates];
+  snapshot.frame = self.view.bounds;
+  snapshot.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
```
t0maboro, line 2010: This is newly added, so I think it might need some clarification in the code on why we didn't need that in legacy logic and why we need it now.

ios/legacy/RNSScreen.mm
```diff
@@ -1907,8 +1909,6 @@ - (int)getIndexOfView:(UIView *)view
   return (int)[[self.screenView.reactSuperview reactSubviews] indexOfObject:view];
 }
 
-// Since the view of the controller can be a snapshot of type `UIView`,
-// when we want to check props of ScreenView, we need to get them from _initialView
```
t0maboro, line 1911: Is this commit no longer valid at all? I think we may lose some context completely removing it

ios/legacy/RNSScreen.mm
```diff
@@ -1751,9 +1753,9 @@ - (void)handleAnimation
 
 - (void)notifyTransitionProgress:(double)progress closing:(BOOL)closing goingForward:(BOOL)goingForward
 {
-  if ([self.view isKindOfClass:[RNSScreenView class]]) {
-    // if the view is already snapshot, there is not sense in sending progress since on JS side
-    // the component is already not present
+  // if the screen was already deleted by React, there is no sense in sending progress
+  // since on JS side the component is already not present
+  if ([self.view isKindOfClass:[RNSScreenView class]] && !((RNSScreenView *)self.view).isInvalidated) {
```
t0maboro: nit: I'd consider splitting the condition for a better readability, sth like that:
```
    if ([self.view isKindOfClass:[RNSScreenView class]]) {
        auto screenView = (RNSScreenView *)self.view;
        if (screenView.isInvalidated) {
            return;
        }
```

t0maboro (approved):
Left some minor comments; the runtime seems to be working ok.
I'll be testing more before the next release and will focus on that (if any doubts come, I'll gate that behind the FF before the release), so I think it's fine to merge in the current shape
Please remember to add backport-to-v4 label

ios/legacy/RNSScreen.mm
kmichalikk: Before the patch, controller.view could be the screen or the snapshot, and the _initialView held the actual screen. Now the _initialView is always the screen, so leaving the comment would be misleading. Technically, I could also remove _initialView and just return self.view - what do you think?

ios/legacy/RNSScreen.mm
kmichalikk: [54f9faa](https://github.com/software-mansion/react-native-screens/pull/4407/commits/54f9faada8701e0ab010a510e8e888ef64a6be87) I went with sth different

ios/legacy/RNSScreen.mm
kmichalikk: [54f9faa](https://github.com/software-mansion/react-native-screens/pull/4407/commits/54f9faada8701e0ab010a510e8e888ef64a6be87)

ios/legacy/RNSScreen.mm
t0maboro: +1 for removing the `initialView` if it duplicates the same value as `.view` now

t0maboro (approved):
looks good, 1 additional comment about `initialView` left above

ios/legacy/RNSScreen.mm
kmichalikk: [4b6bef0](https://github.com/software-mansion/react-native-screens/pull/4407/commits/4b6bef04d561d0a4cada9d8dd542534335440003)

PR #4408, opened 2026-07-27
fix(iOS, Stack v4): fall back to inheriting orientation if it's not defined on the screen

### LINKED ISSUE #4351: [iOS] Screen screenOrientation is ignored when it hosts a nested hasTwoStates ScreenContainer (e.g. react-navigation bottom-tabs / drawer)
### Description

On iOS, a native-stack screen's `orientation` prop is silently ignored whenever that screen hosts a nested `ScreenContainer` running in navigation mode (`hasTwoStates`). This is exactly what `@react-navigation/bottom-tabs` (and `@react-navigation/drawer`) render for their scenes when no scene animation is configured. As a result the device rotates freely even though the host screen requested e.g. `orientation: 'portrait'`.
 
**Expected:** a screen that sets `orientation: 'portrait'` stays locked to portrait, regardless of whether it happens to host a tab/drawer navigator.
 
**Actual:** the screen rotates freely (behaves as `all-but-upside-down`). The tab/drawer scene wrappers cannot carry `orientation` themselves — bottom-tabs passes only `enabled`/`activityState` to them — so there is no way to work around it from JS.
 
A sibling screen in the same stack that does *not* host a nested container (a plain screen) locks to portrait correctly, which makes the difference easy to see side by side.
 
### Likely cause
 
The regression is in the child-first orientation resolution introduced in [#3014](https://github.com/software-mansion/react-native-screens/pull/3014) (`RNSOrientationProviding` / `evaluateOrientation`). In `ios/RNSScreen.mm`, `-[RNSScreen evaluateOrientation]` falls back to `RNSOrientationFromUIInterfaceOrientationMask([self supportedInterfaceOrientations])`. When nothing in the subtree sets an orientation, that override returns `UIInterfaceOrientationMaskAllButUpsideDown`, which converts back to the concrete `RNSOrientationAllButUpsideDown` rather than `RNSOrientationInherit`. An unset tab scene therefore reports a concrete opinion that overrides the ancestor host screen's `portrait` prop as it bubbles up.

### Steps to reproduce

1. Allow landscape at the app level (`Info.plist`-`UISupportedInterfaceOrientations`)
2. Render a native-stack navigator with `screenOptions={{ orientation: 'portrait' }}`.
3. Create a screen hosting a bottom tab navigator by `createBottomTabNavigator()` (no scene animation).
4. Open the tabs screen and rotate the device to landscape → **it rotates (bug).**
 
Minimal, complete reproduction - also available from attached expo snack:
 
```tsx
import { View } from 'react-native';
import { createBottomTabNavigator } from '@react-navigation/bottom-tabs';
 
const Tab = createBottomTabNavigator();
 
const Tabs = () => (
  <Tab.Navigator>
    <Tab.Screen name="A" component={View} />
    <Tab.Screen name="B" component={View} />
  </Tab.Navigator>
);
 
  return (
    <NavigationContainer>
      <Stack.Navigator screenOptions={{ orientation: 'portrait' }}>
        <Stack.Screen name="Tabs" component={Tabs} />
      </Stack.Navigator>
    </NavigationContainer>
  );
}
```
 
A dependency-free variant reproduces the same override without react-navigation:
inside a `ScreenStack`, a `Screen` with `screenOrientation="portrait"` that
contains `<ScreenContainer hasTwoStates enabled>` with a child `Screen` (no
orientation).

https://snack.expo.dev/@snowsuno/react-native-screens-orientation-bug-repro

### Screens version

4.18.0 (reproducable from 4.15.0 ~)

### React Native version

0.85.3

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

Fabric (New Architecture)

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

kligarski:
## Description

Fixes orientation evaluation for Stack v4 when orientation prop is not defined. It now falls back to `RNSOrientationInherit` instead of `RNSOrientationAll`/`RNSOrientationAllButUpsideDown`.

Closes https://github.com/software-mansion/react-native-screens/issues/4351.

### Details

In https://github.com/software-mansion/react-native-screens/pull/3014 we introduced a mechanism for evaluating orientation across nested containers. For backward compatibility, we used `RNSScreen.supportedInterfaceOrientations` for evaluating orientation in Stack v4 which returns hard-coded `UIInterfaceOrientationMaskAll`/`UIInterfaceOrientationMaskAllButUpsideDown` if the screen and its children doesn't set any orientation. Due to new bottom-up evaluation behavior, the lowest `RNSScreen`'s `supportedInterfaceOrientations` would be queried and if it returned the default hard-coded value, it would override parent's defined orientation.

In this PR, we're still performing similar evaluation as `RNSScreen.supportedInterfaceOrientations` but we fall back to `RNSOrientationInherit` when no orientation is set so that we don't override any parent's defined orientation.

This brings one other change in behavior - if orientation is not defined at all, we fall back to supported orientations from `Info.plist` instead of using hard-coded `UIInterfaceOrientationMaskAll`/`UIInterfaceOrientationMaskAllButUpsideDown`. If the configuration in `Info.plist` is different that hard-coded defaults, the behavior will be different.

To mitigate this and any potential unforseen problems, we added a feature flag (enabled by default) that controls this behavior. Disabling it restores behavior prior to this PR.

## Changes

- return `RNSOrientationInherit` in `RNSScreen.evaluateOrientation` when no orientation is defined in children subtree instead of hard-coded defaults
- add `iosOrientationInheritanceFixEnabled` to control this behavior
- update in-code docs, `GUIDE_FOR_LIBRARY_AUTHORS.md`

| before           | after            |
| ---------------- | ---------------- |
| <video src="https://github.com/user-attachments/assets/cc6a8d6b-f4ee-4423-9cc8-2f790eb3b116" /> | <video src="https://github.com/user-attachments/assets/1b1d55dd-1c5f-4cbb-b58a-7bcdbb128729" /> |

## Test plan

Run `Test4351`.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/apps/src/tests/issue-tests/Test4351.tsx
+++ b/apps/src/tests/issue-tests/Test4351.tsx
@@ -1,0 +1,120 @@
+import * as React from 'react';
+import { Button, StyleSheet, Text, View } from 'react-native';
+import { NavigationContainer, ParamListBase } from '@react-navigation/native';
+import {
+  createNativeStackNavigator,
+  NativeStackNavigationProp,
+} from '@react-navigation/native-stack';
+import { createBottomTabNavigator } from '@react-navigation/bottom-tabs';
+
+type StackParamList = {
+  Home: undefined;
+  Plain: undefined;
+  Tabs: undefined;
+};
+
+type NavigationProp<ParamList extends ParamListBase> = {
+  navigation: NativeStackNavigationProp<ParamList>;
+};
+
+type StackNavigationProp = NavigationProp<StackParamList>;
+
+const Stack = createNativeStackNavigator<StackParamList>();
+const Tab = createBottomTabNavigator();
+
+function HomeScreen({ navigation }: StackNavigationProp) {
+  return (
+    <View style={styles.container}>
+      <Text style={styles.title}>Issue #4351 – orientation inheritance</Text>
+      <Text style={styles.paragraph}>
+        The stack is locked to portrait. Push each screen and rotate the device
+        to landscape to compare behavior.
+      </Text>
+      <View style={styles.buttons}>
+        <Button
+          title="Push plain screen (control)"
+          onPress={() => navigation.push('Plain')}
+        />
+        <Button
+          title="Push screen with tabs (repro)"
+          onPress={() => navigation.push('Tabs')}
+        />
+      </View>
+    </View>
+  );
+}
+
+function PlainScreen() {
+  return (
+    <View style={styles.container}>
+      <Text style={styles.title}>Plain screen</Text>
+      <Text style={styles.paragraph}>
+        This leaf stack screen inherits `portrait` from the stack. It should
+        stay portrait whether the fix is enabled or not.
+      </Text>
+    </View>
+  );
+}
+
+function TabContent({ name }: { name: string }) {
+  return (
+    <View style={styles.container}>
+      <Text style={styles.title}>Tab {name}</Text>
+      <Text style={styles.paragraph}>
+        These tab screens do not set `orientation`.{'\n\n'}
+        Fix ON (default): stays portrait (defers to the stack's `portrait`).
+        {'\n\n'}
+        Fix OFF: the tab screen forces `allButUpsideDown`, so the device can
+        rotate to landscape – reproducing the bug.
+      </Text>
+    </View>
+  );
+}
+
+function TabsScreen() {
+  return (
+    <Tab.Navigator>
+      <Tab.Screen name="A">{() => <TabContent name="A" />}</Tab.Screen>
+      <Tab.Screen name="B">{() => <TabContent name="B" />}</Tab.Screen>
+    </Tab.Navigator>
+  );
+}
+
+export default function App() {
+  return (
+    <NavigationContainer>
+      <Stack.Navigator screenOptions={{ orientation: 'portrait' }}>
+        <Stack.Screen name="Home" component={HomeScreen} />
+        <Stack.Screen name="Plain" component={PlainScreen} />
+        <Stack.Screen
+          name="Tabs"
+          component={TabsScreen}
+          options={{ headerShown: true }}
+        />
+      </Stack.Navigator>
+    </NavigationContainer>
+  );
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    alignItems: 'center',
+    justifyContent: 'center',
+    padding: 24,
+  },
+  title: {
+    fontSize: 20,
+    fontWeight: '600',
+    marginBottom: 12,
+    textAlign: 'center',
+  },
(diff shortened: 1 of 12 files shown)


src/components/tabs/screen/TabsScreen.types.ts
```diff
@@ -110,9 +110,13 @@ export interface TabsScreenPropsBase {
...
    * - some components (like `SplitHost`) may choose not to query
    *   its child components,
-   * - Stack v4 implementation **ALWAYS** returns some supported
-   *   orientations (`allButUpsideDown` by default), overriding
-   *   orientation from tab screen.
+   * - Stack v4 (legacy) implementation returns some supported orientations,
+   *   overriding orientation from the tab screen **unless** the Stack v4
+   *   screen has no explicit `orientation`
+   *   and `featureFlags.experiment.iosOrientationInheritanceFixEnabled` is
+   *   enabled (the default), in which case it returns `inherit` and no longer
```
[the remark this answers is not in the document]
kligarski: https://github.com/software-mansion/react-native-screens/pull/4408/commits/953e0b76b1c9ad391702b117cbf32efdf235a034

apps/src/tests/issue-tests/Test4351.tsx
```diff
@@ -0,0 +1,120 @@
+import * as React from 'react';
+import { Button, StyleSheet, Text, View } from 'react-native';
+import { NavigationContainer, ParamListBase } from '@react-navigation/native';
+import {
+  createNativeStackNavigator,
+  NativeStackNavigationProp,
```
t0maboro: can be `import type`

apps/src/tests/issue-tests/Test4351.tsx
```diff
@@ -0,0 +1,120 @@
+import * as React from 'react';
+import { Button, StyleSheet, Text, View } from 'react-native';
+import { NavigationContainer, ParamListBase } from '@react-navigation/native';
```
t0maboro: `ParamListBase` can be `import type`

t0maboro (approved):

apps/src/tests/issue-tests/Test4351.tsx
kligarski: https://github.com/software-mansion/react-native-screens/pull/4408/commits/db3b98833ca68df10acf38321f5734ac9c6e96e6

apps/src/tests/issue-tests/Test4351.tsx
kligarski: https://github.com/software-mansion/react-native-screens/pull/4408/commits/db3b98833ca68df10acf38321f5734ac9c6e96e6

PR #4411, opened 2026-07-27
fix(iOS, Tabs): re-layout tab bar after async icon load to prevent label truncation on iOS 26

davut:
## Description

On iOS 26, `RNSTabBarAppearanceCoordinator` assigns `tabBarItem.image` from an asynchronous image-loader completion block and calls `setNeedsLayout` without `layoutIfNeeded`. The deferred pass can run before the image lands, leaving the label slot measured for the wrong content — the label renders ellipsised (e.g. `Главная` → `Гла…`) until the user selects that tab, which forces the bar to lay out again.

It heals **per item**: only the tab you select re-lays out, so an unselected tab stays clipped. It isn't length-driven either — in my app `Профиль` (7 chars) renders fine while `Главная` (7 chars) clips.

Reported downstream as:
- react-navigation/react-navigation#12908
- expo/expo#42364

## Changes

One layout invalidation: after the async image is assigned, force the pass that a tab tap would have triggered, so the slot is re-measured with the image in place. Guarded to iOS 26.

## Test plan

Production Expo app using `expo-router/unstable-native-tabs` (react-native-screens 4.26.1 + this change via patch-package), iPhone 17 Pro, iOS 26.4. The app loads 10 raster tab icons at mount, so the async path fires reliably.

- **Unpatched:** two of four labels truncated on the very first entry into the tabs. Tapping tabs healed them one at a time; an unselected tab stayed clipped.
- **Patched:** all labels render fully on first mount, and after an in-session language switch (full navigator remount + restore push) with no tab tap in between.
- No regressions in cold launch, tab switching, badge updates, or a full native-host remount.

## Notes

An earlier revision of this PR also patched `RNSTabsScreenComponentView` (title writes) and `RNSTabsScreenViewController` (`viewDidAppear`). Both are removed. I could not reproduce a title-change truncation in a minimal app across seven configurations — raster icons, SF Symbols, no icons, shorter labels, navigator remount, an uncached icon swap in the same commit, and my app's full `NativeTabs` prop set — and neither hunk is needed to fix the app once the async-image path is handled.

I have only verified iOS 26.4. Whether 18.x is affected is unconfirmed; happy to test and lift the guard if it is.

### DIFF
--- a/ios/tabs/RNSTabBarAppearanceCoordinator.mm
+++ b/ios/tabs/RNSTabBarAppearanceCoordinator.mm
@@ -163,6 +163,13 @@
   if ([screenParentViewController isKindOfClass:[UITabBarController class]]) {
     UITabBarController *tabBarVC = (UITabBarController *)screenParentViewController;
     [tabBarVC.tabBar setNeedsLayout];
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
+    if (@available(iOS 26, *)) {
+      // Without this the deferred pass can measure the label slot before the
+      // image lands, truncating the label until a tab is tapped.
+      [tabBarVC.tabBar layoutIfNeeded];
+    }
+#endif // Check for iOS >= 26.0
   }
 }
 
--- a/ios/tabs/RNSTabBarAppearanceCoordinator.mm
+++ b/ios/tabs/RNSTabBarAppearanceCoordinator.mm
@@ -169,7 +169,7 @@
       // image lands, truncating the label until a tab is tapped.
       [tabBarVC.tabBar layoutIfNeeded];
     }
-#endif // Check for iOS >= 26.0
+#endif // RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
   }
 }
 


kmichalikk (changes requested):
Hi @davut 
Thanks for the PR.

I tested the patch and it does fix the issue. But I'm not sure we need all these layout calls, and I'm not sold on the comments in the code.

What I can tell is that the `layoutIfNeeded` when assigning the image fixed the described problem. I cannot however reproduce the problem with setting the title.

https://github.com/user-attachments/assets/ffdf9947-edda-4486-a4a0-c123e15de2c2

If I'm missing some paths please give me a way to test before / after.

Another question, is this an iOS 26+ problem, or should I observe it in older versions as well? The expo issue you linked specifies 17.x 18.x but shows recording of iOS 26. If this is actually 26+ issue, lets wrap the code with version guards (`#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)` + if available)

Before landing, I'd like to have only the minimal changes that solve the problem, and minimal short comments.

davut:
You were right, and my title hunk was wrong. Dropped it.

I couldn't reproduce the title change either. Tried a minimal SDK 57 app on unpatched 4.26.2 with PNG icons, SF Symbols, no icons, shorter labels, a full navigator remount, an uncached icon swap in the same commit, and my app's entire `NativeTabs` config. Labels rendered fine every time.

So I went and bisected the patch in the app where it does reproduce, and the title write was never the trigger. With the patch removed the labels are already truncated on the **first entry into the tabs**, before anything touches a title — and tapping heals them one at a time, only the tab you select. That's your async image path. My app loads 10 raster tab icons at mount, so those completion blocks land after the bar has laid out. The "language switch" I described just remounts the navigator and re-requests the icons, same path one step removed. In the minimal repro the images were already cached by the time I changed a title, which is why nothing ever showed up there.

Rebuilt with only the `RNSTabBarAppearanceCoordinator` hunk: clean on first mount, and clean after an in-session language switch with no tab tap in between.

So this is just that one hunk now, guarded, one-line comment. Title and `viewDidAppear` hunks are gone.

On versions — every report I've seen is iOS 26 and I've only verified on 26.4, so I guarded it as you asked. I haven't checked 18.x. I do have a deterministic repro now, so if you want that boundary pinned down before merging, say the word and I'll test it and drop the guard if 18.x needs it too.

ios/tabs/RNSTabBarAppearanceCoordinator.mm
```diff
@@ -163,6 +163,13 @@ - (void)updateTabBarItem:(nullable UITabBarItem *)tabBarItem
   if ([screenParentViewController isKindOfClass:[UITabBarController class]]) {
     UITabBarController *tabBarVC = (UITabBarController *)screenParentViewController;
     [tabBarVC.tabBar setNeedsLayout];
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
+    if (@available(iOS 26, *)) {
+      // Without this the deferred pass can measure the label slot before the
+      // image lands, truncating the label until a tab is tapped.
+      [tabBarVC.tabBar layoutIfNeeded];
+    }
+#endif // Check for iOS >= 26.0
```
kmichalikk:
```suggestion
#endif // RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
```

kmichalikk (approved):
Hi @davut 
Thanks, this looks good now, but please verify your case on 18.x before we merge.

davut:
Checked 18.x — not affected, so the guard looks right.

iPhone 16 Pro on iOS 18.5, same app and the same 10 raster tab icons that reproduce this every time on 26. The guard's `@available(iOS 26, *)` is false on 18.5, so that build behaves exactly like an unpatched one there.

Four cold starts straight into the tabs, labels render fully every time — including `Мои ответы`, which is *longer* than the `Продавцы` that clips on 26.4. Same app, same point in the flow, only the OS differs.

That lines up with the layout difference: on 26 the selected tab puts icon and label side by side inside the pill, on 18 the label sits under the icon with the whole tab width to itself. Only the 26 layout seems to care that the image landed late.

One caveat so I'm not overstating it: this is "not reproducible with this setup on 18.5", not proof that no pre-26 path exists anywhere. If you'd rather be safe I'm happy to drop the guard, but I've got no evidence 18.x needs it.

t0maboro:
Sorry, but it seems that this PR introduces a regression in our examples.

At the moment, not only are the labels no longer laid out correctly, but there's also a visible flash in the tab bar when navigating to a screen nested inside the stack when the tab bar has appearance defined. Additionally, the title is not always positioned correctly.

Since this is a blocker for the 4.27 release and introduces a new regression, I'm reverting this PR and creating a follow-up task to further investigate the misaligned title and icon issue.

ticket: https://github.com/software-mansion/react-native-screens-labs/issues/1127

https://github.com/user-attachments/assets/886a32ab-291c-43bc-98e1-d42820b512f4

PR #4412, opened 2026-07-28
chore(test, Android): add e2e test for Stack v5 toolbar menu commands

LKuchno:
## Description

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1694

Adds Detox e2e coverage for the existing `test-stack-toolbar-menu-commands-android` scenario (Android toolbar menu items on the v5 stack header — the `toolbarMenu` prop plus the imperative `updateToolbarMenuElements` command).

The test automates all 16 scenario steps, covering:

- initial menu render from props and `onPress` reporting the correct item id,
- imperative commands: absent fields preserve the current value, `undefined` (sent as `null`) resets to the prop's regular default,
- a props update on any single slot rebuilding the whole menu and dropping command-applied state on **all** items,
- safe targeting of an id that is not currently in the menu (excluded slot / unknown id) — no crash, no state leaking into the re-included slot.

The only thing left uncovered is the **order** of the overflow entries; presence and visibility are asserted instead. `e2eCoverage` in `scenario-description.ts` is therefore bumped from `tbd` to `incomplete`, and `scenario.md` documents the gap.

## Changes

- Added `FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-commands-android.e2e.ts`.
- Added `CLASS_NAME_ANDROID_MENU_DROP_DOWN_LIST_VIEW` to `FabricExample/e2e/native-class-names.ts` — every menu-entry matcher is scoped with `withAncestor` to this class so entries in the popup window are not confused with same-titled text elsewhere in the hierarchy.
- Test screen (`apps/src/tests/.../test-stack-toolbar-menu-commands-android/index.tsx`): added `testID`s to the scroll view, the command pickers, the *Send Command* button, the result text and the per-slot switches/pickers; made the picker/switch labels unique (`title` → `cmd title`, `title` → `slot N title`, …) so they no longer collide between the command section and the three slot sections.
- Updated `scenario.md` (E2E test section) and `scenario-description.ts` (`e2eCoverage: 'tbd'` → `'incomplete'`).

No library code is touched — this is test-only.

### DIFF
--- a/FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-commands-android.e2e.ts
+++ b/FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-commands-android.e2e.ts
@@ -1,0 +1,389 @@
+import { device, expect, element, by } from 'detox';
+import {
+  describeIfAndroid,
+  selectSingleFeatureTestsScreen,
+} from '../../e2e-utils';
+import { CLASS_NAME_ANDROID_MENU_DROP_DOWN_LIST_VIEW } from '../../native-class-names';
+
+const SCROLLVIEW_ID = 'toolbar-menu-commands-scrollview';
+const HEADER_TITLE = 'Toolbar Menu Commands Test';
+
+/**
+ * Budget for the overflow popup's open / dismiss animation. Detox's idle sync
+ * does not cover popup window animations, so every assertion that straddles one
+ * has to wait it out explicitly instead of asserting on the spot.
+ */
+const MENU_ANIMATION_TIMEOUT = 2000;
+
+/**
+ * Every title this scenario can ever put into the toolbar menu. Each assertion
+ * checks the full set — the expected titles must be visible and every other one
+ * must not exist — so a stale/leaked menu entry fails the test instead of
+ * slipping through an "only check what I expect" assertion.
+ */
+const ALL_TITLES = [
+  'Title A',
+  'Title B',
+  'Title C',
+  'Changed',
+  'Long Title',
+] as const;
+
+type MenuTitle = (typeof ALL_TITLES)[number];
+
+/**
+ * Mirrors the option `testID` that `SettingsPicker` derives from its `label`.
+ * @see apps/src/shared/SettingsPicker.tsx
+ */
+function optionId(pickerLabel: string, option: string): string {
+  return `${pickerLabel.split(' ').join('-')}-${option}`.toLowerCase();
+}
+
+/**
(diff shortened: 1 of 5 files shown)


FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-commands-android.e2e.ts
```diff
@@ -0,0 +1,389 @@
...
+        ).not.toExist();
+      }
+    }
+  } finally {
+    // Closed in `finally` so a failed assertion — including the gate above —
+    // does not leave the popup covering the screen. This suite is stateful, so
+    // every later step would otherwise fail on an unrelated error.
+    await device.pressBack();
+    await waitForScreen();
+  }
```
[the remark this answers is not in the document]
LKuchno: https://github.com/software-mansion/react-native-screens/pull/4412/changes/87a530c6926ae7e98fbed37813b1c897976ac5ff

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-commands-android.e2e.ts
```diff
@@ -0,0 +1,389 @@
...
+    await waitForMenuItem(expectedVisible[0]);
+
+    for (const title of expectedVisible) {
+      await expect(
+        element(
+          by
+            .text(title)
+            .withAncestor(by.type(CLASS_NAME_ANDROID_MENU_DROP_DOWN_LIST_VIEW)),
+        ),
+      ).toBeVisible();
```
kligarski, line 221: I guess that the order of options in the menu can be checked by reading frame's `y`. Not sure if that's necessary but we could then make it `e2eCoverage: 'full'`.

kligarski (approved):
We can try to check the order of items by reading frames but not sure if that's necessary - up to you.

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-commands-android.e2e.ts
LKuchno: https://github.com/software-mansion/react-native-screens/pull/4412/changes/f1fdcd22629532552293e9c473387f6ce045c98f

kligarski: Let's just verify that the values for `frame.y` seem correct and it's not just `0` or something like this. If that's all correct then we're good to proceed. Thanks.

PR #3363, opened 2025-10-31
fix(Android, Stack): Make MountingCoordinator vector access thread safe

t0maboro:
## Description

Adding the mutex for method access.

(Hopefully) Fixes https://github.com/software-mansion/react-native-screens/issues/3358

## Test code and steps to reproduce

Regression testing on app from: https://github.com/software-mansion/react-native-screens/issues/3234

## Checklist

- [x] Included code example that can be used to test this change
- [x] Ensured that CI passes

PR #4413, opened 2026-07-28
fix(Android): use-after-free of RNSScreenRemovalListener registered as mounting override delegate

### LINKED ISSUE #4151: [Android] SIGSEGV in librnscreens.so with Fabric / New Architecture
### Description

We are seeing a native Android crash reported by Crashlytics that appears to originate in `librnscreens.so`.

This occurs with the React Native New Architecture (Fabric) enabled. The crash appears to happen during navigation / screen transitions.

The crash has been observed on `react-native-screens@4.24.7`.

I searched existing issues and found related discussions, but could not confirm that they match this exact crash signature:

- https://github.com/software-mansion/react-native-screens/issues/3249
- https://github.com/software-mansion/react-native-screens/pull/3250

We are currently evaluating whether upgrading to `4.25.2` resolves the issue, but I could not find a confirmed fix for this specific SIGSEGV in the release notes.

### Crash stack

```text
Crashed: Thread: SIGSEGV  0x000000e500000009
#00 pc 0x14c94 librnscreens.so (BuildId: ace33169f7be2c2f4fe1d0c784b2794a302153c7)
#01 pc 0x14a58 librnscreens.so (BuildId: ace33169f7be2c2f4fe1d0c784b2794a302153c7)
#02 pc 0x14a58 librnscreens.so (BuildId: ace33169f7be2c2f4fe1d0c784b2794a302153c7)
#03 pc 0x15680 librnscreens.so (BuildId: ace33169f7be2c2f4fe1d0c784b2794a302153c7)
#04 pc 0x155c0 librnscreens.so (BuildId: ace33169f7be2c2f4fe1d0c784b2794a302153c7)
#22 pc 0x66e090 libart.so
#23 pc 0x65f4fc libart.so
#26 pc 0x669380 libart.so
```

### Package versions

```json
{
  "react": "19.1.1",
  "react-native": "0.82.1",
  "react-native-screens": "4.24.7",
  "@react-navigation/native": "7.1.25",
  "@react-navigation/native-stack": "7.8.6"
}
```

### Android configuration

```properties
newArchEnabled=true
hermesEnabled=true
minSdkVersion=34
compileSdkVersion=36
targetSdkVersion=36
Android Gradle Plugin=8.9.2
```

### Expected behavior

Navigation and screen transitions should not crash the native Android process.

### Actual behavior

Crashlytics reports a native SIGSEGV inside `librnscreens.so`.

### Additional context

- We have not yet identified a deterministic reproduction path.
- The crash is observed in production via Firebase Crashlytics.
- Fabric/New Architecture is enabled.
- We are planning to upgrade to `react-native-screens@4.25.2`, but have not yet confirmed whether the issue is resolved there.
- Please let us know if additional symbolication, tombstones, or native backtraces would help identify the root cause.

### Questions

1. Is this crash signature known and potentially related to the Android Fabric issues discussed in #3249 / #3250?
2. Has a fix already landed in a version newer than `4.24.7`?
3. What additional diagnostics would help narrow down the root cause?

### Steps to reproduce

1. Build and run the Android app with the New Architecture enabled (`newArchEnabled=true`) and Hermes enabled.
2. Use `react-native-screens@4.24.7`.
3. Navigate through screens using `@react-navigation/native-stack` and trigger screen transitions repeatedly.
4. Observe that the app occasionally crashes in production with a native SIGSEGV in `librnscreens.so`.

**Notes:** This crash is intermittent and we have not yet found a deterministic local reproduction.

No Snack or public repository is available at the moment.

### Screens version

 4.24.7

### React Native version

0.82.1

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Release mode

### Device

Real device

### Device model

Samsung Galaxy A35 5G (`SM-A356U`)

### Acknowledgements

Yes

mbSmaga:
## Description

Fixes an Android SIGSEGV in `facebook::react::MountingCoordinator::pullTransaction`
(thread `mqt_v_js`). It surfaces in two shapes at the same
`shouldOverridePullTransaction()` dispatch, depending on what recycles the freed
block: `SEGV_ACCERR` with the fault address inside `[anon:scudo:primary]`, or
`SEGV_MAPERR` with fault address `0x0` when the vtable slot reads back as zero.
The second looks like a core-renderer null deref; see the analysis in the thread.

Mechanism — an unsynchronized lazy init in `NativeProxy::nativeAddMutationsListener`:

1. Two threads reach `nativeAddMutationsListener` concurrently on a cold launch.
   `ScreensModule.initialize()` registers its lifecycle listener *before* calling
   `setupFabric()`; when the host is already resumed, `onHostResume()` is dispatched
   to the UI thread at registration time — so `setupFabric()` runs on the module-init
   thread while the same call is already queued on the UI thread.
2. Both threads can pass the `if (!screenRemovalListener_)` null check. The racing
   `shared_ptr` assignments are UB: libc++'s move-assign swaps pointer and control
   block as two independent words, so the member can be left holding one thread's
   object pointer with the other thread's control block — and the other thread's
   temporary then destroys and frees the pointed-to listener on release.
3. The registered delegate now reports `use_count=1, expired=0` forever while pointing
   at freed memory. The next `pullTransaction` locks the `weak_ptr` successfully and
   virtual-calls through a recycled vtable slot — SIGSEGV. (A listener destroyed
   *normally* through its own control block is harmless: `expired=1`, `lock()` returns
   null, and core already null-checks that path.)

History: #3241 added the `onHostResume` entry path (its description notes "the order
of lifecycle events is not guaranteed"); #3363 then added `coordinatorsMutex_` because
this function is entered concurrently — but guarded only the vector operations below,
leaving the listener init unsynchronized. A second tear site: `setMountingOverrideDelegate`
takes its `weak_ptr` by value, so registration constructs it from the racing member
with two separate loads.

A second, related bug fixed here: the old callback captured raw `this` and read
`javaPart_` live at call time, while `invalidateNative()` nulls `javaPart_` with the
listener still registered (core's `mountingOverrideDelegates_` is append-only — no
`removeMountingOverrideDelegate` as of 0.87). A notification delivered between
invalidation and the hybrid's destruction called `GetObjectClass` on a null jobject.

Reports of this signature:

- Closes #4151
- Expensify/App#93842 (active, bountied)
- leon-zym/plogkit#54 (react-native 0.86 / Expo SDK 57, e2e stop/launch cycle)

## Changes

- `RNSScreenRemovalListener` is now a process-lifetime singleton (function-local
  static `shared_ptr` in `NativeProxy.cpp`): thread-safe initialization removes the
  racing lazy init entirely, and the entry in core's non-removable delegate list can
  never dangle.
- The listener holds a mutex-guarded swappable callback instead of being constructed
  around one. The callback captures the JNI global reference by value — never `this` —
  so it cannot dereference a finalized `NativeProxy`.
- `setListener` returns a monotonic ownership token; `invalidateNative()` disarms via
  `clearListener(token)`, which is a no-op when a newer install owns the callback
  (protects apps running a second `ReactHost`, e.g. brownfield/hybrid setups, from a
  stale proxy's late teardown).
- Install and invalidate serialize on a per-proxy mutex, so the stored token always
  matches the last install and the install-time copy of the global reference cannot
  race the invalidation write.
- A disarmed listener passes the transaction through untouched (same
  `MountingTransaction` the function already returned) and releases the Java reference
  so the proxy stays collectable.

The durable fix belongs in react-native core (a `removeMountingOverrideDelegate` API
called from delegate owners' teardown); this patch makes the listener safe without it
and stays correct once such an API exists.

## Test plan

The race branch is directly measurable — a far more powerful signal than waiting for
the ~1-in-1000 crash. Minimal repro against **unpatched** code: add one log line
inside the lazy-init branch of `NativeProxy::nativeAddMutationsListener`:

```cpp
// unpatched NativeProxy.cpp, inside `if (!screenRemovalListener_)`:
auto tmp = std::make_shared<RNSScreenRemovalListener>([this](int tag) { /* unchanged */ });
__android_log_print(ANDROID_LOG_ERROR, "RNSInitRace", "alloc listener=%p tid=%d",
    (void *)tmp.get(), gettid());
screenRemovalListener_ = std::move(tmp);
```

then cold-relaunch in a loop:

```bash
while true; do
  adb shell am force-stop <pkg>
  adb shell am start -W -n <pkg>/<main-activity>
  sleep 15
done
```

Two `RNSInitRace` lines with different `tid`s in one launch prove both threads
entered the branch; two distinct `listener=` addresses prove the double allocation.

Results:

- In one app reproducing the crash, both threads executed the init branch on 7/9
  cold launches and it crashed on 3/9; with the init serialized: 0/20 and 0/20. In a
  dev build of the app from Expensify/App#93842: double-init on 1/150 launches
  unguarded vs 0/1000 serialized, with the racy entry window still opening at an
  unchanged rate (30.0% vs 30.4%) — the fix removes the race, not the code path.
- This exact patch ships today via patch-package in two production codebases
  (react-native-screens 4.25.0 and 4.26.2 — files identical to `main`'s
  `cpp/legacy/` copies). Independent verification in the thread below: physical
  arm64 device, 6/128 launches crashing unpatched → 0/88 patched.
- No new automated test file: the race is not deterministically triggerable without
  test-only timing code (a `usleep` widener in the pull path). Behavioral regression
  of the removal path is covered by the existing FabricExample transition e2e suites,
  and the new code compiles in CI for all ABIs.

## Checklist

- [ ] For visual changes, included screenshots / GIFs / recordings documenting the change. — N/A, no visual changes.
- [ ] For API changes, updated relevant public types. — N/A, no public JS/TS API change (internal C++ member removed).
- [ ] Ensured that CI passes — no workflow run has been triggered on this PR yet (fork).

### DIFF
--- a/android/src/main/cpp/NativeProxy.cpp
+++ b/android/src/main/cpp/NativeProxy.cpp
@@ -2,6 +2,7 @@
 #include <react/fabric/Binding.h>
 #include <react/renderer/scheduler/Scheduler.h>
 
+#include <memory>
 #include <string>
 
 #include "NativeProxy.h"
@@ -10,6 +11,21 @@ using namespace facebook;
 using namespace react;
 
 namespace rnscreens {
+
+namespace {
+// Process-lifetime owner. MountingCoordinator::mountingOverrideDelegates_ is
+// append-only with no removal API in react-native core (as of 0.87), so a
+// listener that is destroyed (here: by Java GC finalizing the NativeProxy
+// hybrid) leaves a stale weak_ptr that can still be locked and
+// virtual-dispatched from MountingCoordinator::pullTransaction on the JS
+// thread -> SIGSEGV.
+// Function-local static: thread-safe init, no static-init-order dependency.
+const std::shared_ptr<RNSScreenRemovalListener> &removalListener() {
+  static const std::shared_ptr<RNSScreenRemovalListener> instance =
+      std::make_shared<RNSScreenRemovalListener>();
+  return instance;
+}
+} // namespace
 
 NativeProxy::NativeProxy(jni::alias_ref<NativeProxy::javaobject> jThis)
     : javaPart_(jni::make_global(jThis)) {}
@@ -32,15 +48,14 @@ void NativeProxy::nativeAddMutationsListener(
   auto uiManager =
       fabricUIManager->getBinding()->getScheduler()->getUIManager();
 
-  if (!screenRemovalListener_) {
-    screenRemovalListener_ =
-        std::make_shared<RNSScreenRemovalListener>([this](int tag) {
-          static const auto method =
-              javaPart_->getClass()->getMethod<void(jint)>(
-                  "notifyScreenRemoved");
-          method(javaPart_, tag);
-        });
-  }
+  // Capture a copy of the global ref, never `this`: the listener outlives this
+  // NativeProxy, which Java GC destroys on the finalizer thread.
+  removalListenerToken_ =
+      removalListener()->setListener([javaPart = javaPart_](int tag) {
+        static const auto method =
+            javaPart->getClass()->getMethod<void(jint)>("notifyScreenRemoved");
+        method(javaPart, tag);
+      });
 
   cleanupExpiredMountingCoordinators();
 
@@ -79,7 +94,7 @@ void NativeProxy::addMountingCoordinatorIfNeeded(
       });
 
   if (!wasRegistered) {
-    coordinator->setMountingOverrideDelegate(screenRemovalListener_);
+    coordinator->setMountingOverrideDelegate(removalListener());
     coordinatorsWithMountingOverrides_.push_back(coordinator);
   }
 }
@@ -90,6 +105,9 @@ jni::local_ref<NativeProxy::jhybriddata> NativeProxy::initHybrid(
 }
 
 void NativeProxy::invalidateNative() {
+  // Disarm before the hybrid is collected; also releases the captured global
+  // ref so the Java NativeProxy stays collectable.
+  removalListener()->clearListener(removalListenerToken_);
   javaPart_ = nullptr;
 }
 
--- a/android/src/main/cpp/NativeProxy.cpp
+++ b/android/src/main/cpp/NativeProxy.cpp
@@ -13,17 +13,11 @@ using namespace react;
 namespace rnscreens {
 
 namespace {
-// Process-lifetime owner. Two threads can reach nativeAddMutationsListener
-// concurrently on a cold launch: ScreensModule.initialize() on the module-init
-// thread, and onHostResume() on the UI thread (queued before setupFabric()
-// returns). The previous lazy `if (!screenRemovalListener_)` init raced there:
-// concurrent shared_ptr assignments are UB and can leave the registered
-// delegate pointing at freed memory while its control block still reports it
-// alive, so the next MountingCoordinator::pullTransaction locks the weak_ptr
-// successfully and virtual-calls into recycled heap -> SIGSEGV.
-// A function-local static leaves nothing to race (thread-safe init, no
-// static-init-order dependency), and a process-immortal instance also suits
-// core's delegate list, which is append-only with no removal API (as of 0.87).
+// Keep this listener process-lifetime to avoid races during initialization,
+// when nativeAddMutationsListener() can be called concurrently from different
+// threads. A function-local static initializes thread-safely, and an immortal
+// instance suits core's delegate list, which is append-only. For more details:
+// https://github.com/software-mansion/react-native-screens/pull/4413
 const std::shared_ptr<RNSScreenRemovalListener> &removalListener() {
   static const std::shared_ptr<RNSScreenRemovalListener> instance =
       std::make_shared<RNSScreenRemovalListener>();
--- a/android/src/main/cpp/NativeProxy.cpp
+++ b/android/src/main/cpp/NativeProxy.cpp
@@ -13,13 +13,17 @@ using namespace react;
 namespace rnscreens {
 
 namespace {
-// Process-lifetime owner. MountingCoordinator::mountingOverrideDelegates_ is
-// append-only with no removal API in react-native core (as of 0.87), so a
-// listener that is destroyed (here: by Java GC finalizing the NativeProxy
-// hybrid) leaves a stale weak_ptr that can still be locked and
-// virtual-dispatched from MountingCoordinator::pullTransaction on the JS
-// thread -> SIGSEGV.
-// Function-local static: thread-safe init, no static-init-order dependency.
+// Process-lifetime owner. Two threads can reach nativeAddMutationsListener
+// concurrently on a cold launch: ScreensModule.initialize() on the module-init
+// thread, and onHostResume() on the UI thread (queued before setupFabric()
+// returns). The previous lazy `if (!screenRemovalListener_)` init raced there:
+// concurrent shared_ptr assignments are UB and can leave the registered
+// delegate pointing at freed memory while its control block still reports it
+// alive, so the next MountingCoordinator::pullTransaction locks the weak_ptr
+// successfully and virtual-calls into recycled heap -> SIGSEGV.
+// A function-local static leaves nothing to race (thread-safe init, no
+// static-init-order dependency), and a process-immortal instance also suits
+// core's delegate list, which is append-only with no removal API (as of 0.87).
 const std::shared_ptr<RNSScreenRemovalListener> &removalListener() {
   static const std::shared_ptr<RNSScreenRemovalListener> instance =
       std::make_shared<RNSScreenRemovalListener>();
@@ -48,14 +52,20 @@ void NativeProxy::nativeAddMutationsListener(
   auto uiManager =
       fabricUIManager->getBinding()->getScheduler()->getUIManager();
 
-  // Capture a copy of the global ref, never `this`: the listener outlives this
-  // NativeProxy, which Java GC destroys on the finalizer thread.
-  removalListenerToken_ =
-      removalListener()->setListener([javaPart = javaPart_](int tag) {
-        static const auto method =
-            javaPart->getClass()->getMethod<void(jint)>("notifyScreenRemoved");
-        method(javaPart, tag);
-      });
+  {
+    // Capture a copy of the global ref, never `this`: the listener outlives
+    // this NativeProxy, and the old live read of javaPart_ at call time raced
+    // invalidateNative(). installMutex_ keeps the copy from racing the
+    // invalidation write and keeps the stored token matching the last install.
+    std::lock_guard<std::mutex> lock(installMutex_);
+    removalListenerToken_ =
+        removalListener()->setListener([javaPart = javaPart_](int tag) {
+          static const auto method =
+              javaPart->getClass()->getMethod<void(jint)>(
+                  "notifyScreenRemoved");
+          method(javaPart, tag);
+        });
+  }
 
   cleanupExpiredMountingCoordinators();
 
@@ -106,7 +116,9 @@ jni::local_ref<NativeProxy::jhybriddata> NativeProxy::initHybrid(
 
 void NativeProxy::invalidateNative() {
   // Disarm before the hybrid is collected; also releases the captured global
-  // ref so the Java NativeProxy stays collectable.
+  // ref so the Java NativeProxy stays collectable. Serialized with the install
+  // path so the token is never stale and javaPart_ is not nulled mid-capture.
+  std::lock_guard<std::mutex> lock(installMutex_);
   removalListener()->clearListener(removalListenerToken_);
   javaPart_ = nullptr;
 }
--- a/android/src/main/cpp/NativeProxy.h
+++ b/android/src/main/cpp/NativeProxy.h
@@ -16,8 +16,6 @@ class NativeProxy : public jni::HybridClass<NativeProxy> {
  public:
   std::vector<std::weak_ptr<const facebook::react::MountingCoordinator>>
       coordinatorsWithMountingOverrides_;
-  // Guarded by installMutex_ (serializes listener install vs invalidateNative).
-  uint64_t removalListenerToken_{0};
   static auto constexpr kJavaDescriptor =
       "Lcom/swmansion/rnscreens/NativeProxy;";
   static jni::local_ref<jhybriddata> initHybrid(
@@ -34,6 +32,7 @@ class NativeProxy : public jni::HybridClass<NativeProxy> {
   // invalidateNative: the install-time copy of javaPart_ must not race the
   // invalidation write, and the stored token must match the last install.
   std::mutex installMutex_;
+  uint64_t removalListenerToken_{0};
 
   explicit NativeProxy(jni::alias_ref<NativeProxy::javaobject> jThis);
 
--- a/android/src/main/cpp/NativeProxy.h
+++ b/android/src/main/cpp/NativeProxy.h
@@ -4,6 +4,7 @@
 #include <react/fabric/JFabricUIManager.h>
 #include "RNSScreenRemovalListener.h"
 
+#include <cstdint>
 #include <mutex>
 #include <string>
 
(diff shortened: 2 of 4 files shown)


imdrak:
Independent verification — this fixes a crash we had been chasing in a production RN 0.85.3 app.

Verified on `react-native-screens` 4.25.2, RN 0.85.3 (New Architecture + Hermes), `@react-navigation/native-stack` 7.13.x, Galaxy S25 Ultra (Android 16 / API 36), debug build. Repro is a looped cold start into a deep link that replaces the entire native stack during startup — crash at ~3–4 s uptime on `mqt_v_js`, no UI interaction.

| Build | Launches | Crashes |
| -- | --: | --: |
| 4.25.2 unpatched, no instrumentation | 40 | 2 |
| 4.25.2 unpatched, with delegate instrumentation | 88 | 4 |
| **4.25.2 unpatched, total** | **128** | **6 (4.7%)** |
| 4.25.2 + this PR | 88 | **0** |

Applied on 4.25.2 by path-mapping `cpp/legacy/RNSScreenRemovalListener.*` → `cpp/…` (contents are byte-identical); applied without fuzz. All 88 post-patch launches were confirmed to have actually loaded JS, so the zero isn't a broken-run artifact. Deep-link entry, tab switching and stack push/pop all behave normally.

*Edited to correct the unpatched tally: an intermediate instrumented run (40 launches, 2 crashes) was left out of the first count. The rate is unchanged (~4.7%) and the patched result stands at 0/88.*

We independently landed on the same delegate before finding this PR: instrumenting the registration sites and comparing register `x0` at the fault against the registered pointer **within the same process run** matched `RNSScreenRemovalListener` exactly in both crashes, with the listener destroyed and reallocated at the same address milliseconds earlier.

## Triage note: this bug has two faces

Resolving the offsets in `libreactnative.so` (RN 0.85.3 prebuilt), `pullTransaction+520/+524` are:

```
1130a04: ldr x8, [x8]   ; x8 = vtable[0]      <-- +520
1130a08: blr x8         ; virtual dispatch    <-- +524
```

— the `shouldOverridePullTransaction()` dispatch. Depending on what the recycled block holds, the *same* dispatch faults as either:

- `+524`, `SEGV_ACCERR`, fault addr in `[anon:scudo:primary]` — "trying to execute non-executable memory"
- `+520`, `SEGV_MAPERR`, fault addr **`0x0`** — "null pointer dereference"

Both occurred on the same build. The second one reads like a core-renderer null deref (it cost us a detour through `MountingCoordinator::revoke()` before the disassembly ruled it out), so reports carrying that signature are probably this same issue. Incidentally the `tbz` immediately before those instructions was taken as *non-null*, which confirms at instruction level the `weak_ptr::lock()`-returns-non-null behaviour you describe.

Happy to share the per-launch logs or symbolized tombstones if that would help review.

mbSmaga:
Thanks @imdrak - this is an excellent confirmation, and the disassembly detail is the
strongest part of it: the `tbz` taken as non-null right before the
`shouldOverridePullTransaction()` dispatch is instruction-level evidence that
`weak_ptr::lock()` returned non-null for a dead object. That's the signature that
separates this bug from any ordinary teardown: a listener destroyed normally through
its own control block reports `expired=1` and `lock()` returns null (core null-checks
that path), so only the racing lazy init leaving a mismatched {pointer, control block}
pair produces a lockable-but-freed delegate. I've just updated the PR description to
state that mechanism explicitly.

Your ACCERR/MAPERR split is useful too — it fits the reading that the fault class
follows whatever content recycles the freed chunk (pointer-dense heap data lands in
mapped non-executable memory → `SEGV_ACCERR`; string data sends the jump to an
unmapped ASCII-looking address → `SEGV_MAPERR`). I haven't captured a MAPERR case
myself, so a symbolized tombstone of one — plus the logs you offered — would be very
welcome.

And thank you for running it on arm64 hardware: my own soak data is all x86_64
emulator, so 6/128 → 0/88 on a physical device is exactly the coverage that was
missing.

imdrak:
Here's the MAPERR case, plus the per-launch data.

**`crash-3-MAPERR-null.txt`** is attached directly — the symbolized tombstone, all 91 frames, with only the app id, install path and deep link anonymized.

The rest is in **`rm663-evidence-bundle.zip`**:

- `crash-{1,2,4,5,6}-*.txt` — the five ACCERR tombstones, for contrast.
- `fault-mode-analysis.txt` — disassembly and registers for all six, in one place.
- `per-launch-index.md` — all 216 launches with signal code, fault address, faulting frame, and a "did JS actually run" column.
- `clean-run-for-contrast.txt` — an unpatched launch that didn't crash, where the destroyed and newly-registered addresses differ; `clean-run-patched.txt` — a patched launch.
- `repro-harness.sh` — the loop, parameterized (`SERIAL` / `PKG` / `LINK`).

One detail came out differently from the reading in your comment, and it changes what a duplicate search matches.

## In the MAPERR case the `blr` never runs

```
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0000000000000000
Cause: null pointer dereference
    x0  b40000796f9c1da0   <-- the dead delegate, taken as live by the tbz
    x8  0000000000000000   <-- the vtable pointer read back as zero
    lr  000000767eb57a00  pc  000000767eb57a04
backtrace:
  #00 pc 0000000001130a04  libreactnative.so (MountingCoordinator::pullTransaction(bool) const+520)
  #01 pc 0000000000c633a4  libreactnative.so (FabricUIManagerBinding::schedulerDidFinishTransaction(...)+72)
```

Against the three instructions from before — `ldr x8,[x0]` at +516, `ldr x8,[x8]` at +520, `blr x8` at +524 — the first load returned zero, so the fault is the second *load*, not the jump. `lr` settles it, since `blr` would have overwritten it:

| | fault addr | `x8` | `lr` resolves to | |
| -- | -- | -- | -- | -- |
| MAPERR ×1 | `0x0` | `0` | **+516** | still the return address of the `bl` at +512 — the `blr` never ran |
| ACCERR ×5 | heap address | == fault addr | **+528** | exactly what `blr x8` pushes — it did run |

(Resolved against each run's own load base; they differ per process.)

Your pointer-dense reading holds for the ACCERR side — in all five, `vtable[0]` came back as an address in the same scudo primary region, `0x78eff16ba8` … `0x78eff33b28`. The MAPERR one just doesn't get that far: the vtable slot itself was zero, so there's no jump and no ASCII-looking target. A jump to an unmapped ASCII-ish address would be a third shape — MAPERR at +524 with a non-zero fault address — and it didn't appear in our 128 unpatched launches.

Worth knowing for dedup: this face has no `[anon:scudo:*]` frame and its fault address is `0x0`, so it won't match a search for a garbage address or for "trying to execute non-executable memory". What it does match is `SEGV_MAPERR` + `fault addr 0x0` + `pullTransaction+520` at `#00`, on `mqt_v_js`, seconds into process life, with no `stopSurface` / `reload` / activity-destroy in the log. It read to us like a core-renderer null deref at first and cost a detour through `revoke()` / `baseRevision_`, which is the main reason I'm spelling it out. The same is true of "fault inside recycled scudo heap" in the description — it fits the ACCERR half; entirely your call whether widening it is worth it.

## What this particular tombstone does and doesn't show

The MAPERR launch is from the run where only the reanimated registration site was instrumented, so it doesn't carry the `x0` == screens-listener match. It does show `x0` (`b40000796f9c1da0`) is *not* the reanimated proxy, which logged as alive at `0xb4000077ef97e5b0` in the same run. The two positively attributed crashes are `crash-5` and `crash-6`, both ACCERR.

## Per-launch data

`per-launch-index.md` is generated from the log files rather than transcribed, so it re-derives 6/128 → 0/88 independently. The last column asserts each launch actually ran JS: crashed launches show 10–11 `ReactNativeJS` lines (died mid-startup), clean ones 15–19, none zero.

  
  
[crash-3-MAPERR-null.txt](https://github.com/user-attachments/files/30543956/crash-3-MAPERR-null.txt)
[rm663-evidence-bundle.zip](https://github.com/user-attachments/files/30543957/rm663-evidence-bundle.zip)

t0maboro:
Hi @mbSmaga , I'm adding this to my queue. Since this change is being introduced into the core setup, it requires more thorough testing on our side before proceeding. Ideally, I'd like to work on an example that would allow me to verify the behavior and add it to our example app for regression testing. I'll try to post some more detailed feedback in the upcoming days.

android/src/main/cpp/NativeProxy.cpp
```diff
@@ -11,6 +12,25 @@ using namespace react;
...
+// concurrently on a cold launch: ScreensModule.initialize() on the module-init
+// thread, and onHostResume() on the UI thread (queued before setupFabric()
+// returns). The previous lazy `if (!screenRemovalListener_)` init raced there:
+// concurrent shared_ptr assignments are UB and can leave the registered
+// delegate pointing at freed memory while its control block still reports it
+// alive, so the next MountingCoordinator::pullTransaction locks the weak_ptr
+// successfully and virtual-calls into recycled heap -> SIGSEGV.
+// A function-local static leaves nothing to race (thread-safe init, no
+// static-init-order dependency), and a process-immortal instance also suits
+// core's delegate list, which is append-only with no removal API (as of 0.87).
```
t0maboro: Generally, the inline comment in the code should provide enough context without diving into too much details. I'd move those details to the PR description. I'd suggest shortening this comment to sth like this:
```
// Keep this listener process-lifetime to avoid races during initialization,
// when nativeAddMutationsListener() can be called concurrently from different threads.
// A function-local static is safe here because core's delegate list is append-only.
// For more details, see: https://github.com/software-mansion/react-native-screens/pull/4413
```

android/src/main/cpp/NativeProxy.h
```diff
@@ -13,9 +14,10 @@ using namespace facebook::jni;
 
 class NativeProxy : public jni::HybridClass<NativeProxy> {
  public:
-  std::shared_ptr<RNSScreenRemovalListener> screenRemovalListener_;
   std::vector<std::weak_ptr<const facebook::react::MountingCoordinator>>
       coordinatorsWithMountingOverrides_;
+  // Guarded by installMutex_ (serializes listener install vs invalidateNative).
+  uint64_t removalListenerToken_{0};
```
t0maboro: Can this be private?

cpp/legacy/RNSScreenRemovalListener.h
```diff
@@ -4,17 +4,36 @@
...
+
+  // This instance is process-immortal (see NativeProxy.cpp): every
+  // MountingCoordinator keeps a weak_ptr to it for the life of its surface and
+  // react-native core has no removeMountingOverrideDelegate (as of 0.87).
+  // Swapping the callback instead of replacing the object means registration
+  // never re-runs, so there is no lazy init left to race.
+  // setListener returns an ownership token; clearListener is a no-op unless the
+  // token matches the latest install, so a stale proxy's late teardown (second
+  // ReactHost, late finalization) cannot disarm the callback a newer proxy
+  // installed.
```
t0maboro: same as above, I'd propose sth like that:

/// RN core lacks some `removeMountingOverrideDelegate` API
// We swap callbacks using an ownership token to prevent initialization races, 
// and to ensure that the teardown from the stale proxy doesn't clear the active 
// listener. See https://github.com/software-mansion/react-native-screens/pull/4413 for details.

t0maboro (commented):
@mbSmaga thanks for this PR! I verified the thread race scenario might occur and changes you propose make sense. Added a few suggestions to the code.

android/src/main/cpp/NativeProxy.cpp
mbSmaga: Trimmed to your version. Kept the shape but split the two reasons apart: the thread-safe init is what removes the race, while the append-only delegate list is what makes an immortal instance acceptable.
204d36b4

android/src/main/cpp/NativeProxy.h
mbSmaga: Moved to private:, directly under the installMutex_ that guards it. coordinatorsWithMountingOverrides_ above it is also only used internally; I left it alone to keep the diff scoped.
204d36b4

cpp/legacy/RNSScreenRemovalListener.h
mbSmaga: Applied, close to your wording. Normalized the leading /// to //, and moved "prevents initialization races" off the token, that's the singleton's job. The token only stops a stale proxy's teardown from clearing a newer listener. 204d36b4

mbSmaga:
Thanks for the review @t0maboro - all three applied in 204d36b4, plus a rebase onto current main. Notes in the threads.

@imdrak - you were right, my earlier reading was wrong: lr at +516 shows the blr never ran, the vtable slot itself read back as zero. The description now names both crash shapes so the fault addr 0x0 face stops reading as a core-renderer null deref. Thanks for the tombstones and the per-launch index.

t0maboro (approved):
thank you!

PR #4414, opened 2026-07-28
chore(test): add e2e test for stack toolbar menu title props (Android)

LKuchno:
## Description

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1695

Adds the Detox e2e test for the `test-stack-toolbar-menu-title-android` scenario,
which covers the title-related Android toolbar menu item props on the v5 stack
header: `title`, `titleCondensed`.

The scenario previously carried `e2eCoverage: 'tbd'`; it is now `'incomplete'`,
with the automated and manual-only steps spelled out in `scenario.md`.

## Changes

- **New e2e suite** — `FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-title-android.e2e.ts`,
  12 tests covering scenario steps 1–8 and 12–16:
  - `titleCondensed` labels the Toolbar action button and falls back to `title`
    when unset,
  - the overflow menu always shows the full `title`, never the condensed one,
  - the text-next-to-icon label appears only when there is room (hidden in
    portrait, shown in landscape),
  - `tooltipText` has no effect on overflow rows,
  - `updateToolbarMenuElements` updates and resets `title` / `titleCondensed`
    live.
- **Test screen** — added testIDs for the scroll view, every picker and the
  send-command button. Picker labels are now prefixed (`Slot <n> …`, `cmd …`)
  so that the option-row testIDs `SettingsPicker` derives from them stay unique
  across the three slots. The picker option types are exported so the e2e test
  derives its values from them instead of restating the lists; those exports are
  type-only and the emitted JS is unchanged.
- **`native-class-names.ts`** — registered the Android classes the suite matches
  on: `ActionMenuItemView` (Toolbar action button), `MenuPopupWindow$MenuDropDownListView`
  (overflow popup list) and `LinearLayout`.
- **`scenario.md` / `scenario-description.ts`** — filled in the E2E section and
  bumped coverage to `incomplete`.

Assertions are scoped by native class rather than by bare text, so
"shown on the Toolbar" and "shown in the overflow menu" are distinguished
structurally instead of relying on a string being absent from the whole screen.


FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-title-android.e2e.ts
```diff
@@ -0,0 +1,320 @@
...
+/**
+ * `SettingsPicker` derives the testID of its label and of every option row from
+ * the `label` it is given, replacing spaces with dashes and lowercasing the
+ * result. The screen labels its per-item pickers `Slot <n> <prop>` and its
+ * command pickers `cmd <prop>`, so the two builders below are that same
+ * derivation — renaming a label on the screen breaks the ids here.
+ */
+const pickerId = (label: string) => `${label}-picker`.toLowerCase();
+const optionId = (label: string, value: string) =>
+  `${label}-${value}`.toLowerCase();
```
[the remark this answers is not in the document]
LKuchno: I'd prefer to keep this as-is. The input labels are already dash-separated, making the transform a no-op that doesn't actually reduce call-site duplication.

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-title-android.e2e.ts
```diff
@@ -0,0 +1,369 @@
...
+/**
+ * The app draws edge to edge, so the bottom of the settings list is covered by
+ * the system navigation bar. Detox stops scrolling the moment a view is 75%
+ * visible, which parks it in exactly that strip — and a tap there lands on the
+ * Home button, sends the app to the launcher and fails every later step with
+ * "No activities in stage RESUMED". Keep this fraction of the list's height
+ * free below a control before tapping it; it is comfortably more than a
+ * navigation bar is tall.
+ */
+const NAV_BAR_CLEARANCE_RATIO = 0.1;
```
kligarski: Maybe instead we can modify `scrollUntilVisible` to make sure view is 100% visible (or use this only for this test)

```tsx
export async function scrollUntilVisible(id: string, scrollViewId: string) {
  await waitFor(element(by.id(id)))
    .toBeVisible(100)           // you can define how many % visible the view should be
    .whileElement(by.id(scrollViewId))
    .scroll(600, 'down', Number.NaN, 0.85);
}
```

If this won't work, we can also consider putting the scroll view inside `SafeAreaView` with bottom enabled.

apps/src/tests/single-feature-tests/stack-v5/test-stack-toolbar-menu-title-android/index.tsx
```diff
@@ -237,36 +242,51 @@ function MainScreen() {
...
+          onPress={sendCommand}
+          testID="send-command-button"
         />
-        <Button title="Send Command" onPress={sendCommand} />
 
         <Text style={styles.heading}>Result</Text>
-        <Text style={styles.result}>Last clicked: {lastClicked ?? '—'}</Text>
+        {/* The e2e test taps this label to dismiss the overflow menu, so it
+            must stay non-interactive. */}
+        <Text testID="last-clicked-result" style={styles.result}>
```
kligarski: I don't see `last-clicked-result` used in e2e.

FabricExample/e2e/native-class-names.ts
```diff
@@ -64,3 +64,8 @@ export const CLASS_NAME_UI_LABEL = 'UILabel';
 
 export const CLASS_NAME_ANDROID_APP_COMPAT_IMAGE_BUTTON =
   'androidx.appcompat.widget.AppCompatImageButton';
+export const CLASS_NAME_ANDROID_ACTION_MENU_ITEM_VIEW =
+  'androidx.appcompat.view.menu.ActionMenuItemView';
+export const CLASS_NAME_ANDROID_MENU_DROP_DOWN_LIST_VIEW =
+  'androidx.appcompat.widget.MenuPopupWindow$MenuDropDownListView';
+export const CLASS_NAME_ANDROID_LINEAR_LAYOUT = 'android.widget.LinearLayout';
```
kligarski: I don't see it used anywhere.

LKuchno: CLASS_NAME_ANDROID_LINEAR_LAYOUT was not used as you mention - removed it, but CLASS_NAME_ANDROID_MENU_DROP_DOWN_LIST_VIEW is used ie in lines 185-188
`const overflowRow = (title: MenuText): Detox.NativeMatcher =>
  by
    .text(title)
    .withAncestor(by.type(CLASS_NAME_ANDROID_MENU_DROP_DOWN_LIST_VIEW));`

PR #4422, opened 2026-07-29
chore(iOS, Split): Migrate RNSSplitNavigationController to obj-c

t0maboro:
## Description

Next part of dropping Swift from our codebase; here I am migrating the RNSSplitNavigationController. The code should look similar to the previous implementation; the main change was around key-value observer syntax.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1702

## Changes

- RNSSplitNavigationController: swift -> obj-c

N/A

## Test plan

Played for a while with SplitBaseApp and TestSplitPressables (on this one focused on pressables)

## Checklist

- [ ] Ensured that CI passes


t0maboro:
adding backport to v4 as we need it for SPM compatibility on 4.x path

ios/split/RNSSplitNavigationController.h
```diff
@@ -0,0 +1,32 @@
...
+
+#import <UIKit/UIKit.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class RNSSplitNavigationController;
+
+/// @brief A protocol that observes origin changes in an RNSSplitNavigationController’s view frame.
+///
+/// The subscriber will be notified when the view's origin changes.
```
kmichalikk: I'd rather stick to /* */ block comments in Obj-C

ios/split/RNSSplitNavigationController.mm
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  // This cast should be safe, as only our RNSSplitNavigationControllerViewFrameContext can reach that.
+  CGRect oldFrame = [oldValue CGRectValue];
+  CGRect newFrame = [newValue CGRectValue];
+
+  if (!CGPointEqualToPoint(oldFrame.origin, newFrame.origin)) {
+    [self onViewOriginChange];
+  }
+}
```
kmichalikk: I wonder if we could abstract it somehow out of the nav controller, the swift version was cleaner but here we have some responsibility mismatch IMO

kmichalikk (changes requested):

ios/split/RNSSplitNavigationController.h
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4422/changes/7386764089e52826ab61c239cad26c727689f1b0

ios/split/RNSSplitNavigationController.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4422/changes/f3bd48612f81176d019b05aa6bf54e83cc529f27

ios/split/RNSSplitNavigationControllerFrameObserver.mm
```diff
@@ -0,0 +1,78 @@
+#import "RNSSplitNavigationControllerFrameObserver.h"
+
+static void *RNSSplitNavigationViewFrameContext = &RNSSplitNavigationViewFrameContext;
+
+@implementation RNSSplitNavigationControllerFrameObserver {
+  RNSSplitNavigationController *__weak _navigationController;
+  UIView *__weak _observedView;
+}
```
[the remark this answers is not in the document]
kmichalikk: let's think about it

ios/split/RNSSplitNavigationControllerFrameObserver.mm
```diff
@@ -0,0 +1,78 @@
...
+
+  if (!CGPointEqualToPoint(oldFrame.origin, newFrame.origin)) {
+    [self notifyViewFrameOriginDidChange];
+  }
+}
+
+- (void)notifyViewFrameOriginDidChange
+{
+  RNSSplitNavigationController *navigationController = _navigationController;
+  [navigationController.viewFrameOriginChangeObserver splitNavCtrlViewDidChangeFrameOrigin:navigationController];
```
kmichalikk: I don't like how we get the controller, access it properties and call methods on them. Here we should only observe view frame and notify the navigation controller. A method on navigationController that calls this inside would be better for separation. If you think we're going too deep, you may do the whole observer thing in another PR.

kmichalikk (changes requested):

ios/split/RNSSplitNavigationControllerFrameObserver.mm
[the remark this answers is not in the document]
t0maboro: if the observed view deallocates with the observer still attached, KVO takes care of the cleanup + FrameObserver is strongly owned by RNSSplitNavigationController, so we should avoid cycle

ios/split/RNSSplitNavigationControllerFrameObserver.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4422/commits/63ae0aa594e43a19a932f84b290084d661c9491b

t0maboro: should we consider a similar change in the bottom accessory? https://github.com/software-mansion/react-native-screens/blob/main/ios/tabs/bottom-accessory/RNSTabsBottomAccessoryHelper.mm#L192-L197

ios/split/RNSSplitNavigationController.h
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * @class RNSSplitNavigationController
+ * @brief A subclass of UINavigationController, creates a view that wraps view associated with
+ * RNSSplitScreenController.
+ *
+ * This subclass is responsible for tracking when the underlying view's frame origin changes,
+ * allowing for syncing the ShadowTree layout.
+ *
+ * It observes origin changes via key-value observer and notifies a delegate.
```
kmichalikk: nit: the comment could be more specific

ios/split/RNSViewFrameOriginChangeDelegate.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#import <UIKit/UIKit.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class RNSSplitNavigationController;
+
+@protocol RNSViewFrameOriginChangeDelegate <NSObject>
```
kmichalikk: One thing that I had been reminded about in my CRs is that this sounds generic as if you could use it for different kind of views, but it's specific to SplitNavigationController. Let's rename that + method name to something specific.

ios/split/RNSSplitNavigationControllerFrameObserver.h
```diff
@@ -0,0 +1,28 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * @class RNSSplitNavigationControllerFrameObserver
+ * @brief Class responsible for observing frame origin changes of RNSSplitNavigationController's view.
+ */
+@interface RNSSplitNavigationControllerFrameObserver : NSObject
+
+- (instancetype)initWithSplitNavigationController:(RNSSplitNavigationController *)navigationController;
```
kmichalikk: Alternatively, we could just pass RNSViewFrameOriginChangeDelegate here and avoid any chain calling

ios/split/RNSSplitNavigationControllerFrameObserver.mm
```diff
@@ -0,0 +1,72 @@
...
+  id newValue = change[NSKeyValueChangeNewKey];
+  if (![oldValue isKindOfClass:[NSValue class]] || ![newValue isKindOfClass:[NSValue class]]) {
+    return;
+  }
+
+  CGRect oldFrame = [oldValue CGRectValue];
+  CGRect newFrame = [newValue CGRectValue];
+
+  if (!CGPointEqualToPoint(oldFrame.origin, newFrame.origin)) {
+    [_navigationController viewFrameOriginDidChange];
```
kmichalikk: we have two options based on other comments: make the observer more generic, or leave it SplitNavigationController-specific. For the latter, let's make the method name different than the RNSViewFrameOriginChangeDelegate one, otherwise it's confusing IMO.

kmichalikk (changes requested):
Sorry, it doesn't feel right to me still :(

ios/split/RNSSplitNavigationController.h
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4422/changes/0a241b4e363137390fb3a0f83af040c7149cd9dd

ios/split/RNSViewFrameOriginChangeDelegate.h
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4422/changes/853916116f9bd7e3a17ab08c2be3e6deb1a7f2da

ios/split/RNSSplitNavigationControllerFrameObserver.h
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4422/changes/6f8a8a86a5a42d458ed2f26671b1a924a3713940

ios/split/RNSSplitNavigationControllerFrameObserver.mm
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4422/changes/6f8a8a86a5a42d458ed2f26671b1a924a3713940

kmichalikk (approved):

PR #4428, opened 2026-07-30
chore(test): e2e test for test-stack-toolbar-menu-groups-android

LKuchno:
## Description

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1697

Adds Detox e2e coverage for the **Stack Toolbar Menu Groups (Android)** scenario,
which previously had `e2eCoverage: 'tbd'`.

The interesting part is that Detox cannot read a menu item's checked state on
Android directly. Instead each item's rendered state is asserted on the widget
AppCompat inflates inside its `ListMenuItemView` row, addressed by the row's
title. Since `setGroupCheckable(group, true, exclusive)` gives multi-toggle
groups a `CheckBox` and single-selection groups a `RadioButton`, the widget
class doubles as an assertion about the group type — so the tests also catch a
group silently losing its `singleSelection`, which the manual scenario does not
check. `toHaveToggleValue` then reads the checked state.

The `onSelectionChange` / `onPress` payloads are asserted separately on the
toast they produce, via the `dismissNextToast` / `expectNoToast` helpers.

## Changes

- New `FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-groups-android.e2e.ts`
  — 18 cases covering steps 1-34. They run as one continuous flow, like a manual
  run: each starts from the state the previous one left, and the menu is only
  rebuilt where a step flips a props switch.
- `FabricExample/e2e/e2e-utils.ts` — shared helpers extracted for this test:
  - `selectPickerOption` / `toggleSettingsSwitch` drive a `SettingsPicker` /
    `SettingsSwitch` and assert the resulting state. `selectPickerOption`
    returns early when the picker already shows the target value, so a repeated
    command costs one read and no gesture.
  - `rewindAndScrollUntilVisible` scrolls to the top first, since Detox's
    `whileElement` only scrolls one way.
  - `scrollUntilVisible` takes optional `pixels` / `startPercentage`. Defaults
    match the previous hardcoded values, so existing callers are unchanged.
  - `dismissNextToast` / `expectNoToast` for asserting the toast queue.
- Added `CLASS_NAME_ANDROID_CHECK_BOX` and `CLASS_NAME_ANDROID_RADIO_BUTTON` to
  `FabricExample/e2e/native-class-names.ts`.
- Added testIDs to the test screen (scroll view, four command pickers, send
  button, three config switches). No behavioural change to the screen.
- `scenario-description.ts`: `e2eCoverage` `'tbd'` -> `'full'`; `scenario.md`
  E2E section updated.


FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-groups-android.e2e.ts
```diff
@@ -0,0 +1,544 @@
...
+  await tapById('send-command-button');
+}
+
+async function openOverflowMenu() {
+  await element(by.label(OVERFLOW_MENU_LABEL)).tap();
+}
+
+async function closeMenu() {
+  await device.pressBack();
+}
```
t0maboro: called only from `finally` block, shouldn't we ensure that the menu is open before executing `pressBack`, is there any risk for dropping the screen?

FabricExample/e2e/e2e-utils.ts
```diff
@@ -40,11 +40,32 @@ export function isIOSVersionAtLeast(version: string): boolean {
...
+) {
   await waitFor(element(by.id(id)))
     .toBeVisible()
     .whileElement(by.id(scrollViewId))
-    .scroll(600, 'down', Number.NaN, 0.85);
+    .scroll(pixels, 'down', Number.NaN, startPercentage);
+}
+
+/** Rewinds to the top first — `whileElement` only scrolls one way. */
+export async function rewindAndScrollUntilVisible(
```
t0maboro, line 60: Assuming that there will be a follow-up PR that will utilize features exported from utils - then having exports for the fuctions that we're expecting to use in tests is fine. Applies to all below.

t0maboro (approved):

FabricExample/e2e/e2e-utils.ts
LKuchno: https://github.com/software-mansion/react-native-screens/pull/4428/changes/bbcaeaab500d56a6e4f98e51034e50c3dcc3ed4f
I change e2e test to use already existing functions from e2e-utils file :)

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-groups-android.e2e.ts
LKuchno: Extending function with check if Menu is open:
https://github.com/software-mansion/react-native-screens/pull/4428/changes/d21ceb7fb678421df429dd8e7237fbe8a2d58e49

Issue #1754, opened 2023-04-12
Memory Leak. iOS. Strong reference cycle between RNSScreen and RNSScreenView


mkondakov:
### Description

Instance of RNSScreen class creates strong reference cycle with instance of RNSScreenView class. This issue is critical in our project where user can open and close react-native app inside native iOS app many times.

### Steps to reproduce

1. clone project
2. open TestRNSScreens folder in terminal
3. run "pod install"
4. open "TestRNSScreens.xcworkspace" in Xcode
5. run
6. click on "Open RN App" button
7. click on "close" button
8. repeat 6 and 7 steps few times
9. check memory usage in Xcode

### Snack or a link to a repository

https://github.com/mkondakov/RNSScreensMemoryLeak

### Screens version

3.20.0

### React Native version

0.71.6

### Platforms

iOS

### JavaScript runtime

None

### Workflow

React Native (without Expo)

### Architecture

None

### Build type

Release mode

### Device

Real device

### Device model

all models

### Acknowledgements

Yes

ajstokar:
I believe we are seeing the same issue.  Navigating back and forth between 2 screens causes memory to consistently go up.

mkondakov83:
@kkafar do you have any progress on this issue? We have found that this memory leak is the main reason why application crashes because of low memory conditions. In our app we open and close react-native instances a lot.

kkafar:
Hey, small update from my side. 
I'm investigating the issue but I've not determined root cause of the leak yet unfortunately. 

1. It seems that in standard environment (fully React Native application) there is no leak

<img width="2995" alt="image" src="https://github.com/software-mansion/react-native-screens/assets/50801299/67e15022-c0c8-4fa2-8e8a-140ec558bf67">

You can see in Xcode log console that I've pushed & popped few screens, but in the end there are only two `RNSScreen` controllers allocated -- the first one (for the first screen) and most recent one for the second screen. 

However, at the same time memory usage has risen a bit: 

<img width="601" alt="image" src="https://github.com/software-mansion/react-native-screens/assets/50801299/f7837a44-a67e-43fd-aa5d-1ecfaedfc815">

But as I can detect no apparent leak it nudges me towards conclusion that GC simply was not triggered yet. 


2. I can confirm the leak being present in your reproduction, however it is not clear for me, what change in application setup introduces the key difference here.

Just letting you know that I'm looking, will be back once I have solution.

mkondakov:
@kkafar 
[...]

PR #4433, opened 2026-07-31
fix(ios): deallocate screens on Fabric pop (break RNSScreen<->RNSScreenView retain cycle)

haroon-afzal-tahir:
## Description

On the New Architecture (Fabric), a native-stack screen is **never deallocated after it is popped** — the `RNSScreen` view controller, its `RNSScreenView`, and the entire React subtree they retain stay alive for the rest of the process. Over a long session this grows until the app runs out of memory (in our app it surfaced as a Hermes `GCBase::oom`).

The cause is a retain cycle:

- `RNSScreenView` holds its controller strongly — `RNSScreen.h`: `@property (nonatomic, retain) RNSScreen *controller;`
- `RNSScreen` holds the view back strongly — `self.view` **and** the non-`__weak` `_initialView` ivar (set in `initWithView:`).

The library already contains the break: `-[RNSScreenView invalidateImpl]` nils `_controller` (deferred to the main queue). But its only trigger is `-invalidate` (`RCTInvalidating`), which is a Paper-era hook that is **never called on Fabric**, so the cycle is never broken and every popped screen leaks.

Refs #1754.

## Changes

- Call `[screenChildComponent invalidateImpl]` from `-[RNSScreenStack unmountChildComponentView:index:]` (right after `addSnapshotToView`), so the existing teardown runs on the Fabric unmount path.
- The `_controller = nil` is deferred to the next main-queue tick (unchanged `invalidateImpl` behavior), so the in-flight pop animation — which uses the snapshot added just above — is unaffected. Once the pop finishes and the navigation controller releases the `RNSScreen`, the whole screen deallocates.
- `+[RNSScreenView shouldBeRecycled]` returns `NO`, so the component view is never reused after unmount — nilling the controller here is safe.

No JS/public API change.

## Test plan

Minimal repro: a `@react-navigation/native-stack` with two screens; repeatedly push then pop the second screen while recording **Xcode Instruments → Allocations**, filtered to `RNSScreen` / `RNSScreenView`.

- **Before:** `# Persistent` for `RNSScreen` / `RNSScreenView` increases by one on every push/pop and never decreases.
- **After:** those instances become transient — deallocated on pop; `# Persistent` stays flat.

Verified on iOS with RN New Architecture (Fabric) enabled.

## Checklist

- [x] For API changes, updated relevant public types. — n/a, no API change
- [x] Included steps that can be used to test this change (Instruments repro above).
- [ ] For visual changes, included screenshots / GIFs / recordings. — n/a, no visual change
- [ ] Ensured that CI passes.

### DIFF
--- a/ios/legacy/RNSScreenStack.mm
+++ b/ios/legacy/RNSScreenStack.mm
@@ -1317,6 +1317,16 @@ RNS_IGNORE_SUPER_CALL_END
   RNSScreenView *screenChildComponent = (RNSScreenView *)childComponentView;
   [screenChildComponent.controller addSnapshotToView];
 
+  // Break the RNSScreenView <-> RNSScreen (controller) retain cycle when a screen
+  // is popped. `invalidateImpl` nils `_controller` on the next main-queue tick, so
+  // the running pop animation (which uses the snapshot added above) is unaffected;
+  // once it completes and the navigation controller releases the RNSScreen, the
+  // whole screen (view + controller + its React subtree) deallocates. Its usual
+  // trigger `-invalidate` (RCTInvalidating) is never called on the New
+  // Architecture, so without this the cycle leaks every pushed screen (see #1754).
+  // `+shouldBeRecycled` is NO, so the view is never reused after this.
+  [screenChildComponent invalidateImpl];
+
   RCTAssert(screenChildComponent.reactSuperview == self,
             @"Attempt to unmount a view which is mounted inside different view. (parent: %@, child: %@, index: %@)",
             self,


kkafar:
Hey @haroon-afzal-tahir!

Thanks for the report, however the analysis you provide is strictly wrong:

> The library already contains the break: -[RNSScreenView invalidateImpl] nils _controller (deferred to the main queue). But its only trigger is -invalidate (RCTInvalidating), which is a Paper-era hook that is never called on Fabric, so the cycle is never broken and every popped screen leaks.

`- [RCTComponentViewProtocol invalidate]` message exists on strictly Fabric related [`RCTComponentViewProtocol`](https://github.com/react/react-native/blob/6c4cf29689d6c330239bdaf1b974eff80552f35f/packages/react-native/React/Fabric/Mounting/RCTComponentViewProtocol.h#L114-L118) and is called for every single component removed from view hierarchy that is not a subject to view-recycling mechanism [(link)](https://github.com/react/react-native/blob/6c4cf29689d6c330239bdaf1b974eff80552f35f/packages/react-native/React/Fabric/Mounting/RCTComponentViewRegistry.mm#L111).

If there is a leak, please report it as an issue with a reproduction attached.

PR #4435, opened 2026-07-31
Android: honor gestureEnabled for form sheet dismissal

fabriziocucci:
## Description

On Android a formSheet can be swiped away even with `gestureEnabled` set to `false`. On iOS `gestureEnabled={false}` maps to `modalInPresentation` and blocks dismissal by drag, so the two platforms disagree. This brings Android to parity: `gestureEnabled={false}` keeps the sheet from being dismissed by dragging.

The gamma form sheet already exposes `preventNativeDismiss`. This targets the legacy path that react-navigation and expo-router still use, so apps on the current stack API get the parity now.

## Changes

- The legacy bottom sheet behaviour hardcoded `isHideable = true`. It now uses `isHideable = screen.isGestureEnabled` in `ScreenModalFragment` and `SheetDelegate`.
- With `gestureEnabled={false}` the drag stops at the smallest detent instead of dismissing. The sheet stays draggable between detents.
- No public API changes. `gestureEnabled` already exists; its effect now covers the Android form sheet.

## Test plan

Minimal repro (react-navigation native stack). Run on Android, open the sheet, drag it down.

- Before this change: the sheet is dismissed.
- After: it stops at the smallest detent (`0.3`).
- `gestureEnabled: true`: unchanged, drag down dismisses.
- iOS: unaffected.

```tsx
import { Button, Text, View } from 'react-native';

  return (
    <View style={{ flex: 1, justifyContent: 'center' }}>
      <Button title="Open sheet" onPress={() => navigation.navigate('Sheet')} />
    </View>
  );
}

function Sheet() {
  return (
    <View style={{ flex: 1, padding: 24 }}>
      <Text>Drag me down. With gestureEnabled false I stop at the smallest detent instead of closing.</Text>
    </View>
  );
}

  return (
    <NavigationContainer>
      <Stack.Navigator>
        <Stack.Screen
          name="Sheet"
          component={Sheet}
          options={{
            presentation: 'formSheet',
            gestureEnabled: false,
            sheetAllowedDetents: [0.3, 0.6, 1.0],
            sheetGrabberVisible: true,
          }}
        />
      </Stack.Navigator>
    </NavigationContainer>
  );
}
```

## Checklist

- [x] For API changes, updated relevant public types. (No API changes in this PR.)
- [ ] Ensured that CI passes.

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/legacy/ScreenModalFragment.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/legacy/ScreenModalFragment.kt
@@ -233,7 +233,9 @@ class ScreenModalFragment :
         check(containerHeight != null) { "[RNScreens] Failed to find window height during bottom sheet behaviour configuration" }
 
         behavior.apply {
-            isHideable = true
+            // Honor gestureEnabled so gestureEnabled = false blocks drag-to-dismiss on
+            // Android, matching iOS (modalInPresentation). Still draggable between detents.
+            isHideable = screen.isGestureEnabled
             isDraggable = true
         }
 
--- a/android/src/main/java/com/swmansion/rnscreens/legacy/bottomsheet/SheetDelegate.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/legacy/bottomsheet/SheetDelegate.kt
@@ -187,7 +187,9 @@ class SheetDelegate(
         }
 
         behavior.apply {
-            isHideable = true
+            // Honor gestureEnabled so gestureEnabled = false blocks drag-to-dismiss on
+            // Android, matching iOS (modalInPresentation). Still draggable between detents.
+            isHideable = screen.isGestureEnabled
             isDraggable = true
         }
 


t0maboro:
Hi @fabriziocucci , thank you for opening this PR! I think having parity here definitely makes sense, but this change requires some additional work. On iOS, when the `gestureEnabled` value changes dynamically, the sheet reacts to the prop update, whereas on Android it currently doesn't. This happens because `configureBottomSheetBehaviour` is called only:
1. when the sheet is created
2. when insets are changed

so this flow is currently independent of `gestureEnabled` prop changes.

I can add this to my backlog and try to investigate it further. At the moment, though, most of our capacity is focused on the new v5 implementation, so I can't give a concrete timeline for when I'll be able to pick this up.

(note for me: also noticed that tapping the backdrop on Android always dismisses the sheet, while on iOS it's dependent on `gestureEnabled` prop - that's a completely separate topic that shouldn't be covered in this PR, but sth we should investigate to align the implementation across platforms)

fabriziocucci:
Thanks, that's a fair point and I hadn't traced `configureBottomSheetBehaviour` that far. You're right that a dynamic `gestureEnabled` change won't be picked up on Android the way it is on iOS.

Two things though. Today `gestureEnabled={false}` has no effect at all on the Android form sheet, so the sheet dismisses on drag regardless of what the prop says. And in practice `gestureEnabled` is almost always set once in screen options rather than toggled at runtime, so honouring it at configure time covers how the prop is actually used in the overwhelming majority of apps.

That leaves the dynamic case, which is real and worth fixing, just much less common. This PR doesn't make it any harder to get to: the follow-up is still "reconfigure the behaviour when the prop changes" and it's orthogonal to the two lines here.

Agreed the backdrop behaviour is a separate topic and shouldn't be in this PR.

So: is there a route where this lands as a stop-gap and the dynamic update gets tracked separately? Given your capacity is on v5, I'm happy to take the follow-up myself.

Mostly trying to understand what the options are.

t0maboro:
Okay, I agree that we can try merging it in this form.

If someone needs the dynamic case later, we can always track it as another feature request, but we should clearly mention in this PR description that the dynamic case isn't covered. For example, a use case like "prevent dismissal until some action inside the modal is taken, then unblock" would be a good reason to support dynamic updates.

I'll attach an example to have this covered in testing before the next release. From my side, I think we can proceed with this for now.

t0maboro:
One more thing I've noticed. In the downstream implementation of React Navigation, `gestureEnabled` is always set to `false` on Android, so passing the prop from it won't have any effect for now - just marking it as sth that we need to be aware of: https://github.com/react-navigation/react-navigation/blob/7.x/packages/native-stack/src/views/NativeStackView.native.tsx#L371-L377 . Passing this value through would require additional coordination with the native stack implementation in react-navigation.
Since this has been the case for several years, I'd prefer to avoid changing that behavior and instead look for a different solution (e.g., exposing a dedicated prop for that). Additionally, we'd need to consider other presentation types as well. I'm not sure it would be good to mark this API available on Android when it won't have any effect for the basic Screen presentation.

fabriziocucci:
We've been running this patch in production on Expo SDK 56 for a while and the prop does take effect on Android.

Two form sheets in the same build, identical apart from this option:
- `gestureEnabled: false`: drag down stops at the smallest detent.
- `gestureEnabled: true`: drag down dismisses as before.

So on the formSheet path the value is reaching `Screen` and `isHideable` follows it. Worth noting we're on expo-router. I've digged a bit into the code and found the line you linked, so it seems the sheet path takes a different route to the native side than the card path does?

Your point about other presentation types stands though. I only have evidence for formSheet and if `card` silently ignores it then having the prop work in one place and not the other is not amazing, although the net effect is still positive.

So what would you prefer: scope this to formSheet and document it as such or go with a dedicated prop like you suggested?

Happy either way.

PR #4437, opened 2026-07-31
feat(Stack v5, iOS): Add identifier for header items animation

kmichalikk:
Closes https://github.com/software-mansion/react-native-screens-labs/issues/1711

## Description

This PR adds `identifier` prop to new iOS header implementation. The default heuristic uses visual similarity when animating items between screens. When the item that should be considered logically the same looks differently between screens, it probably going to be crossfaded instead, but when identifier is added, UIKit knows to match those 2.

> [!note]
> For now, user is expected to give valid values for identifier, with 1:1 matches between, no extra validation is present.

## Changes

- added `identifier` prop
- added SFT for testing

> [!note]
> Having the "same" items across screens with different symbols / views is required to break the heuristic, otherwise it works without the identifier. See the video attached.

https://github.com/user-attachments/assets/daf8d133-52a2-4411-8691-714079f0af5e

## Test plan

Use the SFT added in the PR. Validate the flow present in the video.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/stack/screen/RNSStackScreenHeaderCoordinator.mm
+++ b/ios/stack/screen/RNSStackScreenHeaderCoordinator.mm
@@ -471,6 +471,14 @@
                                                                    withHeaderEventsDelegate:_eventsDelegate
                                                                             withImageLoader:_imageLoader];
 
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
+  if (@available(iOS 26.0, *)) {
+    if (item.identifier != nil) {
+      barButtonItem.identifier = item.identifier;
+    }
+  }
+#endif // RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
+
   if (item.menu != nil && item.itemId != nil) {
     RNSStackHeaderMenuToggleStateTracker *tracker = [_trackerRegistry trackerForItemId:item.itemId];
     __weak auto weakSelf = self;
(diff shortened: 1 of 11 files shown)


kmichalikk:
Haven't tested yet what happens when:
- ~user uses the same identifier for many items~ only visual issues, up to developer to keep it correct
- ~the item is in overflow menu for one of the screens~ native handles this in some way, nothing to do
- ~the items are `imageSource` and loaded asynchronously~ items not visible until loaded on push, they work correctly when going back
- ~running on iOS 18~ prop not available, I didn't see a regression

ios/stack/screen/RNSStackScreenHeaderCoordinator.mm
```diff
@@ -471,6 +471,14 @@ - (UIBarButtonItem *)buildBarButtonItemForItem:(id<RNSStackHeaderItemDataProvidi
                                                                    withHeaderEventsDelegate:_eventsDelegate
                                                                             withImageLoader:_imageLoader];
 
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
+  if (@available(iOS 26.0, *)) {
+    if (item.identifier != nil) {
+      barButtonItem.identifier = item.identifier;
```
t0maboro, line 477: Q: What happens when someone resets the value explicitly, changing the identifier to `undefined` in JS?

t0maboro (approved):

kligarski (approved):

ios/stack/screen/RNSStackScreenHeaderCoordinator.mm
kmichalikk: This works but results in item rebuild & you no longer have the identifier. Same for changing it to sth else.

PR #4440, opened 2026-08-03
fix(iOS): add fall back to window.safeAreaInsets when no ancestor provider is found 

Loloekk:
## Description

On iOS, SafeAreaView did not apply insets unless it had an ancestor provider such as Screen, tabs, or a stack container. SafeAreaView from react-native-safe-area-context similarly requires a SafeAreaProvider.

Using the SafeAreaView itself as a fallback provider (as RNSAC does with return self) does not work with our implementation. We apply insets as margin, which changes the view’s frame. UIKit then recomputes safeAreaInsets for that smaller frame, which updates the margin again and causes an infinite layout loop.

When no ancestor provider is found, this change falls back to window.safeAreaInsets. That gives stable system chrome insets (status bar, home indicator, etc.) without the feedback loop. This matches the intended no-provider case: there is no navigation chrome to account for, so window-level system insets are the correct source.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1661

## Changes

- Updated `ios/safe-area/RNSSafeAreaViewComponentView.mm` to fall back to window insets when no ancestor provider is found, instead of skipping the update

| Before | After |
| --- | --- |
| simulator_screenshot_B58A1CD0-07CC-4BF9-B77A-3368FE70E8CD | simulator_screenshot_8E9A1FA1-625E-485C-95C5-653F60CAC2BC |

## Test plan
- Minimal repro: `Test4440` — `SafeAreaView` from `react-native-screens/experimental` with no provider screen; verify insets on iOS.
- Checked `TabsContainer` inside `SafeAreaView` — works as expected (no overlaps or other layout issues).

## Checklist

- [ ] Ensured that CI passes


apps/src/tests/issue-tests/Test4440.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
+import { Text, View } from 'react-native';
+import { SafeAreaView } from 'react-native-screens/experimental';
+export default function App() {
+  return (
+    // SafeAreaView applies insets as margin (not padding), so backgroundColor
+    // does not paint into the system safe-area — the parent shows through there.
+    <SafeAreaView
+      style={{ flex: 1, backgroundColor: 'blue' }}
```
t0maboro, line 9: nit: we have our own color palette, which you can use `import { Colors } from '@apps/shared/styling';`

apps/src/tests/issue-tests/Test4440.tsx
```diff
@@ -0,0 +1,18 @@
...
+      edges={{ top: true, bottom: true }}>
```
t0maboro, line 10: I'd extend this test a bit, allowing all 4 edges to be programmatically configurable using some Buttons and state, having an option to test more valid configurations in a single example.

ios/safe-area/RNSSafeAreaViewComponentView.mm
```diff
@@ -79,13 +79,28 @@ - (void)safeAreaProviderInsetsDidChange:(NSNotification *)notification
...
+  [super safeAreaInsetsDidChange];
+
   if (_providerView == nil) {
+    [self updateStateIfNeeded];
+  }
+}
+
+- (UIEdgeInsets)fallbackSafeAreaInsets
+{
+  return self.window.safeAreaInsets;
```
t0maboro, line 93: I have a concern about this approach that window insets aren't respecting the view's position, while on native iOS own `safeAreaInsets` are calculated with respect to the view hierarchy. Reading the value from the window may incorrectly apply the margin in e.g. modals, which are positioned below the top inset.
I see the problem with using `self` as an insets fallback combined with applying margins instead of padding.

Right now I'm considering a different approach, which would be exposing an explicit `SafeAreaProvider` component to the public API that would implement `RNSSafeAreaProviding` with its own `safeAreaInsets` + change notifications, and log a warning when there's no provider at all. Anyway, I'd wait with the final decision for @kkafar, as he did much more research regarding SAV.

t0maboro (commented):
left some comments, I'd put this PR on hold until @kkafar return from PTO

ios/safe-area/RNSSafeAreaViewComponentView.mm
kkafar: I want to cc @kligarski here.

I'm not certain about right direction here. Would it make sense to simply make a root view controller of each hierarchy a default provider in such cases? wdyt?

kligarski: > Would it make sense to simply make a root view controller of each hierarchy a default provider in such cases? wdyt?

I'm not sure what you mean here. We would need to make React's root view controller implement `RNSSafeAreaProviding`?

Regarding the issue highlighted by @t0maboro, I think I agree with the concern regarding the solution in this PR. I think we should consider either separate provider or changing how SAV work from margin to padding - I don't think we had any particular reason why we went with the margin approach but I might be forgetting something here.

kkafar: @kligarski No, I mean that we can use the root view controller / react root view as a search barrier and use it's `safeAreaInsets` as the baseline there. 

If we went for this, we'd look for a top level view controller in each view hierarchy hierarchy (main react one, or modal) and query it's view for the safe area insets.

kligarski: I'm not sure if it really helps with the issue mentioned by @t0maboro. I guess that for modals and top inset it would help but in general case if the safe area is placed partially inside the _non-safe area_, this still won't work as intelligently as UIKit usually handles it (at least I think so).

image

kkafar: I don't want to engage into making reproduction right now, but I'm not convinced that root view of modal view hierarchy will have those safe area insets like in the drawing of yours.

I think this is something for @Loloekk to verify.

kligarski: The example is a general example, not related to modal in any way.

Here is something what I meant (I had to do the repro to explain it well):

https://github.com/user-attachments/assets/336c33dd-bac8-4f33-8bc0-6b175691d625

Green box is a child of the red box with a top constraint to the safe area of the red box. Even if red box is only partially obstructed by the status bar, the green box starts just at the start of safe area, no additional distance (distance between top of red box and top of green box is not constant to status bar height but it adapts).

<details>
<summary>UIKit repro</summary>

```swift
import UIKit

class SceneDelegate: UIResponder, UIWindowSceneDelegate {
    var window: UIWindow?

    func scene(_ scene: UIScene,
               willConnectTo session: UISceneSession,
               options connectionOptions: UIScene.ConnectionOptions) {
        guard let windowScene = scene as? UIWindowScene else { return }
        let window = UIWindow(windowScene: windowScene)
        window.rootViewController = SafeAreaDebugVC()
        self.window = window
        window.makeKeyAndVisible()
    }
}

final class SafeAreaDebugVC: UIViewController {

    private let container = UIView()   // red   – offset N from top of screen
    private let child     = UIView()   // green – pinned to container's safe area top
    private let marker    = UIView()   // blue  – root view safe area top, for reference

    private let slider = UISlider()
    private let readout = UILabel()
    private var containerTop: NSLayoutConstraint!

    override func viewDidLoad() {
        super.viewDidLoad()
        view.backgroundColor = .black

        container.backgroundColor = .systemRed
        child.backgroundColor     = .systemGreen
        marker.backgroundColor    = .systemBlue

        view.addSubview(container)
        container.addSubview(child)
        view.addSubview(marker)
        view.addSubview(slider)
        view.addSubview(readout)

        readout.numberOfLines = 0
        readout.textColor = .white
        readout.font = .monospacedSystemFont(ofSize: 13, weight: .regular)

        slider.minimumValue = -40
        slider.maximumValue = 160
        slider.value = 10
        slider.addTarget(self, action: #selector(sliderChanged), for: .valueChanged)

        [container, child, marker, slider, readout].forEach {
            $0.translatesAutoresizingMaskIntoConstraints = false
        }

        containerTop = container.topAnchor.constraint(equalTo: view.topAnchor, constant: 10)

        NSLayoutConstraint.activate([
            containerTop,
            container.leadingAnchor.constraint(equalTo: view.leadingAnchor),
            container.trailingAnchor.constraint(equalTo: view.trailingAnchor),
            container.heightAnchor.constraint(equalToConstant: 300),

            // child: pinned to the CONTAINER's own safe area top
            child.topAnchor.constraint(equalTo: container.safeAreaLayoutGuide.topAnchor),
            child.leadingAnchor.constraint(equalTo: container.leadingAnchor, constant: 24),
            child.trailingAnchor.constraint(equalTo: container.trailingAnchor, constant: -24),
            child.heightAnchor.constraint(equalToConstant: 60),

            // reference line at the root view's safe area top
            marker.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor),
            marker.leadingAnchor.constraint(equalTo: view.leadingAnchor),
            marker.widthAnchor.constraint(equalToConstant: 24),
            marker.heightAnchor.constraint(equalToConstant: 2),

            slider.leadingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leadingAnchor, constant: 24),
            slider.trailingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.trailingAnchor, constant: -24),
            slider.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor, constant: -24),

            readout.leadingAnchor.constraint(equalTo: slider.leadingAnchor),
            readout.trailingAnchor.constraint(equalTo: slider.trailingAnchor),
            readout.bottomAnchor.constraint(equalTo: slider.topAnchor, constant: -16),
        ])
    }

    @objc private func sliderChanged() {
        containerTop.constant = CGFloat(slider.value.rounded())
        view.layoutIfNeeded()   // forces safe area propagation + updateReadout via layout pass
    }

    override func viewDidLayoutSubviews() {
        super.viewDidLayoutSubviews()
        readout.text = """
        offset               : \(containerTop.constant)
        root.safeArea.top    : \(view.safeAreaInsets.top)
        container.minY       : \(container.frame.minY)
        container.safeArea   : \(container.safeAreaInsets.top)
        child.minY (in cont) : \(child.frame.minY)
        child top in window  : \(child.convert(child.bounds, to: nil).minY)
        """
    }
}

```

</details>

Loloekk: @kkafar @kligarski I’ve checked this for modals and for the general example.

Both of you are right, using the root view controller’s insets works correctly for modals. It does not work in the example corresponding to the native one above — it takes static insets, independent of where the SAV is placed.

As I discussed with @t0maboro when I started this PR, you decided to use margin rather than padding so the whole SAV stays in the safe area, not only its content. If the SAV has a background, that background should only be in the safe area.

So the only way I see is to implement a SAV provider. Of course that provider also has to be used in the right place: if it wraps both the SAV and the blue banner (as in the drawing), it will behave the same as using the window’s insets.

PR #4448, opened 2026-08-04
chore(test): Android test-stack-toolbar-nested-menu-android coverage

LKuchno:
## Description

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1715

Adds full Detox e2e coverage for the `test-stack-toolbar-nested-menu-android`
scenario — Android toolbar menus with nested submenus. The scenario screen
already existed with `e2eCoverage: 'tbd'`; this PR implements the automation
for every step and flips the coverage to `full`.

The new spec exercises the four things the scenario describes:

- submenu structure and header resolution (`menuTitle`, with a fallback to
  `title`) at every nesting level, including a doubly nested submenu,
- click handling for items at every nesting level,
- imperative commands targeting leaf items, submenu containers, and items two
  levels down (`title`, `hidden`, `menuTitle`),
- props-update rebuild semantics — a single prop change drops all
  command-applied state and rebuilds the tree.

Two scenario steps (19 and 27) leave a submenu entry with no text for Detox to
match, so they are reached by row position within the overflow menu instead;
the absence of a submenu header is asserted indirectly, since a header would
otherwise show up as an extra match for the text it renders.

## Changes

- Added `FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-nested-menu-android.e2e.ts`
  (37 test cases across 9 `describe` blocks).
- `e2e-utils.ts`:
  - `getMatches`/`countMatches` accept an `orEmpty` option so an expected
    absence resolves to `[]`/`0` instead of throwing. It is opt-in precisely
    because it also swallows a crashed app or a dropped Detox connection.
  - Extracted `waitUntil(predicate, { timeout, interval, message })` for
    conditions Detox's `waitFor` cannot express. On Android `waitFor` is one
    native call retrying inside the app, so it asserts one property of one view
    and the matcher needs a literal `atIndex` — which rules out waiting on a
    match count, or on the last match while the count is still settling. Both
    come up here; the full rationale is in the function's JSDoc. `message`
    accepts a builder function so the timeout error can report what the last
    poll observed.
  - Documented that `tapTopmost` needs a freshly built matcher — Detox's
    `atIndex` rewrites the matcher in place on Android.
- `native-class-names.ts`: added `ListMenuItemView` (a popup menu row, the
  anchor for addressing widgets by item title) and `AppCompatImageView`
  (`group_divider` / `submenuarrow`).
- `test-stack-simple-nav.e2e.ts`: rewrote its two local polling helpers on top
  of the shared `waitUntil`. No behavior change.
- Test screen: added `testID`s to the pickers, switches, send-command button
  and result text, and exported the option types the e2e spec reuses.
- `scenario.md`: documented the e2e coverage and the phone-form-factor
  prerequisite — from 720dp smallest-width up, AppCompat opens submenus as
  cascading popups next to their parent rather than replacing it, which the
  test does not account for.


apps/src/tests/single-feature-tests/stack-v5/test-stack-toolbar-nested-menu-android/index.tsx
```diff
@@ -15,6 +15,7 @@ import {
 } from 'react-native-screens';
 import { scenarioDescription } from './scenario-description';
 
+// The option types are exported for the e2e test covering this screen.
```
t0maboro: nit: this comment is related to `type AllIds` rather than `ALL_IDS` - I'd move it lower

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-nested-menu-android.e2e.ts
```diff
@@ -0,0 +1,741 @@
...
+/**
+ * Sets a picker to `option`, then closes it: open option rows stay in the
+ * hierarchy and would collide with the `by.text` menu matchers. Re-picking the
+ * current value is skipped — it costs several taps and two scroll passes.
+ */
+async function selectOption(
+  pickerId: string,
+  pickerLabel: string,
+  option: string,
+) {
```
t0maboro, line 120: Same signature as for `FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-commands-android.e2e.ts`, the concept seems pretty similar as well. Do you think it would be possible to unify these at some point? Asking about a potential further refactor rather than making a change in this PR.

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-nested-menu-android.e2e.ts
```diff
@@ -0,0 +1,741 @@
...
+ * from popping the test screen itself. A left-over popup is never a local
+ * failure — every later matcher would resolve against the popup window.
+ */
+async function closeMenus() {
+  // One press per popup of the deepest path (overflow, submenu, nested
+  // submenu), plus one to notice an unexpected extra level.
+  const MAX_POPUP_DEPTH = 4;
+
+  for (let i = 0; i < MAX_POPUP_DEPTH; i++) {
+    if (await isScreenReachable()) {
```
t0maboro, line 230: Q: Any reason to rely on the screen reachability rather than `CLASS_NAME_ANDROID_MENU_DROP_DOWN_LIST_VIEW` presence?

t0maboro (approved):

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-nested-menu-android.e2e.ts
LKuchno: This function will be moved to e2e-utils in PR #4441 
Its already implemented here https://github.com/software-mansion/react-native-screens/blob/9757ee1ec363001f8c275ca87de1440ac6d0ed1b/FabricExample/e2e/single-feature-tests/tabs/test-tabs-tab-bar-layout-direction.e2e.ts 
but waiting for few more PR to be merged to don't make this changes twice :)

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-nested-menu-android.e2e.ts
LKuchno: Checking host screen visibility is better assertion than probing MenuDropDownListView, which can match a popup that's mid-dismiss and trigger an extra Back press that pops the test screen itself. This check prevents over-popping and guarantees what the next test actually needs: that matchers are targeting the main app window again. Single-press helpers like closeMenuIfOpen can stick with the class probe since they only hit Back once.

apps/src/tests/single-feature-tests/stack-v5/test-stack-toolbar-nested-menu-android/index.tsx
LKuchno: https://github.com/software-mansion/react-native-screens/pull/4448/changes/30ad07be1058af03c963f7040d1adaaefbb73b71

PR #4464, opened 2026-08-05
chore(test): e2e test for test-stack-toolbar-menu-disabled-android

LKuchno:
## Description

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1716

Adds Detox e2e coverage for the `test-stack-toolbar-menu-disabled-android`
scenario - the `disabled` prop on Android toolbar menu elements. The scenario
screen already existed with `e2eCoverage: 'tbd'`; this PR covers every step
(1–22) and flips the coverage to `incomplete`.

The spec covers the `disabled` state for each kind of menu element:

- via props - the toolbar-pinned action item, an overflow action item, the two
  checkable group items, a submenu container, and an item inside that submenu,
- via the `updateToolbarMenuElements` view command - the toolbar-pinned action
  item, the submenu and a checkable item, including the three-state reset
  (`undefined` clears the override and falls back to the prop value), matching
  scenario steps 18–22,
- the invariants around it - a disabled checkable item keeps its check when
  tapped, a disabled submenu does not open, and a props update rebuilds the
  menu and resets the toggle states.

Coverage is `incomplete` rather than `full` because Detox cannot read a tint:
the greyed-out appearance itself (`iconTintColorNormal` vs
`iconTintColorDisabled` on the action-bar button, and the greyed label of
disabled rows) is left to manual verification. The steps that assert one
(2, 4, 5, 8, 10, 12, 15, 18, 21) are covered through the native `enabled`
attribute instead — AppCompat mirrors `MenuItem.isEnabled()` onto the item
view — plus **Last Event** staying put after a tap. So every step has
automated coverage of what a disabled element must *do*; only how it *looks*
stays manual.

Two things the popup hierarchy forces:

- Assertions target `ListMenuItemView` (the popup row), not the title
  `TextView` inside it — `View.setEnabled` does not propagate to children, so
  only the row reflects the disabled state.
- Detox resolves matchers against a single window, so while a popup holds focus
  nothing behind it is addressable. Every step that straddles a menu opening or
  dismissing is bracketed by explicit waits, and menus are closed through a
  `withMenu` helper that runs even when an assertion throws — a popup left over
  would fail every later step in this stateful suite.

## Changes

- Added `FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-disabled-android.e2e.ts`
  (26 test cases across 9 `describe` blocks).
- Test screen: added `testID`s (scroll view, result text, both pickers,
  send-command button, per-item switches) and extracted the menu titles into an
  `ELEMENT_TITLES` constant. `AllIds`, `CmdDisabledOption`, `ItemLabels`,
  `ElementTitles`, `NoEvent` and `HeaderTitle` are now exported so the e2e spec
  imports them as types — a rename on the screen fails type-checking in the
  spec instead of failing the test at runtime.
- `scenario.md` / `scenario-description.ts`: documented what is and is not
  automated, and set `e2eCoverage: 'incomplete'`.


FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-disabled-android.e2e.ts
```diff
@@ -0,0 +1,564 @@
...
+// `waitForScreen` reported rather than thrown — stacked popups keep it false.
+async function isScreenAddressable(): Promise<boolean> {
+  return waitFor(element(by.id(SCROLLVIEW_ID)))
+    .toBeVisible()
+    .withTimeout(MENU_ANIMATION_TIMEOUT_MS)
+    .then(
+      () => true,
+      () => false,
+    );
+}
```
t0maboro, line 100: should be possible to reuse waitForScreen, sth like that
```
async function isScreenAddressable(): Promise<boolean> {
  return waitForScreen().then(
    () => true,
    () => false,
  );
}
```

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-disabled-android.e2e.ts
```diff
@@ -0,0 +1,564 @@
...
+
+// Mirrors the option `testID` that `SettingsPicker` derives from its `label`.
+// @see apps/src/shared/SettingsPicker.tsx
+function optionId(pickerLabel: string, option: string): string {
+  return `${pickerLabel.split(' ').join('-')}-${option}`.toLowerCase();
+}
+
+// Closing the picker again matters: its option rows stay in the hierarchy and
+// would collide with the `by.text` matchers used for the toolbar menu items.
+async function selectOption(
```
t0maboro, line 261: Some methods are generic enough that we might consider extracting them to utils, non-blocking, just marking for a potential further refactor

t0maboro (approved):

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-disabled-android.e2e.ts
LKuchno: I belive most of then are already addressed in this PR ie. selectOption:
https://github.com/software-mansion/react-native-screens/pull/4441/changes

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-disabled-android.e2e.ts
LKuchno: https://github.com/software-mansion/react-native-screens/pull/4464/changes/828dec3e232cc3b6a17cdf982ec255c6c4304103

PR #4474, opened 2026-08-06
feat(release-tests): prepare setup script for automated app builds

Loloekk:
Closes https://github.com/software-mansion/react-native-screens-labs/issues/643#issuecomment-5425467473

## Description

Adds a `release_tests` setup script that creates a fresh React Native app and builds it against a chosen React Native version and `react-native-screens` version. This makes release verification faster and more repeatable without manually wiring each playground app.

**Usage**

```bash
node scripts/release_tests/create_playground.js
node scripts/release_tests/create_playground.js -h   # full help
```

### Setup options

- **React Native version** (`-r` / `--rn-version`)
- **react-native-screens version** (`-s` / `--screens-version`) — use the current worktree (`current`), or a branch, tag, or commit from the local/remote [react-native-screens](https://github.com/software-mansion/react-native-screens) repository. By default the script take a local ref.
- **From-origin** (`-o` / `--from-origin`) — take RNS from the remote repository
- **Template app** (`-t` / `--template`) — one of the templates:
  - `tabsAndStack` — Stack v5 (RNS 5.x)
  - `tabsAndStack4.x` — legacy Stack v4 + Tabs
  - `empty` — keep the default RN app
  - user could add his own template, and use it to create a new app
- **app name** (`-a` / `--app-name`) sets the name of the created app
- **Help** (`-h` / `--help`) — explains all flags

### Running the app

After setup, the script can build and launch the app with `--run`. Run-related flags are only allowed when `--run` is set.

- **Variant** (`-v`) — `debug` or `release`
- **Platform** (`-p`) — `ios`, `android`, or `both`
- **Device flags** — same idea as yarn/RN CLI (`--ios-simulator`, `--ios-device`, `--ios-udid`, `--android-device`). At most one device can be selected per platform.

### Notes

- RNS depends on the `kconv` library. Older RN versions (e.g. around 0.84) don’t pull it in automatically. Ruby &lt; 3.4 provides it via the stdlib; on Ruby 3.4+ with those RN versions, setup fails without it. The script detects this and adds the `nkf` gem (which includes `kconv`).
- Before running, the script kills any Metro process on port `8081`.
- For Android, start the emulator first (or connect a device). `--android-device` expects an adb id from `adb devices` (e.g. `emulator-5554`), not the AVD name. If the emulator is not running, the build may still succeed but the app will not launch.
- All screens versions other than `current` are cloned into a temporary folder and prepared there.
- From the template app, only `App.tsx` and `src/` (if present) are copied.
- Project init uses:
  `npx @react-native-community/cli@latest init ${appName} --version ${rnVersion} --skip-install`
- Logs from running programs are written to `setup.log`.

### 1. Conflicting / invalid flags

- [x] `node scripts/release_tests/create_playground.js -s current -o`
- [x] `node scripts/release_tests/create_playground.js -v release`
- [x] `node scripts/release_tests/create_playground.js -p ios`
- [x] `node scripts/release_tests/create_playground.js --ios-simulator "iPhone 17"`
- [x] `node scripts/release_tests/create_playground.js --run --ios-simulator "iPhone 17" --ios-device "Karol"`
- [x] `node scripts/release_tests/create_playground.js --run -p ios --android-device "emulator-5554"`
- [x] `node scripts/release_tests/create_playground.js --run -p android --ios-simulator "iPhone 17"`

### 2. Happy paths

- [x] Setup only (current worktree, default template):
  ```bash
  node scripts/release_tests/create_playground.js -a SmokeCurrent
  ```
  Check `setup.log`: init + copy `tabsAndStack` + pack/install current screens; no native build.

- [x] Setup + run both platforms:
  ```bash
  node scripts/release_tests/create_playground.js -r 0.84.0 -s current -t tabsAndStack -a SmokeBoth --run -p both --android-device "emulator-5554" --ios-simulator "iPhone 17"
  ```
  Check log: Metro 8081 freed, Android/iOS launch, Ruby 3.4 `nkf` handling if needed, navigation works.

- [x] Legacy 4.x + from-origin:
  ```bash
  node scripts/release_tests/create_playground.js -r 0.84.0 -s branch:4.26-stable -o -t tabsAndStack4.x -a SmokeLegacy --run -p ios --ios-simulator "iPhone 17"
  ```
  Check log: remote clone (not current pack), app runs.

- [x] Commit + release:
  ```bash
  node scripts/release_tests/create_playground.js -r 0.84.0 -s commit:886754c6e0387c9a486c8c6f601abf7957557763 -o -a SmokeGamma --run -v release -p ios --ios-simulator "iPhone 17" -t tabsAndStack4.x
  ```
  Check log: `--mode Release`, app runs.

- [x] Empty templeate, Android only:
  ```bash
  node scripts/release_tests/create_playground.js -t empty -s tag:4.27.0 -o -a SmokeEmpty --run -p android --android-device "emulator-5554"
  ```
  Check log: no template copy, no iOS steps, Android launches.
  
## Checklist

- [x] Ensured that CI passes


release_tests/create_playground/config.js
```diff
@@ -0,0 +1,309 @@
...
+                                           current                       — current working tree
+                                           branch:<name> | tag:<name>    — clone that branch or tag
+                                           commit:<sha>                  — from local repo; use -f to fetch from remote
+                                           <ref>                         — auto-detect branch, tag, or commit
+                                         Ref versions prefer the local git repository when the ref
+                                         exists there; otherwise they are taken from origin.
+        -f, --force-fetch                Force fetch the screens version from the remote repository (origin).
+                                         Use this to bypass the local git cache (e.g., after force pushes).
+                                         Required to fetch a commit from remote (-s commit:<sha> -f).
+                                         Mutually exclusive with --screens-version 'current'.
```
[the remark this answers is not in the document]
Loloekk: [77cc041](https://github.com/software-mansion/react-native-screens/pull/4474/commits/77cc04111323799269915faaf96800bbc697c556)

release_tests/examples/tabsAndStackExperimental/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Imports
+
+- `Tabs` from `react-native-screens`
+- `Stack`, `SafeAreaView` from `react-native-screens/experimental`
+
+## Setup
+
+```bash
+node setup_app.js -s 4.26-stable -r 0.84.0 -f -e tabsAndStackExperimental -g
```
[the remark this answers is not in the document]
Loloekk: [1cf93f1](https://github.com/software-mansion/react-native-screens/pull/4474/commits/1cf93f1a7edafe821dad992712035e3a9d0c7817)

release_tests/examples/tabsAndStack/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+- `Stack`, `Tabs` from `react-native-screens`
+- `SafeAreaView` from `react-native-screens/experimental`
+
+## Setup
+
+```bash
+node setup_app.js -s current -e tabsAndStack
+# or
+node setup_app.js -s 5.0.0-alpha.1 -r 0.84.0 -e tabsAndStack
```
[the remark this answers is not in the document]
Loloekk: [1cf93f1](https://github.com/software-mansion/react-native-screens/pull/4474/commits/1cf93f1a7edafe821dad992712035e3a9d0c7817)

release_tests/create_playground/tasks/buildApp.js
```diff
@@ -0,0 +1,104 @@
...
+
+function buildIosRunCmd(config) {
+  let cmd = `yarn run ios --mode ${config.capitalizedVariant}`;
+  if (config['ios-udid']) {
+    cmd += ` --udid "${config['ios-udid']}"`;
+  } else if (config['ios-device']) {
+    cmd += ` --device "${config['ios-device']}"`;
+  } else if (config['ios-simulator']) {
+    cmd += ` --simulator "${config['ios-simulator']}"`;
+  }
```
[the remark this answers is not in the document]
Loloekk: Will not fix, it's internal CLI with trusted inputs only

release_tests/create_playground/tasks/buildApp.js
```diff
@@ -0,0 +1,104 @@
...
+  const gemfilePath = path.join(appPath, 'Gemfile');
+  const gemfile = fs.readFileSync(gemfilePath, 'utf8');
+
+  if (!gemfileHasGem(gemfile, 'nkf')) {
+    console.log(
+      `\n⚠️ Gemfile is missing 'nkf' (needed on Ruby 3.4+). Adding it...`,
+    );
+    runCommand('bundle add nkf', appPath, logPath);
+  }
+}
```
t0maboro, line 45: I'm wondering whether we need to rely on the exact ruby version here, testing with
```
bundle exec ruby -e "require 'kconv'"
```
you can detect the failing case and add `"nkf"` gem - marking this version as a workaround ofc, because since RN 0.85 `nkf` is included in the template, so we can easily cleanup this code once we drop support for 84

release_tests/create_playground/tasks/buildApp.js
```diff
@@ -0,0 +1,104 @@
...
+  const { paths } = config;
+
+  runTask('Installing iOS Pods', paths.log, () => {
+    const iosDir = path.join(paths.app, 'ios');
+    const gammaEnv = config.gamma ? 'RNS_GAMMA_ENABLED=1' : '';
+
+    ensureRuby34Gems(paths.app, {
+      runCommand,
+      logPath: paths.log,
+    });
```
t0maboro: Necessary only until we have 0.84 supported, later RN added `nkf` Gem to the template: https://react-native-community.github.io/upgrade-helper/?from=0.84.1&to=0.85.0 - I'd add some checks for 0.84 and add some comment or update the method name describing that it can be cleaned up when we drop support for 84.

release_tests/create_playground/tasks/buildApp.js
```diff
@@ -0,0 +1,104 @@
...
+
+    runCommand(
+      `${gammaEnv} bundle install && ${gammaEnv} bundle exec pod install`,
```
t0maboro: gammaEnv is necessary only for `pod install`; can be dropped from here

release_tests/create_playground/tasks/buildApp.js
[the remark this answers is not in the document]
t0maboro: We don't know how this might evolve or who might use it in the future, so I'm looking at this more from a robustness perspective rather than as a severity issue. Imo worth considering.

release_tests/create_playground/utils.js
```diff
@@ -0,0 +1,106 @@
...
+
+    console.log('--------------------------------------------------');
+    console.log(`💡 Check ${logFile} for the full output.`);
+    process.exit(1);
+  }
+}
+
+function freePort(port = METRO_PORT) {
+  try {
+    execSync(`lsof -tiTCP:${port} -sTCP:LISTEN | xargs kill -9`, {
```
t0maboro: Not only Metro might run on this port, we should check whether we're not terminating any other process.

release_tests/create_playground/tasks/cloneScreens.js
```diff
@@ -0,0 +1,127 @@
+const { execSync } = require('child_process');
+
+function getRemoteUrl(screensPath) {
+  return execSync('git config --get remote.origin.url', {
```
t0maboro: edge-case, but when remote is missing, this command fails, would be nice to add some error handling

release_tests/create_playground/tasks/cloneScreens.js
```diff
@@ -0,0 +1,127 @@
...
+    cwd: screensPath,
+  })
+    .toString()
+    .trim();
+}
+
+function refExistsLocally(screensPath, target) {
+  try {
+    execSync(`git rev-parse --verify "${target}^{commit}"`, {
```
t0maboro: I'd use `execFileSync` instead of `execSync`.

release_tests/create_playground/tasks/cloneScreens.js
```diff
@@ -0,0 +1,127 @@
...
+      cwd: screensPath,
+      stdio: 'ignore',
+    });
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+function isBranchOrTag(screensPath, target) {
```
t0maboro: this function tries to do 2 completely different things, I'd separate branch and tag checks

release_tests/create_playground/tasks/cloneScreens.js
```diff
@@ -0,0 +1,127 @@
...
+      cwd: screensPath,
+      stdio: 'ignore',
+    });
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+function looksLikeCommitHash(ref) {
```
t0maboro: looksLikeCommitHash -> matchesCommitShaFormat

release_tests/create_playground/tasks/cloneScreens.js
```diff
@@ -0,0 +1,127 @@
...
+      `git clone --single-branch --branch "${target}" "${source}" "${tempCloneDir}"`,
+    );
+  }
+
+  function checkoutCommit(source, { fetch = false } = {}) {
+    git(`git clone --no-checkout "${source}" "${tempCloneDir}"`);
+    if (fetch) {
+      git(`git fetch origin "${target}"`, tempCloneDir);
+    }
+    git(`git checkout "${target}"`, tempCloneDir);
```
t0maboro: can we try the checkout first, and fetch as a fallback?

release_tests/create_playground/tasks/copyExample.js
```diff
@@ -0,0 +1,35 @@
+const fs = require('fs');
+const path = require('path');
+
+function copyExample(config, { runTask }) {
+  if (config['example-app'] === 'empty') {
+    console.log(`🔍 No example app selected. Skipping...\n`);
```
t0maboro: Shouldn't we rather load our "default" app if nothing is provided?

release_tests/create_playground/tasks/initApp.js
```diff
@@ -0,0 +1,22 @@
+const fs = require('fs');
+const logger = require('../logger');
+
+function initApp(config, { runTask, runCommand }) {
+  const { paths, appName } = config;
+
+  runTask('Cleaning old app folder', paths.log, () => {
+    fs.mkdirSync(paths.playground, { recursive: true });
```
t0maboro, line 9: creating a directory shouldn't belong to the task that clearly describes its action as performing removal

release_tests/create_playground/tasks/initApp.js
```diff
@@ -0,0 +1,22 @@
...
+    fs.rmSync(paths.app, { recursive: true, force: true });
+    logger.append(paths.log, `Removed directory: ${paths.app}\n`);
+  });
+
+  runTask('Initializing React Native app', paths.log, () => {
+    runCommand(
+      `npx @react-native-community/cli@latest init "${appName}" --version "${config['rn-version']}" --skip-install`,
```
t0maboro: nit: `--skip-git-init` option can be added

t0maboro: rn-version can be extracted from config for a symmetry with appName

release_tests/create_playground/tasks/setupScreens.js
```diff
@@ -0,0 +1,119 @@
...
+  getRemoteUrl,
+  refExistsLocally,
+  cloneScreensRef,
+} = require('./cloneScreens');
+
+function installPackedPackage(packFileName, config, { runTask, runCommand }) {
+  runTask('Installing packed package in app', config.paths.log, () => {
+    fs.writeFileSync(path.join(config.paths.app, 'yarn.lock'), '');
+    runCommand(
+      `yarn add ./${packFileName}`,
```
t0maboro: yarn cache might install the old copy of the package if the package name is constant, have you tested whether installing tarballs with the same name, but different contents behaves properly?

release_tests/create_playground/tasks/setupScreens.js
```diff
@@ -0,0 +1,119 @@
...
+    );
+  });
+}
+
+function buildAndPackScreens(sourceDir, packFile, config, { runCommand }) {
+  runCommand('yarn install', sourceDir, config.paths.log);
+  runCommand('yarn prepare', sourceDir, config.paths.log);
+
+  const output = runCommand('npm pack', sourceDir, config.paths.log, true);
+  const rawPackFile = output.trim().split('\n').pop();
```
t0maboro: `npm pack --json` returns an object from which you might be able to get the tarball name

release_tests/create_playground/tasks/setupScreens.js
```diff
@@ -0,0 +1,119 @@
...
+    () => {
+      console.log(`\n📦 Building package from current working tree...\n`);
+      buildAndPackScreens(config.paths.screens, packFile, config, utils);
+    },
+  );
+
+  installPackedPackage(packFileName, config, utils);
+}
+
+function setupGitScreens(config, utils) {
```
t0maboro: I'd rename:
setupCurrentScreens to setupScreensFromWorkingTree
setupGitScreens to setupScreensFromRef
these seem to be more descriptive for me

release_tests/create_playground/index.js
```diff
@@ -0,0 +1,43 @@
...
+  'ios-simulator': config['ios-simulator'] ?? '',
+  'ios-device': config['ios-device'] ?? '',
+  'ios-udid': config['ios-udid'] ?? '',
+  'android-device': config['android-device'] ?? '',
+});
+console.log('--------------------------------------------------');
+
+logger.initLog(config.paths.log);
+
+console.time('⏳ Total execution time');
```
t0maboro, line 32: not necessary, same for end timestamp below

release_tests/create_playground/config.js
```diff
@@ -0,0 +1,311 @@
...
+  }
+
+  return { type, target };
+}
+
+function getConfig() {
+  const { values: config } = parseArgs({
+    options: {
+      'rn-version': {
+        type: 'string',
```
t0maboro, line 45: I'd consider writing in TS rather than using literals as types, if there are no blockers for that

release_tests/create_playground/config.js
```diff
@@ -0,0 +1,311 @@
...
+                                         exists there; otherwise they are taken from origin.
+        -f, --force-fetch                Force fetch the screens version from the remote repository (origin).
+                                         Use this to bypass the local git cache (e.g., after force pushes).
+                                         Mutually exclusive with --screens-version 'current'.
+        -e, --example-app <app>          Name of the example folder to copy (default: 'tabsAndStack').
+                                         Copies 'App.tsx' from 'examples/<app>'. If a 'src' directory
+                                         exists, it will also be copied. Use 'empty' to skip copying
+                                         and keep the default RN App.tsx.
+                                         Available: 'tabsAndStack' (Stack v5 from main export; RNS 5.x),
+                                         'tabsAndStackExperimental' (Stack v5 from experimental; RNS 4.x + gamma),
```
t0maboro: imo `tabsAndStackExperimental` can be skipped, we should focus on Stack v5 for 5.x releases and Stack v4 for 4.x releases only

release_tests/create_playground/config.js
```diff
@@ -0,0 +1,311 @@
...
+    parseScreensVersion(config['screens-version']);
+
+  if (screensRefType === 'current' && config['force-fetch']) {
+    console.error(
+      `\n❌ FATAL ERROR: Invalid flag combination. You cannot use '--force-fetch' when '--screens-version' is set to 'current'.`,
+    );
+    console.error(
+      `Explanation: The 'current' option uses your current working directory directly. Fetching from origin only applies when you target a specific branch or commit (e.g., -s branch:main -f).\n`,
+    );
+    process.exit(1);
```
t0maboro: shouldn't we use `fatal` for consistency?

release_tests/create_playground/config.js
```diff
@@ -0,0 +1,311 @@
...
+      `Cannot use iOS device flags (${iosTargetFlags.join(
+        ', ',
+      )}) when '--platform' is 'android'.`,
+    );
+  }
+
+  const run = Boolean(config.run);
+
+  if (run && config.gamma && platform === 'android') {
+    fatal(`Cannot use '--gamma' when '--platform' is 'android'.`);
```
t0maboro: Shouldn't make any difference on Android, this check shouldn't be necessary + "both" also covers Android, so this check introduces inconsistency for now.

release_tests/create_playground/tasks/buildApp.js
Loloekk: I add this in [a0d4402](https://github.com/software-mansion/react-native-screens/pull/4474/commits/a0d440224820b2e1ae9ed5750b3fc9b1e3fbdee9)

release_tests/create_playground/tasks/buildApp.js
Loloekk: Instead of checking the RN version, I check whether the Gemfile already includes the nkf gem. I also added comments noting that these helpers can be removed once we drop support for 0.84 -  [59508a6](https://github.com/software-mansion/react-native-screens/pull/4474/commits/59508a60be7ae7914b25a2187f1425e0acd03adc)

release_tests/create_playground/tasks/buildApp.js
Loloekk: Fixed in [e8a76a4](https://github.com/software-mansion/react-native-screens/pull/4474/commits/e8a76a499f4a8333bb160aa1a478c26a9503a9b9)

release_tests/create_playground/config.js
Loloekk: Fixed in [9b96556](https://github.com/software-mansion/react-native-screens/pull/4474/commits/9b96556cdd44eaeb8ce65a7802b04c87419264fc)

release_tests/create_playground/config.js
Loloekk: Ok, so I will drop `tabsAndStackExperimental` and `gamma` flag.
Edit: Dropped in [abf16a6](https://github.com/software-mansion/react-native-screens/pull/4474/commits/abf16a6b57cb7c30e40193c9c03aab5788fbf343)

release_tests/create_playground/config.js
Loloekk: There are not any blockers, I will do it later

release_tests/create_playground/tasks/buildApp.js
Loloekk: We decided to drop gamma flag entirely - [abf16a6](https://github.com/software-mansion/react-native-screens/pull/4474/commits/abf16a6b57cb7c30e40193c9c03aab5788fbf343)

release_tests/create_playground/config.js
Loloekk: We decided to drop gamma flag entirely - [abf16a6](https://github.com/software-mansion/react-native-screens/pull/4474/commits/abf16a6b57cb7c30e40193c9c03aab5788fbf343)

release_tests/create_playground/index.js
Loloekk: I’d keep this — the script can take a while, and the total time is useful when iterating on the setup. Fine with removing the emoji if you want.

release_tests/create_playground/tasks/initApp.js
Loloekk: Fixed in [9b96556](https://github.com/software-mansion/react-native-screens/pull/4474/commits/9b96556cdd44eaeb8ce65a7802b04c87419264fc)

Loloekk: Fixed in [1418c49](https://github.com/software-mansion/react-native-screens/pull/4474/commits/1418c49f5e36709057b232d84607fa6c2a584406)

release_tests/create_playground/tasks/copyExample.js
Loloekk: Yes, we do it, the log is misleading. By default we load our example app. When the user passes `empty` as a `--example-app`, we skip copying and keep the default RN App.tsx. 

I change the log in [a0cf8d9](https://github.com/software-mansion/react-native-screens/pull/4474/commits/a0cf8d9bab7f8a10cf705ec1827e92544253e742).

release_tests/create_playground/tasks/setupScreens.js
Loloekk: Fixed in [a2b1218](https://github.com/software-mansion/react-native-screens/pull/4474/commits/a2b1218432c2047207220b17b2902a22ebe34365)

release_tests/create_playground/tasks/cloneScreens.js
Loloekk: Fixed in [a2b1218](https://github.com/software-mansion/react-native-screens/pull/4474/commits/a2b1218432c2047207220b17b2902a22ebe34365)

release_tests/create_playground/tasks/cloneScreens.js
Loloekk: Fixed in [d7f9cbf](https://github.com/software-mansion/react-native-screens/pull/4474/commits/d7f9cbfbcd2eb723923167254a42905b39d1aa40),
I also rename `cloneBranch` to `cloneBranchOrTag`

release_tests/create_playground/tasks/initApp.js
Loloekk: Fixed in [b2b69a6](https://github.com/software-mansion/react-native-screens/pull/4474/commits/b2b69a6205b3d2c6e87db9d7a6eaa0d6992f94ca)

release_tests/create_playground/tasks/cloneScreens.js
Loloekk: Fixed in [701325b](https://github.com/software-mansion/react-native-screens/pull/4474/commits/701325b67dfc1d1d3859242805a250393d4e23aa)
I also add printing an error message when the task fails in [4ab7813](https://github.com/software-mansion/react-native-screens/pull/4474/commits/4ab78130a75fd91db758380774edd9f4cf57b72d)

release_tests/create_playground/tasks/setupScreens.js
Loloekk: I haven't tested it yet, but Yarn does not cache globally by default ([Yarnpkg - Caching](https://yarnpkg.com/features/caching)). For each screen setup, we create a new blank app with an empty cache. Furthermore, even if we were to use a global cache, Yarn compares cache files by their content checksum, so stale caches shouldn't be a problem anyway ([Yarnpkg - Configuration](https://yarnpkg.com/configuration/yarnrc)).

Loloekk: I have tested installing tarballs with the same name, but different contents behavior and it wasn't cached.

release_tests/create_playground/tasks/cloneScreens.js
Loloekk: Yes, we can. It avoids an unnecessary fetch when the commit is already available after clone. Fixed in [cb968d4](https://github.com/software-mansion/react-native-screens/pull/4474/commits/cb968d48e37d98534581b4ebb67470409aef96ce)

release_tests/create_playground/tasks/buildApp.js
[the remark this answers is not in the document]
Loloekk: Addressed by switching execSync to execFileSync in [daa9602](https://github.com/software-mansion/react-native-screens/pull/4474/commits/daa96028c883cc8bca76aefceaa821147cab8859)

release_tests/create_playground/tasks/cloneScreens.js
Loloekk: Fixed in [daa9602](https://github.com/software-mansion/react-native-screens/pull/4474/commits/daa96028c883cc8bca76aefceaa821147cab8859)

release_tests/create_playground/tasks/cloneScreens.js
```diff
@@ -0,0 +1,127 @@
...
+  logPath,
+  runCommand,
+  forceFetch,
+}) {
+  const git = (cmd, cwd = releaseTestsPath) => runCommand(cmd, cwd, logPath);
+
+  function cloneBranch(source) {
+    git(
+      `git clone --single-branch --branch "${target}" "${source}" "${tempCloneDir}"`,
+    );
```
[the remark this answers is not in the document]
Loloekk: Addressed by switching execSync to execFileSync in [daa9602](https://github.com/software-mansion/react-native-screens/pull/4474/commits/daa96028c883cc8bca76aefceaa821147cab8859)

release_tests/create_playground/utils.js
Loloekk: I've switched to running the command with `execFileSync`, and I now handle cases with multiple processes holding the port as well as a process other than Metro. I try to kill the process with `SIGTERM`, and after 3s I use `SIGKILL`.

release_tests/create_playground/tasks/setupScreens.js
Loloekk: Fixed if [298e395](https://github.com/software-mansion/react-native-screens/pull/4474/commits/298e395ccfc4ce195dcb4e766bafbcb4458efd2f)

release_tests/create_playground/config.js
Loloekk: After thinking about it, I don’t think converting this script to TS is worth it. It’s a small CLI, and adding a compile/runner step would be overhead for something this simple. I'd keep it in JS.

release_tests/create_playground/tasks/buildApp.js
```diff
@@ -0,0 +1,117 @@
+const fs = require('fs');
+const path = require('path');
+
+// remove this helper when we drop support for 0.84.
+function gemfileHasGem(gemfile, gemName) {
+  return new RegExp(`^\\s*gem\\s+['"]${gemName}['"]`, 'm').test(gemfile);
+}
+
+// remove this helper when we drop support for 0.84.
+function canRequireKconv(appPath, { runCommand, logPath }) {
```
t0maboro:
```suggestion
function isKconvAvailable(appPath, { runCommand, logPath }) {
```

release_tests/create_playground/tasks/buildApp.js
```diff
@@ -0,0 +1,117 @@
...
+  }
+}
+
+// remove this helper when we drop support for 0.84.
+function ensureNkfGem(appPath, { runCommand, logPath }) {
+  const gemfilePath = path.join(appPath, 'Gemfile');
+  const gemfile = fs.readFileSync(gemfilePath, 'utf8');
+
+  if (gemfileHasGem(gemfile, 'nkf')) {
+    // if the gemfile has the gem name, return
```
t0maboro: this comment doesn't add any value, should be removed

release_tests/create_playground/tasks/buildApp.js
```diff
@@ -0,0 +1,117 @@
...
+    return;
+  }
+
+  if (canRequireKconv(appPath, { runCommand, logPath })) {
+    // if the gem can require kconv, return
```
t0maboro: this comment doesn't add any value, should be removed

release_tests/create_playground/tasks/buildApp.js
```diff
@@ -0,0 +1,117 @@
...
+    return;
+  }
+
+  console.log(
+    `\n⚠️ Gemfile is missing 'nkf' (required to load kconv). Adding it...`,
+  );
+  runCommand('bundle', ['add', 'nkf'], appPath, logPath);
+  // `bundle add nkf` deafult make bundle install, so we don't need to run it again
```
t0maboro: I'd put it above the line it touches; in this form, it can cause some confusion about whether there was some content below that comment at some point

release_tests/create_playground/tasks/buildApp.js
```diff
@@ -0,0 +1,117 @@
...
+
+// remove this helper when we drop support for 0.84.
+function canRequireKconv(appPath, { runCommand, logPath }) {
+  try {
+    runCommand(
+      'bundle',
+      ['exec', 'ruby', '-e', "require 'kconv'"],
+      appPath,
+      logPath,
+      true,
```
t0maboro: not sure about that, because it will produce an error in the output for which we make recovery internally. putting some log before that's indicating that this command might fail is good enough for me.

release_tests/create_playground/tasks/buildApp.js
t0maboro:
```suggestion
  // `bundle add nkf` by default makes bundle install, so we don't need to run it again
```

release_tests/create_playground/tasks/buildApp.js
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  if (canRequireKconv(appPath, { runCommand, logPath })) {
+    // if the gem can require kconv, return
+    return;
+  }
+
+  console.log(
+    `\n⚠️ Gemfile is missing 'nkf' (required to load kconv). Adding it...`,
+  );
```
t0maboro, line 42: I think that the logging system should be unified; sometimes we're using logger, sometimes console.log, for now we don't have a single source of truth

release_tests/create_playground/tasks/cloneScreens.js
```diff
@@ -0,0 +1,150 @@
...
+      .toString()
+      .trim();
+  } catch {
+    throw new Error("Cannot find git remote 'origin'.");
+  }
+}
+
+function refExistsLocally(screensPath, target) {
+  try {
+    execFileSync('git', ['rev-parse', '--verify', `${target}^{commit}`], {
```
t0maboro: 1. What if branch and tag have the same name? How do we differentiate what we're checking?
2. Do we need to have that method, when `isLocalBranch` and `isLocalTag` are stricter checks?

release_tests/create_playground/tasks/cloneScreens.js
```diff
@@ -0,0 +1,150 @@
...
+        cwd: screensPath,
+        stdio: 'ignore',
+      },
+    );
+    return true;
+  } catch {
+    // not a local tag
+    return false;
+  }
+}
```
t0maboro: the only difference from `isLocalBranch` seems to be the path in show-ref command - I'd extract the logic to some helper `hasLocalRef` and use it for both `isLocalBranch` and `isLocalTag`

release_tests/create_playground/tasks/copyExample.js
```diff
@@ -0,0 +1,35 @@
+const fs = require('fs');
+const path = require('path');
+
+function copyExample(config, { runTask }) {
+  if (config['example-app'] === 'empty') {
```
t0maboro: if 'empty' is a sentinel, I'd extract it to a constant and use the constant everywhere.

release_tests/create_playground/tasks/copyExample.js
```diff
@@ -0,0 +1,35 @@
...
+    console.log(`🔍 Selected empty example app. Skipping...\n`);
```
t0maboro, line 7: same as above, I'd think about unifying the logging system

release_tests/create_playground/tasks/copyExample.js
```diff
@@ -0,0 +1,35 @@
...
+        config['example-app'],
+      );
+
+      const sourceAppFile = path.join(exampleDir, 'App.tsx');
+      const targetAppFile = path.join(config.paths.app, 'App.tsx');
+
+      fs.copyFileSync(sourceAppFile, targetAppFile);
+
+      const sourceSrcDir = path.join(exampleDir, 'src');
+      const targetSrcDir = path.join(config.paths.app, 'src');
```
t0maboro, line 27: Can we put the whole content in `App.tsx`. I don't see any reason for maintaining 2 files.

release_tests/create_playground/tasks/setupScreens.js
```diff
@@ -0,0 +1,121 @@
...
+const logger = require('../logger');
+const {
+  getRemoteUrl,
+  refExistsLocally,
+  cloneScreensRef,
+} = require('./cloneScreens');
+
+function installPackedPackage(packFileName, config, { runTask, runCommand }) {
+  runTask('Installing packed package in app', config.paths.log, () => {
+    fs.writeFileSync(path.join(config.paths.app, 'yarn.lock'), '');
```
t0maboro, line 12: this deserves some comment why are we preparing empty lockfile

release_tests/create_playground/tasks/buildApp.js
t0maboro: typo in 'default'

release_tests/create_playground/utils.js
```diff
@@ -0,0 +1,180 @@
...
+const os = require('os');
+const path = require('path');
+const { execFileSync, execSync } = require('child_process');
+const logger = require('./logger');
+
+const CAPTURE_MAX_BUFFER = 64 * 1024 * 1024;
+const METRO_PORT = 8081;
+
+function formatCommand(file, args = []) {
+  return [file, ...args].join(' ');
```
t0maboro: The formatting should include quoting

```
function formatCommand(file, args = []) {
  return [file, ...args].join(' ');
}

console.log(formatCommand("file", ["--flag", "some value with spaces"]))
```

when you'll have some value with spaces and just copy and paste the output to the console, `some value with spaces` will be treated by the shell as a list of subsequent arguments

release_tests/create_playground/utils.js
```diff
@@ -0,0 +1,180 @@
...
+
+    console.log('--------------------------------------------------');
+    console.log(`💡 Check ${logFile} for the full output.`);
+    process.exit(1);
+  }
+}
+
+function isMetroRespondingOnPort(port) {
+  try {
+    const response = execSync(
```
t0maboro: execFileSync?

release_tests/create_playground/utils.js
```diff
@@ -0,0 +1,180 @@
...
+  try {
+    process.kill(pid, 0); // signal 0 only checks existence
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+function sleepSync(ms) {
+  Atomics.wait(new Int32Array(new SharedArrayBuffer(4)), 0, 0, ms);
```
t0maboro, line 157: these exceptions https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/wait#exceptions aren't problematic in our case, right?

release_tests/create_playground/tasks/buildApp.js
Loloekk: Fixed in [ad0ddf1](https://github.com/software-mansion/react-native-screens/pull/4474/commits/ad0ddf1f5080b61cd46a3a57327489ee11cbbcf6)

release_tests/create_playground/tasks/buildApp.js
Loloekk: Fixed in [ad0ddf1](https://github.com/software-mansion/react-native-screens/pull/4474/commits/ad0ddf1f5080b61cd46a3a57327489ee11cbbcf6)

release_tests/create_playground/tasks/buildApp.js
Loloekk: Fixed in [ad0ddf1](https://github.com/software-mansion/react-native-screens/pull/4474/commits/ad0ddf1f5080b61cd46a3a57327489ee11cbbcf6)

release_tests/create_playground/tasks/buildApp.js
Loloekk: Fixed in [ad0ddf1](https://github.com/software-mansion/react-native-screens/pull/4474/commits/ad0ddf1f5080b61cd46a3a57327489ee11cbbcf6)

Loloekk: Fixed in [ad0ddf1](https://github.com/software-mansion/react-native-screens/pull/4474/commits/ad0ddf1f5080b61cd46a3a57327489ee11cbbcf6)

release_tests/create_playground/tasks/buildApp.js
Loloekk: Fixed in [544ab1a](https://github.com/software-mansion/react-native-screens/pull/4474/commits/544ab1a982826e4544cc71b3294768afd1819700)

release_tests/create_playground/tasks/setupScreens.js
Loloekk: Without a local yarn.lock Yarn would pick up the repo root's yarn.lock and treat the app as part of the screens project. An empty yarn.lock marks the app directory as its own project root, so `yarn add` installs here.

I've added comment in [7c14e0d](https://github.com/software-mansion/react-native-screens/pull/4474/commits/7c14e0df67ef248d3a83f415e8098f02fb951ba0)

release_tests/create_playground/tasks/buildApp.js
Loloekk: I've decided to use logger for writing logs to setup.log. There all logs from running command and the information about moving packages ect. Console.log i'm using to inform user about proces progress.

release_tests/create_playground/tasks/copyExample.js
Loloekk: Same as this [comment](https://github.com/software-mansion/react-native-screens/pull/4474#discussion_r3802699521)

Loloekk:
I decided to force the user to use a reference type every time, instead of the current option. The script also does not try to fetch from origin if the target is absent locally. To fetch the target from origin, --force-fetch is required.

Fetch only if -f is used: [e78b08e](https://github.com/software-mansion/react-native-screens/pull/4474/commits/e78b08ead9915abe5d673cc3af15be08edfc0f32)
Force using the target type: [be5e38c](https://github.com/software-mansion/react-native-screens/pull/4474/commits/be5e38cfc6f1e8f44946cad8a9bd6f904dda2807)
I also refactored the errors: [638691b](https://github.com/software-mansion/react-native-screens/pull/4474/commits/638691b5c3ce28881b7b06eda7867f6b77a3dea5)

release_tests/create_playground/tasks/cloneScreens.js
Loloekk: Fixed by refactor - [comment](https://github.com/software-mansion/react-native-screens/pull/4474#issuecomment-5328202211)

release_tests/create_playground/tasks/cloneScreens.js
Loloekk: Fixed by refactor - [comment](https://github.com/software-mansion/react-native-screens/pull/4474#issuecomment-5328202211)

release_tests/create_playground/tasks/copyExample.js
Loloekk: I decided to also copy the `src` directory to enable the user to create some more complicated examples. I know my examples could be put wholly in `App.tsx`, but someone may want to check the library on something else. This solution enables that.

release_tests/create_playground/tasks/cloneScreens.js
Loloekk: Fixed by refactor - [comment](https://github.com/software-mansion/react-native-screens/pull/4474#issuecomment-5328202211)

release_tests/create_playground/tasks/copyExample.js
Loloekk: Fixed in [21d184f](https://github.com/software-mansion/react-native-screens/pull/4474/commits/21d184f3645d23a4c45d81c68a6783945a46bde1)

I also extract the 'current' screens version to a constant

release_tests/create_playground/utils.js
Loloekk: Fixed in [b55c338](https://github.com/software-mansion/react-native-screens/pull/4474/commits/b55c3385cfbb454f42fec08f45781c1bad074ffc).

release_tests/create_playground/utils.js
Loloekk: Fixed in [2f467d3](https://github.com/software-mansion/react-native-screens/pull/4474/commits/2f467d3a033fe2af40ce0fed9881824db2ecc010)

release_tests/create_playground/utils.js
Loloekk: Yes. It may throw if we use it incorrectly, or if the current thread cannot be blocked. In our case we use it correctly, and the thread our script runs on can always be blocked because it is run by Node.

release_tests/create_playground/tasks/cloneScreens.js
```diff
@@ -0,0 +1,69 @@
+const { execFileSync } = require('child_process');
+const { fatal } = require('../utils');
+
+function getRemoteUrl(screensPath) {
+  try {
+    return execFileSync('git', ['config', '--get', 'remote.origin.url'], {
```
t0maboro: any reason for "bare" `execFileSync` instead of `runCommand`

release_tests/create_playground/tasks/setupScreens.js
```diff
@@ -0,0 +1,116 @@
...
+        const useLocal = !fromOrigin;
+        const remoteUrl = useLocal ? null : getRemoteUrl(screensPath);
+
+        if (useLocal) {
+          console.log(
+            `\n📂 Using version '${target}' (${refType}) from the local repository.`,
+          );
+        } else {
+          console.log(
+            `\n☁️ Fetching version '${target}' (${refType}) from the network (${remoteUrl})...`,
```
t0maboro: This log isn't safe imo, because remote URL can be configured with some credentials: https://stackoverflow.com/questions/18935539/authenticate-with-github-using-a-token

scripts/release_tests/create_playground/config.js
```diff
@@ -0,0 +1,299 @@
...
+        default: 'both',
+      },
+      'app-name': {
+        type: 'string',
+        short: 'a',
+        default: 'PlaygroundApp',
+      },
+      run: {
+        type: 'boolean',
+        default: false,
```
t0maboro, line 102: sth is wrong here; when you test with: `node release_tests/create_playground.js --run=false` , the configuration will interpret `run` arg as `true`

Image

release_tests/examples/tabsAndStack/README.md
```diff
@@ -0,0 +1,23 @@
+# tabsAndStack
+
+Tabs + Stack v5 from the main `react-native-screens` export.
+
+## Requirements
+
+|         |                                                                                                         |
+| ------- | ------------------------------------------------------------------------------------------------------- |
+| **RNS** | 5.x                                                                                                     |
+| **RN**  | Fabric required (RN ≥ 0.76). Match the RNS ↔ RN matrix from the library README for the release you use. |
```
t0maboro: mentioning `(RN ≥ 0.76)` isn't correct, because Fabric became default in 0.76 iirc, but it could be used in older RN versions as well, afair 0.67 already had that impl. I'd just leave "Fabric required"

Loloekk: Fixed in [258e29c](https://github.com/software-mansion/react-native-screens/pull/4474/commits/258e29c69c90c422aa484fff31430b5ed71d395e)

scripts/release_tests/create_playground/config.js
Loloekk: Fixed in [64f8e25](https://github.com/software-mansion/react-native-screens/pull/4474/commits/64f8e251510dbdc708958eb971cc23e59012471e)

release_tests/create_playground/tasks/setupScreens.js
Loloekk: Fixed in [037ce3e](https://github.com/software-mansion/react-native-screens/pull/4474/commits/037ce3ebdb0f5e41c57fdbe9f8691b77e7fafdcd)

scripts/release_tests/create_playground/tasks/initApp.js
```diff
@@ -0,0 +1,35 @@
...
+      'npx',
+      [
+        '@react-native-community/cli@latest',
+        'init',
+        appName,
+        '--version',
+        rnVersion,
+        '--skip-install',
+        '--skip-git-init',
+      ],
```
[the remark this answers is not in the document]
Loloekk: The latest version is better option for the future.

scripts/release_tests/create_playground/templates/tabsAndStack/src/tabsAndStack.tsx
```diff
@@ -0,0 +1,144 @@
...
+    const n = routes.filter(r => r.activityMode === 'attached').length;
+    setRoutes(prev => [
+      ...prev,
+      {
+        key: `details-${Date.now()}`,
+        title: `Details #${n}`,
+        activityMode: 'attached',
+      },
+    ]);
+  }, [routes]);
```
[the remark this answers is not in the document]
Loloekk: The key of this example is to check if tabs and stack do not crash on runtime. It's a tap-driven smoke test, duplicate Date.now() keys won't happen in practice.

release_tests/examples/tabsAndStack/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+- `Stack`, `Tabs` from `react-native-screens`
+- `SafeAreaView` from `react-native-screens/experimental`
+
+## Setup
+
+```bash
+node release_tests/create_playground.js -s current -e tabsAndStack
+# or
+node release_tests/create_playground.js -s tag:5.0.0-alpha.1 -r 0.84.0 -e tabsAndStack
```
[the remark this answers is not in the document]
Loloekk: It isn't matter

release_tests/create_playground/utils.js
```diff
@@ -0,0 +1,197 @@
...
+      .filter(Boolean);
+  } catch (error) {
+    if (error?.code === 'ENOENT') {
+      throw new Error(
+        `Cannot free port ${port}: 'lsof' is not installed (required to detect Metro).`,
+      );
+    }
+    if (typeof error?.status === 'number' && error.status === 1) {
+      return; // nothing listening on the port
+    }
```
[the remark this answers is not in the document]
Loloekk: Fixed in [2f5ca87](https://github.com/software-mansion/react-native-screens/pull/4474/commits/2f5ca87d6b9a2a53a3969a2f85b5ccb0d30466ac).

-Q is not available on all macOS lsof versions. When nothing is listening, lsof exits 1 with empty stderr.

t0maboro (approved):

PR #2035, opened 2024-02-15
fix(iOS): Check for multiple screens while changing screen orientation

uzegonemad:
## Description
This fixes two issues with scenes on iOS:

1. A casting issue, which allows non-window scenes on iOS, e.g. CarPlay's CPTemplateApplicationScene
3. An assumed loading order - the UIWindowScene is not guaranteed to be the first scene (or even present) in a multi-scene app

Fixes #1857

**Edit:** I originally reported that this would only fix the second crash reported in #1857. Upon review of the first crash in that issue, I believe this PR actually addresses both crashes.

## Test code and steps to reproduce
There really isn't any simple code that can be used to test this, as this scenario requires having an app that implements multiple scenes, at least one of which being CarPlay. It would be several hundred lines of code.

## Checklist

- [x] Ensured that CI passes

PR #4486, opened 2026-08-10
chore(iOS): silence release build warnings

### LINKED ISSUE #4319: [Xcode 26.4] iOS Release warnings in react-native-screens 4.26.0
## Description

`react-native-screens` 4.26.0 emits four source warning categories in a clean iOS Release simulator build under Xcode 26.4.1. All are resolved by small source-level changes; no warning suppression is needed.

## Environment

- react-native-screens: 4.26.0
- React Native: 0.86.0
- Expo: 57.0.4
- Xcode: 26.4.1
- SDK: iOS 26.2 simulator
- configuration: Release, Hermes, new architecture

## Reproduction

1. Generate a clean iOS project with the versions above.
2. Build with `xcodebuild -configuration Release -sdk iphonesimulator CODE_SIGNING_ALLOWED=NO build`.
3. Inspect the unfiltered build log.

## Observed warnings and source locations

- `ios/RNSScreen.mm`: deprecated application-global `statusBarFrame` fallback
- `ios/RNSScreenWindowTraits.mm`: deprecated `UIApplication.keyWindow` access despite a scene already being available
- `ios/RNSScreenStackHeaderConfig.mm`: `resizeModeFromCppEquiv` is unused in Release when debug-only callers are absent
- `ios/conversion/RNSConversions-Tabs.mm`: non-exhaustive switch because `RNSTabsScreenSystemItemNone` is asserted before the switch but not handled

## Expected behavior

The Release build should be warning-free. Scene-local status bar/window APIs can replace the deprecated globals, the debug-only helper can be compiled only in debug builds, and the enum switch can remain exhaustive while preserving the assertion.

Loloekk:
## Description

Silence Release build warnings.
Closes https://github.com/software-mansion/react-native-screens/issues/4319

Follow-up:
- Clean up remaining warnings from Stack v5
- Refactor `enforceDesiredDeviceOrientation` in `ios/legacy/RNSScreenWindowTraits.mm`

## Changes

- Make the `RNSTabsScreenSystemItem` switch exhaustive (include `None`) and fix the fallback assert (`RCTAssert(true, …)` → `RCTAssert(false, …)`) so unknown values actually fail in debug (`ios/conversion/RNSConversions-Tabs.mm`)
- Remove unused `getStatusBarHeightIsModal` from `ios/legacy/RNSScreen.mm`.
- Guard `resizeModeFromCppEquiv` with `#ifndef NDEBUG` in `ios/legacy/RNSScreenStackHeaderConfig.mm`
- Replace deprecated `UIApplication.keyWindow` and the manual `connectedScenes` lookup with `RCTKeyWindow()` in `ios/legacy/RNSScreenWindowTraits.mm`. This is safe because `RCTKeyWindow()` filters to `UIWindowScene` and selects the foreground-active scene before returning keyWindow, which should be at least as safe for multi-scene setups as choosing the first UIWindowScene.

## Test plan

- Confirm these 4 warnings no longer appear 
- Run FabricExample app a check `apps/src/tests/single-feature-tests/tabs/test-tabs-system-item-ios`

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/conversion/RNSConversions-Tabs.mm
+++ b/ios/conversion/RNSConversions-Tabs.mm
@@ -323,12 +323,11 @@ RNSTabsScreenSystemItem RNSTabsScreenSystemItemFromReactRNSTabsScreenSystemItem(
   }
 }
 
-UITabBarSystemItem RNSTabsScreenSystemItemToUITabBarSystemItem(RNSTabsScreenSystemItem systemItem)
+std::optional<UITabBarSystemItem> RNSTabsScreenSystemItemToUITabBarSystemItem(RNSTabsScreenSystemItem systemItem)
 {
   switch (systemItem) {
     case RNSTabsScreenSystemItemNone:
-      RCTAssert(false, @"Attempt to convert tabs systemItem none to UITabBarSystemItem");
-      return UITabBarSystemItemSearch;
+      return std::nullopt;
     case RNSTabsScreenSystemItemBookmarks:
       return UITabBarSystemItemBookmarks;
     case RNSTabsScreenSystemItemContacts:
@@ -354,8 +353,7 @@ UITabBarSystemItem RNSTabsScreenSystemItemToUITabBarSystemItem(RNSTabsScreenSyst
     case RNSTabsScreenSystemItemTopRated:
       return UITabBarSystemItemTopRated;
   }
-  RCTAssert(false, @"Attempt to convert unknown tabs screen systemItem to UITabBarSystemItem [%d]", systemItem);
-  return UITabBarSystemItemSearch;
+  return std::nullopt;
 }
 
 #if RNS_TABS_BOTTOM_ACCESSORY_AVAILABLE
--- a/ios/conversion/RNSConversions-Tabs.mm
+++ b/ios/conversion/RNSConversions-Tabs.mm
@@ -354,7 +354,7 @@ UITabBarSystemItem RNSTabsScreenSystemItemToUITabBarSystemItem(RNSTabsScreenSyst
     case RNSTabsScreenSystemItemTopRated:
       return UITabBarSystemItemTopRated;
   }
-  RCTAssert(false, @"Attempt to convert unknown tabs screen systemItem to UITabBarSystemItem [%d]", systemItem);
+  RCTAssert(false, @"Attempt to convert unknown tabs screen systemItem to UITabBarSystemItem [%ld]", (long)systemItem);
   return UITabBarSystemItemSearch;
 }
 
--- a/ios/conversion/RNSConversions-Tabs.mm
+++ b/ios/conversion/RNSConversions-Tabs.mm
@@ -325,9 +325,10 @@ RNSTabsScreenSystemItem RNSTabsScreenSystemItemFromReactRNSTabsScreenSystemItem(
 
 UITabBarSystemItem RNSTabsScreenSystemItemToUITabBarSystemItem(RNSTabsScreenSystemItem systemItem)
 {
-  RCTAssert(systemItem != RNSTabsScreenSystemItemNone,
-            @"Attempt to convert tabs systemItem none to UITabBarSystemItem");
   switch (systemItem) {
+    case RNSTabsScreenSystemItemNone:
+      RCTAssert(false, @"Attempt to convert tabs systemItem none to UITabBarSystemItem");
+      return UITabBarSystemItemSearch;
     case RNSTabsScreenSystemItemBookmarks:
       return UITabBarSystemItemBookmarks;
     case RNSTabsScreenSystemItemContacts:
@@ -353,7 +354,7 @@ UITabBarSystemItem RNSTabsScreenSystemItemToUITabBarSystemItem(RNSTabsScreenSyst
     case RNSTabsScreenSystemItemTopRated:
       return UITabBarSystemItemTopRated;
   }
-  RCTAssert(true, @"Attempt to convert unknown tabs screen systemItem to UITabBarSystemItem [%d]", systemItem);
+  RCTAssert(false, @"Attempt to convert unknown tabs screen systemItem to UITabBarSystemItem [%d]", systemItem);
   return UITabBarSystemItemSearch;
 }
 
--- a/ios/legacy/RNSScreenWindowTraits.mm
+++ b/ios/legacy/RNSScreenWindowTraits.mm
@@ -101,7 +101,8 @@
 {
 #if !TARGET_OS_TV && !TARGET_OS_VISION
   dispatch_async(dispatch_get_main_queue(), ^{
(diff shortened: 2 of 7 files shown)


ios/conversion/RNSConversions-Tabs.mm
```diff
@@ -325,9 +325,10 @@ RNSTabsScreenSystemItem RNSTabsScreenSystemItemFromReactRNSTabsScreenSystemItem(
 
 UITabBarSystemItem RNSTabsScreenSystemItemToUITabBarSystemItem(RNSTabsScreenSystemItem systemItem)
 {
-  RCTAssert(systemItem != RNSTabsScreenSystemItemNone,
-            @"Attempt to convert tabs systemItem none to UITabBarSystemItem");
   switch (systemItem) {
+    case RNSTabsScreenSystemItemNone:
+      RCTAssert(false, @"Attempt to convert tabs systemItem none to UITabBarSystemItem");
+      return UITabBarSystemItemSearch;
```
t0maboro: I'm not convinced that returning an arbitrary value is the best approach here. It relies on `None` only being set programmatically from the internal scope, as well as on the conversion being properly gated at every call site.
I'd suggest using `std::optional` instead and adding explicit error handling at the places where this method is used. That would make the invalid/unavailable state explicit.

ios/legacy/RNSScreenWindowTraits.mm
```diff
@@ -130,17 +130,8 @@ + (void)enforceDesiredDeviceOrientation
...
-        // when an app supports multiple scenes (e.g. CarPlay), it is possible that
-        // UIWindowScene is not the first scene, or it may not be present at all
-        UIWindowScene *scene = nil;
-        for (id connectedScene in array) {
-          if ([connectedScene isKindOfClass:[UIWindowScene class]]) {
-            scene = connectedScene;
-            break;
-          }
-        }
+        UIWindow *window = RCTKeyWindow();
```
t0maboro: `RCTKeyWindow()` is already called above in the same method - we can extract it to variable

ios/legacy/RNSScreenWindowTraits.mm
```diff
@@ -130,17 +130,8 @@ + (void)enforceDesiredDeviceOrientation
...
-
-        // when an app supports multiple scenes (e.g. CarPlay), it is possible that
-        // UIWindowScene is not the first scene, or it may not be present at all
-        UIWindowScene *scene = nil;
-        for (id connectedScene in array) {
-          if ([connectedScene isKindOfClass:[UIWindowScene class]]) {
-            scene = connectedScene;
-            break;
-          }
-        }
```
t0maboro, line 143: Okay, I had some concerns, but looking at https://github.com/software-mansion/react-native-screens/pull/2035/ contents, it seems that this won't cause a regression regarding the CarPlay case, because `RCTKeyWindow` already checks if the scene is `UIWindowScene` https://github.com/react/react-native/blob/v0.87.0-rc.4/packages/react-native/React/Base/RCTUtils.mm#L618-L620 . Based on the #2035 contents, I am not sure how to perform regression testing on that, so I'm just making conclusions based on the implementation.

Anyway, it would be nice to extend the PR description with that information to clarify why we treat this change as safe

t0maboro (commented):
left some comments for further discussion

ios/conversion/RNSConversions-Tabs.mm
Loloekk: Fixed in [1096f9f](https://github.com/software-mansion/react-native-screens/pull/4486/commits/1096f9fd2d38fa5d1ee24d12767b9e78460aed01)

ios/legacy/RNSScreenWindowTraits.mm
Loloekk: Fixed in [c28e1aa](https://github.com/software-mansion/react-native-screens/pull/4486/commits/c28e1aaf5be014290df56177b2bc61663aada852)

ios/legacy/RNSScreenWindowTraits.mm
Loloekk: I've updated the changes section with an explanation of the decision.

ios/tabs/screen/RNSTabsScreenComponentView.mm
```diff
@@ -163,12 +163,18 @@ - (void)createTabBarItem
 {
   UITabBarItem *tabBarItem = nil;
   if (_systemItem != RNSTabsScreenSystemItemNone) {
-    UITabBarSystemItem systemItem = rnscreens::conversion::RNSTabsScreenSystemItemToUITabBarSystemItem(_systemItem);
-    tabBarItem = [[UITabBarItem alloc] initWithTabBarSystemItem:systemItem tag:0];
+    std::optional<UITabBarSystemItem> systemItem =
+        rnscreens::conversion::RNSTabsScreenSystemItemToUITabBarSystemItem(_systemItem);
+    if (!systemItem) {
+      RCTLogError(
```
t0maboro, line 170: `RCTLog.h` should be imported to avoid transitive dependencies

ios/tabs/screen/RNSTabsScreenComponentView.mm
```diff
@@ -163,12 +163,18 @@ - (void)createTabBarItem
 {
   UITabBarItem *tabBarItem = nil;
   if (_systemItem != RNSTabsScreenSystemItemNone) {
-    UITabBarSystemItem systemItem = rnscreens::conversion::RNSTabsScreenSystemItemToUITabBarSystemItem(_systemItem);
-    tabBarItem = [[UITabBarItem alloc] initWithTabBarSystemItem:systemItem tag:0];
+    std::optional<UITabBarSystemItem> systemItem =
```
t0maboro, line 167: `auto` should work

ios/tabs/screen/RNSTabsScreenComponentView.mm
Loloekk: Fixed in [9c30409](https://github.com/software-mansion/react-native-screens/pull/4486/commits/9c304096d9b931893e013d5679740a31ae2dc36b)

Loloekk: I've add it also in `ios/tabs/RNSTabBarAppearanceCoordinator.mm`.

ios/tabs/screen/RNSTabsScreenComponentView.mm
Loloekk: I think it's better to keep std::optional<UITabBarSystemItem> to clarify the code. If you feel strongly that the change is necessary, I'll do it.

t0maboro: the current shape is ok

t0maboro (approved):
looks good, thanks!

PR #4493, opened 2026-08-12
chore(test): e2e test for test-stack-lift-on-scroll-android

LKuchno:
## Description

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1733

Adds Detox e2e coverage for the Android **lift-on-scroll** header behaviour, covering the assertable half of the existing `test-stack-lift-on-scroll-android` scenario.

The app bar's lifted state is read through its `elevation`, which Material's app bar state list animator drives off the same `state_lifted` that paints the tonal `liftOnScrollColor`. That makes the *transition* assertable even though the colour itself is not: the test asserts **that** the app bar lifts and unlifts, never how it looks.

## Changes

- New e2e spec `FabricExample/e2e/single-feature-tests/stack-v5/test-stack-lift-on-scroll-android.e2e.ts`, covering:
  - app bar attach / detach as `hidden` and `headerConfig enabled` are toggled,
  - the `small` header staying pinned while content scrolls,
  - a transparent header overlaying the content,
  - the lifted state following scroll position for `liftOnScroll` = `true` / `false` / `undefined`.
- `FabricExample/e2e/native-class-names.ts`: added `CLASS_NAME_ANDROID_APP_BAR_LAYOUT`. Detox resolves `by.type` with `isAssignableFrom`, so it also matches the library's internal `StackHeaderAppBarLayout` subclasses.
- Test screen (`apps/src/tests/single-feature-tests/stack-v5/test-stack-lift-on-scroll-android/index.tsx`):
  - added `testID`s to the scroll view and the config controls,
  - added a bottom sentinel so the test can assert the content actually scrolled instead of inferring it from the scroll action not throwing,
  - added top padding when `transparent` is on, so the overlaying header no longer buries the controls that toggle it back off,
  - exported `TriState` for reuse in the spec.
- `scenario.md`: replaced the "Incomplete - we can't detect the elevation/color flash with Detox" note with the actual coverage split, and listed what stays manual-only.

### DIFF
--- a/FabricExample/e2e/single-feature-tests/stack-v5/test-stack-lift-on-scroll-android.e2e.ts
+++ b/FabricExample/e2e/single-feature-tests/stack-v5/test-stack-lift-on-scroll-android.e2e.ts
@@ -260,6 +260,8 @@ async function expectLiftFollowsScroll() {
   await expectAppBarFlat();
   await scrollAwayFromTop();
   await expectAppBarLifted();
+  await scrollUntilVisible(BOTTOM_MARKER, SCROLL_VIEW);
+  await expectAppBarLifted();
   // Has to be the edge jump: unlifting needs offset zero exactly. That makes
   // this usable only with an opaque header — every caller has one.
   await scrollToTopEdge();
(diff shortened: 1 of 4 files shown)


apps/src/tests/single-feature-tests/stack-v5/test-stack-lift-on-scroll-android/scenario.md
```diff
@@ -44,7 +57,7 @@ Incomplete - we can't detect the elevation/color flash with Detox.
 
 2. Set `liftOnScroll` = `false`, then scroll.
 
-- [ ] The header does not lift — it stays flat regardless of scroll position.
+- [ ] The header does not lift - it stays flat regardless of scroll position.
```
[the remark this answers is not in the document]
LKuchno: I think this change is unnecessary, current step description is sufficient.

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-lift-on-scroll-android.e2e.ts
```diff
@@ -0,0 +1,402 @@
...
+}
+
+/**
+ * Asserts the lifted state is scroll-independent. Note this cannot assert a
+ * *value*: an app bar that is not liftable at all rests at the animator's
+ * "enabled" elevation, not at zero — only its invariance under scrolling is the
+ * claim being made.
+ */
+async function expectLiftUnaffectedByScroll() {
+  const elevationAtTop = await appBarElevation();
```
t0maboro, line 276: Q: I just want to understand the timing here. We have `waitForAppBarElevation`, and I'm not sure whether we should consider using it in this case.

t0maboro (approved):

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-lift-on-scroll-android.e2e.ts
LKuchno: I don't think we should. waitForAppBarElevation returns as soon as its predicate passes, so testing a negative assertion (elevation === before) would pass immediately on the first sample without validating later frames.

We don't need polling here as lifts trigger on the first offset change and last ~150 ms. By the time the scroll helper's own waitFor(...).not.toBeVisible() settles, the animation is already done.

t0maboro: ok

PR #4497, opened 2026-08-13
fix(Android, Tabs): expose tab item testID as viewIdResourceName

### LINKED ISSUE #4496: [Android] Bottom tab item testID is exposed only as a View tag, so UiAutomator-based drivers (Maestro, Appium) cannot match it
## Description

On Android, a bottom tab item's `testID` is written to the view's `tag` property only. UiAutomator-based drivers read `AccessibilityNodeInfo.getViewIdResourceName()` (surfaced as `resource-id`), never the view tag, so those drivers cannot select tab items by id at all. The selector silently matches nothing rather than failing loudly, which makes it easy to misdiagnose as a flaky tap.

`TabsHostA11yCoordinator.setA11yPropertiesToTabItem` sets:

```kotlin
// when matching view by id, espresso driver seems to look for tag property
findTabItemView(menuItem.itemId)?.tag = tabsScreen.tabBarItemTestID
```

The existing comment names the target: Espresso, where it does work. But Maestro and Appium drive through UiAutomator, so for them the id is invisible.

This is inconsistent with how React Native exposes `testID` for every ordinary view. `ReactAccessibilityDelegate` does `info.viewIdResourceName = testId`, which is why `id:` selectors work everywhere else in an RN app on Android but stop working on native tab items specifically.

Introduced with the feature in #3497, which notes Android "is more complicated when it comes to the item" — this is that complication surfacing for non-Espresso drivers. Worth noting the intent: the JS-side counterpart, [expo/expo#47472](https://github.com/expo/expo/pull/47472), added `testID` to `NativeTabs.Trigger` explicitly "so native tabs can be matched in end-to-end tests". So this is not a feature request — it is that promised behaviour not holding on Android for any driver other than Espresso.

## Steps to reproduce

1. `npx create-expo-app@latest rnstabs && cd rnstabs`
2. Add a `NativeTabs` layout with an explicit `testID` on a trigger:

```tsx
// app/(tabs)/_layout.tsx
import { NativeTabs } from 'expo-router/unstable-native-tabs';

export default function Layout() {
  return (
    <NativeTabs>
      <NativeTabs.Trigger name="index" testID="tab-home">
        <NativeTabs.Trigger.Label>Home</NativeTabs.Trigger.Label>
        <NativeTabs.Trigger.Icon md="home" />
      </NativeTabs.Trigger>
      <NativeTabs.Trigger name="search" testID="tab-search">
        <NativeTabs.Trigger.Label>Search</NativeTabs.Trigger.Label>
        <NativeTabs.Trigger.Icon md="search" />
      </NativeTabs.Trigger>
    </NativeTabs>
  );
}
```

3. Build and run a debug build on an Android emulator (`npx expo run:android`).
4. Dump the accessibility tree, no test framework required:

```
adb shell uiautomator dump /sdcard/ui.xml && adb shell cat /sdcard/ui.xml | tr '>' '>\n' | grep navigation_bar_item
```

**Expected:** the tab item nodes carry `resource-id="tab-home"` / `resource-id="tab-search"`.
**Actual:** the tab item nodes carry no `resource-id` at all. Equivalently, a Maestro step `- tapOn: { id: "tab-search" }` matches nothing, while `- tapOn: { text: "Search" }` works.

For contrast, any ordinary RN `<View testID="foo">` in the same app does show `resource-id="foo"`.

## Snack or a link to a repository

Not reproducible in a Snack: `unstable-native-tabs` needs a custom native build, and the symptom is only observable through the Android accessibility tree (`uiautomator dump`), which Snack cannot expose. The steps above build a minimal app from `create-expo-app` in a couple of commands and verify with a single `adb` command, no test framework involved. Happy to push a repo if that is still preferred.

## Screens version

4.26.2 (tested). Still present in **4.27.0** and on **`main`** by source inspection: both keep the `?.tag = tabsScreen.tabBarItemTestID` assignment with no `viewIdResourceName`.

## React Native version

0.86.2

## Platforms

Android (iOS is unaffected, `testID` maps to `accessibilityIdentifier` there and matches correctly)

## Environment

- expo-router 57.0.11 (`unstable-native-tabs`)
- Maestro 2.6.0
- Android API 34, Pixel 8 emulator

## Suggested fix

Keep the tag assignment for Espresso, and additionally mirror the id onto the accessibility node:

```kotlin
findTabItemView(menuItem.itemId)?.let { menuView ->
    menuView.tag = tabsScreen.tabBarItemTestID

    ViewCompat.setAccessibilityDelegate(
        menuView,
        object : AccessibilityDelegateCompat() {
            override fun onInitializeAccessibilityNodeInfo(host: View, info: AccessibilityNodeInfoCompat) {
                super.onInitializeAccessibilityNodeInfo(host, info)
                info.viewIdResourceName = tabsScreen.tabBarItemTestID
            }
        },
    )
}
```

Calling `super` first preserves Material's own node population, so the `contentDescription` and the "tab, N of M" announcement are unaffected.

Verified as a local patch on 4.26.2 against a Pixel 8 (API 34). After the change the hierarchy shows `resource-id=tab-search` with `content-desc` still holding the localized label and `selected` still tracking, and a flow tapping four tabs by id passes. Without it, the same flow matches nothing.

Happy to open a PR with this if the approach looks right.

## Why it matters

Any Maestro or Appium suite migrating from a JS tab bar to `NativeTabs` loses id-based selection of its primary navigation. Both workarounds are poor: matching on the visible label couples tests to copy and to the active locale, and setting `accessibilityLabel` to the test id replaces the Android `contentDescription`, so TalkBack announces the test id on the app's primary navigation.

mackbrowne:
## Description

Closes #4496.

A bottom tab item's `testID` is written to the view's `tag` property only:

```kotlin
// when matching view by id, espresso driver seems to look for tag property
findTabItemView(menuItem.itemId)?.tag = tabsScreen.tabBarItemTestID
```

As the comment says, that works for Espresso. But UiAutomator-based drivers — Maestro and Appium — match on `AccessibilityNodeInfo.getViewIdResourceName()` (exposed as `resource-id`) and never read the view tag. For them the id is invisible, and a selector silently matches nothing rather than failing loudly, which reads as a flaky tap.

This is inconsistent with the rest of React Native. `ReactAccessibilityDelegate` does `info.viewIdResourceName = testId` for every ordinary view, which is why `id` selectors work everywhere else in an RN app on Android but stop working on native tab items specifically.

Worth noting the intent behind the prop: the JS-side counterpart, expo/expo#47472, added `testID` to `NativeTabs.Trigger` explicitly "so native tabs can be matched in end-to-end tests". On Android that currently only holds for Espresso.

## Changes

- `TabsHostA11yCoordinator.setA11yPropertiesToTabItem` now also sets `viewIdResourceName` on the tab item's accessibility node, via a `AccessibilityDelegateCompat`.
- The existing `tag` assignment is kept, so Espresso is unaffected.
- `super.onInitializeAccessibilityNodeInfo` is called first, so Material's own node population — including the `contentDescription` and the "tab, N of M" announcement — is preserved.
- The delegate reads `tabsScreen.tabBarItemTestID` at node-init time rather than capturing it, so a prop update is reflected without reinstalling the delegate.

No public API change, and no visual change, so the visual documentation section is omitted.

## Test plan

Verified on a Pixel 8 emulator (API 34) against `react-native-screens` 4.26.2 applied as a patch, with `expo-router` 57.0.11 `unstable-native-tabs` and Maestro 2.6.0. The same change against `main`'s shape is what this PR contains; `ktlint 1.3.0` (matching `android/spotless.gradle`) reports clean on the modified file.

Minimal example:

```tsx
import { NativeTabs } from 'expo-router/unstable-native-tabs';

export default function Layout() {
  return (
    <NativeTabs>
      <NativeTabs.Trigger name="index" testID="tab-home">
        <NativeTabs.Trigger.Label>Home</NativeTabs.Trigger.Label>
        <NativeTabs.Trigger.Icon md="home" />
      </NativeTabs.Trigger>
      <NativeTabs.Trigger name="search" testID="tab-search">
        <NativeTabs.Trigger.Label>Search</NativeTabs.Trigger.Label>
        <NativeTabs.Trigger.Icon md="search" />
      </NativeTabs.Trigger>
    </NativeTabs>
  );
}
```

Inspect the accessibility tree, no test framework required:

```
adb shell uiautomator dump /sdcard/ui.xml && adb shell cat /sdcard/ui.xml | tr '>' '>\n' | grep navigation_bar_item
```

| | tab item nodes |
| --- | --- |
| before | no `resource-id` |
| after | `resource-id="tab-home"` / `resource-id="tab-search"` |

In both cases `content-desc` still holds the label and `selected` still tracks the active tab, so the screen-reader behaviour is unchanged. Equivalently, a Maestro step `- tapOn: { id: "tab-search" }` matches nothing before the change and taps the tab after it; in a real suite all four tabs of an app shell flow now tap by id.

I have not been able to run the repo's own Android e2e suite locally, so I would appreciate CI confirming that. Happy to add a `Test*.tsx` example under `apps/src/tests` if you would like one for this.

## Checklist

- [ ] For visual changes, included screenshots / GIFs / recordings documenting the change. *(no visual change)*
- [x] For API changes, updated relevant public types. *(no API change)*
- [ ] Ensured that CI passes *(cannot run the Android e2e suite locally; relying on CI)*

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/tabs/host/TabsHostA11yCoordinator.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/tabs/host/TabsHostA11yCoordinator.kt
@@ -3,6 +3,9 @@ package com.swmansion.rnscreens.tabs.host
 import android.os.Build
 import android.view.MenuItem
 import android.view.View
+import androidx.core.view.AccessibilityDelegateCompat
+import androidx.core.view.ViewCompat
+import androidx.core.view.accessibility.AccessibilityNodeInfoCompat
 import androidx.core.view.children
 import androidx.core.view.get
 import com.google.android.material.bottomnavigation.BottomNavigationView
@@ -23,8 +26,26 @@ class TabsHostA11yCoordinator(
             menuItem.contentDescription = tabsScreen.tabBarItemAccessibilityLabel
         }
 
-        // when matching view by id, espresso driver seems to look for tag property
-        findTabItemView(menuItem.itemId)?.tag = tabsScreen.tabBarItemTestID
+        findTabItemView(menuItem.itemId)?.let { tabItemView ->
+            // when matching view by id, espresso driver seems to look for tag property
+            tabItemView.tag = tabsScreen.tabBarItemTestID
+
+            // UiAutomator-based drivers (Maestro, Appium) read the node's viewIdResourceName instead,
+            // so expose the testID there as well — this is what React Native's ReactAccessibilityDelegate
+            // does for every other view. Read through to the screen so a prop update is reflected.
+            ViewCompat.setAccessibilityDelegate(
+                tabItemView,
+                object : AccessibilityDelegateCompat() {
+                    override fun onInitializeAccessibilityNodeInfo(
+                        host: View,
+                        info: AccessibilityNodeInfoCompat,
+                    ) {
+                        super.onInitializeAccessibilityNodeInfo(host, info)
+                        info.viewIdResourceName = tabsScreen.tabBarItemTestID
+                    }
+                },
+            )
+        }
     }
 
     // Third-party SDKs can freely assign their own ids to id-less views in CustomBottomNavigationView's subtree and there is no way to


kkafar (commented):
cc @kmichalikk 

I recall you've implemented the a11y for Tabs. Could you review this PR please?

kmichalikk (approved):
Looks good, verified with maestro

Thanks!

kmichalikk:
@coderabbitai full review

Issue #3559, opened 2026-01-21
Bug: formSheet corners show shimmer when app theme differs from system theme


marcospgp:
### Description

When an app uses a different theme than the system (e.g., app forces dark mode while phone is in 
  light mode), formSheet presentations show a visible shimmer/flash at the rounded corners during  
  the present/dismiss animation.                                                                   
                                                                                                   
  ### Steps to reproduce                                                                           
                                                                                                   
  1. Set phone to light mode                                                                       
  2. App forces dark mode (via in-app theme preference, not `userInterfaceStyle` in app.json)      
  3. Present a screen with `presentation: "formSheet"`                                             
  4. Observe white shimmer at rounded corners during animation                                     
                                                                                                   
  ### Expected behavior                                                                            
                                                                                                   
  The formSheet container should respect the app's theme, not the system theme.                    
                                                                                                   
  ### Actual behavior                                                                              
                                                                                                   
  The formSheet container uses the system's interface style for its background. When system is     
  light mode but app content is dark, the light container background shows through at the rounded  
  corners during animation.                                                                        
                                                                                                   
  ### Root cause                                                                                   
                                                                                                   
  `UISheetPresentationController` uses the system's interface style. The presented view            
  controller's `overrideUserInterfaceStyle` is not being set to match the app's theme.             
                                                                                                   
  ### Workarounds attempted (none work)                                                            
                                                                                                   
  - `userInterfaceStyle: "dark"` in app.json - forces entire app to dark, prevents dynamic themes  
  - `contentStyle: { backgroundColor }` - doesn't affect the sheet container                       
  - `card` color in React Navigation theme - doesn't affect formSheet container                    
                                                                                                   
  ### Proposed fix                                                                                 
                                                                                                   
  Set `overrideUserInterfaceStyle` on the presented view controller (`_controller`) based on a new 
  `userInterfaceStyle` prop on Screen.                                                             
                                                                                                   
  ### Environment                                                                                  
                                                                                                   
  - react-native-screens: 4.16.0                                                                   
  - react-native: 0.76.x                                                                           
  - expo: 53                                                                                       
  - Platform: iOS 17+    

### Steps to reproduce

see above

### Snack or a link to a repository

see above

### Screens version

see above

### React Native version

see above

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

willashley23:
[...]

PR #4499, opened 2026-08-13
Fix light border around iOS 26 modal corners

regnerus:
## Summary

- Dark native modals could show a thin light border around their bottom corners on iOS 26.
- This change removes that border without changing the modal presentation or swipe-to-dismiss behavior.

## Description

While testing a dark modal on an iPhone running iOS 26 or newer, I noticed a thin light line following the rounded bottom corners. The line remains visible after the presentation animation has finished.

The bottom line is caused by redundant software clipping where the modal already follows the physical display corners. A top line can also appear because UIKit's native backing view can retain the system background color, which contrasts with the modal content along the antialiased edge.

Changing the React Native view background, `contentStyle`, and navigation theme did not help. The line comes from UIKit's modal container instead.

On iOS 26, UIKit gives large modals an opaque system background. It also rounds the bottom of the modal in software, even when the modal already reaches the bottom and sides of the physical display. The antialiased edge of that second rounding exposes a small amount of the light system background.

This change detects that specific layout, matches the native backing view to the modal content, and removes only the redundant bottom corner radii. The native top corners are kept as they are. Smaller or differently positioned modals continue to use UIKit's original corner configuration.

The behavior is limited to iPhones running iOS 26 or newer.

Apple has confirmed that large sheets becoming opaque is intentional on iOS 26:

https://developer.apple.com/forums/thread/795880

Possibly related to #3559, although that issue describes a temporary shimmer during `formSheet` animations. This issue is a permanent line on a stationary `modal`.

### Before
before

 ### After
after

## Test plan

- [x] Present a dark native modal on an iPhone running iOS 26 or newer — verify the light line is gone.
- [x] Verify the top modal corners still use UIKit's original radius.
- [x] Swipe down on the modal — verify interactive dismissal still works.
- [x] Build the FabricExample with Xcode 27 and the iOS Simulator SDK 27.

## Checklist
- [x] Ensured that CI passes

### DIFF
--- a/ios/legacy/RNSScreen.mm
+++ b/ios/legacy/RNSScreen.mm
@@ -1561,6 +1561,42 @@ Class<RCTComponentViewProtocol> RNSScreenCls(void)
   if (isDisplayedWithinUINavController || isTabScreen || self.screenView.isPresentedAsNativeModal) {
     [self.screenView updateBounds];
   }
+
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
+  if (@available(iOS 26.0, *)) {
+    if (self.screenView.isPresentedAsNativeModal &&
+        self.traitCollection.userInterfaceIdiom == UIUserInterfaceIdiomPhone) {
+      UIView *clippingView = self.view.superview;
+      UIView *shadowView = clippingView.superview;
+      UIView *transitionView = shadowView.superview;
+      if (transitionView != nil) {
+        CGRect shadowFrame = [transitionView convertRect:shadowView.bounds fromView:shadowView];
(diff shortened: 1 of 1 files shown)


kkafar (changes requested):
Hey @regnerus,

the PR introduces obvious regression:

https://github.com/user-attachments/assets/c98a7a49-58d5-4623-af9f-83c93c6b54bb

which makes it unsuitable for proceeding.

Additionally, I do not observe any clipping when testing. Sheet with detent 1.0 or a small range of nearby values fills the screen fully, unless you use detent like 0.9895 or something - but that's not clipping - that's background underneath the modal. 

We need a reproducer here and obviously prevent regressions.

PR #4502, opened 2026-08-14
fix(android): respect gestureEnabled for form sheets

omer-ayhan:
## Summary

Android `formSheet` presentations currently remain draggable even when `gestureEnabled: false` is set.

This change makes the Material bottom-sheet behavior use the existing `screen.isGestureEnabled` value in both Android form-sheet implementations:

- `SheetDelegate`
- `ScreenModalFragment`

With this change, `gestureEnabled: false` prevents swipe-to-dismiss while preserving the existing default behavior when the option is enabled or omitted.

## Root cause

The Android form-sheet setup hardcodes `isDraggable = true`, even though the screen's `gestureEnabled` prop is already propagated to native code.

## Validation

- `./android/gradlew -p android spotlessCheck -q`
- `git diff --check`
- The same patch compiled successfully in an Expo Android host app.
- A standalone `assembleDebug` from this clean checkout was attempted but was blocked by the environment's React Native Maven metadata lookup (`SAXNotRecognizedException`), before compiling the changed source.

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/legacy/ScreenModalFragment.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/legacy/ScreenModalFragment.kt
@@ -234,7 +234,7 @@ class ScreenModalFragment :
 
         behavior.apply {
             isHideable = true
-            isDraggable = true
+            isDraggable = screen.isGestureEnabled
         }
 
         when (screen.sheetDetents.count) {
--- a/android/src/main/java/com/swmansion/rnscreens/legacy/bottomsheet/SheetDelegate.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/legacy/bottomsheet/SheetDelegate.kt
@@ -188,7 +188,7 @@ class SheetDelegate(
 
         behavior.apply {
             isHideable = true
-            isDraggable = true
+            isDraggable = screen.isGestureEnabled
         }
 
         // There is a guard internally that does not allow the callback to be duplicated.


t0maboro:
In the downstream implementation of React Navigation, `gestureEnabled` is always set to `false` on Android, so passing the prop from it won't have any effect for now - just marking it as sth that we need to be aware of: https://github.com/react-navigation/react-navigation/blob/7.x/packages/native-stack/src/views/NativeStackView.native.tsx#L371-L377 . Passing this value through would require additional coordination with the native stack implementation in react-navigation.
Since this has been the case for several years, I'd prefer to avoid changing that behavior and instead look for a different solution (e.g., exposing a dedicated prop for that). Additionally, we'd need to consider other presentation types as well. I'm not sure it would be good to mark this API available on Android when it won't have any effect for the basic Screen presentation.

omer-ayhan:
> In the downstream implementation of React Navigation, `gestureEnabled` is always set to `false` on Android, so passing the prop from it won't have any effect for now - just marking it as sth that we need to be aware of: https://github.com/react-navigation/react-navigation/blob/7.x/packages/native-stack/src/views/NativeStackView.native.tsx#L371-L377 . Passing this value through would require additional coordination with the native stack implementation in react-navigation. Since this has been the case for several years, I'd prefer to avoid changing that behavior and instead look for a different solution (e.g., exposing a dedicated prop for that). Additionally, we'd need to consider other presentation types as well. I'm not sure it would be good to mark this API available on Android when it won't have any effect for the basic Screen presentation.

Thanks for the clarification @t0maboro . That makes sense.

After checking the referenced codebase I agree that changing the existing gestureEnabled behavior could be misleading on Android, especially since React Navigation currently sets it to false there. I'll close this PR and instead investigate a dedicated sheet-specific prop, coordinated between react-native-screens and React Navigation, with the other sheet presentation types considered as well.

PR #4507, opened 2026-08-17
fix(Android, Stack v5): change order of toolbar menu item prop application to ensure correct overflow behavior

kligarski:
## Description

Moves applying `showAsAction` prop **AFTER** setting an icon to ensure that correct overflow behavior is applied. Prior to this PR, in some configurations the icon would not appear in the toolbar menu. Instead, overflow icon would be visible but it wouldn't open the overflow menu, making the menu item not available.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1714.

## Changes

- move applying `showAsAction` to the bottom of `StackHeaderToolbarMenuApplicator.applyMenuElementOptions`

| Before           | After            |
| ---------------- | ---------------- |
| <video src="https://github.com/user-attachments/assets/c18f3ba9-ba83-44f2-acdb-de0b4843cdfc" /> | <video src="https://github.com/user-attachments/assets/de62c583-9652-4220-be29-200562f2a9a0" /> |

## Test plan

Run `test-stack-toolbar-menu-show-as-action` SFT.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/stack/header/toolbar/StackHeaderToolbarMenuApplicator.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/stack/header/toolbar/StackHeaderToolbarMenuApplicator.kt
@@ -291,7 +291,6 @@ internal object StackHeaderToolbarMenuApplicator {
         }
         options.hidden?.let { menuItem.isVisible = !it }
         options.disabled?.let { menuItem.isEnabled = !it }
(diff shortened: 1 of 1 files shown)


LKuchno (approved):
Tested. Behavior is as described in scenario, problem from reported issue is not present anymore.
Thank you! ⭐️

t0maboro (approved):

PR #4517, opened 2026-08-18
fix(iOS, Stack v5): Opt out of largeTitleDisplayMode by default when headerConfig is not present

kmichalikk:
Closes https://github.com/software-mansion/react-native-screens-labs/issues/1699

## Description

This PR fixes the issue with wrong header mode (large instead of regular) when headerConfig is not present in the screen.

## Changes

- opted out of large header when initializing the current screen, before headerConfig is attached

| Before | After |
| --- | --- |
| image | image |

## Test plan

You can use `test-stack-prevent-native-dismiss-nested-stack` which already has a nested stack inside. Observe that it works correctly when headerConfig is commented out.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/stack/screen/RNSStackScreenController.mm
+++ b/ios/stack/screen/RNSStackScreenController.mm
@@ -19,6 +19,13 @@
     _screenView = componentView;
     _headerCoordinator = [[RNSStackScreenHeaderCoordinator alloc] initWithScreenController:self];
     _containerItemSupport = [RNSContainerItemSupport new];
+
+#if !TARGET_OS_TV
+    // NavigationBar is configured to always prefer large title and the actual configuration is
+    // handled by headerConfig updating the display mode on navigationItem.
+    // In case the headerConfig is not provided, we opt out of large title by default.
+    self.navigationItem.largeTitleDisplayMode = UINavigationItemLargeTitleDisplayModeNever;
+#endif // !TARGET_OS_TV
   }
   return self;
 }


ios/stack/screen/RNSStackScreenController.mm
```diff
@@ -19,6 +19,13 @@ - (instancetype)initWithComponentView:(RNSStackScreenComponentView *)componentVi
     _screenView = componentView;
     _headerCoordinator = [[RNSStackScreenHeaderCoordinator alloc] initWithScreenController:self];
     _containerItemSupport = [RNSContainerItemSupport new];
+
+#if !TARGET_OS_TV
+    // NavigationBar is configured to always prefer large title and the actual configuration is
+    // handled by headerConfig updating the display mode on navigationItem.
+    // In case the headerConfig is not provided, we opt out of large title by default.
+    self.navigationItem.largeTitleDisplayMode = UINavigationItemLargeTitleDisplayModeNever;
```
t0maboro, line 27: just for catching context, was there any reason why we decided to hardcode `prefersLargeTitles` to `YES` rather than parametrizing it by e.g., `largeTitleEnabled` or any other prop as we did in v4?

t0maboro (approved):

kligarski (commented):
If there is no header config, I think we should hide the navigation bar completely. If so, is this change necessary?

ios/stack/screen/RNSStackScreenController.mm
kmichalikk: I somewhat didn't like the idea of setting both each time. I researched what is the correct way of setting those but I don't remember the conclusions. `prefersLargeTitles` is a global switch and `displayMode` is per screen, so I thought it would be more "defensive" to not add another runtime changes and only use one toggle.

kmichalikk:
@kligarski fair, I'm going to do it in another PR and close this.

Issue #3096, opened 2025-08-04
Exception java.lang.IndexOutOfBoundsException: getChildDrawingOrder() returned invalid index 3 (child count is 3)


sunny635533:
### Description

I found that a large number of users reported this error from the google play backend of the online log. At present to track ". Com. Swmansion rnscreens. ScreenStack. PerformDraw (ScreenStack. Kt: 334) "here, but I am unable to reproduce the problem. May I ask if there are any others who have encountered similar problems?

### Steps to reproduce

I am unable to reproduce the problem.Only these Log:
```
Exception java.lang.IndexOutOfBoundsException: getChildDrawingOrder() returned invalid index 3 (child count is 3)
  at android.view.ViewGroup.getAndVerifyPreorderedIndex (ViewGroup.java:2239)
  at android.view.ViewGroup.dispatchDraw (ViewGroup.java:4522)
  at android.view.View.draw (View.java:23881)
  at android.view.View.updateDisplayListIfDirty (View.java:22729)
  at android.view.View.draw (View.java:23604)
  at android.view.ViewGroup.drawChild (ViewGroup.java:4766)
  at com.facebook.react.views.view.ReactViewGroup.drawChild (ReactViewGroup.java:972)
  at android.view.ViewGroup.dispatchDraw (ViewGroup.java:4549)
  at com.facebook.react.views.view.ReactViewGroup.dispatchDraw (ReactViewGroup.java:943)
  at android.view.View.draw (View.java:23881)
  at com.facebook.react.views.view.ReactViewGroup.draw (ReactViewGroup.java:934)
  at android.view.View.updateDisplayListIfDirty (View.java:22729)
  at android.view.View.draw (View.java:23604)
  at android.view.ViewGroup.drawChild (ViewGroup.java:4766)
  at android.view.ViewGroup.dispatchDraw (ViewGroup.java:4549)
  at android.view.View.updateDisplayListIfDirty (View.java:22720)
  at android.view.View.draw (View.java:23604)
  at android.view.ViewGroup.drawChild (ViewGroup.java:4766)
  at android.view.ViewGroup.dispatchDraw (ViewGroup.java:4525)
  at android.view.View.updateDisplayListIfDirty (View.java:22720)
  at android.view.View.draw (View.java:23604)
  at android.view.ViewGroup.drawChild (ViewGroup.java:4766)
  at android.view.ViewGroup.dispatchDraw (ViewGroup.java:4525)
[...]

PR #4519, opened 2026-08-18
fix(Android, Stack v4+): append SwipeRefreshLayout transition filler

### LINKED ISSUE #4513: Placeholder View added to SwipeRefreshLayout during screen transition is never removed → getChildDrawingOrder() returned invalid index N (child count is N)
### Description

`startTransitionRecursive` inserts a placeholder `View` into every `SwipeRefreshLayout` it walks, and nothing ever removes it. When that placeholder is later removed by React's reconciliation, `SwipeRefreshLayout.mCircleViewIndex` is left pointing past the end of the child list until the next measure pass — and a draw that lands in that window crashes.

**Screen.kt (4.16.0)** — the add side:

```kotlin
private fun startTransitionRecursive(parent: ViewGroup?) {
    parent?.let {
        for (i in 0 until it.childCount) {
            val child = it.getChildAt(i)
            if (parent is SwipeRefreshLayout && child is ImageView) {
                // ...
                // We add a simple View to bump the number of children to make it work.
                // TODO: find a better way to handle this scenario
                it.addView(View(context), i)
            } else {
                child?.let { view -> it.startViewTransition(view) }
            }
            // ...
```

…and the end side, which only ends transitions and never removes what was added:

```kotlin
private fun endTransitionRecursive(parent: ViewGroup) {
    parent.children.forEach { childView ->
        parent.endViewTransition(childView)
        // ...
    }
}
```

**Why it produces exactly this message.** AndroidX `SwipeRefreshLayout` recomputes `mCircleViewIndex` only in `onMeasure`, and orders drawing with:

```java
protected int getChildDrawingOrder(int childCount, int i) {
    if (mCircleViewIndex < 0) return i;
    else if (i == childCount - 1) return mCircleViewIndex;   // draw the circle last
    else if (i >= mCircleViewIndex) return i + 1;
    else return i;
}
```

Sequence:

1. Steady state — children are `[content, circle]`, `mCircleViewIndex = 1`, `childCount = 2`. Fine.
2. Screen transition — the placeholder is inserted **before** the circle: `[content, placeholder, circle]`. After the next measure, `mCircleViewIndex = 2`, `childCount = 3`. Still fine.
3. The placeholder is removed (it is not part of the React tree, so reconciliation drops it) → `childCount = 2`, but `mCircleViewIndex` is still `2` until `onMeasure` runs again.
4. A draw in that window calls `getChildDrawingOrder(2, 1)`; `i == childCount - 1`, so it returns `mCircleViewIndex` = **2**. `ViewGroup.getAndVerifyPreorderedIndex` rejects it:

```
java.lang.IndexOutOfBoundsException: getChildDrawingOrder() returned invalid index 2 (child count is 2)
    at android.view.ViewGroup.getAndVerifyPreorderedIndex(ViewGroup.java:2125)
    at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4427)
    at android.view.View.draw(View.java:26349)
    at android.view.View.updateDisplayListIfDirty(View.java:25099)
    at android.view.View.draw(View.java:26032)
    at android.view.ViewGroup.drawChild(ViewGroup.java:4688)
    at com.facebook.react.views.view.ReactViewGroup.drawChild(ReactViewGroup.kt:900)
```

The reported index is always exactly `childCount`, which is the signature of this off-by-one rather than of a generic drawing-order desync.

Worth noting that React Native's own `ViewGroupDrawingOrderHelper.getChildDrawingOrder` has a recovery path for stale indices that cannot fire, because `update()` clears the field while the guard below re-tests a local copy captured before the call — so a stale array is returned anyway. That is a separate RN issue, but it means nothing upstream of this rescues the bad index either.

**Production impact:** 16 crashes across 13 distinct users in 9 days, Android 9–16, ~10 different device models (Samsung, Xiaomi, Motorola, Infinix). ~80% of the crashes happen while pushing a screen, not while popping. Every occurrence carries the same numbers: returned index == child count.

### Steps to reproduce

Reliably observed in production, not yet reduced to a minimal case:

1. A native stack with two screens, **each** containing a `FlatList` (or `ScrollView`) with a `refreshControl={<RefreshControl … />}`.
2. Navigate from one to the other repeatedly on a physical Android device, release build, New Architecture enabled.
3. The crash appears on a small fraction of transitions — device- and timing-dependent, since it needs a draw between the placeholder's removal and the next measure pass.

I do not have a minimal repro, and I would rather be upfront about that than open a low-signal report — #3096 was closed as `missing-repro` with the same symptom.

What I can offer instead is the mechanism above, which is derived from the 4.16.0 source and the AndroidX implementation, and predicts the exact error string (returned index always equal to child count). Happy to build a repro app if that is what unblocks this; I would also be glad to test a patch against real traffic, since the crash reproduces on our users within hours.

Two directions that look plausible from the outside:

- track the inserted placeholder and remove it in `endTransitionRecursive`, so the child count returns to its previous value at a controlled point rather than whenever reconciliation gets to it; or
- skip both branches for `SwipeRefreshLayout`'s `CircleImageView` — never mutating the child list removes the window entirely, at the cost of the circle not being preserved through the transition.

### Screens version

4.16.0

### React Native version

0.81.5 (Expo SDK 54)

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

Expo (bare/prebuild)

### Build type

Release

### Device

Real device

### Device model

Reproduced across ~10 models — Samsung Galaxy A03/A05/A07/A55/A75/M53/S20 FE, Xiaomi Redmi (2312DRAABG, 23122PCD1I, 25078PC3EG), Infinix X6525B. Android 9 through 16.

### Acknowledgements

Yes

drhops:
## Description

`Screen.startTransitionRecursive` inserts a filler `View` into every `SwipeRefreshLayout`
it walks, at the progress circle's own index:

```kotlin
if (parent is SwipeRefreshLayout && child is ImageView) {
    // ...
    // We add a simple View to bump the number of children to make it work.
    // TODO: find a better way to handle this scenario
    it.addView(View(context), i)   // `i` is the circle's index
}
```

That placement pushes the circle one slot down the child list. AndroidX `SwipeRefreshLayout` recomputes `mCircleViewIndex` **only in `onMeasure`**, and consumes it on every draw:

```java
protected int getChildDrawingOrder(int childCount, int i) {
    if (mCircleViewIndex < 0)      return i;
    else if (i == childCount - 1)  return mCircleViewIndex;  // draw the circle last
    else if (i >= mCircleViewIndex) return i + 1;
    else                            return i;
}
```

So if a layout pass lands while the filler is in place, the layout caches the *inflated* circle index. When the content child is then removed into a view transition, `childCount` drops but the cache does not — and the next draw returns an index past the end of the list. `ViewGroup.getAndVerifyPreorderedIndex` throws:

```
java.lang.IndexOutOfBoundsException: getChildDrawingOrder() returned invalid index 2 (child count is 2)
  at android.view.ViewGroup.getAndVerifyPreorderedIndex(ViewGroup.java:2239)
  at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4522)
  ...
  at com.swmansion.rnscreens.ScreenStack.drawAndRelease(ScreenStack.kt:64)
  at com.swmansion.rnscreens.ScreenStack.performDraw(ScreenStack.kt:334)
```

The reported index always equals the child count, which is the signature of this particular cache going stale.

### Why appending fixes it

Walking every state, with `C` = content, `O` = circle, `X` = filler:

| state | children (before) | count / cached | children (after) | count / cached |
| measured while filler present | `C X O` | 3 / **2** | `C O X` | 3 / **1** |
| content detached into the transition | `X O` | 2 / **2** → 💥 | `O X` | 2 / **1** → ok |

The cached index only has to stay inside the child list, and the circle only has to be the last child *drawn*. Appending satisfies both in every state, because the circle's index no longer moves when a filler is added. In the healthy case the two placements produce **identical paint order** — `content, filler, circle` either way — because `getChildDrawingOrder` already hoists the circle to last regardless of where it sits.

In the stale window the patched layout draws the circle first and the filler last, which is the one behavioural difference. The filler is a bare `View` with no background, so it paints nothing and cannot occlude the circle.

Any placement *before* the circle has the same defect — inserting at index `0` gives `X C O` / cached `2`, which fails identically once `C` leaves. Appending is the only placement that survives.

### Why the filler itself has to stay

It looks removable, but it is load-bearing. With no filler at all, the content child leaving takes `childCount` to 1 against a cached index of 1:

```
getChildDrawingOrder() returned invalid index 1 (child count is 1)
```

That is what the filler is compensating for — it holds `childCount` above the stale cached index. This change keeps that property and only moves where the filler sits.

### Why it is hard to reproduce interactively

If the pop unmounts the whole subtree in a single mount transaction, **no layout pass runs between the filler insert and the content child's detachment**, so the cached index is still the pre-filler value (`1`) and *both* placements are safe. The crash needs something to measure the outgoing `SwipeRefreshLayout` while the filler is in place.

In the production traces we have, the frames underneath `ScreenStack.performDraw` include `RecyclerView`/ViewPager2 — a pager (`react-native-pager-view` under a tab view) hosting the refresh control. The pager keeps offscreen pages mounted and measures them, which supplies exactly the layout pass the bug needs while the exit animation is still running.

The `invalid index 3 (child count is 3)` variant in #3096 is the same defect one slot further along, consistent with a `SwipeRefreshLayout` that survives a removal cycle with a filler still attached (nothing removes them) and accumulates another on the next one.

Closes #4513.
Likely the same root cause as #3096.

## Changes

- `legacy/Screen.kt`: append the transition filler (`it.addView(View(context))`) rather than inserting it at the circle's index, and record why the placement matters.

## Test plan

**Deterministic reproduction (JVM, Robolectric).** Reproduced and verified against `react-native-screens` 4.23.0 on RN 0.83.6, Robolectric 4.15.1, `@Config(sdk = [34])`. This drives the real `Screen.startRemovalTransition()` path — it throws on stock 4.23.0 and passes with this change:

<details><summary>RefreshControlScreenRemovalTest.kt</summary>

```kotlin
@RunWith(RobolectricTestRunner::class)
@Config(sdk = [34])
class RefreshControlScreenRemovalTest {
    private lateinit var activity: Activity
    private lateinit var reactContext: ReactApplicationContext

    @Before
    fun setUp() {
        activity = Robolectric.buildActivity(Activity::class.java).setup().get()
        DisplayMetricsHolder.initDisplayMetricsIfNotInitialized(activity)
        reactContext = BridgeReactContext(RuntimeEnvironment.getApplication())
    }

    /** Reproduces the crash against upstream's filler placement, independently of the fix. */
    @Test
    fun `filler inserted ahead of the progress circle crashes the next draw`() {
        val refreshLayout = ReactSwipeRefreshLayout(reactContext)
        val content = View(activity)
        refreshLayout.addView(content, 0)
        activity.setContentView(refreshLayout)

        layOut(refreshLayout)
        assertEquals("circle starts after the content child", 1, indexOfCircle(refreshLayout))

        // Screen.kt today: `it.addView(View(context), i)`, with `i` the circle's index.
        refreshLayout.addView(View(activity), indexOfCircle(refreshLayout))
        layOut(refreshLayout)
        assertEquals("filler pushed the circle down a slot", 2, indexOfCircle(refreshLayout))

        refreshLayout.startViewTransition(content)
        refreshLayout.removeView(content)
        assertEquals(2, refreshLayout.childCount)

        try {
            draw(refreshLayout)
            fail("expected the stale circle index to be rejected while drawing")
        } catch (e: IndexOutOfBoundsException) {
            assertTrue(
                "expected the drawing-order failure, got: ${e.message}",
                e.message.orEmpty().contains("getChildDrawingOrder() returned invalid index 2"),
            )
        }
    }

    /** The same sequence through the real removal path. Fails on 4.23.0, passes with this change. */
    @Test
    fun `refresh control still draws after a patched screen removal`() {
        val screen = Screen(ThemedReactContext(reactContext, activity, null, -1))
        val refreshLayout = ReactSwipeRefreshLayout(reactContext)
        val content = View(activity)
        refreshLayout.addView(content, 0)
        screen.addView(refreshLayout)
        activity.setContentView(screen)

        // Measure the refresh layout itself: ReactViewGroup sizes its children from the shadow
        // tree rather than measuring them, so laying out the Screen alone never refreshes the cache.
        layOut(refreshLayout)
        assertEquals("circle starts after the content child", 1, indexOfCircle(refreshLayout))

        screen.startRemovalTransition()
        assertEquals("screens adds one filler view", 3, refreshLayout.childCount)
        // The measure that caches the circle's index while the filler is in place.
        layOut(refreshLayout)

        refreshLayout.removeView(content)
        assertEquals(2, refreshLayout.childCount)

        draw(screen)   // throws IndexOutOfBoundsException on stock 4.23.0
    }

    private fun layOut(view: View) {
        view.measure(
            MeasureSpec.makeMeasureSpec(WIDTH, MeasureSpec.EXACTLY),
            MeasureSpec.makeMeasureSpec(HEIGHT, MeasureSpec.EXACTLY),
        )
        view.layout(0, 0, WIDTH, HEIGHT)
    }

    private fun draw(view: View) {
        view.draw(Canvas(Bitmap.createBitmap(WIDTH, HEIGHT, Bitmap.Config.ARGB_8888)))
    }

    private fun indexOfCircle(parent: ViewGroup): Int =
        (0 until parent.childCount).first { parent.getChildAt(it) is ImageView }

    private companion object {
        const val WIDTH = 1080
        const val HEIGHT = 1920
    }
}
```

</details>

Verified by reverting `Screen.kt` in `node_modules` to the current upstream form and re-running: the second test fails with `getChildDrawingOrder() returned invalid index 2 (child count is 2)`, and passes again with the filler appended.

**Production evidence.** This is the fix we are shipping. The crash affected 18 users on our latest production build, skewed to Android 16 and Samsung devices, on `react-native-screens` 4.23.0 with RN 0.83.6 / Fabric.

**Repro.** I do not have a reliable interactive on-device reproducer — the timing window is narrow, and per the analysis above it needs a layout pass on the refresh layout during the exit animation. What I do have is a deterministic reproduction of the exact mechanism, production crash data, and the state analysis showing the change is a no-op whenever the cache is fresh.


Caulez:
Reporter of #4513 here. Confirming this from production, including the part that is hardest to take on faith: that the filler is load-bearing.

We shipped the other candidate fix, dropping the filler entirely, to our whole Android user base yesterday, on `react-native-screens` 4.16.0 / RN 0.81.5 / Fabric. It produced exactly the one-slot-earlier failure this PR's analysis predicts:

```
IndexOutOfBoundsException: getChildDrawingOrder() returned invalid index 1 (child count is 1)
```

Volume, same app, same screens, two consecutive releases:

| build | variant | crashes |
|---|---|---|
| 1.2.3+16 | upstream (`addView(view, i)`) | 16 events / 13 users in 9 days |
| 1.3.0+17 | filler removed | 194 events / 71 users in 1 day |

So removing the filler does not close the window, it moves the off-by-one one slot and makes it fire on every screen removal with a `RefreshControl` rather than only in the narrow timing window. We have switched to this PR's placement.

@t0maboro this also explains why the reproduction attempt above cannot crash: it has the filler insert and the content detach, but nothing measures the `SwipeRefreshLayout` in between, so the cached index is never the inflated one.

rakesh-stan:
@Caulez, so is it working fine with this fix in your app, without any crash reports related to this issue?

kkafar (commented):
Hey, thanks for the PR. 

As @t0maboro mentioned in the linked issue - we'll wait with landing this until the issues author confirms that the patch solves the problem. 

The patch you propose itself if fine I guess. It doesn't improve the approach, however it doesn't degrade it either. 

My thoughts are, that we could try to actually remove the placeholder views when ending the view transition, but let's see whether the patch works first.

I guess you're running this patch in your production? How s it going?

kkafar:
Oh, I've just noticed that @Caulez confirmed it here 🤦🏻 😄 

Edit: but I don't know what conclusions to draw from that comment - let's give @Caulez some time to give us a conclusive update on the matter.

kkafar (commented):
Haven't verified it yet, but one possible explanation of why the override with index is used there came to my mind.

The view could be added BEFORE the image view, because later, when react unmounts, it unmounts the dummy view from that index and that prevents the loading-indicator from disappearing.

There are few disclaimers needed: I'm not even sure, that React manages that view hierarchy or not, but IIRC that loading indicator is customisable via prop on ScrollView, therefore react might actually add/remove views there. 

Second thing - I've no idea whether it's a thing or not - it's just an idea that crossed my mind. 

Third thing - I guess disappearing loading indicator is better than a crash, therefore I'll consider proceeding here.

Fourth thing - We need to definitely test this & see whether there is a regression or not. I dunno when we'll have resources to test this properly. No promises here.

drhops:
Thanks for looking into this @kkafar!  I can't speak for @Caulez, but we deployed this PR as a patch in our app release on Friday and it appears to be resolved / haven't seen the crash on the new version since:

Screenshot 2026-08-24 at 12 11 43 PM

Caulez:
Conclusive update.

We shipped this PR's placement to our entire Android base in 1.3.1 (versionCode 19), live since Aug 21, on `react-native-screens` 4.16.0 / RN 0.81.5 / Fabric. Part of the base hasn't updated yet, so the two builds are running side by side: same app, same screens, same days. That makes this an A/B rather than a before/after.

| build | filler | Android users active, last 3d | crashes, Aug 22-24 |
|---|---|---|---|
| 1.3.0+17 | removed (the wrong fix) | 131 | 171 (115 / 39 / 17 per day) |
| 1.3.1+19 | this PR (appended) | 458 | 0 |

Totals for the issue, by release: 1092 on `1.3.0+17`, 15 on `1.2.3+16` (upstream `addView(view, i)`), zero on `1.3.1+19`. The last event on the fixed build is not "a few hours ago", it does not exist. Meanwhile the broken build was still firing today, from 3.5x fewer users.

@kkafar on the index hypothesis: I don't think React owns that slot. The filler is constructed in `Screen.kt` as a bare `View(context)` and added directly to the `SwipeRefreshLayout`, so it never enters the React view registry and no unmount can target its index. Paint order is unchanged either way, since `getChildDrawingOrder` hoists the circle to last wherever it sits. Three days and 458 users in, we have no reports of a stuck refresh indicator, though I'll say plainly that we have no telemetry for that, only the absence of complaints.

Removing the placeholder views at the end of the transition sounds strictly better than either placement, and I'd be happy to run that in production too, if you want the same measurement on it.

kkafar:
Thanks for the updates!

That sounds convincing. We'll have @Loloekk give this problem a deeper look, testing for regressions. If our testing is good, we'll proceed here.

Loloekk (approved):

Loloekk:
Hey, 
I have checked your changes. It looks great. In issue test 4519 [241d9af](https://github.com/software-mansion/react-native-screens/pull/4519/commits/241d9af5abb1104a521c4c92e2010c54b56c4c07) the circle does not disappear during popping. 

| Before | After |
|--------|-------|
| <video src="https://github.com/user-attachments/assets/1144e496-5585-40cb-bc2d-3c0307fbdf2e" controls="controls" width="100%"></video> | <video src="https://github.com/user-attachments/assets/ddac7575-0369-4723-a9b5-7f8e4a3e4603" controls="controls" width="100%"></video> |


The edited line is hit only when we use JS back button. Stopped animation is caused by hitting a breakpoint.

PR #4524, opened 2026-08-19
fix(android) transparent header backgroundColor dropped by Props 2.0

delekta:
## Description

On Android Fabric with Props 2.0 (`enablePropsUpdateReconciliationAndroid`), `ScreenStackHeaderConfig` `backgroundColor="transparent"` never reaches native. JS still has the color (`processColor` is `0`). Native never calls `setBackgroundColor`, so the toolbar keeps Material `colorPrimary`.

iOS is unaffected (it does not use the `getDiffProps` path).

Android `HostPlatformColor::UndefinedColor` is `0`, and so is `processColor('transparent')` / `'#00000000'` / `'rgba(0,0,0,0)'`. Without `generateOptionalProperties`, codegen emits `SharedColor backgroundColor{}` whose default is also `0`. On CREATE, Props 2.0 diffs against that default, omits the key, and Kotlin never paints.

This is a CREATE-path bug. A red → transparent UPDATE diffs against the previous color and can hide it.

**Minimal repro:** https://github.com/delekta/screens-transparent-header-bug-repro  
RN `0.87.0`, screens `4.27.0`, screens-only (no React Navigation required).

Do **not** disable Props 2.0 as the library fix.

## Changes

- Set `generateOptionalProperties: true` on `RNSScreenStackHeaderConfig` so omitted is `nullopt` and transparent is `optional(0)`. The diff then emits `0` and native `setBackgroundColor(0)` runs.
- The flag is component-wide (optional strings, colors, mixed arrays), so native readers are updated:
  - Android C++: `title.empty()` → `!title.has_value() || title->empty()` in `RNSScreenShadowNode.cpp`
  - iOS Fabric `updateProps`: unwrap `std::optional` for `RCTNSStringFromStringNilIfEmpty`, `RCTUIColorFromSharedColor`, and header bar-button vectors
- Kotlin ViewManagers are unchanged (`Int?` already). Unset color still skips the setter (theme bar). Transparent now applies `0`.

Please also cherry-pick onto `main` (5.0 / `src/fabric/legacy/…`) if that spec is still used there.

Chip **2. transparent** on the repro (CREATE / remount). Expected: lime page shows through the header. Fail: leftover Material bar.

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/bf95f40b-49e7-4016-81b1-418823cb434d" />  | <video src="https://github.com/user-attachments/assets/10f0a3cd-6a60-4316-9432-d4320d9a6c60" />  |

Opaque red (chip 1) should stay red in both.

## Test plan

Repro: https://github.com/delekta/screens-transparent-header-bug-repro

```sh
npm install
npm start
npm run android
```

Keep `ENABLE_PROPS_2_0 = true` in `MainApplication.kt`. Each chip remounts the stack so 2–4 are CREATE.

- Chip **1. Red** — header and page both red (control).
- Chip **2. transparent** — lime page; header must look lime (see-through). Fail = leftover Material bar.
- Chips **3 / 4** — `#00000000` and `rgba(0,0,0,0)` same as transparent.
- Chip **5** — `#FFFFFF00` still works (alpha 0, RGB ≠ 0).
- Chip **6** — dark, then UPDATE to transparent still works.
- Control: set `ENABLE_PROPS_2_0 = false`, rebuild; chip 2 already correct without this PR.
- iOS: header title, colors, and bar button items unchanged.

Same bug via native-stack: in the repro `index.js` import `./App.native-stack` instead of `./App`.

No new example files in this repo; the repro app is the test.

## Checklist

- [ ] Ensured that CI passes
```

### DIFF
--- a/common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.cpp
+++ b/common/cpp/react/renderer/components/rnscreens/RNSScreenShadowNode.cpp
@@ -122,7 +122,7 @@ void RNSScreenShadowNode::appendChild(
           ? 0.f
           : findHeaderHeight(
                 headerProps.titleFontSize,
-                headerProps.title.empty(),
+                !headerProps.title.has_value() || headerProps.title->empty(),
                 applyTopInset)
                 .value_or(0.f);
 
--- a/ios/RNSScreenStackHeaderConfig.mm
+++ b/ios/RNSScreenStackHeaderConfig.mm
@@ -15,6 +15,7 @@
 #import <react/renderer/components/rnscreens/RCTComponentViewHelpers.h>
 #import <react/utils/ManagedObjectWrapper.h>
 #import <rnscreens/RNSScreenStackHeaderConfigComponentDescriptor.h>
+#import <optional>
 #import "RCTImageComponentView+RNSScreenStackHeaderConfig.h"
 #import "RNSBackBarButtonItem.h"
 #import "RNSBarButtonItem.h"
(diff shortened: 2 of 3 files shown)


t0maboro:
@delekta Thanks for the PR! After digging in, I'm convinced that the root cause is in RN core, not in our spec. As you've already described, on Android, `HostPlatformColor::UndefinedColor` is 0, which is also `processColor('transparent')`. The Props 2.0 diffing drops `'transparent'` and emits `0`, resulting in using the theme-default value for every `SharedColor` prop, including core's own components: `<Switch thumbColor="transparent">`.

I've reported it to upstream https://github.com/react/react-native/issues/58085. We need to hold this PR until we receive some feedback from Meta about that behavior, because I am unwilling to merge a workaround for a single component when the real RC is still in the core and it touches other libraries/native components as well.

PR #4526, opened 2026-08-20
fix(Android, Tabs): prevent native view state restoration in TabsScreen

### LINKED ISSUE #4523: [Android, Tabs] ClassCastException when TabsScreen restores stale Fabric view state
### Description

On Android with Fabric and native bottom tabs, `TabsScreen` allows `FragmentManager` to save and restore the native state of its descendant views.

React Native can reuse an Android view ID after a view has been unmounted. When a tab fragment is reattached, Android may restore state saved for the previous view into a different view type that now has the same ID.

We observed this crash in production:

```text
java.lang.ClassCastException:
android.view.AbsSavedState$1 cannot be cast to
android.widget.CompoundButton$SavedState
```

`CompoundButton.onRestoreInstanceState()` expects `CompoundButton.SavedState`, but it receives state previously saved by a different view with the same Android ID.

The regular `Screen` implementation already prevents this by overriding `dispatchSaveInstanceState` and `dispatchRestoreInstanceState` as no-ops:

https://github.com/software-mansion/react-native-screens/blob/4.27.0/android/src/main/java/com/swmansion/rnscreens/Screen.kt#L220-L228

That behavior was introduced in:

https://github.com/software-mansion/react-native-screens/pull/313

`gamma/tabs/screen/TabsScreen.kt` is a separate `ViewGroup` and does not contain the same overrides:

https://github.com/software-mansion/react-native-screens/blob/4.27.0/android/src/main/java/com/swmansion/rnscreens/gamma/tabs/screen/TabsScreen.kt

The issue reproduces on 4.26.2. Source inspection shows that 4.27.0 and the current `main` branch also lack these overrides in `TabsScreen`.

React Native owns this view hierarchy, so Android should not persist and restore descendant state independently. `TabsScreen` should follow the same behavior as `Screen`.

A proposed fix is:

```kotlin
override fun dispatchSaveInstanceState(
    container: SparseArray<Parcelable>,
) {
    // React Native owns this view hierarchy.
}

override fun dispatchRestoreInstanceState(
    container: SparseArray<Parcelable>,
) {
    // React Native owns this view hierarchy.
}
```

A deterministic Robolectric regression test fails before these overrides with the exact `ClassCastException` above and passes after adding them.

### Minimal reproduction

- [Repository](https://github.com/anirudhsama/react-native-screens-tabs-state-repro)
- [Screen recording](https://github.com/anirudhsama/react-native-screens-tabs-state-repro/blob/main/artifacts/reproduction.mp4)
- [Android crash log](https://github.com/anirudhsama/react-native-screens-tabs-state-repro/blob/main/artifacts/crash-log.txt)

The repository contains a one-tap Expo app and a deterministic Robolectric test pinned to `react-native-screens@4.26.2`. The test fails with the exact exception on the unpatched dependency and passes after adding the two `TabsScreen` state-dispatch overrides.

### Steps to reproduce

The state collision can be reproduced without relying on timing:

1. Create a `TabsScreen`.
2. Add a `CheckBox` with an Android view ID.
3. Save hierarchy state from a plain `View` using the same ID.
4. Restore that hierarchy state through `TabsScreen`.
5. Observe `AbsSavedState$1 cannot be cast to CompoundButton$SavedState`.

The core reproduction is:

```kotlin
val childId = 42
val tabsScreen = TabsScreen(themedReactContext).apply {
    addView(CheckBox(activity).apply { id = childId })
}

val staleHierarchyState = SparseArray<Parcelable>()

View(activity)
    .apply { id = childId }
    .saveHierarchyState(staleHierarchyState)

tabsScreen.restoreHierarchyState(staleHierarchyState)
```

Before the proposed fix, the final call crashes in `CompoundButton.onRestoreInstanceState()`. After the fix, `TabsScreen` ignores the stale native hierarchy state and the test passes.

### Screens version

4.26.2

Also confirmed by source inspection in 4.27.0 and current `main`.

### React Native version

0.86.2

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow with prebuild

### Architecture

Fabric / New Architecture

### Build type

Release mode in production. The regression is also reproducible in an API 33 Robolectric test.

### Device

Real devices in production. The deterministic reproduction uses Robolectric API 33.

### Acknowledgements

Yes

kkafar:
## Description

On Android with Fabric and native bottom tabs, `TabsScreen` is the view returned from `TabsScreenFragment.onCreateView`, so `FragmentManager` saves and restores the native hierarchy state (`View#saveHierarchyState` / `View#restoreHierarchyState`) of the entire react-owned subtree beneath it.

View ids in that subtree are react tags, which are not stable identities across remounts. Restored state can therefore be applied to a different view type that now has the same id, crashing in widgets that cast the saved state unchecked, e.g.:

```text
java.lang.ClassCastException:
android.view.AbsSavedState$1 cannot be cast to
android.widget.CompoundButton$SavedState
```

React Native owns this view hierarchy and keeps it alive, so native save/restore of descendant state is both unnecessary and unsound here. Legacy `Screen` has suppressed it since #313 — this PR applies the same treatment to `TabsScreen`.

Note: `StackScreenFragment`'s view root has the same gap; a follow-up PR will address Stack v5 separately.

Closes #4523.

## Changes

- Override `dispatchSaveInstanceState` / `dispatchRestoreInstanceState` as no-ops in `TabsScreen`, mirroring `Screen` (legacy) behavior.

## Test plan

- Reporter's deterministic reproduction: https://github.com/anirudhsama/react-native-screens-tabs-state-repro — contains a one-tap Expo app and a Robolectric test that fails with the exact `ClassCastException` on the unpatched library and passes with these overrides (this repository currently has no Android unit test infrastructure, so the test is not ported here).
- Manual reproduction from the issue:

```kotlin
val childId = 42
val tabsScreen = TabsScreen(themedReactContext).apply {
    addView(CheckBox(activity).apply { id = childId })
}

val staleHierarchyState = SparseArray<Parcelable>()

View(activity)
    .apply { id = childId }
    .saveHierarchyState(staleHierarchyState)

tabsScreen.restoreHierarchyState(staleHierarchyState)
```

Before this change the final call crashes in `CompoundButton.onRestoreInstanceState`; after it, the stale native state is ignored.

- Verified locally: `yarn lint-android` passes, `:react-native-screens:compileDebugKotlin` builds clean.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/tabs/screen/TabsScreen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/tabs/screen/TabsScreen.kt
@@ -2,6 +2,8 @@ package com.swmansion.rnscreens.tabs.screen
 
 import android.content.res.Configuration
 import android.graphics.drawable.Drawable
+import android.os.Parcelable
+import android.util.SparseArray
 import android.view.ViewGroup
 import androidx.fragment.app.Fragment
 import com.facebook.react.uimanager.ThemedReactContext
@@ -126,6 +128,17 @@ class TabsScreen(
         r: Int,
         b: Int,
     ) = Unit
+
+    override fun dispatchSaveInstanceState(container: SparseArray<Parcelable>) {
+        // Do nothing. React Native owns this view hierarchy and keeps it alive, so there is no
+        // need to serialize/deserialize native view state. View ids in this subtree are react
+        // tags, which are not stable identities - restoring state by id can apply state saved
+        // by one view type to a different one, crashing e.g. in CompoundButton (see #4523).
+    }
+
+    override fun dispatchRestoreInstanceState(container: SparseArray<Parcelable>) {
+        // Ignore restoring instance state too, as we are not saving anything anyways.
+    }
 
     internal fun setTabsScreenDelegate(delegate: TabsScreenDelegate?) {
         tabsScreenDelegate = WeakReference(delegate)


kligarski (approved):
I haven't checked the runtime but it looks reasonable.

PR #4527, opened 2026-08-20
fix(Android, Stack v5): prevent native view state restoration in stack screens

kkafar:
## Description

Follow-up to #4526 (Tabs), covering the same defect in Stack v5, as flagged in the discussion of #4523.

`StackHeaderCoordinatorLayout` is the view returned from `StackScreenFragment.onCreateView`, so `FragmentManager` saves and restores the native hierarchy state (`View#saveHierarchyState` / `View#restoreHierarchyState`) of the entire react-owned subtree beneath it — the `StackScreen` content as well as any react-owned header views hosted in the app bar.

View ids in that subtree are react tags, which are not stable identities across remounts. Restored state can therefore be applied to a different view type that now has the same id, crashing in widgets that cast the saved state unchecked, e.g.:

```text
java.lang.ClassCastException:
android.view.AbsSavedState$1 cannot be cast to
android.widget.CompoundButton$SavedState
```

React Native owns this view hierarchy and keeps it alive, so native save/restore of descendant state is both unnecessary and unsound here. Legacy `Screen` has suppressed it since #313; #4526 applied the same treatment to `TabsScreen`.

References #4523 (the issue itself is closed by the Tabs fix in #4526).

## Changes

- Override `dispatchSaveInstanceState` / `dispatchRestoreInstanceState` as no-ops in `StackHeaderCoordinatorLayout` — the fragment view root, so the override severs native state save/restore for the whole subtree, including react-owned header views that a `StackScreen`-level override would miss.

## Test plan

- Same failure mechanism as #4523; the reporter's deterministic reproduction for Tabs is https://github.com/anirudhsama/react-native-screens-tabs-state-repro — the Stack v5 equivalent substitutes the fragment view root:

```kotlin
val childId = 42
val stackScreen = StackScreen(themedReactContext)
val fragmentRoot = StackHeaderCoordinatorLayout(activity, stackScreen, canNavigateBack = false)
stackScreen.addView(CheckBox(activity).apply { id = childId })

val staleHierarchyState = SparseArray<Parcelable>()

View(activity)
    .apply { id = childId }
    .saveHierarchyState(staleHierarchyState)

fragmentRoot.restoreHierarchyState(staleHierarchyState)
```

Before this change the final call crashes in `CompoundButton.onRestoreInstanceState`; after it, the stale native state is ignored. (This repository currently has no Android unit test infrastructure, so the test is not ported here.)

- Verified locally: `yarn lint-android` passes, `:react-native-screens:compileDebugKotlin` builds clean.

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/stack/header/StackHeaderCoordinatorLayout.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/stack/header/StackHeaderCoordinatorLayout.kt
@@ -2,7 +2,9 @@ package com.swmansion.rnscreens.stack.header
 
 import android.annotation.SuppressLint
 import android.content.Context
+import android.os.Parcelable
 import android.util.Log
+import android.util.SparseArray
 import android.view.ViewGroup.LayoutParams.MATCH_PARENT
 import android.widget.FrameLayout
 import androidx.activity.OnBackPressedDispatcherOwner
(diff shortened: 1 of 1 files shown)


kligarski (approved):
Same as previous PR, looks good but I haven't checked the runtime.

PR #4533, opened 2026-08-20
feat(Android, Stack v5): expose `statusBarScrimColor` for the header

kligarski:
## Description

In edge-to-edge apps the stack header draws under the status bar. Whenever
scroll flags let header content leave the screen, the title and buttons pass
straight through the status-bar area with nothing masking them. This PR exposes
`statusBarScrimColor` prop and enables a sensible scrim by default, so status
bar is always displayed on a solid background instead of obstructing the
toolbar.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/905.

### Details

Material provides two independent status-bar scrim mechanisms, depending on
header type.

#### Small header — `AppBarLayout.statusBarForeground`

The small header is an `AppBarLayout` (ABL) hosting a `MaterialToolbar`. ABL can
draw a `statusBarForeground` drawable _after_ (above) all of its children, sized
to the full width × the top window inset and translated against the scroll
offset — so the strip stays glued to the top of the window while the toolbar
scrolls away underneath it.

We want the default strip to always match the bar's _effective_ background
color, including mid-flight during the lift-on-scroll animation (the background
animates between `backgroundColor` and its blend with
`scrolledBackgroundColor`). Material has a built-in sync for this, but it is
narrowly keyed: when a drawable is set as `statusBarForeground`, ABL captures
its color (possible only for `ColorDrawable`, `MaterialShapeDrawable` and
`ColorStateListDrawable`) and re-tints the strip on each animation frame _only
if the captured color equals the theme's `colorSurface`_. Any custom background
color breaks the sync (the strip freezes while the bar animates).

So instead we install a `PaintDrawable` as the strip — it is not
color-extractable, the captured color is `null`, and the built-in sync is
permanently inert regardless of theme — and drive the tint ourselves from the
public `AppBarLayout.LiftOnScrollProgressListener`. That listener is invoked
from the same `ValueAnimator` frame with the same per-frame mixed color the
built-in sync would apply, so the result is pixel-identical to the native effect
while working for arbitrary colors. The lift animation only runs on lifted-state
_changes_, so when color props change while the bar is already lifted, the
applicator jumps the strip straight to the blended end color (same pattern as
the existing background jump).

#### Medium / large headers — `CollapsingToolbarLayout.statusBarScrim`

Collapsing headers pair the ABL with a `CollapsingToolbarLayout` (CTL), which
has its own mechanism: `statusBarScrim`. It shares the alpha animation of the
content scrim (fades in when the header collapses, out when it expands) and is
drawn _after_ all children and after the collapsing title — unlike the content
scrim, which is drawn _below_ the toolbar child and therefore cannot mask
toolbar content.

The default matters here: with scroll-only flags (no `exitUntilCollapsed`) the
CTL exits the screen but leaves the toolbar behind the status-bar area.
`statusBarScrim` property can be used to mask it. We default it to the content
scrim color (`scrolledBackgroundColor`), which makes the masking seamless and is
a visual no-op in the common opaque `exitUntilCollapsed` configuration.

One artifact needed a workaround: both CTL scrims share the same `scrimAlpha`
and overlap in the status-bar strip, so mid-fade the strip composites two layers
at partial alpha — visibly darker than either scrim duration of the fade. The
fix removes the overlap: our content scrim (the bottom layer) is a small
`ColorDrawable` subclass that skips the strip band whenever a status bar scrim
is installed. The band is exactly `[-offset, -offset + topInset]` in CTL
coordinates — the same rect CTL assigns to the status bar scrim — and the
content scrim's visible top edge always aligns with the strip top, so the
exclusion is a single clip. The offset is tracked with an
`OnOffsetChangedListener` and the inset by observing `dispatchApplyWindowInsets`
(both ABL and CTL install their own `OnApplyWindowInsetsListener`, so setting
ours would silently replace Material's).

#### Defaults and policy

| Header             | `statusBarScrimColor` unset                                                              | Explicit color             |
| ------------------ | ---------------------------------------------------------------------------------------- | -------------------------- |
| `small`            | constant strip following the bar's effective background color through the lift animation | static strip of that color |
| `medium` / `large` | scrim in the content scrim color, fading with the collapse                               | that color, same fade      |

- The default scrim is installed only when the color it follows resolves to a
  fully opaque color — translucent headers stay see-through instead of
  getting a double-composited band. An explicit color is always honored.
- `'transparent'` disables the scrim entirely.
- The `small` strip is constant while the `medium`/`large` scrim shows only
  when collapsed — this asymmetry mirrors the native Material widgets.

## Changes

- Added the `statusBarScrimColor` prop (fabric spec, public Android types,
  config plumbing, view manager) reusing the `BACKGROUND_COLORS`
  invalidation flag.
- `StackHeaderAppBarLayout.Small`: installed a `PaintDrawable` status bar
  strip and a `LiftOnScrollProgressListener` keeping it in sync with the
  bar's effective color; the listener is detached when an explicit color is
  set.
- `StackHeaderApplicator`: new `applyStatusBarScrim` step dispatching per
  header type, with the opacity policy in a single
  `resolveStatusBarScrimColor` helper; invisible collapsing scrims are
  normalized to `null`.
- `StackHeaderContentScrimDrawable` + exclusion plumbing in
  `StackHeaderAppBarLayout.Collapsing`: the content scrim skips the
  status-bar strip while a status bar scrim is installed, so the two fading
  scrims never stack.
- Added the `test-stack-header-status-bar-scrim-android` SFT.

## Visual documentation

https://github.com/user-attachments/assets/1f3d8a01-fc6a-4398-81db-659bd10064ab

## Test plan

Run `test-stack-header-status-bar-scrim-android`,
`test-stack-header-background-android`.

## Checklist

- [x] Ensured that CI passes

---

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/stack/header/appbar/StackHeaderAppBarLayout.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/stack/header/appbar/StackHeaderAppBarLayout.kt
@@ -2,6 +2,8 @@ package com.swmansion.rnscreens.stack.header.appbar
 
 import android.annotation.SuppressLint
 import android.content.Context
+import android.graphics.Color
+import android.graphics.drawable.PaintDrawable
 import android.view.LayoutInflater
 import android.view.ViewGroup.LayoutParams.MATCH_PARENT
 import android.view.ViewGroup.LayoutParams.WRAP_CONTENT
@@ -56,8 +58,28 @@ internal sealed class StackHeaderAppBarLayout(
         internal val titleTextView: TextView
         internal val subtitleTextView: TextView
 
+        // Keeps the status bar strip in sync with the bar's effective color through
+        // the lift animation (same per-frame mixed color Material applies to the
+        // background). Detached when an explicit statusBarScrimColor is set.
+        private var statusBarScrimSyncEnabled = false
+        private val statusBarScrimSync =
+            object : LiftOnScrollProgressListener() {
+                override fun onUpdate(
+                    elevation: Float,
+                    backgroundColor: Int,
+                    progress: Float,
+                ) {
+                    statusBarForeground?.setTint(backgroundColor)
+                }
+            }
+
         init {
             addView(toolbar)
+
+            // PaintDrawable is not color-extractable, so AppBarLayout's built-in
+            // colorSurface-keyed strip tint sync can never activate and fight the
+            // listener above (or an explicit scrim color) — regardless of theme.
+            statusBarForeground = PaintDrawable().apply { setTint(Color.TRANSPARENT) }
 
             toolbar.title = TITLE_PLACEHOLDER
             toolbar.subtitle = SUBTITLE_PLACEHOLDER
@@ -65,6 +87,16 @@ internal sealed class StackHeaderAppBarLayout(
             subtitleTextView = toolbar.findTextViewWithText(SUBTITLE_PLACEHOLDER)
             toolbar.title = ""
             toolbar.subtitle = ""
+        }
+
+        internal fun setStatusBarScrimSyncEnabled(enabled: Boolean) {
+            if (statusBarScrimSyncEnabled == enabled) return
+            statusBarScrimSyncEnabled = enabled
+            if (enabled) {
+                addLiftOnScrollProgressListener(statusBarScrimSync)
+            } else {
+                removeLiftOnScrollProgressListener(statusBarScrimSync)
+            }
         }
 
         private fun MaterialToolbar.findTextViewWithText(text: String): TextView =
--- a/android/src/main/java/com/swmansion/rnscreens/stack/header/appbar/StackHeaderAppBarLayout.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/stack/header/appbar/StackHeaderAppBarLayout.kt
@@ -7,8 +7,10 @@ import android.graphics.drawable.PaintDrawable
 import android.view.LayoutInflater
 import android.view.ViewGroup.LayoutParams.MATCH_PARENT
 import android.view.ViewGroup.LayoutParams.WRAP_CONTENT
+import android.view.WindowInsets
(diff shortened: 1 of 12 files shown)


android/src/main/java/com/swmansion/rnscreens/stack/header/appbar/StackHeaderAppBarLayout.kt
```diff
@@ -115,6 +152,51 @@ internal sealed class StackHeaderAppBarLayout(
                 "[RNScreens] Collapsing StackHeaderAppBarLayout must be MEDIUM or LARGE type."
             }
             addView(collapsingToolbarLayout)
```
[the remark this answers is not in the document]
kligarski: image

src/components/stack/header/StackHeaderConfig.android.types.ts
```diff
@@ -793,6 +793,35 @@ export interface StackHeaderConfigPropsAndroid {
...
    */
   scrolledBackgroundColor?: ColorValue | undefined;
+  /**
+   * @summary Color of the scrim drawn in the status-bar area, masking header
+   * content that scrolls under the status bar in edge-to-edge apps.
+   *
+   * @description
+   * For the `small` header, the scrim is a constant strip pinned to the top of
+   * the window, drawn above the toolbar content. When not provided, the strip
+   * follows the bar's effective background color: {@link backgroundColor} at
```
t0maboro: "follows the bar's effective background color" - I'm wondering whether explicit mention that it's the appbar's bg color is carrying some context, because we're mentioning both appbar and statusbar in this doc

t0maboro (approved):

src/components/stack/header/StackHeaderConfig.android.types.ts
kligarski: https://github.com/software-mansion/react-native-screens/pull/4533/commits/8d8070decb4b3ac2631f462c99482698a4452b20

PR #4534, opened 2026-08-21
chore(test): e2e test for test-tabs-bottom-accessory-visibility-ios

LKuchno:
## Description

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1745

Adds Detox e2e coverage for the iOS `bottomAccessoryHidden` prop on `TabsHost`, covering the end state of each step of the `test-tabs-bottom-accessory-visibility-ios` scenario on iPhone.

Both `hidden` (`bottomAccessoryHidden`) and `rendered` (`bottomAccessory` set / `undefined`) remove the accessory from the native hierarchy, so presence is asserted the same way for both: the RNS accessory view exists and is visible with its content, and its frame sits fully above the tab bar's top edge.

`bottomAccessory` only exists on iOS 26+, so both bottom-accessory suites now run only there: a shared `describeIfiOS26` / `describeIfiPadOS26` in `e2e-utils` (`isIOSVersionAtLeast('26.0')`) skips them on older iOS and on Android.

Animation quality (smooth slide out/in, no blank frame) stays manual - Detox samples settled state, not animation frames.

## Changes

- New spec `FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-visibility-ios.e2e.ts` - one `it` per scenario step: baseline, `hidden` on/off, `rendered` off/on, and the combined `hidden` + `rendered` sequence.
- `FabricExample/e2e/e2e-utils.ts`: added `describeIfiOS26` and `describeIfiPadOS26`.
- `test-tabs-bottom-accessory-layout-ios.e2e.ts`: iPhone and `@ipad` suites switched to the iOS 26+ gates.
- Test screen `index.tsx`: added `testID`s only (`bottom-accessory-visibility-scrollview`, `rendered-switch`, `hidden-switch`, `bottom-accessory-text`).
- `scenario.md`: `## E2E test` section states the coverage split; `scenario-description.ts`: `e2eCoverage` → `'incomplete'`.

### DIFF
--- a/FabricExample/e2e/e2e-utils.ts
+++ b/FabricExample/e2e/e2e-utils.ts
@@ -39,7 +39,7 @@ export function isIOSVersionAtLeast(version: string): boolean {
 }
 
 /**
- * Suites for iOS 26+-only features (e.g. `bottomAccessory`, the header overflow
+ * Suites for iOS 26+ only features (e.g. `bottomAccessory`, the header overflow
  * button). `isIOSVersionAtLeast` is false on Android, so these stay iOS-only
  * and additionally self-skip on iOS 18 and older.
  */
--- a/FabricExample/e2e/e2e-utils.ts
+++ b/FabricExample/e2e/e2e-utils.ts
@@ -47,7 +47,7 @@ export const describeIfiOS26 = isIOSVersionAtLeast('26.0')
   ? describe
   : describe.skip;
 
-export const describeIfiPadIOS26 =
+export const describeIfiPadOS26 =
   isIPadTarget && isIOSVersionAtLeast('26.0') ? describe : describe.skip;
 
 export type ScrollOptions = {
--- a/FabricExample/e2e/e2e-utils.ts
+++ b/FabricExample/e2e/e2e-utils.ts
@@ -38,6 +38,18 @@ export function isIOSVersionAtLeast(version: string): boolean {
   );
 }
 
+/**
+ * Suites for iOS 26+-only features (e.g. `bottomAccessory`, the header overflow
+ * button). `isIOSVersionAtLeast` is false on Android, so these stay iOS-only
+ * and additionally self-skip on iOS 18 and older.
+ */
+export const describeIfiOS26 = isIOSVersionAtLeast('26.0')
+  ? describe
+  : describe.skip;
+
+export const describeIfiPadIOS26 =
+  isIPadTarget && isIOSVersionAtLeast('26.0') ? describe : describe.skip;
+
 export type ScrollOptions = {
   /** Pixels per step. Smaller steps avoid overshooting a short row. */
   pixels?: number;
--- a/FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-layout-ios.e2e.ts
+++ b/FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-layout-ios.e2e.ts
@@ -1,10 +1,10 @@
 import { device, expect, element, by } from 'detox';
 import { expect as jestExpect } from '@jest/globals';
 import {
-  describeIfiOS,
+  describeIfiOS26,
+  describeIfiPadIOS26,
   selectSingleFeatureTestsScreen,
   forceTapByLabeliOS,
-  describeIfiPad,
   getElementAttributes,
 } from '../../e2e-utils';
 import { IosElementAttributes } from 'detox/detox';
@@ -161,7 +161,7 @@ async function verifyConfigTabInitialContent() {
   ).toBeVisible();
 }
 
-describeIfiOS('Tabs bottomAccessory (iOS)', () => {
+describeIfiOS26('Tabs bottomAccessory (iOS 26+)', () => {
   beforeAll(async () => {
     await device.reloadReactNative();
     await selectSingleFeatureTestsScreen(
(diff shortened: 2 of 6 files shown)


FabricExample/e2e/e2e-utils.ts
```diff
@@ -44,6 +38,18 @@ export function isIOSVersionAtLeast(version: string): boolean {
...
+/**
+ * Suites for iOS 26+ only features (e.g. `bottomAccessory`, the header overflow
+ * button). `isIOSVersionAtLeast` is false on Android, so these stay iOS-only
+ * and additionally self-skip on iOS 18 and older.
+ */
+export const describeIfiOS26 = isIOSVersionAtLeast('26.0')
+  ? describe
+  : describe.skip;
+
+export const describeIfiPadIOS26 =
```
kligarski:
```suggestion
export const describeIfiPadOS26 =
```
nit

FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-layout-ios.e2e.ts
```diff
@@ -311,7 +311,7 @@ describeIfiOS('Tabs bottomAccessory (iOS)', () => {
   });
 });
 
-describeIfiPad('@ipad Tabs bottomAccessory (iPad)', () => {
+describeIfiPadIOS26('@ipad Tabs bottomAccessory (iPad, iOS 26+)', () => {
```
kligarski:
```suggestion
describeIfiPadOS26('@ipad Tabs bottomAccessory (iPadOS 26+)', () => {
```
nit

FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-visibility-ios.e2e.ts
```diff
@@ -0,0 +1,187 @@
...
+    .not.toExist()
+    .withTimeout(TRANSITION_TIMEOUT_MS);
+  await expect(element(by.id(ACCESSORY_TEXT))).not.toExist();
+
+  jestExpect(await getTabButtonWidth()).toBeLessThan(
+    tabButtonWidthWithAccessory,
+  );
+}
+
+async function expectConfigScreenResponsive() {
```
kligarski: I'm not sure if checking visibility can be considered responsiveness check but maybe we don't really need this at all? In next steps we're always toggling something so we check responsiveness in a way.

FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-visibility-ios.e2e.ts
```diff
@@ -0,0 +1,187 @@
...
+  CLASS_NAME_RNS_TABS_BOTTOM_ACCESSORY,
+  CLASS_NAME_UI_TAB_BAR,
+  CLASS_NAME_UI_TAB_BAR_BUTTON_IOS26,
+} from '../../native-class-names';
+
+/**
+ * Covers the end state of each `scenario.md` step: after every `hidden` /
+ * `rendered` toggle the accessory is either shown with its content above the
+ * tab bar, or gone. Presence is cross-checked against the tab item pill
+ * (`_UITabButton`), which spans the accessory's width while one is attached.
```
kligarski: I wonder, do we need to test this behavior at all? Isn't it enough that the accessory is visible or not? Do we really care about the buttons changing size? I guess it's unlikely that Apple will change this (but I recall that in iOS 26 betas the buttons would not change the size when bottom accessory was shown).

But I'm leaving this up to you to decide, I'm okay with both keeping and removing it.

kligarski (approved):
I left some suggestions but overall looks good.

FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-visibility-ios.e2e.ts
LKuchno: I removed it. The manual scenario doesn't check the pill resizing either, so the e2e was asserting more than the scenario it covers. It was also the most fragile assertion in the file: it relied on the private _UITabButton class and on undocumented layout behavior that already varied between iOS 26 betas, so it could break CI on a future iOS update without a real regression. The remaining checks (accessory presence/absence, its content, and its frame sitting above the tab bar) cover the actual feature.

https://github.com/software-mansion/react-native-screens/pull/4534/changes/f77d760d1cdba2167065d074c2f6eb519be99c33

FabricExample/e2e/e2e-utils.ts
LKuchno: https://github.com/software-mansion/react-native-screens/pull/4534/changes/704170c1d4691142f1d7b578901dcb87e9e87917

FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-layout-ios.e2e.ts
LKuchno: https://github.com/software-mansion/react-native-screens/pull/4534/changes/704170c1d4691142f1d7b578901dcb87e9e87917

FabricExample/e2e/single-feature-tests/tabs/test-tabs-bottom-accessory-visibility-ios.e2e.ts
LKuchno: Good point, I removed it - it was an additional check not mentioned in the manual scenario and not really needed looking at the other actions performed.

https://github.com/software-mansion/react-native-screens/pull/4534/changes/f77d760d1cdba2167065d074c2f6eb519be99c33

PR #4535, opened 2026-08-21
fix(iOS, Stack v5): Refactor header coordination to hide header when no headerConfig is provided

kmichalikk:
Closes https://github.com/software-mansion/react-native-screens-labs/issues/1699

## Description

Currently on v5 when no headerConfig is provided native header is rendered empty, with back button present. Additionally, the header is set to large variant by default, and that looks bad. This PR hides the header entirely when no config is provided, matching the behavior in v4.

> [!important]
> Without visible back button, screen dismiss gestures are disabled and swiping results in dismissing from outer stack. This is native behavior but indesirable in our case. We prevented this in v4 with a custom gesture recognizer, which is not implemented yet for v5, but will be in the future (https://github.com/software-mansion/react-native-screens-labs/issues/824). 

## Changes

- moved the header coordinator wiring header config earlier so we can catch the fact when screen `viewWillAppear` fires
- added method for setting header visibility
- added `screen.viewWillAppear` that hides the header if no config provider is present

n/a

## Test plan

Use simple navigation SFT, hide the header on one screen and navigate; use test-stack-header-subviews-ios, hide / show the header via prop updates. Verify that it hides.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/stack/header/RNSStackHeaderConfigComponentView.h
+++ b/ios/stack/header/RNSStackHeaderConfigComponentView.h
@@ -5,6 +5,7 @@
 #import "RNSStackHeaderConfigDataProviding.h"
 #import "RNSStackHeaderEventsDelegate.h"
 #import "RNSStackHeaderItemInvalidationDelegate.h"
+#import "RNSStackScreenHeaderCoordinator.h"
 #import "RNSViewFrameChangeDelegate.h"
 
 NS_ASSUME_NONNULL_BEGIN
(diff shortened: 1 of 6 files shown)


kligarski (changes requested):
https://github.com/user-attachments/assets/12d38e18-7c41-47ea-ab21-53c3d7caced8

Popping the screen with native back gesture doesn't work now. I think that we need to do the workaround from v4 if we want the back gesture to work with hidden navigation bar.

kmichalikk:
this is expected, we don't have gesture recognizer delegate wired up -- this is a larger task for another PR

kligarski:
1. There is no mention of it in the PR description and test plan.
2. Should we even merge this PR now if it breaks one of the most basic functionalities of the Stack?

kmichalikk:
@kligarski after today's discussion - can we proceed or would you like me to change something?

the logic to match v4 is planned to land at some point in the future, right now it matches what UIKit does by default

kligarski (approved):
Let's update PR description as I asked above and link an issue regarding the problem there.

PR #4536, opened 2026-08-21
fix(Example): Upgrade kotlin to 2.3.*

t0maboro:
## Description

For RCA, please see: https://github.com/software-mansion/react-native-screens/pull/4443 . We've internally discussed with @kkafar that the Kotlin upgrade should be safe because of their backward compatibility policy. This still diverges app from the RN template, but seems to be slightly better approach than applying the workaround with resolve RNS sources to its realpath.

Please note that bumping the `kotlinVersion` alone does nothing. The versionless KGP classpath entry resolves to the version pinned by `@react-native/gradle-plugin` that pins Kotlin 2.2.0, so the module kept compiling with Kotlin 2.2.* After pinning KGP explicitly, the resolved compiler is 2.3.*.

## Changes

- upgrade Kotlin to 2.3, remove stale workaround

N/A

## Test plan

How to reproduce the issue:

1. Downgrade kotlin to 2.2
2. Build FabricExample
3. Make any edit in `ScreenStackHeaderConfig.onUpdate`
4. Build FabricExample again - the incremental compilation fails witj

```
e: file:///Users/tomaszboron/react-native-screens/FabricExample/node_modules/react-native-screens/android/src/main/java/com/swmansion/rnscreens/legacy/ScreenStackHeaderConfig.kt:35:71 Cannot access 'class PointerEventsBoxNoneImpl : ReactPointerEventsView': it is internal in file.
e: file:///Users/tomaszboron/react-native-screens/FabricExample/node_modules/react-native-screens/android/src/main/java/com/swmansion/rnscreens/legacy/ScreenStackHeaderConfig.kt:264:32 Cannot access 'fun trySetWindowTraits(screen: Screen, activity: Activity?, context: ReactContext?): Unit': it is internal in 'com/swmansion/rnscreens/legacy/ScreenWindowTraits'.
e: file:///Users/tomaszboron/react-native-screens/FabricExample/node_modules/react-native-screens/android/src/main/java/com/swmansion/rnscreens/legacy/ScreenStackHeaderConfig.kt:361:75 Cannot access 'var isHiddenBySearchBar: Boolean': it is internal in 'com/swmansion/rnscreens/legacy/ScreenStackHeaderSubview'.
e: file:///Users/tomaszboron/react-native-screens/FabricExample/node_modules/react-native-screens/android/src/main/java/com/swmansion/rnscreens/legacy/ScreenStackHeaderConfig.kt:362:22 Cannot access 'fun setHiddenBySearchBar(hidden: Boolean): Unit': it is internal in 'com/swmansion/rnscreens/legacy/ScreenStackHeaderSubview'.
```

This should NOT happen after changing the Kotlin version to 2.3

## Checklist

- [ ] Ensured that CI passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the Android build configuration for improved compatibility with the latest Kotlin version.
  * Removed a workaround that could interfere with incremental compilation and linked project builds.

### DIFF
--- a/FabricExample/android/build.gradle
+++ b/FabricExample/android/build.gradle
@@ -25,6 +25,7 @@ buildscript {
     dependencies {
         classpath("com.android.tools.build:gradle")
         classpath("com.facebook.react:react-native-gradle-plugin")
+        // TODO: Can be removed when RN ships Kotlin >= 2.3.0
         // Pin KGP explicitly, because the version-less classpath entry resolves to the
         // version pinned by RN's gradle-plugin.
         classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:${kotlinVersion}")
--- a/FabricExample/android/build.gradle
+++ b/FabricExample/android/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         compileSdkVersion = 37
         targetSdkVersion = 36
         ndkVersion = "27.1.12297006"
-        kotlinVersion = "2.2.0"
+        kotlinVersion = "2.3.21"
     }
     repositories {
         google()
@@ -25,7 +25,9 @@ buildscript {
     dependencies {
         classpath("com.android.tools.build:gradle")
         classpath("com.facebook.react:react-native-gradle-plugin")
-        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin")
+        // Pin KGP explicitly, because the version-less classpath entry resolves to the
+        // version pinned by RN's gradle-plugin.
+        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:${kotlinVersion}")
         classpath rnrepoClasspath()
     }
 }
--- a/FabricExample/android/settings.gradle
+++ b/FabricExample/android/settings.gradle
@@ -4,16 +4,3 @@ extensions.configure(com.facebook.react.ReactSettingsExtension) { ex -> ex.autol
 rootProject.name = 'FabricExample'
 include ':app'
 includeBuild('../node_modules/@react-native/gradle-plugin')
-
-// Kotlin 2.2 incremental compilation can't see a module's own `internal` declarations 
-// when the project path goes through a symlink (see: https://youtrack.jetbrains.com/issue/KT-80039): 
-// KGP registers the previous compilation output as a canonicalized
-// "friend path" while the classpath keeps the symlinked path, so the visibility
-// check never matches. node_modules/react-native-screens is a symlink to the repo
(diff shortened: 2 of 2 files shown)


FabricExample/android/build.gradle
```diff
@@ -25,7 +25,9 @@ buildscript {
     dependencies {
         classpath("com.android.tools.build:gradle")
         classpath("com.facebook.react:react-native-gradle-plugin")
-        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin")
+        // Pin KGP explicitly, because the version-less classpath entry resolves to the
+        // version pinned by RN's gradle-plugin.
```
kligarski, line 30: Maybe we should add information that this pin can be removed when RN bumps Kotlin, similar to previous `// TODO: Can be removed when RN ships Kotlin >= 2.3.0 which resolved the issue.`. When somebody sees this randomly in a year or so it will have a higher chance of being cleaned up.

kligarski (approved):
Tested reproduction using this PR - build passed correctly after changes to `onUpdate`.

FabricExample/android/build.gradle
kkafar: +1 here

kkafar (approved):
Just a single request for a comment documenting when that code can be cleaned up again. 

Thanks.

FabricExample/android/build.gradle
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4536/commits/dec858416e16deaafbcaedf2085610eca942f6de

FabricExample/android/build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
         compileSdkVersion = 37
         targetSdkVersion = 36
         ndkVersion = "27.1.12297006"
-        kotlinVersion = "2.2.0"
+        kotlinVersion = "2.3.21"
```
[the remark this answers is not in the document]
t0maboro: Meta doesn't strictly follow these compatibility ranges in the Upgrade Helper either. KGP `2.2.0` lists Gradle `7.6.3-8.14` and AGP `7.3.1-8.10.0` as supported, which seems inconsistent with their claim that AGP 9 is supported for RN 0.87.

Given that RN core itself doesn't strictly follow this compatibility table and the backward compatibility of KGP 2.3.x when Meta considers 2.2.x to work correctly, I don't think this is an issue we need to address right now.

This only affects `FabricExample` anyway. If it becomes a problem, we can align with RN template and return to the workaround based on canonical paths.

PR #4544, opened 2026-08-24
feat(iOS, Stack v5): Add `backButtonMenuEnabled` prop for iOS header

kmichalikk:
Closes https://github.com/software-mansion/react-native-screens-labs/issues/848

## Description

This PR adds `backButtonMenuEnabled` to Stack v5 header on iOS. The new handling should be free of the limitations of v4 version (see https://github.com/software-mansion/react-native-screens/pull/3303). The old implementation used custom back button whereas this one provides a custom UINavigationBar class that disables the context menu interaction on the button and nothing else. The custom class is supported natively without major hacks via [dedicated init function](https://developer.apple.com/documentation/uikit/uinavigationcontroller/init(navigationbarclass:toolbarclass:)?language=objc) for the navigation controller.

Unfortunately, to block both long press and mouse right click (iPad, different interaction), accounting for UIKit randomly rebuilding the item whenever it wants and resetting any overwritten values, I had to use KVO to listen for `contextMenuInteractionEnabled` and write our value back. One may argue this is no less hacky than previous implementation, but this does not break any other prop combinations.

## Changes

- Added `backButtonMenuEnabled` prop
- Added custom `RNSStackNavigationBar` class which allows for suppressing back button menu
- updated the SFT for back button

https://github.com/user-attachments/assets/87626fe1-8559-487e-b4cb-e9a81cf681b7

## Test plan

Use `test-stack-back-button-ios` SFT.

## Checklist

- [ ] Ensured that CI passes


src/components/stack/header/StackHeaderConfig.ios.tsx
```diff
@@ -155,6 +156,7 @@ function StackHeaderConfig(
       collapsable={false}
       backButtonTitle={backButtonTitle}
       backButtonDisplayMode={backButtonDisplayMode}
+      backButtonMenuEnabled={backButtonMenuEnabled ?? true}
```
kligarski:
```suggestion
      backButtonMenuEnabled={backButtonMenuEnabled}
```

ios/stack/host/RNSStackNavigationController.mm
```diff
@@ -15,7 +16,12 @@ @implementation RNSStackNavigationController {
 
 - (instancetype)init
 {
-  if (self = [super init]) {
+#if !TARGET_OS_TV
+  self = [super initWithNavigationBarClass:RNSStackNavigationBar.class toolbarClass:nil];
+#else
```
kligarski:
```suggestion
#else // !TARGET_OS_TV
```

ios/stack/screen/RNSStackScreenController.mm
```diff
@@ -57,6 +57,9 @@ - (void)viewWillAppear:(BOOL)animated
 {
   [super viewWillAppear:animated];
   [self.headerCoordinator updateNavigationBarVisibilityAnimated:animated];
+#if !TARGET_OS_TV
+  [self.headerCoordinator updateBackButtonMenuEnabled];
+#endif // !TARGET_OS_TV
```
kligarski, line 62: These 2 functions are always called together - I wonder if the header coordinator should expose a method that runs both of these and use it instead so that in the future this doesn't get out of sync (e.g. only one of them is called). But it's not a big issue for now and you can argue that they are unrelated + sometimes some additional methods are called on the coordinator. But I just worry that we're going to add 3 more similar methods and everywhere we will have a block of these and then somebody will forget to call one of them.

But I'm okay with keeping it as is - up to you.

ios/stack/host/RNSStackNavigationBarCoordinator.mm
```diff
@@ -9,6 +10,15 @@ - (void)setHidden:(BOOL)hidden
   [navigationController setNavigationBarHidden:hidden animated:animated];
 }
 
+#if !TARGET_OS_TV
+- (void)setBackButtonMenuEnabled:(BOOL)enabled forNavigationController:(UINavigationController *)navigationController
+{
+  if ([navigationController.navigationBar isKindOfClass:RNSStackNavigationBar.class]) {
```
kligarski: Silent fail or an assertion?

kligarski (approved):
Looks good.

ios/stack/screen/RNSStackScreenController.mm
kmichalikk: I think I'll soon refactor it all in separate PR

ios/stack/host/RNSStackNavigationBarCoordinator.mm
kmichalikk: [83e84d2](https://github.com/software-mansion/react-native-screens/pull/4544/commits/83e84d283204c6bf4c34efc95ef32c31632e2c11) added a log

ios/stack/host/RNSStackNavigationController.mm
kmichalikk: [83e84d2](https://github.com/software-mansion/react-native-screens/pull/4544/commits/83e84d283204c6bf4c34efc95ef32c31632e2c11)

src/components/stack/header/StackHeaderConfig.ios.tsx
kmichalikk: [83e84d2](https://github.com/software-mansion/react-native-screens/pull/4544/commits/83e84d283204c6bf4c34efc95ef32c31632e2c11)

ios/stack/host/RNSStackNavigationBarCoordinator.mm
kligarski: Not sure if log visible with `RNS_DEBUG_LOGGING=1` is enough - I think that having other class than `RNSStackNavigationBar` is something we never expect to happen so it might be a good idea to hard assert here.

kmichalikk: [b7b10c6](https://github.com/software-mansion/react-native-screens/pull/4544/commits/b7b10c6bed5e0a30f1e1c9ffa5536b021608b0e8) added a full assertion

PR #4546, opened 2026-08-24
chore(test): e2e test for test-stack-header-item-identifier-ios

LKuchno:
## Description

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1750

Adds a Detox e2e test for the `test-stack-header-item-identifier-ios` scenario (Stack v5, iOS 26). The suite asserts the settled end state of identifier-driven header item matching: which SF Symbol item sits at which trailing position after each push (via horizontal frame order), and — with separators enabled — that each item sits in its own `_UINavigationBarPlatterView`, with an explicit distinct-platter check so a shared-platter regression fails at the platter assertion rather than as an ordering error.

The separators suite follows the scenario's step 6 path: it pushes through the stack first and pops back to screen One before flipping the toggle, so the reconfiguration of already-materialized native items is exercised, not just items configured from birth. The custom-views step (scenario step 4) is left manual-only per review feedback — automating it would only check that the custom views render, not the real identifier-matching functionality. Transition-quality checks (flashing, blur/crossfade) and the identifiers-off step (identical end state) also remain manual-only, as documented in the scenario.

## Changes

- Added `FabricExample/e2e/single-feature-tests/stack-v5/test-stack-header-item-identifier-ios.e2e.ts` gated with `describeIfiOS26`.
- `FabricExample/e2e/e2e-utils.ts` - added the shared `describeIfiOS26`.
- `FabricExample/e2e/native-class-names.ts` - registered `_UINavigationBarPlatterView` (the iOS 26 liquid-glass bar-item bubble).
- `scenario.md` — filled in the E2E coverage section (automated vs manual-only); `scenario-description.ts` — `e2eCoverage: 'tbd'` → `'incomplete'`.

### DIFF
--- a/FabricExample/e2e/e2e-utils.ts
+++ b/FabricExample/e2e/e2e-utils.ts
@@ -53,9 +53,6 @@ export const describeIfiOS26 = isIOSVersionAtLeast('26.0')
   ? describe
   : describe.skip;
 
-export const describeIfiPadIOS26 =
-  isIPadTarget && isIOSVersionAtLeast('26.0') ? describe : describe.skip;
-
 export type ScrollOptions = {
   /** Pixels per step. Smaller steps avoid overshooting a short row. */
   pixels?: number;
--- a/FabricExample/e2e/e2e-utils.ts
+++ b/FabricExample/e2e/e2e-utils.ts
@@ -218,12 +218,16 @@ export async function selectPickerOption(
   await expect(element(by.id(pickerId))).toHaveText(expected);
 }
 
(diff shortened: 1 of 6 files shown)


kmichalikk (approved):
I'm afraid that this more of a "test of the test rather than the feature", especially with the custom views, but on the other hand we haven't really tested dynamically moving items in header, and this does that. We may skip testing the custom views, I think. Other than that, LGTM.

kmichalikk (approved):

PR #4547, opened 2026-08-25
chore(test): manual scenario for test-svm-configures-scroll-view

LKuchno:
## Description

Adds the missing `scenario.md` for `test-svm-configures-scroll-view` and makes the screen usable for manually verifying scroll edge effects: a floating selector at the bottom switches the marker's `scrollEdgeEffects.top` between `automatic` / `hard` / `soft` / `hidden` at runtime, so a tester can compare all values while scrolling.

The per-value visual expectations in the scenario were verified on an iOS 26.5 simulator: `hard` renders a blurred band ending in a sharp dividing line, `soft` and `automatic` render a gradual fade, `hidden` disables the effect entirely. Rectangle colors are now taken once at module scope, so switching the effect no longer recolors the list on re-render (`generateNextColor` advances a global counter on every call).

The scenario is iOS-only (scroll edge effects require iOS 26+); on Android the screen remains a prepared environment for native-side debugging, as in the original PR #3674.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1635

## Changes

- `test-svm-configures-scroll-view/index.tsx` — floating bottom selector (chip row) bound to `scrollEdgeEffects.top`; rectangle colors hoisted to module-scope `RECT_COLORS` for render stability
- `test-svm-configures-scroll-view/scenario.md` — new manual scenario: per-value visual reference and 7 steps covering each effect value, runtime switching, and switch stability
- `test-svm-configures-scroll-view/scenario-description.ts` — details updated (runtime selector, iOS 26+ support note), `platforms` narrowed to `['ios']`, `e2eCoverage` set to `incomplete` (the checks are visual and not automatable with Detox)


kkafar (approved):
Looks good, thank you!

PR #4552, opened 2026-08-26
fix(iOS): restore back chevron after RTL changes

### LINKED ISSUE #4395: [iOS]: Stack back chevron has the wrong direction after switching from RTL back to LTR
### Description

After switching the app from LTR to RTL using I18nManager, and then switching back to LTR, the native stack back button is only partially restored. The back button returns to the correct left-side placement for LTR, but its chevron still points right. In LTR, the chevron should point left.

Tested on both the iOS Simulator and a physical device.
I have also observed the same behavior with earlier versions of Expo and `react-native-screens`.
The issue is resolved only after manually terminating and relaunching the app. Reloading it with Expo's `reloadAppAsync()` does not correct the chevron direction.

### Steps to reproduce

1. Install the dependencies: npm install
2. Build and launch the app on iOS: npx expo run:ios
3. Ensure the app initially uses LTR layout.
4. Open the **Explore** tab.
5. Enable **Right-to-left layout** and wait for the app to reload.
6. Open the **Home** tab.
7. Tap **Open extra screen**.
8. Observe that the back button has the correct placement and chevron direction for RTL.

<details>
<summary>RTL — working correctly</summary>
Image
</details>

9. Use the back button to return to the Home screen.
10. Open the **Explore** tab.
11. Disable **Right-to-left layout** and wait for the app to reload into LTR.
12. Open the **Home** tab.
13. Tap **Open extra screen** again.
14. Observe that the back button is correctly placed on the left, but its chevron incorrectly points right.

<details>
<summary>LTR after switching back — incorrect</summary>
Image
</details>

https://github.com/croshim/expo-57-rtl-issue

### Screens version

4.26.2

### React Native version

0.86.0

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Build type

Release mode

### Device

Real device

### Device model

iPhone 14 Pro (iOS 26) + iOS simulator

### Acknowledgements

Yes

huytdps13400:
## Description

Closes #4395.

After switching from forced RTL back to LTR, UIKit can restore the navigation-item placement while retaining the mirrored back-indicator image. The legacy stack previously relied on `semanticContentAttribute` and scoped `UIAppearance`, which do not reliably refresh the native back indicator after a runtime direction change.

The stack now updates layout direction through UIKit traits:

- iOS 17+: `UINavigationController.traitOverrides.layoutDirection`
- iOS < 17: the parent view controller's `setOverrideTraitCollection:forChildViewController:`, matching the native Tabs implementation

The explicit navigation-controller semantic value remains synchronized on every iOS version because legacy stack gestures and custom transitions consume it directly. Search bars inherit the navigation controller's native RTL behavior without a field-level direction override.

## Changes

- Refresh stack layout direction through UIKit traits.
- Keep the legacy semantic direction synchronized after applying traits so back gestures retain their native RTL direction.
- Let UIKit render the complete search bar in RTL rather than forcing only its text field to LTR.
- Keep semantic updates scoped to the current navigation controller instead of registering global `UIAppearance` rules.

## Reproduction in FabricExample

The existing `Test654` demonstrates the issue:

1. Build and launch `FabricExample` on iOS.
2. Enable **Right to left** on the main screen.
3. Open **Issue Tests**, search for `Test654`, and open it.
4. Tap **Tap me for second screen**; verify the RTL back item and edge gesture match native UIKit behavior.
5. Restart at the main screen, disable **Right to left**, and repeat.
6. Verify the first LTR visit shows a left-side, left-pointing back chevron.

| Before | After |
| --- | --- |
| LTR back chevron incorrectly points right | LTR back chevron correctly points left |

## Verification

- `corepack yarn check-types` on the repository-pinned Node 23.11.0
- `corepack yarn prepare`
- Prior branch verification: full `FabricExample` Release build and manual LTR → RTL → LTR chevron flow on iOS 26.5
- Latest review correction is pushed for the repository iOS build and E2E workflows to validate

## Checklist

- [x] Included reproduction steps using the existing example app.
- [x] Included before/after screenshots.
- [x] No public API change.
- [ ] CI passes on the latest head.

### DIFF
--- a/ios/legacy/RNSScreenStack.h
+++ b/ios/legacy/RNSScreenStack.h
@@ -10,8 +10,15 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-NS_INLINE BOOL RNSNavigationControllerIsRTL(UINavigationController *navigationController)
+/**
+ * Returns whether the navigation controller currently uses right-to-left layout.
+ */
+NS_INLINE BOOL RNSNavigationControllerIsRTL(UINavigationController *_Nullable navigationController)
 {
+  if (navigationController == nil) {
+    return NO;
+  }
+
   UISemanticContentAttribute semanticContentAttribute = navigationController.view.semanticContentAttribute;
   if (semanticContentAttribute == UISemanticContentAttributeForceRightToLeft) {
     return YES;
--- a/ios/legacy/RNSScreenStack.h
+++ b/ios/legacy/RNSScreenStack.h
@@ -10,6 +10,18 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+NS_INLINE BOOL RNSNavigationControllerIsRTL(UINavigationController *navigationController)
+{
+  UISemanticContentAttribute semanticContentAttribute = navigationController.view.semanticContentAttribute;
+  if (semanticContentAttribute == UISemanticContentAttributeForceRightToLeft) {
+    return YES;
+  }
+  if (semanticContentAttribute == UISemanticContentAttributeForceLeftToRight) {
+    return NO;
+  }
+  return navigationController.traitCollection.layoutDirection == UITraitEnvironmentLayoutDirectionRightToLeft;
+}
+
 @interface RNSNavigationController : UINavigationController <RNSViewControllerDelegate,
                                                              RNSTabsSpecialEffectsSupporting
 #if !TARGET_OS_TV
--- a/ios/legacy/RNSScreenStack.mm
+++ b/ios/legacy/RNSScreenStack.mm
@@ -823,10 +823,9 @@ RNS_IGNORE_SUPER_CALL_END
   if (customAnimationOnSwipePropSetAndSelectedAnimationIsCustom) {
     if ([gestureRecognizer isKindOfClass:[RNSScreenEdgeGestureRecognizer class]]) {
       UIRectEdge edges = ((RNSScreenEdgeGestureRecognizer *)gestureRecognizer).edges;
-      BOOL isRTL = _controller.view.semanticContentAttribute == UISemanticContentAttributeForceRightToLeft;
+      BOOL isRTL = RNSNavigationControllerIsRTL(_controller);
       BOOL isSlideFromLeft = topScreen.stackAnimation == RNSScreenStackAnimationSlideFromLeft;
-      // if we do not set any explicit `semanticContentAttribute`, it is `UISemanticContentAttributeUnspecified` instead
-      // of `UISemanticContentAttributeForceLeftToRight`, so we just check if it is RTL or not
+      // If no explicit semantic fallback is set, the navigation controller's trait collection determines direction.
       BOOL isCorrectEdge = (isRTL && edges == UIRectEdgeRight) ||
           (!isRTL && isSlideFromLeft && edges == UIRectEdgeRight) ||
           (isRTL && isSlideFromLeft && edges == UIRectEdgeLeft) || (!isRTL && edges == UIRectEdgeLeft);
@@ -897,7 +896,7 @@ RNS_IGNORE_SUPER_CALL_END
     translation = [gestureRecognizer translationInView:gestureRecognizer.view].x;
     velocity = [gestureRecognizer velocityInView:gestureRecognizer.view].x;
     distance = gestureRecognizer.view.bounds.size.width;
-    BOOL isRTL = _controller.view.semanticContentAttribute == UISemanticContentAttributeForceRightToLeft;
+    BOOL isRTL = RNSNavigationControllerIsRTL(_controller);
     if (isRTL) {
       translation = -translation;
       velocity = -velocity;
@@ -1018,7 +1017,7 @@ RNS_IGNORE_SUPER_CALL_END
   NSDictionary *gestureResponseDistanceValues = topScreen.gestureResponseDistance;
   float x = [gestureRecognizer locationInView:gestureRecognizer.view].x;
   float y = [gestureRecognizer locationInView:gestureRecognizer.view].y;
-  BOOL isRTL = _controller.view.semanticContentAttribute == UISemanticContentAttributeForceRightToLeft;
+  BOOL isRTL = RNSNavigationControllerIsRTL(_controller);
   if (isRTL) {
     x = _controller.view.frame.size.width - x;
   }
--- a/ios/legacy/RNSScreenStackHeaderConfig.mm
+++ b/ios/legacy/RNSScreenStackHeaderConfig.mm
@@ -797,12 +797,10 @@ RNS_IGNORE_SUPER_CALL_END
     [parentViewController
         setOverrideTraitCollection:[UITraitCollection traitCollectionWithLayoutDirection:layoutDirection]
             forChildViewController:navCtrl];
-    return;
   }
 
-  // A detached navigation controller has no parent that can override its traits yet.
-  // Preserve the legacy fallback until the next configuration pass after attachment.
-  if (parentViewController == nil && navCtrl.view.semanticContentAttribute != self.direction) {
+  // Keep the explicit legacy value synchronized because stack gestures and animations inspect it before traits.
+  if (navCtrl.view.semanticContentAttribute != self.direction) {
     navCtrl.view.semanticContentAttribute = self.direction;
     navCtrl.navigationBar.semanticContentAttribute = self.direction;
     [[UIButton appearanceWhenContainedInInstancesOfClasses:@[ navCtrl.navigationBar.class ]]
--- a/ios/legacy/RNSScreenStackHeaderConfig.mm
+++ b/ios/legacy/RNSScreenStackHeaderConfig.mm
@@ -603,6 +603,7 @@ RNS_IGNORE_SUPER_CALL_END
 #if !TARGET_OS_TV
           RNSSearchBar *searchBar = subview.subviews[0];
           searchBarPresent = true;
+          searchBar.controller.searchBar.semanticContentAttribute = config.direction;
           navitem.searchController = searchBar.controller;
           navitem.hidesSearchBarWhenScrolling = searchBar.hideWhenScrolling;
 #if RNS_IPHONE_OS_VERSION_AVAILABLE(16_0)
@@ -778,7 +779,9 @@ RNS_IGNORE_SUPER_CALL_END
     if (navCtrl.traitCollection.layoutDirection != layoutDirection) {
       navCtrl.traitOverrides.layoutDirection = layoutDirection;
     }
-  } else {
+  } else
+#endif // RNS_IPHONE_OS_VERSION_AVAILABLE(17_0)
+  {
     UIViewController *parentViewController = navCtrl.parentViewController;
     if (parentViewController != nil && navCtrl.traitCollection.layoutDirection != layoutDirection) {
       [parentViewController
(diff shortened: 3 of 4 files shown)


kligarski (commented):
Hi, @huytdps13400. Thank you for the PR.

1. Do you think it would be possible to also include a fix for iOS < 17, in a similar way how it's done in Tabs via `- (void)setOverrideTraitCollection:(nullable UITraitCollection *)collection forChildViewController:(UIViewController *)childViewController`?

2. Also, it would be nice to include reproduction steps that are possible to do in our example app in the PR description. We can use already existing `Test654` to showcase the bug:

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/da2c8b1a-51fb-4aca-90ac-c34a1dd623ea" /> | <video src="https://github.com/user-attachments/assets/c404e17e-f2c7-4571-8e67-c9c5a7c5de7a" />

3. In the "after" video above with the fix from this PR, after changing to RTL, the search bar's icon in `issue-test` screen is missing which is not the case in the "before" video. 

| before | after |
| --- | --- |
| Image | Image |

This is a regression and we need to investigate and fix it before proceeding further with the PR.

huytdps13400:
Thanks for the detailed review. I addressed all three points in 8fbe2ecf:

1. iOS < 17 now uses the parent view controller's `setOverrideTraitCollection:forChildViewController:`, following the Tabs implementation.
2. I updated the PR description with a complete FabricExample / `Test654` reproduction flow.
3. I reproduced the missing search icon on iOS 26.5 and traced it to the navigation-controller trait override changing the inherited search-field layout. The search field now preserves its existing LTR content layout when the RTL navigation override is active, so the default magnifying-glass icon remains visible.

I also updated stack gesture/custom-animation RTL detection to use the effective navigation-controller trait direction, with the legacy semantic value retained as a fallback.

Verification:

- `corepack yarn check-types` ✅
- `RNScreens` native target build ✅
- full FabricExample Release build ✅
- manual iPhone 17 Pro / iOS 26.5 flow ✅: search icon visible before and during editing; `Test654` second screen shows the RTL back item and chevron; no crash after changing direction

Detox itself could not start because this machine does not have `applesimutils`; I ran the same Release-app flow manually instead. The repository-wide Jest command also collects unrelated React Navigation submodule generated/Playwright tests (43 suites / 632 tests pass, 20 configuration/submodule suites fail), so I documented that limitation in the PR description.

ios/legacy/RNSScreenStack.h
```diff
@@ -10,6 +10,18 @@
...
+{
+  UISemanticContentAttribute semanticContentAttribute = navigationController.view.semanticContentAttribute;
+  if (semanticContentAttribute == UISemanticContentAttributeForceRightToLeft) {
+    return YES;
+  }
+  if (semanticContentAttribute == UISemanticContentAttributeForceLeftToRight) {
+    return NO;
+  }
+  return navigationController.traitCollection.layoutDirection == UITraitEnvironmentLayoutDirectionRightToLeft;
+}
```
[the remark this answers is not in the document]
huytdps13400: Fixed in b123464b. The helper now accepts a nullable navigation controller and returns `NO` before dereferencing `nil`. The updated native target compiles successfully, and `yarn check-types` passes.

huytdps13400:
Addressed the iOS < 17 stale-direction case in b123464b.

The attached-controller path no longer returns immediately after applying the parent trait override. It now also synchronizes the explicit legacy semantic value and navigation-bar appearance state, so `RNSNavigationControllerIsRTL` cannot prefer an old detached-controller direction during gestures or custom transitions.

I also made the RTL helper explicitly nullable and documented its behavior.

Verification:

- `RNScreens` CocoaPods target build (Debug, iOS Simulator) ✅
- `corepack yarn check-types` ✅
- `git diff --check` ✅

huytdps13400:
Addressed the cross-stack appearance leak in 54fd2d45.

I checked the history of these rules: they were introduced in #2185, whose description already notes that the `UIAppearance` proxy can override another native navigation bar. With the new per-controller trait override on iOS < 17, those global rules are no longer needed.

The legacy path now updates only the current `navCtrl.view` and `navCtrl.navigationBar`. Child layout is inherited from that instance plus the parent trait override, so one stack cannot register direction rules for controls created later in another stack.

Verification:

- `RNScreens` CocoaPods target build (Debug, iOS Simulator) ✅
- `corepack yarn check-types` ✅
- no `appearanceWhenContainedInInstancesOfClasses` usage remains in this header path
- `git diff --check` ✅

ios/legacy/RNSScreenStackHeaderConfig.mm
```diff
@@ -604,6 +604,16 @@ + (void)updateViewController:(UIViewController *)vc
           RNSSearchBar *searchBar = subview.subviews[0];
           searchBarPresent = true;
           navitem.searchController = searchBar.controller;
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(17_0)
+          if (@available(iOS 17.0, *)) {
+            if (config.direction == UISemanticContentAttributeForceRightToLeft) {
+              // Overriding the navigation controller's layout direction makes UIKit drop the default search icon.
+              // Keep the search field's existing left-to-right content layout while the navigation bar is mirrored.
+              searchBar.controller.searchBar.searchTextField.semanticContentAttribute =
```
kligarski: Forcing the search bar to always appear in LTR is not a proper fix. In native iOS apps, it is in RTL and we must not change this behavior.

| screens after current fix | native RTL |
| --- | --- |
| Image | Image |

ios/legacy/RNSScreenStack.mm
kligarski, line 1: Go back gesture direction is incorrect.

| screens after current fix | native RTL |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/17d0a242-f146-4fb0-9c10-337850d8e692" /> | <video src="https://github.com/user-attachments/assets/8352b4c5-55d6-4e40-b20f-08f2d256ca21" /> |

kligarski (changes requested):

ios/legacy/RNSScreenStackHeaderConfig.mm
huytdps13400: Fixed in 209c9ada. I removed the searchTextField LTR override entirely, so the search bar now inherits the navigation controller trait and matches native UIKit RTL behavior.

ios/legacy/RNSScreenStack.mm
huytdps13400: Fixed in 209c9ada. The iOS 17 trait-override branch returned before synchronizing navCtrl.view.semanticContentAttribute, which the legacy gesture path still reads. The trait override now falls through to the scoped semantic update on every iOS version, while the parent trait override remains limited to pre-iOS 17.

huytdps13400:
@kligarski Addressed both review findings in `209c9ada`: the search field is no longer forced LTR, and the iOS 17 path now keeps the scoped navigation-controller semantic direction synchronized after applying `traitOverrides`, restoring the legacy back-gesture direction without reintroducing global appearance rules. `yarn check-types` and `yarn prepare` pass on Node 23.11.0; the latest iOS CI/E2E rerun is pending.

ios/legacy/RNSScreenStackHeaderConfig.mm
```diff
@@ -764,21 +764,41 @@ - (void)configureBackItem:(nullable UINavigationItem *)prevItem
...
+    }
+  }
+#else
+  UIViewController *parentViewController = navCtrl.parentViewController;
+  if (parentViewController != nil && navCtrl.traitCollection.layoutDirection != layoutDirection) {
+    [parentViewController
+        setOverrideTraitCollection:[UITraitCollection traitCollectionWithLayoutDirection:layoutDirection]
+            forChildViewController:navCtrl];
+  }
+#endif // RNS_IPHONE_OS_VERSION_AVAILABLE(17_0)
```
kligarski: nit: this will avoid duplication

```suggestion
#if RNS_IPHONE_OS_VERSION_AVAILABLE(17_0)
  if (@available(iOS 17.0, *)) {
    if (navCtrl.traitCollection.layoutDirection != layoutDirection) {
      navCtrl.traitOverrides.layoutDirection = layoutDirection;
    }
  } else
#endif // RNS_IPHONE_OS_VERSION_AVAILABLE(17_0)
  {
    UIViewController *parentViewController = navCtrl.parentViewController;
    if (parentViewController != nil && navCtrl.traitCollection.layoutDirection != layoutDirection) {
      [parentViewController
          setOverrideTraitCollection:[UITraitCollection traitCollectionWithLayoutDirection:layoutDirection]
              forChildViewController:navCtrl];
    }
  }
```

ios/legacy/RNSScreenStackHeaderConfig.mm
kligarski: Unfortunately we're back to broken search bar icon in RTL.

obraz

ios/legacy/RNSScreenStackHeaderConfig.mm
```diff
@@ -764,21 +764,41 @@ - (void)configureBackItem:(nullable UINavigationItem *)prevItem
...
+
+  // Keep the explicit legacy value synchronized because stack gestures and animations inspect it directly.
+  if (navCtrl.view.semanticContentAttribute != self.direction) {
+    navCtrl.view.semanticContentAttribute = self.direction;
     navCtrl.navigationBar.semanticContentAttribute = self.direction;
```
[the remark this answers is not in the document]
huytdps13400: Fixed in f9d7224f. The navigation controller view and navigation bar are now compared and updated independently, so a matching view can no longer suppress a stale navigation-bar direction update.

ios/legacy/RNSScreenStackHeaderConfig.mm
huytdps13400: Applied in f9d7224f. The availability branch now uses the suggested preprocessor shape, keeping the pre-iOS 17 fallback in one block.

ios/legacy/RNSScreenStackHeaderConfig.mm
huytdps13400: Fixed in f9d7224f without forcing the text field LTR. The complete UISearchBar now receives config.direction before it is attached to the navigation item, so UIKit creates and lays out the native RTL search field consistently. I verified this in an isolated UIKit harness on iOS 26.5 with the same navigation-controller trait override: the Arabic placeholder remains RTL and the magnifying-glass icon is present on the right. RTL search bar with right-side search icon

huytdps13400:
@kligarski Follow-up `f9d7224f` addresses the latest review: it applies the deduplicated availability branch, synchronizes the view and navigation bar independently, and sets the whole `UISearchBar` to the requested direction before attachment rather than forcing only the text field LTR. A minimal UIKit harness on iOS 26.5 confirmed native RTL placeholder layout with the magnifying-glass icon visible on the right; screenshot is attached in the review thread. Typecheck and package build pass on Node 23.11.0.

ios/legacy/RNSScreenStackHeaderConfig.mm
```diff
@@ -764,21 +765,37 @@ - (void)configureBackItem:(nullable UINavigationItem *)prevItem
...
+      self.direction != UISemanticContentAttributeForceRightToLeft) {
+    return;
+  }
+
+  UITraitEnvironmentLayoutDirection layoutDirection = self.direction == UISemanticContentAttributeForceRightToLeft
+      ? UITraitEnvironmentLayoutDirectionRightToLeft
+      : UITraitEnvironmentLayoutDirectionLeftToRight;
+
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(17_0)
+  if (@available(iOS 17.0, *)) {
```
kligarski, line 778:
```suggestion
  if (@available(iOS 17.0, tvOS 17.0, *)) {
```

ios/legacy/RNSScreenStackHeaderConfig.mm
```diff
@@ -764,21 +765,37 @@ - (void)configureBackItem:(nullable UINavigationItem *)prevItem
...
-  if ((self.direction == UISemanticContentAttributeForceLeftToRight ||
-       self.direction == UISemanticContentAttributeForceRightToLeft) &&
-      // iOS 12 cancels swipe gesture when direction is changed. See #1091
-      navCtrl.view.semanticContentAttribute != self.direction) {
-    // This is needed for swipe back gesture direction
-    navCtrl.view.semanticContentAttribute = self.direction;
+  if (self.direction != UISemanticContentAttributeForceLeftToRight &&
+      self.direction != UISemanticContentAttributeForceRightToLeft) {
+    return;
+  }
```
kligarski, line 771: I checked with the debugger and it seems like due to how codegen and prop diffing works, even though codegen spec specifies `'ltr'` as default value for the prop, it remains `UISemanticContentAttributeUnspecified` therefore we early return. We should initialize `  _direction = UISemanticContentAttributeForceLeftToRight;` in `initProps` and make sure nothing breaks.

kligarski (commented):
I left 2 more suggestions but I haven't observed any problems with runtime now so we're close to finishing the PR. Thanks.

PR #4555, opened 2026-08-26
fix(Android, Stack v4+): transition RecyclerView as a unit on screen removal

drhops:
**Note: this fix was discovered via https://github.com/callstack/react-native-pager-view/issues/1005#issuecomment-5095233311 and all credit goes to @mozzius**

## Description

When a screen is removed, `Screen.startTransitionRecursive` calls `startViewTransition` on every descendant so the outgoing screen keeps drawing through the exit animation. That works by delaying `ViewGroup`'s detachment contract and the transitioning view keeps its parent when it is removed:

```java
// ViewGroup.removeFromArray
if (!(mTransitioningViews != null && mTransitioningViews.contains(children[index]))) {
    children[index].mParent = null;      // skipped while transitioning
}
```

`RecyclerView` owns the attach/detach lifecycle of its children, and rejects any holder whose parent is still set:

```java
// RecyclerView.Recycler.recycleViewHolderInternal
if (holder.isScrap() || holder.itemView.getParent() != null) {
    throw new IllegalArgumentException("Scrapped or attached views may not be recycled. ...");
}
```

So once a screen removal has marked a `ViewPager2` page, the next recycle throws:

```
java.lang.IllegalArgumentException: Scrapped or attached views may not be recycled.
  isScrap:false isAttached:true androidx.viewpager2.widget.ViewPager2$RecyclerViewImpl{...},
  adapter:com.reactnativepagerview.ViewPagerAdapter@...

  at ...RecyclerView$Recycler.recycleViewHolderInternal(RecyclerView.java:7071)
  at ...RecyclerView$LayoutManager.removeAndRecycleViewAt(RecyclerView.java:9741)
  at ...LinearLayoutManager.recycleChildren(LinearLayoutManager.java:1464)
  at ...LinearLayoutManager.fill(LinearLayoutManager.java:1613)
  at ...RecyclerView.scrollStep(RecyclerView.java:2047)
  at ...RecyclerView$ViewFlinger.run(RecyclerView.java:5835)
```

## Repro Steps:

Setup: a RecyclerView inside of a native stack:

1. A native stack. ScreenStackViewManager.removeViewAt is the only caller of startRemovalTransition()
2. A RecyclerView anywhere in the popped screen's subtree. Depth is irrelevant and `<PagerView>` is the example in our app.
3. ≥3 pages, so a page actually leaves the retained window and recycleChildren runs.

Minimal test screen

```
import React from 'react';
import { Button, Text, View } from 'react-native';
import PagerView from 'react-native-pager-view';

  return <Button title="Open pager" onPress={() => navigation.navigate('Pager')} />;
}

function Pager() {
  return (
    <PagerView style={{ flex: 1 }} initialPage={0}>
      {Array.from({ length: 6 }, (_, i) => (
        <View key={i} style={{ flex: 1, backgroundColor: i % 2 ? '#cde' : '#edc' }}>
          <Text style={{ fontSize: 64, textAlign: 'center', marginTop: 240 }}>{i}</Text>
        </View>
      ))}
    </PagerView>
  );
}

export default function Test2461() {
  return (
    <NavigationContainer independent>
      <Stack.Navigator>
        <Stack.Screen name="Pager" component={Pager} />
      </Stack.Navigator>
    </NavigationContainer>
  );
}

```

On Android: open the pager screen, drag horizontally and release mid-page, then immediately press Back while it's still gliding. Use a back button rather than back gesture. Repeat a couple times given the animation settle window is only a few hundred milliseconds.

### Notes

Fixes the crash reported in #2461 (closed without a fix).
Downstream: callstack/react-native-pager-view#1005.

## Changes

- `legacy/Screen.kt`: skip recursion into a `RecyclerView` in `startTransitionRecursive`, and mirror
  the skip in `endTransitionRecursive`.
- `android/build.gradle`: declare `androidx.recyclerview:recyclerview` explicitly. It was already on
  the compile classpath transitively via Material, but the new import should not rely on that. Pinned
  at `1.1.0`, which is what Material already resolves.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -221,6 +221,7 @@ dependencies {
     implementation 'androidx.fragment:fragment-ktx:1.8.9'
     implementation 'androidx.transition:transition-ktx:1.7.0'
     implementation 'androidx.coordinatorlayout:coordinatorlayout:1.2.0'
+    implementation 'androidx.recyclerview:recyclerview:1.1.0'
     implementation 'androidx.swiperefreshlayout:swiperefreshlayout:1.1.0'
     implementation 'com.google.android.material:material:1.14.0'
     implementation "androidx.core:core-ktx:1.17.0"
--- a/android/src/main/java/com/swmansion/rnscreens/legacy/Screen.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/legacy/Screen.kt
@@ -16,6 +16,7 @@ import androidx.annotation.RequiresApi
 import androidx.coordinatorlayout.widget.CoordinatorLayout
 import androidx.core.view.children
 import androidx.fragment.app.Fragment
+import androidx.recyclerview.widget.RecyclerView
 import androidx.swiperefreshlayout.widget.SwipeRefreshLayout
 import com.facebook.react.bridge.ReactContext
 import com.facebook.react.uimanager.PixelUtil
@@ -478,7 +479,8 @@ class Screen(
                 endTransitionRecursive(childView.toolbar)
             }
 
-            if (childView is ViewGroup) {
+            // Mirrors the skip in `startTransitionRecursive`.
+            if (childView is ViewGroup && childView !is RecyclerView) {
                 endTransitionRecursive(childView)
             }
         }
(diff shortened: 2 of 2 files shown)


android/build.gradle
```diff
@@ -221,6 +221,7 @@ dependencies {
     implementation 'androidx.fragment:fragment-ktx:1.8.9'
     implementation 'androidx.transition:transition-ktx:1.7.0'
     implementation 'androidx.coordinatorlayout:coordinatorlayout:1.2.0'
+    implementation 'androidx.recyclerview:recyclerview:1.1.0'
```
kkafar, line 224: This is less than perfect and what actually stops me from landing this change. 

Additional dependency just for a symbol is a lot. 

What's the impact of this on package size? 
Also why outdated 1.1.0, when newer versions are available?

drhops: @kkafar this dependency is already included transitively, so this change just makes it direct/explicit given we now rely on it in the code.  1.1.0 because that's what was already being included transitively, and it shouldn't affect package size at all.

PR #4525, opened 2026-08-20
chore(test): mark stack overflow icon e2e coverage as not automated

LKuchno:
## Description

This PR documents that the `test-stack-overflow-icon-android` scenario cannot be automated with Detox. The scenario's checkpoints verify the overflow button's rendered icon and tint colors, which Detox cannot assert: `getAttributes()` exposes no color/drawable data and the e2e suite does no screenshot comparison. Additionally, the pressed and focused tints would require mid-press observation and hardware-keyboard focus, neither of which Detox can drive. The scenario is marked as manual-only.

The test screen is also migrated from the `Component:` route config to the `element:` form, following #4495.

## Changes

- **scenario.md**: Replace the "TBD" e2e section with "Incomplete: Not automated." plus an explanation of the Detox limitations
- **scenario-description.ts**: Set `e2eCoverage` from `'tbd'` to `'incomplete'`
- **index.tsx**: Migrate `StackContainer` route config from `Component: MainScreen` to `element: <MainScreen />`

PR #4556, opened 2026-08-26
feat(Android, Stack v5): support for host-level color scheme configuration

kligarski:
## Description

Adds support for configuring host-level color scheme of the Stack. Color scheme propagation works with Tabs as well.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1748.

#### Source of color scheme configuration

In Tabs, the tab bar is a part of the container on both platforms. That's why color scheme configuration is exposed only on the Host component. 

For Stack, it's different between the platforms. On iOS, `UINavigationBar` (and `UIToolbar`) are also part of the `UINavigationController` but on Android each screen has its own independent `AppBarLayout`. There are no views which are a part of the container that would require adaptation to color scheme.

We decided for now to expose the same configuration (host-level) for Android implementation, same as Tabs for API consistency. All screens inside the `StackContainer` will use the theme resolved by the container. In the future, we plan to expose per-screen/per-header-config configuration as well (https://github.com/software-mansion/react-native-screens-labs/issues/1749). 

Something that might be worth considering is whether iOS implementation of Tabs and Stack might also use screen-level configuration - it shouldn't impact shared views like navigation bar or tab bar but on iOS this might influence e.g. keyboard appearance.

#### `StackScreenFragment` <-> `StackContainer` communication

When appearance is modified by `react-native`'s `Appearance` module, only fragments are informed about configuration change; views don't receive it (more details: https://github.com/software-mansion/react-native-screens/pull/3167). That's why `StackScreenFragment`s need to inform `StackContainer` about a change (so that we're able to resolve correct system theme).

In Tabs, this information goes from `TabsScreenFragment` (native impl) to `TabsContainer` (native impl) via `TabsScreen` (React view). This isn't ideal because we're relying on React view to pass information that can belong fully on the native side.

image

For Stack, we decided to keep it native-only by introducing `StackScreenFragmentDelegate` interface that will allow fragments to inform the delegate (`StackContainer`) about configuration change directly. Tabs implementation should be refactored in a similar way in the future.

#### `colorSchemeCoordinator` in `StackHeaderCoordinatorLayout`

Even though we currently don't expose per-screen/per-header-config configuration of the color scheme, I decided to re-use  `colorSchemeCoordinator` in `StackHeaderCoordinatorLayout`. By using it, we get color scheme propagation "for free" by making `StackHeaderCoordinatorLayout` `ColorSchemeProviding`. We modify the `wrappedContext`'s theme according to the coordinator. It will also make exposing screen-level config easier in the future. One minor detail: we don't need to receive `onConfigurationChange` calls from views not fragments because container-level coordinator will receive them and propagate the change to screens. System's color scheme is always resolved by top coordinator in the hierarchy so stale `systemUiNightMode` is not a problem.

#### Updating the color scheme of the header

Unfortunately, to fully adapt to new color scheme, we need to recreate the `MaterialToolbar` (which we can do by rebuilding the entire header). Otherwise, toolbar items and back button icon/overflow icon in the toolbar don't update their font and ripple color + the overflow menu remains in previous color scheme. This happens due to how `MaterialToolbar`'s internals copy the theme and don't fully react to theme change in the `wrappedContext`.

This however causes 2 problems:

1. Rebuilding the header expands the collapsing header. We already track if the header is fully collapsed (see https://github.com/software-mansion/react-native-screens/pull/4436) so we can use it to restore collapsed state. Currently, I added this only to the theme change. I think we can consider applying this fix on every rebuild but this requires some investigation - when scroll flags are changed, we want to reset the state and currently we do so by expanding the header unconditionally. Here's the internal ticket to track this: https://github.com/software-mansion/react-native-screens-labs/issues/1192.

2. Rebuilding the header clears the toolbar state which includes user's radio/checkbox selection and any updates applied via view commands. This is unacceptable but there doesn't seem to be a simple way to prevent this - we can't re-apply the menu from previous toolbar. This will require additional serialization/deserialization process - here's the internal ticket to track this: https://github.com/software-mansion/react-native-screens-labs/issues/1755.

#### Native-originating header update

In order to force header rebuild, for now I added `forcedFlags` which allow to bypass invalidation flags that live on the provider. But this isn't a clean solution - if in the future prop update causes a native color scheme update, we will apply the update twice - from color scheme coordinator's callback and then after React transaction - this is unnecessary. Also, while testing I found this related bug: https://github.com/software-mansion/react-native-screens-labs/issues/1756 - when Stack is nested in Tabs, if we change the tab and go back, header is not reconstructed. This happens because tab fragment is detached which results in view hierarchy being destroyed. When the view is recreated, invalidation flags on the provider remain empty which skips the entire header update. This should be handled in follow up PR. See the ticket for more details - I am considering refactoring the invalidation flags by moving them to the native implementation. I'm open for feedback here.

## Changes

- handle `colorScheme` prop on `StackHost`/`StackContainer` by adding `colorSchemeCoordinator` and implementing `ColorSchemeProviding`
- add `StackScreenFragmentDelegate` to inform `StackContainer` about configuration change in one of the fragments
- resolve text appearance from `wrappedContext` to handle correct color scheme
- resolve default icons from toolbar's context to work around cache bug
- add SFT

https://github.com/user-attachments/assets/0896a236-75ef-4035-9a86-19549929c7b6

## Test plan

Run `test-stack-color-scheme` SFT.

## Checklist

- [x] Ensured that CI passes

---

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/stack/header/StackHeaderCoordinatorLayout.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/stack/header/StackHeaderCoordinatorLayout.kt
@@ -338,7 +338,11 @@ internal class StackHeaderCoordinatorLayout(
                 else -> R.style.Theme_Material3Expressive_DayNight_NoActionBar
             },
         )
-        if (uiNightMode == appliedUiNightMode) return
+
+        if (uiNightMode == appliedUiNightMode) {
+            return
+        }
+
         appliedUiNightMode = uiNightMode
         currentProvider?.let {
             // A rebuild is forced because MaterialToolbar snapshots its theme at construction:
--- a/android/src/main/java/com/swmansion/rnscreens/stack/header/StackHeaderCoordinatorLayout.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/stack/header/StackHeaderCoordinatorLayout.kt
@@ -317,6 +317,9 @@ internal class StackHeaderCoordinatorLayout(
 
     override fun removeColorSchemeListener(listener: ColorSchemeListener) = colorSchemeCoordinator.removeColorSchemeListener(listener)
 
+    // No onConfigurationChanged override is needed: this view never sets its own colorScheme,
+    // so resolution always delegates to the parent provider, which does handle system changes.
+
     override fun onAttachedToWindow() {
         super.onAttachedToWindow()
         colorSchemeCoordinator.setup(this) { applyUiNightMode(it) }
@@ -455,6 +458,8 @@ internal class StackHeaderCoordinatorLayout(
     // region Teardown
 
     internal fun tearDown() {
+        colorSchemeCoordinator.teardown()
+
         removeHeader()
 
         stackScreenWrapper.removeView(stackScreen)
--- a/android/src/main/java/com/swmansion/rnscreens/stack/header/StackHeaderCoordinatorLayout.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/stack/header/StackHeaderCoordinatorLayout.kt
@@ -2,6 +2,7 @@ package com.swmansion.rnscreens.stack.header
 
 import android.annotation.SuppressLint
 import android.content.Context
+import android.content.res.Configuration
 import android.os.Parcelable
 import android.util.Log
 import android.util.SparseArray
@@ -14,6 +15,9 @@ import com.facebook.react.bridge.ReactContext
 import com.google.android.material.R
 import com.google.android.material.appbar.AppBarLayout
 import com.google.android.material.appbar.MaterialToolbar
+import com.swmansion.rnscreens.common.colorscheme.ColorSchemeCoordinator
+import com.swmansion.rnscreens.common.colorscheme.ColorSchemeListener
+import com.swmansion.rnscreens.common.colorscheme.ColorSchemeProviding
 import com.swmansion.rnscreens.stack.header.appbar.StackHeaderAppBarLayout
 import com.swmansion.rnscreens.stack.header.appbar.StackHeaderScrollingViewBehavior
 import com.swmansion.rnscreens.stack.header.config.OnHeaderConfigurationAttachListener
@@ -36,7 +40,8 @@ internal class StackHeaderCoordinatorLayout(
     context: Context,
     internal val stackScreen: StackScreen,
     private val canNavigateBack: Boolean,
-) : CoordinatorLayout(context) {
+) : CoordinatorLayout(context),
+    ColorSchemeProviding {
     // region Config attach / detach
 
     private var currentProvider: StackHeaderConfigurationProviding? = null
@@ -173,8 +178,12 @@ internal class StackHeaderCoordinatorLayout(
         activity?.onBackPressedDispatcher?.onBackPressed()
     }
 
-    private fun processUpdate(provider: StackHeaderConfigurationProviding) {
-        val needsRebuild = provider.invalidationFlags.needsRebuild
+    private fun processUpdate(
+        provider: StackHeaderConfigurationProviding,
+        forcedFlags: StackHeaderInvalidationFlags = StackHeaderInvalidationFlags.NONE,
+    ) {
+        val effectiveFlags = provider.invalidationFlags or forcedFlags
+        val needsRebuild = effectiveFlags.needsRebuild
         if (needsRebuild) {
             resetHeader()
             if (provider.hidden) {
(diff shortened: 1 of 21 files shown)


apps/src/tests/single-feature-tests/stack-v5/test-stack-color-scheme/index.tsx
```diff
@@ -0,0 +1,341 @@
...
+
+  const [headerType, setHeaderType] = useState<StackHeaderTypeAndroid>('small');
+  const [lastSelection, setLastSelection] = useState<string | null>(null);
+
+  const headerConfigRef = useRef<StackHeaderConfigRef>(null);
+  const iconSeedRef = useRef(0);
+
+  const handleGroupChange = useCallback(
+    (groupId: string, selectedIds: string[]) => {
+      setLastSelection(`${groupId}: ${JSON.stringify(selectedIds)}`);
```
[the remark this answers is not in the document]
kligarski: https://github.com/software-mansion/react-native-screens/pull/4556/commits/3d58afa6835e056f66e88295874672e55aa87854

apps/src/tests/single-feature-tests/stack-v5/test-stack-color-scheme/index.tsx
```diff
@@ -0,0 +1,341 @@
...
+
+function TestStackColorScheme() {
+  const [hostColorScheme, setHostColorScheme] =
+    useState<StackHostColorScheme>('inherit');
+  const [reactColorScheme, setReactColorScheme] =
+    useState<Appearance.ColorSchemeOverride>('auto');
+
+  useEffect(() => {
+    Appearance.setColorScheme(reactColorScheme);
+  }, [reactColorScheme]);
```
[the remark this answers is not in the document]
kligarski: https://github.com/software-mansion/react-native-screens/pull/4556/commits/696382a8032c9563fa116e2a2998dc709be3c1ab

android/src/main/java/com/swmansion/rnscreens/stack/header/StackHeaderCoordinatorLayout.kt
```diff
@@ -292,6 +301,61 @@ internal class StackHeaderCoordinatorLayout(
...
+
+    private fun applyUiNightMode(uiNightMode: Int) {
+        wrappedContext.setTheme(
+            when (uiNightMode) {
+                Configuration.UI_MODE_NIGHT_YES -> R.style.Theme_Material3Expressive_Dark_NoActionBar
+                Configuration.UI_MODE_NIGHT_NO -> R.style.Theme_Material3Expressive_Light_NoActionBar
+                else -> R.style.Theme_Material3Expressive_DayNight_NoActionBar
+            },
+        )
+        if (uiNightMode == appliedUiNightMode) return
```
kmichalikk: nit: let's keep parentheses - it's more readable IMO

android/src/main/java/com/swmansion/rnscreens/stack/header/StackHeaderCoordinatorLayout.kt
```diff
@@ -173,8 +178,12 @@ internal class StackHeaderCoordinatorLayout(
         activity?.onBackPressedDispatcher?.onBackPressed()
     }
 
-    private fun processUpdate(provider: StackHeaderConfigurationProviding) {
-        val needsRebuild = provider.invalidationFlags.needsRebuild
+    private fun processUpdate(
+        provider: StackHeaderConfigurationProviding,
+        forcedFlags: StackHeaderInvalidationFlags = StackHeaderInvalidationFlags.NONE,
+    ) {
+        val effectiveFlags = provider.invalidationFlags or forcedFlags
```
kmichalikk, line 185: I'm not sure about naming it "forced" when it's only the fallback if `invalidationFlags` is falsy

android/src/main/java/com/swmansion/rnscreens/stack/screen/StackScreenFragmentDelegate.kt
```diff
@@ -0,0 +1,12 @@
+package com.swmansion.rnscreens.stack.screen
+
+import android.content.res.Configuration
+
+internal interface StackScreenFragmentDelegate {
```
kmichalikk, line 5: Let's have a more descriptive name for the interface, could even be StackScreenFragmentConfigurationUpdateDelegate

also nit, why Delegate and not Listener? I think that delegate is more tied to the calling object, as if the caller calls the delegate in its own interest, whereas Listener doesn't have these assumptions

apps/src/tests/single-feature-tests/stack-v5/test-stack-color-scheme/scenario.md
```diff
@@ -0,0 +1,181 @@
...
+
+**OS test creation version:** Android: API Level 37.
+
+<!-- TODO: add iOS versions-->
+
+## E2E test
+
+Incomplete: Not automated. Detox does not have access to color attributes
+natively, making it impossible to reliably verify if the native header color
+has changed in response to a style update.
```
kmichalikk, line 20: are you sure? I remember hearing something about measuring colors, but I don't remember if it was in context of detox or ai agent; there is supposedly a takeScreenshot tool in detox (?)

kmichalikk (approved):
Some nitpicks, otherwise LGTM

I noticed that on pixel fold the when changing color scheme, there are dark flashes for a split second, and on another device there are light flashes (pretty visible when the header is dark vs light). You may want to look at that. Could it just be different default background?

I also observed weird glitches, header leaving dark empty bar instead of hiding, not working back button, but they all went away upon reload and didn't come back.

apps/src/tests/single-feature-tests/stack-v5/test-stack-color-scheme/scenario.md
kligarski: Same comment in other tests, e.g. https://github.com/software-mansion/react-native-screens/pull/4525/changes#diff-0c6a9b4797a75f8fcc3e20a6f1c3df1877b0cf2bdeb73ff56cc7b09acde79dabR15-R19

android/src/main/java/com/swmansion/rnscreens/stack/header/StackHeaderCoordinatorLayout.kt
kligarski: This is an `or` operation on two integers backing the flags. Even if invalidation flag on the provider is not invalidated, we force the update of `forcedFlags`, hence the name.

android/src/main/java/com/swmansion/rnscreens/stack/screen/StackScreenFragmentDelegate.kt
kligarski: 1. The interface is meant to establish communication between `StackScreenFragment`s and `StackContainer`. Currently, it's only handling configuration callback but it can be used in the future for any other necessary callbacks - that's why I called this a generic delegate, not a specific use case one. Similar pattern is used in e.g. `StackHeaderDelegate`, `StackContainerDelegate`.
2. In UIKit which uses a delegate pattern a lot, I've always seen 2 types of callbacks - ones asking for a decision/object (`should...`, `animationControllerForOperation:...`) and other ones with just some kind of notification (`navigationController:didShowViewController:animated:`, `tabBarController:didSelectViewController:`, `tabBarController:displayOrderDidChangeForGroup:`, ...). Delegate introduced in the PR seems to me to fall into the latter category. If we still want to change it, I'd do this in a separate PR with other Stack delegates. Let me know what you think.

android/src/main/java/com/swmansion/rnscreens/stack/header/StackHeaderCoordinatorLayout.kt
kligarski: https://github.com/software-mansion/react-native-screens/pull/4556/commits/e71eefa00032a44041c9360393f0ff7fa3dc437b

kligarski:
> I noticed that on pixel fold the when changing color scheme, there are dark flashes for a split second, and on another device there are light flashes

When header is rebuilt, it flashes before it is laid out correctly. I created a ticket to research if we can prevent this: https://github.com/software-mansion/react-native-screens-labs/issues/1770.

Regarding the Fold bug, the screen which shows up during a transition is a completely different screen (with different configuration than current). This seems to me like an emulator glitch. Might be worth to check a real device in the future though.

https://github.com/user-attachments/assets/ba4e1ecc-e475-41a9-93b5-b0456e04b546

> I also observed weird glitches, header leaving dark empty bar instead of hiding,
> not working back button,

I've seen the second one sometimes but I couldn't reproduce it - let's maybe wait for the back button navigation refactor to land and if we see it once again, we'll try to investigate further.

While testing now, the first one also happened to me. I managed to find a repro for it. I created a ticket to track this: https://github.com/software-mansion/react-native-screens-labs/issues/1771.

EDIT: I found a potential fix. Will create a PR soon.

android/src/main/java/com/swmansion/rnscreens/stack/screen/StackScreenFragmentDelegate.kt
kmichalikk: I'm still not sure but I'm not going to keep arguing. This is fine.

PR #4558, opened 2026-08-26
fix(Android, Tabs): use ViewIdGenerator to identify MenuItems instead of sequential ids

Loloekk:
## Description

On Android, tab `MenuItem`s were identified with sequential ids. Those ids can collide with view tags assigned by the React Native renderer. Tapping the 4th tab dispatch a tap to a RN view with tag `4`.

This PR assigns each tab a `MenuItem` id from `ViewIdGenerator`, which generates ids that do not collide with RN tags.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1108

## Changes

- Persist a `menuItemId` from `ViewIdGenerator` on each `TabsScreenFragment` and use that id when creating/looking up `MenuItem`s, instead of deriving the id from the fragment index.
- In `updateNavigationMenuStructure` (`TabsContainer.kt`), if the menu order no longer matches `tabsModel`, clear and rebuild the menu. With sequential ids, a shuffle was treated as an illegal state. With stable unique ids, tab reordering can desync the menu, so rebuilding is required.
- Update other functions which are using old ids 

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/f3535e6b-ac94-4416-a7d7-08257d427bb9" controls="controls" width="100%"></video> | <video src="https://github.com/user-attachments/assets/dd77bc6c-fd7f-458f-9d6f-f72e5eae3821" controls="controls" width="100%"></video> |

## Test plan

- Check If this problem do not occur in [issue 1108](https://github.com/software-mansion/react-native-screens-labs/issues/1108)
- Check test `test-tabs-stale-update-rejection`
- Check test `test-tabs-simple-nav`

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/tabs/container/TabsContainer.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/tabs/container/TabsContainer.kt
@@ -530,17 +530,19 @@ class TabsContainer internal constructor(
     }
 
     private fun updateNavigationMenuStructure() {
-        if (bottomNavigationView.menu.size != tabsModel.size) {
-            // Most likely first render or some tab has been removed. Let's nuke the menu (easiest option).
-            bottomNavigationView.menu.clear()
+        val menu = bottomNavigationView.menu
+        val isMenuInSync =
+            menu.size == tabsModel.size &&
+                tabsModel.withIndex().all { (index, fragment) ->
+                    menu.getItem(index).itemId == fragment.menuItemId
+                }
+
+        if (isMenuInSync) {
+            return
         }
-        tabsModel.forEachIndexed { index, fragment ->
-            val menuItem =
-                bottomNavigationView.menu.getOrCreateMenuItemForFragmentAt(
-                    index,
-                    fragment.tabsScreen,
-                )
-            check(fragmentIndexForMenuItemId(menuItem.itemId) == index) { "[RNScreens] Illegal state: menu items are shuffled" }
+        menu.clear()
+        tabsModel.forEach { fragment ->
+            menu.getOrCreateMenuItemForFragment(fragment)
         }
     }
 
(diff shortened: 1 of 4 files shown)


Loloekk:
@coderabbitai full review

kmichalikk:
@coderabbitai full review

android/src/main/java/com/swmansion/rnscreens/tabs/container/TabsContainer.kt
```diff
@@ -530,17 +530,19 @@ class TabsContainer internal constructor(
...
+                tabsModel.withIndex().all { (index, fragment) ->
+                    menu.getItem(index).itemId == fragment.menuItemId
+                }
+
+        if (isMenuInSync) {
+            return
+        }
+        menu.clear()
+        tabsModel.forEach { fragment ->
+            menu.getOrCreateMenuItemForFragment(fragment)
```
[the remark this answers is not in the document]
Loloekk: We do not support changing the tabs container at runtime, so restoring a menu item is unnecessary. The check on the marked line is only a safety measure for the futures.

android/src/main/java/com/swmansion/rnscreens/tabs/container/TabsContainer.kt
```diff
@@ -717,12 +719,10 @@ class TabsContainer internal constructor(
...
     }
 
-    private fun getFragmentForMenuItemId(itemId: Int): TabsScreenFragment? = tabsModel.getOrNull(fragmentIndexForMenuItemId(itemId))
+    private fun getFragmentForMenuItemId(itemId: Int): TabsScreenFragment? = tabsModel.find { it.menuItemId == itemId }
 
     private fun getMenuItemIdForFragment(tabsScreenFragment: TabsScreenFragment): Int? =
-        tabsModel.indexOfFirst { it === tabsScreenFragment }.takeIf { it != -1 }?.let {
-            menuItemIdForFragmentAtIndex(it)
-        }
+        tabsScreenFragment.menuItemId.takeIf { tabsModel.any { it === tabsScreenFragment } }
```
kkafar, line 725: Why isn't it just `tabsScreenFragment.menuItemId`?

kkafar (commented):
Thanks! Looks & works fine. I have one question regarding the implementation. Please answer it before we proceed.

android/src/main/java/com/swmansion/rnscreens/tabs/container/TabsContainer.kt
Loloekk: The previous behaviour checked under the hood whether this fragment is present in tabsModel. I know it should always be present, but there was also an it != -1 check. I decided to keep that check even though the code is now simpler.

kkafar: Okay, this convinces me. If I'd done that - I'd simplify it, but there is nothing wrong in keeping the semantics exactly as before - we can modify it if needed later.

kkafar (approved):
Okay, looks great then! Let's proceed.

kkafar:
One thing -> I've also marked this PR with `action:backport-to-v4` - it's an important fix - we should backport it.

PR #4562, opened 2026-08-27
refactor(iOS): polish silence of warnings for v5 components

Loloekk:
## Description

This PR is a follow-up to [PR 4508](https://github.com/software-mansion/react-native-screens/pull/4508). It polish uncorrect silence of warnings.

## Changes

- Use `[[maybe_unused]]` for static integers in  `ios/split/RNSSplitHostController.mm`
- In `ios/stack/host/RNSStackNavigationController.mm`, wrap only the `dumpStackModel` for-loop with `#ifdef RNS_DEBUG_LOGGING`

## Test plan

- Confirm these 4 warnings no longer appear
- Build and run the app in both debug and release

## Checklist

- [x] Ensured that CI passes

### DIFF
--- a/ios/stack/host/RNSStackNavigationController.mm
+++ b/ios/stack/host/RNSStackNavigationController.mm
@@ -126,8 +126,8 @@
 
 - (void)dumpStackModel
 {
-#ifdef RNS_DEBUG_LOGGING
   RNSLog(@"[RNScreens] StackContainer [%ld] MODEL BEGIN", self.view.tag);
+#ifdef RNS_DEBUG_LOGGING
   for (UIViewController *viewController in self.viewControllers) {
     RNSLog(@"[RNScreens] %@", static_cast<RNSStackScreenComponentView *>(viewController.view).screenKey);
   }
--- a/ios/stack/host/RNSStackNavigationController.mm
+++ b/ios/stack/host/RNSStackNavigationController.mm
@@ -126,8 +126,8 @@
 
 - (void)dumpStackModel
 {
-  RNSLog(@"[RNScreens] StackContainer [%ld] MODEL BEGIN", self.view.tag);
 #ifdef RNS_DEBUG_LOGGING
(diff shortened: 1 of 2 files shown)


ios/stack/host/RNSStackNavigationController.mm
```diff
@@ -126,8 +126,8 @@ - (void)performContainerUpdateIfNeeded
 
 - (void)dumpStackModel
 {
-#ifdef RNS_DEBUG_LOGGING
   RNSLog(@"[RNScreens] StackContainer [%ld] MODEL BEGIN", self.view.tag);
+#ifdef RNS_DEBUG_LOGGING
```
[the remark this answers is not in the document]
Loloekk: Intentional. MODEL BEGIN should stay behind RNSLog and remain silent in release. The #ifdef only needs to cover the loop (unused viewController). We didn't mean to log this in all builds.

kkafar, line 129: I don't think this particular change is for better. I'm gonna split a hair in two here, just for educational purposes. Usually I'd fine with this change.

If `RNS_DEBUG_LOGGING` is not set, I think that the intention here is not to print anything. Obviously, both versions of the code achieve that - that's great. But now, consider what the code communicates in both of the versions.

In the original one - the intention is unambiguous. You've wrapped whole function body with the `ifdef`. You clearly communicate the intention. 

In now updated version you rely on internal behavior of `RNSLog`. If that ever changes, the behaviour won't we consistent anymore - you'll print header but no screen details from the for-loop below. Additionally, reading first time through that code it rather looks like the header should always be printed, but the for-loop below - not. 

I'd suggest leaving this one unmodified.

Loloekk: Fixed in [516766e](https://github.com/software-mansion/react-native-screens/pull/4562/commits/516766e5f111875b9ac1851b0f2ec3401f82258c)

kkafar (approved):
Looks good. Thanks!

PR #4569, opened 2026-08-28
chore(test): replace local e2e helpers with shared e2e-utils exports

LKuchno:
## Description

First of the two PRs proposed in the e2e helpers [RFC-1754](https://github.com/software-mansion/react-native-screens-labs/pull/1754): the behavioral cleanup. Spec-local helpers that re-implemented (or slightly diverged from) `e2e-utils` exports are replaced by the shared ones, constants that had drifted per spec are consolidated, and the two helper families that were never shared — the iOS header context menu and the Android Stack v5 toolbar overflow menu — are reconciled to a single implementation in place, in `e2e-utils.ts`. No directory restructuring; that is the follow-up PR.

This changes test behavior in a few deliberate places, all verified on device: shared settle waits (5000 ms) for the Android overflow menu, `selectPickerOption`'s early return and value assertion applied everywhere, `expectTopmostVisible` used instead of stale local copies, and `tapBarBackButton` gaining a settle wait and topmost-match tap on both platforms (its Android branch previously matched an id that nothing sets).

## Changes

- `e2e-utils.ts`
  - Helpers moved here from spec-local implementations — and, where specs had diverged, reconciled into one shared version — rather than newly written: `getSingleMatch`, `getFrame`, `readTopmostText`, `readText`, `tapTopmostButton`, `expectLastClicked`, `expectTopmostButtons`, `overflowMenuText`, `actionMenuItem`, `menuItemImage`, `expectOverflowMenuOrder`, Stack v5 toolbar matchers (`stackV5Toolbar`, `stackV5AppBar`, `stackV5BackButton`, `stackV5HeaderTitle`), `expectIconActionItem` / `expectTextActionItem` / `expectNoActionItem`, `waitForRouteName` / `waitForTopmostRoute` (route-key polling for stacked/nested screens), `openHeaderTitleMenu` / `headerTitle`, bottom-accessory getters + `expectBottomAccessoryAboveTabBar`, and `waitForMenuItem` / `tapMenuItem` inside `createOverflowMenuHelpers`.
  - `selectPickerOption` takes an optional `control` (pickers outside a scroll view are tapped in place) and asserts by label on iOS / text on Android. `scrollUntilVisible` gained a `direction` option (kept scoped to testID targets); `rewindAndScrollUntilVisible` rejects `direction` at the type level since it only ever rewinds up then scrolls down.
  - `getElementAttributes` removed in favor of `getSingleMatch`; the two section navigators share one implementation; `DEFAULT_TIMEOUT_MS` replaces scattered `3000` literals.
  - iOS context-menu and header-item helpers (`openContextMenu`, `dismissContextMenu`, `menuRow`, `checkmarkFor`, `chevronFor`, `menuRowIcon`, `submenuTitleRow`, `headerItem`, `barButtonIcon`) consolidated here from six header specs.
  - Doc comments trimmed to the contract.
- `native-class-names.ts`: new `CLASS_NAME_ANDROID_TOOLBAR` (generic `androidx.appcompat.widget.Toolbar` — matches both the legacy `CustomToolbar` and Stack v5's `MaterialToolbar`, since Detox resolves `by.type` with `isAssignableFrom`) and `CLASS_NAME_UI_CONTEXT_MENU_PLATTER_TRANSITION_VIEW` (was a bare literal in one spec).
- `elements/back-button.ts`: waits for the button, taps the topmost match on both platforms; iOS 26 ambiguity handled with a compound matcher. The Android matcher is its own — an `AppCompatImageButton` scoped to the generic `Toolbar` (`CLASS_NAME_ANDROID_TOOLBAR`) rather than the shared `stackV5BackButton`, since that one is scoped to `MaterialToolbar` specifically and would miss the legacy header.
- `helpers/isVersionEqualOrHigherThan.ts`: argument names changed to `version` / `minimumVersion`.
- 29 single-feature specs and `issue-tests/Test758`: local copies of `isIOSVersionAtLeast`, `optionId`, scroll/picker/switch helpers, overflow-menu open/close/wait logic, toolbar matchers and attribute readers dropped in favor of the shared exports; unused imports removed.
- Every changed suite was run on device: iOS (iPhone 17, iOS 26) header, tabs and stack suites; Android (Pixel 9) toolbar-menu, tabs and stack suites. The two `test-tabs-ime-insets-android` keyboard cases fail on an AVD with `hw.keyboard=yes` identically before and after this change.

### DIFF
--- a/FabricExample/e2e/e2e-utils.ts
+++ b/FabricExample/e2e/e2e-utils.ts
@@ -1,4 +1,5 @@
-import { device, expect, element, by } from 'detox';
+import { expect as jestExpect } from '@jest/globals';
+import { device, expect, element, by, waitFor } from 'detox';
 import {
   AndroidElementAttributes,
   IosElementAttributes,
@@ -6,13 +7,28 @@ import {
 } from 'detox/detox';
 import isVersionEqualOrHigherThan from './helpers/isVersionEqualOrHigherThan';
 import {
+  CLASS_NAME_ANDROID_ACTION_MENU_ITEM_VIEW,
+  CLASS_NAME_ANDROID_APP_COMPAT_IMAGE_VIEW,
   CLASS_NAME_ANDROID_CHECK_BOX,
   CLASS_NAME_ANDROID_LIST_MENU_ITEM_VIEW,
   CLASS_NAME_ANDROID_MENU_DROP_DOWN_LIST_VIEW,
   CLASS_NAME_ANDROID_RADIO_BUTTON,
+  CLASS_NAME_RNS_TABS_BOTTOM_ACCESSORY,
+  CLASS_NAME_UI_BUTTON_BAR_BUTTON,
+  CLASS_NAME_UI_CONTEXT_MENU_CELL,
+  CLASS_NAME_UI_CONTEXT_MENU_CELL_CONTENT_VIEW,
+  CLASS_NAME_UI_CONTEXT_MENU_LIST_VIEW,
+  CLASS_NAME_UI_CONTEXT_MENU_PLATTER_TRANSITION_VIEW,
+  CLASS_NAME_UI_CONTEXT_MENU_SUBMENU_TITLE_VIEW,
+  CLASS_NAME_UI_IMAGE_VIEW,
+  CLASS_NAME_UI_MODERN_BAR_BUTTON,
+  CLASS_NAME_UI_TAB_BAR,
 } from './native-class-names';
 
 const { getIOSVersionNumber } = require('../../scripts/e2e/ios-devices.js');
+
+/** Default for waits Detox's idle sync already mostly covers. */
+const DEFAULT_TIMEOUT_MS = 3000;
 
 export const describeIfiOS =
   device.getPlatform() === 'ios' ? describe : describe.skip;
@@ -21,10 +37,8 @@ export const describeIfAndroid =
   device.getPlatform() === 'android' ? describe : describe.skip;
 
 /**
- * Detox exposes no runtime UIUserInterfaceIdiom query, so the idiom is inferred
- * from the simulator name requested via `RNS_APPLE_SIM_NAME` (see
- * scripts/e2e/ios-devices.js). iPad-only suites self-skip on the default iPhone
- * run and execute only with e.g. RNS_APPLE_SIM_NAME="iPad Pro 13-inch (M4)".
+ * Inferred from `RNS_APPLE_SIM_NAME` (Detox has no idiom query), e.g.
+ * RNS_APPLE_SIM_NAME="iPad Pro 13-inch (M4)". See scripts/e2e/ios-devices.js.
  */
 export const isIPadTarget =
   device.getPlatform() === 'ios' &&
@@ -32,11 +46,7 @@ export const isIPadTarget =
 
 export const describeIfiPad = isIPadTarget ? describe : describe.skip;
 
-/**
- * True when running on iOS at `version` or newer. Use it to pick between the
- * legacy and iOS 26 variants of the private UIKit class names in
- * ./native-class-names.
- */
+/** `true` on iOS at `version` or newer; `false` on Android. */
 export function isIOSVersionAtLeast(version: string): boolean {
   return (
     device.getPlatform() === 'ios' &&
@@ -44,11 +54,7 @@ export function isIOSVersionAtLeast(version: string): boolean {
   );
 }
 
-/**
- * Suites for iOS 26+ only features (e.g. `bottomAccessory`, the header overflow
- * button). `isIOSVersionAtLeast` is false on Android, so these stay iOS-only
- * and additionally self-skip on iOS 18 and older.
- */
+/** Suites for iOS 26+ only features; skipped on Android and older iOS. */
 export const describeIfiOS26 = isIOSVersionAtLeast('26.0')
   ? describe
   : describe.skip;
@@ -61,27 +67,40 @@ export type ScrollOptions = {
   pixels?: number;
   /** Swipe start, as a fraction of height. `NaN` leaves it to Detox. */
   startPercentage?: number;
+  /** Scroll direction; `whileElement` only ever scrolls one way. */
+  direction?: 'up' | 'down';
 };
 
+/** A `testID`, or any matcher for targets that carry no `testID`. */
+export type ScrollTarget = string | NativeMatcher;
+
+function toMatcher(target: ScrollTarget): NativeMatcher {
+  return typeof target === 'string' ? by.id(target) : target;
+}
+
 export async function scrollUntilVisible(
-  id: string,
+  target: ScrollTarget,
   scrollViewId: string,
-  { pixels = 600, startPercentage = 0.85 }: ScrollOptions = {},
+  {
+    pixels = 600,
+    startPercentage = 0.85,
+    direction = 'down',
+  }: ScrollOptions = {},
 ) {
-  await waitFor(element(by.id(id)))
+  await waitFor(element(toMatcher(target)))
     .toBeVisible()
     .whileElement(by.id(scrollViewId))
-    .scroll(pixels, 'down', Number.NaN, startPercentage);
+    .scroll(pixels, direction, Number.NaN, startPercentage);
 }
 
 /** Rewinds to the top first — `whileElement` only scrolls one way. */
 export async function rewindAndScrollUntilVisible(
-  id: string,
+  target: ScrollTarget,
   scrollViewId: string,
   options: ScrollOptions = {},
 ) {
   await element(by.id(scrollViewId)).scrollTo('top');
-  await scrollUntilVisible(id, scrollViewId, options);
+  await scrollUntilVisible(target, scrollViewId, options);
 }
 
 export async function selectIssueTestScreen(screenName: string) {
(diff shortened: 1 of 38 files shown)


FabricExample/e2e/e2e-utils.ts
```diff
@@ -61,27 +70,40 @@ export type ScrollOptions = {
...
 /** Rewinds to the top first — `whileElement` only scrolls one way. */
 export async function rewindAndScrollUntilVisible(
-  id: string,
+  target: ScrollTarget,
   scrollViewId: string,
   options: ScrollOptions = {},
 ) {
   await element(by.id(scrollViewId)).scrollTo('top');
-  await scrollUntilVisible(id, scrollViewId, options);
+  await scrollUntilVisible(target, scrollViewId, options);
```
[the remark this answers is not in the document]
LKuchno: Fixed in 17d8d7ee6 — `rewindAndScrollUntilVisible` no longer accepts `direction` at all (`Omit<ScrollOptions, 'direction'>`): it rewinds to the top, so only a downward scan can reach anything, and an upward scan from the top is unsatisfiable by construction. The `direction` option remains only on `scrollUntilVisible`, which has a real bidirectional caller and no rewind. The unused `ScrollTarget` matcher generalization was reverted in the same commit.

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-back-button-android.e2e.ts
```diff
@@ -20,38 +20,29 @@ const PUSH_ANOTHER = 'PUSH ANOTHER';
...
 // exist". The toolbar is asserted first, otherwise a header that never
 // rendered would pass too. Each screen builds its own toolbar, hence the index.
 async function expectNoBackButton() {
-  const toolbarMatcher = by.type(CLASS_NAME_ANDROID_MATERIAL_TOOLBAR);
-  const count = await countMatches(toolbarMatcher);
-  await expect(element(toolbarMatcher).atIndex(count - 1)).toBeVisible();
-  await expect(element(backButtonMatcher)).not.toExist();
+  await expectTopmostVisible(stackV5Toolbar);
+  await expect(element(stackV5BackButton())).not.toExist();
 }
```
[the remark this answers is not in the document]
LKuchno: Not applicable here, and the suggested scoping isn't expressible in Detox. On Android a covered screen keeps its toolbar but *loses* its chevron under a headered top screen (documented on `stackV5BackButton` in e2e-utils), and the only covered screen in this suite is the stack root, which never has one — so "zero chevrons anywhere" is the correct expected state, and the global `not.toExist()` is stronger than a topmost-scoped check, not weaker. `expectNoBackButton` also asserts the topmost toolbar is visible first, which rules out the one real caveat (a headerless top screen leaving the covered chevron visible). As for the fix: "descendant of the last `stackV5Toolbar()` match" can't be written — `withAncestor` takes a matcher, not an indexed match, frames can't disambiguate (stacked toolbars share coordinates), and hierarchy order can't distinguish "topmost has none, covered has one" from "topmost has the only one".

FabricExample/e2e/e2e-utils.ts
```diff
@@ -571,25 +798,25 @@ export type OverflowMenuControl = {
...
       .toBeVisible()
       .withTimeout(MENU_ANIMATION_TIMEOUT_MS);
   };
 
+  /** `waitForScreen` reported rather than thrown — a stacked popup keeps it false. */
+  const isScreenAddressable = () =>
+    waitForScreen().then(
+      () => true,
+      () => false,
+    );
```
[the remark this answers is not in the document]
LKuchno: https://github.com/software-mansion/react-native-screens/pull/4569/changes/e12472eeb6a9539a599520b95e0dbb0eb84cd6c4

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-prevent-native-dismiss-single-stack.e2e.ts
```diff
@@ -171,14 +55,6 @@ describeIfAndroid('Stack v5: prevent native dismiss - single stack', () => {
     }
   }
 
-  /**
-   * Asserts no `onNativeDismissPrevented` toast is on screen. Every step clears
-   * its own toasts, so an unexpected one always lands at position 1.
-   */
-  async function expectNoToast(): Promise<void> {
-    await expect(element(by.label(toastLabel(1)))).not.toExist();
-  }
```
t0maboro, line 180: Instead of looking for a specific label, we now match any label that satisfies the regex. The matching criteria seem to be too loose for me now.

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-batch-commands-android.e2e.ts
```diff
@@ -22,62 +24,43 @@ import { CLASS_NAME_ANDROID_ACTION_MENU_ITEM_VIEW } from '../../native-class-nam
 const SCROLLVIEW_ID = 'toolbar-menu-batch-commands-scrollview';
 const HEADER_TITLE = 'Toolbar Menu Batch Commands Test';
 const SCROLL_STEP = { pixels: 300 };
+const SETTINGS_CONTROL = { scrollViewId: SCROLLVIEW_ID, ...SCROLL_STEP };
 
-const EVENT_TIMEOUT_MS = 3000;
 // Generous on purpose: the failing-image cases wait on a network error, and a
 // slow one must not read as a stuck batch.
-const IMAGE_LOAD_TIMEOUT_MS = 90000;
+const IMAGE_LOAD_TIMEOUT_MS = 30000;
```
t0maboro: why this line was changed? it doesn't seem to be related to the e2e-utils refactor

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-groups-android.e2e.ts
```diff
@@ -188,56 +132,12 @@ async function sendCommand({ id, checked, title, hidden }: CommandSpec) {
...
-  }
-
-  if (pressCount > 0) {
-    await waitForScreen();
-  }
-}
+const { waitForScreen, closeMenuIfOpen, closingMenuAfter, withOverflowMenu } =
+  createOverflowMenuHelpers({ scrollViewId: SCROLLVIEW_ID });
 
 async function tapMenuItem(title: string) {
```
t0maboro: Was this left intentionally? It has the same signature as the helper from utils

FabricExample/e2e/e2e-utils.ts
```diff
@@ -1,333 +1,123 @@
...
+  CLASS_NAME_UI_MODERN_BAR_BUTTON,
+  CLASS_NAME_UI_TAB_BAR,
 } from './native-class-names';
 
 const { getIOSVersionNumber } = require('../../scripts/e2e/ios-devices.js');
 
+// ---------------------------------------------------------------------------
+// Platform and OS-version gates
+// → structure PR: framework/platform.ts
+// ---------------------------------------------------------------------------
```
t0maboro, line 37: assuming that's the marking for further file split in followup, if not, should be removed

FabricExample/e2e/e2e-utils.ts
```diff
@@ -489,11 +466,17 @@ export const MENU_ANIMATION_TIMEOUT_MS = 5000;
...
 
+/** Outlasts a popup's exit animation without paying the full animation timeout. */
+const MENU_DISMISS_PROBE_TIMEOUT_MS = 1000;
+
 /** The accessibility label AppCompat gives the overflow button. */
-const OVERFLOW_MENU_LABEL = 'More options';
+export const OVERFLOW_MENU_LABEL = 'More options';
+
+/** The popup hosting the overflow menu (or, once opened, a submenu). */
+export const overflowMenuMatcher = (): NativeMatcher =>
```
t0maboro: no need to export, used only in this file

FabricExample/e2e/e2e-utils.ts
```diff
@@ -647,5 +680,468 @@ export function createOverflowMenuHelpers({
...
+/**
+ * Dismisses the menu at any submenu depth by tapping UIKit's dismiss overlay
+ * near its leading edge (a tap on the menu itself only pops one level).
+ */
+export async function dismissContextMenu(
+  timeout = CONTEXT_MENU_ANIMATION_TIMEOUT_MS,
+) {
+  await tapWithinFrame(
+    await getFrame(
+      by.type(CLASS_NAME_UI_CONTEXT_MENU_PLATTER_TRANSITION_VIEW),
```
t0maboro, line 873: (if possible to launch on iOS 18)
Q: Does iOS 18 have platters? iirc it's liquid-glass-related stuff.

FabricExample/e2e/e2e-utils.ts
LKuchno: Yes, this heading are added to make further change easier.

FabricExample/e2e/e2e-utils.ts
LKuchno: I've checked it on iOS18.6 and it works too. The naming is a bit of a coincidence: _UIContextMenuPlatterTransitionView is UIKit's private name for the context-menu dismiss overlay and has existed, well before liquid glass. It's unrelated to the actual liquid-glass "platter" UI (e.g. _UINavigationBarPlatterView for the iOS 26 nav-bar button bubbles) - Apple just reused the word "platter" in both class names.

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-groups-android.e2e.ts
LKuchno: It was intentional but during check I find out there is way to reuse it partially in this test too.
The signature matched, but the local version was a bare tap with no wait/assertions, needed because this file also uses it to open the submenu via "More", which doesn't close the popup like a real leaf tap does (so it can't use the shared helper's "gone after tap" check).

**Refactored:**
leaf-item taps now use the shared tapMenuItem from e2e-utils directly, and the local override is scoped down to just the "More" submenu-trigger tap in openSubmenu, which is where the divergence is actually needed.

https://github.com/software-mansion/react-native-screens/pull/4569/changes/51eb206d892f0d993041f88c403bcf8aa936f1cb

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-toolbar-menu-batch-commands-android.e2e.ts
LKuchno: It's not related, analyzing tests I thought we don't need that long timeout but it should be changed in separate PR - I revert this change and if it will be needed i will introduce this change in dedicated PR.
https://github.com/software-mansion/react-native-screens/pull/4569/changes/bdf556e8587ce40409881a3cfcc0edba9f91693c

FabricExample/e2e/single-feature-tests/stack-v5/test-stack-prevent-native-dismiss-single-stack.e2e.ts
LKuchno: I refactor this function, kept regex for cases when expected not existing message is not provided, and old matching way for those with toast message provided. Accordingly e2e test using this function were updated with messages - where it was applicable.
https://github.com/software-mansion/react-native-screens/pull/4569/changes/fcbb74f3c9f1eacb650ebc284d846d1b896e69f5

FabricExample/e2e/e2e-utils.ts
LKuchno: https://github.com/software-mansion/react-native-screens/pull/4569/changes/f4337248842bbcc1fa4b07a18ecfd69e23603221

FabricExample/e2e/helpers/isVersionEqualOrHigherThan.ts
```diff
@@ -24,23 +24,24 @@ function assertSupportedVersionString(
 
 function compareVersions(
   version: MajorVersion | MajorMinorVersion,
-  versionToCompare: MajorVersion | MajorMinorVersion,
+  minimumVersion: MajorVersion | MajorMinorVersion,
 ) {
   const [majorA, minorA = '0'] = version.split('.').map(Number);
-  const [majorB, minorB = '0'] = versionToCompare.split('.').map(Number);
+  const [majorB, minorB = '0'] = minimumVersion.split('.').map(Number);
```
kligarski: nit: I'd keep `versionToCompare` name here because this function is just a comparator that accepts 2 numbers and returns an integer. In `isVersionEqualOrHigherThan` the name makes sense though.

kligarski (approved):

FabricExample/e2e/helpers/isVersionEqualOrHigherThan.ts
LKuchno: Thank you! 
https://github.com/software-mansion/react-native-screens/pull/4569/changes/45ce9f2b99dbf5abbef2cd6b239629334da43e76

PR #4573, opened 2026-08-31
fix(Android, FormSheet v5): Move the sheet above the keyboard

t0maboro:
## Description

`FormSheet` on Android has ignored the software keyboard so far. The sheet now follows the keyboard: whichever detent it rests at, it is pushed up by the keyboard height as far as the screen allows, and only the part that can't be pushed above the keyboard shrinks the content box. The sheet moves together with the keyboard animation.

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1276

## Changes

- `FormSheetDimensionsCoordinator` tracks the IME inset and derives `keyboardLift = max(0, ime.bottom - systemBars.bottom)`, the part of the keyboard that sticks out above the navigation bar.
- `FormSheetDetents` extends every resting height by the lift, capped at the container height. The peek height accounts for Material adding the keyboard inset back on its own, **the middle detent is kept at least 2 px below the largest one so Material can still tell the two states apart**, and the content box keeps its size while the sheet rises and shrinks only once the sheet hits the top of the screen.
- `FormSheetKeyboardCoordinator` defines our own `WindowInsetsAnimationCompat.Callback` in place of Material's. It records the sheet's layout `top` before the keyboard insets are applied, translates the sheet back there once the end-state layout is in place and drives `translationY` to zero along the keyboard animation. Tracking is enabled by `FormSheetPresentationManager` only while the sheet rests in `PRESENTED`; otherwise the enter/exit animators own the translation.

| Android | iOS |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/325f98f8-f2e5-4fa8-9a48-773209c4d106" /> | <video src="https://github.com/user-attachments/assets/fc39516a-b7fe-462d-b67e-9680fc99e4a0" /> |

## Test plan

Added a dedicated SFT (maybe it should be CIT 🤔).

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/modals/formsheet/native/coordinator/FormSheetDimensionsCoordinator.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/modals/formsheet/native/coordinator/FormSheetDimensionsCoordinator.kt
@@ -4,7 +4,6 @@ import android.view.ViewGroup
 import android.widget.FrameLayout
 import androidx.core.view.ViewCompat
 import androidx.core.view.WindowInsetsCompat
-import androidx.core.view.doOnLayout
 import com.swmansion.rnscreens.modals.formsheet.native.core.FormSheetAvailableHeightProvider
 import com.swmansion.rnscreens.modals.formsheet.native.core.FormSheetContainer
 import com.swmansion.rnscreens.modals.formsheet.native.core.FormSheetDialog
@@ -14,12 +13,12 @@ import com.swmansion.rnscreens.modals.formsheet.native.model.FormSheetDetents
 internal class FormSheetDimensionsCoordinator(
     private val dialog: FormSheetDialog,
     private val container: FormSheetContainer,
-    private val bottomSheetView: FrameLayout?,
     private val behaviorController: FormSheetBehaviorController?,
 ) : FormSheetContentSizeChangeDelegate,
     FormSheetAvailableHeightProvider.OnAvailableHeightMeasuredListener {
     private var lastTopInset = 0
     private var lastBottomInset = 0
+    private var lastImeInset = 0
     private var currentDetents: FormSheetDetents? = null
     private var currentInitialDetentIndex: Int = 0
     private var shouldApplyInitialDetent: Boolean = false
@@ -34,38 +33,20 @@ internal class FormSheetDimensionsCoordinator(
     internal fun setup() {
         dialog.availableHeightProvider.availableHeightListener = this
         setupWindowInsetsListener()
-
-        bottomSheetView?.let { view ->
-            disableMaterialInsetsAnimationCallback(view)
-        }
     }
 
     private fun setupWindowInsetsListener() {
         ViewCompat.setOnApplyWindowInsetsListener(container) { _, insets ->
             val topInset = getTopInset(insets)
             val bottomInset = getBottomInset(insets)
-            if (topInset != lastTopInset || bottomInset != lastBottomInset) {
+            val imeInset = getImeInset(insets)
+            if (topInset != lastTopInset || bottomInset != lastBottomInset || imeInset != lastImeInset) {
                 lastTopInset = topInset
                 lastBottomInset = bottomInset
+                lastImeInset = imeInset
                 invalidateGeometry()
             }
             insets
-        }
-    }
-
-    /**
-     * BottomSheetBehavior registers an internal `WindowInsetsAnimationCallback` on the
-     * sheet view during its first `onLayoutChild`. That callback drives `translationY` to follow
-     * animated inset changes, what interferes with our slide-in custom animation.
-     *
-     * We manage insets ourselves by setting a fixed height for FormSheetContainer, so we can
-     * clear the Material's callback to remove the conflict entirely.
-     *
-     * This method must run after the first layout pass.
-     */
-    private fun disableMaterialInsetsAnimationCallback(view: FrameLayout) {
-        view.doOnLayout {
-            ViewCompat.setWindowInsetsAnimationCallback(it, null)
         }
     }
 
@@ -116,12 +97,14 @@ internal class FormSheetDimensionsCoordinator(
      * our container are measured, so the values applied here are picked up by the very same traversal.
      */
     private fun resolveGeometry(sheetAvailableSpace: Int) {
+        val bottomInset = effectiveBottomInset
+
         currentDetents?.let { detents ->
             behaviorController?.updateSheetBehavior(
                 detents = detents,
                 sheetAvailableSpace = sheetAvailableSpace,
                 contentHeightForFitToContents = currentContentHeight,
-                nativeContainerPaddingBottom = lastBottomInset,
+                nativeContainerPaddingBottom = bottomInset,
                 initialDetentIndex = currentInitialDetentIndex,
                 applyInitialDetent = shouldApplyInitialDetent,
             )
@@ -129,8 +112,8 @@ internal class FormSheetDimensionsCoordinator(
         }
 
         val sheetContainerHeight =
-            currentDetents?.sheetContainerHeight(sheetAvailableSpace, lastTopInset, lastBottomInset, currentContentHeight)
-                ?: (sheetAvailableSpace - lastTopInset - lastBottomInset).coerceAtLeast(0)
+            currentDetents?.sheetContainerHeight(sheetAvailableSpace, lastTopInset, bottomInset, currentContentHeight)
+                ?: (sheetAvailableSpace - lastTopInset - bottomInset).coerceAtLeast(0)
 
         val layoutParams =
             container.layoutParams
@@ -142,6 +125,11 @@ internal class FormSheetDimensionsCoordinator(
             container.layoutParams = layoutParams
         }
     }
+
+    private val effectiveBottomInset: Int
+        get() = maxOf(lastBottomInset, lastImeInset)
+
+    private fun getImeInset(insetsCompat: WindowInsetsCompat): Int = insetsCompat.getInsets(WindowInsetsCompat.Type.ime()).bottom
 
     private fun getTopInset(insetsCompat: WindowInsetsCompat): Int =
         insetsCompat
--- a/android/src/main/java/com/swmansion/rnscreens/modals/formsheet/native/coordinator/FormSheetDimensionsCoordinator.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/modals/formsheet/native/coordinator/FormSheetDimensionsCoordinator.kt
@@ -97,14 +97,13 @@ internal class FormSheetDimensionsCoordinator(
      * our container are measured, so the values applied here are picked up by the very same traversal.
      */
     private fun resolveGeometry(sheetAvailableSpace: Int) {
-        val bottomInset = effectiveBottomInset
-
         currentDetents?.let { detents ->
             behaviorController?.updateSheetBehavior(
                 detents = detents,
                 sheetAvailableSpace = sheetAvailableSpace,
                 contentHeightForFitToContents = currentContentHeight,
-                nativeContainerPaddingBottom = bottomInset,
+                nativeContainerPaddingBottom = lastBottomInset,
+                keyboardLift = keyboardLift,
                 initialDetentIndex = currentInitialDetentIndex,
                 applyInitialDetent = shouldApplyInitialDetent,
             )
@@ -112,8 +111,13 @@ internal class FormSheetDimensionsCoordinator(
         }
 
         val sheetContainerHeight =
-            currentDetents?.sheetContainerHeight(sheetAvailableSpace, lastTopInset, bottomInset, currentContentHeight)
-                ?: (sheetAvailableSpace - lastTopInset - bottomInset).coerceAtLeast(0)
+            currentDetents?.sheetContainerHeight(
+                sheetAvailableSpace,
+                lastTopInset,
+                lastBottomInset,
+                currentContentHeight,
+                keyboardLift,
+            ) ?: (sheetAvailableSpace - lastTopInset - lastBottomInset - keyboardLift).coerceAtLeast(0)
 
         val layoutParams =
             container.layoutParams
@@ -126,8 +130,13 @@ internal class FormSheetDimensionsCoordinator(
         }
     }
 
-    private val effectiveBottomInset: Int
-        get() = maxOf(lastBottomInset, lastImeInset)
+    /**
+     * The part of the keyboard inset that sticks out above the bottom system inset. Material pads the sheet by
+     * the larger of the two, so this is exactly how much the sheet has to be extended to keep its detent-sized
+     * part above the keyboard.
+     */
+    private val keyboardLift: Int
+        get() = (lastImeInset - lastBottomInset).coerceAtLeast(0)
 
     private fun getImeInset(insetsCompat: WindowInsetsCompat): Int = insetsCompat.getInsets(WindowInsetsCompat.Type.ime()).bottom
 
--- a/android/src/main/java/com/swmansion/rnscreens/modals/formsheet/native/model/FormSheetDetents.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/modals/formsheet/native/model/FormSheetDetents.kt
@@ -60,8 +60,9 @@ internal class FormSheetDetents(
     ): Int {
         /*
          * We add the `bottomInset` to the `contentHeight` so that the Material BottomSheet
-         * is laid out behind the system navigation bar. The sheet's container covers the insets,
-         * while the RN content is strictly constrained to `contentHeight`.
+         * is laid out behind the system navigation bar or the keyboard (the sheet grows by
+         * the keyboard height and gets pushed above it, following iOS approach). The sheet's
+         * container covers the insets, while the RN content is strictly constrained to `contentHeight`.
          */
         if (contentHeight <= 0) {
             // Avoid collapsing the sheet before the React content has been laid out and measured.
@@ -102,7 +103,8 @@ internal class FormSheetDetents(
             if (contentHeight <= 0) {
                 return (containerHeight - topInset - bottomInset).coerceAtLeast(0)
             }
-            return contentHeight.coerceAtMost(containerHeight)
+            // Content taller than the safe area is aligned with it.
+            return contentHeight.coerceAtMost((containerHeight - topInset - bottomInset).coerceAtLeast(0))
         }
 
         // Bottom inset is always fully subtracted because the sheet is in its dedicated window and it's
--- a/android/src/main/java/com/swmansion/rnscreens/modals/formsheet/native/model/FormSheetDetents.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/modals/formsheet/native/model/FormSheetDetents.kt
@@ -92,6 +92,7 @@ internal class FormSheetDetents(
         val middleDetentHeight =
             heightAt(1, containerHeight, keyboardLift)
                 .coerceAtMost(maxAllowedHeight(containerHeight, keyboardLift) - MIDDLE_DETENT_MIN_GAP)
+                .coerceAtLeast(1)
         return middleDetentHeight.toFloat() / containerHeight
     }
 
--- a/android/src/main/java/com/swmansion/rnscreens/modals/formsheet/native/model/FormSheetDetents.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/modals/formsheet/native/model/FormSheetDetents.kt
@@ -33,59 +33,82 @@ internal class FormSheetDetents(
 
     private fun heightFractionAt(index: Int): Double = detents[index]
 
+    // Height of the sheet resting at the given detent, measured down to the bottom edge of the container.
     private fun heightAt(
         index: Int,
         containerHeight: Int,
-    ): Int = (heightFractionAt(index) * containerHeight).roundToInt()
-
-    private fun firstHeight(containerHeight: Int): Int = heightAt(0, containerHeight)
+        keyboardLift: Int,
+    ): Int = ((heightFractionAt(index) * containerHeight).roundToInt() + keyboardLift).coerceAtMost(containerHeight)
 
     /**
      * Height handed to Material as `peekHeight`. Material treats the peek height as the content height above
-     * the bottom system inset and adds that inset back (`BottomSheetBehavior.calculatePeekHeight`), while every
+     * the bottom inset and adds that inset back (`BottomSheetBehavior.calculatePeekHeight`), while every
      * other metric (`maxHeight`, `halfExpandedRatio`) describes the sheet down to the screen edge. The inset is
      * subtracted here so the lowest detent is resolved against the same reference as the other ones.
+     * Material's inset covers the keyboard as well, so the lift is subtracted along with the system inset.
      */
     internal fun peekHeight(
         containerHeight: Int,
         bottomInset: Int,
-    ): Int = (firstHeight(containerHeight) - bottomInset).coerceAtLeast(0)
+        keyboardLift: Int = 0,
+    ): Int = (heightAt(0, containerHeight, keyboardLift) - bottomInset - keyboardLift).coerceAtLeast(0)
 
-    internal fun maxAllowedHeight(containerHeight: Int): Int = heightAt(count - 1, containerHeight)
+    internal fun maxAllowedHeight(
+        containerHeight: Int,
+        keyboardLift: Int = 0,
+    ): Int = heightAt(count - 1, containerHeight, keyboardLift)
 
     internal fun maxAllowedHeightForFitToContents(
         containerHeight: Int,
         contentHeight: Int,
         bottomInset: Int,
+        keyboardLift: Int = 0,
     ): Int {
         /*
-         * We add the `bottomInset` to the `contentHeight` so that the Material BottomSheet
-         * is laid out behind the system navigation bar or the keyboard (the sheet grows by
-         * the keyboard height and gets pushed above it, following iOS approach). The sheet's
-         * container covers the insets, while the RN content is strictly constrained to `contentHeight`.
+         * We add the `bottomInset` and the `keyboardLift` to the `contentHeight` so that the Material BottomSheet
+         * is laid out behind the system navigation bar or the keyboard. The sheet's container covers the insets,
+         * while the RN content is strictly constrained to `contentHeight`.
          */
         if (contentHeight <= 0) {
             // Avoid collapsing the sheet before the React content has been laid out and measured.
             return containerHeight
         }
-        return (contentHeight + bottomInset).coerceAtMost(containerHeight)
+        return (contentHeight + bottomInset + keyboardLift).coerceAtMost(containerHeight)
     }
 
-    internal fun halfExpandedRatio(): Float {
+    /**
+     * Ratio of the middle detent's sheet height to the container height, handed to Material as `halfExpandedRatio`.
+     * Material resolves the half-expanded position from it as `(int) (parentHeight * (1 - ratio))`.
+     *
+     * The keyboard lift can push both the middle and the largest detent to the top of the container. Material needs
+     * the half-expanded position strictly below the expanded one so the middle detent is kept at least
+     * [MIDDLE_DETENT_MIN_GAP] px shorter than the largest one.
+     */
+    internal fun halfExpandedRatio(
+        containerHeight: Int,
+        keyboardLift: Int = 0,
+    ): Float {
         check(count == MAX_DETENTS) { "[RNScreens] Exactly $MAX_DETENTS detents are required for halfExpandedRatio." }
-        return heightFractionAt(1).toFloat()
+        val middleDetentHeight =
+            heightAt(1, containerHeight, keyboardLift)
+                .coerceAtMost(maxAllowedHeight(containerHeight, keyboardLift) - MIDDLE_DETENT_MIN_GAP)
+        return middleDetentHeight.toFloat() / containerHeight
     }
 
     internal fun expandedOffsetFromTop(
         containerHeight: Int,
         topInset: Int = 0,
+        keyboardLift: Int = 0,
     ): Int {
         check(count == MAX_DETENTS) { "[RNScreens] Exactly $MAX_DETENTS detents are required for expandedOffsetFromTop." }
-        return largestDetentTopOffset(containerHeight) + topInset
+        return largestDetentTopOffset(containerHeight, keyboardLift) + topInset
     }
 
     // Distance from the top of the window to the top of the largest detent's surface.
-    private fun largestDetentTopOffset(containerHeight: Int): Int = containerHeight - maxAllowedHeight(containerHeight)
+    private fun largestDetentTopOffset(
+        containerHeight: Int,
+        keyboardLift: Int,
+    ): Int = containerHeight - maxAllowedHeight(containerHeight, keyboardLift)
 
     /**
      * Material's BottomSheetDialog dynamically applies padding when its content overlaps
@@ -97,25 +120,32 @@ internal class FormSheetDetents(
         topInset: Int,
         bottomInset: Int,
         contentHeight: Int = 0,
+        keyboardLift: Int = 0,
     ): Int {
+        val bottomPadding = bottomInset + keyboardLift
+
         if (isFitToContents) {
             // Until we have a measured content height, fall back to a safe-area height so Yoga can lay out.
             if (contentHeight <= 0) {
-                return (containerHeight - topInset - bottomInset).coerceAtLeast(0)
+                return (containerHeight - topInset - bottomPadding).coerceAtLeast(0)
             }
             // Content taller than the safe area is aligned with it.
-            return contentHeight.coerceAtMost((containerHeight - topInset - bottomInset).coerceAtLeast(0))
+            return contentHeight.coerceAtMost((containerHeight - topInset - bottomPadding).coerceAtLeast(0))
         }
 
-        // Bottom inset is always fully subtracted because the sheet is in its dedicated window and it's
+        // Bottom padding is always fully subtracted because the sheet is in its dedicated window and it's
         // anchored to the bottom.
         // Top inset is subtracted only by the amount the sheet actually overlaps it.
-        val topOverlap = (topInset - largestDetentTopOffset(containerHeight)).coerceAtLeast(0)
-        return (maxAllowedHeight(containerHeight) - topOverlap - bottomInset).coerceAtLeast(0)
+        val topOverlap = (topInset - largestDetentTopOffset(containerHeight, keyboardLift)).coerceAtLeast(0)
+        return (maxAllowedHeight(containerHeight, keyboardLift) - topOverlap - bottomPadding).coerceAtLeast(0)
     }
 
     companion object {
         const val MAX_DETENTS = 3
         const val FIT_TO_CONTENTS_DETENT_VALUE = -1.0
+
+        // Material resolves the half-expanded position in float precision and the truncation to int can eat 1 px
+        // which would put the middle detent back on the expanded position.
+        private const val MIDDLE_DETENT_MIN_GAP = 2
     }
 }
--- a/apps/src/tests/single-feature-tests/form-sheet/test-form-sheet-keyboard/scenario.md
+++ b/apps/src/tests/single-feature-tests/form-sheet/test-form-sheet-keyboard/scenario.md
@@ -1,0 +1,138 @@
+# Test Scenario: Keyboard Integration
+
+## Details
+
+**Description:** Verify how a `FormSheet` reacts to the system keyboard when a TextInput inside it gets focused. Two sheets are covered: one with two detents (`[0.6, 1.0]`) and one with `detents="fitToContents"`. Each has a text input at the top and at the bottom of its content.
+
+**OS test creation version:** iOS: 26.5, Android: API Level 36.
+
+## E2E test
+
+TBD: Planned, but will be implemented separately.
+
+## Prerequisites
+
+- iPhone: device or simulator. On the simulator make sure the software keyboard is shown.
+- Android: phone device or emulator. The emulator has to use the on-screen keyboard.
+
+## Note
+
+- Keyboard handling differs between platforms and both behaviors are expected:
+  - **iOS:** UIKit handles it. A sheet resting at a detent smaller than the largest one grows to the largest detent while the keyboard is shown. A sheet resting at its largest detent (including `fitToContents`) is pushed up so that it sits above the keyboard; its view becomes taller by the keyboard height, the extra area is hidden behind the keyboard. Content laid out at the bottom of a sheet resting at the largest detent might end up under the keyboard.
+  - **Android:** the sheet reacts to the keyboard insets. At a lower detent the sheet moves up by the keyboard height, at the largest detent the content box shrinks so that it stays above the keyboard, a `fitToContents` sheet moves up as a whole. The sheet moves together with the keyboard animation.
+
+## Steps - iPhone
+
+### Baseline
+
+1. Launch the app and navigate to the **Keyboard Integration** screen.
+
+- [ ] The host screen shows the "FormSheet Test" title, the "Open FormSheet (detents)" and "Open FormSheet (fitToContents)" buttons.
+
+---
+
+### Detents – keyboard at the lower detent
+
+2. Tap "Open FormSheet (detents)".
+
+- [ ] The sheet presents at the lower detent (0.6). "Top input" is at the top of the sheet, "Bottom input" and "Dismiss from JS" at its bottom.
+
+3. Tap "Top input".
+
+- [ ] The keyboard slides in and the sheet grows to the largest detent (1.0) at the same time. "Top input" is focused and visible. The content is laid out for the taller sheet, "Bottom input" ends up under the keyboard.
+
+4. Press the Return key on the keyboard.
+
+- [ ] The keyboard hides and the sheet returns to the lower detent (0.6). The content is laid out for the smaller sheet again, nothing is clipped and no empty space is left.
+
+---
+
+### Detents – keyboard at the largest detent
+
+5. Drag the sheet up to the largest detent (1.0), then tap "Bottom input".
+
+- [ ] The keyboard slides in, the sheet stays at 1.0. "Bottom input" is focused but covered by the keyboard – this matches the native sheet behavior.
+
+6. With the keyboard shown, swipe the sheet down past the lower detent.
+
+- [ ] Both the keyboard and the sheet are dismissed. The host screen is undimmed and both "Open FormSheet" buttons are pressable again.
+
+---
+
+### fitToContents
+
+7. Tap "Open FormSheet (fitToContents)".
+
+- [ ] The sheet presents with a height matching its content.
+
+8. Tap "Bottom input".
+
+- [ ] The keyboard slides in and the sheet moves up so that its whole content sits above the keyboard. "Bottom input" is focused and visible.
+
+9. Press the Return key on the keyboard.
+
+- [ ] The keyboard hides and the sheet returns to its resting position at the bottom of the screen.
+
+10. Tap "Dismiss from JS".
+
+- [ ] The sheet dismisses and the host screen is undimmed.
+
+## Steps - Android
+
+### Baseline
+
+1. Launch the app and navigate to the **Keyboard Integration** screen.
+
+- [ ] The host screen shows the "FormSheet Test" title, the "Open FormSheet (detents)" and "Open FormSheet (fitToContents)" buttons.
+
+---
+
+### Detents – keyboard at the lower detent
+
+2. Tap "Open FormSheet (detents)".
+
+- [ ] The sheet presents at the lower detent (0.6). "Top input" is visible at the top of the sheet; "Bottom input" and "Dismiss from JS" are not visible yet (the content box is laid out to the largest detent).
+
+3. Tap "Top input".
+
+- [ ] The keyboard slides in and the sheet moves up by the keyboard height, following the keyboard animation – no jump before or after it. "Top input" is focused and visible.
+
+4. Press the system back button (or use the back gesture).
+
+- [ ] The keyboard hides and the sheet moves back down together with it, settling at the lower detent (0.6). The sheet stays presented.
+
+---
+
+### Detents – keyboard at the largest detent
+
+5. Drag the sheet up to the largest detent (1.0), then tap "Bottom input".
+
+- [ ] The keyboard slides in and the content box shrinks to the area above the keyboard. "Bottom input" is focused and visible right above the keyboard, together with "Dismiss from JS".
+
+6. Press the system back button.
+
+- [ ] The keyboard hides, the content box grows back to the full sheet height and "Bottom input" moves back to the bottom of the sheet. The sheet stays at 1.0.
+
+7. Tap "Top input", then – with the keyboard shown – swipe the sheet down past the lower detent.
+
+- [ ] Both the keyboard and the sheet are dismissed. The host screen is undimmed and both "Open FormSheet" buttons are pressable again.
+
+---
+
+### fitToContents
+
+8. Tap "Open FormSheet (fitToContents)".
+
+- [ ] The sheet presents with a height matching its content.
+
+9. Tap "Bottom input".
+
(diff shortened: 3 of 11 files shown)


android/src/main/java/com/swmansion/rnscreens/modals/formsheet/native/coordinator/FormSheetDimensionsCoordinator.kt
```diff
@@ -114,21 +95,23 @@ internal class FormSheetDimensionsCoordinator(
...
     private fun resolveGeometry(sheetAvailableSpace: Int) {
+        val bottomInset = effectiveBottomInset
+
         currentDetents?.let { detents ->
             behaviorController?.updateSheetBehavior(
                 detents = detents,
                 sheetAvailableSpace = sheetAvailableSpace,
                 contentHeightForFitToContents = currentContentHeight,
-                nativeContainerPaddingBottom = lastBottomInset,
+                nativeContainerPaddingBottom = bottomInset,
```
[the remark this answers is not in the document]
t0maboro: the bottom inset (especially ime) is accounted internally by the behavior, so the translation works properly: https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java#L1730

apps/src/tests/single-feature-tests/form-sheet/test-form-sheet-keyboard/scenario.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Detents – keyboard at the lower detent
+
+2. Tap "Open FormSheet (detents)".
+
+- [ ] The sheet presents at the lower detent (0.6). "Top input" is visible at the top of the sheet; "Bottom input" and "Dismiss from JS" are not visible yet (the content box is laid out to the largest detent).
+
+3. Tap "Top input".
+
+- [ ] The keyboard slides in and the sheet moves up by the keyboard height, following the keyboard animation – no jump before or after it. "Top input" is focused and visible.
```
LKuchno: Actually currently when keyboard is extended  "Bottom input" and "Dismiss from JS".

https://github.com/user-attachments/assets/b51cfa94-eed8-4cf7-a54a-d05aa40d431a

If this is expected I would add information in this step about this.

apps/src/tests/single-feature-tests/form-sheet/test-form-sheet-keyboard/scenario.md
```diff
@@ -0,0 +1,138 @@
...
+
+4. Press the system back button (or use the back gesture).
+
+- [ ] The keyboard hides and the sheet moves back down together with it, settling at the lower detent (0.6). The sheet stays presented.
```
LKuchno: Note that: "Bottom input" and "Dismiss from JS" are not visible again.

LKuchno (commented):
Looks good, I left just one suggestion for Android part to avoid confusion what is expected.

apps/src/tests/single-feature-tests/form-sheet/test-form-sheet-keyboard/scenario.md
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4573/commits/14bff3034ec7f6bd78c002a90c2311b274da9cdd

apps/src/tests/single-feature-tests/form-sheet/test-form-sheet-keyboard/scenario.md
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4573/commits/14bff3034ec7f6bd78c002a90c2311b274da9cdd

apps/src/tests/single-feature-tests/form-sheet/test-form-sheet-keyboard/scenario.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Detents – keyboard at the lower detent
+
+2. Tap "Open FormSheet (detents)".
+
+- [ ] The sheet presents at the lower detent (0.6). "Top input" is visible at the top of the sheet; "Bottom input" and "Dismiss from JS" are not visible yet (the content box is laid out to the largest detent).
+
+3. Tap "Top input".
+
+- [ ] The keyboard slides in and the sheet moves up by the keyboard height, following the keyboard animation – no jump before or after it. "Top input" is focused and visible. On Android, the "Bottom input" and "Dismiss from JS" are moved to the visible area.
```
LKuchno, line 130: This step is under '## Steps - Android', so 'On Android' is redundant here.   

```suggestion
- [ ] The keyboard slides in and the sheet moves up by the keyboard height, following the keyboard animation – no jump before or after it. "Top input" is focused and visible. "Bottom input" and "Dismiss from JS" are moved to the visible area.
```

apps/src/tests/single-feature-tests/form-sheet/test-form-sheet-keyboard/scenario.md
```diff
@@ -0,0 +1,138 @@
...
+
+4. Press the system back button (or use the back gesture).
+
+- [ ] The keyboard hides and the sheet moves back down together with it, settling at the lower detent (0.6). The sheet stays presented. On Android, the "Bottom input" and "Dismiss from JS" are moved outside the visible area.
```
LKuchno, line 134: Same here:

```suggestion
- [ ] The keyboard hides and the sheet moves back down together with it, settling at the lower detent (0.6). The sheet stays presented. "Bottom input" and "Dismiss from JS" are moved outside the visible area.
```

LKuchno (approved):
LGTM!
You can remove 'On Android' as updated steps are under '## Steps - Android'.

kligarski (commented):
I tested the changes with additional generated test and I found some problems.

1. Middle detent of 3 and single detent doesn't seem to react to keyboard.

https://github.com/user-attachments/assets/7aa79819-6b8d-4f85-b39c-9f408c666df8

https://github.com/user-attachments/assets/2ceb850f-2e83-440b-9fb3-17d5eb56e284

2. Might be related to first one - shouldn't the sheet expand to full in this scenario?

https://github.com/user-attachments/assets/3ccce9ca-653d-49e7-a0d4-cf0c1c726adb

3. If both bottom screen and sheet have keyboard focus, when sheet is presented/dismissed, the keyboard flashes - not sure if this was also the case in previous implementation?

https://github.com/user-attachments/assets/8e4822f3-bc22-4c4d-b765-3d459e94d288

4. This might be a problem in the test itself but I wanted to add it here in case you've seen it before. When I added nestedScrollEnabled, the sheet doesn't respond to touches. Only when I toggled the prop in runtime via fast refresh, it starts working.

https://github.com/user-attachments/assets/910b7cd6-a637-4db4-a48f-807afeaf4598

5. Sheet below nested sheet also receives ime inset - not sure if this is intended. Additionally when nested sheet is dismissed, the sheet below doesn't animate back.

https://github.com/user-attachments/assets/4016bdb4-806a-49cd-925d-8065ca8cfbf1

https://github.com/user-attachments/assets/3a8b3cf2-d081-46e0-98b8-f2b073302d5c

<details>
<summary>Here is the generated test screen I used</summary>

```tsx
import React, { useCallback, useEffect, useRef, useState } from 'react';
import {
  Keyboard,
  Pressable,
  ScrollView,
  StyleSheet,
  Text,
  TextInput,
  type TextInputInstance,
  View,
} from 'react-native';
import { FormSheet } from 'react-native-screens';
import { createScenario } from '@apps/tests/shared/helpers';
import { Colors } from '@apps/shared/styling';

type Detents = number[] | 'fitToContents' | undefined;

type InputRef = { current: TextInputInstance | null };

type SheetPreset = {
  key: string;
  title: string;
  hint: string;
  detents: Detents;
  /**
   * Number of filler blocks the sheet starts with. Drives the content height,
   * which is the input of `fitToContents` measuring and of the "does the content
   * still fit above the keyboard" cases.
   */
  blocks: number;
  initialDetentIndex?: number | 'last';
  preventNativeDismiss?: boolean;
  /** Wraps the body in a ScrollView. Not usable together with `fitToContents`. */
  scrollable?: boolean;
  /** Pushes the last input to the very bottom of the sheet with a flex spacer. */
  bottomAnchored?: boolean;
};

const PRESETS: SheetPreset[] = [
  {
    key: 'kb-two-detents',
    title: 'Two detents [0.5, 1.0]',
    hint: 'Baseline. At 0.5 the sheet moves up by the keyboard height; at 1.0 the content box shrinks.',
    detents: [0.5, 1.0],
    blocks: 3,
  },
  {
    key: 'kb-tiny-detent',
    title: 'Single tiny detent [0.2]',
    hint: 'The sheet is shorter than the keyboard. Nothing should be pushed off-screen or clipped at the top.',
    detents: [0.2],
    blocks: 2,
  },
  {
    key: 'kb-near-full',
    title: 'Single detent [0.98]',
    hint: 'Just below full height: not the largest detent, but almost no room to move up into.',
    detents: [0.98],
    blocks: 8,
  },
  {
    key: 'kb-near-full-pair',
    title: 'Detents [0.5, 0.97, 1.0]',
    hint: 'The 0.97 / 1.0 boundary. Focus at 0.97 and at 1.0 and compare - the two must not fight each other.',
    detents: [0.5, 0.97, 1.0],
    blocks: 8,
  },
  {
    key: 'kb-three-detents',
    title: 'Detents [0.25, 0.6, 1.0]',
    hint: 'Three detents (the maximum Android supports). Focus an input while resting at each one and check the reported index.',
    detents: [0.25, 0.6, 1.0],
    blocks: 8,
  },
  {
    key: 'kb-three-detents-2',
    title: 'Detents [0.25, 0.3, 1.0]',
    hint: 'Three detents (the maximum Android supports). Focus an input while resting at each one and check the reported index.',
    detents: [0.25, 0.3, 1.0],
    blocks: 8,
  },
  {
    key: 'kb-full-only',
    title: 'Single detent [1.0], bottom-anchored input',
    hint: 'Largest detent only. The bottom input sits at the very bottom - exactly where the content box has to shrink.',
    detents: [1.0],
    blocks: 3,
    bottomAnchored: true,
  },
  {
    key: 'kb-initial-last',
    title: 'Detents [0.4, 1.0], initialDetentIndex: last',
    hint: 'Opens directly at 1.0. Combine with focus mode "autoFocus" - the keyboard animates during the presentation.',
    detents: [0.4, 1.0],
    initialDetentIndex: 'last',
    blocks: 5,
  },
  {
    key: 'kb-fit-short',
    title: 'fitToContents, tiny content',
    hint: 'A short sheet that must be lifted above the keyboard as a whole.',
    detents: 'fitToContents',
    blocks: 1,
  },
  {
    key: 'kb-fit-medium',
    title: 'fitToContents, content that still fits',
    hint: 'Tall enough to matter, short enough to fit above the keyboard. Grow it (+1) until it stops fitting.',
    detents: 'fitToContents',
    blocks: 5,
  },
  {
    key: 'kb-fit-overflow',
    title: 'fitToContents, content taller than the screen',
    hint: 'Content overflows the safe area even without the keyboard; the keyboard takes away even more room.',
    detents: 'fitToContents',
    blocks: 30,
  },
  {
    key: 'kb-scroll',
    title: 'Detents [0.6, 1.0] + ScrollView',
    hint: 'Scrolling and keyboard insets at once. Scroll the focused input under/over the keyboard.',
    detents: [0.6, 1.0],
    blocks: 14,
    scrollable: true,
  },
  {
    key: 'kb-prevent-dismiss',
    title: 'Detents [0.5, 1.0], preventNativeDismiss',
    hint: 'Swipe the sheet down hard with the keyboard shown - the sheet must bounce back and not get stuck.',
    detents: [0.5, 1.0],
    preventNativeDismiss: true,
    blocks: 3,
  },
  {
    key: 'kb-default-detents',
    title: 'No detents prop (default)',
    hint: 'Whatever the default resolves to, it has to react to the keyboard the same way an explicit detent does.',
    detents: undefined,
    blocks: 4,
  },
];

/**
 * Detent configurations applied to an already presented sheet by the "cycle" chip.
 * Includes the array <-> `fitToContents` switch, which changes the native view tree.
 *
 * Keep every entry at 3 detents or fewer - Android throws for anything above that.
 */
const DETENT_CYCLE: Detents[] = [
  [0.25],
  [0.5, 1.0],
  [0.97],
  [0.97, 1.0],
  [1.0],
  'fitToContents',
];

type FocusMode = 'off' | 'autoFocus' | 'onDidAppear';

const FOCUS_MODES: FocusMode[] = ['off', 'autoFocus', 'onDidAppear'];

const FOCUS_MODE_HINTS: Record<FocusMode, string> = {
  off: 'Nothing is focused when the sheet opens.',
  autoFocus:
    'The top input mounts with autoFocus - the keyboard animates while the sheet is still PRESENTING, so keyboard tracking is not enabled yet.',
  onDidAppear:
    'The top input is focused from onDidAppear - the keyboard animates once the sheet is PRESENTED and tracking is enabled.',
};

/** Delay used by the "race the keyboard animation" chips - lands mid keyboard animation. */
const RACE_DELAY_MS = 250;

function describeDetents(detents: Detents): string {
  if (detents === undefined) {
    return 'default';
  }
  if (detents === 'fitToContents') {
    return 'fitToContents';
  }
  return `[${detents.join(', ')}]`;
}

function useKeyboardMetrics() {
  const [metrics, setMetrics] = useState({ isVisible: false, height: 0 });

  useEffect(() => {
    const showSubscription = Keyboard.addListener('keyboardDidShow', event =>
      setMetrics({
        isVisible: true,
        height: Math.round(event.endCoordinates.height),
      }),
    );
    const hideSubscription = Keyboard.addListener('keyboardDidHide', () =>
      setMetrics({ isVisible: false, height: 0 }),
    );

    return () => {
      showSubscription.remove();
      hideSubscription.remove();
    };
  }, []);

  return metrics;
}

function Chip({
  label,
  onPress,
  isActive,
}: {
  label: string;
  onPress: () => void;
  isActive?: boolean;
}) {
  return (
    <Pressable
      testID={`chip-${label}`}
      style={[styles.chip, isActive === true && styles.chipActive]}
      onPress={onPress}>
      <Text
        style={[styles.chipLabel, isActive === true && styles.chipLabelActive]}>
        {label}
      </Text>
    </Pressable>
  );
}

/**
 * Controls are laid out in horizontally scrollable strips rather than wrapped rows,
 * so that they stay ~40dp tall. Otherwise they would eat the whole content box of
 * the small-detent presets and push the inputs out of the visible area.
 */
function ChipStrip({ children }: { children: React.ReactNode }) {
  return (
    <ScrollView
      horizontal
      style={styles.chipStrip}
      contentContainerStyle={styles.chipStripContent}
      showsHorizontalScrollIndicator={false}
      keyboardShouldPersistTaps="handled">
      {children}
    </ScrollView>
  );
}

function StripLabel({ label }: { label: string }) {
  return <Text style={styles.stripLabel}>{label}</Text>;
}

type SheetBodyProps = {
  preset: SheetPreset;
  detents: Detents;
  detentIndex: number | null;
  blocks: number;
  setBlocks: (updater: (previous: number) => number) => void;
  isPulsing: boolean;
  onTogglePulse: () => void;
  keyboardHeight: number;
  isKeyboardVisible: boolean;
  focusMode: FocusMode;
  topInputRef: InputRef;
  bottomInputRef: InputRef;
  numericInputRef: InputRef;
  onCycleDetents: () => void;
  onResetDetents: () => void;
  onDismiss: () => void;
  onOpenNested: () => void;
  onFocusThen: (action: 'dismiss' | 'grow' | 'detents' | 'nested') => void;
  onRapidFocusBlur: () => void;
  onLater: (action: () => void, delayMs: number) => void;
};

function SheetBody({
  preset,
  detents,
  detentIndex,
  blocks,
  setBlocks,
  isPulsing,
  onTogglePulse,
  keyboardHeight,
  isKeyboardVisible,
  focusMode,
  topInputRef,
  bottomInputRef,
  numericInputRef,
  onCycleDetents,
  onResetDetents,
  onDismiss,
  onOpenNested,
  onFocusThen,
  onRapidFocusBlur,
  onLater,
}: SheetBodyProps) {
  const isFitToContents = detents === 'fitToContents';
  // A flex spacer only has room to expand when the body fills a fixed-height sheet.
  const isBottomAnchored = preset.bottomAnchored === true && !isFitToContents;

  const content = (
    <>
      <Text style={styles.sheetTitle}>{preset.title}</Text>
      <Text style={styles.status}>
        {describeDetents(detents)} · idx{' '}
        {detentIndex === null ? '—' : detentIndex} · blocks {blocks} · kb{' '}
        {isKeyboardVisible ? `${keyboardHeight}px` : 'hidden'} · {focusMode}
      </Text>

      {/* The primary input goes first so that it stays reachable even in a sheet
          that is only a fraction of the screen tall. */}
      <TextInput
        testID="sheet-top-input"
        ref={topInputRef}
        autoFocus={focusMode === 'autoFocus'}
        style={styles.input}
        placeholder="Top input"
        placeholderTextColor={Colors.NavyLight60}
      />

      <ChipStrip>
        <StripLabel label="size" />
        <Chip
          label="-1"
          onPress={() => setBlocks(value => Math.max(0, value - 1))}
        />
        <Chip label="+1" onPress={() => setBlocks(value => value + 1)} />
        <Chip label="1" onPress={() => setBlocks(() => 1)} />
        <Chip label="5" onPress={() => setBlocks(() => 5)} />
        <Chip label="12" onPress={() => setBlocks(() => 12)} />
        <Chip label="30" onPress={() => setBlocks(() => 30)} />
        <Chip
          label="pulse 2<->10"
          isActive={isPulsing}
          onPress={onTogglePulse}
        />
        <Chip
          label="grow @2s"
          onPress={() => onLater(() => setBlocks(value => value + 6), 2000)}
        />
        <Chip
          label="shrink @2s"
          onPress={() => onLater(() => setBlocks(() => 1), 2000)}
        />
        <StripLabel label="detents" />
        <Chip label="cycle" onPress={onCycleDetents} />
        <Chip label="cycle @2s" onPress={() => onLater(onCycleDetents, 2000)} />
        <Chip label="reset" onPress={onResetDetents} />
      </ChipStrip>

      <ChipStrip>
        <StripLabel label="focus" />
        <Chip label="top" onPress={() => topInputRef.current?.focus()} />
        <Chip
          label="numeric"
          onPress={() => numericInputRef.current?.focus()}
        />
        <Chip label="bottom" onPress={() => bottomInputRef.current?.focus()} />
        <Chip label="hide kb" onPress={() => Keyboard.dismiss()} />
        <Chip label="rapid focus/blur" onPress={onRapidFocusBlur} />
        <StripLabel label={`race +${RACE_DELAY_MS}ms`} />
        <Chip label="focus -> close" onPress={() => onFocusThen('dismiss')} />
        <Chip label="focus -> grow" onPress={() => onFocusThen('grow')} />
        <Chip label="focus -> detents" onPress={() => onFocusThen('detents')} />
        <Chip label="focus -> nested" onPress={() => onFocusThen('nested')} />
        <StripLabel label="sheet" />
        <Chip label="nested" onPress={onOpenNested} />
        <Chip label="close" onPress={onDismiss} />
        <Chip label="close @2s" onPress={() => onLater(onDismiss, 2000)} />
      </ChipStrip>

      <TextInput
        testID="sheet-numeric-input"
        ref={numericInputRef}
        style={styles.input}
        keyboardType="numeric"
        placeholder="Numeric input (different keyboard height)"
        placeholderTextColor={Colors.NavyLight60}
      />

      <TextInput
        testID="sheet-multiline-input"
        style={[styles.input, styles.multilineInput]}
        multiline
        placeholder="Multiline input - type to make the content grow while the keyboard is up"
        placeholderTextColor={Colors.NavyLight60}
      />

      {Array.from({ length: blocks }, (_, index) => (
        <View key={index} style={styles.block}>
          <Text style={styles.blockLabel}>block {index + 1}</Text>
        </View>
      ))}

      {isBottomAnchored && <View style={styles.flexSpacer} />}

      <TextInput
        testID="sheet-bottom-input"
        ref={bottomInputRef}
        style={styles.input}
        placeholder="Bottom input"
        placeholderTextColor={Colors.NavyLight60}
      />
    </>
  );

  if (preset.scrollable === true && !isFitToContents) {
    return (
      <View style={styles.sheetBodyFill}>
        <ScrollView
          nestedScrollEnabled={true}
          style={styles.sheetScroll}
          contentContainerStyle={styles.sheetScrollContent}
          keyboardShouldPersistTaps="handled">
          {content}
        </ScrollView>
      </View>
    );
  }

  return (
    <View
      style={[
        styles.sheetBody,
        isFitToContents ? styles.sheetBodyWrap : styles.sheetBodyFill,
      ]}>
      {content}
    </View>
  );
}

function TestFormSheetKeyboardStress() {
  const [activePreset, setActivePreset] = useState<SheetPreset | null>(null);
  const [isOpen, setIsOpen] = useState(false);
  const [isNestedOpen, setIsNestedOpen] = useState(false);
  // Bumped on every open. Only used as the body key in the `autoFocus` mode, where
  // the body has to remount for `autoFocus` to fire again; in the other modes the body
  // is kept mounted so opening the sheet does not re-measure content from scratch.
  const [session, setSession] = useState(0);
  const [blocks, setBlocks] = useState(0);
  const [isPulsing, setIsPulsing] = useState(false);
  const [detentCycleIndex, setDetentCycleIndex] = useState<number | null>(null);
  const [detentIndex, setDetentIndex] = useState<number | null>(null);
  const [focusMode, setFocusMode] = useState<FocusMode>('off');

  const topInputRef = useRef<TextInputInstance | null>(null);
  const bottomInputRef = useRef<TextInputInstance | null>(null);
  const numericInputRef = useRef<TextInputInstance | null>(null);
  const timersRef = useRef<ReturnType<typeof setTimeout>[]>([]);

  const keyboard = useKeyboardMetrics();

  useEffect(
    () => () => {
      timersRef.current.forEach(clearTimeout);
      timersRef.current = [];
    },
    [],
  );

  const runLater = useCallback((action: () => void, delayMs: number) => {
    timersRef.current.push(setTimeout(action, delayMs));
  }, []);

  const detents =
    detentCycleIndex === null
      ? activePreset?.detents
      : DETENT_CYCLE[detentCycleIndex];

  useEffect(() => {
    if (!isPulsing || !isOpen) {
      return;
    }
    const intervalId = setInterval(
      () => setBlocks(value => (value > 5 ? 2 : 10)),
      1200,
    );
    return () => clearInterval(intervalId);
  }, [isPulsing, isOpen]);

  const openPreset = useCallback((preset: SheetPreset) => {
    setActivePreset(preset);
    setBlocks(preset.blocks);
    setIsPulsing(false);
    setDetentCycleIndex(null);
    setDetentIndex(null);
    setSession(value => value + 1);
    setIsOpen(true);
  }, []);

  const closeSheet = useCallback(() => setIsOpen(false), []);

  const cycleDetents = useCallback(
    () =>
      setDetentCycleIndex(value =>
        value === null ? 0 : (value + 1) % DETENT_CYCLE.length,
      ),
    [],
  );

  const resetDetents = useCallback(() => setDetentCycleIndex(null), []);

  const openNested = useCallback(() => setIsNestedOpen(true), []);

  const focusThen = useCallback(
    (action: 'dismiss' | 'grow' | 'detents' | 'nested') => {
      topInputRef.current?.focus();
      runLater(() => {
        switch (action) {
          case 'dismiss':
            closeSheet();
            break;
          case 'grow':
            setBlocks(value => value + 8);
            break;
          case 'detents':
            cycleDetents();
            break;
          case 'nested':
            openNested();
            break;
        }
      }, RACE_DELAY_MS);
    },
    [runLater, closeSheet, cycleDetents, openNested],
  );

  const rapidFocusBlur = useCallback(() => {
    topInputRef.current?.focus();
    runLater(() => Keyboard.dismiss(), 150);
    runLater(() => bottomInputRef.current?.focus(), 300);
    runLater(() => Keyboard.dismiss(), 450);
    runLater(() => topInputRef.current?.focus(), 600);
  }, [runLater]);

  const handleDidAppear = useCallback(() => {
    if (focusMode === 'onDidAppear') {
      topInputRef.current?.focus();
    }
  }, [focusMode]);

  return (
      <ScrollView
        style={styles.list}
        contentContainerStyle={styles.listContent}
        keyboardShouldPersistTaps="handled">
        <Text style={styles.title}>FormSheet x keyboard stress (Android)</Text>

        <Text style={styles.groupLabel}>
          Focus mode (applies to the next opened sheet)
        </Text>
        <View style={styles.chipRow}>
          {FOCUS_MODES.map(mode => (
            <Chip
              key={mode}
              label={mode}
              isActive={focusMode === mode}
              onPress={() => setFocusMode(mode)}
            />
          ))}
        </View>
        <Text style={styles.hint}>{FOCUS_MODE_HINTS[focusMode]}</Text>

        <TextInput
          testID="host-input"
          style={styles.input}
          placeholder="Host input - focus it, then open a sheet with the keyboard already up"
          placeholderTextColor={Colors.NavyLight60}
        />
        <Text style={styles.status}>
          keyboard: {keyboard.isVisible ? `${keyboard.height}px` : 'hidden'}
        </Text>

        {PRESETS.map(preset => (
          <Pressable
            key={preset.key}
            testID={preset.key}
            style={styles.presetButton}
            onPress={() => openPreset(preset)}>
            <Text style={styles.presetTitle}>{preset.title}</Text>
            <Text style={styles.presetHint}>{preset.hint}</Text>
          </Pressable>
        ))}
      </ScrollView>

      <FormSheet
        isOpen={isOpen}
        detents={detents}
        initialDetentIndex={activePreset?.initialDetentIndex}
        preventNativeDismiss={activePreset?.preventNativeDismiss}
        prefersGrabberVisible
        onDidAppear={handleDidAppear}
        onDetentChanged={event => setDetentIndex(event.nativeEvent.index)}
        onNativeDismiss={closeSheet}>
        {activePreset !== null && (
          <SheetBody
            key={focusMode === 'autoFocus' ? `autofocus-${session}` : 'stable'}
            preset={activePreset}
            detents={detents}
            detentIndex={detentIndex}
            blocks={blocks}
            setBlocks={setBlocks}
            isPulsing={isPulsing}
            onTogglePulse={() => setIsPulsing(value => !value)}
            keyboardHeight={keyboard.height}
            isKeyboardVisible={keyboard.isVisible}
            focusMode={focusMode}
            topInputRef={topInputRef}
            bottomInputRef={bottomInputRef}
            numericInputRef={numericInputRef}
            onCycleDetents={cycleDetents}
            onResetDetents={resetDetents}
            onDismiss={closeSheet}
            onOpenNested={openNested}
            onFocusThen={focusThen}
            onRapidFocusBlur={rapidFocusBlur}
            onLater={runLater}
          />
        )}
      </FormSheet>

      <FormSheet
        isOpen={isNestedOpen}
        detents={[0.5, 1.0]}
        onNativeDismiss={() => setIsNestedOpen(false)}>
        <View style={[styles.sheetBody, styles.sheetBodyFill]}>
          <Text style={styles.sheetTitle}>Nested sheet</Text>
          <TextInput
            testID="nested-input"
            style={styles.input}
            placeholder="Nested input"
            placeholderTextColor={Colors.NavyLight60}
          />
          <ChipStrip>
            <Chip label="close nested" onPress={() => setIsNestedOpen(false)} />
            <Chip
              label="close both"
              onPress={() => {
                setIsNestedOpen(false);
                closeSheet();
              }}
            />
          </ChipStrip>
          <Text style={styles.hint}>
            Stacked on top of the sheet below. Focus here, then dismiss this
            sheet with the keyboard still shown - the sheet underneath has to
            settle back at its own detent.
          </Text>
        </View>
      </FormSheet>
    </View>
  );
}

  container: {
    flex: 1,
    backgroundColor: Colors.offBackground,
  },
  list: {
    flex: 1,
  },
  listContent: {
    padding: 16,
    paddingBottom: 48,
  },
  title: {
    fontSize: 20,
    fontWeight: 'bold',
    color: Colors.text,
    marginBottom: 12,
  },
  hint: {
    fontSize: 12,
    color: Colors.NavyLight80,
    marginTop: 8,
  },
  status: {
    fontSize: 11,
    color: Colors.NavyLight100,
    marginTop: 2,
  },
  groupLabel: {
    fontSize: 12,
    fontWeight: '600',
    color: Colors.NavyLight80,
    marginTop: 12,
    marginBottom: 4,
  },
  chipRow: {
    flexDirection: 'row',
    flexWrap: 'wrap',
    gap: 6,
  },
  chipStrip: {
    flexGrow: 0,
    marginTop: 6,
  },
  chipStripContent: {
    alignItems: 'center',
    gap: 6,
    paddingRight: 12,
  },
  stripLabel: {
    fontSize: 10,
    fontWeight: '700',
    textTransform: 'uppercase',
    color: Colors.NavyLight60,
    marginLeft: 4,
  },
  chip: {
    paddingHorizontal: 10,
    paddingVertical: 6,
    borderRadius: 14,
    borderWidth: 1,
    borderColor: Colors.cardBorder,
    backgroundColor: Colors.cardBackground,
  },
  chipActive: {
    backgroundColor: Colors.primary,
    borderColor: Colors.primary,
  },
  chipLabel: {
    fontSize: 12,
    color: Colors.primary,
  },
  chipLabelActive: {
    color: Colors.White,
  },
  presetButton: {
    padding: 12,
    marginTop: 8,
    borderRadius: 8,
    borderWidth: 1,
    borderColor: Colors.cardBorder,
    backgroundColor: Colors.cardBackground,
  },
  presetTitle: {
    fontSize: 15,
    fontWeight: '600',
    color: Colors.text,
  },
  presetHint: {
    fontSize: 12,
    color: Colors.NavyLight80,
    marginTop: 2,
  },
  sheetBody: {
    backgroundColor: Colors.background,
    paddingHorizontal: 12,
    paddingTop: 8,
    paddingBottom: 12,
  },
  sheetBodyFill: {
    flex: 1,
    backgroundColor: Colors.background,
  },
  sheetBodyWrap: {
    paddingBottom: 20,
  },
  sheetScroll: {
    flex: 1,
  },
  sheetScrollContent: {
    paddingHorizontal: 12,
    paddingTop: 8,
    paddingBottom: 24,
  },
  sheetTitle: {
    fontSize: 13,
    fontWeight: '600',
    color: Colors.text,
  },
  input: {
    height: 44,
    borderWidth: 1,
    borderColor: Colors.cardBorder,
    borderRadius: 8,
    paddingHorizontal: 12,
    marginTop: 6,
    color: Colors.text,
    backgroundColor: Colors.cardBackground,
  },
  multilineInput: {
    height: undefined,
    minHeight: 44,
    paddingTop: 12,
    paddingBottom: 12,
  },
  block: {
    height: 40,
    marginTop: 6,
    borderRadius: 6,
    justifyContent: 'center',
    paddingHorizontal: 12,
    backgroundColor: Colors.PurpleLight40,
  },
  blockLabel: {
    fontSize: 12,
    color: Colors.text,
  },
  flexSpacer: {
    flex: 1,
    minHeight: 12,
  },
});

export default createScenario(TestFormSheetKeyboardStress, scenarioDescription);
```

And a checklist (also generated, might not be valid 100%):

```md
# FormSheet × keyboard — manual test plan (Android)

Screen: **FormSheet → Keyboard Stress (Android)**
(`test-form-sheet-keyboard-stress-android`).

Android only. Emulator must use the on-screen keyboard
(*Settings → System → Languages & input → Physical keyboard → Use on-screen
keyboard*, or unplug the hardware keyboard in AVD settings).

---

## Reference behavior

| Sheet state | expected |
| --- | --- |
| resting **below** the largest detent | the whole sheet moves up by the keyboard height |
| resting **at** the largest detent | the content box shrinks to the area above the keyboard; the sheet stays put |
| `fitToContents` | the whole sheet moves up above the keyboard |

In all three cases the sheet must move **together with the keyboard animation** —
driven by `FormSheetKeyboardCoordinator`'s `WindowInsetsAnimation` callback, so:

- no jump *before* the animation starts,
- no correction *after* it ends,
- no clipping of the sheet's top edge or rounded corners,
- `translationY` back to exactly `0` when the animation finishes (a leftover
  offset shows up as the sheet sitting slightly too high/low afterwards).

Every step assumes: **no crash, no stuck sheet, no stuck backdrop, no sheet left
translated once the keyboard is gone.**

---

## Device matrix

Run at least section A on each row; run everything on one primary device.

- [ ] API 30+ (real `WindowInsetsAnimation`), **gesture navigation**
- [ ] API 30+, **3-button navigation** (different bottom inset)
- [ ] ⚠ API 29 or lower (`WindowInsetsAnimationCompat` fallback path — the
      `onProgress` driving `translationY` behaves differently there)
- [ ] A device/emulator with a **display cutout** or tall status bar

---

## A. Baseline — detents vs. keyboard

Focus mode: **off**.

- [ ] **Two detents [0.5, 1.0]** → tap **Top input** while resting at 0.5.
      The sheet moves up by the keyboard height, in sync with the keyboard.
- [ ] Same sheet → back button to hide the keyboard. The sheet moves back down
      with the keyboard and settles exactly at 0.5. The sheet is **not**
      dismissed by that back press.
- [ ] Drag to **1.0**, tap **Bottom input**. The sheet stays at 1.0 and the
      content box shrinks — the bottom input ends up right above the keyboard.
- [ ] Hide the keyboard at 1.0. The content box grows back to the full sheet
      height, bottom input returns to the bottom.
- [ ] **No detents prop (default)** → focus **Top input**. Same handling as the
      explicit-detent equivalent.
- [ ] **Detents [0.25, 0.6, 1.0]** → at *each* of the three detents, focus an
      input and hide the keyboard again. The `idx` value in the sheet header
      must match where the sheet actually rests, before and after.
- [ ] Same sheet → check the middle detent (0.6) behaves like a lower detent
      (moves up) and 1.0 behaves like the largest one (content box shrinks).

> Android supports at most **3** detents — `FormSheetBehaviorController` throws
> `IllegalStateException` above that, and nothing validates it on the JS side.
> Worth deciding in review whether that should fail louder / earlier.

## B. Extreme detents

- [ ] **Single tiny detent [0.2]** → focus **Top input**. The sheet is shorter
      than the keyboard. It must move fully above the keyboard, must not be
      pushed off the top of the screen, and must not be clipped.
- [ ] Same sheet → chip `hide kb`. The sheet returns exactly to 0.2, no double
      animation, no residual offset.
- [ ] ⚠ **Single detent [0.98]** → focus **Top input**. There is almost no room
      to move up into. Check the sheet does not overshoot the top inset and does
      not flicker between "move up" and "shrink content box" handling.
- [ ] ⚠ **Detents [0.5, 0.97, 1.0]** → focus at **0.97**, hide, then focus at
      **1.0**. 0.97 must be handled as a non-largest detent (move up) and 1.0 as
      the largest one (shrink). The sheet must not snap between them on its own.
- [ ] Same sheet → with the keyboard shown at 0.97, drag up to 1.0 and back down.
      Dragging while the keyboard is up must not desync the sheet position.
- [ ] **Single detent [1.0], bottom-anchored input** → focus the **Bottom
      input** (it sits at the very bottom). It must end up above the keyboard,
      not behind it.

## C. Content size changes — keyboard hidden (sanity pass first)

Use **fitToContents, content that still fits**.

- [ ] Chips `+1` / `-1` a few times — the sheet animates to each new height, no
      flicker, no double animation.
- [ ] Chip `30` — content overflows; the sheet clamps to the available safe area
      instead of growing past the top inset.
- [ ] Chip `1` — the sheet shrinks back down cleanly.

## D. Content size changes **while the keyboard is shown** ⚠

The "jump around" class. For each: focus an input first, keep the keyboard up,
then press the size chip.

- [ ] **fitToContents, content that still fits** → focus **Top input**, then
      `+1` repeatedly until the content stops fitting above the keyboard. The
      transition from "fits" to "does not fit" is smooth, the sheet stays above
      the keyboard, the top edge never goes under the status bar.
- [ ] Same, going back down: `-1` repeatedly. No leftover gap between the sheet
      and the keyboard.
- [ ] **fitToContents, content taller than the screen** → focus **Top input**.
      The content already overflows and the keyboard takes away more room. The
      sheet stays pinned to the safe area and is not pushed off-screen.
- [ ] Same sheet → chip `1` with the keyboard up. The sheet collapses to a small
      height and stays glued above the keyboard.
- [ ] ⚠ **fitToContents, tiny content** → focus **Top input**, then chip
      `pulse 2<->10`. Watch ~10 s. The sheet resizes repeatedly while the
      keyboard is up — it must stay above the keyboard and must not drift up or
      down cumulatively.
- [ ] Turn `pulse 2<->10` off. The sheet settles at a stable height.
- [ ] **Two detents [0.5, 1.0]** at 0.5 → focus **Top input**, then chip
      `grow @2s`. The content grows inside the shifted sheet; the sheet position
      and detent must not be disturbed.
- [ ] Same → chip `shrink @2s` with the keyboard up.
- [ ] **Single detent [1.0], bottom-anchored input** → focus **Bottom input**,
      then `+1` / `30`. The content box is already shrunk by the keyboard;
      growing the content must not push anything behind the keyboard.
- [ ] ⚠ Any `fitToContents` preset → focus the **multiline input** and type
      several lines. The content grows continuously while the keyboard is up;
      the sheet must follow smoothly, not stutter or fight the keyboard.
- [ ] ⚠ **fitToContents, tiny content** → turn on `pulse 2<->10` *first*, then
      focus **Top input** while it is pulsing. Keyboard animation and content
      resize start at the same time.

## E. Detents changed at runtime ⚠

The `cycle` chip walks: `[0.25]` → `[0.5, 1.0]` → `[0.97]` → `[0.97, 1.0]` →
`[1.0]` → `fitToContents`.

- [ ] **Two detents [0.5, 1.0]** → keyboard hidden, press `cycle` through the
      whole ring once. Every transition animates cleanly.
- [ ] ⚠ Same, but focus **Top input** first and `cycle` through the whole ring
      **with the keyboard up**. Each new configuration must re-settle correctly
      relative to the keyboard.
- [ ] ⚠ Specifically check the array → `fitToContents` step and the
      `fitToContents` → `[0.25]` step with the keyboard shown (the native view
      tree changes shape there).
- [ ] With the keyboard up, `cycle` to `[0.97]` then to `[1.0]` — the switch from
      "move up" to "shrink content box" handling must not double-animate.
- [ ] Chip `reset` with the keyboard up — returns to the preset's detents
      without losing the keyboard offset.
- [ ] ⚠ Focus **Top input**, then chip `cycle @2s` — the detents change *while*
      the keyboard animation is still running.

## F. Autofocus / focus timing ⚠

This is the `isTrackingEnabled` boundary: in `autoFocus` mode the keyboard
animation starts while the sheet is still `PRESENTING`, so keyboard tracking is
deliberately off; in `onDidAppear` mode it starts once the sheet is `PRESENTED`
and tracking is on. **The end state must be identical in both modes.**

Set the focus mode chip at the top of the host screen *before* opening a sheet.

**Mode `autoFocus`:**

- [ ] ⚠ **Two detents [0.5, 1.0]**
- [ ] ⚠ **Single tiny detent [0.2]**
- [ ] ⚠ **Single detent [0.98]**
- [ ] ⚠ **Detents [0.4, 1.0], initialDetentIndex: last** (opens directly at 1.0)
- [ ] ⚠ **fitToContents, tiny content**
- [ ] ⚠ **fitToContents, content taller than the screen**
- [ ] For each of the above: the sheet ends up correctly positioned with the
      keyboard shown — no jump at the end of the presentation animation, no
      sheet overlapping the keyboard, no sheet stuck part-way.
- [ ] For each: hide the keyboard afterwards — the sheet returns to its detent
      correctly. A missed start position shows up here as a wrong resting spot.
- [ ] ⚠ Open one of them, dismiss it while the keyboard is still up, and
      immediately open it again.

**Mode `onDidAppear`:**

- [ ] **Two detents [0.5, 1.0]**
- [ ] **fitToContents, content that still fits**
- [ ] **Detents [0.4, 1.0], initialDetentIndex: last**
- [ ] Compare each against the same preset in `autoFocus` mode — same end state,
      only the timing differs.

**Keyboard already up before the sheet opens:**

- [ ] Focus the **host input** at the top of the screen and — without hiding the
      keyboard — open **Two detents [0.5, 1.0]**. The sheet presents correctly
      with the keyboard already visible.
- [ ] Same with **fitToContents, content that still fits**.
- [ ] Same with **Single tiny detent [0.2]**.
- [ ] Same in focus mode `autoFocus` — the keyboard stays up across the
      transition while focus moves from the host input into the sheet.

## G. Racing the keyboard animation ⚠

The `focus -> …` chips focus the top input and fire the action 250 ms later,
mid keyboard animation. This targets the exit animator now starting from the
live `translationY` instead of `0`.

- [ ] ⚠ **Two detents [0.5, 1.0]** → `focus -> close`. The sheet slides out from
      wherever it currently is — no teleport to the bottom, no visible jump at
      the start of the exit animation.
- [ ] ⚠ **fitToContents, content that still fits** → `focus -> close`.
- [ ] ⚠ **Single detent [0.98]** → `focus -> close`.
- [ ] ⚠ **Single tiny detent [0.2]** → `focus -> close`.
- [ ] ⚠ `focus -> grow` on **fitToContents, content that still fits**.
- [ ] ⚠ `focus -> detents` on **Two detents [0.5, 1.0]**.
- [ ] ⚠ `focus -> nested` on **Two detents [0.5, 1.0]** — a second sheet stacks
      on top while the keyboard is still animating.
- [ ] ⚠ Chip `rapid focus/blur` on **Two detents [0.5, 1.0]** (focus → blur →
      focus → blur → focus at 150 ms steps). The sheet must not accumulate
      offset; the final state matches "keyboard shown".
- [ ] ⚠ Chip `rapid focus/blur` on **fitToContents, tiny content**.
- [ ] ⚠ Chip `rapid focus/blur` on **Single detent [1.0], bottom-anchored
      input** (the shrink/grow path rather than the translate path).
- [ ] Chip `close @2s` — focus an input first so the sheet dismisses with the
      keyboard fully shown. Keyboard and sheet go away together.

## H. Keyboard height changes without a hide/show ⚠

The IME inset changes without a full hide → show cycle, so the animation starts
from a non-zero inset.

- [ ] ⚠ **Two detents [0.5, 1.0]** → chip `top`, then chip `numeric` (a shorter
      keyboard). The sheet adjusts to the new keyboard height — no gap, no
      overlap.
- [ ] ⚠ Then chip `bottom` (back to the default keyboard).
- [ ] ⚠ Same sequence on **fitToContents, content that still fits**.
- [ ] ⚠ Same sequence on **Single detent [1.0], bottom-anchored input**.
- [ ] With a keyboard focused, open/close the emoji or clipboard panel of the
      IME (Gboard) — any height change must be tracked.
- [ ] Switch IME language with the globe key while an input is focused.

## I. Scroll views

- [ ] **Detents [0.6, 1.0] + ScrollView** → focus **Top input**, then scroll the
      content. Scrolling works while the keyboard is up; insets are correct.
- [ ] Same → scroll down to the **Bottom input** and focus it.
- [ ] Same → with the keyboard up, drag the sheet between 0.6 and 1.0.
- [ ] Same → scroll to the top of the ScrollView and keep dragging down with the
      keyboard shown — the sheet drag takes over without losing the offset.
- [ ] Same → horizontal chip strips inside the sheet still scroll while the
      keyboard is up (the coordinator uses `DISPATCH_MODE_STOP`, so verify inner
      views are not starved of touch/inset handling).

## J. Dismissal with the keyboard shown

- [ ] **Two detents [0.5, 1.0]** → focus an input, then swipe the sheet down
      past the lowest detent. Keyboard and sheet go away together, the backdrop
      clears, the host screen is interactive again.
- [ ] Same → focus an input, then chip `close` (JS dismissal).
- [ ] Same → focus an input, then tap the backdrop above the sheet.
- [ ] Same → focus an input, then press the system **back button twice** (first
      press hides the keyboard, second dismisses the sheet).
- [ ] Same → focus an input, then use the **back gesture** instead. The keyboard
      hides; the same gesture must not also dismiss the sheet.
- [ ] ⚠ **Detents [0.5, 1.0], preventNativeDismiss** → focus an input, then
      swipe down hard. The sheet bounces back to its detent, the keyboard state
      stays consistent, and the sheet is not left translated or stuck.
- [ ] Same → repeat the hard swipe 3–4 times in a row with the keyboard up.
- [ ] Same → focus an input, then back button — the keyboard hides but the sheet
      stays presented.
- [ ] **fitToContents, content taller than the screen** → focus an input, then
      swipe down to dismiss.

## K. Stacked sheets

- [ ] **Two detents [0.5, 1.0]** → chip `nested`, focus the **Nested input**.
      The nested sheet handles the keyboard; the sheet below stays put.
- [ ] With the nested keyboard up, chip `close nested`. The sheet underneath
      settles back at its own detent and the keyboard state stays consistent.
- [ ] Focus the parent's **Top input**, keep the keyboard up, then chip `nested`
      — the nested sheet presents over a keyboard-offset parent.
- [ ] With the nested sheet's keyboard up, chip `close both`.
- [ ] ⚠ **fitToContents, content that still fits** as the parent → repeat the
      first two items.
- [ ] ⚠ **Single tiny detent [0.2]** as the parent → repeat the first two items.

## L. Interruptions & environment

- [ ] **Two detents [0.5, 1.0]** → focus an input, then rotate the device with
      the keyboard shown (portrait ↔ landscape). Sheet and keyboard stay
      consistent; rotate back.
- [ ] **fitToContents, content that still fits** → same rotation test.
- [ ] Landscape only: repeat section A step 1 (the keyboard is proportionally
      much taller relative to the screen there).
- [ ] Focus an input, background the app (Home), foreground it again. Sheet
      position and keyboard state are restored consistently.
- [ ] Focus an input, then open the notification shade and dismiss it.
- [ ] Focus an input, then trigger a configuration change (dark mode toggle /
      font size change in developer settings).
- [ ] Split-screen / multi-window: open a sheet and focus an input.
- [ ] Open and dismiss 5 sheets in a row with the keyboard up each time — no
      leaked backdrops, no growing lag, no leaked inset-animation callbacks.
```

</details>

t0maboro:
> I tested the changes with additional generated test and I found some problems. [...]

1 & 2. https://github.com/software-mansion/react-native-screens/pull/4573/commits/df04275fbfb44f5f063c1836835064f00c145f66

3. It's the IME re-targeting between windows when the sheet's window goes away; similar behavior observed for v4 & native dialog

https://github.com/user-attachments/assets/65ed503b-95d5-4a77-8ffc-9b6c06fe7a86

https://github.com/user-attachments/assets/a1f24aa4-e136-4dca-b4ef-533409253689

4. Reproduced without keyboard changes; unrelated to keyboard changes tracked separately in https://github.com/software-mansion/react-native-screens-labs/issues/1815

5. each dialog window receives the IME insets and only the target window gets the animation, which is exactly how a native stacked BottomSheetDialog behaves

https://github.com/user-attachments/assets/5a192ba9-0351-47fd-ae3a-5ded4666e186

apps/src/tests/single-feature-tests/form-sheet/test-form-sheet-keyboard/scenario.md
```diff
@@ -0,0 +1,202 @@
...
+  - **Android:** the sheet reacts to the keyboard insets. Whichever detent it rests at, the sheet is pushed up by the keyboard height as far as the screen allows, so its detent-sized part stays above the keyboard (the extra area is hidden behind the keyboard, like on iOS). Only the part that can't be pushed above the keyboard shrinks the content box - a sheet resting at the full-height detent stays in place and its content box shrinks by the keyboard height. A `fitToContents` sheet moves up as a whole. The sheet moves together with the keyboard animation.
+
+## Steps - iPhone
+
+### Baseline
+
+1. Launch the app and navigate to the **Keyboard Integration** screen.
+
+- [ ] The host screen shows the "FormSheet Test" title and the four "Open FormSheet" buttons: detents, three detents, single detent, fitToContents.
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4573/changes/54f61db4ef599f8c275157c465fd00adb3a46dad

android/src/main/java/com/swmansion/rnscreens/modals/formsheet/native/model/FormSheetDetents.kt
```diff
@@ -33,58 +33,82 @@ internal class FormSheetDetents(
...
+     */
+    internal fun halfExpandedRatio(
+        containerHeight: Int,
+        keyboardLift: Int = 0,
+    ): Float {
         check(count == MAX_DETENTS) { "[RNScreens] Exactly $MAX_DETENTS detents are required for halfExpandedRatio." }
-        return heightFractionAt(1).toFloat()
+        val middleDetentHeight =
+            heightAt(1, containerHeight, keyboardLift)
+                .coerceAtMost(maxAllowedHeight(containerHeight, keyboardLift) - MIDDLE_DETENT_MIN_GAP)
```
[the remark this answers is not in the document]
t0maboro: https://github.com/software-mansion/react-native-screens/pull/4573/changes/989ea714683aa3a1b5fb537ab2b5c51fb29df1d4

kligarski (approved):
To be honest, the calculations with `keyboardLift` are starting to become hard to comprehend but I guess it's just Material.

The runtime looks good.

PR #4574, opened 2026-08-31
fix(Android): stop reading uninitialized state from requestLayout() so R8 optimization cannot crash it

### LINKED ISSUE #4586: Android: crash on first screen mount with R8 optimization enabled (null check in ScreenContainer.requestLayout removed)
### Description

With R8 optimization enabled in a release build, the app crashes as soon as the first screen is mounted:

```
FATAL EXCEPTION: main
java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.Class java.lang.Object.getClass()' on a null object reference
    at com.facebook.react.modules.core.ReactChoreographer.postFrameCallback
    at com.swmansion.rnscreens.ScreenContainer.requestLayout
    at android.view.View.setFlags(View.java:18980)
    at android.view.ViewGroup.initViewGroup(ViewGroup.java:731)
    at android.view.ViewGroup.<init>(ViewGroup.java:724)
    at com.swmansion.rnscreens.ScreenContainer.<init>
    at com.swmansion.rnscreens.ScreenStack.<init>
    at com.swmansion.rnscreens.ScreenStackViewManager.createViewInstance

Expected: the app launches and renders the stack, as it does with R8 optimization off.
```
Cause: ViewGroup's constructor calls requestLayout() before the subclass is initialized. ScreenContainer overrides requestLayout() and reads layoutCallback, which is still null at that point. The code guards this with `layoutCallback != null` plus `@Suppress("SENSELESS_COMPARISON")`. R8's optimizer sees a final field written once with a non-null object, decides the check can never fail, and deletes it. Confirmed by reading the post-R8 dex of ScreenContainer.requestLayout. CustomToolbar and TabsHost have the same pattern.

This depends on the R8 version. R8 8.12.14 (bundled by AGP 8.11) still keeps the check. R8 9.4.14 removes it. So the library currently works only while consuming apps keep R8 optimization off, which AGP 9 makes harder: it rejects proguard-android.txt (the default file carrying -dontoptimize) by default, and Google's current guidance is proguard-android-optimize.txt.

### Steps to reproduce

1. Any app that renders a native stack (react-navigation native stack, which mounts a ScreenStack).
2. In android/app/build.gradle, release build type: minifyEnabled true and proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'.
3. In android/settings.gradle, inside pluginManagement, pin the newer R8:
   buildscript {
       repositories { google(); mavenCentral() }
       dependencies { classpath("com.android.tools:r8:9.4.14") }
   }
4. Build the release variant and launch it.
5. The app crashes on the first mounted screen with the trace above. Repeat with -dontoptimize (the legacy proguard-android.txt) and it launches.

https://github.com/TaduJR/rnscreens-r8-optimizer-crash

### Screens version

4.25.0

### React Native version

0.86.0

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

Expo bare workflow

### Build type

Release mode

### Device

Android emulator

### Device model

Google Android emulator, arm64, API 37

### Acknowledgements

Yes

TaduJR:
Fixes https://github.com/software-mansion/react-native-screens/issues/4586

## Description

With R8 optimization enabled, the app crashes as soon as the first screen is mounted:

```
FATAL EXCEPTION: main
java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.Class java.lang.Object.getClass()' on a null object reference
	at com.facebook.react.modules.core.ReactChoreographer.postFrameCallback(SourceFile:4)
	at com.swmansion.rnscreens.ScreenContainer.requestLayout(SourceFile:21)
	at android.view.View.setFlags(View.java:18980)
	at android.view.ViewGroup.initViewGroup(ViewGroup.java:731)
	at android.view.ViewGroup.<init>(ViewGroup.java:724)
	at com.swmansion.rnscreens.ScreenContainer.<init>(SourceFile:1)
	at com.swmansion.rnscreens.ScreenStack.<init>(SourceFile:1)
	at com.swmansion.rnscreens.ScreenStackViewManager.createViewInstance(SourceFile:6)
```

`ViewGroup`'s constructor calls `requestLayout()` (through `initViewGroup` and `setFlags`) before the subclass is initialized. `ScreenContainer` overrides `requestLayout()` and reads `layoutCallback`, which is still null at that point. The code guarded this with `layoutCallback != null` plus `@Suppress("SENSELESS_COMPARISON")`. R8's optimizer sees a `final` field written once with a non-null object, decides the check can never fail, and deletes it. Reproduced with R8 9.4.14. R8 8.12.14, which AGP 8.11 bundles, still keeps the check.

So the library only works with R8 optimization off. AGP 9 rejects `proguard-android.txt` by default, the file that carried `-dontoptimize`, and Google's guidance is `proguard-android-optimize.txt`, so apps following it hit this. `CustomToolbar` and `TabsHost` have the same pattern. `TabsHost` is not legacy code.

An earlier revision of this PR typed the fields as nullable. That only changes Kotlin metadata, not bytecode, and the post-R8 dex came out identical to the crashing build. This revision removes the fields instead.

No existing issue covers this.

## Changes

`requestLayout()` no longer reads anything initialized after `super()`. It is gated by Booleans, whose default value is correct while the superclass constructor runs, and posts a fresh callback each time. `StackHost` already works this way.

- `legacy/ScreenContainer.kt`: removed the `layoutCallback` field, its body is now `forceSubtreeMeasureAndLayoutPass()`.
- `legacy/CustomToolbar.kt`: same, as `forceMeasureAndLayoutPass()`.
- `tabs/host/TabsHost.kt`: took over the two pending flags from `TabsHostLayoutCoordinator`, which held no other state, and that class is deleted. Its two scheduling methods and their docs moved into `TabsHost`.

One behaviour difference: a `requestLayout()` during construction used to be skipped and now posts one deferred measure/layout pass, which runs on the fully constructed view. `StackHost` already does this. Kotlin emits no write for `= false` initializers, so a flag set during construction is not reset afterwards and nothing posts twice.

`StackHost` and `StackHeaderSubview` also override `requestLayout()` and are unaffected: one guards on a `Boolean`, the other uses a safe call.

## Test plan

Reproduced on react-native-screens 4.25.0: release build, `minifyEnabled true`, `proguard-android-optimize.txt`, R8 9.4.14 via `classpath("com.android.tools:r8:9.4.14")` in `settings.gradle`, rendering a `ScreenStack`. The code is unchanged in 4.27.0 and on `main`.

Verified by reading the post-R8 dex of `ScreenContainer.requestLayout` with `dexdump -d`, R8 9.4.14, no keep rules for this library:

- unchanged code: null check gone from the dex, app crashes on launch.
- nullable fields (previous revision): dex identical to the above.
- this revision: a Boolean test followed by `postFrameCallback` with a new callback, no field read left. The app launches and renders with no fatal exception.

Compiles clean against React Native 0.87.0-rc.3 (`:react-native-screens:compileDebugKotlin` in `FabricExample`), zero errors, no new warnings on the changed files. No automated test is added because the failure only exists after R8 optimization runs.

## Checklist

- [x] Ensured that CI passes


android/src/main/java/com/swmansion/rnscreens/legacy/CustomToolbar.kt
```diff
@@ -89,21 +77,30 @@ open class CustomToolbar(
...
                 ReactChoreographer
                     .getInstance()
-                    .postFrameCallback(
-                        ReactChoreographer.CallbackType.NATIVE_ANIMATED_MODULE,
-                        layoutCallback,
-                    )
+                    .postFrameCallback(ReactChoreographer.CallbackType.NATIVE_ANIMATED_MODULE) {
+                        isLayoutEnqueued = false
+                        forceMeasureAndLayoutPass()
+                    }
```
kkafar, line 89: This won't fly that easily unfortunately.

It changes the timing when the callback is posted, e.g. it fires an additional callback, when previously `layoutCallback == null`. 

This is a very fragile code that would require careful testing & there is lack of capacity on our side at the moment.

TaduJR:
Check here https://github.com/software-mansion/react-native-screens/issues/4586#issuecomment-5574530893

PR #4578, opened 2026-09-01
fix(Android, Stack v5): scope header back button to its own stack

kkafar:
## Description

With a stack nested inside another stack, two headers are visible, each with its own back button. Pressing the **outer** header's back button is expected to pop the outer screen together with its whole nested stack (which is how iOS behaves). On Android it popped only the top screen of the **inner** stack instead.

Root cause: the header back button called `activity.onBackPressedDispatcher.onBackPressed()`, i.e. global back dispatch. The activity-level `OnBackPressedDispatcher` runs the newest enabled callback in its deque, which for a nested stack is the inner `FragmentManager`'s internal back callback — so a button with scoped ("pop *my* screen") semantics was routed through an unscoped mechanism.

The header back button press is now routed directly to the `StackContainer` that owns the pressed header, which resolves `preventNativeDismiss` itself and pops its own stack. System back / gesture-back is intentionally unchanged and still goes through the activity dispatcher (reworking that path is planned as a follow-up).

Closes software-mansion/react-native-screens-labs#1649.

## Changes

- Added internal `StackHeaderBackPressHandler` (SAM). `StackContainer` implements it and passes itself when creating `StackScreenFragment`; the fragment (holding it weakly) hands it to `StackHeaderCoordinatorLayout`, whose navigation-icon click now invokes the handler instead of the activity's `OnBackPressedDispatcher`.
- `StackContainer.handleHeaderBackButtonPress` guards against stale presses (pressed screen must be the current top, stack depth > 1, container attached), then pops via `popBackStackImmediate(screenKey, POP_BACK_STACK_INCLUSIVE)` on its own `FragmentManager`. The no-arg form must not be used here — it delegates the pop to the child primary-navigation fragment's back stack, reintroducing the bug.
- Added a recursive veto query to the `Container` / `ContainerItem` interfaces: `wantsToPreventStackNativeDismiss(): ContainerItem?`. An item consults its nested container first (non-null answer wins), otherwise answers for itself via its `preventNativeDismiss` flag. A nested `StackContainer` polls its items back-to-front (every screen dismissed with the subtree gets a vote, deepest wins); `TabsContainer` consults only its active item. On a veto, the vetoing screen fires `onNativeDismissPrevented`; otherwise the pop proceeds.
- `ContainerItem` is now public (was internal) so the public `Container` interface can reference it in the new query's signature.
- Updated `test-stack-lifecycle-events` and `test-stack-prevent-native-dismiss-nested-stack` scenarios and a stale e2e spec comment to describe the new behavior.

Setup: outer stack `Home → NestedStack`, inner stack `NestedHome → NestedA`, both headers visible, outer back button pressed on `NestedA`.

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/8ee31389-894b-4d1d-8624-8788518425f3" alt="before" /> | <video src="https://github.com/user-attachments/assets/8c14d4fe-0828-4ef6-86b8-7e2c9cac0423" alt="before" /> |

## Test plan

Verified manually on a Pixel 9 Pro emulator (API 36), FabricExample debug build, with the test screens launched directly via `apps/App.tsx`:

- `apps/src/tests/single-feature-tests/stack-v5/test-stack-lifecycle-events` — the fixed case (step 14: outer back button on `NestedA` → `Home`, Android toast order captured empirically and written into the scenario) plus regressions: single-stack header pop, outer back button at the nested root, inner back button popping within the nested stack, system back popping innermost-first and exiting the container at the nested root. No JS-side errors from the multi-screen dismiss (`onDismiss` storm handled cleanly).
- `apps/src/tests/single-feature-tests/stack-v5/test-stack-prevent-native-dismiss-nested-stack` — steps 4, 5, 6, 7, 10, 16, 18: header-button interception on enabled screens (one toast per press, incl. rapid repeated presses), system-back interception unchanged, disabled-flag pops unchanged.
- `test-stack-prevent-native-dismiss-single-stack` — header path is identical by construction with no nesting (same code path as the B-screen checks above).
- Existing Detox specs for both prevent-dismiss scenarios need no assertion changes.

Repro of the original issue: push `NestedStack`, push `NestedA`, tap the **outer** header back button — before this change only `NestedA` pops.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/android/src/main/java/com/swmansion/rnscreens/common/container/Container.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/common/container/Container.kt
@@ -9,4 +9,12 @@ interface Container {
      * (if any).
      */
     fun resolveCurrentContentScrollView(): ViewGroup?
+
+    /**
+     * Asked when this container's whole subtree is about to be natively dismissed,
+     * e.g. the screen hosting this container is popped. Returns the item that
+     * vetoes the dismissal via `preventNativeDismiss`, or null when nothing
+     * in the subtree does.
+     */
+    fun wantsToPreventStackNativeDismiss(): ContainerItem?
 }
--- a/android/src/main/java/com/swmansion/rnscreens/stack/screen/StackScreenFragment.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/stack/screen/StackScreenFragment.kt
@@ -2,12 +2,14 @@ package com.swmansion.rnscreens.stack.screen
 
 import android.content.res.Configuration
 import android.os.Bundle
+import android.util.Log
 import android.view.Gravity
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import androidx.fragment.app.Fragment
 import androidx.transition.Slide
+import com.swmansion.rnscreens.stack.header.StackHeaderBackPressHandler
 import com.swmansion.rnscreens.stack.header.StackHeaderCoordinatorLayout
 import java.lang.ref.WeakReference
 
@@ -15,7 +17,11 @@ internal class StackScreenFragment(
     internal val stackScreen: StackScreen,
     private val canNavigateBack: Boolean,
     private val delegate: WeakReference<StackScreenFragmentDelegate>,
+    backPressHandler: StackHeaderBackPressHandler,
 ) : Fragment() {
+    // Weakly held so that a fragment retained by FragmentManager past container teardown
+    // does not keep the container view subtree alive.
+    private val backPressHandler: WeakReference<StackHeaderBackPressHandler> = WeakReference(backPressHandler)
     private var screenLifecycleEventEmitter: StackScreenAppearanceEventsEmitter? = null
 
     /**
@@ -48,7 +54,11 @@ internal class StackScreenFragment(
         inflater: LayoutInflater,
         container: ViewGroup?,
         savedInstanceState: Bundle?,
-    ): View = StackHeaderCoordinatorLayout(requireContext(), stackScreen, canNavigateBack)
+    ): View =
+        StackHeaderCoordinatorLayout(requireContext(), stackScreen, canNavigateBack) { pressedScreen ->
+            backPressHandler.get()?.handleHeaderBackButtonPress(pressedScreen)
+                ?: Log.w(TAG, "[RNScreens] Header back button press dropped - handler is gone")
+        }
 
     override fun onViewCreated(
         view: View,
@@ -116,5 +126,9 @@ internal class StackScreenFragment(
     private fun teardownPreventNativeDismissCallback() {
         requireNativeDismissBackPressedCallback.remove()
         preventNativeDismissBackPressedCallback = null
+    }
+
+    companion object {
+        private const val TAG = "StackScreenFragment"
     }
 }
--- a/android/src/main/java/com/swmansion/rnscreens/stack/screen/StackScreenFragment.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/stack/screen/StackScreenFragment.kt
@@ -19,8 +19,6 @@ internal class StackScreenFragment(
     private val delegate: WeakReference<StackScreenFragmentDelegate>,
     backPressHandler: StackHeaderBackPressHandler,
 ) : Fragment() {
-    // Weakly held so that a fragment retained by FragmentManager past container teardown
-    // does not keep the container view subtree alive.
     private val backPressHandler: WeakReference<StackHeaderBackPressHandler> = WeakReference(backPressHandler)
     private var screenLifecycleEventEmitter: StackScreenAppearanceEventsEmitter? = null
 
--- a/android/src/main/java/com/swmansion/rnscreens/stack/screen/StackScreenFragment.kt
+++ b/android/src/main/java/com/swmansion/rnscreens/stack/screen/StackScreenFragment.kt
@@ -17,9 +17,8 @@ internal class StackScreenFragment(
     internal val stackScreen: StackScreen,
     private val canNavigateBack: Boolean,
     private val delegate: WeakReference<StackScreenFragmentDelegate>,
-    backPressHandler: StackHeaderBackPressHandler,
+    private val backPressHandler: WeakReference<StackHeaderBackPressHandler>,
 ) : Fragment() {
-    private val backPressHandler: WeakReference<StackHeaderBackPressHandler> = WeakReference(backPressHandler)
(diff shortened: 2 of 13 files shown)


android/src/main/java/com/swmansion/rnscreens/tabs/container/TabsContainer.kt
```diff
@@ -779,6 +780,10 @@ class TabsContainer internal constructor(
         return selectedTab.tabsScreen.findContentScrollView()
     }
 
+    // Only the active item is consulted - a preventing screen inside an inactive tab
+    // does not veto the dismissal.
+    override fun wantsToPreventStackNativeDismiss(): ContainerItem? = selectedTab.tabsScreen.wantsToPreventStackNativeDismiss()
```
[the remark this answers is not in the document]
kkafar: Updated locally. Will push in a moment.

apps/src/tests/single-feature-tests/stack-v5/test-stack-prevent-native-dismiss-nested-stack/scenario.md
```diff
@@ -276,3 +297,48 @@ Manual only (not automated):
...
+
+- [ ] The gesture pops normally back to **NestedB**, whose `Key` is unchanged
+      from step 30. No toast is shown — the preventing screens below
+      (**NestedB**, **NestedHome**) do not over-block after the restart.
+
+32. On **NestedB** (top screen again, prevent Enabled), perform a system
+    gesture-back.
+
+- [ ] The gesture is intercepted: the "Native dismiss prevented - NestedB"
+      toast appears and the app stays on **NestedB**.
```
kkafar, line 344: This does not work currently. Will be handled by the second PR in the stack.

android/src/main/java/com/swmansion/rnscreens/stack/screen/StackScreenFragment.kt
```diff
@@ -2,20 +2,24 @@ package com.swmansion.rnscreens.stack.screen
...
 import com.swmansion.rnscreens.stack.header.StackHeaderCoordinatorLayout
 import java.lang.ref.WeakReference
 
 internal class StackScreenFragment(
     internal val stackScreen: StackScreen,
     private val canNavigateBack: Boolean,
     private val delegate: WeakReference<StackScreenFragmentDelegate>,
+    backPressHandler: StackHeaderBackPressHandler,
 ) : Fragment() {
+    private val backPressHandler: WeakReference<StackHeaderBackPressHandler> = WeakReference(backPressHandler)
```
t0maboro: why we're not passing WeakRef. similarly as for `delegate`?

apps/src/tests/single-feature-tests/stack-v5/test-stack-prevent-native-dismiss-single-stack/scenario.md
```diff
@@ -168,3 +168,15 @@ is Disabled - including the pop expected after toggling **B** to Disabled.
 - [ ] The most recent toggle takes effect before the back press: since
       prevent is Enabled at press time, back is intercepted (toast shown, app
       stays on **B**).
+
+### Android: prevent native dismiss mechanism keeps working after activity backgrounding
+
+14. From **Home**, tap **Push B**. On **B** tap back button. 
+
+    - [ ] A green toast appears and the app remains on **B**.
```
t0maboro: nit: other checkboxes doesn't have indentations

t0maboro (approved):

t0maboro (approved):

android/src/main/java/com/swmansion/rnscreens/stack/screen/StackScreenFragment.kt
kkafar: I think we don't need a weak ref in either case. This is just a convention to indicate lack of ownership, but overall this won't leak anyway. 

As per unifying the way we receive the reference - I'll fix it up. Thanks!

kkafar: https://github.com/software-mansion/react-native-screens/pull/4578/commits/d25add2c50b3565cea0f1086e8f9ad2cdf97900f

android/src/main/java/com/swmansion/rnscreens/common/container/Container.kt
```diff
@@ -9,4 +9,12 @@ interface Container {
...
      */
     fun resolveCurrentContentScrollView(): ViewGroup?
+
+    /**
+     * Asked when this container's whole subtree is about to be natively dismissed,
+     * e.g. the screen hosting this container is popped. Returns the item that
+     * vetoes the dismissal via `preventNativeDismiss`, or null when nothing
+     * in the subtree does.
+     */
+    fun wantsToPreventStackNativeDismiss(): ContainerItem?
```
kligarski, line 19: I'm not sure about the name - `container.wantsToPreventStackNativeDismiss()` sounds to me like an action rather than a question, especially when it's a method, not a prop (which makes sense because we're traversing the tree here). 

Also, it returns an item or null but I'd expect a boolean from a method starting with "wants". Additionally, do we need the "Stack" in the name of the method? Won't there be any other container that might natively dismiss some items? E.g. closing a column in Split, if we implement it on Android?

I thought about `findItemPreventingNativeDismiss()` but I don't like that it sounds like finding an item is the expected result when really it's not the typical case.

kligarski (approved):
Looks good. I'm not sure about the name of the container interface function but I don't have a strong better suggestion.

android/src/main/java/com/swmansion/rnscreens/common/container/Container.kt
kkafar: I kind of agree with your points here except the part:

> [...] Won't there be any other container that might natively dismiss some items? E.g. closing a column in Split, if we implement it on Android?

If we do - we'll just refactor the method. 

I'll try to think of something more fitting, but will also proceed with this PR. If I have something I'll refactor it higher in the PR stack.

apps/src/tests/single-feature-tests/stack-v5/test-stack-prevent-native-dismiss-single-stack/scenario.md
kkafar: +1 here. 

I like it better when there are indentations, but +1, it introduces inconsistency.

kkafar: https://github.com/software-mansion/react-native-screens/pull/4578/commits/494e68da6ab0ff96f956573c609d168eb99252df

Issue #1775, opened 2023-05-29
Weird rotation when navigation from landscape screen to portrait one


vargajacint:
### Description

Basically, whenever we go from landscape screen to portrait one, the portrait one exhibits an ugly rotation from landscape view to correct portrait position.

https://user-images.githubusercontent.com/11645756/199097249-69ba8800-013f-44bc-b28b-37833c5b8a87.mp4

[This issue ](https://github.com/react-navigation/react-navigation/issues/10966)perfectly describes the problem!

### Steps to reproduce

1. Create an empty project using React Navigation library
2. Create a BottomStackNavigator and create a Native Screen on the same level
3. Change the Native Screen `orientation` property to `landscape_right`
4. Navigate from the BottomStackNavigator to the Native Screen and navigate back using `goBack` function.

### Snack or a link to a repository

https://snack.expo.dev/@vargajacint/weird-rotation-effect-issue

### Screens version

3.18.0

### React Native version

0.71.0

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

feedthedevil:
Hi. Any updates?

tboba:
Hi @feedthedevil, we are aware of the problem - @kkafar made a quick fix for it in #1848, but we still wonder if this PR should be merged. We're not happy how the animation looks like and also the transitions between screens are quite buggy on Fabric. The work is still in progress ⚠️

Issue #1774, opened 2023-05-23
Facing crash issue when putting app in background. 


pyshagrwl:
### Description

**Environment:** Please check below version of lib we are using
- React Native (Android) Version : 0.68.7
- "@react-navigation/native": "6.0.16",
- "@react-navigation/stack": "6.3.7",
- "react-native-screens": "3.20.0",
-  Hermes is enabled for our project 

When we are putting our application in background and when coming back to application we are facing crash issue. Please check below crash logs for the same. 

```
Fatal Exception: java.lang.RuntimeException: Unable to start activity ComponentInfo{com.my.app.uat/com.my.app.home.LandingActivity}: androidx.fragment.app.Fragment$l: Unable to instantiate fragment com.swmansion.rnscreens.ScreenFragment: calling Fragment constructor caused an exception
       at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3780)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3942)
       at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:109)
       at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
       at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2345)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loopOnce(Looper.java:233)
       at android.os.Looper.loop(Looper.java:344)
       at android.app.ActivityThread.main(ActivityThread.java:8212)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:584)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1034)

Caused by androidx.fragment.app.Fragment$l: Unable to instantiate fragment com.swmansion.rnscreens.ScreenFragment: calling Fragment constructor caused an exception
       at androidx.fragment.app.Fragment.instantiate(Fragment.java:681)
       at androidx.fragment.app.FragmentContainer.instantiate(FragmentContainer.java:57)
       at androidx.fragment.app.FragmentManager$3.instantiate(FragmentManager.java:507)
       at androidx.fragment.app.FragmentState.instantiate(FragmentState.java:81)
[...]

PR #4594, opened 2026-09-03
fix(Android, Stack v5): preventNativeDismiss survives activity restart

kkafar:
## Description

Stacked on #4578 (header chevron scoped to its own stack).

Two defects in `preventNativeDismiss` for the Android **system back press** (Stack v5), one pre-existing hole, and a tabs fix the second one depends on.

**1. Interception died after one activity stop/start** (app switch, lock screen, home button). `OnBackPressedDispatcher` hands a back press to the *last-added enabled* callback. Every `FragmentManager` registers its pop callback with the lifecycle-owner overload, which removes it on the owner's `ON_STOP` and re-appends it at the deque tail on `ON_START`. Our per-fragment `PreventNativeDismissCallback` used the plain overload, so its slot never moved; one background cycle let every `FragmentManager` callback leapfrog it.

Fix: each `StackContainer` owns one veto-only `OnBackPressedCallback`, registered with the **same lifecycle owner** its `FragmentManager` uses (the parent fragment when nested, `FragmentActivity`'s fragment host at the root). `LifecycleRegistry` re-dispatches `ON_START` in registration order, so after every restart the deque is rebuilt as `[fm, veto]` per level, deeper levels later. `FragmentManager` keeps executing pops, so predictive back is unaffected.

**2. A veto fired from a covered branch.** Lower stack screens stay resumed with their views attached, so a stack nested in a covered screen kept its veto live: with a preventing screen below, system back on the screen pushed above it was vetoed instead of popping it. The old per-fragment callback had the same hole. Fix: the veto is enabled only while the container is on the primary-navigation path, the rule `FragmentManager` applies to its own callback (`isPrimaryNavigation(parent)`), and is re-evaluated from `Fragment.onPrimaryNavigationFragmentChanged` (forwarded by `StackScreenFragment` and `TabsScreenFragment`; the latter is needed because `FragmentManager`'s recursion skips the tab's own stack container).

**3. Tabs never set a primary navigation fragment.** Required by 2, and a fix on its own: `FragmentManager` enables a child manager's back callbacks only on the primary-navigation path, so system back could never pop inside a tab's stack. `TabsContainer` now marks the selected tab primary in both of its transactions.

Root owner note: obtained through `FragmentManager.getHost()`, `@RestrictTo(LIBRARY)`, lint-suppressed; no public accessor exists. The activity's own lifecycle starts after every fragment and would put the root veto behind all nested `FragmentManager` callbacks. A host that is not a `LifecycleOwner` degrades to the activity with a warning.

### Behavior changes

- A stack nested in a tab pops on system back (was: press fell through to the enclosing stack or finished the activity).
- A v5 stack whose owner fragment is never made primary (v5 inside a legacy v4 `Screen`, brownfield RN-in-a-Fragment) no longer vetoes system back. `FragmentManager`'s own pop was already off there, so the press stops being swallowed.
- Unchanged, now documented: a lone root screen with the flag still blocks the activity from finishing and reports `onNativeDismissPrevented`.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1775
Closes https://github.com/software-mansion/react-native-screens-labs/issues/1774

## Changes

- `StackContainer`: `System back veto` region. Lifecycle-owned veto callback (registered on attach, removed on detach), enabled iff `isOnPrimaryNavigationPath() && top item vetoes`, recomputed eagerly on model changes, on attach/detach, on prop changes and on primary-navigation changes; re-dispatch to the next callback when the state went stale; ancestor invalidation walk. Keeps the resolved `FragmentManagerWithOwner` as its single source for the manager and the parent fragment.
- `Container.onOwnerPrimaryNavigationFragmentChanged()`: new hook; `StackScreenFragment` and `TabsScreenFragment` forward `onPrimaryNavigationFragmentChanged` to the container nested in their screen. `TabsContainer` implements it as a no-op.
- `TabsContainer`: `setPrimaryNavigationFragment(selected)` in both tab-switch transactions; null-safe `selectedTabOrNull` for the veto query.
- `FragmentManagerHelper`: `findFragmentManagerWithOwnerForView` returns manager, lifecycle owner, parent fragment and dispatcher; a root host that is not a `LifecycleOwner` falls back to the activity with `Log.w`; `findFragmentManagerForView` is non-null (it never returned null).
- `StackScreen`: `preventNativeDismiss` changes invalidate the owning container through `StackHost`, which also covers the window before the fragment view exists.
- Removed: `PreventNativeDismissCallback`, `PreventNativeDismissChangeObserver`, top-fragment bookkeeping in `StackScreenFragment`, `updateTopFragment` + `OnCommitCallbackOp`.

## Test plan

Scenarios (Android, direct `apps/App.tsx` launch, see each scenario's prerequisites):

- `apps/src/tests/single-feature-tests/stack-v5/test-stack-prevent-native-dismiss-nested-stack/scenario.md`: steps 1-26 regression, 27-32 interception after background/foreground (#1775), 33-39 screens pushed above a preventing nested stack, including a flag flipped while covered and a cancelled predictive gesture.
- `.../test-stack-prevent-native-dismiss-single-stack/scenario.md`: steps 1-15 regression, 16-17 lone root screen with the flag.
- `apps/src/tests/component-integration-tests/tabs-stack-v5/test-stack-prevent-native-dismiss-tabs-in-stack/scenario.md`: pop inside a tab's stack (#1774), unselected tab does not veto, tab switch, outer push above the tabs host, restart.

Detox: `FabricExample/e2e/single-feature-tests/stack-v5/test-stack-prevent-native-dismiss-{single,nested}-stack.e2e.ts` (chevron and Pop paths; system back is manual).

Status: static checks green (`yarn lint-android`, `assembleDebug`, CI). On-device pass of the three scenarios pending, hence draft. The restart defect was reproduced on device before the fix; the covered-branch and tabs-in-stack behaviors are source-derived until that pass runs.

## Checklist

- [x] Ensured that CI passes

https://claude.ai/code/session_01NsiBwyHRt1aTiDEMMUQk66


android/src/main/java/com/swmansion/rnscreens/stack/host/StackContainer.kt
```diff
@@ -392,7 +362,140 @@ internal class StackContainer(
...
+     * navigation branch. Lower stack screens stay resumed with their views attached, so a container
+     * nested in a covered screen would otherwise keep vetoing pops of the screens above it. Hence
+     * `isEnabled` is additionally gated on the owner fragment being on the primary-navigation path,
+     * the same `isPrimaryNavigation(parent)` rule FragmentManager applies to its own callback. The
+     * gate is re-evaluated on attach and whenever the owner reports a primary-navigation change
+     * (see [onOwnerPrimaryNavigationFragmentChanged]).
+     *
+     * Known limitation: a container re-attached while its owner is already started and its fragments
+     * survive in the FragmentManager lands after its own nested FragmentManagers until the next
+     * stop/start cycle (rare: Fabric remove+insert of an existing view, clipped subviews).
```
kkafar, line 399: I estimate that this is an extremely rare case for a container to be detached and re-attached w/o actually popping the whole container. 

I guess we'll see how it goes.

PR #4599, opened 2026-09-04
refactor(Android, Stack v5): move invalidation flags to native implementation

kligarski:
## Description

A stack v5 nested in tabs could permanently lose its header after a tab round trip: no app bar, and the content losing its collapsing scroll behavior, until some prop change happened to force a structural rebuild. The root cause was an ownership mistake in the header update flow — the invalidation flags ("what has not been applied yet") lived on the React-owned configuration, which outlives the native view that consumes them.

This PR moves the flags to the native consumer, unifies every update source (React props, color scheme, adoption, window attach) behind a single accumulate-then-flush path owned by `StackContainer`, and then — now that the flow can apply pure deltas — retains the header view itself across fragment view destruction, so a tab switch no longer rebuilds the header at all.

Along the way it fixes a handful of adjacent bugs that all came from the same place: the lift-on-scroll target lost on window reattach, the collapsed header re-expanding after any runtime configuration change, and a double rebuild on first mount when a color scheme override is set.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1192.
Closes https://github.com/software-mansion/react-native-screens-labs/issues/1756.

#### Why the invalidation flags moved to the native implementation

`invalidationFlags` used to live on `StackHeaderConfig` — the React-owned configuration object — and the coordinator cleared the bits it had applied. But that flag set is not a property of the configuration; it is consumer state: _"what the current consumer has not applied yet."_ Storing consumer state on the producer only works while the two share a lifetime, and here they do not:

- `StackHeaderConfig` is React-owned and lives as long as the shadow node.
- `StackHeaderCoordinatorLayout` is the fragment's view. Tabs drive the child fragment manager with `detach()`/`attach()`, which caps children at `CREATED` and destroys every `StackScreenFragment` view on every tab switch.

So on reattach a brand new coordinator adopted a surviving configuration whose flags had already been consumed by the previous, now-dead coordinator — typically `NONE`. `needsRebuild` was `false` and `appBarLayout` was `null`, so every branch of `processUpdate` was skipped, and nothing would ever raise the flags again. Only a prop change that raised `STRUCTURE`/`SUBVIEWS` (`type`, `hidden`, `transparent`, `maxLines`, a subview add/remove) resurrected the header.

With `pendingFlags` owned by the coordinator, adoption is _fully dirty by construction_ — `invalidate(ALL)` — because a freshly adopted configuration is, by definition, something this coordinator has applied nothing of.

#### The path of updates

Everything now funnels through one path, and that path is the container's — the shape `TabsHost` / `TabsContainer` already use: the host owns the React commit boundary, the container owns the update pass. The configuration observer protocol is a single call, `onInvalidated(flags)`: accumulate on the consumer's side, cheap, callable at any time, from any source. The configuration no longer listens to the `UIManager` at all; it is a plain producer.

The batch boundary lives on `StackContainer`, announced by its owner:

- `beginUpdateBatch()` — `StackHost.willMountItems`. More updates are coming; hold header flushes.
- `endUpdateBatch()` — `StackHost.didMountItems`, after the sorted push/pop operations were handed over. Clears the flag and runs `performContainerUpdateIfNeeded()`.

`performContainerUpdateIfNeeded()` is the one container update: pending navigation operations first, then a header flush over every fragment still in `stackModel`. Operations go first so a screen popped in the same commit is not updated on its way out. A screen pushed in the same commit has no coordinator yet; adoption and window attach build its header as before.

The coordinator keeps its `pendingFlags` and a single gate, reading the batch state through `StackUpdateBatchStateProviding` (implemented by the container, passed down through the fragment like the back press handler):

```kotlin
internal fun flushPendingUpdates() {
    val provider = currentProvider ?: return
    if (pendingFlags.isEmpty) return
    if (updateBatchStateProvider.get()?.isUpdatePending == true) return   // batch in progress
    if (!isAttachedToWindow) return                                       // theme not resolved yet; accumulate
    processUpdate(provider)
}
```

Every source is the same two lines — invalidate, then try to flush — and the gate decides whether it applies now or in the container pass:

| Source                                                        | Path                                                                        |
| ------------------------------------------------------------- | --------------------------------------------------------------------------- |
| React prop change                                             | setter → `onInvalidated(flag)` → gate holds → `endUpdateBatch` container pass |
| Adoption (first mount and every reattach)                     | `invalidate(ALL)` + flush                                                   |
| Color scheme change (prop, system config, or RN `Appearance`) | `applyUiNightMode` → `invalidate(STRUCTURE)` + flush                        |
| Window attach                                                 | `invalidate(LIFT_ON_SCROLL)` + flush                                        |
| Async prop icon resolution, content scroll view change        | `onInvalidated(...)` — outside a batch the gate is open, so it applies now  |

The last row is worth calling out: those config-side sites previously each carried their own `if (!isInsideMountTransaction) flushUpdates()` guard, because the configuration was the one deciding when to apply. They are now plain invalidations.

Native-originated updates (adoption, color scheme, window attach, async icon loads) do not go through the container; they only consult its batch state. Two concrete improvements from the unification: a `colorScheme` prop change arriving in the same commit as header prop changes used to produce a synchronous rebuild followed by a second update, and now coalesces into a single `processUpdate`; and push/pop and header updates of one commit are applied in a defined order, instead of depending on `UIManagerListener` registration order across the host and each header config.

#### Why `isUpdatePending`

Prop setters fire one at a time inside a mount transaction, so applying each invalidation as it arrives would mean N header updates (and potentially N rebuilds) per commit. The consumer therefore needs to ask "is more coming?" — and that is `isUpdatePending`.

The question is answered by the native container, not by React code: `StackHost` translates `willMountItems`/`didMountItems` into `beginUpdateBatch`/`endUpdateBatch`, and nothing below the host knows about mount transactions. An owner that never begins a batch gets immediate application of every header update, which is the right default for a non-React owner. The property is deliberately a _hint_, not a command: the "can I apply right now?" decision stays in one place, alongside the coordinator's own `isAttachedToWindow` condition, and it is exactly what lets an async icon load that completes outside a commit apply without waiting for the next one.

#### Deferring the first flush until the view is attached

The header used to be built during coordinator construction (`init` → adoption), i.e. before `onAttachedToWindow` had a chance to pin the color scheme. With a `colorScheme` override that meant the first build ran under the un-pinned DayNight theme and was immediately thrown away and rebuilt.

The `!isAttachedToWindow` clause in the flush gate fixes that: an early flush only accumulates, `onAttachedToWindow` resolves the scheme first and then flushes once. Both first mount and reattach build the header exactly once, under the right theme. (A scheme change _inside_ `setup()` flushes by itself, and the trailing flush then no-ops.)

`onAttachedToWindow` also re-raises `LIFT_ON_SCROLL` on every attach. `AppBarLayout.onDetachedFromWindow` calls `clearLiftOnScrollTargetView()` and never re-resolves it.

#### Preserving the collapsed state across updates

The color scheme work landed a narrow version of this: a rebuilt app bar starts expanded, so `applyUiNightMode` captured `isAppBarFullyCollapsed` and re-asserted it afterwards. With the color path no longer rebuilding synchronously, that capture/restore had to move into `processUpdate` — and once there, it generalizes to every update for free.

What this changes:

- **Fixes an unhandled bug.** Changing `expandedTitleAppearance.fontSize` on a fully collapsed `medium`/`large` header changes the CTL height (`extraHeightForTitles`), and nothing re-asserted the offset afterwards — the header jumped. New SFT step 16 in `test-stack-header-title-appearance-android`.
- **Subsumes the subtitle workaround.** The `maxLines == 1 && isAppBarFullyCollapsed` `setExpanded` block in `applyTitleAndSubtitle` is gone, along with the `isAppBarFullyCollapsed` parameter threaded into the applicator. The `ctl.requestLayout()` next to it stays — `setExpanded` requests layout on the _AppBarLayout_, which does not mark the CTL child dirty, so the title/subtitle vertical split would never be recomputed.
- **Covers every rebuild trigger**, not just the theme one: `type`, `maxLines`, `transparent`, `collapsedTitleGravityMode`, subview add/remove.
- **Changes documented behavior for scroll flags.** `applyScrollFlags` deliberately snaps expanded, and the JS docs promised it. The restore now runs after it and wins, so a _fully_ collapsed header survives a scroll-flag change (and expands when the new flags cannot collapse); a _partially_ collapsed one still snaps expanded. `StackHeaderConfig.android.types.ts` is updated accordingly. This is a product decision, not a correctness one — the mechanics are safe either way.

Two supporting details: `isAppBarFullyCollapsed` is now cleared in `removeHeader()` (the header is genuinely gone) rather than `resetHeader()` (which also runs for a plain rebuild) — otherwise two rebuilds in the same frame, with no layout in between to re-populate the flag, would lose the collapsed state. And the `hidden` early-return in `processUpdate` now just calls `removeHeader()`, which is the same three calls it was making inline plus that clear.

Fractional offsets still reset to expanded on a rebuild. Material exposes no public API for an arbitrary app bar offset; the closest bypasses the behavior's scrim and drag-callback updates and would need a post-layout hook, producing a visible expanded → fractional jump. Which is part of why the next section exists.

#### Why the retain approach

Even with all of the above, a tab round trip still destroyed and rebuilt the entire header: a visible recreation flash, and a collapse state that could only ever be restored to _fully_ collapsed — a partial offset was always lost, since rebuild-and-restore has no way to express one.

But the fragment itself survives `detach()`/`attach()`; only its view is destroyed. So `StackScreenFragment` now caches its `StackHeaderCoordinatorLayout` and returns the same instance from `onCreateView`, with `tearDown()` moved from `onDestroyView` to `onDestroy`. The `FragmentManager` removes the view from its container before `onDestroyView`, so the cached instance can be returned as-is. The pattern is already precedented in this repo — `TabsScreenFragment.onCreateView` does the same.

Retention preserves the _exact_ offset, fractional included, because the offset lives in `ViewOffsetHelper.offsetTop`, owned by the behavior in the app bar's layout params. With no saved state and no pending action, `BaseBehavior.onLayoutChild` simply re-clamps and re-applies the current offset and re-dispatches `onOffsetChanged`, lifted state included. Nothing has to be reconstructed.

The update flow refactor is what makes retention nearly free — no new machinery:

- The `!isAttachedToWindow` gate means props changed while the tab is detached accumulate and apply as in-place deltas at reattach.
- `appliedUiNightMode` survives with the view and dedupes the scheme re-check in `onAttachedToWindow`.
- The `LIFT_ON_SCROLL` re-invalidation on attach covers exactly what Android clears on window detach.

Two things that deliberately stay: the adoption `invalidate(ALL)` (first mount is an adoption too, and rebuilds still happen inside a living coordinator on theme/`type`/`hidden` changes), and the toolbar menu controller lifecycle.

#### Known issue, handled in a follow-up: a re-shown header comes back expanded

Retention surfaced an asymmetry in `hidden`. Toggling it on and off _while the tab is detached_ keeps the collapse (only the latest value is ever read, so `removeHeader()` never runs), but hiding the header, looking at the screen, and then un-hiding it brings it back expanded — `hidden` invalidates `STRUCTURE`, so it takes the rebuild path through `removeHeader()`, which clears the collapse memory.

This will be handled in a follow-up PR. Ticket with more details: https://github.com/software-mansion/react-native-screens-labs/issues/1781.

## Changes

**Native — update flow**

- `StackHeaderConfigurationProviding`: dropped `invalidationFlags` / `clearInvalidationFlags`.
- `StackHeaderConfigurationObserver`: `onConfigChanged(config)` → `onInvalidated(flags)` (a `fun interface`).
- `StackHeaderConfig`: no longer a `UIManagerListener` and no longer stores or clears flags; prop setters invalidate through the observer, and the per-site `isInsideMountTransaction` flush guards are gone.
- `StackUpdateBatchStateProviding` (new, `stack/host/`): `isUpdatePending`, implemented by `StackContainer`, passed to `StackScreenFragment` and the coordinator as a `WeakReference`.
- `StackContainer`: `beginUpdateBatch()` / `endUpdateBatch()`; `performContainerUpdateIfNeeded()` now runs pending operations and then `flushPendingHeaderUpdates()` on every fragment in `stackModel`.
- `StackHost`: `willMountItems` → `beginUpdateBatch()`; `didMountItems` → `enqueuePendingOperations(...)` → `endUpdateBatch()`.
- `StackContainerUpdateCoordinator`: `executePendingOperationsIfNeeded` → `enqueuePendingOperations`; it only hands the sorted operations to the container and no longer calls `performContainerUpdateIfNeeded()` itself.
- `StackScreenFragment`: `flushPendingHeaderUpdates()` forwards to the retained coordinator.
- `StackHeaderCoordinatorLayout`: owns `pendingFlags`; new `invalidate()` / `flushPendingUpdates()` with a single gate (nothing pending / batch in progress per the container / not attached to window); `processUpdate` snapshots and clears the flags at entry and lost its `forcedFlags` parameter; adoption invalidates `ALL`; `onAttachedToWindow` resolves the color scheme, re-raises `LIFT_ON_SCROLL` and flushes once.
- `StackHeaderInvalidationFlags`: removed `APPEARANCE`, `clearing()` and `isNotEmpty`.

**Native — collapsed state**

- The `wasFullyCollapsed` capture/restore moved from `applyUiNightMode` into `processUpdate`, applying to every update.
- `isAppBarFullyCollapsed` is cleared in `removeHeader()` instead of `resetHeader()`; the `hidden` branch of `processUpdate` now calls `removeHeader()`.
- `StackHeaderApplicator.applyTitleAndSubtitle` lost its `isAppBarFullyCollapsed` parameter and the `setExpanded` workaround.

**Native — view retention**

- `StackScreenFragment` caches the `StackHeaderCoordinatorLayout` and returns it from `onCreateView`; `tearDown()` moved from `onDestroyView` to `onDestroy`.

**JS / docs / tests**

- `StackHeaderConfig.android.types.ts`: the `scrollFlag*` remark now documents that a fully collapsed header is preserved (partial collapse still snaps expanded, and a header that can no longer collapse expands).
- New CIT `tabs-stack-v5/test-stack-tabs-stack-in-tabs-header-persistence` (Android only) — header, offset, menu state and delta-vs-rebuild changes across tab round trips, with and without a `colorScheme` override.
- `test-stack-header-title-appearance-android/scenario.md`: new step 16 for the appearance-change-while-collapsed case.
- `test-stack-subviews-android/scenario.md`: corrected a known issue — a rebuild removes and re-adds only the app bar, so the content scroll position was never actually reset (what read as a reset was the header re-expanding and pushing content down), and a full collapse now survives the rebuild.

https://github.com/user-attachments/assets/884f14f9-5ee0-42e4-a8c6-481fda72e7b1

## Test plan

Run new `test-stack-tabs-stack-in-tabs-header-persistence` CIT. Note that step 19 (popping in the nested stack) is currently blocked, see: https://github.com/software-mansion/react-native-screens-labs/issues/1774.

You can also use SFTs:

- `test-stack-header-title-appearance-android` — new step 16, plus the existing appearance matrix.
- `test-stack-subviews-android` — subview add/remove while fully collapsed; scroll-flag switches (the behavior change from "always snaps expanded").
- `test-stack-color-scheme` — the scheme change while collapsed path, now going through the shared update path.
- `test-stack-toolbar-menu-*` — menu state across theme rebuilds and tab round trips.

## Checklist

- [x] Ensured that CI passes

---


t0maboro (approved):

android/src/main/java/com/swmansion/rnscreens/stack/header/config/StackHeaderConfig.kt
```diff
@@ -411,13 +384,16 @@ internal class StackHeaderConfig(
...
+        get() = isInsideMountTransaction
+
     override fun willMountItems(uiManager: UIManager) {
         isInsideMountTransaction = true
     }
 
     override fun didMountItems(uiManager: UIManager) {
         isInsideMountTransaction = false
-        flushUpdates()
+        configObserver?.onFlushRequested()
```
kkafar: This is interesting. I've read the PR description but it still does not sit well with me. 

Why do we want to synchronise the updates from the "configuration source" at all? 
Shouldn't the host be responsible for giving the container an impulse to flush the updates and then container would trigger the appearance / header coordinators? 

I'm not saying that we 100% change it, but I want to understand whether this flow is deliberate. 

My argument here is that having this go thorough container would mean that the appearance / structure update is well coordinated with other updates, such as tab change. 

In the diff you propose, I think, it relies on `UIManagerListener` callback order, doesn't it? Of course, valid question here is whether it matters at all when header structure / appearance is updated in relation to tab update / other updates.

kkafar (commented):
Looks good, thank you. I have one question regarding the update flow. Let's answer it before we proceed.

android/src/main/java/com/swmansion/rnscreens/stack/header/config/StackHeaderConfig.kt
kligarski: To be honest, I only though about batching the prop updates and consolidating React-triggered updates with native-triggered updates (like color scheme coordinator's update). I haven't thought about interaction with Tabs here.

We should be able to move the flush request to the host I think.

kligarski: https://github.com/software-mansion/react-native-screens/pull/4599/commits/cfe3625bb4fc3e8c98338dc4009482999c4ef86a

kkafar (approved):
Thank you! 

Looks good now.

PR #4606, opened 2026-09-07
fix(Android, Stack v5): restore valid header collapse state after re-showing the header

kligarski:
## Description

Re-showing a hidden stack v5 header always brought it back **expanded**, even when the content underneath was scrolled into its middle. That is a state no gesture can produce, and getting into it jumps the content by the full header height. The same defect hit a header that was `hidden` from the moment its screen mounted.

This PR makes the collapse memory a tri-state — "collapsed", "expanded", "unknown" — and lets the content scroll position decide when, and only when, the memory is unknown.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1781.

#### The problem

`hidden` invalidates `STRUCTURE`, so it always takes the rebuild path, and `processUpdate` early-returns through `removeHeader()` — which clears `isAppBarFullyCollapsed`. A rebuilt app bar starts expanded and there is nothing left to re-assert against, so the header comes back at full height regardless of where the content sits. A header hidden from mount never had a memory in the first place.

The base PR's view retention made this visible rather than causing it: before retention a tab round trip rebuilt everything anyway, so there was no state to lose.

#### What we considered

- **Today** — the memory dies with the app bar, so a re-shown header is always expanded.
- **A** — `wasFullyCollapsed || contentScrolled`, evaluated on _every_ update, guarded with `!scrollFlagEnterAlways`.
- **B** — `wasFullyCollapsed && contentScrolled`, with the memory kept alive through the hidden period.
- **C (this PR)** — tri-state memory. `isAppBarFullyCollapsed: Boolean?`, where `null` means "no offset observed since the header was last removed"; `removeHeader()` sets `null` instead of `false`, and the restore becomes `wasFullyCollapsed ?: isContentScrolled()`.
- **D** — retain the `StackHeaderAppBarLayout` instance across `hidden` instead of destroying it, so the exact offset comes back with the view, and decide its legality by comparing the content scroll position at hide time with the one at re-show.

D is the only option that reaches full fidelity — rows 2, 3 and 12 — and it does so with no scroll-flag reasoning. But it sits on top of C rather than replacing it (the tri-state rule is still needed wherever there is no retained instance), making it a net addition for outcomes that are already legal. Details in https://github.com/software-mansion/react-native-screens-labs/issues/1784.

#### Cases to consider

- ✅ **faithful** — comes back in the state it was hidden in.
- ➖ **different** — not the hide-time state, but a legal one: reachable by gesture, and one gesture from correction.
- ❌ **broken** — gesture-unreachable: an expanded app bar above content scrolled into its middle.

Default flags (`scroll` + `exitUntilCollapsed` + `snap`) unless the row says otherwise:

| #   | Case (hide → re-show unless noted)                                      | Today | A   | B   | C (this PR) |
| --- | ----------------------------------------------------------------------- | ----- | --- | --- | ----------- |
| 1   | Expanded, content at top                                                | ✅    | ✅  | ✅  | ✅          |
| 2   | Partially collapsed, content at top                                     | ➖    | ➖  | ➖  | ➖          |
| 3   | Fully collapsed, content at top                                         | ➖    | ➖  | ➖  | ➖          |
| 4   | Fully collapsed, content scrolled                                       | ❌    | ✅  | ✅  | ✅          |
| 5   | Scrolled back to the top while hidden                                   | ✅    | ✅  | ✅  | ✅          |
| 6   | Scrolled further down while hidden                                      | ❌    | ✅  | ✅  | ✅          |
| 7   | Hidden from mount, content scrolled                                     | ❌    | ✅  | ❌  | ✅          |
| 8   | Rebuild while hidden (`type`, color scheme), content scrolled           | ❌    | ✅  | ✅  | ✅          |
| 9   | Un-hide landing on tab reattach, content scrolled                       | ❌    | ✅  | ✅  | ✅          |
| 10  | Tab reattach, partial offset (no hide involved)                         | ✅    | ✅  | ➖  | ✅          |
| 11  | Tab reattach, fully collapsed, content at top (no hide)                 | ✅    | ✅  | ➖  | ✅          |
| 12  | `enterAlways`, whole bar re-entered above scrolled content              | ✅    | ✅  | ✅  | ➖          |
| 13  | `enterAlways + enterAlwaysCollapsed`, fully collapsed, content scrolled | ❌    | ❌  | ✅  | ✅          |

- **Rows 2/3** are the deliberate trade. Once the header is gone, "content at the top" cannot tell "never moved" from "scrolled back" (row 5), so both resolve to expanded. No approach fixes them without the hide-time snapshot.
- **Rows 10/11 are what kills B.** Its memory survives the hide, so its "content at top → expand" clause also fires on tab reattach — and by Material's pre-scroll ordering a partially collapsed header _always_ has its content at 0. Both land on legal states, but they turn ✅ into ➖ on every tab round trip. B needs an explicit "this is an un-hide, not a reattach" marker to avoid that.
- **Row 13 is what kills A.** The `!scrollFlagEnterAlways` guard — the price of running the content term on live-header paths — is too broad: with `enterAlwaysCollapsed` only the toolbar re-enters (`AppBarLayout.getDownNestedPreScrollRange`), so expanded-over-scrolled is unreachable there and the guard suppresses a restore that is needed. Dropping the guard fixes row 13, but then the term fires on live updates too and rows 10-12 regress under the `enterAlways` family.
- **Row 12 is C's only ➖** — the fidelity A's guard bought, traded away for having no flag reasoning at all.
- **Degenerate flags** (none set, or `small` + `exitUntilCollapsed`) are safe by construction under every approach: `totalScrollRange == 0`, so `setExpanded(false, false)` resolves to offset 0, i.e. expanded. `snap` never appears — it decides where a gesture comes to rest, not how state is restored.

#### Why this approach

**It is simple.** One field goes from `Boolean` to `Boolean?`, plus a one-line helper. The content term is scoped to exactly the transition that destroyed the memory, so live updates, plain rebuilds and tab reattach keep today's behavior unchanged — there is no scroll-flag branching anywhere.

**It always lands on a valid state.** "Fully collapsed over scrolled content" and "expanded with the content at the top" are gesture-reachable under _every_ scroll-flag combination, because scrolling down always collapses the bar before the content moves. So the rule cannot produce the shape it is fixing — with the present flags or any future ones. This is the whole reason the geometry analysis A needed disappears.

**The "incorrect" restores are all okay.** Every ➖ above is a state the user could have reached themselves and can undo with a single gesture. There is no ❌ left, and none of the alternatives removed a ➖ without adding a ❌ somewhere else.

Two smaller wins over a one-shot "the header was removed" boolean: after an un-hide the state stays `null` until the first real offset dispatch, so a second rebuild in the same frame still consults the content instead of losing the intent; and a repeated `removeHeader()` while already hidden (e.g. a device dark-mode toggle) is idempotent. Both config-swap orderings now settle in a legal state too — add-before-remove keeps the memory and re-collapses the adopted config's fresh app bar, remove-before-add falls through to the content rule.

#### Known issue

Re-showing a `medium`/`large` header with `scrollFlagExitUntilCollapsed: false` fades its status bar scrim in, which reads as a flash: https://github.com/software-mansion/react-native-screens-labs/issues/1782.

## Changes

**Native**

- `StackHeaderCoordinatorLayout`: `isAppBarFullyCollapsed` is now `Boolean?`; `removeHeader()` sets it to `null`; the restore is `wasFullyCollapsed ?: isContentScrolled()`.
- New `isContentScrolled()` helper on top of the existing `findContentScrollView()` plumbing (already used for lift-on-scroll) — no new resolution path.

**Tests / docs**

- New SFT `test-stack-header-hidden-restore-android`: the full rule — at-top drop, scroll-while-hidden in both directions, rebuild while hidden, per-flag variants (including the small header and the no-flags degradation), partial-offset reset, and hidden-from-mount via a pushed screen. The screen exposes individual scroll-flag switches and uses `ScrollViewMarker`, so the registered content-scroll-view resolution path is exercised.
- CIT `test-stack-tabs-stack-in-tabs-header-persistence`: **Note** and step 17 rewritten (a removed header no longer "always comes back expanded"), new step 18 for the detached un-hide over scrolled content, later steps renumbered.

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/be0854d6-c7e7-4d79-8e89-0bf23e036189" /> | <video src="https://github.com/user-attachments/assets/43215ca1-fe6f-4302-b9e4-71b2db3494f9" /> |

## Test plan

`test-stack-header-hidden-restore-android` (SFT) — all 18 steps; it is the rule in full.

`test-stack-tabs-stack-in-tabs-header-persistence` (CIT) — steps 16-18 for the detached path, plus steps 6-9 and 14-15 as a regression check that tab reattach is untouched. Note that step 20 is still blocked, see https://github.com/software-mansion/react-native-screens-labs/issues/1774.

## Checklist

- [ ] Ensured that CI passes

---


apps/src/tests/single-feature-tests/stack-v5/test-stack-header-hidden-restore-android/scenario.md
```diff
@@ -0,0 +1,195 @@
...
+- Scroll-flag steps only use valid combinations
+  (`scrollFlagEnterAlwaysCollapsed` requires `scrollFlagEnterAlways`; both
+  are meaningful only with `scrollFlagExitUntilCollapsed` off).
+- A `small` header cannot collapse while `scrollFlagExitUntilCollapsed` is on
+  (its pinned height equals its full height), which is why the small-header
+  step turns that flag off.
+- When `medium/large` header with `scrollFlagExitUntilCollapsed: false` is
+  re-shown, its status bar scrim flashes (it fades in). This is a known
+  issue, see
+  https://github.com/software-mansion/react-native-screens-labs/issues/1782.
```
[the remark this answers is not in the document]
kligarski: This is an internal issue tracking repository.

apps/src/tests/single-feature-tests/stack-v5/test-stack-header-hidden-restore-android/scenario.md
```diff
@@ -0,0 +1,195 @@
...
+      and re-enters only once the content is scrolled back to the top.
+
+9. **scrollFlagEnterAlways** on (with **scrollFlagExitUntilCollapsed** still
+   off).
+
+- [ ] The header comes back fully collapsed, and a small upward drag
+      re-enters the whole header while the content stays mid-list.
+
+10. With the flags from step 9, drag up so the whole header re-enters while
+    the content stays scrolled, then toggle **hidden** on and off.
```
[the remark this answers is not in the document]
kligarski: https://github.com/software-mansion/react-native-screens/pull/4606/commits/79846e21d643a1041ae946912a8d4df4263be3a5

android/src/main/java/com/swmansion/rnscreens/stack/header/StackHeaderCoordinatorLayout.kt
```diff
@@ -242,18 +243,23 @@ internal class StackHeaderCoordinatorLayout(
...
-            if (wasFullyCollapsed) {
+            if (wasFullyCollapsed ?: isContentScrolled()) {
                 appBar.setExpanded(false, false)
             }
         }
 
         onMaybeHeaderLayoutChanged()
     }
 
+    // If the content scroll view can scroll further up, we consider it scrolled.
```
kmichalikk: = didn't reach the top

this is a bit confusing because scrolling up is pulling finger down? Maybe let's reword this.

android/src/main/java/com/swmansion/rnscreens/stack/header/StackHeaderCoordinatorLayout.kt
```diff
@@ -328,7 +334,7 @@ internal class StackHeaderCoordinatorLayout(
 
     private fun removeHeader() {
         resetHeader()
-        isAppBarFullyCollapsed = false
+        isAppBarFullyCollapsed = null
```
kmichalikk, line 333: nit to consider: Would replacing this int with a proper enum that has some STATE_UNKNOWN be more readable at the cost of more verbose code? Wouldn't it be better?

apps/src/tests/single-feature-tests/stack-v5/test-stack-header-hidden-restore-android/scenario.md
```diff
@@ -0,0 +1,195 @@
...
+   change **type** to `medium`, then toggle **hidden** off.
+
+- [ ] The header comes back as a `medium` header, **fully collapsed** - a
+      rebuild request arriving while hidden does not disturb the rule.
+
+7. Set **type** back to `large` and scroll the content back to the top.
+
+- [ ] The header stays fully collapsed across the type change - the header is
+      live, so its own collapse state is re-asserted - and expands only once
+      the content is back at the top.
```
kmichalikk: this explanation is not necessary

apps/src/tests/single-feature-tests/stack-v5/test-stack-header-hidden-restore-android/scenario.md
```diff
@@ -0,0 +1,195 @@
...
+
+---
+
+### Scroll-flag variants
+
+Each step: set the flags as listed, scroll down until the content is visibly
+scrolled, toggle **hidden** on and off.
+
+8. **scrollFlagExitUntilCollapsed** off (plain scrolling header).
```
kmichalikk: nit: **Set** x off / **Set** x on

apps/src/tests/single-feature-tests/stack-v5/test-stack-header-hidden-restore-android/scenario.md
```diff
@@ -0,0 +1,195 @@
...
+      state is dropped (see **Note**).
+
+11. **scrollFlagEnterAlwaysCollapsed** on as well. After checking the first
+    result, scroll up so the toolbar re-enters, then toggle **hidden** on and
+    off again.
+
+- [ ] The header comes back fully collapsed; an upward scroll re-enters only
+      the toolbar, and the full height returns only once the content reaches
+      the top.
+- [ ] After the second hide/re-show the re-entered toolbar is dropped: the
```
kmichalikk, line 139:
```suggestion
      the top.
      
- [ ] After the second hide/re-show the re-entered toolbar is dropped: the
```

apps/src/tests/single-feature-tests/stack-v5/test-stack-header-hidden-restore-android/scenario.md
```diff
@@ -0,0 +1,195 @@
...
+16. Scroll to the top. Tap **Push Details**.
+
+- [ ] The _Details_ screen has no header from the moment it appears and its
+      content starts below the status bar.
+
+17. Scroll the _Details_ content down a little, then toggle its **hidden**
+    switch off.
+
+- [ ] The header appears **fully collapsed** - the very first build of this
+      header lands on scrolled content and must not expand above it.
```
kmichalikk, line 191: Is this "fully collapsed" in this case?

Image

kmichalikk (commented):
The assumptions seems reasonable & work as described.

I observed strange glitches with some flags when hiding / showing - is there anything we can do to make them go away?

https://github.com/user-attachments/assets/e741b4bb-95fe-4cbf-8992-1c993a46477a

android/src/main/java/com/swmansion/rnscreens/stack/header/StackHeaderCoordinatorLayout.kt
kligarski: I thought that this would explain `canScrollVertically(-1)` better but I changed it. Scrolling up = scrolling towards the top = on mobile, dragging down.

https://github.com/software-mansion/react-native-screens/pull/4606/commits/680212d9168561ba445c6cd5de320afda7184a6f

android/src/main/java/com/swmansion/rnscreens/stack/header/StackHeaderCoordinatorLayout.kt
kligarski: I don't really see any benefit and `null` meaning no data seems clear to me but I can change it if we want to.

apps/src/tests/single-feature-tests/stack-v5/test-stack-header-hidden-restore-android/scenario.md
kligarski: https://github.com/software-mansion/react-native-screens/pull/4606/commits/680212d9168561ba445c6cd5de320afda7184a6f

apps/src/tests/single-feature-tests/stack-v5/test-stack-header-hidden-restore-android/scenario.md
kligarski: https://github.com/software-mansion/react-native-screens/pull/4606/commits/680212d9168561ba445c6cd5de320afda7184a6f

apps/src/tests/single-feature-tests/stack-v5/test-stack-header-hidden-restore-android/scenario.md
kligarski: https://github.com/software-mansion/react-native-screens/pull/4606/commits/680212d9168561ba445c6cd5de320afda7184a6f

apps/src/tests/single-feature-tests/stack-v5/test-stack-header-hidden-restore-android/scenario.md
kligarski: If you try to scroll down (drag up), the header doesn't collapse any further - this means it's fully collapsed (and this is the case here).

kligarski:
> The assumptions seems reasonable & work as described.
> 
> I observed strange glitches with some flags when hiding / showing - is there anything we can do to make them go away?

1. I created an issue for incorrect background color with `exitUntilCollapsed`: https://github.com/software-mansion/react-native-screens-labs/issues/1790
2. Issue for status bar scrim fade-in animation appearing when re-showing the header in collapsed state already exists: https://github.com/software-mansion/react-native-screens-labs/issues/1782
3. The content changing size after a delay happens due to lack of synchronous state updates on Android. I also apply SafeAreaView when header is hidden here which might cause additional jumps - this might be an interesting case to consider for our approach to handling the safe area.

apps/src/tests/single-feature-tests/stack-v5/test-stack-header-hidden-restore-android/scenario.md
```diff
@@ -0,0 +1,196 @@
...
+- Scroll-flag steps only use valid combinations
+  (`scrollFlagEnterAlwaysCollapsed` requires `scrollFlagEnterAlways`; both
+  are meaningful only with `scrollFlagExitUntilCollapsed` off).
+- A `small` header cannot collapse while `scrollFlagExitUntilCollapsed` is on
+  (its pinned height equals its full height), which is why the small-header
+  step turns that flag off.
+- When `medium/large` header with `scrollFlagExitUntilCollapsed: false` is
+  re-shown, its status bar scrim flashes (it fades in). This is a known
+  issue, see
+  https://github.com/software-mansion/react-native-screens-labs/issues/1782.
```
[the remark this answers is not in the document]
kligarski: https://github.com/software-mansion/react-native-screens/pull/4606#discussion_r3947851144

apps/src/tests/single-feature-tests/stack-v5/test-stack-header-hidden-restore-android/scenario.md
```diff
@@ -0,0 +1,196 @@
+# Test Scenario: Stack header hidden restore
+
+## Details
+
+**Description:** Re-showing a hidden stack v5 header must restore the collapse
+state. The rule under test: the header comes back **fully collapsed** when the
+content is scrolled away from the top and **expanded** otherwise, independent
+of the configured scroll flags. This prevents the broken shape of an expanded
+app bar pushed above content that is scrolled into its middle, for every
+scroll-flag combination, including a header that was hidden from the moment
```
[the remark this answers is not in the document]
kligarski: No scroll => collapsed == expanded. I don't think that specifying this edge case is really helping. The general rule is described here and a step specifies the edge case.

kmichalikk (approved):

PR #4609, opened 2026-09-07
feat(JS): add stack state config, update tabs one 

Loloekk:
## Description

This PR adds a way to set the initial screens on a stack. Pass an array of route names; they are pushed from left to right, so the last one ends up on top. Names do not have to be unique — a repeated name is pushed again as a separate instance.

This PR also adds a type check for defaultRouteName for tabs.

To avoid using an incorrect screen name when defining the initial state of a stack or tabs, it is necessary to use `defineStackRouteConfigs` / `defineTabRouteConfigs`. Thanks to that, we get an IDE error if something is wrong. 

Closes https://github.com/software-mansion/react-native-screens-labs/issues/1786

## Changes

- Add the `initialRouteNames` prop for stack,
- Update tabs witch a type check

No visual changes

## Test plan

It will be checked through use in other tests.

## Checklist

- [ ] Ensured that CI passes


apps/src/tests/issue-tests/index.ts
```diff
@@ -212,6 +212,7 @@ export { default as Test4361 } from './Test4361';
 export { default as Test4423 } from './Test4423';
 export { default as Test4519 } from './Test4519';
 export { default as Test4571 } from './Test4571';
+export { default as Test9999 } from './Test4609';
```
kkafar: Seems like a typo

apps/src/tests/issue-tests/Test4609.tsx
kkafar, line 1: Does this issue really need a separate regression test? It can be checked on any other test, right?

kkafar (commented):
Looks good overall. I've noticed a typo & unnecessary test. Let's discuss about it first before proceeding.

apps/src/tests/issue-tests/Test4609.tsx
Loloekk: Yes, but currently we do not use this feature anywhere,. Should I edit any other test or add a new one as single feature test?

kkafar: The `issue-tests` are meant to test library features. We don't really test our "test code", which is the case of example container implementations. 

I recommend dropping this test entirely & we'll see it in action soon when I add tests related to the native back handling on Android.

Loloekk: I removed this test in [80b5595](https://github.com/software-mansion/react-native-screens/pull/4609/commits/80b55954c81240dc667f8e009fa534c050923c3e)

apps/src/tests/issue-tests/index.ts
Loloekk: I removed this test in [80b5595](https://github.com/software-mansion/react-native-screens/pull/4609/commits/80b55954c81240dc667f8e009fa534c050923c3e)

kkafar (approved):
Looks good, thank you!

PR #4610, opened 2026-09-07
feat(iOS, Tabs): support deferring transition until first state update

satya164:
## Description

When using `lazy` rendering in react navigation, screens are rendered only after the first visit. Since tab switch is controlled natively, and JS render can lag behind, this usually shows a blank screen before the content is rendered, which is undesirable.

## Changes

This adds a prop to defer the transition until state is updated for the first time for a screen. We override the native transition to achieve this, as the built-in transition is private, this is an approximation. But it's better than flash of a blank screen.

Currently this is iOS-only.

> [!NOTE]
> I don't have much familiarity with the iOS APIs, so the changes were authored with Codex.

### iOS 18

**Before**

https://github.com/user-attachments/assets/e4d26485-e499-480e-96cf-e168f465c752

**After**

https://github.com/user-attachments/assets/0272a007-9318-4bbf-bf28-4ffbcc884b43

### iOS 26

**Before**

https://github.com/user-attachments/assets/8129aa78-0f26-46f9-be7b-dfa61cbb9776

**After**

https://github.com/user-attachments/assets/c1bc54fe-a490-4c5d-bbff-d750e69b89ff

### iOS 27

**Before**

https://github.com/user-attachments/assets/5ebbc1d8-2c0f-41cb-a9d2-e9ad692a0a44

**After**

https://github.com/user-attachments/assets/9dd26a02-bd0b-4ab3-8be2-715043c6b214

## Test plan

Added an example for the feature in `FabricExample`.

## Known issues / untested

- Animation is approximate
- Scroll edge affects changes earlier without animation
- Overflow is untested

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/ios/tabs/host/RNSTabBarController.mm
+++ b/ios/tabs/host/RNSTabBarController.mm
@@ -1,4 +1,5 @@
 #import "RNSTabBarController.h"
+#import <QuartzCore/QuartzCore.h>
 #import <React/RCTAssert.h>
 #import <React/RCTLog.h>
 #import <objc/message.h>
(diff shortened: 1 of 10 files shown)


ios/tabs/host/RNSTabBarController.mm
```diff
@@ -873,8 +1109,9 @@ - (nullable UIViewController *)popToRootMoreNavigationController:
   }
 
   auto *poppedViewControllers = [moreNavigationController popToRootViewControllerAnimated:animated];
-  RCTAssert(poppedViewControllers != nil && poppedViewControllers.count == 1,
-            @"[RNScreens] Expected exactly one view controller to be popped");
+  RCTAssert(
+      poppedViewControllers != nil && poppedViewControllers.count == 1,
+      @"[RNScreens] Expected exactly one view controller to be popped");
   return [poppedViewControllers firstObject];
 }
```
satya164, line 1116: Unrelated formatting changes are due to the repo's pre-commit hook.

PR #4611, opened 2026-09-08
feat(Android, Stack v5): handle `direction` prop and improve RTL collapsing header workaround

kligarski:
## Description

Stack v5 on Android did not fully support RTL. The collapsing header's collapsed title was drawn over the trailing subview and the toolbar menu, and push/pop transitions always slid the same way regardless of direction.

This PR fixes both and exposes a `direction` prop on `StackHost` (`inherit` | `ltr` | `rtl`) so the stack's layout direction can be set explicitly, independently of `I18nManager` in accordance with [RFC-996](https://github.com/software-mansion/react-native-screens-labs/blob/main-issue-tracker/rfcs/0996-rtl-and-dark-mode.md). The prop is Android-only for now.

Closes https://github.com/software-mansion/react-native-screens-labs/issues/891.

### Details

**The `direction` prop.** On Android it maps to React Native's `style.direction`, which sets the native `layoutDirection` on the host view; the whole native subtree — app bar, subviews, menu, back button and the screen content — inherits it. `inherit` therefore falls back to whatever a parent view resolves to, which in practice is the `I18nManager` setting.

**The collapsing header workaround, reworked.** `CollapsingToolbarLayout` tracks the collapsed title area by injecting a `MATCH_PARENT` placeholder view into the toolbar, which inherits the toolbar's default `START` gravity — the same layout group our leading subview goes into. Under RTL the toolbar walks that group in reverse child order, so the placeholder is positioned first and consumes the entire layout cursor ([material-components-android#1867](https://github.com/material-components/material-components-android/issues/1867)).

The previous workaround force-measured the app bar to trigger the placeholder's creation and then moved it to index 0. It depended on an `isRTL` flag plumbed through the React-owned header config, and it did not fix the layout fully - the collapsing header's collapsed title was drawn over the trailing subview and the toolbar menu.

It is replaced by `StackHeaderToolbar`, which intercepts the placeholder in `onViewAdded` and gives it `CENTER_HORIZONTAL` gravity — the group the toolbar positions last, clamped into whatever the other children left over, which is the correct rect in both directions. As a side effect the `isRTL` flag is gone from the header config: direction is now read natively, where it belongs.

**Rebuild on direction change.** Not everything in the header re-resolves on its own when the direction flips at runtime — most visibly the back button icon, which has to swap its arrow direction. `StackHeaderCoordinatorLayout` now remembers the direction the header was built against and triggers a structural rebuild from `onRtlPropertiesChanged`, so layout and appearance are both correct after a switch.

## Changes

- Added the `direction` prop (`StackHostDirection`) to `StackHost` and exported the type.
- Added `StackHeaderToolbar` for the collapsing header and removed the old RTL workaround from `StackHeaderApplicator`, along with `isRTL` from the header configuration interface.
- Rebuilt the header on layout direction change in `StackHeaderCoordinatorLayout`.
- Switched fragment transitions from `Gravity.RIGHT` / `Gravity.LEFT` to `Gravity.END` / `Gravity.START` so push/pop mirror under RTL.
- Added the `test-stack-layout-direction` SFT and its scenario.

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/6bc71053-7be1-41bf-b76d-6ad0aa046473" /> | <video src="https://github.com/user-attachments/assets/0f239b03-8705-42e7-b6de-3393e8d9a1f9" /> |

## Test plan

New scenario: `apps/src/tests/single-feature-tests/stack-v5/test-stack-layout-direction`.

## Checklist

- [ ] Ensured that CI passes

---


apps/src/tests/single-feature-tests/stack-v5/test-stack-layout-direction/scenario-description.ts
```diff
@@ -0,0 +1,11 @@
+import type { ScenarioDescription } from '@apps/tests/shared/helpers';
+
+export const scenarioDescription: ScenarioDescription = {
+  name: 'Stack Layout Direction',
+  key: 'test-stack-layout-direction',
+  details:
+    'Tests how the stack header and navigation handle system, React Native, and StackHost layout direction.',
+  platforms: ['android'], // TODO: add iOS
```
kmichalikk, line 8: I don't think this is ever going to be merged with iOS since the config is different. Let's rename the test to *-android

apps/src/tests/single-feature-tests/stack-v5/test-stack-layout-direction/scenario.md
```diff
@@ -0,0 +1,232 @@
...
+
+**Description:** Validates the `direction` prop on `StackHost`, which controls
+the layout direction of the stack's native container and everything it holds:
+the Material 3 app bar (title, subtitle, leading/center/trailing subviews,
+back button, toolbar menu) and the screen content. Success means every
+direction-relative element resolves against the effective direction, that
+switching direction at runtime re-lays the header out immediately and
+correctly in both directions, and that push/pop transitions mirror.
+
+The prop is currently Android-only. On Android it works through React Native's
```
kmichalikk, line 13: No need to say that it's android only if the test is marked for android

apps/src/tests/single-feature-tests/stack-v5/test-stack-layout-direction/scenario.md
```diff
@@ -0,0 +1,232 @@
...
+  RTL the two swap. `C` stays centered in both. The `centerSubview` switch is
+  shown only for the `small` header, which is the only type that supports it.
+- The default back arrow comes from the theme (`homeAsUpIndicator`) and is
+  auto-mirrored, so it points left in LTR and right in RTL. A custom
+  `backButtonIcon` supplied as a bitmap is **not** mirrored - that is expected,
+  and out of scope here.
+- `contentInsetStart` / `contentInsetEnd` under RTL are covered by
+  `test-stack-header-content-insets-android`.
+- `medium` is deliberately not in the header-type picker: it shares the whole
+  collapsing code path with `large` and differs only in type scale.
```
kmichalikk, line 52: Let's make this shorter; this could very well be 3 short points

apps/src/tests/single-feature-tests/stack-v5/test-stack-layout-direction/scenario.md
```diff
@@ -0,0 +1,232 @@
...
+### Baseline
+
+1. Navigate to the **Stack Layout Direction** scenario. It opens with
+   `direction = inherit`, `type = small`, `menu = none`, leading and trailing
+   subviews on and the center subview off.
+
+- [ ] React Native RTL label is `I18nManager.isRTL == false`.
+- [ ] Title `Direction` and subtitle `Subtitle` are left-aligned.
+- [ ] `L·48` sits just after the title, `T·48` on the right edge.
+- [ ] The content row reads `START` on the left, `END` on the right.
```
kmichalikk, line 67: these should be indented so they are under 1. when rendered

android/src/main/java/com/swmansion/rnscreens/stack/header/appbar/StackHeaderToolbar.kt
```diff
@@ -0,0 +1,58 @@
...
+import android.view.View
+import android.view.ViewGroup.LayoutParams.MATCH_PARENT
+import com.google.android.material.appbar.MaterialToolbar
+import com.swmansion.rnscreens.stack.header.subview.StackHeaderSubview
+
+/**
+ * Toolbar of a collapsing header.
+ *
+ * CollapsingToolbarLayout tracks the collapsed title area by injecting a MATCH_PARENT placeholder
+ * view into the toolbar, and it inherits the toolbar's default START gravity - the same layout
```
kmichalikk, line 14: "match parent placeholder view" is a mental shortcut that may not be understood, please be more descriptive of what you do and what is the heuristic here, maybe the heuristic for asserting the view could be a separate function?

PR #4622, opened 2026-09-10
feat(iOS, Stack v5): Add button and prominentButton text appearance

kmichalikk:
## Description

This PR adds text attributes to `buttonAppearance` and `prominentButtonAppearance` for both `standardAppearance` and `scrollEdgeAppearance`. These are: `color`, `fontFamily`, `fontSize`, `fontStyle`, `fontWeight`. They can be different depending on button state: normal, highlighted, disabled and focused.

- normal state: as you would guess
- highlighted state: when the button is pressed and held
- disabled: when the button is disabled; won't show highlighted state
- focused: I wasn't able to observe it, but it's included nonetheless

> [!important]
> Changing the values at runtime doesn't work for iOS 26, and has minor bugs on iOS 27. I suggest testing the changes on iOS 18 and 27.

## Changes

- added text attributes to button and prominent button for each button state

https://github.com/user-attachments/assets/7a9a3808-5aa7-4dbf-8bec-b7507c60a1d3

## Test plan

Use `test-stack-header-item-appearance-ios`

## Checklist

- [ ] Ensured that CI passes

---


kmichalikk:
I need your input on what to do with this:

When resetting text color, even clearing the appearance entirely, the text switches to lightblue for no apparent reason. Confirmed on iOS 18 and iOS 27, breakpoints show nil appearance assigned. In UIKit comments the default is stated to be black. I don't want to put workarounds in the code to handle this.

https://github.com/user-attachments/assets/82108763-a8b3-4534-9d24-851ffc8e8a34

src/components/stack/header/StackHeaderConfig.ios.types.ts
```diff
@@ -606,6 +1026,13 @@ export interface StackHeaderConfigPropsIOS {
    * @summary Appearance of the header when the edge of scrollable content
    * is not aligned with the edge of the header.
    *
+   * @description
+   * Limitations:
+   * - On iOS 26, header items resolve the `buttonFont*` / `prominentButtonFont*` attributes once,
+   *   from whichever appearance is in effect at the moment the items are attached to the navigation bar,
+   *   and do not update on property changes or scroll edge transitions.
+   * - On iOS 27, changing text color at runtime doesn't work reliably, other props appear to be working
```
kligarski, line 1034: nit: not sure if we should state in the docs that something "appears to be working", maybe we should just remove the final part after comma. Applies to both appearance props.

src/components/stack/header/StackHeaderConfig.ios.types.ts
```diff
@@ -455,6 +455,426 @@ export interface StackHeaderAppearanceIOS {
...
+   * @summary Specifies the font family used for plain header items in the
+   * highlighted (pressed) state.
+   *
+   * @description When unset, UIKit derives the highlighted (pressed) appearance from
+   * the normal state. Explicitly set per-state attributes merge into the state's defaults,
+   * so e.g. a font-only group keeps the system highlighted (pressed) text color.
+   *
+   * @platform ios
+   */
+  buttonHighlightedFontFamily?: TextStyle['fontFamily'] | undefined;
```
kligarski, line 558: I wonder if we should match iOS Tabs appearance with nested objects or keep flattened hierarchy here. Recently, we prefer the flattened hierarchies but in this case I'm not sure if it's better for the user. For example: 

```ts
appearance: {
  prominentButtonFontFamily: ...,
  prominentButtonFontSize: ...,
  prominentButtonFontWeight: ...,
  prominentButtonFontStyle: ...,
  prominentButtonFontColor: ...,
  prominentHighlightedButtonFontFamily: ...,
  prominentHighlightedButtonFontSize: ...,
  prominentHighlightedButtonFontWeight: ...,
  prominentHighlightedButtonFontStyle: ...,
  prominentHighlightedButtonFontColor: ...,
}

// vs

appearance: {
  prominentButton: {
    normal: {
      fontFamily: ...,
      fontSize: ...,
      fontWeight: ...,
      fontStyle: ...,
      fontColor: ...,
    },
    highlighted: {
      fontFamily: ...,
      fontSize: ...,
      fontWeight: ...,
      fontStyle: ...,
      fontColor: ...,
    }
  }
}


```

apps/src/tests/single-feature-tests/stack-v5/test-stack-header-item-appearance-ios/scenario.md
```diff
@@ -15,7 +17,13 @@ TBD
 
 - iOS / iPadOS simulator
 
-## Steps on iPhone
+## Note
+
+- runtime changes of appearance don't work on iOS 26, but does work on 18 and works partially on 27 (there may be bugs, or some may not update)
+- runtime switch between `standardAppearance` and `scrollEdgeAppearance` doesn't work on iOS 26
+- menu overflow doesn't appear on iOS 18 by default, some steps won't work
```
kligarski, line 24: "some steps won't work" is not a great instruction for the tester, on the other hand I don't think we should create separate steps for iOS 18, 26 and 27 because it's going to take a very long time to create and test every time. To be honest, I'm not sure what to do here.

kligarski (approved):

PR #4624, opened 2026-09-10
feat(iOS, Stack v5): Add back button appearance

kmichalikk:
## Description

This PR adds `backButtonAppearance` prop which is identical to other button appearances besides not including `disabled` state. When not specified, the back button defaults to regular `buttonAppearance`.

> [!note]
> For iOS >= 26, the back button text is not displayed by default. If custom title is specified, it is only displayed if there is enough room (both header items and the title has higher priority). The example makes the screen title empty so the back title fits and is displayed.

## Changes

- Added `backButtonAppearance` prop
- Updated SFT

https://github.com/user-attachments/assets/eec48642-15a7-44c7-b83a-539179ff97d3

> [!important]
> There's a bug on iOS 27 when using back button with custom title and forcing it to collapse / show multiple times, that I submitted [here](https://developer.apple.com/forums/thread/846016). The button may also flash sometimes, this is also native.

## Test plan

Use `test-stack-header-item-appearance-ios`.

## Checklist

- [ ] Ensured that CI passes

---


src/components/stack/header/StackHeaderConfig.ios.types.ts
```diff
@@ -470,44 +470,49 @@ export interface StackHeaderAppearanceIOS {
   /**
    * @summary Specifies the font family used for plain header items.
    *
-   * @description Applies to the text of header items with the `plain` variant,
-   * including the back button title.
+   * @description Applies to the text of header items with the `plain` variant.
+   * Also applies to the back button title, but only while no `backButton*`
```
kligarski, line 474: I'm not sure if it's clear that this applied to `backButton*` props only in `StackHeaderAppearanceIOS`

apps/src/tests/single-feature-tests/stack-v5/test-stack-header-item-appearance-ios/scenario.md
```diff
@@ -81,14 +81,33 @@ TBD
     
     - [ ] The back button title follows the "button" slot configuration
 
-10. On the details screen, enable "scrollEdgeAppearance" and configure the slots
+10. On the details screen, configure the "backButton" slot differently from the
```
kligarski, line 84: Maybe we should add some appearance presets so that the configuration doesn't take that much time and so that the tester doesn't need to think about what props should they select for each appearance?

kligarski (commented):
1. The back button flashes - is this expected?

https://github.com/user-attachments/assets/f1469785-cb97-473c-8bf9-d31803344970

2. Changes in runtime are buggy but I guess it's just UIKit. After popping and pushing it seems to correct itself.

https://github.com/user-attachments/assets/b5b085ab-6bc9-44f0-b82e-8abfa4d7f88d

PR #4629, opened 2026-09-10
fix(iOS): create headerSearchPaths list dynamically for spm

kacperzolkiewski:
## Description

Create headerSearchPaths list dynamically for spm

## Changes

Change static `headerSearchPaths` list into dynamic headers lookup.

N/A

## Test plan

CI pass

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,43 +1,30 @@
 // swift-tools-version: 6.0
+import Foundation
 import PackageDescription
 
-// Header search paths for quoted cross-directory imports (SPM has no CocoaPods header maps).
-// Keep in sync when adding new ios/ subdirectories that contain headers.
-let headerSearchPaths: [String] = [
-    "ios",
-    "ios/bridging",
-    "ios/conversion",
-    "ios/helpers/container",
-    "ios/helpers/image",
-    "ios/helpers/scroll-view",
-    "ios/legacy",
-    "ios/legacy/events",
-    "ios/legacy/integrations",
-    "ios/legacy/utils",
-    "ios/modals/form-sheet",
-    "ios/modals/utils",
-    "ios/safe-area",
-    "ios/scroll-to-top-guard",
-    "ios/scroll-view-marker",
-    "ios/scroll-view-marker/conversion",
-    "ios/split",
-    "ios/split/conversion",
-    "ios/stack",
-    "ios/stack/conversion",
-    "ios/stack/header",
-    "ios/stack/host",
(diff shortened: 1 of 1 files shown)


t0maboro (commented):
When I add a new header, I don't see `headerSearchPaths` being updated. The cache only seems to get invalidated after I make some change to `Package.swift`. Is this expected?

```sh
swift package dump-package > log1.txt
mkdir -p ios/newdir && echo '// x' > ios/newdir/NewHeader.h 
swift package dump-package > log2.txt
```

At this point,
`log1 === log2`

After making some change in `Package.swift`

```sh
echo '// x' >> Package.swift
swift package dump-package > log3.txt
```

`log1 !== log3`

and new headerSearchPaths are appended.

Thus, I'm not convinced whether we shouldn't update headerSearchPaths explicitly.

kacperzolkiewski:
> When I add a new header, I don't see `headerSearchPaths` being updated. The cache only seems to get invalidated after I make some change to `Package.swift`. Is this expected?
> 
> ```shell
> swift package dump-package > log1.txt
> mkdir -p ios/newdir && echo '// x' > ios/newdir/NewHeader.h 
> swift package dump-package > log2.txt
> ```
> 
> At this point, `log1 === log2`
> 
> After making some change in `Package.swift`
> 
> ```shell
> echo '// x' >> Package.swift
> swift package dump-package > log3.txt
> ```
> 
> `log1 !== log3`
> 
> and new headerSearchPaths are appended.
> 
> Thus, I'm not convinced whether we shouldn't update headerSearchPaths explicitly.

@t0maboro  Yes, the cache is only invalidated for SPM when `Package.swift` changes, as you noticed. So that's the downside of dynamic header imports: whenever you update the header lists, you'd need to clean node_modules, iOS/build and rebuild iOS to invalidate the cache.

t0maboro:
> So that's the downside of dynamic header imports: whenever you update the header lists, you'd need to clean node_modules, iOS/build and rebuild iOS to invalidate the cache.

For current development, especially as we're about to start extracting the core and files begin moving around locations more frequently, I think this could become frustrating. It might be worth getting a second opinion on this, but I'd prefer keeping headerSearchPaths explicit and having a failed CI check if they aren't updated accordingly.

A full clean of `node_modules` and the `ios` directory sounds like a regression compared to what we currently do with CocoaPods, and also compared to what we do with explicit paths in SPM so far.

PR #4632, opened 2026-09-10
feat(iOS, Tabs): add `tabBarHiddenAnimationEnabled` prop (#4627)

### LINKED ISSUE #4627: iOS: tab bar visibility is never animated — setTabBarHidden:animated: is hardcoded to NO
### Description

`RNSTabsHostComponentView` applies the `tabBarHidden` prop with the animation flag hardcoded to `NO`:

https://github.com/software-mansion/react-native-screens/blob/main/ios/tabs/host/RNSTabsHostComponentView.mm#L246

```objc
if (newComponentProps.tabBarHidden != oldComponentProps.tabBarHidden) {
  _tabBarHidden = newComponentProps.tabBarHidden;
#if RNS_IPHONE_OS_VERSION_AVAILABLE(18_0)
  if (@available(iOS 18.0, *)) {
    [_controller setTabBarHidden:_tabBarHidden animated:NO];
  } else
#endif
  {
    _controller.tabBar.hidden = _tabBarHidden;
  }
}
```

`-[UITabBarController setTabBarHidden:animated:]` exists precisely so the transition can be animated, but the animated variant is never requested and there is no prop to ask for it. The tab bar is therefore always drawn or not drawn, never moved.

On iOS 26 this is very visible, because the Liquid Glass tab bar is a floating pill over the content: toggling `hidden` makes it pop in and out instead of sliding. In our app it is the difference between a hide-on-scroll that reads as motion and one that reads as a flash, and we have had it reported as a bug by testers.

Changing the flag to `YES` produces the transition UIKit already knows how to draw. We are currently shipping exactly that as a one-line patch:

```diff
-      [_controller setTabBarHidden:_tabBarHidden animated:NO];
+      [_controller setTabBarHidden:_tabBarHidden animated:YES];
```

A prop would be better than flipping the default, since some apps may rely on the instant behaviour — something like `tabBarHiddenAnimated` (defaulting to the current `NO` to stay backwards compatible), or accepting an animation config alongside `tabBarHidden`. Happy to open a PR for whichever shape you prefer.

Note the pre-iOS-18 branch (`_controller.tabBar.hidden = ...`) cannot animate at all, so this only concerns the iOS 18+ path.

### Steps to reproduce

1. Render `NativeTabs` (expo-router) on iOS 26 with the default Liquid Glass tab bar.
2. Drive the `hidden` prop from scroll position, so it becomes `true` when the user scrolls down and `false` when scrolling stops or reverses.
3. Scroll down until the bar hides, then let it come back.
4. Observe the bar appear instantly, with no transition — most obvious when the bar returns over contrasting content.

No isolated repro yet — the behaviour follows directly from the hardcoded flag quoted above, and reproduces with any component that toggles `tabBarHidden`. Happy to put a minimal example together if that would help triage.

### Screens version

4.26.2 (also verified unchanged on 4.27.0, on `main`, and on 5.0.0-alpha.2 — line 246 there)

### React Native version

0.85.3 (Expo SDK 56)

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Build type

Debug mode

### Device

iOS simulator, Real device

### Device model

iPhone 17 Pro simulator (iOS 26.5), iPhone 16 Pro Max (iOS 26.x)

### Acknowledgements

Yes

ilcato:
## Description

`tabBarHidden` is applied on iOS 18+ through `[UITabBarController setTabBarHidden:animated:]`, with the `animated` flag hardcoded to `NO`. UIKit is able to animate that transition, but the library never asks for it, so the tab bar can only pop in and out. On iOS 26, where the Liquid Glass tab bar floats over the content, this reads as a flash rather than as a bar arriving or leaving.

This PR adds the `ios.tabBarHiddenAnimationEnabled` prop to `TabsHost`, as agreed with @kkafar in the linked issue. It defaults to `true`, so the animated transition becomes the out-of-the-box experience; apps that rely on the instant behaviour can opt out with `false`.

Closes #4627.

## Changes

- `TabsHostIOSNativeComponent.ts`: new codegen prop `tabBarHiddenAnimationEnabled` (`WithDefault<boolean, true>`).
- `TabsHost.ios.types.ts`: new documented `tabBarHiddenAnimationEnabled` prop in `TabsHostPropsIOS`; `TabsHost.ios.tsx` forwards `ios?.tabBarHiddenAnimationEnabled` to the native component.
- `TabsHost.types.ts`: `tabBarHidden` docs now point to the iOS prop that controls the animation.
- `RNSTabsHostComponentView.{h,mm}`: new `tabBarHiddenAnimationEnabled` property (default `YES` in `resetProps`), read in `updateProps` **before** `tabBarHidden` so that both props changing in the same update use the new setting, and passed as the `animated:` argument.
- `test-tabs-tab-bar-hidden` SFT: iOS-only switch for `ios.tabBarHiddenAnimationEnabled`, and `scenario.md` updated with the animated expectation plus two steps that exercise the opt-out.

Notes for reviewers:

- The prop is placed under `ios` because the animation only exists on the iOS path; on Android `tabBarHidden` toggles `isVisible` on the `BottomNavigationView` and there is nothing to animate. Happy to move it to the top level if you'd rather keep it next to `tabBarHidden`.
- On iOS < 18 the `tabBar.hidden` fallback is unchanged and the prop is a no-op, as documented.
- `Detox` synchronises with UIKit animations, so the existing `test-tabs-tab-bar-hidden.e2e.ts` assertions should keep passing with the animated default. I could not run the iOS e2e suite locally; please let me know if you'd like the e2e test extended to cover the new switch as well.

Recorded on an iPhone 17 Pro simulator, iOS 26.5, `FabricExample` → `Single Feature Tests` → `Tabs` → `Tab Bar Hidden`. The first two hide/show cycles run with `ios.tabBarHiddenAnimationEnabled: true` (the new default), then the switch is turned off and the two following cycles show the previous, immediate behaviour.

https://github.com/user-attachments/assets/fbfd5338-500c-4e6c-9362-6802b6a07a3e

Measured on the recording (30 fps), counting frames in which the tab bar strip is neither fully shown nor fully hidden:

| Cycle | `tabBarHiddenAnimationEnabled` | Hide | Show |
| --- | --- | --- | --- |
| 1 | `true` | 0 transition frames | 4 transition frames (~130 ms) |
| 2 | `true` | 1 transition frame | 4 transition frames (~130 ms) |
| 3 | `false` | 0 transition frames | 0 transition frames |
| 4 | `false` | 0 transition frames | 0 transition frames |

UIKit animates the alpha of the bar rather than sliding it, so the visible difference is a short fade on reveal instead of a single-frame pop.

## Test plan

- `yarn lint-js`, `yarn check-types` and `prettier --check` on the changed files pass locally.
- `RNSTabsHostComponentView.{h,mm}` are clean under the repository `.clang-format`.
- `FabricExample` builds for the iOS Simulator with the change (Xcode 26.6, RN 0.87).
- Manual: `Single feature tests` → `Tabs` → `Tab Bar Hidden` (`apps/src/tests/single-feature-tests/tabs/test-tabs-tab-bar-hidden`), following the updated `scenario.md`: toggle `tabBarHidden` with the animation switch on (default) and off, and compare.

## Checklist

- [ ] Ensured that CI passes

### DIFF
--- a/apps/src/tests/single-feature-tests/tabs/test-tabs-tab-bar-hidden/scenario-description.ts
+++ b/apps/src/tests/single-feature-tests/tabs/test-tabs-tab-bar-hidden/scenario-description.ts
@@ -4,7 +4,7 @@ export const scenarioDescription: ScenarioDescription = {
   name: 'Tab Bar Hidden',
   key: 'test-tabs-tab-bar-hidden',
   details:
-    'Test tabBarHidden prop on TabsHost - toggle to show/hide the tab bar at runtime',
+    'Test tabBarHidden prop on TabsHost - toggle to show/hide the tab bar at runtime. On iOS, also toggles ios.tabBarHiddenAnimationEnabled to compare animated and immediate transitions.',
   platforms: ['ios', 'android'],
   e2eCoverage: 'full',
   smokeTest: false,
--- a/apps/src/tests/single-feature-tests/tabs/test-tabs-tab-bar-hidden/index.tsx
+++ b/apps/src/tests/single-feature-tests/tabs/test-tabs-tab-bar-hidden/index.tsx
@@ -1,6 +1,6 @@
 import { SettingsSwitch } from '@apps/shared/SettingsSwitch';
 import React from 'react';
-import { ScrollView, Text } from 'react-native';
+import { Platform, ScrollView, Text } from 'react-native';
 import { scenarioDescription } from './scenario-description';
 import { createScenario } from '@apps/tests/shared/helpers';
 import {
@@ -25,6 +25,17 @@ function ConfigScreen() {
         onValueChange={value => updateHostConfig({ tabBarHidden: value })}
         testID="tab-bar-hidden-switch"
       />
+      {Platform.OS === 'ios' && (
+        <SettingsSwitch
+          style={{ marginBottom: 15 }}
+          label="ios.tabBarHiddenAnimationEnabled"
+          value={hostConfig.ios?.tabBarHiddenAnimationEnabled ?? true}
+          onValueChange={value =>
+            updateHostConfig({ ios: { tabBarHiddenAnimationEnabled: value } })
+          }
+          testID="tab-bar-hidden-animation-enabled-switch"
+        />
+      )}
     </ScrollView>
   );
 }
--- a/apps/src/tests/single-feature-tests/tabs/test-tabs-tab-bar-hidden/scenario.md
+++ b/apps/src/tests/single-feature-tests/tabs/test-tabs-tab-bar-hidden/scenario.md
@@ -2,7 +2,7 @@
 
 ## Details
 
-**Description:** This test scenario focuses on the visibility management of the tab bar. It validates the tabBarHidden property, ensuring that the UI responds dynamically to state changes without layout shifts or persistence errors.
+**Description:** This test scenario focuses on the visibility management of the tab bar. It validates the tabBarHidden property, ensuring that the UI responds dynamically to state changes without layout shifts or persistence errors. On iOS, it also validates the `ios.tabBarHiddenAnimationEnabled` property, which controls whether the transition is animated (iOS 18+).
 
 **OS test creation version:** iOS: 18.6 and 26.2, Android: API Level 36.
 
@@ -23,8 +23,16 @@ Full: Covers all manual scenario steps.
 
 2. Toggle `tabBarHidden` to `true`.
 
-- [ ] Tab bar should disappear immediately.
+- [ ] Tab bar should disappear. On iOS 18+, the transition should be animated (`ios.tabBarHiddenAnimationEnabled` defaults to `true`); on Android and iOS < 18 the tab bar disappears immediately.
 
 3. Toggle back to `false`.
 
-- [ ] Tab bar should reappear.
+- [ ] Tab bar should reappear, with the same animation behavior as in step 2.
+
+4. (iOS only) Toggle `ios.tabBarHiddenAnimationEnabled` to `false`, then repeat steps 2-3.
+
+- [ ] Tab bar should disappear and reappear immediately, without animation.
+
+5. (iOS only) Toggle `ios.tabBarHiddenAnimationEnabled` back to `true`, then repeat steps 2-3.
+
+- [ ] Tab bar should disappear and reappear with animation again.
--- a/ios/tabs/host/RNSTabsHostComponentView.h
+++ b/ios/tabs/host/RNSTabsHostComponentView.h
@@ -44,6 +44,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, readonly) BOOL tabBarHidden;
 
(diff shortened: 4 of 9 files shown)


apps/src/tests/single-feature-tests/tabs/test-tabs-tab-bar-hidden/scenario-description.ts
```diff
@@ -4,7 +4,7 @@ export const scenarioDescription: ScenarioDescription = {
   name: 'Tab Bar Hidden',
   key: 'test-tabs-tab-bar-hidden',
   details:
-    'Test tabBarHidden prop on TabsHost - toggle to show/hide the tab bar at runtime',
+    'Test tabBarHidden prop on TabsHost - toggle to show/hide the tab bar at runtime. On iOS, also toggles ios.tabBarHiddenAnimationEnabled to compare animated and immediate transitions.',
```
kkafar:
```suggestion
    'Test tabBarHidden prop on TabsHost - toggle to show/hide the tab bar at runtime.',
```

This is an unnecessary change. It's supposed to be a short description displayed in the example app.

kkafar (commented):
Thanks for the PR!

I've tested the runtime on iOS and it indeed works as expected. 

https://github.com/user-attachments/assets/538fd512-abcf-4313-912f-afb178b73025

The native code is fine, only a single remark to the test metadata.

Thinking out loud here: 
One thing I wonder is whether we can have feature parity with Android here. If yes, then we'd put this prop in different place - `TabsHost.tabBarHiddenAnimationEnabled` instead of `TabsHost.ios.tabBarHiddenAnimationEnabled`. I did some initial research and it seems that unfortunately Material's BottomNavigationView does not support such behaviour of of the box. 

We'll see however later whether we can come up with some sensible implementation. 

Let's keep this PR limited do iOS for now. Later we'll come up with something for Android before 5.0 stabilisation.

ilcato:
@kkafar I agree with you on keeping this as an iOS only PR.

kkafar (approved):
Thanks! 

Looks good then. Let's have the CI pass and we'll land it.

kkafar (approved):
I've also verified that during the very first render, if: `tabBarHidden === true && tabBarHiddenAnimationEnabled === true` there is no glitch, where the tab bar would animate away AFTER first drawn frame. At least I can not observe any such animation.

https://github.com/user-attachments/assets/5be2aec9-753e-4d2d-bce9-40079aba7126

kligarski:
> I've also verified that during the very first render, if: `tabBarHidden === true && tabBarHiddenAnimationEnabled === true` there is no glitch, where the tab bar would animate away AFTER first drawn frame. At least I can not observe any such animation.

@kkafar Did you check iOS 18? Please see https://github.com/software-mansion/react-native-screens-labs/issues/517.

kkafar:
No I did not. Let me verify it immediately.

kkafar:
There is a problem indeed - I'll patch it soon. 


https://github.com/user-attachments/assets/5dd3e023-21af-4909-b2dc-a8ccfe58c6c2

PR #4091, opened 2026-05-24
fix(iOS): remove FormSheet scroll view KVO observer synchronously on invalidate

phewphewb:
## Description

Fixes https://github.com/software-mansion/react-native-screens/issues/4090

After dismissing FormSheet, pushed screen ScrollView is clipped to sheet height. 
FormSheet KVO observer not removed before scroll view is reused by the next screen. 

## Changes

- Remove the FormSheet scroll view KVO observer synchronously at the start of `-invalidateImpl`, instead of inside the `dispatch_async` block introduced in #3728
- Keep deferring `_controller - nil` to the async block (unchanged)
-  Add `Test4090` as a minimal repro for dismissing a form sheet and immediately pushing a screen with a full-height `ScrollView`

#### Before

https://github.com/user-attachments/assets/a7d056a2-01a5-4cd9-b8dd-0ed90b59e325 

#### After

https://github.com/user-attachments/assets/6a7b2eb5-0ef1-48f1-82ad-e7ee21a1c6a2

## Checklist

- [x] Included code example that can be used to test this change.
- [x] For visual changes, included screenshots / GIFs / recordings documenting the change.
- [ ] For API changes, updated relevant public types.
- [ ] Ensured that CI passes

Issue #4090, opened 2026-05-24
After dismissing FormSheet, pushed screen ScrollView is clipped to sheet height


phewphewb:
### Description

On iOS, dismissing a **form sheet** and immediately **pushing** another screen can leave the pushed screen’s `ScrollView` clipped to the former sheet height (~50% of the screen).

## Video

Reproduced in **FabricExample**.

https://github.com/user-attachments/assets/28d3a742-e82a-4a23-9590-0aabb7664ad3

## Expected behavior

Push screen’s scroll view uses full screen height; scrolling covers the full viewport.

## Actual behavior

Scroll view only scrolls within ~sheet height, appears clipped.

### Steps to reproduce

1. Present a screen with `presentation: 'formSheet'` and `sheetAllowedDetents: [0.5]`.
2. From the sheet, dismiss and navigate to a push screen in one handler (`goBack()` + `navigate(...)`).
3. On the push screen, use a full-height `ScrollView` with `headerShown: true` and `headerTransparent: true`.

(TestNEXT in the provided repro repo, can be renamed to the issue number later)

### Snack or a link to a repository

https://github.com/software-mansion/react-native-screens/commit/9ae08e359fb69b7d610bc394c208cb610d700649

### Screens version

main

### React Native version

0.85.0

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Acknowledgements

Yes

phewphewb:
> Hey! 👋
> 
> The issue doesn't seem to contain a [minimal reproduction](https://stackoverflow.com/help/minimal-reproducible-example).
> 
> Could you provide a [snack](https://snack.expo.dev/) or a link to a GitHub repository under your username that reproduces the problem?

It does have a repro example. The repository link is in the description.

PR #4652, opened 2026-09-13
fix(iOS): skip form sheet scroll view frame correction once the screen is invalidated

### LINKED ISSUE #4651: [iOS] Pushed screen's ScrollView is clipped to the dismissed formSheet's height (invalidated sheet still corrects the recycled scroll view)
### Description

On iOS (Fabric), when a `formSheet` screen is replaced by a pushed screen in the same navigation action (`router.replace` from the sheet in expo-router, or `goBack()` + `navigate()` as in `Test4090`), the pushed screen's `ScrollView` comes up sized to the sheet's detent height instead of the screen. With `sheetAllowedDetents: [0.5]` on an iPhone 17 Pro Max the `UIScrollView` frame is `{0, 0, 440, 464}` while its content view is laid out for the full 956pt screen: everything below 464pt is neither painted nor tappable, and the view does not scroll. Every later screen that Fabric hands the same recycled `UIScrollView` instance is clipped the same way until the app restarts.

This is the symptom of #4090. #4091 (in 4.26.0) removes the sheet's KVO observer synchronously in `invalidateImpl`, but that is not the path that resizes the scroll view here. With `NSLog` in `correctScrollViewFrame:` and `invalidateImpl` on 4.26.0:

```
17:10:32.309  invalidate screen=0x37b0ba200 pres=6 sv=0x155a6c380
17:10:32.313  correct    screen=0x37b0ba200 pres=6 frame={{0, 0}, {440, 464}} sv=0x155a6c380 svframe={{0, 0}, {440, 956}}
```

The dismissed sheet's `RNSScreenView` (`pres=6` = formSheet) gets one more layout pass 4ms after it has been invalidated, `applyFrameCorrectionForDescendantScrollView` still finds the scroll view down its old subview chain, and `correctScrollViewFrame:` sets that scroll view — which Fabric has already recycled into the pushed screen and laid out at 956pt — back to the sheet's 464pt frame. The KVO observer was already removed, so #4091's guard never runs.

Still reproduces on 4.27.0 (latest) and the code in `applyFrameCorrectionForDescendantScrollView` / `updateBounds` is unchanged on `4.28-stable` and `main` (`ios/legacy/RNSScreen.mm`).

**Suggested fix:** `RNSScreenView` already tracks `invalidated` (set in `invalidateImpl`, since 4.27.0). Returning early from `applyFrameCorrectionForDescendantScrollView` when it is set fixes this; verified in a production app on 4.26.0 with an equivalent patch (four sheet → push round trips and later pushes all full height).

```objc
- (void)applyFrameCorrectionForDescendantScrollView
{
  if (_invalidated) {
    return;
  }
  RCTScrollViewComponentView *scrollView = [self tryFindDescendantScrollView];
  ...
```

Screens with a loading state that mount their `ScrollView` a frame later are unaffected, which is why this is easy to miss in example apps.

### Steps to reproduce

1. Clone the repository below, `npm install`, `npx expo run:ios`.
2. Tap **Open sheet** (a `formSheet` with `sheetAllowedDetents: [0.5]` and a `ScrollView`).
3. Tap **Open detail 1** (the row calls `router.replace("/detail/1")`; the detail screen is a plain full-height `ScrollView` with 30 rows).
4. Only rows 0–9 render; the rest of the list is missing and cannot be scrolled to. Go back and tap **Push detail 1 directly** to see the same screen at full height.

https://github.com/saddlepaddle/rns-formsheet-clips-pushed-screen

### Screens version

4.26.0 (also reproduced on 4.27.0)

### React Native version

0.86.3 (Expo SDK 57)

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Build type

Debug mode

### Device

iOS simulator

### Device model

iPhone 17 Pro Max (iOS 26.0)

### Acknowledgements

Yes

saddlepaddle:
## Description

Fixes #4651 (same symptom as #4090).

On iOS (Fabric), when a `formSheet` is replaced by a pushed screen in one navigation action (`goBack()` + `navigate()`, or expo-router's `router.replace`), the pushed screen's `ScrollView` comes up sized to the sheet's detent height instead of the screen, and so does every later screen that Fabric hands the same recycled `UIScrollView` instance.

#4091 removes the sheet's KVO observer synchronously in `invalidateImpl`, but that is not the path that resizes the scroll view. The dismissed sheet's `RNSScreenView` still receives a layout pass after it has been invalidated; `updateBounds` → `applyFrameCorrectionForDescendantScrollView` finds the scroll view down its old subview chain, and `correctScrollViewFrame:` sets it back to the sheet frame. By then Fabric has recycled that scroll view into the new screen and laid it out at full height. With `NSLog` on 4.26.0:

```
17:10:32.309  invalidate screen=0x37b0ba200 pres=formSheet sv=0x155a6c380
17:10:32.313  correct    screen=0x37b0ba200 pres=formSheet frame={{0, 0}, {440, 464}} sv=0x155a6c380 svframe={{0, 0}, {440, 956}}
```

## Changes

- `ios/legacy/RNSScreen.mm`: `applyFrameCorrectionForDescendantScrollView` returns early once the screen has been invalidated (`_invalidated`, set in `invalidateImpl`). An invalidated screen has been deleted by React; it must not size a scroll view it no longer owns.

`Test4651`, FabricExample on an iPhone 17 Pro Max simulator (iOS 26), after **Open FormSheet → Open Detail (row 1)**. The sheet is `presentation: "formSheet"` with `sheetAllowedDetents: [0.5]`; the detail screen is a plain full-height `ScrollView` with 40 rows.

| Before (main) | After |
| --- | --- |
| Before | After |

Before: the detail `UIScrollView` frame is `{0, 0, 440, 464}` (the `[0.5]` detent); rows 10–39 are missing and the list does not scroll. After: full height, all rows scrollable, also on a second sheet → row round trip.

## Test plan

New `apps/src/tests/issue-tests/Test4651.tsx`. `Test4090` does not trip this on `main` because its Purple screen wraps the `ScrollView` in a `View`; `Test4651` has the `ScrollView` as the screen's only child on both the sheet and the pushed screen, which is what the reporting app (and expo-router's form sheets) do.

1. `FabricExample`, `apps/App.tsx` rendering `Test4651`, iPhone 17 Pro Max simulator (iOS 26).
2. Tap **Open FormSheet**, then **Open Detail (row 1)**.
3. Before this change the detail scroll view is clipped to the sheet height (see captures); after it fills the screen. **Push Detail directly** from Home is full height in both cases.
4. Also verified in the reporting app (react-native-screens 4.26.0 with the same early return as a patch): four sheet → push round trips and later pushes all at full height.

Minimal standalone reproduction (stock `create-expo-app`, SDK 57, expo-router): https://github.com/saddlepaddle/rns-formsheet-clips-pushed-screen

## Checklist


kkafar (commented):
I confirm that I can reproduce the problem when I revert the fix:

https://github.com/user-attachments/assets/3d5f191e-24a9-4f69-a85c-19e810c9b2de

I think that the fix is also sound - it does not make sense to update a frame from an invalidated view. My expectation is that that - if a view is invalidated, that means it has received `DELETE` mutation - it is already detached from the hierarchy anyway, because there had to be a preceding `REMOVE` mutation. Therefore I don' understand one thing: how come that the `RNSScreenView` still receives a layout pass! 

I need to verify this before I approve.

kkafar (approved):
Okay, I understand it now!

The thing is simple - the view has indeed received `REMOVE` mutation beforehand, however `RNSScreenStack` does not unmount the removed view from HostTree immediately - it keeps it attached and only **schedules** a block on the main queue - which likely runs in next main loop iteration, while layout is calculated from "before sleep" main run loop step (before commands are sent to render server).

Everything seems reasonable now. Thank you!

Issue #4347, opened 2026-07-16
[iOS] Metro-bundled PNG can flash in native header item during transition


maxlapides:
### Description

On iOS, images used by native stack header bar button items can be assigned after UIKit's initial layout in release builds. This can cause an empty button or fallback label to flash during a transition, and can leave the button with an incorrect initial width.

For example, React Navigation's native-stack API accepts an image icon and converts it to react-native-screens' `imageSource` or `templateSource` shape:

```tsx
unstable_headerRightItems: () => [
  {
    type: 'button',
    label: 'Add',
    icon: {
      type: 'image',
      source: require('./header-add.png'),
    },
    onPress: () => {},
  },
]
```

In `RNSBarButtonItem`, image and template sources are resolved through `RNSImageLoadingHelper` / `RCTImageLoader`. The helper notes that synchronous loading is not guaranteed in release mode. The completion block can therefore set `self.image` after the bar button item has already been laid out.

The same artwork resolves correctly and synchronously when compiled into the client app's asset catalog and passed through the existing `{ type: 'xcasset', name }` path, because `RNSBarButtonItem` calls `UIImage imageNamed:` directly.

Expected behavior: local or named asset images used by native header items are available for the initial UIKit layout and appear consistently during transitions.

Actual behavior: the image may be missing during initial layout in an iOS release build. Switching to the direct XCAsset path fixes the issue.

https://github.com/user-attachments/assets/1f06ed09-c9ba-48c8-80a2-681867a5ebea

One possible fix, suggested in the related React Navigation discussion, would be to internally use the XCAsset / `UIImage imageNamed:` path when an asset name is supplied through an image source such as `{ uri: 'asset_name' }`.

Related React Navigation discussion: https://github.com/react-navigation/react-navigation/pull/13183

### Steps to reproduce

1. Configure a native stack screen with an `unstable_headerRightItems` button whose icon uses a required local image, as shown above.
2. Build and run the app in release mode on iOS with the New Architecture enabled.
3. Navigate to and from the screen repeatedly so that UIKit creates and lays out the native header item during a transition.
4. Observe that the image can be absent during the initial layout. An empty label produces a temporarily empty button; a visible fallback label can flash before the image arrives, and the initial width can be incorrect.
5. Compile the same image into the main app XCAsset catalog and pass it as `{ type: 'xcasset', name: 'header-add' }` directly to react-native-screens. The icon resolves synchronously and the layout issue no longer occurs.

[...]

PR #4672, opened 2026-09-17
feat: Add a preferred loading mode for header images

erickreutz:
Refs #4347.

Local PNG header icons can arrive after UIKit's initial bar-button layout in Release builds. With Liquid Glass, buttons can appear empty and oversized until the icons arrive. Expo Updates can resolve images to local cache files outside the main bundle, including during an embedded launch.

This adds an opt-in loading preference to image-backed header icons:

```tsx
{
  type: 'imageSource',
  imageSource: require('./icon.png'),
  preferredLoadingMode: 'synchronous',
}
```

`preferredLoadingMode` accepts `automatic` or `synchronous`. Omitting it preserves existing behavior. The preference also applies to iOS `templateSource` icons. It is a preference rather than a guarantee: unsupported sources and failed reads or decodes fall back to the existing loader. HTTP/HTTPS images stay on that loader.

On iOS, `RNSImageLoadingHelper` attempts synchronous loading for local PNG file URLs, preserving source scale and rendering mode. Both legacy bar-button items and the newer stack header use it. On Android, the header image loader supports local PNG files, packaged asset PNGs, and PNG resources in the app, preserving Android resource density. Toolbar menu items, back icons, and overflow icons pass the preference through. Tab image loading remains unchanged.

Synchronous loading performs file I/O and image construction on the UI thread and bypasses the normal image-loader cache for supported sources. It is intended for small header icons. There are no file-size or pixel-dimension limits, so callers should not opt large images into this path.

The example app exposes the preference in the iOS header-icon and Android toolbar-icon scenarios. `Test4672` exercises legacy iOS push/pop using Screens directly. This PR does not change Expo Router or React Navigation APIs.

## Validation evidence

The upstream example builds in Release for iOS and Android. Five iOS header rendering tests pass, including original and template images with omitted, automatic, and synchronous preferences. The legacy example also pushes and pops on a physical iPhone with the opted-in PNG visible.

Separate native probes exercise the actual loaders: 20 iOS helper cases pass on both iOS 18.6 and 26.5, covering callback timing, source scale, rendering mode, and fallback behavior. Nine Android loading cases pass, including equal resource dimensions between synchronous loading and Fresco. These probes are local validation tools, not part of the library. The committed unit tests cover preference transport and unchanged non-image icons.

## Original regression recordings

These physical-device Release recordings show the original missing-image problem and the earlier unconditional local-PNG patch. They are not new OTA validation of the opt-in API in this revision.

| Before | Earlier patch, embedded launch | Earlier patch, downloaded OTA |
| --- | --- | --- |
| Before | Embedded | OTA |

The baseline uses separate right-side buttons and the patched recordings use a shared group. They demonstrate the missing-image phase, not identical grouping. A brief transition blur remains. The new preference has not yet been retested through a downloaded Expo OTA.


ios/helpers/image/RNSImageLoadingHelper.mm
```diff
@@ -13,6 +13,18 @@ + (void)loadImageSyncIfPossibleFromJsonSource:(nonnull NSDictionary *)jsonImageS
   RCTImageSource *imageSource = [RCTConvert RCTImageSource:jsonImageSource];
   RCTAssert(imageSource != nil, @"[RNScreens] Expected nonnil image source");
 
+  // Expo Updates can resolve local icons outside NSBundle, where RCTImageLoader loads asynchronously.
+  // Supply these PNGs before UIKit lays out and animates the bar button item.
+  NSURL *url = imageSource.request.URL;
+  if (url.isFileURL && [url.pathExtension.lowercaseString isEqualToString:@"png"]) {
+    NSData *data = [NSData dataWithContentsOfURL:url];
+    UIImage *image = data ? [UIImage imageWithData:data scale:imageSource.scale ?: 1.0] : nil;
```
kkafar: How does this check whether the to-be-loaded-file is local or remote?

ios/helpers/image/RNSImageLoadingHelper.mm
```diff
@@ -13,6 +13,18 @@ + (void)loadImageSyncIfPossibleFromJsonSource:(nonnull NSDictionary *)jsonImageS
   RCTImageSource *imageSource = [RCTConvert RCTImageSource:jsonImageSource];
   RCTAssert(imageSource != nil, @"[RNScreens] Expected nonnil image source");
 
+  // Expo Updates can resolve local icons outside NSBundle, where RCTImageLoader loads asynchronously.
+  // Supply these PNGs before UIKit lays out and animates the bar button item.
```
kkafar: RNSImageLoadingHelper is a general purpose helper - remove the comment that refers to some particular view behavior (bar button item here).

kkafar (commented):
Hey!

I have mixed feelings here. I think that if we iron out the patch we could proceed here, but my thoughts are that it might be constraint-depended (e.g. number of icons, size of icons) whether we want to load them synchronously or not. Therefore, we might think of some mechanism (a prop? where should it leave? feature flag? don't like it) to let developer decide whether he wants to load the icon synchronously or not. Maybe this information could be passed together with the icon? Dunno yet. Will return here.

ios/helpers/image/RNSImageLoadingHelper.mm
erickreutz: `url.isFileURL` is the check here. HTTP/HTTPS URLs fall through to the existing loader. We confirmed on the phone that Expo Updates was resolving the image to a local `.expo-internal/….png` file. This reads that cached file; it doesn’t download anything.

ios/helpers/image/RNSImageLoadingHelper.mm
erickreutz: Yep, agreed. I’ll remove the bar-button-specific wording.

ios/helpers/image/RNSImageLoadingHelper.mm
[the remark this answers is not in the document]
erickreutz: Fair point. The current path doesn’t put any limits on the image. If we keep this automatic, checking both file size and pixel dimensions makes sense. I’d like to settle the opt-in vs. automatic behavior with the maintainer first.

erickreutz:
Yeah, that makes sense. Our icons are small, so this works well for us, but I get the concern about making it the default for every local PNG.

A per-image opt-in sounds reasonable. In our case, we’d need to pass that through Expo Router and React Navigation into Screens, so it would involve changes across all three.

Would you lean toward that, or keeping this inside Screens with size limits and falling back to the existing loader for larger images? Happy to work through either approach.

ios/helpers/image/RNSImageLoadingHelper.mm
kkafar: Okay, that's sits with me. Thanks for the explanation.

kkafar:
> A per-image opt-in sounds reasonable. In our case, we’d need to pass that through Expo Router and React Navigation into Screens, so it would involve changes across all three.
> Would you lean toward that, or keeping this inside Screens with size limits and falling back to the existing loader for larger images? Happy to work through either approach.

My thoughts are as follows: 

I'd approach this with UIKit-like API, e.g. [preferredPlacement](https://developer.apple.com/documentation/uikit/uitab/preferredplacement?language=objc). The prop expresses preference, but setting it does give no guarantee what will actually happen. I'd do this, because we can not actually guarantee to load an image synchronously, e.g. if the image is remote we don't want to block the UI thread. Moreover we can later extend this mechanism easily with size / type constraints depending on our needs. 

Another aspect is that per-icon configuration seems favourable - such fine-grained control covers more use cases than only e.g. a global switch. Developer knows his assets and can hint how should they be loaded.

Having that said, I think we need two things:

1. sensible approach to modeling the API and data model underneath.

    Would be great to convey this additional information w/o vast refactor of current 
    data model used for transferring prop information between JS and native. 

2. Support this feature also on Android.

I believe that if we add support in screens, there won't be much resistance to downstream it to expo-router / react-navigation. If we have some solid proposition, I can forward this to respective maintainers, so that they are engaged & aware at early stage.

erickreutz:
Yeah, I like that approach. What about `preferredLoadingMode: 'automatic' | 'synchronous'` on the icon?

```tsx
icon: {
  type: 'imageSource',
  imageSource: require('./icon.png'),
  preferredLoadingMode: 'synchronous',
}
```

Leaving it out would keep things working as they do today. Setting it to `'synchronous'` would mean “try to load this right away.” We’d start with local PNGs, including bundled images and OTA assets already on disk. Anything else, or anything that fails to load, would use the existing loader. That also leaves room to add size limits later.

I’d wire it into the existing icon config and support it on both iOS and Android.

Does that sound like what you had in mind?

erickreutz:
Pushed this up with `preferredLoadingMode` on the image icon:

```tsx
{
  type: 'imageSource',
  imageSource: require('./icon.png'),
  preferredLoadingMode: 'synchronous',
}
```

Omitting it or using `'automatic'` keeps the existing behavior. `'synchronous'` tries to load supported local PNGs immediately and falls back to the existing loader otherwise. Remote images still use the existing loader.

It’s wired into both iOS header implementations and Android’s header icons. This stays within Screens, with no Expo Router or React Navigation changes.

I tested Release builds on both platforms, including push/back on my phone. The loading checks also cover scale, tinting, and fallback behavior. I haven’t rerun the new opt-in API through a downloaded OTA yet.

The tradeoff is still main-thread work, so this is meant for small header icons. I left out size limits since this is now an explicit opt-in.

erickreutz:
@kkafar let me know what you think!

PR #4673, opened 2026-09-17
fix(iOS, Tabs): prevent tab bar hidden animation on first render

kkafar:
## Description

Follow-up to review from <https://github.com/software-mansion/react-native-screens/pull/4632>.

@kligarski pointed out that if we have `tabBarHidden == true` and `tabBarHiddenAnimationEnabled == true`
in first render, then on iOS 18 and below we'll see animation of the tab bar hiding away
after the first render. On iOS > 18 this also happens, but the animation is fade instead of slide
and it does not bother that much.

## Changes

Now, we animate the tab bar only when the tabBar is attached to window.
Important note: hidden tab bar is still attached to the window.

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/38765295-8376-4ee8-b393-e41331f6a41c" alt="before" /> | <video src="https://github.com/user-attachments/assets/63faa6a0-3f21-4576-bf2f-b2fd05bff895" alt="after" /> |

## Test plan

Added dedicated single-feature-test: `test-tabs-tab-bar-initially-hidden`.

## Checklist

- [x] Ensured that CI passes


kligarski (approved):
Looks good, I checked it on iOS 18, 26 and 27. Thanks.

PR #4585, opened 2026-09-02
test(FormSheet v5): Add test for ScrollView nested in FormSheet

t0maboro:
## Description

Adds a test scenario checking that a `ScrollView` nested in the sheet drives it the way the native sheets do: swipe up expands, swipe down at the top edge collapses and then dismisses and dragging the non-scrollable header always moves the sheet.

On Android the `ScrollView` has to opt into nested scrolling (`nestedScrollEnabled`) so that `BottomSheetBehavior` receives its scroll events. Without it the sheet consumes every downward drag and a scrolled list can never be scrolled back by touch. 

Closes: https://github.com/software-mansion/react-native-screens-labs/issues/1275

## Changes

- Added new scenario

## Before & after - visual documentation

| Android | iOS |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/eab5d0bf-d987-42d2-aeea-88b2d3b8dfd3" /> | <video src="https://github.com/user-attachments/assets/c9d73af0-a172-44f3-9982-dd8a0de4a5f5" /> |

## Test plan

Added new scenario

## Checklist

- [ ] Included code example that can be used to test this change.
- [ ] For visual changes, included screenshots / GIFs / recordings documenting the change.
- [ ] For API changes, updated relevant public types.
- [ ] Ensured that CI passes

PR #4675, opened 2026-09-18
feat(iOS, Tabs): Migrate to UITab API for iOS >= 18

kmichalikk:
> [!caution]
> This PR brings a MAJOR change to tabs and MUST be carefully reviewed and tested on all possible scenarios.

## Description

Migrates the children management of `RNSTabBarController` from the legacy `viewControllers`-based API to the modern `UITab`-based API (`UITabBarController.tabs` / `selectedTab`) on iOS 18+. The legacy path remains in place for iOS < 18, and tvOS.

UIKit ties newer tab-bar features to the `UITab` API (e.g. the system search tab treatment, `UISearchTab.automaticallyActivatesSearch`), so adopting it is necessary. 

> [!note]
> `UISearchTab` is not adopted here - left for a followup PR. The `search` system item gets a plain `UITab`, it should behave exactly the same on iOS 26, will work on 27 too after followup.

  ## Changes

- **Children management (iOS 18+)**: children updates build the `tabs` array, reusing existing `UITab` instances by view controller identity. The new tab `identifier` is set to `screenKey` by default, but this may be changed in a followup.
- **New delegate pair** `tabBarController:shouldSelectTab:` / `didSelectTab:previousTab:` mirroring the legacy delegate logic (repeated selection special effect, `preventNativeSelection`). Unlike the legacy delegate, `didSelectTab` also fires for **programmatic** selection changes — those are recognized (navigation state already points at the target) and ignored.
- **Per-tab configuration sync**, split per coordinator responsibilities:
    - new `RNSTabBarItemsCoordinator` — title & badge onto `UITab`s, accessibility onto their `UITabBarItem`s; also absorbed the legacy-path a11y update from the controller (`updateTabBarItemsA11yIfNeededInScreenControllers:`),
    - `RNSTabBarAppearanceCoordinator` — icon mirroring (`syncIconsOfTabs:`, `selectedImage` gated to iOS 26.1+ where the API exists) and `updateUserInterfaceStyleOfTabBar:forScreenView:` (moved out of the controller),
- **Repaint orchestration**: `tabBarItemsDidChange` syncs config and, only when something actually changed, re-sets the same `tabs` array — mutating a live `UITab` in place does **not** repaint the rendered bar buttons; the re-set is the only thing that does. The repaint is skipped while the More controller is active, because `selectedTab` is stale then (More has no `UITab`) and the re-set would yank the selection away. The appearance coordinator calls it after asynchronously loaded icons land (the tab doesn't observe `tabBarItem`).
- **Accessibility**: written to the tab-managed `UITabBarItem` only. The rendered bar buttons read `accessibilityIdentifier`/`accessibilityLabel` from the item, not from the `UITab` (`UITab.accessibilityIdentifier` never reaches them).
- **More navigation controller handling**: with `UITab` API, UIKit selects More with **no tab-bar delegate callback at all**. The universal signal (verified on 18.6 / 26.5 / 27.0) is `UINavigationControllerDelegate` `navigationController:willShowViewController:` on the More controller — it fires for list appearance, hosted-tab re-display, pushes and pop-to-list. The handler:
    - emits `didSelectMoreTab` when the list shows,
    - progresses navigation state when a hosted tab is (re-)displayed (no other delegate reports it),
    - enforces `preventNativeSelection` on re-display (pop to list + prevented event).

    There is one case where `dispatch_async` seems needed and I couldn't find a workaround: iOS 26 updates `selectedViewController` *before* the `willAppear` callbacks, iOS 27 only *after* them, and a synchronous handler would read the outgoing selection on iOS 27, failing the assertion already present in code.
    
    The existing push interceptor (ISA-swizzled `pushViewController:animated:`) still gates More-list row pushes on both paths. Since no delegate callback exists to install the More hooks lazily on the `UITab` path, they are installed eagerly on container updates and size-class changes (UIKit resets the More controller's delegate & ISA whenever it leaves the tab bar); `canHaveMoreNavigationController` gates the `moreNavigationController` getter, which *creates* the controller on access.

Nothing should change visually vs main branch.

## Test plan

> [!caution]
> This PR brings a MAJOR change to tabs and MUST be carefully reviewed and tested on all possible scenarios.

Some of the important things to verify:
- rendering of more controller, restoring correct previously selected tabs
- showing and hiding more controller on iPad
- selection prevention for regular tabs and more controller
- runtime changes to styling, including icon updates
- tab accessibility: tab ids and labels

  ## Checklist

  - [x] For API changes, updated relevant public types. *(no public API changes)*
  - [ ] Ensured that CI passes


FabricExample/e2e/single-feature-tests/tabs/test-tabs-system-item-ios.e2e.ts
```diff
@@ -1,5 +1,5 @@
 import { expect as jestExpect } from '@jest/globals';
-import { device, expect, element, by } from 'detox';
+import { device, expect, element, by, waitFor } from 'detox';
```
kkafar: `waitFor` seems unused

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -462,8 +475,137 @@ - (void)tabBarController:(UITabBarController *)tabBarController
...
+    /*
+     * `selectedTab` is stale while More is active (More has no `UITab`) - repainting now
+     * could yank the selection away from More.
+     */
+    UITab *_Nullable previouslySelectedTab = self.selectedTab;
+    UIViewController *_Nullable selectedViewController = self.selectedViewController;
+    if (selectedViewController != nil &&
+        ([self isViewControllerTheMoreNavigationController:selectedViewController] ||
+         [self isViewControllerHostedByMoreNavigationController:selectedViewController])) {
+      return;
```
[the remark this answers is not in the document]
kmichalikk: recheck this

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -462,8 +475,137 @@ - (void)tabBarController:(UITabBarController *)tabBarController
...
+
+  NSMutableArray<__kindof UITab *> *tabs = [NSMutableArray arrayWithCapacity:_tabScreenControllers.count];
+  for (RNSTabsScreenViewController *screenController in _tabScreenControllers) {
+    [tabs addObject:[self tabForTabScreenController:screenController]];
+  }
+  self.tabs = tabs;
+
+  if (previouslySelectedTab != nil && [tabs containsObject:previouslySelectedTab]) {
+    self.selectedTab = previouslySelectedTab;
+  }
```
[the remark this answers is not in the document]
kmichalikk: recheck this

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -462,8 +475,137 @@ - (void)tabBarController:(UITabBarController *)tabBarController
...
+
+  UIViewController *viewController = selectedTab.viewController;
+
+  /*
+   * Unlike the legacy delegate, this also fires for programmatic selection changes - those
+   * have already progressed the navigation state, so they match the selected screen key.
+   */
+  if (_navigationState == nil ||
+      [_navigationState.selectedScreenKey isEqualToString:[self screenKeyForViewController:viewController]]) {
+    return;
```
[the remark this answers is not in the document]
kmichalikk: recheck this

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -191,7 +203,28 @@ - (void)didMoveToParentViewController:(UIViewController *)parent
...
-  RNSLog(@"TabBar: %@ didSelectItem: %@", tabBar, item);
+#if RNS_MORE_NAVIGATION_CONTROLLER_AVAILABLE && RNS_IPHONE_OS_VERSION_AVAILABLE(18_0)
+  if (@available(iOS 26.1, *)) {
+    // The only direct "user tapped More" signal on the UITab path - no UITab delegate covers More.
+    // Mirrors the More branch of the legacy `shouldSelectViewController:`: enforce selection
+    // prevention on the More stack top before UIKit displays it.
+    if (self.tabs.count > 0 && [self isMoreNavigationControllerPresentInTabBar] &&
+        item == self.moreNavigationController.tabBarItem) {
+      [self prepareForMoreNavigationControllerHandlingIfNeeded];
+      [self disableNavigationBarInMoreNavigationController];
```
kmichalikk, line 214: Old api does this "ad hoc" in `userDidSelectViewController` but it's too late in the context of a new API. Verified empirically + new tab api doesn't get notified of the view controller is `(did/should)selectTab`

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -338,7 +327,7 @@ - (void)userDidRepeatViewControllerSelection:(nonnull UIViewController *)viewCon
     return;
   }
 
-  [self updateNavigationStateOnModelUpdate];
```
kmichalikk, line 341: this method was a single-line wrapper and not used in every instance, it's more readable and consistent without the indirection IMO

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -365,21 +354,26 @@ - (void)userDidSelectViewController:(nonnull UIViewController *)viewController
     // Instead, we emit a dedicated event so JS knows the More tab was tapped.
     [_observerRegistry emitDidSelectMoreTabWithCurrentState:_navigationState sender:self];
   } else {
-    [self updateNavigationStateOnModelUpdate];
-    auto *updateContext = [[RNSTabsNavigationStateUpdateContext alloc] initWithNavState:_navigationState
-                                                                             isRepeated:NO
-                                                              hasTriggeredSpecialEffect:NO
-                                                                           actionOrigin:RNSTabsActionOriginUser];
-    [_observerRegistry emitDidUpdateStateTo:_navigationState withContext:updateContext sender:self];
```
kmichalikk, line 373: this on the other hand I chose to extract into `emitUserSelectionUpdateWithOrigin`

ios/tabs/screen/RNSTabsScreenComponentView.mm
```diff
@@ -177,6 +177,14 @@ - (void)createTabBarItem
...
     tabBarItem = [[UITabBarItem alloc] init];
   }
+
+  if (@available(iOS 26.1, *)) {
+    // TODO: This is an ugly workaround and I would love to see it replaced.
+    // With UITab-managed children (iOS >= 26.1) a replacement item does not repaint first time.
+    // Assigning a throwaway item first flips the internal logic so that the real assignment
+    // below paints synchronously.
+    _controller.tabBarItem = [[UITabBarItem alloc] init];
+  }
```
kmichalikk: This is the only quick-and-not-overly-complicated code that fixes the problem with not updating system item. I'd argue to keep it in this PR. It's self-contained and the execution doesn't go here if we're not changing things at runtime + it doesn't make us change test cases + it makes hot reload still work.

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -462,7 +453,169 @@ - (void)tabBarController:(UITabBarController *)tabBarController
...
+
+/// Always builds a FRESH instance, seeded with current content - the bar reads a `UITab` only
+/// when its instance enters the tabs array.
+- (UITab *)makeTabForTabScreenController:(RNSTabsScreenViewController *)screenController API_AVAILABLE(ios(18.0))
+{
+  __weak RNSTabsScreenViewController *weakScreenController = screenController;
+
+  return [[UITab alloc] initWithTitle:screenController.title ?: @""
+                                image:nil
+                           identifier:[screenController.tabScreenComponentView screenKey]
```
kmichalikk, line 542: For now, identifier is just screenKey, but I'm thinking about exporting it in another PR

ios/tabs/screen/RNSTabsScreenComponentView.mm
```diff
@@ -177,13 +177,28 @@ - (void)createTabBarItem
   } else {
     tabBarItem = [[UITabBarItem alloc] init];
   }
+
+  [self applyTabBarItemRepaintWorkaround];
```
kmichalikk, line 181: This is the only quick-and-not-overly-complicated code that fixes the problem with not updating system item. I'd argue to keep it in this PR. It's self-contained and the execution doesn't go here if we're not changing things at runtime + it doesn't make us change test cases + it makes hot reload still work.

ios/tabs/screen/RNSTabsScreenComponentView.mm
```diff
@@ -200,9 +215,8 @@ - (void)updateTabBarItem
 
   [self updateTabBarItemTitle:evaluatedTitle];
 
-  if (![tabBarItem.badgeValue isEqualToString:_badgeValue]) {
-    tabBarItem.badgeValue = _badgeValue;
-  }
+  // The badge is missing here intentionally. It is applied by `RNSTabBarAppearanceCoordinator`
+  // to be correctly caught by UITab update.
```
kmichalikk, line 219: I'm setting it with appearance and it does seem to work, but rabbit makes good case that there is tab badgeValue. I don't know whether I should leave it as is or: reassiging badge value upon tab creation and then finding a tab to update the value here. Probably more logic to do this.

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -191,7 +203,28 @@ - (void)didMoveToParentViewController:(UIViewController *)parent
 
 - (void)tabBar:(UITabBar *)tabBar didSelectItem:(UITabBarItem *)item
 {
-  RNSLog(@"TabBar: %@ didSelectItem: %@", tabBar, item);
+#if RNS_MORE_NAVIGATION_CONTROLLER_AVAILABLE && RNS_IPHONE_OS_VERSION_AVAILABLE(18_0)
+  if (@available(iOS 26.1, *)) {
+    // The only direct "user tapped More" signal on the UITab path - no UITab delegate covers More.
+    // Mirrors the More branch of the legacy `shouldSelectViewController:`: enforce selection
+    // prevention on the More stack top before UIKit displays it.
+    if (self.tabs.count > 0 && [self isMoreNavigationControllerPresentInTabBar] &&
```
kkafar: What is the first part of the condition about: `self.tabs.count > 0`? Does it mean to check if we're using the UITab API?

If so -> I'd recommend we have it done differently. We should create a *single decision point* / *source of truth* of whether we rely on UITab or UIViewController based API, e.g. `if ([self usesUITabApi])` and in every place were we need to decide whether the new API is in use or not - we should rely on value returned by that method.

If not -> I think then that you assume that `availbility` of iOS 26.1 implies usage of the new API -> then I'd like the `@available(iOS 26.1, *)` part to also be extracted into compile time macro, so the condition communicates clearly the intention behind it.

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -191,7 +203,28 @@ - (void)didMoveToParentViewController:(UIViewController *)parent
 
 - (void)tabBar:(UITabBar *)tabBar didSelectItem:(UITabBarItem *)item
 {
-  RNSLog(@"TabBar: %@ didSelectItem: %@", tabBar, item);
+#if RNS_MORE_NAVIGATION_CONTROLLER_AVAILABLE && RNS_IPHONE_OS_VERSION_AVAILABLE(18_0)
+  if (@available(iOS 26.1, *)) {
```
kkafar: Can we extract 26.1 here into an compile time macro (with `#define`)? So that we can decide from a single place when the new API is activated?

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -191,7 +203,28 @@ - (void)didMoveToParentViewController:(UIViewController *)parent
...
+      UIViewController *_Nullable poppedViewController =
+          [self popToRootInMoreNavigationControllerRespectSelectionPrevention:YES animated:NO];
+      if (poppedViewController != nil) {
+        [self
+            onDidPreventUserFromSelectingViewControllerWithKey:[self screenKeyForViewController:poppedViewController]];
+      }
+
+      if (![self isSelectedViewControllerTheMoreNavigationController]) {
+        [_observerRegistry emitDidSelectMoreTabWithCurrentState:_navigationState sender:self];
+      }
```
kkafar: I do not understand this condition. 

I thought that if the UITab API is used, then the `UITabBarController.selectedViewController` field is not kept up to date, is it? 
Will it be updated to point to `moveNavigationController` in case the "more controller UITab" is selected and present in `UITabBarController.selectedTab`?

ios/tabs/host/RNSTabBarController.mm
kkafar: You're right to the fact that it was a single-line wrapper. Please note, however,  that it's name, together with it's doc comment, carried additional information you now lose, when you remove this wrapper. 
It communicated that you progress the *navigation state* due to previous update of the UITabBarController's underlying model. This was a counter part to the `updateSelectedViewControllerTo:withKey:actionOrigin:`, which went other way around - it updated the UIKit model in response to *navigation state* change.

In the state you propose - we have only a bare call to navigation state update, forfeiting all this additional semantics. 

I'll insist here to restore that to the original state, because I think that the additional semantics information is useful, and we don't achieve much, beside losing it, when removing this one layer of indirection.

If you find this kind of debate interesting, I highly recommend you watch this video by Logan Smith: https://www.youtube.com/watch?v=2OMRWPOSw9s

ios/tabs/host/RNSTabBarController.mm
kkafar: I'm fine with extracting the event creation & call to observer creation into a separate method - this is a good idea. 

I'd suggest the `updateNavigationOnModelUpdate` call to be restored though. I've given my reasoning in some other comment earlier.

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -462,7 +453,169 @@ - (void)tabBarController:(UITabBarController *)tabBarController
...
+  [self setSelectedViewController:screenController];
+}
+
+- (UITab *)tabForTabScreenController:(RNSTabsScreenViewController *)screenController API_AVAILABLE(ios(18.0))
+{
+  for (UITab *tab in self.tabs) {
+    if (tab.viewController == screenController) {
+      return tab;
+    }
+  }
```
kkafar: Why not just query `screenController.tab`?

ios/tabs/host/RNSTabBarController.mm
kkafar: What do you mean by "exporting it"?

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -462,7 +453,169 @@ - (void)tabBarController:(UITabBarController *)tabBarController
...
+
+  return [self makeTabForTabScreenController:screenController];
+}
```
kkafar: Shouldn't all of the methods that use UITab & other symbols added in iOS 18 be compiled out, if the SDK is not available? My question is: will this code build from XCode when the SDK 18 or newer is not available? 

I might be wrong here, but I believe that these methods should be surrounded with ifdefs.

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -462,7 +453,169 @@ - (void)tabBarController:(UITabBarController *)tabBarController
...
+
+- (void)applySelectedScreenController:(nonnull UIViewController *)screenController
+{
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_1)
+  if (@available(iOS 26.1, *)) {
+    if (![self isMoreNavigationControllerTabBarItemSelected]) {
+      UITab *tab = [self findTabForScreenKey:[self screenKeyForViewController:screenController]];
+      RCTAssert(tab != nil,
+                @"[RNScreens] No installed UITab for screenKey: %@",
+                [self screenKeyForViewController:screenController]);
```
kkafar: Do we need this whole search machinery here? If you have screenController, it just has plain `UIViewController.tab` property, which you can query. 

Seems like an overkill here.

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -462,7 +453,169 @@ - (void)tabBarController:(UITabBarController *)tabBarController
...
+  BOOL repeatedSelectionHandledBySpecialEffect =
+      [static_cast<RNSTabsScreenViewController *>(tab.viewController) tabScreenSelectedRepeatedly];
+
+  auto *updateContext =
+      [[RNSTabsNavigationStateUpdateContext alloc] initWithNavState:_navigationState
+                                                         isRepeated:YES
+                                          hasTriggeredSpecialEffect:repeatedSelectionHandledBySpecialEffect
+                                                       actionOrigin:RNSTabsActionOriginUser];
+  [_observerRegistry emitDidUpdateStateTo:_navigationState withContext:updateContext sender:self];
+}
```
kkafar: this method is not a part of UITabBarController delegate - it should not be in this code section.

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -462,7 +453,169 @@ - (void)tabBarController:(UITabBarController *)tabBarController
...
+
+  return [[UITab alloc] initWithTitle:screenController.title ?: @""
+                                image:nil
+                           identifier:[screenController.tabScreenComponentView screenKey]
+               viewControllerProvider:^UIViewController *(UITab *) {
+                 return weakScreenController;
+               }];
+}
+
+#pragma mark UITabBarControllerDelegate (UITab API)
```
kkafar: I like the fact of separation of one type of API of UITabBarControllerDelegate from another - but let's have these two code sections colocated in the source file.

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -462,7 +453,169 @@ - (void)tabBarController:(UITabBarController *)tabBarController
...
+
+  BOOL repeatedSelectionHandledBySpecialEffect =
+      [static_cast<RNSTabsScreenViewController *>(tab.viewController) tabScreenSelectedRepeatedly];
+
+  auto *updateContext =
+      [[RNSTabsNavigationStateUpdateContext alloc] initWithNavState:_navigationState
+                                                         isRepeated:YES
+                                          hasTriggeredSpecialEffect:repeatedSelectionHandledBySpecialEffect
+                                                       actionOrigin:RNSTabsActionOriginUser];
+  [_observerRegistry emitDidUpdateStateTo:_navigationState withContext:updateContext sender:self];
```
kkafar: You might also move this to a helper method, similarly as what you have done in the other place I commented on earlier.

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -462,7 +453,169 @@ - (void)tabBarController:(UITabBarController *)tabBarController
...
+  }
+  _isHandlingUserTabSelection = NO;
+
+  // Counterpart of `userDidSelectViewController:`. Asserts stay in the synchronously-mutated
+  // tab domain - `selectedViewController` may still lag mid-transition (iOS 27).
+  RCTAssert(self.selectedTab == selectedTab, @"[RNScreens] Expected UIKit to update selectedTab");
+  [self progressNavigationState:[self screenKeyForViewController:selectedTab.viewController]
+                     withOrigin:RNSTabsActionOriginUser];
+  [self emitSelectionUpdateWithOrigin:RNSTabsActionOriginUser];
+  _isHandlingExplicitSelectionUpdate = NO;
```
kkafar: I'd be good to delegate both paths (UITab and UIViewController one) to the same `userDidSelectTab/ViewController` method. This avoid duplication and we'll have a single place to maintain. At least for now, it seems that these methods can be easily kept in sync. Please not, that if you create the `- [RNSTabBarController usesUITabApi]` method I've suggested somewhere above - you'll have easy time in adjusting the `- [RNSTabBarController userDidSelectViewController:]` method.

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -553,45 +705,43 @@ - (void)updateSelectedViewControllerInner
...
   }
 
-  if (currSelectedViewController == nextSelectedViewController && _navigationState != nil) {
-    // Nothing to do, we don't allow for programmatic repeat selection, unless
-    // we're during first render.
+  // While More is active `self.selectedViewController` is the More controller itself - a matching
+  // request is a repeat only when it targets what the More stack currently tops.
+  BOOL isRepeatedSelection = [self isMoreNavigationControllerTabBarItemSelected]
+      ? nextSelectedViewController == [self resolveMoreNavigationController].topViewController
+      : nextSelectedViewController == self.selectedViewController;
```
kkafar: I have similar remarks as above - is it safe to use the `self.selectedViewController` here?

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -553,45 +705,43 @@ - (void)updateSelectedViewControllerInner
...
   } else {
-    self.tabBar.overrideUserInterfaceStyle = screenViewController.tabScreenComponentView.userInterfaceStyle;
+    self.tabBar.overrideUserInterfaceStyle = nextSelectedViewController.tabScreenComponentView.userInterfaceStyle;
   }
 
   RNSLog(@"Change selected view controller to: %@", nextSelectedViewControllerKey);
-  BOOL hasStateProgressed = [self updateSelectedViewControllerTo:nextSelectedViewController
-                                                         withKey:nextSelectedViewControllerKey
-                                                    actionOrigin:_pendingStateUpdate.actionOrigin];
+  [self progressNavigationState:nextSelectedViewControllerKey withOrigin:_pendingStateUpdate.actionOrigin];
```
kkafar: Similar remarks to `updateNavigationStateOnModelUpdate` here.

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -614,8 +764,8 @@ - (void)updateTabBarAppearance
 
 - (void)updateTabBarA11yIfNeeded
 {
-  for (UIViewController *tabViewController in self.viewControllers) {
-    auto screenView = static_cast<RNSTabsScreenViewController *>(tabViewController).tabScreenComponentView;
+  for (RNSTabsScreenViewController *tabViewController in _tabScreenControllers) {
```
kkafar: Why do we iterate over _tabScreenControllers here? The old code iterated only over the "installed" view controllers? Why the change in semantics here?

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -686,6 +832,23 @@ - (nonnull NSString *)screenKeyForViewController:(nonnull UIViewController *)vie
...
+- (nullable UITab *)findTabForScreenKey:(nullable NSString *)screenKey API_AVAILABLE(ios(18.0))
+{
+  if (screenKey == nil) {
+    return nil;
+  }
+  for (UITab *tab in self.tabs) {
+    if ([[self screenKeyForViewController:tab.viewController] isEqualToString:screenKey]) {
+      return tab;
+    }
+  }
```
kkafar: You can just look at tab's identifier. 

I can suggest here creating a category / extension on `UITab` class with a screenKey / navigationKey method on it, so that the semantics are clear when used.

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -769,7 +928,7 @@ - (BOOL)canHaveMoreNavigationController
   // https://developer.apple.com/documentation/uikit/uitabbarcontroller?language=objc#The-More-navigation-controller
   // The count is documented. Size class check is empirical, to tighten the condition and have less
   // false positives. If we ever find it not correct, we can safely remove it.
-  return self.viewControllers.count >= kMinCountOfVCsForMoreVCPresence &&
+  return _tabScreenControllers.count >= kMinCountOfVCsForMoreVCPresence &&
```
kkafar: also change in semantics here? Deliberate?

ios/tabs/host/RNSTabBarController.mm
```diff
@@ -783,7 +942,7 @@ - (BOOL)isViewControllerHostedByMoreNavigationController:(nonnull UIViewControll
   }
 
   // Guard: VC must be one we manage (excludes arbitrary external VCs).
-  if ([self.viewControllers indexOfObject:viewController] == NSNotFound) {
+  if (![_tabScreenControllers containsObject:static_cast<RNSTabsScreenViewController *>(viewController)]) {
```
kkafar: same here

ios/tabs/screen/RNSTabsScreenComponentView.mm
```diff
@@ -177,13 +177,28 @@ - (void)createTabBarItem
...
+  [self applyTabBarItemRepaintWorkaround];
   _controller.tabBarItem = tabBarItem;
 }
 
-- (void)updateTabBarItem
+/**
+ * TODO: This is an ugly workaround and I would love to see it replaced.
+ * With UITab-managed children (iOS >= 26.1) a replacement item does not repaint first time.
+ * Assigning a throwaway item first flips the internal logic so that the real assignment
+ * that follows paints synchronously. Remove once UIKit internals no longer require it.
```
kkafar: I don't think I understand exactly what the problem was from this description. 

What does it mean for the tab bar item to "repaint first time"?

You mean it's appearance configuration changes and that is not reflected on the screen?

ios/tabs/screen/RNSTabsScreenComponentView.mm
kkafar: So that method updates the title only right now. Would it also make sense to move this to appearance coordinator? 

I guess the badge value & text title are not appearance per se, therefore they were not a part of appearance coordinator logic. Now, when we just move the badgeValue update from here we:

1. split the "model update" into two places, title updated here and badge value updated somewhere else,
2. put "model update" into appearance update. 

I think that we either create "model coordinator" that will fire after appearance coordinator, or we consider renaming appearance coordinator to configuration coordinator or something like this. I'm open to suggestions here.

kkafar (changes requested):
Looks good overall, thank you. 
I have a series of remarks, though -> let's resolve them before we proceed. Leaving them below. 

> This PR brings a MAJOR change to tabs and MUST be carefully reviewed and tested on all possible scenarios.

If this PR does not contain breaking changes please use different wording. Breaking changes are only MAJOR-grade changes.

ios/tabs/host/RNSTabBarController.mm
kmichalikk: I mean adding a prop to configure it

ios/tabs/host/RNSTabBarController.mm
kmichalikk: https://github.com/software-mansion/react-native-screens/pull/4585/commits/1f8362dd93a946124276d1a76da65a56158a1e2c

ios/tabs/host/RNSTabBarController.mm
kmichalikk: https://github.com/software-mansion/react-native-screens/pull/4675/commits/b2c7f8a05daf0ef575ddef24bb63092252322e8c

ios/tabs/host/RNSTabBarController.mm
kmichalikk: https://github.com/software-mansion/react-native-screens/pull/4675/commits/b2c7f8a05daf0ef575ddef24bb63092252322e8c

ios/tabs/host/RNSTabBarController.mm
kmichalikk: https://github.com/software-mansion/react-native-screens/pull/4675/commits/b2c7f8a05daf0ef575ddef24bb63092252322e8c

ios/tabs/host/RNSTabBarController.mm
kmichalikk: https://github.com/software-mansion/react-native-screens/pull/4675/commits/1f8362dd93a946124276d1a76da65a56158a1e2c

ios/tabs/host/RNSTabBarController.mm
kmichalikk: https://github.com/software-mansion/react-native-screens/commit/1f8362dd93a946124276d1a76da65a56158a1e2c

ios/tabs/host/RNSTabBarController.mm
kmichalikk: https://github.com/software-mansion/react-native-screens/commit/70f4aa72842032a6052a5d9ca04b72c57394fd59

ios/tabs/host/RNSTabBarController.mm
kmichalikk: https://github.com/software-mansion/react-native-screens/commit/aaeb0b69bc8f1a7038f758bcdefb684be294b27c

this code was placed in a different method that ran for both paths, so this was a check for tab api, yes, but now this code is wrapped with availability check so here it's unneeded

ios/tabs/host/RNSTabBarController.mm
kmichalikk: https://github.com/software-mansion/react-native-screens/commit/58ff129b18c99192ada47f64917c9151b03ff7ad

Added a comment, looked for a better place to call that but haven't found one. The `didSelectTab:` doesn't fire for More tab.

ios/tabs/host/RNSTabBarController.mm
kmichalikk: I was going to argue that the shape of the function and flow is too different, and crucially the  `userDidSelectViewController` asserts that the `selectedViewController` has settled, which it didn't when I last checked. I tried to show the exact flow that crashes, but it didn't crash.... I looked for 2h in reflogs and agent memories and haven't found the commit that crashed (I swear it did, I found that on the second PR with UISearchTab and I needed to adjust here just for that find), so I'm in a point where I need to rethink this. It may work as you described, I need more time to figure it out.

PR #4676, opened 2026-09-18
chore(iOS, FabricExample): use scene lifecycle by default 

kligarski:
## Description

`FabricExample` still booted through the legacy `UIApplicationDelegate` lifecycle unless you explicitly built with `RNS_USE_SCENE_DELEGATE=1`. Apple deprecated that lifecycle in iOS 26 and starting from iOS 27, the app crashes without the new scene lifecycle. This PR flips the example app to use the scene delegate by default.

The legacy path stays reachable — `RNS_USE_SCENE_DELEGATE=0` (or `false`) restores it, which is useful for reproducing issues coming from apps that have not adopted scenes yet.

### Detox and the scene lifecycle

The first version of this PR broke iOS e2e: every spec calling `device.reloadReactNative()` crashed with `Invalid AppDelegate: Could not access rootViewFactory`. On the new architecture Detox reloads by reading `rootViewFactory` / `reactNativeFactory` from `UIApplication.sharedApplication.delegate`. Our `SceneDelegate` owned the factory and `AppDelegate` was empty, so Detox found nothing.

There are two upstream conventions for where the factory lives:

- **React Native 0.88+ / community template** — `SceneDelegate` owns `reactNativeFactory` and calls the new `startReactNative(withModuleName:in:connectionOptions:)`; `AppDelegate` is an empty shell. This won't work with current version of Detox. **We should revisit this with the RN 0.88 bump.**
- **Expo** — `AppDelegate` creates and stores the factory in `didFinishLaunching`; the scene delegate fetches it, creates the window and starts React Native (`ExpoReactNativeFactoryProvider`, `ExpoAppSceneDelegate`). Expo's own Detox-tested `bare-expo` app uses this layout.

Since the React Native direction is `SceneDelegate` owning the factory, this PR keeps that layout and only adds a read-only `@objc var reactNativeFactory` proxy on `AppDelegate` that resolves the factory from the foreground-active connected scene's `SceneDelegate`. Detox reads it via KVC, so `reloadReactNative` works again, and nothing moves back to the app delegate. Multiple scenes (iPad multi-window) keep one factory per scene, as before. This is a temporary solution until Detox adds support for RN 0.88 and scene lifecycle.

## Changes

- `FabricExample/ios/Podfile`: replaced the `ENV['RNS_USE_SCENE_DELEGATE'] == '1'` opt-in with `rns_scene_delegate_enabled?`, which defaults to `true` and only opts out on `0`/`false`.
- `FabricExample/ios/FabricExample/Info.plist`: committed the `UIApplicationSceneManifest` entry so a fresh clone uses `SceneDelegate` without needing `pod install` to inject it.
- `FabricExample/ios/FabricExample.xcodeproj/project.pbxproj`: added `RNS_USE_SCENE_DELEGATE` to `SWIFT_ACTIVE_COMPILATION_CONDITIONS` for both Debug and Release.
- `FabricExample/ios/AppDelegate.swift`: the scene branch exposes a read-only `@objc var reactNativeFactory` proxy to the active scene's factory, for Detox compatibility.
- `FabricExample/ios/Podfile.lock`: refreshed `PODFILE CHECKSUM`.

## Test plan

1. `cd FabricExample/ios && bundle exec pod install` — `Info.plist` keeps the scene manifest and the `RNS_USE_SCENE_DELEGATE` flag stays set in the project.
2. Build & run `FabricExample` on iOS (Debug and Release). The app comes up through `SceneDelegate.swift` (`AppDelegate.swift` is compiled out by `#if !RNS_USE_SCENE_DELEGATE`).
3. `RNS_USE_SCENE_DELEGATE=0 bundle exec pod install` — the scene manifest is removed from `Info.plist` and the compilation flag is unset; the app then boots through `AppDelegate.swift` as before.
4. `cd FabricExample && yarn build-e2e-ios && yarn test-e2e-ios` — specs using `device.reloadReactNative()` pass.
5. Make sure that CI passes (build & e2e).

## Checklist

- [x] Ensured that CI passes

---


kligarski:
Let's see if the e2e passes or some changes are necessary.

kmichalikk (approved):

kligarski:
Moving to draft until we decide how to handle Detox e2e.

kmichalikk (approved):
The sorting in AppDelegate looks weird, you may leave some explanation. Otherwise GTG

kligarski:
> The sorting in AppDelegate looks weird, you may leave some explanation.

https://github.com/software-mansion/react-native-screens/pull/4676/commits/079331dc5f83fcb2e5992a255fcfe85ea1f4d30f

kkafar (commented):
Can we make it so when running detox build it'll use the old setup? I'd like to have that working on iOS 27 nicely in day-to-day development. 

@LKuchno I think you should be aware of these problems, since you try right now to get our CI working on iOS 27. Get in touch with @kligarski.

kligarski:
> Can we make it so when running detox build it'll use the old setup? I'd like to have that working on iOS 27 nicely in day-to-day development.

After adding `reactNativeDelegate` getter to AppDelegate, Detox works correctly.

PR #4678, opened 2026-09-18
feat(iOS): add `visibilityPriority` prop for header items 

Loloekk:
## Description

This PR adds support for [UIBarButtonItemVisibilityPriority](https://developer.apple.com/documentation/uikit/uibarbuttonitemvisibilitypriority?language=objc)
through a new `visibilityPriority` prop, available on iOS 27 and higher.

It lets you set the priority of header items. The priority is either one of the
strings `'low'`, `'standard'` (the default) or `'high'`, which map to the
matching UIKit constants, or a raw number. When the items no longer fit in the
header, the ones with the lowest priority are moved into the overflow menu.

Apple does not document the numeric values behind the named constants, so the
strings are resolved to the UIKit constants natively instead of being mapped to
numbers in JS.

## Changes

- Add the `visibilityPriority` prop and pass it to `barButtonItem.visibilityPriority` on the native side,
- Add the `Stack Header Item Visibility Priority (iOS)` single feature test.

## Visual documentation

https://github.com/user-attachments/assets/8c388147-36b3-4b86-8d3b-4c502f1c1ae1

## Test plan

Run the 'Stack Header Item Visibility Priority (iOS)' single feature test on an iOS 27 Pro Max

## Checklist

- [ ] Ensured that CI passes


apps/src/tests/single-feature-tests/stack-v5/test-stack-header-item-visibility-priority-ios/scenario.md
```diff
@@ -0,0 +1,56 @@
...
+    - [ ] The two icons are listed in the menu
+6. Press the "search" to hide the menu
+    - [ ] The resizing item is in the header, at its full width, and both
+        icons are in the overflow menu
+7. Press the button to reach `low`
+    - [ ] The resizing item disappears, both icons stay in the header
+8. Press the button to reach `1000`
+    - [ ] Same as `high`
+9. Press the resizing item in the header
+    - [ ] Same as 1, except the button still reads `visibilityPriority: 1000`
```
kmichalikk: Don't write same as 1, please write the expected outcome.

apps/src/tests/single-feature-tests/stack-v5/test-stack-header-item-visibility-priority-ios/scenario.md
```diff
@@ -0,0 +1,56 @@
...
+10. Press the button to reach `standard`
+    - [ ] Same as 1
+
+## Steps on iPhone (iOS 26)
+
+11. Press the resizing item 
+    - [ ] The priority is ignored (the property does not exist before iOS 27).
+        After you press the resizing item, it disappears and does not come back
+        for any of the priorities.
```
kmichalikk: Unneeded, leave a note that this is only supported on iOS >= 27

ios/stack/header/RNSStackHeaderItemComponentView.mm
```diff
@@ -17,6 +17,42 @@
...
+{
+  if (value == nil) {
+    return nil;
+  }
+
+#if RNS_IPHONE_OS_VERSION_AVAILABLE(27_0)
+  if (@available(iOS 27.0, *)) {
+    if ([value isKindOfClass:NSNumber.class]) {
+      return @([(NSNumber *)value integerValue]);
+    }
```
kmichalikk: Please research what this is: https://developer.apple.com/documentation/uikit/uibarbuttonitemvisibilitypriority/init(rawvalue:)

could we use it instead and return the actual visibility priority type from this function?

ios/stack/header/RNSStackHeaderItemComponentView.mm
```diff
@@ -17,6 +17,42 @@
 
 namespace react = facebook::react;
 
+/**
+ * Resolves the `visibilityPriority` prop, which is either one of the named
+ * values or a raw number, into a `UIBarButtonItemVisibilityPriority` value.
+ * The names are resolved here rather than in JS, so that they keep pointing at
+ * whatever the UIKit constants hold.
+ */
+static NSNumber *_Nullable RNSVisibilityPriorityFromPropValue(id _Nullable value)
```
kmichalikk: nit: Please look for a better place for this function, if there is any, maybe some utils file

PR #4679, opened 2026-09-18
feat(iOS, Tabs): Use `UISearchTab` with `automaticallyActivatesSearch` prop

### LINKED ISSUE #4671: [iOS 27 SDK] role="search" loses the detached search-tab chrome; the legacy UITabBarSystemItemSearch dress-up only happens under the iOS 26 SDK
### Description

`<NativeTabs.Trigger role="search">` renders as a detached search tab when the app is built against the iOS 26 SDK. Rebuild the same source against the iOS 27 SDK and the tab collapses back into the main tab bar as an ordinary tab, with its label shown.

The device OS is identical in both cases. Only the SDK the binary is linked against changes.

This is the visual half of what #3999 / discussion #4000 described. That report noted, correctly, that "iOS 26 dresses up the legacy system item with the new chrome automatically, which is why the visual is right." Under the iOS 27 SDK that automatic dress-up no longer happens, so the visual is now wrong too, not just the auto-activation behavior.

**Why it happens**

`RCTConvert+RNSTabs.mm` maps `'search'` to `RNSTabsScreenSystemItemSearch`, and `RNSConversions-Tabs.mm` converts that to `UITabBarSystemItemSearch`. `RNSTabsScreenComponentView.mm` then builds the item with `[[UITabBarItem alloc] initWithTabBarSystemItem:tag:]`.

`UISearchTab` is never instantiated. I checked 4.26.2 and 4.28.0; neither references `UITab`, `UISearchTab` or `UITabRole` anywhere under `ios/`. Under the iOS 26 SDK that did not matter for appearance. Under the iOS 27 SDK it does.

**Evidence**

Two binaries of the same app, read with `xcrun vtool -show-build-version`:

| Build | platform | minos | sdk | search tab |
| Xcode 26.6 (EAS image `macos-tahoe-26.5-xcode-26.6`) | IOS | 16.4 | **26.5** | detached |
| Xcode 27.0 (27A266a), local | IOSSIMULATOR | 16.4 | **27.0** | inside the bar |

**Impact**

An app that adopted `role="search"` on the documented expo-router path gets a different tab bar the first time it is rebuilt with a newer Xcode, with no source change and nothing in the build output saying so.

**What would fix it**

Instantiate `UISearchTab` for `role="search"` on iOS 26+, as discussion #4000 already proposes for `automaticallyActivatesSearch`. Under the iOS 27 SDK that is now the only way to get the appearance as well as the behavior.

### Steps to reproduce

1. Build an app with four `NativeTabs.Trigger`s, one carrying `role="search"`.
2. Build it with Xcode 26 and run it on an iOS 27 device or simulator. The search tab is detached from the main tab bar.
3. Build the same source with Xcode 27 and run it on the same iOS 27 device. The search tab is now inside the main tab bar.

A Snack cannot express this, because the variable is the Xcode version the binary is linked against, not the JavaScript. The reproduction is any app using the documented separate-search-tab setup, built twice from one commit with two Xcode versions.

I have not published a minimal repository. If that blocks triage, say so and I will put one together; I wanted to report the SDK dependency first, since `vtool` on the two binaries already isolates it and the affected code path is a handful of lines.

One honest caveat: between my two builds the Expo patch versions also moved (expo 57.0.16 to 57.0.22, react-native 0.86.2 to 0.86.3). I inspected the expo-router diff for `native-tabs` across those versions and the only change is icon rendering-mode resolution; `role` still reaches `systemItem` unchanged, and react-native-screens is 4.26.2 in both. So the SDK is the remaining explanation, but I have not yet rebuilt one commit against both SDKs to prove it by construction.

### Screens version

4.26.2 (also inspected 4.28.0; same code path)

### React Native version

0.86.3

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo (development build and EAS Build)

### Build type

Debug mode with a development bundle (local) and release build (TestFlight)

### Device

Real device and iOS simulator

### Device model

iPhone on iOS 27, and iPhone 18 Pro simulator on iOS 27

### Acknowledgements

- [x] I have searched for existing issues and pull requests.

kmichalikk:
Closes https://github.com/software-mansion/react-native-screens/issues/4671

## Checklist

- [ ] Ensured that CI passes


src/components/tabs/screen/TabsScreen.ios.types.ts
```diff
@@ -271,11 +271,34 @@ export interface TabsScreenPropsIOS {
...
+   * screen's lifetime. Changing `systemItem` to or from `'search'` on a mounted tab screen is
+   * therefore not supported - it triggers an assertion in development builds. To change the
+   * search role at runtime, remount the tab screen (e.g. by changing its React `key`).
+   * Changes between the other system item values are applied dynamically.
+   *
    * @see {@link https://developer.apple.com/documentation/uikit/uitabbaritem/systemitem|UITabBarItem.SystemItem}
    *
    * @platform ios
    */
   systemItem?: TabsScreenSystemItem | undefined;
```
satya164, line 285: correct me if i'm wrong, but aren't we overloading `systemItem` meaning? as i understand, `UITabBarItem.SystemItem` is for choosing the built-in localized icon and label for a tab based on predefined sets. so changing the class to something else doesn't fit with native UIKit semantics.

imo we should have a separate prop to explicitly choose to use `UISearchTab`, or `role: "search"` to match SwiftUI behavior with an explicit prop.

it'd also be nice to expose the new `prominentTabIdentifier` which is exposed in SwiftUI as `role: "prominent"`

kmichalikk: In 26, this specific role made the item go to the side, so I thought this would make sense to keep + the linked issue leans towards it, but I'm open to sth else once I get the core to work

prominent is blocked by the UITab API rework, but should be done shortly after that, the API for that will probably be different than role=prominent, but that for another discussion

satya164: > In 26, this specific role made the item go to the side

yes, but this is iOS behavior, so UIKit apps would also need changes. previously there have been arguments towards exposing native behavior directly in screens and let the integration layer handle platform and version differences, so using a UIKit property name to expose custom behavior in screens itself would be non-characteristic imo.

we expose this prop as `tabBarSystemItem` in React Navigation, so I'd prefer to avoid confusion of using a UIKit API for the name while having different behavior.

a separate prop like `role: 'search' | 'prominent'` to mirror the similar SwiftUI API, or some other explicit prop would be better.

cc @kkafar

apps/src/tests/single-feature-tests/tabs/test-tabs-search-tab-activation-ios/scenario.md
```diff
@@ -0,0 +1,80 @@
+# Test Scenario: Search Tab Activation (iOS)
```
kmichalikk, line 1: Should this be an integration test instead?

PR #4703, opened 2026-09-21
feat(iOS, Stack v5):  add `minimizationBehavior` and `restorationBehavior` to control when the navigation bar minimizes and restores on scroll

Loloekk:
…eaderConfig

## Checklist

- [ ] Ensured that CI passes


apps/src/tests/single-feature-tests/stack-v5/test-stack-header-bar-minimization-ios/scenario.md
```diff
@@ -0,0 +1,85 @@
...
+## Note
+
+On iOS < 27 the props are ignored and a warning is logged for any value other
+than `automatic`.
+
+`restorationBehavior: atScrollEdge` is honored by the system only together with
+`minimizationBehavior: onScrollDown`. With other behaviors it falls back to
+`automatic`.
+
+## Steps
```
kmichalikk, line 28: I would like to have a header item present and verify what happens with it when scrolled

kmichalikk (commented):
Looks sensible. However, when I tested on 27.1 (Duo, please verify on regular ones), the automatic behaves differently than in the scenario -- please adjust the steps if that's the case, I thing what shipped in .1 would be the way forward.

apps/src/tests/single-feature-tests/stack-v5/test-stack-header-bar-minimization-ios/scenario.md
Loloekk: I've added the header items in [a862bb7](https://github.com/software-mansion/react-native-screens/pull/4703/commits/a862bb7a5810fc33aeaada432476fd4504538834). 

Regarding the behavior: Version 27.1 is unavailable on standard iPhones. I tested it on version 27.2, and the `automatic` behavior works as `never`. I'm not sure whether the discrepancy you noticed is related to the Duo device or the iOS version.

